PR #6523, opened 2020-07-22
fix(grid): update left and right gutter class names to start and end

### LINKED ISSUE #6510: Add support for narrow grid mode and update gutter classes
We should implement the [narrow grid mode](https://www.carbondesignsystem.com/guidelines/2x-grid/implementation#the-narrow-grid) in a similar way to our condensed grid mode. We should also update our gutter classes from `left` and `right` to `start` and `end` which are flipped for RTL.

alisonjoseph:
Closes #6510

Updates left/right grid class names to start and end

#### Changelog

**New**

- start / end classnames for grid gutter and hang

#### Testing / Reviewing

Ensure nothing is broken. 

Confirm this is the correct way to comment that the left/right should be deprecated??


tw15egan (approved):
LGTM 👍 ✅

tw15egan:
@alisonjoseph seems like it's pulling in the `narrow` code from the other PR?

alisonjoseph:
Oops, I think I fixed it. 🤞

packages/grid/scss/_mixins.scss
```diff
@@ -159,7 +159,7 @@
 // No gutter
 // -----------------------------------------------------------------------------
 
-/// Add `no-gutter` and `no-gutter--{left,right}` classes to the output CSS. These
+/// Add `no-gutter` and `no-gutter--{start,end}` classes to the output CSS. These
```
joshblack, line 162: I think we'll need to keep the previous classes as well otherwise it'll be a breaking change, maybe with a note to remove in v11?

packages/grid/scss/_mixins.scss
```diff
@@ -185,12 +196,21 @@
 // Hang
 // -----------------------------------------------------------------------------
 
-/// Add `hang--left` and `hang--right` classes for a given gutter. These classes are
-/// used alongside `no-gutter--left` and `no-gutter--right` to "hang" type.
+/// Add `hang--start` and `hang--end` classes for a given gutter. These classes are
```
joshblack, line 199: Same note for this, I think

joshblack (changes requested):

joshblack (approved):
Ignore me! Just was a comment change

Issue #6067, opened 2020-05-13
Sort icon on table headers in poor position (especially bad for wide columns)


Solly2:
Sort icon on table headers in poor position (especially bad for wide columns)

On sortable columns the sort icon sits aligned on the hard right ofthe cell.  This means that it is closer to the column to the right of the one which is actually being used for the sort.
![02b](https://user-images.githubusercontent.com/45822417/81813868-e82cdb00-951f-11ea-8fef-cedd97dd2181.png)

This can obviously lead to some usability issues with the user not being able to easily see which column is providing the sort.

As a fix I suggest either moving the icon to 16px to the right of the column header
![02b copy 2](https://user-images.githubusercontent.com/45822417/81814929-46a68900-9521-11ea-83cb-c37e19adeaea.png)

Or leaving the hover state on columns which have been sorted
![02b copy](https://user-images.githubusercontent.com/45822417/81815054-70f84680-9521-11ea-9145-9fdea5d92cd5.png)

aagonzales:
The sorted data table row is actually supposed to keep the darker background color when applied. IBM Cloud brought this up a few months ago and I thought they were going to contribute the change but looks like it hasn't happened.

In general, the aligned is this way to reflect the IBM Language intent. Icons always align to the right side of container they are in. It is done like this through out the system. 

![image](https://user-images.githubusercontent.com/11670886/88109972-042ca880-cb71-11ea-88b9-1dfbf824cd05.png)

tw15egan:
@aagonzales I'll put in a quick fix for the active sorting column so that it retains a selected background, but going to close the issue as won't fix re: icons

PR #6524, opened 2020-07-22
fix(DataTable): retain active state when sort is active

tw15egan:
Refs #6067 

Adds background to Data Table headers that are actively sorting.

#### Changelog

**New**

- `.bx--table-sort--active`  now has a background of `$data-table-column-hover`

#### Testing / Reviewing

Ensure active sort headers retain a background color when a column is sorted and focus moves away


tw15egan:
@aagonzales wasn't sure if there was a different token we wanted to use here instead

aagonzales:
@tw15egan I think that's fine for now. Data table goes into a layering of colors that we didn't expect originally. We can hopefully resolve this to a better token in the future token work.

aagonzales (approved):
The background color looks good. The icon is doing something weird though. (Though it looks like this is happening in the normal dev environment too, so I'll go ahead and approve). 

image

aledavila (approved):

joshblack (approved):

PR #6549, opened 2020-07-24
fix(search): change fill to inherit

### LINKED ISSUE #6432: Using fill="currentColor" for icons causes a regression for certain components in dark themes
Copying over from a report by @SimonFinney in https://github.com/carbon-design-system/carbon/issues/5531

---

This appears to cause regression in dark themes in cases where the icon is dependent on a `fill` from its parent - here's an example of the `Search` component:

Default

This is resolved by removing the `currentColor` on the icon as per the original behavior: 

currentColor removed

But also if the `color` of the parent is set to the `fill` value:

color

We can use this as a workaround for now, but I wanted to raise it for awareness as I suspect this may be a silent issue reoccurring elsewhere and could easily be overlooked on lighter themes.

aledavila:
Closes #6432 

Worked on this with @andreancardona 
Fill color for search close icon was given the wrong class. Added fill to close class instead of button. 

## Testing
Check the regression on dark themes.


tw15egan (approved):

dakahn (approved):
:surfer:

PR #6550, opened 2020-07-24
fix(data-table): fix data-table with sorting in ie11

### LINKED ISSUE #6536: Unnecessary line breaks in sortable table headers on IE11

- [x] `carbon-components`

## Detailed description

On IE11, multi-word table headers have unnecessary line breaks.  

Data Table

> see changed?

Text should all be on one line.

IE11

10.14

See https://react.carbondesignsystem.com/?path=/story/datatable--with-sorting for an example but the problem is not limited to react.

## Additional information

Screen Shot 2020-07-24 at 9 16 44

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6536

Data table with sorting was not rendering properly in IE11 due to unspecified heights /widths

#### Changelog

**New**

- IE11 specific selector for `align-items: center` bug
- Specific heights/ widths added to certain elements

**Changed**

- Rearranged a selector

#### Testing / Reviewing

Test that the Data Table with sorting renders correctly in IE11 and all other browsers


joshblack (approved):

dakahn (approved):
:surfer: :earth_asia:

PR #6569, opened 2020-07-28
fix(icon-build-helpers): update react builder to split up entrypoint

### LINKED ISSUE #6390: Revert changes to icon entrypoint that add file-based re-exports
We're getting reports of test slowdowns when bringing in newer versions of `@carbon/icons-react`. This seemed to happen after we updated the entrypoint to re-export files. When doing a quick profile session in jest, we captured the following flame chart:

image

Users reported that pinning to the version before this entrypoint strategy was introduced solved the problem.

This issue is to revert the re-export based strategy back to the previous flat file entrypoint.

joshblack:
Closes #6390

This PR presents a hybrid approach for handling the challenges discussed in #6390. Instead of re-exporting from each of our ~5000 modules defined, we create a series of chunks that house a configurable number of component exports. We do this as creating a single entrypoint ends up breaking most babel tooling.

As a result, this hybrid approach allows us to configure how we'd like to chunk out our modules while still providing a single entrypoint. It should provide a balanced approach between the two options, and allow us to configure it to optimize performance for file processing and module resolution time.

#### Changelog

**New**

**Changed**

- Update the React icon builder to chunk out exports into dedicated chunks

**Removed**


tw15egan (approved):
Icons are building properly, let me know if there is anything else I can run to test this but looks good to me 👍  ✅

emyarod (approved):

PR #6574, opened 2020-07-29
fix(data-table): make persistent search bar keyboard accessible

### LINKED ISSUE #6573: [TableToolbarSearch]: Search input is not focused when persistent prop is used.
## Detailed description

When **persistent** prop is set on TableToolbarSearch, user is not able to type inside input, when TableToolbarSearch is focused via keyboard (Tab).  There is no cursor inside. The only way to make it work, is to click on it, so user is not able to use only keyboard to interact with this field. Please check code sandbox below.

1. Create table with TableToolbarSearch. Set **persistent** prop
2. Use Tab key to switch focus to TableToolbarSearch.
**Result:** TableToolbarSearch is focused, but focus is not set on input. User can't type any character.
**Expected:** Input if focused, user is able to type into search input, just after focus via keyboard, without need to click it on input.

image

https://codesandbox.io/s/sweet-elgamal-bghmh?file=/src/index.js

janhassel:
Closes #6573

When using `persistent` prop on `TableToolbarSearch` component, the search bar was not keyboard accessible since its tabindex was set to `-1`.

#### Changelog

**Changed**

- Control container and input `tabindex` as well as input's `aria-hidden` based on `expanded || persistent` instead of `expanded` in `TableToolbarSearch`
- Updated DataTable snapshots regarding this

#### Testing / Reviewing

- Testing should make sure PR has no side effects
- For review: run react storybook and change `defaultExpanded` to `persistent` in `with-batch-actions.js` story. Search bar should be in tab order now. For previous behaviour, refer to the code sandbox in the linked issue or change

```js
const searchExpanded = expanded || persistent;
```

in `TableToolbarSearch.js` with

```js
const searchExpanded = expanded;
```
Search bar shouldn't be in the tab order anymore then.


tw15egan:
Thanks for taking the time to contribute a fix! I'm noticing a small issue, however. In the original issue, I do see the problem in the second search bar with `persistent`. However, if I switch the wording to `persistant` (the old, deprecated prop name), it works properly. 

In this PR, it now works for `persistent`, but it now seems broken if I set the prop name to `persistant` 🤔

janhassel:
@tw15egan True, didn't notice that earlier. Should be fixed now! 👍

tw15egan (approved):
Thanks for fixing that 🙏  👍  ✅

joshblack (approved):

dakahn (approved):
Thank you so much for taking care of this 🏄

PR #6585, opened 2020-07-29
fix(Copy): set font-size after reset

tw15egan:
Based on Slack convo with @holmansze 

Screen Shot 2020-07-29 at 2 21 01 PM

Ensures `font-size: 0` is set after any reset styles

#### Changelog

**Changed**

- Moved `bx--copy` to end of the file so it is not overwritten by `bx--copy-btn` block

#### Testing / Reviewing

Go to `.storybook/styles.scss` and set `$css--reset: false` and view CopyButton. It will look wrong.
Pull down PR
Go to `.storybook/styles.scss` and set `$css--reset: false` and view CopyButton. It will look correct


joshblack (approved):

emyarod (approved):
looks good to me

PR #2785, opened 2019-05-20
feat(react): allow proptype displayname match in childOf checks (WIP)

finken2:
Allow displayname match for proptype checks in:
prop-types/childrenOf
prop-types/childrenOfType

The impetus for this was react-hot-loader proptype checks failing. The react-hot-loader wraps classes in a proxy wrapper so the strict class match doesn't work:
https://github.com/gaearon/react-hot-loader#checking-element-types
https://github.com/gaearon/react-hot-loader/issues/304
This causes a react warning in the chrome dev console during development when using the carbon-react dropdown components. 

It seems reasonable to allow displayname to count for these checks. Currently, it appears that these childOf functions are only used in the ListBox set of components.

PR #6595, opened 2020-07-30
fix(ListBoxMenu): use arrayOf(Component) instead of childrenOfType

atikenny:
Relates to [#2785](https://github.com/carbon-design-system/carbon/pull/2785)

Fixes an issue where hot-loader changes the component's instance and === doesn't work properly.

#### Changelog

**Changed**

- ListBoxMenu prop types are validated with arrayOf instead of childrenOfType

#### Testing / Reviewing

- Make sure that no warning or error is thrown when using the ListBoxMenu with ListBoxMenuItem children

#### Screenshot

image

### DIFF
--- a/packages/react/src/components/ListBox/ListBoxMenu.js
+++ b/packages/react/src/components/ListBox/ListBoxMenu.js
@@ -40,6 +40,13 @@ ListBoxMenu.propTypes = {
    */
   children: PropTypes.oneOfType([
     PropTypes.arrayOf(ListBoxMenuItem),
+    /**
+     * allow single item using the workaround for functional components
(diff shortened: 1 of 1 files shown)


emyarod (commented):
looks good to me, but just wondering, do you have an example that showcases the original error output?

atikenny:
@emyarod there is a discussion about it in the linked issue, it basically happens when using hot-reload. It's just a dev time annoyance really nothing else, but would like to keep the noise in our console minimal. As far as I can tell there is no real difference between an array of nodes and the childrenOfType custom prop type.

emyarod (approved):
looks good to me

tw15egan (approved):
👍 ✅

PR #6600, opened 2020-07-31
feat(DataTable): support deep keys in rows

maslof:
Closes  ##

Add support deep keys in rows

#### Changelog

**New**

- support deep keys in rows
- test PR

**Changed**

- normalize function for support deep keys
- storybook shared values
- normalize-test

#### Testing / Reviewing

```js
const rows = [
  {
    id: 'a',
    foo: { deepFoo: 'Foo a' },
    bar: 'Bar a',
    baz: 'Baz a',
  },
  {
    id: 'b',
    foo: { deepFoo: 'Foo b' },
    bar: 'Bar b',
    baz: 'Baz b',
  },
  {
    id: 'c',
    foo: { deepFoo: 'Foo c' },
    bar: 'Bar c',
    baz: 'Baz c',
  },
];

const headers = [
  {
    // Support deep keys
    key: 'foo.deepFoo',
    header: 'Foo',
  },
  {
    key: 'bar',
    header: 'Bar',
  },
  {
    key: 'baz',
    header: 'Baz',
  },
];
```


tw15egan:
Will this break for people _not_ using deep keys in a row? Can you mix and match?

tw15egan (approved):
Tested using old data and works as expected, thanks for adding in this functionality! 

👍  ✅

joshblack:
Hey there @MaslovAnton! 👋  Thanks so much for taking the time to put this contribution together 🙏 

In general, when folks have requested deep key support we would talk about normalizing the data to be rendered in the table so that the structure itself was flat. This tends to work out in most use-cases, but I was curious if you could share more about your use-cases to see where this guidance isn't appropriate.

joshblack:
Going to close this out due to inactivity, feel free to re-open!

chibongho:
We have a use case where the rows have a fairly complex nested structure (not flattened), and we also want to provide a way for 3rd party developers to configure what columns they want to have appear in the table. The 3rd party developers can create their own react Component to render their own desired column in the table (basically a thin wrapper for `<TableCell>` to define their own key to retrieve the data in a row), Given the complex structure, there is no one-size-fits-all way for us to flatten the rows in a way that satisfy all 3rd party developers. They need a way to either define deep keys, or take in the original unflattened row and be able to query it forever they want.

We can get around it by directly key-ing into the unflattened rows instead of using the `cells` values generated by `<DataTable>` like below, but I'm not sure how discouraged it is:

```
const unflattedRows = ...;

<DataTable rows={unflattenedRows}>
  <TableBody>
    {unflattenedRows.map((row, index) => <>
      {row.cells.map(cell => <TableCell key={cell.id}>{cell.value}<TableCell />}
      {unflattenedRows[index].map(urow => <ConfigurableTableCells urow={urow} /> }
    </>}
  </TableBody>
</DataTable>
```

Issue #5638, opened 2020-03-16
[Button-fluid] Adding 3:1 border between touching buttons


shixiedesign:
### Summary

Suggestion is to consider adding `gray-20, 1 px, solid` border between fluid primary and secondary buttons. 

This feature will add a 3:1 distinction between the two UI elements. It's a recommended feature to improve accessibility in data vis (advised by Michael Gower), although there are no clear requirements for UI to do the same, the logic applies as the two buttons should appear to users as distinct, and both interactive. 

### Exploration

![image](https://user-images.githubusercontent.com/15144993/76792328-32e8db80-6799-11ea-9d28-a1be9ddcc47a.png)

aagonzales:
We need @mjabbink and @jeanservaas input on this one.

mjabbink:
Can we see this in context of a layout/usecase? In theory I have not issue with this but need to see more context.

jeanservaas:
I like this and I want to do it in our next release, and I want to add it to button guidance now because we're getting tons of questions about this. This update not only adds clarity and contrast accessibility but in the dark themes, it will prevent the primary and secondary colors from vibrating.

But @shixiedesign I have one question, why not just use the subtle border token rather than going cool gray?

mjabbink:
Agree on both. I don’t think the need is there for cool grey. Unless of course we switched all Grey to Cool Grey.

aagonzales:
Status: ready for development. 

Dev tasks:
- [x] add border to buttons in modal
- [x] add border to button groups in button component

Design tasks:
- [x] update modal images on website
- [x] update in modal kit symbols

laurenmrice:
Closing because this has been added 👍🏻

FabianPiconeDev:
The [official docs](https://www.carbondesignsystem.com/components/button/usage/) says this update is not yet made. Maybe someone can update the docs?

> Although we haven’t made this update yet, we plan to add a 1px border between all fluid buttons that calls the $ui-03 token for subtle borders.

PR #6608, opened 2020-08-04
feat(button): add separator in button sets

emyarod:
Ref #5638

This PR adds a separator between buttons (in button sets only) with a new `$button-separator` token

#### Changelog

**New**

- `ButtonSet` wrapper component
- `$button-separator` token for button sets
- support for vertically stacked button set orientation

**Changed**

- refactor styles and selectors to allow for sets to contain more than two buttons

#### Testing / Reviewing

Confirm the separators appear correct and for button sets only

### DIFF
--- a/packages/components/docs/sass.md
+++ b/packages/components/docs/sass.md
@@ -196,6 +196,7 @@
   - [✅highlight [variable]](#highlight-variable)
   - [✅decorative-01 [variable]](#decorative-01-variable)
   - [✅hover-light-ui [variable]](#hover-light-ui-variable)
+  - [✅button-separator [variable]](#button-separator-variable)
   - [✅skeleton-01 [variable]](#skeleton-01-variable)
   - [✅skeleton-02 [variable]](#skeleton-02-variable)
   - [✅⚠️brand-01 [variable]](#brand-01-variable)
@@ -4227,6 +4228,7 @@ Define theme variables from a map of tokens
   $highlight: map-get($theme, 'highlight') !global;
   $decorative-01: map-get($theme, 'decorative-01') !global;
   $hover-light-ui: map-get($theme, 'hover-light-ui') !global;
+  $button-separator: map-get($theme, 'button-separator') !global;
   $skeleton-01: map-get($theme, 'skeleton-01') !global;
   $skeleton-02: map-get($theme, 'skeleton-02') !global;
   $brand-01: map-get($theme, 'brand-01') !global;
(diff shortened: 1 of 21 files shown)


laurenmrice (changes requested):
We can get rid of the gray 20 rule at the end of the last button. Otherwise everything looks great!

Screen Shot 2020-08-04 at 2 20 18 PM

emyarod:
would we want to just create a `ButtonSet` wrapper component to help create button sets? instead of relying on users adding the `bx--btn-set` class to the container div?

dakahn (approved):

joshblack:
@emyarod I think that makes sense, maybe a `ButtonGroup` component? And folks could specify direction (row, column), or I guess pass in a `stacked` prop?

Group comes up only because that's what I remember seeing in docs, let me know if sets is a better term for these (especially since we have that class already in code)

emyarod:
yeah exactly I was thinking of something like `ButtonSet` since that's the name of the story and the class names in code

laurenmrice (changes requested):
Secondary button in `button sets`: The focus is slightly off and should follow this spec.

button 2

----------------------------------------------------------------------------------------------------------------------------

 Primary button in dark themes for `button sets` and `button stories`: For focus we should be using a transparent 1px space instead of gray, but I am noticing this may be a Button bug in general.
button1

emyarod:
the focus issue should be resolved now, and I will problems that are inherited from the base button styles separately since they are not directly related to the current PR!

laurenmrice (changes requested):
When the Secondary button receives focus, we need the focus to be on top of the rule separator between the buttons. Right now it looks like the focus is getting cut off because it is behind the rule.
Artboard

------------------------------------------------------------------------------------------------------
For the stacked set of buttons, we do not need to include the top rule on the secondary button.

Screen Shot 2020-08-07 at 3 55 47 PM

laurenmrice:
For the new button separator token we need to have the following values:

`button-separator`
White | G10 | G90 | G100 |
| --- | --- | --- | --- |
| G20 `#e0e0e0` | G20 `#e0e0e0` | G100 `#161616` | G100 `#161616` |

laurenmrice (changes requested):
-The stacked secondary button on focus still is getting overlapped by the rule separator.
Screen Shot 2020-08-10 at 11 41 19 AM

------------------------------------------------------------------------------------------------------

-The token being using for the rule separator needs to be Gray 20 #e0e0e0 for the White theme and and Gray 10 theme. It looks like it is using Gray 100 right now.
dark line

laurenmrice (approved):
Looks great !! thank you 🙌🏻

joshblack (approved):

aledavila (approved):

PR #6612, opened 2020-08-04
feat(eslint): enable button-has-type rule

joshblack:
Add rule for `button-has-type` to make sure we explicitly annotate buttons with a type. This catches a frequent bug we have for buttons where they don't have `type="button"`

#### Changelog

**New**

- Enable `react/button-has-type` rule

**Changed**

- Update files to pass eslint rule by adding explicit `type="button"` to buttons

**Removed**

### DIFF
--- a/config/eslint-config-carbon/plugins/react.js
+++ b/config/eslint-config-carbon/plugins/react.js
@@ -26,6 +26,7 @@ module.exports = {
   },
   rules: {
     // react
+    'react/button-has-type': 'error',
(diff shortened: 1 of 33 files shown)


emyarod (commented):
looks good to me, pending snapshot updates

joshblack:
Just updated snapshots, will check back in when the build finishes 👀

tw15egan (approved):

emyarod (approved):

PR #6627, opened 2020-08-06
feat(components): add layout package to components _layout.scss

### LINKED ISSUE #5648: Sass Layout and Breakpoint Variable Access
### Summary
Our team would like access to the breakpoints sass partial similar to what we currently have for spacing, typography, color, etc...

Example using spacing
```scss
@import 'carbon-components/scss/globals/scss/spacing';
.custom-thing {
  margin-right: $layout-03;
}
```

Example using typography
```scss
@import 'carbon-components/scss/globals/scss/typography';
.custom-thing {
  @include type-style('productive-heading-02');
  }
```

In order to do the same for breakpoints, we have to use
```scss
@import 'carbon-components/scss/globals/grid/grid';
```
or hijack another module that imports the breakpoints for us (such as `carbon-components/scss/globals/scss/typography`)

There are two problems with this:
1. Importing grid seems to cause a lot of "stuff" to be done aside from just making some variables and functions available.
2. Importing grid is not practical within a scss module, which we use extensively to isolate individual sub-applications and components within our react framework.

joshblack:
Closes #5648

Add layout package to the `_layout.scss` global entrypoint. This should enable folks to get access to the breakpoints without requiring the vendor directory. This should also unblock https://github.com/carbon-design-system/carbon-website/issues/1466

#### Changelog

**New**

- Add an import to the layout package in our `_layout.scss` global entrypoint for teams to use

**Changed**

**Removed**


andreancardona (approved):
@joshblack looks good to me :)

dakahn (approved):
I like and approve these changes.

PR #6644, opened 2020-08-10
chore(slider): add no wrap white space to range label

aledavila:
Adds a no wrap to the range labels in the rare event that a user needs to have a word instead of a unit increment. However, because it's meant for unit increments a user would need to manually add their own space before the label in the prop. This helps in that use case so the word doesn't wrap and stays on the same line. 

#### Testing / Reviewing

Check that slider and labels look fine.


joshblack:
@aledavila is there an issue or repro of the problem that this is solving? 👀

aledavila:
@joshblack So I had a question on slack about using labels but they had the issue with the spacing. And asked Anna about it. So the issue is that if you input a word it won't have the space since its meant for increments. So you'd have to manually add a space in your prop but if you do so it wraps. So I just made it so it wouldn't wrap. 

Screen Shot 2020-08-10 at 11 11 44 AM
Screen Shot 2020-08-10 at 11 16 51 AM

andreancardona (approved):
@aledavila looks good!

joshblack (approved):

PR #6647, opened 2020-08-11
fix(PaginationNav): add forwardRef

janhassel:
Adds support for passing a `ref` to `PaginationNav` component.

#### Changelog

**New**

- Wrapped PaginationNav component in `React.forwardRef`
- Added passed `ref` to root node of component

**Changed**

- Updated `PublicAPI-test.js.snap` in `packages/react`

#### Testing / Reviewing

- Unit tests should make sure no functionality is altered
- Manual review: create dummy component with a ref like the following and verify it logs the expected reference object.
```jsx
function Dummy() {
  const ref = React.createRef();

  useEffect(() => {
    console.log(ref);
  });

  return <PaginationNav ref={ref} />
}
```


joshblack (approved):

andreancardona (approved):
looks good to merge :)

PR #5918, opened 2020-04-23
feat(text-input): add warn prop

janhassel:
Related issue: https://github.com/carbon-design-system/carbon/issues/6492

This is a draft PR with a sample implementation of a warning state for the TextInput component to start a conversation around the technical side of things.

@aagonzales Right now, the icon is still the round warning icon. I haven't been able to change it yet, since the triangle one doesn't have a separate inner path to color the exclamation mark. I'll open an issue for it shortly.

@joshblack I'd love your input about the implementation and if this would be the right way to go. Also, for accessibility: I believe `aria-invalid` shouldn't be emitted since it could lead to confusion for users that rely on screen readers, would you agree?

PR #6652, opened 2020-08-12
feat(NumberInput): add warn prop

janhassel:
Related: #5918

This PR adds a `warn` prop to the `NumberInput` component to follow the behaviour added to the text input so far (see issue PR above).

#### Changelog

**New**

- Added `props.warn` and `props.warnText` to `NumberInput`

**Changed**

- Updated css selector in `_form.scss` to display warning messages (when input has `data-warn`)

#### Testing / Reviewing

- Errors (invalid state) should always overpower warnings
- Warning message should appear instead of helper text
- Verify visual style with the already implemented warn prop of TextInput


packages/components/src/components/form/_form.scss
```diff
@@ -78,10 +78,11 @@
 
   input[data-invalid],
   .#{$prefix}--number[data-invalid] .#{$prefix}--number__input-wrapper,
+  .#{$prefix}--number[data-warn] .#{$prefix}--number__input-wrapper,
```
joshblack: Was looking at this PR real quick, is `data-warn` to match the `data-invalid` convention above? 👀  If it helps, I think we could go down a class-driven approach for this as the `data-*` states traditionally come from vanilla if I remember correctly

janhassel: Yeah this was mainly to follow the same "predictable" behaviour for the developer. In case a custom style or some logic is in place that utilizes `[data-invalid]` as the css selector, it would be more intuitive to select the warning state with `[data-warn]` than by naming the classes.

I'm open to changing it to a classname though, in which case I would also update the `TextInput.js` to remove its `data-warn` and also soley rely on the `--warning` modifier.

janhassel: @joshblack Just to clarify now, do we want to follow the `[data-warn]` or the `--warning` class convention?

joshblack: Hey @janhassel! So sorry about the delay, `--warning` should be good going forward!

janhassel: @joshblack All good. Changed it accordingly 🙂 (cc @emyarod @andreancardona)

emyarod (approved):

joshblack:
bump @andreancardona  @jeanservaas when you have a chance to review 👀

andreancardona (approved):
@janhassel looks good!

joshblack:
bump @jeanservaas @aagonzales if you have a sec to review 👀

aagonzales (approved):
Looks good! Matches the TextInput warning.

PR #6651, opened 2020-08-11
docs(storybook): updates all stories to CSF format

tw15egan:
Begins work on updating Storybook to `v6`. Since `storiesOf` is deprecated in `v6`, we'll need to update them before we migrate.

### Steps
- [x] Convert from `storiesOf` to `CSF`
- [ ] Upgrade packages to `v6`

#### Changelog

**New**
- `@babel/plugin-proposal-logical-assignment-operators` to fix an error on build

**Changed**
- `downshift` updated from `5.2.1` to `5.2.7` to fix weird `react-is` error (cc @dakahn 😉 )
- All stories have been converted from `storiesOf` format to `CSF`. Step 1 of the docs epic was to do this, so I just ran the codemod across the repo. 

**Removed**

#### Testing / Reviewing

Ensure each story still looks the same

PR #6663, opened 2020-08-13
chore(storybook): Convert stories from storiesOf to CSF format

tw15egan:
Refs #6651 

Begins work on updating Storybook to `v6`. Since `storiesOf` is deprecated in `v6`, we'll need to update them before we migrate.

### Steps
- [x] Convert from `storiesOf` to `CSF` 

#### Changelog

**Changed**
- All stories have been converted from `storiesOf` format to `CSF`. Step 1 of the docs epic was to do this, so I just ran the codemod across the repo. 
- In v6 they will also deprecate the way we annotate stories, so I've also run that codemod to prepare for these changes ([reference link](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#hoisted-csf-annotations))

**Removed**

#### Testing / Reviewing

Ensure each story still looks the same


joshblack (approved):

emyarod (approved):
looks good to me

(although I think this will create merge conflicts for me with the mobile tabs PR 😭)

PR #6678, opened 2020-08-14
chore(storybook): upgrade to v6 [WIP]

tw15egan:
Starting the migration of `storybook` from `5.3.9` to `6.0.7`

Using the storybook migration guide as I go, but still running into some issues. Builds take ~30 minutes, which I'm guessing is due to a change that was made that I haven't fixed on our end yet. 

Migration guide: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-53x-to-60x

Quick notes on what has been changed so far:
- Changed to hoisted annotations ([ref](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#hoisted-csf-annotations))
- `Preview` is changed to `Canvas` and `Props` is changed to `ArgsTable` ([ref](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#60-docs-breaking-changes))
- `preview.js` now contains just the parameter and decorator configurations. Instead of `addParameters` / `addDecorators`, this has been updated to `export const parameters` ([ref](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-addparameters-and-adddecorator))
- `manager.js` now contains the theme creation, action configurations, and custom themes. Theme creation seems like it belongs here, but not too sure about our custom theme and the `configureActions`. ([ref](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#using-managerjs))

@joshblack let me know if you have any ideas I can try out, I'll keep poking around and changing things around 😅


tw15egan:
I yield 💀

PR #6679, opened 2020-08-17
fix(FloatingMenu): ignore disabled nodes for autofocus

### LINKED ISSUE #6677: [OverflowMenu]: second item is not focused when first item is disabled

- [ ] `carbon-components`
- [x] `carbon-components-react` v10.17.0

## Detailed description

When an overflow menu is opened, the first item in the menu is focused.

When the first item in an overflow menu **is disabled** (or if the first n items are disabled), I would expect that the focus be applied to the first non-disabled item. 

Currently in this situation the focus isn't applied inside the menu at all but remains on the trigger.

focus boo

Sandbox: https://codesandbox.io/s/unruffled-pine-yeky6?file=/src/index.js

I suppose we might be able to get around this by modifying OverflowMenu's `selectorPrimaryFocus` to be an nth selector with a :not(disabled) or something similar, but ideally this functionality would be provided by default without any additional config. We used to be able to configure this via OverflowMenuItem's `primaryFocus` but it's been deprecated.

emyarod:
Closes #6677

This PR adds a check for the disabled property on tabbable and focusable nodes in the floating menu autofocus logic

#### Testing / Reviewing

Confirm that disabled nodes are not focused when floating menu components are opened


tw15egan (approved):
Tested locally with Item 1 disabled and focus went to Item 2 on menu open 👍 ✅

packages/react/src/internal/FloatingMenu.js
```diff
@@ -336,8 +336,12 @@ class FloatingMenu extends React.Component {
     const primaryFocusNode = menuBody.querySelector(
       this.props.selectorPrimaryFocus || null
     );
-    const tabbableNode = menuBody.querySelector(selectorTabbable);
-    const focusableNode = menuBody.querySelector(selectorFocusable);
+    const tabbableNodes = menuBody.querySelectorAll(selectorTabbable);
+    const focusableNodes = menuBody.querySelectorAll(selectorFocusable);
+    const [tabbableNode] = [...tabbableNodes].filter((node) => !node.disabled);
+    const [focusableNode] = [...focusableNodes].filter(
```
asudoh:
```suggestion
    const focusableNode = [...focusableNodes].find(
```

packages/react/src/internal/FloatingMenu.js
```diff
@@ -336,8 +336,12 @@ class FloatingMenu extends React.Component {
     const primaryFocusNode = menuBody.querySelector(
       this.props.selectorPrimaryFocus || null
     );
-    const tabbableNode = menuBody.querySelector(selectorTabbable);
-    const focusableNode = menuBody.querySelector(selectorFocusable);
+    const tabbableNodes = menuBody.querySelectorAll(selectorTabbable);
+    const focusableNodes = menuBody.querySelectorAll(selectorFocusable);
+    const [tabbableNode] = [...tabbableNodes].filter((node) => !node.disabled);
```
asudoh:
```suggestion
    const tabbableNode = [...tabbableNodes].find((node) => !node.disabled);
```

asudoh (commented):
Good try - Some suggestions in case you are interested. Thanks @emyarod!

packages/react/src/internal/FloatingMenu.js
emyarod: thanks for the suggestion @asudoh!

asudoh (approved):
LGTM 👍 - BTW have we explored `:disabled()`? Just a consideration. Thanks @emyarod!

andreancardona (commented):
Looks good to me :). @emyarod

andreancardona (approved):
@emyarod looks good to me :)

PR #6700, opened 2020-08-20
feat(data-table): add getCellId in filterRows

### LINKED ISSUE #2593: Provide getCellId in the filterRows API
### Summary

The `filterRows` API for `DataTable` is missing `getCellId` parameter

### Justification

`getCellId` is required for most (if not all) custom filtering cases, so I really think it should be provided in the API https://github.com/IBM/carbon-components-react/blob/master/src/components/DataTable/README.md#filterrows
As a workaround I could use `import { getCellId } from "carbon-components-react/lib/components/DataTable/tools/cells"` but that feels more like a hack than using a public API.

Cell values for each row could be easily retrieved from `filterRows` function using just the parameters that are provided to the function.

### "Must have" functionality

see above

N/A

### Available extra resources

N/A

joshblack:
Closes #2593 

Add `getCellId` to `filterRows` prop

#### Changelog

**New**

**Changed**

- Add `getCellId` to `filterRows` prop

**Removed**


tw15egan (approved):

andreancardona (approved):
@joshblack looks good to me :)

PR #6707, opened 2020-08-21
fix(overflow-menu): check if function exists before calling

### LINKED ISSUE #6705: Overflow using renderIcon throws error in Firefox on click outside
When using renderIcon the overflow sometimes throws an error when clicked outside

```
TypeError: target.matches is not a function
var target = event.target;
  259 |   var triggerEl = _this._triggerRef.current;
  260 | 
> 261 |   if (!menuBody.contains(target) && triggerEl && !target.matches(".".concat(prefix, "--overflow-menu,.").concat(prefix, "--overflow-menu-options"))) {
```

- [ X ] `carbon-components`
- [ X ] `carbon-components-react`

## Detailed description

Error occurs on click outside.

Using renderIcon

Firefox 79.0, 64bit

    "carbon-components": {
      "version": "10.17.0",
    "carbon-components-react": {
      "version": "7.17.0",

> should be aware of?

IBM Aspera

1. Open overflow
2. Click outside
3. It might happen it might not. When it does happen it will happen every time you click outside until reload.

https://codesandbox.io/s/lucid-franklin-pedx0?file=/index.js

## Additional information

Screen Shot 2020-08-20 at 7 06 05 PM

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6705

Check if `target.matches` exists before invoking it

#### Changelog

**Changed**

- Check if `target.matches` exists before invoking it

#### Testing / Reviewing

Using Firefox, open OverflowMenu, and open the menu. Then click anywhere outside the browser, return to the browser, and click anywhere inside to close the menu. There should not be an error.

### DIFF
--- a/packages/react/src/components/OverflowMenu/OverflowMenu.js
+++ b/packages/react/src/components/OverflowMenu/OverflowMenu.js
@@ -405,14 +405,16 @@ class OverflowMenu extends Component {
         (event) => {
           const { target } = event;
           const { current: triggerEl } = this._triggerRef;
-          if (
-            !menuBody.contains(target) &&
-            triggerEl &&
-            !target.matches(
-              `.${prefix}--overflow-menu,.${prefix}--overflow-menu-options`
-            )
-          ) {
-            this.closeMenu();
+          if (typeof target.matches === 'function') {
+            if (
+              !menuBody.contains(target) &&
+              triggerEl &&
+              !target.matches(
+                `.${prefix}--overflow-menu,.${prefix}--overflow-menu-options`
+              )
+            ) {
+              this.closeMenu();
+            }
           }
         },
         !hasFocusin


dabrad26:
I see similar target.matches in 
- pagination-nav
- init-component-by-search
- init-component-by-event
- init-component-by-launcher
- ui-shell/product-switcher
- dropdown
- data-table

Maybe should have this check for all the cases.  IDK what init-component is for, but I assume the others are similar to what overflow was doing.

packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -405,14 +405,16 @@ class OverflowMenu extends Component {
...
           const { current: triggerEl } = this._triggerRef;
-          if (
-            !menuBody.contains(target) &&
-            triggerEl &&
-            !target.matches(
-              `.${prefix}--overflow-menu,.${prefix}--overflow-menu-options`
-            )
-          ) {
-            this.closeMenu();
+          if (target.matches) {
```
dabrad26:
```suggestion
          if (typeof target.matches === 'function') {
```

Which is donehttps://github.com/carbon-design-system/carbon/blob/7a95a7d46c2a36c19fffafcbf1153c7ba19f7f7f/packages/components/src/globals/js/misc/event-matches.js#L19 (function type check).

dabrad26 (changes requested):
IDK of a reason matches would be set to anything but function. But normally for libraries I go for overly cautious :/

dabrad26 (approved):

tw15egan:
I agree, made it match `function` specifically. Not seeing any issues with those other components, I think the main reason this one was throwing was due to the positioning logic of the `FloatingMenu`. 

Thanks again for catching that and taking a look 👍

dabrad26:
Thanks! I am a bit concerned with an event getting triggered in Firefox and then bubbling it in on viewport focus (with no target); that def seems like wrong behavior.  

I hope its a bug in Firefox and they fix it.  But if not I fear it could show as more issue later for other events that rely on target (even if not error can cause the expected behaviors to get impacted).  

One concern I saw was it seemed when clicking in Firefox dev tools around the events backed up and would then hit the app over and over once viewport got focus (causing freezing).  Seems very odd...  Will keep an eye on Firefox's bug reports and open if I don't see others talking about it.

dakahn (approved):

andreancardona (approved):

PR #6714, opened 2020-08-24
feat(type): use unitless line-heights

### LINKED ISSUE #6711: Type styles using hard coded line heights
https://github.com/carbon-design-system/carbon/blob/master/packages/type/scss/_styles.scss

This means if a font size is changed due to a11y purposes or custom/snowflake styles the line height will not adjust with the font size resulting in misaligned or constrained text. The standard method for defining line-height is to use unitless values (e.g. 1.25).

With this, child elements can compute their line heights based on their computed font size, rather than inheriting an arbitrary value from a parent that is more likely to need overriding.

More info:
https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/
https://css-tricks.com/almanac/properties/l/line-height/#unitless-line-heights

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6711

Updates our type package to use a unitless `line-height` value

#### Changelog

**Changed**
- Modified the `scss` and `js` type scale files
- Instead of calling a `carbon--rem()` to set the `line-height`, we just set it to a unitless value
- Typos

#### Testing / Reviewing

Ensure the values match the website documentation, with the exception of `heading-03`. An issue has been made for that here https://github.com/carbon-design-system/carbon-website/issues/1604


dakahn (approved):

joshblack (approved):

joshblack:
The changes look good code-wise, although from the percy snapshots it seems like there might be some unexpected line height changes? I'm not sure how the line-height value was computed but for `productive-heading-06` it seems like the font-size is 42px and the line-height was set to `3.125` which I believe gives us the 131.25px value line-height in that story (`42px * 3.125 = 131.25`)

I think the value of line-height should be computed by proportion instead of relative to the base font-size, so instead of `50px / 16px = 3.125` it'd be `50px / 42px = 1.19` for the line-height value. What do you think @tw15egan? 👀

tw15egan:
@joshblack Ah yeah that would explain the differences... so would we need to compute that for all of them? i.e. `body short 01` would be `18px / 14px = 1.285`?

joshblack:
@tw15egan yeah! Exactly.

tw15egan:
@joshblack updated. After making these changes, I noticed there were still 29 diffs in Percy. Seems like Chrome rounds `15.94px` DOWN to `15px`, which was throwing off any line heights that were set to `1.33`. As a result, I bumped any `1.33` values up to `1.34` 👍

joshblack (approved):

PR #6458, opened 2020-07-13
fix(Tooltip): add focus trap behavior and fix focus management logic

emyarod:
Closes #6443
Refs https://github.com/carbon-design-system/carbon/issues/1264#issuecomment-637159076
Refs https://github.com/carbon-design-system/carbon/issues/4414
Refs https://github.com/carbon-design-system/carbon/pull/5489

This PR implements optional focus trapping in the interactive tooltip and correctly places focus back on the trigger element after the tooltip is dismissed.

#### Changelog

**Changed**

- enable focus trap option in interactive tooltip
- place focus back on tooltip trigger element after tooltip dismissal rather than around elements where the tooltip portal is created/destroyed
- rename `wrapFocus` props to be more generic

#### Testing / Reviewing

Confirm that focus and blur behaves as expected within the interactive tooltip component as well as between interactive siblings to the tooltip. Ensure that focus management remains the same in other `wrapFocus` or `FloatingMenu` components like Modal, ComposedModal, and OverflowMenu

PR #6722, opened 2020-08-25
feat(Tooltip): allow focusTrap prop passthrough

### LINKED ISSUE #6719: Tooltip focus grab is breaking previous usages of Tooltip

- [x] `carbon-components 10.16`
- [x] `carbon-components-react 7.16`

## Detailed description

I am using a Tooltip with a PasswordInput to provide Guidance to the user while they type a new password.  I recently upgraded to the versions above and this is broken.  The Tooltip is grabbing focus like a modal and the input cannot be accessed.

Tooltip.  I think the focus problem would happen with any component the Tooltip is used with.

> see changed?
Before Tooltips could be used with components and not grab focus.  I'd like a way to restore that previous behavior.

Chrome 84.0.4147.105

Versions above.

> should be aware of?
I work for IBM Cloud (Console).   I would need a fix soon or it will block other stuff.

1. Step one
2. Step two
3. Step three
4. etc.

The following code in a sandbox will reproduce:

image

>
> - Style and vanilla JS:
> - React:

## Additional information

- Screenshots or code
- Notes

The component I am working with looks like this (when it was working):
image

emyarod:
Closes #6719 

Related #6458

This PR allows users to set the `focusTrap` prop on Tooltikps to passthrough to the underlying `FloatingMenu`. A check for the `focusTrap` prop is also added to the `FloatingMenu` to prevent moving focus away from tooltip trigger elements

#### Changelog

**New**

- `focusTrap` prop on `Tooltip` to control the underlying `FloatingMenu`

**Changed**

- check `focusTrap` prop before calling focus method in `FloatingMenu`

#### Testing / Reviewing

Confirm that tooltips do not steal focus away from focusable trigger elements when `focusTrap` is false

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip-story.js
+++ b/packages/react/src/components/Tooltip/Tooltip-story.js
@@ -110,7 +110,17 @@ function UncontrolledTooltipExample() {
           triggerText={<div>My text wrapped with tooltip</div>}
           open={value}
           showIcon={false}>
-          Some text
+          <p id="tooltip-body">
+            This is some tooltip text. This box shows the maximum amount of text
+            that should appear inside. If more room is needed please use a modal
+            instead.
+          </p>
+          <div className={`${prefix}--tooltip__footer`}>
+            <a href="/" className={`${prefix}--link`}>
+              Learn More
+            </a>
+            <Button size="small">Create</Button>
+          </div>
         </Tooltip>
       </div>
     </>
--- a/packages/react/src/components/Tooltip/Tooltip-story.js
+++ b/packages/react/src/components/Tooltip/Tooltip-story.js
@@ -29,7 +29,6 @@ const directions = {
 const props = {
   withIcon: () => ({
     direction: select('Tooltip direction (direction)', directions, 'bottom'),
-    focusTrap: boolean('Focus trap (focusTrap)', true),
     triggerText: text('Trigger text (triggerText)', 'Tooltip label'),
     tabIndex: number('Tab index (tabIndex in <Tooltip>)', 0),
     selectorPrimaryFocus: text(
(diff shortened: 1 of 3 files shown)


tw15egan (approved):

joshblack:
Would the intended behavior here be that if the dialog doesn't trap focus (nonmodal dialog) that it should stay in tab order? Similarly, if we are using `role="dialog"` should we be shifting the focus still to the first focusable element?

Side-note, would this be a good candidate for revisiting with the nested roles? (tooltip -> dialog)

emyarod:
yes it should stay in tab order when the focus trap behavior is disabled. we can remove the role attribute when the focus trap is disabled if that's what you mean. and yes this should probably be a use case to consider in those discussions (I'm out of the loop on what explorations have been done)

joshblack:
> I'm out of the loop on what explorations have been done

Honestly there hasn't been much outside of conversations on PRs like this one haha

> yes it should stay in tab order when the focus trap behavior is disabled

Makes sense, was trying to reproduce in the storybook but was having trouble tabbing into the content:

demo

> we can remove the role attribute when the focus trap is disabled if that's what you mean

I believe if we have something with `role="dialog"` we need to shift focus into the dialog, whether or not it traps focus in the dialog.

emyarod:
I'm not sure about how your example looks but there is an uncontrolled tooltip example in the storybook. This prop was only meant to be disabled in the uncontrolled component scenario, since by default we are trapping focus right? To avoid that confusion I guess we can remove the prop from the default story maybe. As for the role attribute, what would be the verdict on that? it sounds like it should be removed when focus trap is disabled

joshblack:
@emyarod just was noting that for [this story](https://deploy-preview-6722--carbon-components-react.netlify.app/?path=/story/tooltip--default-bottom) (which I think is the uncontrolled one) that when `focusTrap` is false in knobs you can't tab into the content of the tooltip. I think whether or not we trap focus that we should keep the contents of the popup in tab order/reachable by keyboard.

emyarod:
right that's what I was referring to with my earlier comment

> This prop was only meant to be disabled in the uncontrolled component scenario, since by default we are trapping focus right? To avoid that confusion I guess we can remove the prop from the default story 

this prop is only relevant in the controlled example so should I just remove it entirely from the uncontrolled example?

joshblack:
Got it, sounds good! Sorry about the confusion 😄

joshblack (approved):

tw15egan:
@emyarod I was unsure if you wanted to add anything else to this PR. Looks good on my end, merge when ready 👍

PR #6738, opened 2020-08-27
fix(navigation): ignore disabled focusable elements

emyarod:
Related #6677 #6679

This PR fixes an issue with our `selectorFocusable` selectors to ignore disabled nodes with a tabindex attribute, which allows us to revert to the original autofocus logic in FloatingMenu

#### Testing / Reviewing

Confirm the autofocus behavior is consistent with the result of #6679

### DIFF
--- a/packages/react/src/internal/keyboard/navigation.js
+++ b/packages/react/src/internal/keyboard/navigation.js
@@ -68,5 +68,5 @@ export const selectorFocusable = `
   a[href], area[href], input:not([disabled]),
   button:not([disabled]),select:not([disabled]),
   textarea:not([disabled]),
-  iframe, object, embed, *[tabindex], *[contenteditable=true]
+  iframe, object, embed, *[tabindex]:not([disabled]), *[contenteditable=true]
 `;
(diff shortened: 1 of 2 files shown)


packages/react/src/internal/keyboard/navigation.js
```diff
@@ -68,5 +68,5 @@ export const selectorFocusable = `
   a[href], area[href], input:not([disabled]),
   button:not([disabled]),select:not([disabled]),
   textarea:not([disabled]),
-  iframe, object, embed, *[tabindex], *[contenteditable=true]
+  iframe, object, embed, *[tabindex]:not([disabled]), *[contenteditable=true]
```
asudoh, line 71: Would this work?

```suggestion
  iframe, object, embed, *[tabindex]:not(:disabled), *[contenteditable=true]
```

asudoh (commented):
Thank you for your quick update @emyarod!

packages/react/src/internal/keyboard/navigation.js
emyarod: yes that also works, I was following the convention from the previous selectors like `textarea:not([disabled])`. would one be preferred over the other here?

asudoh: Oh good point! `:disabled` doesn't work for custom elements, so probably we should keep this as-is.

asudoh (approved):
Thanks @emyarod!

tw15egan (approved):

joshblack (approved):

PR #6744, opened 2020-08-28
fix(tooltip): correct tooltipId prop name

tay1orjones:
Every time we update our jest snapshots we're seeing regenerated values on tooltips like 
```
aria-controls="__carbon-tooltip_5nwulidr9lq"
aria-describedby="__carbon-tooltip_5nwulidr9lq"
```

Recently in https://github.com/carbon-design-system/carbon/pull/6458 the new autogenerated value was added and it [checks against an undocumented `id` prop](https://github.com/carbon-design-system/carbon/pull/6458/files#diff-86c971c455ffb6e81dbdb821e9ea9840R234). We can correct our issue by changing all usages of `tooltipId` to `id`, but this seems to be an unintentional change that should instead be using `tooltipId` if one is passed. 

@emyarod could you confirm if this is correct?

#### Changelog

**Changed**

- update prop usage for unique tooltip identifier prop

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.js
+++ b/packages/react/src/components/Tooltip/Tooltip.js
@@ -231,7 +231,8 @@ class Tooltip extends Component {
    * @private
    */
   _tooltipId =
-    this.props.tooltipId || `__carbon-tooltip_${Math.random().toString(36).substr(2)}`;
+    this.props.tooltipId ||
+    `__carbon-tooltip_${Math.random().toString(36).substr(2)}`;
 
   /**
    * Internal flag for tracking whether or not focusing on the tooltip trigger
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Tooltip/Tooltip.js
```diff
@@ -231,7 +231,7 @@ class Tooltip extends Component {
    * @private
    */
   _tooltipId =
-    this.props.id || `__carbon-tooltip_${Math.random().toString(36).substr(2)}`;
+    this.props.tooltipId || `__carbon-tooltip_${Math.random().toString(36).substr(2)}`;
```
emyarod:
```suggestion
    this.props.tooltipId || this.props.id || `__carbon-tooltip_${Math.random().toString(36).substr(2)}`;
```

should we still account for users passing in an `id` prop or is it intuitive to reach for `tooltipId`, `tooltipBodyId`, and `triggerId`?

emyarod (approved):
looks good to me, just need to format with prettier to pass CI

dakahn (approved):

joshblack (approved):

Issue #3720, opened 2019-08-13
Tooltip onHover


eglepuo:
### Summary

Tooltip works only onClick, it doesn't work on hover and for having tooltip on hover I have to use TooltipDefinition component. But if I want to open tooltip on the same information icon  then I get warning that TooltipDefinition should be used only with strings. I think Tooltip component should have option to work on hover or on click.

### Justification

Most of the people use tooltips on hover and not all the tooltips goes on the text, sometimes they go on icons too.

### "Must have" functionality

Tooltip open option -> onClick or onHover

asudoh:
Hi 👋 thank you for reporting! Actually it was out conscious decision, where our design dictates that interactive tooltip with richer (especially interactive) contents should be opened by interaction, and definition tooltip shouldn't have a richer content. Some discussion is going on (where it was completely slipped my mind) wrt something like bold text can be in definition tooltip. CC @carbon-design-system/design

elizabethsjudd:
There is a whole conversation about the contents of the different types of tooltips starting here that explains why rich text is **not** allowed within definition tooltips: https://github.com/carbon-design-system/carbon/pull/3476#issuecomment-518670586

PR #6758, opened 2020-09-01
Tooltip add hover support

### LINKED ISSUE #6484: [Tooltip/TooltipDefinition] Support for hover or truncation with ellipses
### Summary
Can we have a prop added to allow hover support for the `Tooltip` component?

If not, I see the the `TooltipDefinition` does support hover. However, it's wrapped in a button tag which prevents it from being truncated with an ellipses when the screen size is smaller.

https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/TooltipDefinition/TooltipDefinition.js#L76

Thanks!

rob-richards:
Closes #6484 

This PR allows the user to pass a prop called `hover` to `Tooltip` that opens when the mouse is over the trigger.

#### Changelog

**New**

- `hover` prop on Tooltip to open FloatingMenu
- Test added to ensure that the tooltip is opened when prop is passed in

**Changed**

- Changed mouse events and hover prop to control hover behavior

#### Testing / Reviewing

Pass `hover={true}` to the `Tooltip` component. Place the mouse cursor over the tooltip trigger and observe.

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip-test.js
+++ b/packages/react/src/components/Tooltip/Tooltip-test.js
@@ -174,8 +174,10 @@ describe('Tooltip', () => {
         open: true,
         triggerText: 'Tooltip',
       });
-      // Enzyme doesn't seem to allow state() in a forwardRef-wrapped class component
-      expect(rootWrapper.find('Tooltip').instance().state.open).toEqual(true);
+      setTimeout(() => {
+        // Enzyme doesn't seem to allow state() in a forwardRef-wrapped class component
+        expect(rootWrapper.find('Tooltip').instance().state.open).toEqual(true);
+      }, 200);
     });
 
     it('prop.hover change should update hover state', () => {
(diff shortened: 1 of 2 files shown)


emyarod (commented):
IIRC it was a deliberate decision not to trigger interactive tooltips on hover, not sure if this has changed but here is a previous discussion and further context https://github.com/carbon-design-system/carbon/issues/3720

rob-richards:
Thanks @emyarod. I saw the proposal was set to open by @tw15egan, which is why I opened this PR. Do one of you mind verifying that making the tooltip an _optional_ hover is not within the scope of this component? Thanks!

emyarod:
to my knowledge, optional hover is not within the scope of the interactive tooltip component

tw15egan:
@rob-richards it looks like the original issue has been accepted by design now with the following caveat:

> * Hover + non-interactive tooltip is needed. Need to figure out how to do this from a dev perspective. Interactive tooltips should continue to not support the hover.

We'll need to find a way to only allow a hover prop if there is no interactive content within the tooltip

PR #6762, opened 2020-09-01
chore(release): v10.19.0-rc.0

dakahn:
Release PR for v10.19.0-rc.0


joshblack (commented):
Do all of these packages have release-related bumps, or is it coming from `scss-generator` and `test-utils` changing? If so, we might be able to cut back on the number of releases since those packages are internal only and won't change the resulting shipped package

PR #6770, opened 2020-09-02
docs(Toggles): add label examples to stories; remove a11yProps

dakahn:
Noticed while triaging an issue that we didn't have labels on our Toggle examples. So I threw some on Toggle and ToggleSmall 👍 

Also removed a11yProps and just added the label to the main props object. 

Also removed the unneeded aria label since we have explicit labeling with the `<label>` element 🏄


tw15egan:
Just noticed there is a difference in margin between the skeleton label and the non-skeleton label. Also noticed that there are no skeleton styles... 

If we don't have skeleton styles defined (not sure they are needed for a toggle?) perhaps we should just remove the examples and any styles associated with them

joshblack (approved):
Looks good outside of the feedback from TJ above 👍

joshblack:
bump @dakahn when you have a sec 👀

dakahn:
@tw15egan sorry, coming back to this after -- too too long.

I fixed the spacing issue by getting the skeleton code in line with the regular toggle code. So you think we should maybe scrap Toggle.Skeleton and all it's styles altogether?

joshblack:
For skeleton I bet we could just remove it from the story until we add in the styles for it 👀

tw15egan:
Yeah I'd just remove the skeleton stories until we implement it 😄  Everything else looks great!

tw15egan:
I think we just need to remove the `ToggleSmall` skeleton then we're good to GO ✈️

dakahn:
@tw15egan oh duh, sorry

joshblack:
bump @tw15egan I think it's missing 1 review for the auto-merge if you have a sec to approve 👀

tw15egan (approved):
🎉

PR #6784, opened 2020-09-06
remove redundant proptypes value for locale

panpan-lin:
There are two `en` there. I suppose one of them is redundant.

**Removed**

- remove one of the two `en` in proptypes value for locale


emyarod (approved):
do we need to have separate en-US and en-GB locales? otherwise looks good to me after snapshot updates

tw15egan:
@emyarod I believe that would be handled with the `dateFormat` prop. I'm actually not seeing either `en-US` or `en-GB` on their site, so `en` may just be aliased to `default` 

https://github.com/flatpickr/flatpickr/tree/master/src/l10n

tw15egan (approved):

Issue #6777, opened 2020-09-03
Filtered and non-filtered tags have different heights


nsukaviriya:
Codebase and when we view tags on the Carbon site, the filtered and plain tags have 2 different heights
![image](https://user-images.githubusercontent.com/68748472/92137815-a03a0880-eddb-11ea-9d5c-0195654308d3.png)

However, when designers use the CC Sketch library they are the same height
![image](https://user-images.githubusercontent.com/68748472/92138004-da0b0f00-eddb-11ea-86bb-540f215e2e55.png)

I assume you didn't design them to be of different heights on purpose.

## What package(s) are you using?

Tag, Carbon-components, Angular

- [X ] `carbon-components`
- [ ] `carbon-components-react`

## Detailed description

> Describe in detail the issue you're having.

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
"carbon-components": "10.14.0",
"carbon-components-angular": "^4.12.4"

- Screenshots or code
- Notes

PR #6785, opened 2020-09-06
fix Tag component

GoldenFishe:
Closes #

Fix diff height Tag and Tag (with icon) https://github.com/carbon-design-system/carbon/issues/6777

#### Changelog

**Changed**

- Changed height from `1.5rem` to `1.8rem.` Because Tag with icon have `1.8rem` height


GoldenFishe:
https://github.com/carbon-design-system/carbon/issues/6777 fix

tw15egan:
I believe we need to find a way to reduce the height of the `Filter Tag` button, as Tags are supposed to render at `24px / 1.5rem` 

https://www.carbondesignsystem.com/components/tag/style#structure

GoldenFishe:
Ok. So we need to reduce content height. In Tag with filter we got +4px from "close" button. So I reduce "close" button height from 20px to 16px. And now we have 24px for Tag and 24px for Tag with filter

joshblack (approved):
Looks good with the changes, curious what you think @tw15egan 👀

tw15egan (approved):
Looks great! Thanks for taking the time to fix this 👍  ✅

Issue #6788, opened 2020-09-07
Changing isExpanded props in TreeNode to false doesn't change TreeNode inner state


MaciejRusek:
Changing isExpanded props in TreeNode to false doesn't change TreeNode inner state

## What package(s) are you using?

- [ ] `carbon-components`
- [x] `carbon-components-react`

## Detailed description
As TreeView and TreeNode aren't officially released yet I've copied the source code directly to our app.

The problem is with the `isExpanded` prop. In my scenario, I need the button to collapse all nodes at once. To do so I am using `isExpanded` prop of the TreeNode component.
Information about the expanded or collapsed node is stored in redux. So when collapse all button is clicked I'm setting `isExpanded` prop to false. Unfortunately, it doesn't collapse the nodes. Prop changes to the false but the inner state doesn't (see the gif below).

![bug](https://user-images.githubusercontent.com/57291202/92396877-0ef1cb80-f126-11ea-9204-8c2990ab9f37.gif)

PR #6789, opened 2020-09-08
fix(tree): allow prop to decide about the state

MaciejRusek:
https://github.com/carbon-design-system/carbon/issues/6788

All information is in the issue.


emyarod (changes requested):
this change will break tooltip expansion when it is not controlled by the TreeNode props (try expanding or collapsing a tree node)

this PR should be superseded by https://github.com/carbon-design-system/carbon/pull/6791

joshblack:
Closing in preference of https://github.com/carbon-design-system/carbon/pull/6791 based on the comment above 👀

PR #6791, opened 2020-09-08
fix(TreeNode): sync expanded prop and state

### LINKED ISSUE #6788: Changing isExpanded props in TreeNode to false doesn't change TreeNode inner state

emyarod:
Closes #6788

This PR syncs the tree node expanded prop and state to allow for controlled mode expansion/collapse

#### Changelog

**New**

- TreeView story with global expansion controls

**Changed**

- `TreeNode` expansion prop and state sync

#### Testing / Reviewing

Confirm the treeview story with expansion controls functions as expected and the existing functionality has no regressions


tw15egan (approved):
👍 ✅

joshblack (approved):

PR #6807, opened 2020-09-10
fix(OverflowMenu): stop focusing button onblur

### LINKED ISSUE #6227: OverflowMenu steals back focus when clicking outside and it collapses

- [ ] `carbon-components`

## Detailed description

If I click on a button (or really anything) while the OverflowMenu is expanded.  It steals focus back to the collapsed Menu button instead of focusing on what I clicked on

> see changed?

> should be aware of?

Take a look at the sandbox below.
Expand the Overflow Menu
Now click on the button next to the Overflow Menu.
Notice that the OverflowMenu Icon has stolen the focus back to itself

https://codesandbox.io/s/carbon-overflow-menu-focus-stealing-9k1pl

## Additional information

- Screenshots or code
Here's a picture of the stolen focus after I clicked the Click me button
image

gptt916:
Closes #6227

This PR removes focus button function call from closeMenu, instead only call it when esc key is pressed to close the overflow menu

#### Changelog

**Changed**

- OverflowMenu button no longer aquires focus when the menu is closed via clicking elsewhere on the page.

#### Testing / Reviewing

Leveraging Chrome's focus tracker (https://developers.google.com/web/tools/chrome-devtools/accessibility/focus) verify that the button does not have focus after closing the menu via clicking on the canvas. 

Also verify that closing the menu via the Esc button still results in the menu button aquiring focus.


packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -321,11 +327,11 @@ class OverflowMenu extends Component {
     }
   };
 
-  closeMenu = () => {
-    let wasOpen = this.state.open;
+  closeMenu = (cb) => {
```
emyarod: is this change also intentional? the `closeMenu` method is internal and I don't see any new callback usage

gptt916: Yep, it's being called on line 310 with the callback, moving the call to `focusMenuEl` to be executed only on Escape press, which I believe was the desired behaviour.

emyarod: gotcha looks good to me, let's rename the parameter to something more descriptive like `onCloseMenu` or something

emyarod (approved):

andreancardona (changes requested):
@gptt916 once param is updated to `onCloseMenu` looks good to me!

gptt916:
@andreancardona Changes are ready for review

joshblack (approved):

andreancardona (approved):
@gptt916 🎉

PR #6811, opened 2020-09-10
fix(Toggle): make labelText a node instead of string

### LINKED ISSUE #6810: Toggle, ToggleSmall: labelText should not be limited to strings only
## Title line template: Brief description

`labelText`, `labelA` and `labelB` should not be strings only. There is no technical reason.

- [ ] `carbon-components`

## Detailed description

Toggle, ToggleSmall

> see changed?

Chrome

The latest

> should be aware of?

https://github.com/data-driven-forms/react-forms will be used in https://github.com/ManageIQ/manageiq-ui-classic
No date

1. Check documentation for Toggle/ToggleSmall

https://react.carbondesignsystem.com/?path=/docs/toggle--skeleton

## Additional information

- Notes

I will make a PR myself.

rvsia:
Closes #6810

Change types of `label`s in Toggles to `React.node`

#### Changelog

**Changed**

- A `node` can be used as `labelText`, `labelA` and `labelB` in `Toggle`, `ToggleSmall`

#### Testing / Reviewing

```jsx
<Toggle 
  id="toggle"
  labelText={<small>Very small label here</small>}
  aria-label="Some-aria-label" // or use default "toggle"
/>
```


tw15egan (approved):

dakahn:
@rvsia Is this specifically so you can use a `small` html tag inside a label? 

And for Carbon Team -- do we have precedent for handling styling like this? @carbon-design-system/design

tw15egan:
@dakahn for some background, we've changed other label `propTypes` from `string` to `node` to be more accommodating, since some users may need to render tooltips or even just spans for conditional styling.

rvsia:
@dakahn No, the using of the `small` element is just an example. Actually in our project, we are using a label with tooltip, but that would much more complex to show. 😄 

Another good use case is using of libraries like `react-intl` where elements are used to translate texts.

And as @tw15egan said, in all other components `labelText` is already `node`.

joshblack:
I think this is a good example of something we would want to restrict, given that it's being rendered in a `<label>` right @tw15egan?

I think cases for custom `<span>`s could come up, but often we see this use-case come up when folks want to add interactive elements inside of `<label>`, or other element types, which will have their meaning stripped IIRC. [Link](Interactive_content)

dakahn:
The reason I asked is because if it's a styling concern (like in your example) we can address that. That said allowing any node isn't what we want here. Generally interactive elements in particular _should not_ be used inside a label as that label (even explicit or wrapping labels) will essentially break.

[I've coded up an example demonstrating this here](https://codepen.io/dakahn/full/ZEWoyEr). Using NVDA I focus The label/button (which would be unexpected to the user) 
image
and it is announced as 'Enter name; button' 
image
When I move focus to the textbox it is programmatically unlabeled and announced as only (and confusingly) "edit; blank"

Contrast that with the expected behavior of hitting tab and having the textbox focus and announce itself as "Enter name; edit; blank"
image

It's worth noting as well that in the[ MDN article for label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) they specifically call this out (thanks @joshblack)

> Don't place interactive elements such as anchors or buttons inside a label. Doing so makes it difficult for people to activate the form input associated with the label.

This _can_ be mitigated with the `aria-label` attribute on the input, but adding aria to compensate for bad HTML is -- inadvisable.

tw15egan:
I was just going for consistency https://github.com/carbon-design-system/carbon/blob/459f0fa464a304f8e044a4191e1400b349ddecc4/packages/react/src/components/TextInput/TextInput.js#L217 , https://github.com/carbon-design-system/carbon/blob/459f0fa464a304f8e044a4191e1400b349ddecc4/packages/react/src/components/Checkbox/Checkbox.js#L107

Wouldn't DAP error if a user tries to enter an interactive element in there?

dakahn:
Actually no 😭 In my testing DAP, Accessibility Checker, and Axe all missed this -- a great example of automated accessibility tools being good "low hanging fruit" detectors, but unreliable for anything else.

joshblack:
_Edit: whoops, just saw @dakahn's comment_

---

@tw15egan unfortunately not that I can see 😕

Quick test:

image

And then the a11y tree:

image

Definitely get what you mean by consistency, this is definitely something that exists in the codebase 100%.

tw15egan:
OOOF, then this should probably be reverted across the codebase for the next major version 😬 

Thanks for filling me in!

rvsia:
Oh, I understand your concerns. However, it’s actually impossible to follow the Carbon guideline on providing help information as it requires to put a tooltip in a label https://www.carbondesignsystem.com/patterns/forms-pattern#offering-help .

dakahn:
I'll unblock since we already allow this elsewhere. We can make a note to lock this down in v11 👍

dakahn (approved):

PR #6827, opened 2020-09-11
fix(stylelint-config-carbon): disable custom property pattern

### LINKED ISSUE #6812: Use of --ibm as a custom property name prefix may not be suitable for all
https://github.com/carbon-design-system/carbon/blob/9470f86386ae6e4eb00113b99a10f3ff69bb7b1a/config/stylelint-config-carbon/rules/limit-language-features.js#L56

```
    // Custom property
    // Specify a pattern for custom properties.
    'custom-property-pattern': 'ibm-.+',
```

Any products using the carbon config will be forced to prefix custom properties with 'ibm' this will confuse these properties with carbons own properties.

Alternatively or in addition to updates we should tell the user what the pattern is e.g. 
```
    // Custom property
    // Specify a pattern for custom properties.
    'custom-property-pattern': [
      'ibm-.+',
      {
        message: "custom-property-pattern: Carbon expects '--ibm' prefix'",
      },
    ],
```

joshblack:
Closes #6812 

Disables the custom property pattern we have in stylelint

#### Changelog

**New**

**Changed**

- Disables the custom property pattern we have in stylelint

**Removed**


tw15egan (approved):
👍

andreancardona (approved):
@joshblack 🎉

PR #6842, opened 2020-09-15
docs(DataTable): add playground story

emyarod:
This PR adds a playground story to the DataTable docs so that props like `useZebraStyles`, `size`, and `stickyHeader` can be tested and applied. This is useful for devs and designers when confirming positioning and alignment issues in tables

#### Changelog

**New**

- DataTable Playground story

**Changed**

- add storybook knobs to the default table stories

#### Testing / Reviewing

Ensure the storybook knobs for the default data table stories are correct as well as the playground story


packages/react/src/components/DataTable/DataTable-story.js
```diff
@@ -51,7 +63,7 @@ export default {
 };
 
 export const Usage = () => (
-  <DataTable rows={rows} headers={headers}>
+  <DataTable rows={rows} headers={headers} {...props()}>
```
joshblack:
```suggestion
  <DataTable rows={rows} headers={headers}>
```

packages/react/src/components/DataTable/DataTable-story.js
```diff
@@ -140,7 +152,7 @@ export const BasicTable = () => {
 };
 
 export const WithOverflowMenu = () => (
-  <DataTable rows={rows} headers={headers}>
+  <DataTable rows={rows} headers={headers} {...props()}>
```
joshblack:
```suggestion
  <DataTable rows={rows} headers={headers}>
```

packages/react/src/components/DataTable/DataTable-story.js
```diff
@@ -177,7 +189,7 @@ export const WithOverflowMenu = () => (
 );
 
 export const WithToolbar = () => (
-  <DataTable rows={rows} headers={headers}>
+  <DataTable rows={rows} headers={headers} {...props()}>
```
joshblack:
```suggestion
  <DataTable rows={rows} headers={headers}>
```

packages/react/src/components/DataTable/DataTable-story.js
```diff
@@ -302,7 +314,7 @@ export const WithCheckmarkColumns = () => {
   ];
 
   return (
-    <DataTable rows={rows} headers={headers}>
+    <DataTable rows={rows} headers={headers} {...props()}>
```
joshblack:
```suggestion
    <DataTable rows={rows} headers={headers}>
```

packages/react/src/components/DataTable/DataTable-story.js
```diff
@@ -356,3 +368,5 @@ export const WithCheckmarkColumns = () => {
     </DataTable>
   );
 };
+
+export const Playground = () => require('./stories/default').default(props());
```
joshblack: We should try and inline the example here with knobs so that folks can see it in the story source 👀  Otherwise they just see the `require` statement

joshblack (changes requested):
Ah, totally blanked on the knobs. Thanks for adding in the playground example. Left a couple of comments, namely that for most stories we want people to be able to copy and paste them without any problems. The playground story is different though and doesn't need to follow this example since it uses knobs 👍

packages/react/src/components/DataTable/DataTable-story.js
emyarod: the default story is unused then, so I think we can just remove it

packages/react/src/components/DataTable/DataTable-story.js
```diff
@@ -356,3 +368,45 @@ export const WithCheckmarkColumns = () => {
     </DataTable>
   );
 };
+
+export const Playground = () => (
+  <DataTable
+    rows={rows}
+    headers={headers}
+    {...props}
```
tw15egan:
```suggestion
    {...props()}
```

dakahn (approved):
Seems cool to me pending @tw15egan 's changes 🏄

tw15egan (approved):

joshblack (approved):

PR #6851, opened 2020-09-16
feat(ResponsiveHoc): hoc to make react responsive work with ssr

anthonyshittu:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/packages/react/src/components/ResponsiveHoc/ResponsiveHoc-story.js
+++ b/packages/react/src/components/ResponsiveHoc/ResponsiveHoc-story.js
@@ -1,0 +1,40 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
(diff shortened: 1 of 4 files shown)


joshblack:
Hi there @anthonyshittu! 👋 

Thanks so much for taking the time to make a contribution 🙏  For new work being contributed, we try to point people towards first creating an issue that we can use as a conversation around the contribution before any work begins.

For this specific contribution, we're unlikely to bring in helpers in this way unless they relate to a specific component or come up frequently in product. For the case of window resize, we'd recommend a custom hook in the project or to lean on a library that provides similar functionality.

Hope this makes sense, and thanks again for taking the time here!

joshblack:
Going to close this out based on the note above, let me know if you have any questions!

PR #6868, opened 2020-09-21
feat(FileUploaderDropContainer): validate filetypes from OS file picker

### LINKED ISSUE #6864: Accept prop is not working in fileUploader component on windows [bug]
Sandbox link: https://codesandbox.io/s/github/carbon-design-system/carbon/tree/master/packages/react/examples/drag-and-drop-file-uploader
image (1)
Though in the code it is mentioned that supported extensions are only "jpeg" or "png". I am still able to select .txt file in the File Upload component.
FYI: This issue is only visible in windows operating system, but works as expected in MAC.
Version: "carbon-components": "10.20.0",
    "carbon-components-react": "7.20.0",

emyarod:
Closes #6864

This PR validates file types when using the native OS file picker for the scenario where the user deliberate selects the option to choose all file types rather than remaining on the default OS file picker behavior of only allowing supported file types (see [screenshot in ticket thread](https://github.com/carbon-design-system/carbon/issues/6864#issuecomment-696223509) for more context). This PR also updates the demos to show examples of file type validation

#### Changelog

**New**

- validate file types for native OS file picker

**Changed**

- update component demos

#### Testing / Reviewing

Using the native OS file picker, select the option to pick all file types (rather than keeping the default option of only allowing supported file types) and select an invalid file. The demo should mark the file as invalid

### DIFF
--- a/packages/react/examples/drag-and-drop-file-uploader/src/index.js
+++ b/packages/react/examples/drag-and-drop-file-uploader/src/index.js
@@ -56,6 +56,23 @@ function ExampleDropContainerApp(props) {
       );
       return;
     }
+    // file type validation
+    if (fileToUpload.invalidFileType) {
+      const updatedFile = {
+        ...fileToUpload,
+        status: 'edit',
+        iconDescription: 'Delete file',
+        invalid: true,
+        errorSubject: 'Invalid file type',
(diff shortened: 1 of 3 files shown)


tw15egan (approved):
Working as expected 👍  ✅

aledavila (approved):

PR #6901, opened 2020-09-23
fix(button): restrict transitions to relevant properties

### LINKED ISSUE #6823: Cannot access button in a tooltip via keyboard navigation
When a button is the first focusable element in a tooltip, the button never gets focus when the tooltip is opened via keyboard navigation. If there are other interactive elements in the tooltip after the button, there is no way to put focus on them either using just the keyboard.

## Detailed description
Sandbox example: https://codesandbox.io/s/wizardly-easley-gqs06?file=/src/index.js
When I click on the tooltip with the mouse, it opens and focus goes to the button. However when I tab to the tooltip, it opens but focus does not go to the button and I cannot see a way to get focus on the button with the keyboard. 

I am also seeing this issue in the [storybook](https://react.carbondesignsystem.com/?path=/story/tooltip--default-bottom) when I specify the button as the selectorPrimaryFocus element.

```
```

Focus should go directly to the button when the user tabs to the tooltip and it opens

> What happened instead?
Focus doesn't go to the button and there is no way to put focus on the button with just the keyboard

1. Go to https://codesandbox.io/s/wizardly-easley-gqs06?file=/src/index.js
2. Tab to the tooltip
Note focus does not go to the button

Alternatively add '.bx--tooltip__footer Button' as the selectorPrimaryFocus in the storybook example at https://react.carbondesignsystem.com/?path=/story/tooltip--default-bottom and tab to the tooltip.

image

emyarod:
Closes #6823

This PR specifies the properties for Carbon button transitions so that they do not interfere with focus logic inside the interactive tooltips

#### Changelog

**Changed**

- target only specific properties for button `transition`

#### Testing / Reviewing

Set the button in the tooltip menu as the primary focus element and confirm that it is automatically focused when the tooltip is opened

### DIFF
--- a/packages/components/src/components/button/_mixins.scss
+++ b/packages/components/src/components/button/_mixins.scss
@@ -30,7 +30,10 @@
   border-radius: $button-border-radius;
   outline: none;
   cursor: pointer;
-  transition: all $duration--fast-01 motion(entrance, productive);
+  transition: background $duration--fast-01 motion(entrance, productive),
+    box-shadow $duration--fast-01 motion(entrance, productive),
+    border-color $duration--fast-01 motion(entrance, productive),
+    outline $duration--fast-01 motion(entrance, productive);
 
   &:disabled,
   &:hover:disabled,


joshblack:
bump @aledavila @tw15egan when you all have a sec today 👀

tw15egan (approved):
Confirmed keyboard nav now focuses on the correct element 👍 ✅

aledavila (approved):

PR #6916, opened 2020-09-25
fix(Search): add `disabled` prop

### LINKED ISSUE #6915: Search component missing `disabled` prop and definition
- [ ] `carbon-components`

## Detailed description

The library already has disabled styles for the Search component but the prop definition and storybook knob are missing. Even though adding the `disabled` prop to the component currently works as expected since it's being spread into the underlying `input`, it would be best to document this prop in code and on Storybook

Search

emyarod:
Closes #6915

This PR adds an explicit `disabled` prop and prop type definition to the Search component and updates the corresponding stories. Even though the prop currently works as expected, we should document this prop separately and add it to the storybook

#### Changelog

**New**

- `disabled` prop on Search component and stories

#### Testing / Reviewing

Confirm that the Search component disabled state is still applied when passing in the `disabled` prop, and that there are no style regressions due to missing or incorrect selectors (the styles are already in the library)


tw15egan:
Should the search icon use `disabled` token when disabled? 

Screen Shot 2020-09-25 at 10 33 46 AM

aagonzales:
Yes, the icon should also be `disabled-02`

emyarod:
I don't think the component currently changes the icon color when you add the `disabled` prop so this isn't a regression but I can make that change here as well

tw15egan:
Looks great! Just seeing a tiny flicker of focus if you click directly on the `x` when it is disabled

search

tw15egan (approved):
Looks great! 👍 ✅

joshblack:
bump @andreancardona when you get a sec today 👀

andreancardona (approved):
@emyarod looks good :)

aagonzales (approved):

PR #6920, opened 2020-09-28
fix(storybook): label text prop name for the react select component

sanjitbauli:
In the storybook the prop for Label Text says helperText

Screenshot 2020-09-28 at 1 45 34 PM

Changes:
The pr will fix the prop name and change it to labelText instead of helperText which is incorrect.


emyarod (approved):
looks good to me

joshblack (approved):

PR #6928, opened 2020-09-29
feat(Tabs): create scrollable Tabs variant

### LINKED ISSUE #6798: 10.19 broke tab styles

emyarod:
Closes #6798
Ref #4758

This PR namespaces the styles and adds conditionals throughout the Tabs methods and props to support compatibility for both scrollable and non-scrollable/dropdown Tabs variants

#### Changelog

**New**

- namespaced styles for nonscrollable/dropdown and scrollable variants
- `scrollable` prop to toggle scrollable Tabs variant

**Changed**

- add deprecation notices throughout the component, styles, tests, and stories

#### Testing / Reviewing

Ensure the scrollable and dropdown/non-scrollable tabs function as expected (for scrollable refer to current stable storybook, and for dropdown/non-scrollable refer to a Carbon version prior to #6410)

### DIFF
--- a/packages/components/src/components/tabs/_tabs.scss
+++ b/packages/components/src/components/tabs/_tabs.scss
@@ -22,131 +22,126 @@
     @include reset;
     @include type-style('body-short-01');
 
-    display: flex;
+    position: relative;
     width: 100%;
     height: auto;
-    min-height: rem(40px);
     color: $text-01;
+    @include carbon--breakpoint(md) {
+      min-height: rem(40px);
+      background: none;
+    }
   }
 
   .#{$prefix}--tabs--container {
-    min-height: rem(48px);
+    @include carbon--breakpoint(md) {
+      min-height: rem(48px);
+    }
   }
 
-  .#{$prefix}--tabs__nav {
+  .#{$prefix}--tabs-trigger {
     display: flex;
-    flex-direction: row;
-    width: auto;
-    max-width: 100%;
-    margin: 0;
-    padding: 0;
-    overflow: auto hidden;
-    list-style: none;
-    transition: max-height $duration--fast-01 motion(standard, productive);
-
-    // hide scrollbars
-    scrollbar-width: none;
-
-    &::-webkit-scrollbar {
+    align-items: center;
+    justify-content: space-between;
+    height: rem(40px);
+    padding: 0 $spacing-09 0 $spacing-05;
+    color: $text-01;
+    background-color: $field-01;
+    border-bottom: 1px solid $ui-04;
+    outline: 2px solid transparent;
+    cursor: pointer;
+    @include carbon--breakpoint(md) {
       display: none;
     }
   }
 
-  //-----------------------------
-  // Overflow Nav Buttons
-  //-----------------------------
-  .#{$prefix}--tabs__overflow-indicator--left,
-  .#{$prefix}--tabs__overflow-indicator--right {
-    z-index: 1;
-    flex: 1 0 auto;
-    width: $carbon--spacing-03;
+  .#{$prefix}--tabs-trigger:focus,
+  .#{$prefix}--tabs-trigger:active {
+    @include focus-outline('outline');
   }
 
-  .#{$prefix}--tabs__overflow-indicator--left {
-    margin-right: -$carbon--spacing-03;
-    background-image: linear-gradient(to left, transparent, $ui-background);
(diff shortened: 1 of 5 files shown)


packages/components/src/components/tabs/_tabs.scss
```diff
@@ -18,370 +18,790 @@
...
-    @include type-style('body-short-01');
-
-    display: flex;
-    width: 100%;
-    height: auto;
-    min-height: rem(40px);
-    color: $text-01;
-  }
+  // TODO: remove block in next major release
+  .#{$prefix}--tabs--stable {
```
cal-smith: This would be a major breaking change for us (Angular). This would result in none of the tabs styles applying to our current tabs implementation. Given that you're already scoping the scrollable tabs (`bx--tabs-scrollable`) I believe it should be sufficient to leave `bx--tabs` un-scoped and just override classes as needed within the `bx--tabs-scrollable` scope.

cal-smith (changes requested):
Hey 👋 I've been watching the progress on this issue, and would like to help where I can to get this released as soon as possible. We have a number of consumers that would like to update to a newer `carbon-components` release and this is currently blocking them from that.

packages/components/src/components/tabs/_tabs.scss
emyarod: good catch, I think this was left over from the other PR which will probably be closed in favor of this one

emyarod: @cal-smith thanks again for taking a look, it's great to have your feedback with regression and compatibility testing

tw15egan:
Awesome work getting this up, just noticed a small style issue on the scrollable version in Safari while reviewing. 

Screen Shot 2020-09-29 at 10 48 09 AM

Also noticed some focus issues with the Dropdown variant, but those were present before the scrollable tabs implementation so that can be fixed in another PR. I don't want to hold this fix up 👍

cal-smith (approved):
LGTM on the compatibility front! Thanks for getting this sorted! 😁

joshblack:
Do we need to offer backwards-compat in the React code? It seems like the styles should be sufficient and then we wouldn't need to add conditions for scrollable in several places for `Tabs`/`Tab` that will need to be removed.

emyarod:
yeah we need to add conditions because the internal component method signatures are changed for the scrollable variant. I don't believe that styles are sufficient to switch between the dropdown and scrollable tab behavior. and I also think it would be best not to call any methods associated with scrollable tabs when the dropdown tabs are being used (or vice versa)

emyarod:
image

@tw15egan I'm not seeing that issue in Safari, can you add some more details about how I might be able to replicate that bug?

tw15egan:
Eh, I'm running the MacOS Big Sur beta and noticed an issue that was fixed in a different issue on the new Safari 14.0.0 but was present on Safari 13.1.2. Chalking it up to that 👍

tw15egan (approved):

emyarod:
I see, does the issue also appear on the current version of Tabs in Safari 14?

joshblack:
I think what I'm not understanding is why this needs to be a variant on the React component side. It seemed like the React component should just use the new styles and shouldn't introduce a variant because the scrollable work is the new default for Tabs.

The issue with replacing the styles was that it broke framework implementations, so it seemed like if we had both styles that it would solve that issue for them but then in the React implementation we just use the new styles by default and not worry about adding a variant.

emyarod:
oh I see so we make no changes to our JS and just scope the styles? I guess having the component changes was useful for regression testing but I can remove the backwards compatibility in the JS

joshblack:
@emyarod yeah! Exactly, I think that's all that we need to do to fix for framework implementations 👀

joshblack (approved):

tw15egan (approved):

cal-smith:
Thanks all!

tw15egan:
@cal-smith pushed up an `rc-1`, can you verify this resolves your issue?

cal-smith:
@tw15egan just checked it out, looks like everything is sorted!

Issue #6792, opened 2020-09-08
Recommendations for TreeView accessibility


carmacleod:
Recommendations for improving accessibility for the (experimental) [Carbon unstable_Treeview](https://carbon-components-react.netlify.app/?path=/story/unstable-treeview--default) component.

- [x] Add `aria-current="true"` to the "active" node (PR: #6937)
- [ ] Change our <kbd>Space</kbd> key behavior to match the **Recommended selection model** for **Selection in multi-select trees** in the [APG section for keyboard interaction in a TreeView](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-23) (i.e. <kbd>Space</kbd> toggles selection, <kbd>Enter</kbd> activates)
- [ ] Multi-select contiguous with _mouse_: Add <kbd>Shift + click</kbd> to select contiguous tree nodes from the last selected node to the currently Shift+clicked node (behavior can be tested in Mac Finder tree view, and Windows File Explorer's list view)
  - See also: [Add Shift+click selection to Carbon multi-select datatable](https://github.com/carbon-design-system/carbon/issues/4084)
- [ ] Multi-select contiguous with _keyboard_: Add <kbd>Shift + Space</kbd> to select contiguous tree nodes from the last selected node to the currently focused node
- [ ] Disabled items need to take focus (see the APG section on [Focusability of disabled controls](https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_disabled_controls) which specifically mentions Tree items in a Tree View) (PR: https://github.com/carbon-design-system/carbon/pull/11402)
- [x] Support <kbd>Home</kbd>, <kbd>End</kbd> and <kbd>Ctrl/Cmd+A</kbd> (See the APG's [Keyboard interaction in a TreeView](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-23)) (PR: https://github.com/carbon-design-system/carbon/pull/11328)
- [ ] #11859
- [ ] Consider also supporting optional keyboard selection shortcuts listed in the **Recommended selection model** for **Selection in multi-select trees** in the [APG section for keyboard interaction in a TreeView](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-23)
- [ ] Need to have design discussion about these (related) questions:
  - What happens when multiple items are selected and <kbd>Enter</kbd> is pressed? In Mac Finder tree view and Windows File Explorer list view, multiple files are opened. Would users expect that behavior?
[...]

PR #6937, opened 2020-09-29
feat(TreeNode): set aria-current on current active node

emyarod:
Refs #6792 

This PR couples `aria-current="true"` to the active tree node as an accessibility enhancement to the experimental treeview. The remaining points in the issue will be addressed at a later time when we are ready to reprioritize this effort but the current PR will serve as the demo for the upcoming ARIA deep dive on trees (part 3)

#### Changelog

**New**

- `aria-current` attribute for the active tree node

#### Testing / Reviewing

Confirm that the active tree node receives `aria-current="true"` and all other nodes do not have this attribute


dakahn (approved):

carmacleod:
@emyarod Please see: https://github.com/emyarod/carbon/pull/1

tw15egan (approved):

PR #6945, opened 2020-10-01
fix(content-switcher): provide empty default onChange prop

janhassel:
Prevents the ContentSwitcher from throwing an error when no `onChange` prop is passed yet. Demo of current behaviour (click on "two" or "three"): https://l4ijp.csb.app/

#### Changelog

**New**

- added `() => {}` as default onChange prop

**Changed**

- update PublicAPI snapshot to reflect this change

#### Testing / Reviewing

- Unit tests should make sure no functionality is affected
- Manual: Remove the `onChange` prop from the ContentSwitcher playground story and verify that it does not throw an error as it does in the codesandbox


tw15egan (approved):
👍 ✅

dakahn (approved):

PR #6954, opened 2020-10-01
fix(Modal): set default focus to secondary button

### LINKED ISSUE #6939: Modal: Set default primary focus to something besides primary button

- [ ] `carbon-components`

## Detailed description

Currently the default focused button on a modal is the primary button if it exists. In situations where users are interacting quickly with the UI, they might inadvertently trigger the primary button event which could include things like deleting.

Modal that has `passiveModal = false`

> see changed?

Set default primary focus to be either the secondary button or close button.

Brave Version 1.12.114 Chromium: 84.0.4147.135 (Official Build) (64-bit)

carbon-components-react 7.21.0-rc.0

> should be aware of?

Cloud IKS UI, no

asthas:
Closes #6939

Change default focus to secondary button

#### Testing / Reviewing

- Check storybook for the modal component
image

### DIFF
--- a/packages/react/src/components/Modal/Modal.js
+++ b/packages/react/src/components/Modal/Modal.js
@@ -429,14 +429,16 @@ export default class Modal extends Component {
         )}
         {!passiveModal && (
           <ButtonSet className={`${prefix}--modal-footer`}>
-            <Button kind="secondary" onClick={onSecondaryButtonClick}>
+            <Button
+              kind="secondary"
+              onClick={onSecondaryButtonClick}
+              ref={this.button}>
               {secondaryButtonText}
             </Button>
             <Button
               kind={danger ? 'danger' : 'primary'}
               disabled={primaryButtonDisabled}
-              onClick={onRequestSubmit}
-              ref={this.button}>
+              onClick={onRequestSubmit}>
               {primaryButtonText}
             </Button>
           </ButtonSet>


emyarod (commented):
rather than setting the secondary button to always be the primary focusable element by default, can we add the fallbacks as suggested in the original ticket?

asthas:
@emyarod The previous implementation follows fallbacks as `selectorPrimaryFocus || primary button || close button`

The changes I made change fallbacks as `selectorPrimaryFocus || secondary button || close button`

The same was mentioned in the issue.

emyarod:
not sure if we should directly search for the secondary button, but I think it would be better to still search for the primary button first before the secondary button. before anything else though, I just want to confirm the expected autofocus behavior with design cc @carbon-design-system/design

tw15egan:
I wonder if we can focus the secondary/close button if `danger` is set to true (only for `Danger` modals)?

asthas:
@tw15egan Yeah sure! I can try that . Can you please update  #6939 because it doesn't mention about danger modal?

tw15egan:
@asthas let's wait to get some design input before we do any more work 🙂

laurenmrice:
Briefly spoke with Andrew about this in slack, so posting here. I think we should go with this approach.
1. add secondary button in between primary button and close button in the order of fallbacks (so the order is custom selector -> primary -> secondary -> close)
2. if the current element to focus on is disabled, move to the next one in the order

chrisconnors-ibm:
@asthas -  If the secondary button has focus by default and receives the keyboard handler for return or enter, it is, by definition, the _Primary Button_.

More important than whether a secondary button should be the defacto Primary Button by intercepting the keyboard event that should be going to the Actual Primary Button by definition, but what action the design in question is assigning to each button. There's no reason the default button has to be action that confirms or takes the action the dialogs is referring to; Primary does not have to always be confirmation.

image

If the Primary button is something consequential or dangerous, it's not uncommon to add friction so that the likelihood of inadvertent  keyboard entry is reduced. You see this in patterns where user is forced to type a specific string before the dangerous or irreversible Primary Button is enabled.

tw15egan:
@chrisconnors-ibm I would post that in the actual issue, @asthas just volunteered to open a PR to add this in, not the original issue creator 👍

chrisconnors-ibm:
> @chrisconnors-ibm I would post that in the actual issue, @asthas just volunteered to open a PR to add this in, not the original issue creator 👍

Will do, wanted to get some friction as close to the merge button as possible.

joshblack:
Just wanted to bump this PR, is this something that is still being pursued or is it still blocked? 👀

emyarod:
I believe we can close this PR and possibly the original ticket if there is no response to the latest comment about modal action UX

aledavila:
I'm going to close this PR. If you have additional information feel free to open another issue :)

PR #6962, opened 2020-10-02
feat(Link): add noopener when target="_blank" is applied

### LINKED ISSUE #6910: Automatically set Link rel attribute to "noopener noreferrer" if target is "_blank"
### Summary / Justification

According to [web.dev](https://web.dev/external-anchors-use-rel-noopener/), setting the `target` attribute to "_blank" for a link without "rel=noopener" opens a site to cross-origin exploits.

This issue proposes an "invisible" feature that automatically sets the "rel" attribute to "noopener noreferrer" only if `target` is "_blank".

In the React implementation:

```jsx
<Link
  href="<external link that could be compromised>"
  target="_blank">
  Text
</Link>

// renders as
<a
  href="<external link that could be compromised>"
  class="bx--link"
  target="_blank"
  rel="noopener noreferrer">
  Text
</a>
```

The `rel` attribute can always be overridden by the consumer:

```jsx
<Link
  href="<external link that could be compromised>"
  target="_blank"
  rel={"prefetch"}>
  Text
</Link>
```

emyarod:
Closes #6910

This PR automatically adds `rel="noopener"` to Carbon `<Link />`s when `target="_blank"` is applied. Not sure if we want to break this out into an `ExternalLink` component but this PR serves as a starting point to implement this feature

#### Testing / Reviewing

Confirm that `rel="noopener"` is automatically added to links that also have `target="_blank"`

### DIFF
--- a/packages/react/src/components/Breadcrumb/__tests__/__snapshots__/Breadcrumb-test.js.snap
+++ b/packages/react/src/components/Breadcrumb/__tests__/__snapshots__/Breadcrumb-test.js.snap
@@ -24,6 +24,7 @@ exports[`Breadcrumb should render 1`] = `
             <a
               className="bx--link"
               href="www.carbondesignsystem.com"
+              rel={null}
             >
               Breadcrumb 1
             </a>
@@ -55,6 +56,7 @@ exports[`Breadcrumb should support rendering the current page 1`] = `
             <a
               className="bx--link"
               href="#a"
+              rel={null}
             >
               A
             </a>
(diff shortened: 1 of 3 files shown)


tw15egan:
Do we also need to have it add `noreferrer`? 

I'm good with handling it this way, instead of making people use a new component. Especially since `target="_blank"` isn't a great practice anyways 🤷

emyarod:
the recommendation provided in the issue says `noopener` or `noreferrer`, and since we target modern browsers I went with `noopener`

tw15egan:
Ah nice! I thought you needed to have both. Looks great 👍

tw15egan (approved):

joshblack:
bump @aledavila when you get a chance today 👀

aledavila (approved):

Issue #6143, opened 2020-05-27
the color of the grids do not follow the focus move


pankaj-kumar34:
In chrome, under Service credentials section, use the Tab and Shift+Tab keys to navigate to and back ,when the focus left ,the dialog and button lost. please refer to below .
WCAG 2.1:Use the Tab and Shift+Tab keys to navigate to and back between widgets
Checkpoint 4.1.2 reflow：https://www.ibm.com/able/guidelines/ci162/name_role_value.html
![image](https://user-images.githubusercontent.com/18532515/82975453-1ae3d280-9ffa-11ea-833d-9d106da0ea17.png)

emyarod:
can you create a [reduced test case in Code Sandbox](https://codesandbox.io/s/github/carbon-design-system/carbon/tree/master/packages/react/examples/codesandbox) for easier testing and debugging?

pankaj-kumar34:
https://codesandbox.io/s/awesome-bassi-n49hv?file=/src/index.js

Focus issue, we are able to make it work but when we switch tab using arrow button then the values are not coming as expected.

dakahn:
Can you recreate the issue in that sandbox you posted or is this just on the site you're testing? I'm unable to recreate this bug. Focus follows selected switcher button

pankaj-kumar34:
In image below last printed value is index=2, which is value for _Month_ switch button. But Focused button is _Day_.
![image](https://user-images.githubusercontent.com/18532515/95043510-91a48100-06fa-11eb-961e-7c2e3c4cbc0c.png)

tw15egan:
related: https://github.com/carbon-design-system/carbon/issues/6162

PR: https://github.com/carbon-design-system/carbon/pull/6968

joshblack:
Closed by: https://github.com/carbon-design-system/carbon/pull/6257

PR #6257, opened 2020-06-11
fix(content-switcher): call onChange with correct data with keyboard

joshblack:
Closes #6119 

#### Changelog

**New**

**Changed**

- Update content switcher to call `onChange` with correct data when using keyboard
- Added tests for behavior to prevent regression

**Removed**

PR #6968, opened 2020-10-04
fix(ContentSwitcher): set onChange event data correctly

### LINKED ISSUE #6162: ContentSwitcher selection with arrow keys
## Environment

> Operating system
MacOS

> Browser
Chrome, FireFox

## Detailed description

10

When using the left/right arrow keys to navigate a ContentSwitcher I expected the correct button to be selected with focus on the desired button. 

> What happened instead?
For example:
I tabbed to the first ContentSwitcher item and pressed the right arrow key. The index passed to the onChange event was the index of the currently selected button so the content did not change as the button to the right was not selected. However, focus still moved to that button.
The issue is the same when using the left arrow key. Every subsequent arrow press means focus is on the wrong button and does not align with the content displayed.

1. Tab to ContentSwitcher
2. Press either the left or right arrow key

## Additional information

This is what's captured in the onChange function (see _ref) when clicking on the desired button (Sample Code button)
image

This is what's captured in the onChange function (see _ref) when trying to move off the Sample Code button to another on it's left. (pressing the left arrow key on that button)
image

I would expect the last scenario to say `index: 0` 

Related issue: https://github.ibm.com/mhub/qp-planning/issues/5510

motou:
Closes #6162 

The `onChange` event data by pressing arrow keys should be the same as clicking on the previous/next switch.

#### Changelog

**Changed**

- set the `onChange` event data correctly by pressing arrow keys 

#### Testing / Reviewing

In the https://codesandbox.io/s/sad-moon-ji8u0 the event data should be the same with pressing arrow keys or clicking.

### DIFF
--- a/packages/react/src/components/ContentSwitcher/ContentSwitcher.js
+++ b/packages/react/src/components/ContentSwitcher/ContentSwitcher.js
@@ -98,7 +98,11 @@ export default class ContentSwitcher extends React.Component {
           () => {
             const switchRef = this._switchRefs[nextIndex];
             switchRef && switchRef.focus();
-              this.props.onChange({ index: nextIndex, name: this.props.children[nextIndex].props.name, text: this.props.children[nextIndex].props.text });
+            this.props.onChange({
+              index: nextIndex,
+              name: this.props.children[nextIndex].props.name,
+              text: this.props.children[nextIndex].props.text,
+            });
           }
         );
       }
--- a/packages/react/src/components/ContentSwitcher/ContentSwitcher.js
+++ b/packages/react/src/components/ContentSwitcher/ContentSwitcher.js
@@ -97,7 +97,7 @@ export default class ContentSwitcher extends React.Component {
           () => {
             const switchRef = this._switchRefs[nextIndex];
             switchRef && switchRef.focus();
-            this.props.onChange(data);
+              this.props.onChange({ index: nextIndex, name: this.props.children[nextIndex].props.name, text: this.props.children[nextIndex].props.text });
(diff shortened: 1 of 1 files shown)


emyarod (commented):
thanks for jumping in! it looks like you need to run `prettier` to format the changed file in order to pass CI

but to address one of the points in the original issue, since this would technically be a breaking change for users, would we be fine with this change in between major releases? cc @carbon-design-system/developers-system

tw15egan:
This will also close https://github.com/carbon-design-system/carbon/issues/6143

tw15egan:
I think this is a similar fix to https://github.com/carbon-design-system/carbon/pull/6257

cc @joshblack

motou:
Yes almost the same 😅 but the other fix uses `...data`, which introduces unwanted attribute `key`. That is not what the user expects. See https://github.com/carbon-design-system/carbon/pull/6257/files#r500786562

I agree that it belongs to breaking changes. Moreover, I am thinking if the `onKeyDown` event of the original / destination tab should be triggered as well, just like the mouse event: firstly child component Switch `onClick` event, then ContentSwitcher `onChange` event.

joshblack:
For context, the spread on `data` was to try and match the existing implementation and try and ship the correct one at the same time (e.g. currently it does `this.props.onChange(data)` and so the PR was attempting to maintain that functionality)

Unfortunately, removing that `data` would be a breaking change (as noted above) and we wouldn't be able to ship this until v11 😞

dakahn:
@motou bump

PR #6969, opened 2020-10-05
fix(ProgressStep, StructuredList): fix exports

### LINKED ISSUE #6961: Inconsistent imports for multiple components
## Inconsistent import for multiple components

Some components cannot be automatically import as their import differs from others.

### Describe in detail the issue you're having.

In Data Driven Forms we are using Babel's `transform-imports` to include different versions of imports for CJS/ESM builds. For most Carbon components it's actually pretty straightforward process as the export structure is clear:

```jsx
// .../Component/Component.js

...

export default Component;
```

To transform lines like 
```jsx
import { TextInput } from 'carbon-components-react';
```

we are using this config

```jsx
    'carbon-components-react': {
      transform: (importName) => {
        let res;
        const files = glob.sync(path.resolve(__dirname, `../../node_modules/carbon-components-react/${env === 'cjs' ? 'lib' : 'es' }/**/${importName}.js`));
        if (files.length > 0) {
          res = files[0];
        } else {
          throw new Error(`File with importName ${importName} does not exist`);
        }
        res = res.replace(path.resolve(__dirname, '../../node_modules/'), '');
        res = res.replace(/^\//, '');
        return res;
      },
      preventFullImport: false,
      skipDefaultConversion: false
    },
```

And it works well. However, there are several components, that break the consistency and because of that, the transformation isn't working. I found two: ProgressIndicator and StructuredList. (Maybe there are more, I wasn't exploring the rest of library.)

```jsx
// ../ProgressIndicator/ProgressIndicator.js

export const ProgressIndicator;
export const ProgressStep;
```

```jsx
// ../StructuredList/StructuredList.js

export const StructuredListWrapper;
export const StructuredListCell;
export const StructuredListRow;
export const StructuredListBody;
```

So the absence of the default import and combination of multiple components in one file break the consistency.

### Is this issue related to a specific component?

`ProgressIndicator`
`StructuredList`

### What did you expect to happen? What happened instead? What would you like to see changed?

I would like to change the structure to be in line with all other components:

```jsx
// ../ProgressIndicator/ProgressIndicator.js

export default ProgressIndicator;
```

```jsx
// ../ProgressIndicator/ProgressStep.js

export default ProgressStep;
```

```jsx
// ../StructuredList/StructuredListWrapper.js

export default StructuredListWrapper;
```

```jsx
// ../StructuredList/StructuredListCell.js

export default StructuredListCell;
```

```jsx
// ../StructuredList/StructuredListRow.js

export default StructuredListRow;
```

```jsx
// ../StructuredList/StructuredListBody.js

export default StructuredListBody;
```

Of course, it's possible to **keep** the old exports in files to be backwards compatible. No issues there.

### What version of the Carbon Design System are you using?

`"carbon-components-react": "^7.20.0",`

### What offering/product do you work on? Any pressing ship or release dates we should be aware of?

https://github.com/data-driven-forms/react-forms/

1. use the babel plugin to transform `import { ProgressStep } from 'carbon-components-react';` 

## Additional information

I have no problem with creating a PR to solve this issue.

cc @Hyperkid123

rvsia:
Closes #6961

- make default export for each sub-component
- add hook for error messages

#### Changelog

**Changed**

- `ProgressStep`, `StructuredListBody`, `StructuredListCell`, `StructuredListHead`, `StructuredListInput`, `StructuredListRow` and `StructuredListWrapper` are moved to separate files with default imports.

#### Testing / Reviewing

- [ ] import one of the mentioned components, see no differece
- [ ] import one of them from the old paths (`import { ProgressStep } from '.../ProgressIndicator/ProgressIndicator'` or `import { StructuredListBody } from '.../StructuredList/StructuredList'`) with `NODE_ENV` not set to `production` and see following error:

image


rvsia:
@tw15egan any thoughts?

joshblack:
Hey @rvsia! 👋  So sorry about the delay, last week was a planning week for our group and things were less active than they usually are.

Looking at the parent issue and the PR, I think the preference would be to have people pull components off the entrypoint instead of referring to file paths. If there is an issue with tree-shaking, then definitely let us know and we can take a look!

For context, sometimes there is a preference to have multiple components in the same file and what we want to commit to maintaining (through semver) will be the entrypoint versus the paths of the files in the project.

In the past, we did offer a plugin that tried to re-write the component paths when tree-shaking was failing (I think this was when webpack 2/3 were common and tree-shaking was flaky) where we ultimately kept track of the path of components for teams to avoid having components set as default exports: https://unpkg.com/browse/babel-plugin-transform-carbon-imports@9.1.2/paths.js Unfortunately as more of IBM shifted over to webpack 4, and tree-shaking improved, we saw less of a need for this style of plugin.

Hope this makes sense!

rvsia:
@joshq Thanks for the response!

> I think the preference would be to have people pull components off the entrypoint

The entrypoint still exists as `COMPONENT/index.js`, I just think it's always a good idea to have a consistent file structure. (Don't mix named/default imports, use the same case format, etc.)

But I understand that you don't need this change in your codebase. We can close this PR and return to the issue to find a better solution.

joshblack:
@rvsia thanks so much for understanding! Really appreciate it 🙏

PR #6974, opened 2020-10-05
fix(Tabs): select tab do not vertical scroll into view

### LINKED ISSUE #6973: Tabs component scrolls tabs into view on click, no way to stop this behaviour

- [ ] `carbon-components`

## Detailed description

When using the tabs and tab component, selecting on a tab (click or keyboard) the component will scroll the window such that the tabs component is at the top.

Tabs, tab

> see changed?

This causes unwanted behaviour in my usecase, there should be a way to turn this off, perhaps via a boolean prop.

Firefox, chrome

Latest for all packages (x.21)

> should be aware of?

IBM Cloud Docs

1. Go to https://react.carbondesignsystem.com/?path=/story/tabs--container
2. Open the storybook addons panel (Keyboard press 'A', or via the storybook menu)
3. Drag the addons section such that it covers half the container portion
4. Scroll the panel such that it's at the very top (i.e. showing a gap between the tabs and the top of the storybook container)
image
5. Click on the 2nd tab, it is now scrolled to the top
image

## Additional information

This line here is causing the issue: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Tabs/Tabs.js#L232

Looks like it's scrolling both x and y axis when it should only be scrolling x. I can open a PR to fix if needed.

gptt916:
Closes #6973

Prevent vertical scrolling when selecting tabs

#### Changelog

No longer vertically scrolls on new tab selection.

Use regular (bool) argument for the scrollIntoView api instead of the experimental (object) type argument, ref: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

#### Testing / Reviewing

1. Go to the storybook and open the storybook addons panel (Keyboard press 'A', or via the storybook menu)
2. Drag the addons section such that it covers half the container portion (so the canvas is scrollable)
3. Scroll the panel such that it's at the very top (i.e. showing a gap between the tabs and the top of the storybook container)
image
4. Click on the 2nd tab, it should no longer scroll vertically

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.js
+++ b/packages/react/src/components/Tabs/Tabs.js
@@ -229,7 +229,7 @@ export default class Tabs extends React.Component {
       event.type === 'click'
     ) {
       const currentScrollLeft = this.state.tablistScrollLeft;
-      tab?.tabAnchor?.scrollIntoView({ inline: 'nearest' });
+      tab?.tabAnchor?.scrollIntoView(false, { inline: 'nearest' });
       const newScrollLeft = this.tablist.current.scrollLeft;
       if (newScrollLeft > currentScrollLeft) {
         this.tablist.current.scrollLeft += this.OVERFLOW_BUTTON_OFFSET;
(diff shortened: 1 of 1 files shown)


emyarod (approved):
looks good to me, thanks for the fix

tw15egan (approved):
🎉  Thanks for the quick fix 👍 ✅

PR #6981, opened 2020-10-06
fix(tabs): remove role knob

tw15egan:
Removes the `role` knob, as this prop was removed in #6964 . This was causing the `role` to still appear in the storybook.


joshblack (approved):

carmacleod (commented):
+1

Thanks, @tw15egan !

[Edit: I did select "Approve", but I guess GitHub ignored me. 😄 ]

emyarod (approved):
looks good to me

PR #6989, opened 2020-10-07
fix(Tabs): improve screen reader support

### LINKED ISSUE #6984: Tabs control does not work with screen readers
The Tabs control doesn't work with JAWS or NVDA screen readers. They can sort of be used in reading/browse mode, but they don't recognize the ARIA tabs in focus mode, so they eat the arrow keys, which is not supposed to happen (i.e. if a screen reader recognizes a tablist with tab children, they will pass the arrow keys through to the control so that the user can switch tabs).
Note: Only tried on Windows so far, but I expect to see a similar problem on Mac. Also, didn't try vanilla Carbon Tabs - only react.

The problem is that the "tab" is not correctly marked up. The "tab" role needs to be on "the element that takes focus and does the key event handling". This is true of all interactive ARIA roles - the role needs to be on the element that the user will actually interact with.

In the case of the Carbon Tabs control, the interaction is taking place on the anchor element, but the list item element has all of the ARIA markup and tabindex="-1" as well.

i.e. a Carbon Tab currently looks like this:
```
<li role="tab" tabindex="-1" class="bx--tabs--scrollable__nav-item" aria-selected="false" aria-disabled="false" aria-controls="tab-1__panel" wiid="C11">
    <a id="tab-1" class="bx--tabs--scrollable__nav-link" href="#" tabindex="-1" wiid="C12">Tab label 1</a>
</li>
```

but it needs to _either_ look like the following (i.e. move all the things over to the anchor, and make the li presentational so that screen readers don't get confused by the li's implicit listitem role):
```
<li role="presentation" class="bx--tabs--scrollable__nav-item" wiid="C11">
    <a id="tab-1" role="tab" class="bx--tabs--scrollable__nav-link" href="#" tabindex="-1" aria-selected="false" aria-disabled="false" aria-controls="tab-1__panel" wiid="C12">Tab label 1</a>
</li>
```

OR, alternatively, a Carbon Tab can look like this (i.e. use a button instead of an anchor, so that the context menu for the tab doesn't have linky things in it like "Open link in new window" or "Copy link address"):
```
<li role="presentation" class="bx--tabs--scrollable__nav-item" wiid="C11">
    <button id="tab-1" role="tab" class="bx--tabs--scrollable__nav-link" tabindex="-1" aria-selected="false" aria-disabled="false" aria-controls="tab-1__panel" wiid="C12">Tab label 1</button>
</li>
```

OR, another alternative is to go completely fake and use divs (which is what the [APG](https://w3c.github.io/aria-practices/examples/tabs/tabs-1/tabs.html) did), which would be ok, too, but I'm not sure if that's as nice for "progressive enhancement" (i.e. no js, older browsers, etc).

Happy to work with you to get this working properly.

emyarod:
Closes #6984

This PR updates the `role` attribute on the tab anchor and moves the aria attributes from the parent list item element to the anchor as well. The anchor is also updated to be a button now to prevent things like "Open link in new window" or "Copy link address" from appearing in the context menu

#### Changelog

**Changed**

- deprecate `renderAnchor` (to be replaced with `renderButton`, unless we have a better name in mind)
- move role and aria attributes from list item to inner anchor
- convert anchor to button element

**Removed**

- outdated tests

#### Testing / Reviewing

Ensure the updated tabs match the existing tabs in behavior and appearance and confirm the updated markup matches the expected markup from the referenced issue

### DIFF
--- a/packages/react/src/components/Tab/Tab-test.js
+++ b/packages/react/src/components/Tab/Tab-test.js
@@ -33,16 +33,16 @@ describe('Tab', () => {
       expect(wrapper.props().role).toEqual('presentation');
     });
 
-    it('renders <a> with tabindex set to 0', () => {
+    it('renders <button> with tabindex set to 0', () => {
       expect(wrapper.find('button').props().tabIndex).toEqual(0);
     });
 
-    it('sets tabIndex on <a> if one is passed via props', () => {
+    it('sets tabIndex on <button> if one is passed via props', () => {
       wrapper.setProps({ tabIndex: 2 });
       expect(wrapper.find('button').props().tabIndex).toEqual(2);
     });
 
-    it('uses label to set children on <a> when passed via props', () => {
+    it('uses label to set children on <button> when passed via props', () => {
       expect(wrapper.find('button').props().children).toEqual('firstTab');
     });
 
--- a/packages/react/src/components/Tab/Tab-test.js
+++ b/packages/react/src/components/Tab/Tab-test.js
@@ -46,11 +46,6 @@ describe('Tab', () => {
       expect(wrapper.find('button').props().children).toEqual('firstTab');
     });
 
-    it('sets new href value when passed in via props', () => {
-      wrapper.setProps({ href: '#other-content' });
-      expect(wrapper.find('button').props().href).toEqual('#other-content');
-    });
-
     it(`should not have [className="${prefix}--tabs__nav-item--selected"] by default`, () => {
       expect(wrapper.hasClass(`${prefix}--tabs__nav-item--selected`)).toBe(
         false
--- a/packages/react/src/components/Tab/Tab-test.js
+++ b/packages/react/src/components/Tab/Tab-test.js
@@ -29,8 +29,8 @@ describe('Tab', () => {
       ).toBe(true);
     });
 
-    it('renders <li> with [role="tab"]', () => {
-      expect(wrapper.props().role).toEqual('tab');
+    it('renders <li> with [role="presentation"]', () => {
+      expect(wrapper.props().role).toEqual('presentation');
     });
 
     it('renders <a> with tabindex set to 0', () => {
--- a/packages/react/src/components/Tab/Tab-test.js
+++ b/packages/react/src/components/Tab/Tab-test.js
@@ -21,11 +21,11 @@ describe('Tab', () => {
       expect(wrapper.hasClass('extra-class')).toBe(true);
     });
 
-    it('renders <a> with expected className', () => {
+    it('renders <buttton> with expected className', () => {
       expect(
         // TODO: uncomment and replace assertion in next major version
-        // wrapper.find('a').hasClass(`${prefix}--tabs__nav-link`)
-        wrapper.find('a').hasClass(`${prefix}--tabs--scrollable__nav-link`)
+        // wrapper.find('button').hasClass(`${prefix}--tabs__nav-link`)
+        wrapper.find('button').hasClass(`${prefix}--tabs--scrollable__nav-link`)
       ).toBe(true);
     });
 
@@ -34,25 +34,21 @@ describe('Tab', () => {
     });
 
     it('renders <a> with tabindex set to 0', () => {
-      expect(wrapper.find('a').props().tabIndex).toEqual(0);
+      expect(wrapper.find('button').props().tabIndex).toEqual(0);
     });
 
     it('sets tabIndex on <a> if one is passed via props', () => {
       wrapper.setProps({ tabIndex: 2 });
-      expect(wrapper.find('a').props().tabIndex).toEqual(2);
+      expect(wrapper.find('button').props().tabIndex).toEqual(2);
     });
 
     it('uses label to set children on <a> when passed via props', () => {
-      expect(wrapper.find('a').props().children).toEqual('firstTab');
-    });
-
-    it('sets href as # by default', () => {
-      expect(wrapper.find('a').props().href).toEqual('#');
+      expect(wrapper.find('button').props().children).toEqual('firstTab');
     });
 
     it('sets new href value when passed in via props', () => {
       wrapper.setProps({ href: '#other-content' });
-      expect(wrapper.find('a').props().href).toEqual('#other-content');
+      expect(wrapper.find('button').props().href).toEqual('#other-content');
     });
 
     it(`should not have [className="${prefix}--tabs__nav-item--selected"] by default`, () => {
--- a/packages/react/src/components/Tab/Tab-test.js
+++ b/packages/react/src/components/Tab/Tab-test.js
@@ -21,7 +21,7 @@ describe('Tab', () => {
       expect(wrapper.hasClass('extra-class')).toBe(true);
     });
 
-    it('renders <buttton> with expected className', () => {
+    it('renders <button> with expected className', () => {
       expect(
         // TODO: uncomment and replace assertion in next major version
         // wrapper.find('button').hasClass(`${prefix}--tabs__nav-link`)
--- a/packages/react/src/components/Tab/Tab.js
+++ b/packages/react/src/components/Tab/Tab.js
@@ -179,7 +179,7 @@ export default class Tab extends React.Component {
         {renderButton ? (
           renderButton(buttonProps)
         ) : (
-          <button type="button" {...buttonProps}>
+          <button type="button" role="tab" {...buttonProps}>
             {label}
           </button>
         )}
--- a/packages/react/src/components/Tab/Tab.js
+++ b/packages/react/src/components/Tab/Tab.js
@@ -41,7 +41,7 @@ export default class Tab extends React.Component {
     /**
      * Provide a string that represents the `href` of the Tab
      */
-    href: PropTypes.string.isRequired,
+    href: deprecate(PropTypes.string),
 
     /**
      * The element ID for the top-level element.
--- a/packages/react/src/components/Tab/Tab.js
+++ b/packages/react/src/components/Tab/Tab.js
@@ -10,6 +10,7 @@ import React from 'react';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
 import TabContent from '../TabContent';
+import deprecate from '../../prop-types/deprecate.js';
 
 const { prefix } = settings;
 
@@ -72,7 +73,7 @@ export default class Tab extends React.Component {
      * Useful for using Tab along with react-router or other client
      * side router libraries.
      **/
-    renderAnchor: PropTypes.func,
+    renderAnchor: deprecate(PropTypes.func),
 
     /*
      * An optional parameter to allow overriding the content rendering.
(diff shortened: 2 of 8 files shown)


andreancardona (approved):
@emyarod looks good!

joshblack (approved):

packages/react/src/components/Tab/Tab-test.js
```diff
@@ -21,38 +21,34 @@ describe('Tab', () => {
       expect(wrapper.hasClass('extra-class')).toBe(true);
     });
 
-    it('renders <a> with expected className', () => {
+    it('renders <buttton> with expected className', () => {
```
carmacleod: probably want this to read `<button>` instead of `<buttton>` :)
```suggestion
    it('renders <button> with expected className', () => {
```

packages/react/src/components/Tab/Tab-test.js
```diff
@@ -21,38 +21,34 @@ describe('Tab', () => {
...
-    it('sets href as # by default', () => {
-      expect(wrapper.find('a').props().href).toEqual('#');
+      expect(wrapper.find('button').props().children).toEqual('firstTab');
     });
 
     it('sets new href value when passed in via props', () => {
       wrapper.setProps({ href: '#other-content' });
-      expect(wrapper.find('a').props().href).toEqual('#other-content');
+      expect(wrapper.find('button').props().href).toEqual('#other-content');
     });
```
carmacleod: You don't want to set the href of a button. :)
Woah, href is a prop? What is the use case for setting the href of a tab?

packages/react/src/components/Tab/Tab.js
```diff
@@ -121,7 +122,8 @@ export default class Tab extends React.Component {
       tabIndex,
```
carmacleod, line 122: I don't think tabIndex should be a prop... you don't want authors messing with the tab order of the Tabs, because you want to control the switching when the user types arrow keys...

carmacleod:
Still not working in screen readers. Still need to:
- delete `tabindex="-1"` from the `<li>`s (don't need them to be clickable because the buttons are clickable).
- add `role="tab"` to the `<button>` (because a `tablist` needs to have `tab` children).
- remove the `href` from the button (just to clean up)

Also, I looked more closely at the `tabpanels` this time. There's some work needed there as well - sorry I didn't look earlier.
- remove the `aria-live` attribute from the `tabpanel`. The user does not expect or want the contents of every tabpanel read to them when they are switching tabs in the tablist. They know that they can type the <kbd>Tab</kbd> key to go from the `tablist` to the current `tabpanel` if they want to go through the `tabpanel`'s contents.

- don't need to use `aria-hidden="true"/"false"` on the `tabpanel`, because HTML `hidden` takes care of hiding content from screen readers, so we can clean up some code here (less code always runs faster, and it's easier to read... ;) )

- either the `tabpanel` element itself, or "the first element containing meaningful content inside the tabpanel" must be focusable. In this example, which mostly has non-interactive paragraphs in the tabpanels, each `tabpanel` should have `tabindex="0"` so that the user can type <kbd>Tab</kbd> to go from the `tablist` into the current `tabpanel`. The 4th tabpanel of this example has some focusable content, but I'm not sure whether jumping to the "renderContent" button qualifies as "meaningful content"? I think that what the guidance means about "meaningful content" is more like if the `tabpanel` contains a form with form controls, then it's ok to let the focus go to the first form control instead of giving the `tabpanel` a `tabindex="0"`. This choice will likely need to be a new prop because it's app-specific logic - we can discuss.

packages/react/src/components/Tab/Tab.js
emyarod: it's mainly so that if the user chooses to set a custom tabindex, it is applied to the correct element rather than being spread onto the wrapper element. Typically the user isn't expected to provide a value for this since it's handled automatically by the component itself

emyarod:
I updated the PR with your suggestions @carmacleod . The only exception is the focus order change for the tabpanels, but I think we can make that change separately after discussing the pattern and user expectations more

packages/react/src/components/Tab/Tab.js
carmacleod: Regarding "custom tabindex":
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex
> Avoid using tabindex values greater than 0. Doing so makes it difficult for people who rely on assistive technology to navigate and operate page content.

- https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex
> Using a tabindex greater than 0 is considered an anti-pattern.

emyarod: in that case we can deprecate the prop to give consumers time to prepare for the prop removal

carmacleod:
This is way better! The browsers and screen readers now recognize this as a tablist with tabs! 🎉 
Now we can test the details (and we can look at adding the tabpanels to the focus order). 😄 

- Note that `div#tab-4__panel` is missing `role="tabpanel"` for some reason (which really messes up Firefox).

I will test more later.

emyarod:
> Note that `div#tab-4__panel` is missing `role="tabpanel"` for some reason (which really messes up Firefox).

that tab was rendering some example custom content so I've added the role attribute to that example

carmacleod:
> that tab was rendering some example custom content so I've added the role attribute to that example

It would be better if the component didn't leave the ARIA markup up to the author?

Would it be difficult to code it so that the custom content is added as a child of the tabpanel div? So in the initial state, the tabpanel div is there, but simply empty? It can be hidden until it's selected. (Is the visually-hidden necessary?)

So, instead of:
```
<div id="tab-4__panel" hidden="" aria-labelledby="tab-4" class="bx--visually-hidden"></div>
```

Would it be possible to mark the tabpanel div up like this, and maybe use the div.some-content to add the custom content?
```
<div role="tabpanel" id="tab-4__panel" hidden="" aria-labelledby="tab-4" class="bx--tab-content tab-content"> 
    <div class="some-content"></div> <!-- Either have this div here initially, or add as child when tab selected -->
</div>
```

(Note that we still need to discuss adding tabpanels to the tab order, so the above doesn't add `tabindex="0"`... yet. :)

emyarod:
> It would be better if the component didn't leave the ARIA markup up to the author?

it isn't left up to the author by default but if they require a custom tab wrapper, it is available to them as an option. what you proposed is a possibility, but I think that eliminates existing functionality for the consumer which would be a breaking change. we also have to evaluate whether or not we should restrict tab panel content to a `div` since it would limit the versatility of the component

carmacleod:
> I think that eliminates existing functionality for the consumer

Maybe it can be written in such a way the change is completely transparent to the consumer?

> we also have to evaluate whether or not we should restrict tab panel content to a div since it would limit the versatility of the component

The tabpanel itself is already a div, so maybe just add the content provided by the consumer/author as a child of the tabpanel?

**Edit:** Actually, I think there is something that I am completely not understanding here. The Tabs in this Tabs component are always rendered when selected (i.e. arrow to Tab, see content), so I'm not sure what the `renderContent` prop actually does?

carmacleod:
Finally had a chance to test the scrolling arrow buttons, and I think that having them in the tab order and visible to screen readers is definitely problematic. Please note the following points in @aagonzales's [keyboard navigation design](https://github.com/carbon-design-system/carbon/issues/4758#issuecomment-591662065):
- <kbd>tab</kbd>: traps focuses in the tab component starting at the select tab.
   - press <kbd>tab</kbd> again and focus will move out of the tab component and not the next tab inside the component.
   - clicking <kbd>tab</kbd> at any point while inside tab component will move focus to the next focusable item on page.
- Scrolling arrow buttons will be excluded from screen reader and keyboard arrow selections.
- <kbd>home</kbd> key takes the user to the first tab.
- <kbd>end</kbd> key takes the user to the last tab.

Together with the diagrams, and given that she was answering my scroll arrow button questions from https://github.com/carbon-design-system/carbon/issues/4758#issuecomment-589082541, I interpreted the first 2 design points above to mean that the scroll arrow buttons would not be in the tab order, and they would be hidden from screen readers.

We can do that by adding `aria-hidden="true" tabindex="-1"` to the scrolling arrow buttons (and then the SVG wouldn't need `aria-hidden="true" because descendants of an aria-hidden element are also aria-hidden).

Note also that right-click currently activates the scroll buttons, which feels really odd.

There's also a strange state that a keyboard user can get into where the tablist has the focus. We should make sure that doesn't happen. (maybe just taking the scrolling arrow buttons out of the tab order will fix that? I can test again afterwards).

**Edit:** Just realized I still need to test and think about mobile, and what happens in mobile screen readers if the scroll arrow buttons are taken out of the tab order. Argh. Are you allowed to conditionally take the scroll buttons out of the tab order on desktop and leave them in for mobile (if necessary)? Just asking, for now.

emyarod:
the `renderContent` prop allows the consumer to render any custom content they want to display. In our example we have rendered something similar to the default tab content divs but in theory it can be used to render anything

regarding the suggested changes, the scroll buttons are no longer tab focusable and also marked with `aria-hidden="true"`

> the SVG wouldn't need `aria-hidden="true"

just a side note, this is automatically applied to our icons by default when they are generated

> Note also that right-click currently activates the scroll buttons, which feels really odd.

scroll buttons should now only activate with left mouse button

> There's also a strange state that a keyboard user can get into where the tablist has the focus

I haven't been able to replicate that. Can you provide a list of steps to reliably enter this state?

> Are you allowed to conditionally take the scroll buttons out of the tab order on desktop and leave them in for mobile (if necessary)?

This would require us to differentiate between mobile and desktop clients, which is possible but it would be very hacky and inconsistent I believe

for the <kbd>Home</kbd> and <kbd>End</kbd> key enhancements, I think we can create a separate ticket for that to curb the scope creep that's occurring to this ticket

carmacleod:
This seems better, but the "strange state" is even more pronounced now, so I can't quite test fully yet:

>> There's also a strange state that a keyboard user can get into where the tablist has the focus

> I haven't been able to replicate that. Can you provide a list of steps to reliably enter this state?

Here's how to replicate (saw this in FF and Chrome on Windows - didn't try it anywhere else):
- shrink your browser window so that at least one scroll arrow appears
- either:
   - refresh the browser and then type the <kbd>tab</kbd> key
   - or click on a tab in the tablist (i.e. Tab label 1, or Tab label 2) and then type <kbd>shift+tab</kbd>

Somehow, the `ul` is in the tab order, which is really unexpected, because it doesn't have a `tabindex` or anything. I wondered how this was possible, and I think [this SO answer](https://stackoverflow.com/questions/1948241/how-could-a-ul-element-receive-the-focus-event#answer-59780325) makes sense. So presumably we can give the `ul` a `tabindex="-1"` and that should take it out of the tab order even in this weird scrollable overflow situation. That seems to work for me if I just add `tabindex="-1"` to the `ul` in DevTools. (Of course, I can't test after refreshing the page, but I assume that would work).

Sorry about the scope creep. Some of it was because I can't test the screen readers until the markup reaches the point where screen readers recognize the component.  :)

I've opened separate issues for:
- [home/end keys in tablist](https://github.com/carbon-design-system/carbon/issues/7070)
- [add tabpanel focus api](https://github.com/carbon-design-system/carbon/issues/7078)

Still need to open one for further discussion of renderContent api.
I still need to test mobile to see if the scroll arrow buttons need to be focusable, which may be an issue - not sure yet.

emyarod:
yeah that seemed to be a Windows only issue, which is resolved by adding `tabindex="-1"`. Thank you for opening the issues for the additional features! hopefully this PR is now good to go

carmacleod (approved):
+1

Looks good!
Thanks for your patience, @emyarod !

PR #7005, opened 2020-10-08
feat(accordion): add disabled state to accordion

### LINKED ISSUE #7000: [Accordion] missing disabled state
Accordion needs to have a disabled state.

**Spec:**

Artboard

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7000

Adds in `disabled` state for `Accordion` and `AccordionItem` 

#### Changelog

**New**

- `disabled` prop for `Accordion` and `AccordionItem`

**Changed**

- Added two knobs to Accordion Playground, one to disable the entire `Accordion`, and the other to disable just one item. 

#### Testing / Reviewing

Go to `Accordion Playground` and try disabling the entire Accordion as well as an AccordionItem


joshblack (commented):
For prop naming, what would be the best convention for went to decide between `disabled` and `isDisabled`?

One quick rule of thumb was that `isDisabled` is used for custom components and `disabled` is relegated to HTML elements but not sure how effective that is? 🤷

tw15egan:
@joshblack essentially the `disabled` state here is being passed down to a button, so it is using the correct syntax. Just needed a modifier class to handle disabling the borders, since that is placed on the parent element. I'd prefer to just keep it `disabled` across the board, so we don't make users think "Is this one `isDisabled` or `disabled`?".

I did wonder about other props like `light` vs `isLight`...

joshblack:
@tw15egan I'm down with that, makes a lot of sense to me 👍

joshblack (approved):

laurenmrice (approved):
Looks good to me 👍🏻

I think that the Gray 10 theme may need a value change for the disabled-01 token, but I can look into that and open a separate issue.

joshblack:
bump @aledavila when you get a sec today 👀

aledavila (approved):

PR #7010, opened 2020-10-09
feat(Link): added size support for Link component (#7001)

### LINKED ISSUE #7001: [link] add size variations
Add size variations to link and inline link
- 12px
- 16px

(or other way to control and set the type size)

rzhekova:
Closes #7001

Hello 👋 here are the changes for issue 7001. The link font size can now be set via a `size` prop which at the moment allows only `12px` and `16px` to be passed in as values. However, if you'd prefer I can change those to `sm` and `lg` so they're more in line with the TextInput component. 

Also, updated Storybook and `sass.md` with the new bit of CSS. 

Let me know what you think. Thanks!


packages/components/src/components/link/_link.scss
```diff
@@ -91,6 +91,14 @@
   .#{$prefix}--link--disabled.#{$prefix}--link--inline {
     text-decoration: underline;
   }
+
+  .#{$prefix}--link--12px {
+    font-size: 12px;
```
emyarod: can we convert this to `rem`s?

packages/components/src/components/link/_link.scss
```diff
@@ -91,6 +91,14 @@
...
+  }
+
+  .#{$prefix}--link--16px {
+    font-size: 16px;
```
emyarod: can we convert this to `rem`s?

emyarod (commented):
thanks for jumping in on this issue! would you mind using the `sm` and `lg` naming scheme like TextInput?

asudoh:
Weighing in; Wondering if we can introduce a t-shirt size or a like rather than a size indicating a pixel size. Especially given that Carbon for external facing websites has a theme to bump up the type/container sizes. Thanks!

rzhekova:
Hi guys, changed over to t-shirt sizes and rems. Thanks 🙂

emyarod:
thanks for making those changes @rzhekova ! it looks like there is a merge conflict with your latest test file which needs to be resolved

rzhekova:
Hi @emyarod yeah noticed there were conflicts, just pushed up more changes to resolve them

emyarod (approved):

rzhekova:
@emyarod Awesome, thanks!

tw15egan (approved):
Looks great, thanks for contributing! 🎉  👍  ✅

dakahn (approved):

aagonzales (approved):
Sorry for the delayed review. I was out of office. This looks great! Thanks for contributing!

packages/components/src/components/link/_link.scss
```diff
@@ -91,6 +91,14 @@
   .#{$prefix}--link--disabled.#{$prefix}--link--inline {
     text-decoration: underline;
   }
+
+  .#{$prefix}--link--sm {
+    font-size: rem(12px);
```
vpicone: I think we'll want adjust line height/letter-spacing as well. We could use the `helper-text-01` and `body-long-02` tokens to handle `sm` and `lg` respectively.

vpicone (changes requested):
@aagonzales Right now we're just adjusting font size, we probably would want to tweak line-height and letter spacing as well right?

We could use the `helper-text-01` and `body-long-02` tokens to handle `sm` and `lg` respectively.

packages/components/src/components/link/_link.scss
rzhekova: Hi @vpicone, the classes are now using the tokens you mentioned

vpicone: Awesome! We’ll just wait for design verification to make sure those are the right tokens.

aagonzales: `helper-text-01` isn't actually the correct token. Its more helper text that is associated with fields. We don't really have a token that matches this 12px usage right now. Does it absolutely have to have a token? If so I think we may need to make a new type token that supports it. 

@jeanservaas what do you think?

vpicone: @aagonzales it doesn't have to have a token, we can explicitly define each attribute. But token or not, I kind of assumed we'd want to match the line-height and letter-spacing of similarly sized text.

aagonzales: Yup we would want to match those other attributes. I talked to Jeannie about this and we want to add a new type token for the 12px scenario which has been coming up more. `helper-text-01` is actually going to be deprecated in the next release according to Jeannie.

rzhekova: Hi both, I have a couple of questions regarding what you discussed above. 

If a new token is needed, is it something that has to be done as part of this PR? If no, I can for the time being just add the CSS needed to `--link--sm` and once the token is created it can replace the CSS in another issue perhaps. If yes, that's fine I can add the changes in another commit, assuming to create a new one I just need to use eg `helper-text-01` or some other token as an example. Either way, it would be great if you let me know what the line-height and the letter-spacing need to be set to.

And finally, is `--link--lg` okay using `body-long-02` or would that need changing too?

Thanks!

aagonzales: Yes the new token would be a task outside of this issue. So I guess its ok to proceed with using `help-text-01` for now and we can replace it later (unless dev disagrees with that @vpicone @tw15egan @emyarod )

And I was distracted by that token convo, but the `--lg` link should be using `body-short-02` to match the line height sequencing of reg link which uses `body-short-01`

rzhekova: No worries @aagonzales! I'll wait until a decision's been made around `--link--sm`, and I'll change the other one at that point as well

vpicone: @rzhekkova I don’t think a link variant token makes sense, since it would only be used in this one place anyway. IMO using the helper-text token now will make it easier to find and clean up when we actually deprecate the token.

aagonzales: It wouldn't be link specific. It would be a new general 12px type style. Jeannie has come across other places where she's need a 12px style that isn't a label or caption.

rzhekova: Ok cool, that makes sense. So, in that case I'll just change `--link--lg` to use `body-short-02` as per Anna's message above

vpicone: @aagonzales for sure, that’s was my concern for a link—sm token suggested here. 

Using the existing helper token would make it easier to find and replace instances in the future when the general token is published. Easier than trying to find/replace chunks of css properties.

rzhekova: Hi guys, I've updated the large link to use `body-short-02`. Let me know if any other changes are needed :)

vpicone (approved):
LGTM!

tw15egan:
@aagonzales can you just take one final look before we merge? 🙏

aagonzales (approved):
Looks good!

tw15egan:
@all-contributors add @rzhekova for code

PR #7035, opened 2020-10-13
fix(themes): correct color for highlight token in g10 theme

janhassel:
According to the [color usage](https://www.carbondesignsystem.com/guidelines/color/usage) page, the token `highlight` should use `blue-10` in the Gray 10 theme as opposed to `blue-20` in the White theme. This is also reflected in the Sketch kit (layer styles).

image

However, in code, the gray 10 theme currently uses `blue-20` instead of `blue-10`. This PR fixes that.

#### Changelog

**Changed**

- Corrected color used for token `highlight` in g10 theme

#### Testing / Reviewing

- Verify that the g10 theme emits `blue-10 (#edf5ff)` for `highlight` instead of `blue-20 (#d0e2ff)`

### DIFF
--- a/packages/themes/src/g10.js
+++ b/packages/themes/src/g10.js
@@ -8,7 +8,7 @@ import { adjustLightness } from './tools';
 
 import {
   // Blue
-  blue20,
+  blue10,
   blue40,
   blue60,
   blue70,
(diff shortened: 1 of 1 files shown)


tw15egan (approved):
LGTM, thanks for fixing that 👍 ✅

emyarod (approved):
looks good to me

Issue #5455, opened 2020-02-26
Proposal: Support for 48px versions of a wider range of components


lukefirth:
### Summary

'Tall' (sometimes also called 'extra large') styling support for some component types is available for _some_ components today, mostly inputs via the 'field size' feature (referenced in a sibling issue: #5448 ). This issue requests extending similar support for other interactive components.

### Justification

A good number of our matured products in Watson IoT that have traditionally not been Carbon, are moving towards full adoption. However, a portion of those larger offerings also have small-screen and touch requirements. 

Many of these requirements look like they can be met near immediately through use of the field size features 'extra-large' option referenced in issue #5448 , where others without this type of feature would also benefit from a larger size option for bigger targets.

### Desired UX and success metrics

The set identified where 48px versions would be immediately beneficial:

- [ ] Accordion
- [x] Buttons (default already available at 48px)
- [x] Content switcher
- [x] Date picker
- [x] Dropdowns (already seems to support xl which is 48px input)
- [ ] Overflow menu items
- [x] Number input
- [x] Searchbar (already seems to support xl which is 48px input)
- [ ] Tabs (default - container is already 48px)
- [x] Text input (already seems to support xl which is 48px input)

![Artboard Copy 2](https://user-images.githubusercontent.com/30671949/75388436-9bcfe880-58aa-11ea-9d9b-c9ad154e9619.png)

Note: We'd also make two minor optional recommendations for these 'larger' versions that:

- The spacing between the label and any input is adjusted from the current 8px to 16px, see below:
![Artboard](https://user-images.githubusercontent.com/30671949/75386334-1eef3f80-58a7-11ea-880e-a9a009b191d3.png)

- The default icon sizes used be set to 20px, rather than 16px
![Artboard Copy](https://user-images.githubusercontent.com/30671949/75386542-7db4b900-58a7-11ea-843f-bf48f4d88e40.png)
_(This also gets us around the thicker rendering of the 16px '+' icon at the default 48px button size)_

Note: As with #5448 , a naming convention would be beneficial for describing these various 'versions' of the inputs. Eg: Data table uses 'tall' for scales above the default, and I would recommend that this convention is used here too (rather than the 'extra-large' used currently in the field size prop). 

I can provide more extensive examples, sketch files with these already mocked up and in use in a library etc if required, as well as an internal repo issue link with more extensive documentation.

### "Must have" functionality

- The ability to set the above list of components to a size based on 48px height.

### Specific timeline issues / requests

[...]

PR #7038, opened 2020-10-13
feat(accordion): add size variants to Accordion

tw15egan:
Related to https://github.com/carbon-design-system/carbon/issues/5455

Adds in `sm` (32px) and `xl` (48px) variants to `Accordion` 

#### Changelog

**New**

- `size` prop added to `Accordion`. Adds in two variants `sm` and `xl`. 

**Changed**

- Updated snapshots

#### Testing / Reviewing

Ensure all size variants look correct

### DIFF
--- a/packages/components/src/components/accordion/_accordion.scss
+++ b/packages/components/src/components/accordion/_accordion.scss
@@ -72,6 +72,16 @@
     }
   }
 
+  // Size styles
+  .#{$prefix}--accordion--xl .#{$prefix}--accordion__heading {
+    min-height: rem(48px);
+  }
+
+  .#{$prefix}--accordion--sm .#{$prefix}--accordion__heading {
+    min-height: rem(32px);
+    padding: rem(5px) 0;
+  }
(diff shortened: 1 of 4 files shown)


emyarod (approved):

jeanservaas (commented):
This looks great and I want to approve. I just have one question, is it possible to call this a "Large" variant vs. an "Extra Large" variant? Since we're starting to add large variants everywhere, we should be consistent. The 64px button variant is only called "Large" and the 48px search I believe might even be default.

Thoughts on this?

aledavila (approved):

tw15egan:
@jeanservaas we'll be able to do that on the next major release, but right now we're just aligning it to all the other form input sizes (`sm` and `xl`). In the next major release, we'll switch to `small` and `large` as the two variants.

johnbister (approved):
Weird Al would be proud.

weirdal

jeanservaas (approved):

PR #7045, opened 2020-10-14
fix(Tabs): keep selected tab in view

### LINKED ISSUE #6917: Tabs component: need ability to scroll to the selected tab
When the Tabs component has a long list of tabs inside and the page is first rendered with a selected tab, the selected tab may not be shown.  The user has to press the < or > arrows to look for the selected tab.

I am using carbon-components 10.20.0 and carbon-components-react 7.20.0.

Example is like this:
When I load the page, it shows the list of tabs:
image

But have to scroll to see the selected tab:
image

I think it will cause usability issue from end users.

emyarod:
Closes #6917

This PR makes sure the selected tab in the component state is visible (component scrolls to the selected tab) when the component updates

#### Changelog

**Changed**

- scroll selected tab into view on component update

#### Testing / Reviewing

Change the selected tab and confirm that the tab is in view on initial render in addition to subsequent renders

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.js
+++ b/packages/react/src/components/Tabs/Tabs.js
@@ -145,6 +145,29 @@ export default class Tabs extends React.Component {
 
     this._handleWindowResize();
     window.addEventListener('resize', this._debouncedHandleWindowResize);
+
+    // scroll selected tab into view on mount
+    const {
+      clientWidth: tablistClientWidth,
+      scrollLeft: tablistScrollLeft,
+      scrollWidth: tablistScrollWidth,
+    } = this.tablist.current;
+    const tab = this.getTabAt(this.state.selected);
+    const horizontalOverflow = tablistScrollWidth > tablistClientWidth;
+
+    if (horizontalOverflow) {
+      const leftOverflowNavButtonHidden =
+        tab?.tabAnchor?.getBoundingClientRect().right <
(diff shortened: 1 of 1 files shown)


joshblack:
@emyarod random question, is there a chance that this `cDU` call could be invoked if Tabs is not in the viewport? If so, would this change cause the page to scroll the selected tab into view (in the y direction)?

emyarod:
no it shouldn't be scrolled in the y direction (related https://github.com/carbon-design-system/carbon/pull/6974)

edit: I am assuming you are not referring to the default behavior where changing focus also scrolls the focused element into view

joshblack (approved):

tw15egan:
Seeing an issue where you are not able to click to scroll to other tabs since it seems like it wants to keep the focus on the first tab 

focus-click

emyarod:
@tw15egan I'm unable to replicate that issue. Can you provide more details about how you're getting that to occur?

dakahn:
Seeing the same bug as @tw15egan on Firefox latest Windows 10

emyarod:
can you take another look with the latest PR changes?

joshblack:
bump @dakahn based on comment above

dakahn (approved):

dakahn:
Scrolling works swell! 👍

PR #7056, opened 2020-10-15
feat(DatePicker): set up localization for datepicker

crhuff:
## Changelog

**New**

- Sets default dateFormat and placeholder based off locale
- Modified `zh-tw` to `zh_tw` and `zh-cn` to `zh` when passed to flatpickr
- Allow null override of automatic placeholder
- Pass locale from DatePicker to DatePickerInput to prevent double prop entry

#### Testing / Reviewing

Pass a locale ('en', 'no', 'zh', 'zh-tw') to DatePicker and no format and check if the format is updated.
Pass a locale to DatePicker and a null placeholder and check if the placeholder is empty.


packages/react/src/components/DatePicker/DatePicker.js
```diff
@@ -29,6 +29,38 @@ l10n.en.weekdays.shorthand.forEach((day, index) => {
...
+    case 'pt':
+      return 'd/m/Y';
+    case 'zh':
+    case 'zh_tw':
+    case 'ja':
+      return 'Y/m/d';
+    default:
+      return 'm/d/Y';
+  }
+};
```
asudoh: Weighing in; Carbon core doesn't provide locale data due to the maintenance concern (and thus let application deal with it instead as the current `master` does). e.g. Running below code produces some different results from here, e.g. for `en-CA`:

```javascript
console.log('af:', new Intl.DateTimeFormat('af').format(new Date()));
console.log('af-NA:', new Intl.DateTimeFormat('af-NA').format(new Date()));
console.log('af-ZA:', new Intl.DateTimeFormat('af-ZA').format(new Date()));
console.log('agq:', new Intl.DateTimeFormat('agq').format(new Date()));
console.log('agq-CM:', new Intl.DateTimeFormat('agq-CM').format(new Date()));
...
```

packages/react/src/components/DatePicker/DatePicker.js
```diff
@@ -210,6 +242,8 @@ export default class DatePicker extends Component {
       'uk', // Ukrainian
       'vn', // Vietnamese
       'zh', // Mandarin
+      'zh-cn', // Mandarin alternate
```
asudoh, line 262: The list here should be in sync with https://github.com/flatpickr/flatpickr/tree/v4.6.1/src/l10n, but `zh-cn` doesn't seem to be there.

crhuff: Hmm, yeah this is a product of i18n spitting out zh-cn. I end up re-casting it to `zh` and `zh-tw` to `zh_tw`. We can do that locally, however I felt like enough people might just be passing through that it would be worth it.

packages/react/src/components/DatePicker/DatePicker.js
crhuff: I can certainly switch it over to that instead of using the hardcoded values.

asudoh: Cool, also note that `m/d/Y` thing is just a default value. Application should take care of putting the right prop value for the locale user is using.

emyarod (commented):
thanks for opening this PR! from what I can tell this seems like an enhancement which would add some unwanted maintenance concerns as @asudoh mentions. given that the default prop values can be overridden by the application already, I believe it makes sense for us to continue with the current method of providing locale data rather than setting the default value dynamically

crhuff:
> thanks for opening this PR! from what I can tell this seems like an enhancement which would add some unwanted maintenance concerns as @asudoh mentions. given that the default prop values can be overridden by the application already, I believe it makes sense for us to continue with the current method of providing locale data rather than setting the default value dynamically

Regarding removing the `m/d/Y` default prop, that reason I did that was if a locale is passed but no dateFormat, there's no way to tell if the user specifically wanted `m/d/Y` or just did not pass a value. In this situation, I still set dateFormat to `m/d/Y` if no locale and no dateFormat was defined, but only after finding out whether dateFormat was explicitly set. I'm open to suggestions on updating that though

crhuff:
I've added some updates now to make the date format more regional/less hardcoded @asudoh. Let me know if it's in line with what you're thinking/maintenance free enough for carbon

packages/react/src/components/DatePickerInput/DatePickerInput.js
```diff
@@ -13,6 +13,42 @@ import { Calendar16 } from '@carbon/icons-react';
...
+  if (placeholder !== undefined) return placeholder;
+  if (!locale) return null;
+  let formatObj;
+  try {
+    formatObj = new Intl.DateTimeFormat(upperSubLocale(locale)).formatToParts(
+      new Date()
+    );
+  } catch (err) {
+    console.log(err);
+    formatObj = new Intl.DateTimeFormat('en').formatToParts(new Date());
```
asudoh: This API is not supported by IE11. As @emyarod said, my suggestion was that dealing with date format in application code instead of doing it in our date picker code.

crhuff: I've got it now to fall back to whats currently in carbon, so hopefully this can only be an improvement when possible. If you see any issue with falling back to default props then please let me know

emyarod: the try-catch block would cause IE11 to fall back to the original/current component behavior, and for the browsers that do support the `formatToParts` method, the formatting would be displayed differently. not sure if we want to proceed if this would create inconsistent behavior across browsers

joshblack:
bump @emyarod @aledavila when you get a sec to look 👀

emyarod:
?? we are having an open discussion here https://github.com/carbon-design-system/carbon/pull/7056#pullrequestreview-512239774

packages/react/src/components/DatePickerInput/DatePickerInput.js
crhuff: While I lean towards ignoring IE11 since it's EOL ~ Aug '21, the other option would be to create a date with Jan 31, 1970 or something, then parse the location of each since `Intl.DateTimeFormat(locale).format(new Date())` is supported on all browsers

Example: 
image

emyarod: since this would be setting the default formatting based on locale, I think unit tests against all of the [available locales](https://github.com/unicode-cldr/cldr-core/blob/master/availableLocales.json) should be expected. but aside from that, Carbon has a commitment to supporting IE11 so I don't think we should ignore it until support for IE11 is officially dropped from the library

emyarod:
based on our conversation in Slack, we've decided to keep this enhancement at the PAL level. Until further research has been conducted on `Intl.DateTimeFormat` against the available locales https://github.com/unicode-cldr/cldr-core/blob/master/availableLocales.json in order to determine the default values we can assign for each locale, a table of defaults to locales can be maintained at the application level.

for Carbon to adopt this, we likely would be maintaining our version of the locale list/auto generating the format from locale, which would mean another `locale` property to maintain. This enhancement would probably become more viable at the Carbon level once we're able to drop IE support, which would allow us to more easily test and use `Intl.DateTimeFormat`, but until then I think we can close this for the time being

PR #7060, opened 2020-10-16
chore(project): update yarn to berry 

joshblack:
We were sitting on Yarn v1.22 which was the last stable version of the 1.x branch for a while before Yarn v2 (berry) stabilized.

Now, after adding in support for node_modules in addition to PnP we can hopefully move over to v2 with minimal changes.

This PR adds in support for Yarn berry and their new zero-installs mode, along with updates to our project where needed to make it work.

#### Changelog

**New**

- Yarn cache folder for zero-installs

**Changed**

- Update yarn version to berry
- Update `carbon-components-react` to use dynamic bin location through `npm-which` instead of relying on static path
- Yarn automatically updates bundler now since the name of the bin matches the package name
- Update CI workflows to new install syntax

**Removed**

- Remove yarn v1.22.5 release file and offline-mirror


joshblack:
Note, unfortunately the flags we pass into Yarn (`--offline --frozen-lockfile`) cause it to fail with the new version. I'll try and see if a netlify config file will help with it in the interim

emyarod:
would the netlify config changes need to be merged separately first?

joshblack:
Gonna close this, for now, I think with the latest changes it would work but I think we should wait for things to be more stable before bringing it in

PR #7061, opened 2020-10-16
ci(workflows): add manual dispatch for react release and fix cf version

joshblack:
Add manual dispatch for React storybook release, and fix issue with CF version (we used to pin and now we no longer need to)

#### Changelog

**New**

**Changed**

- Add manual dispatch for React storybook release, and fix issue with CF version

**Removed**


emyarod (approved):

andreancardona (approved):
@joshblack looks good! 🍿

PR #7064, opened 2020-10-16
docs: fix Node.js version requirement

dbrugger:
Closes

```
> yarn install --offline

yarn install v1.22.5
[1/5] 🔍  Validating package.json...
error carbon@0.0.0: The engine "node" is incompatible with this module. Expected version "12.x". Got "10.20.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

#### Changelog

**Changed**

- Match Node.js version requirement in the Contributor guide with version in https://github.com/carbon-design-system/carbon/blob/master/package.json#L8

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -20,7 +20,7 @@ contributing environment. Please take a look at our
 Before contributing to Carbon, you should make sure you have the following tools
 installed:
 
-- [Node.js](https://nodejs.org/en/download/) v10 or above here or follow their
+- [Node.js](https://nodejs.org/en/download/) v12 or above here or follow their
   installation through a package manager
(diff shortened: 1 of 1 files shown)


joshblack (approved):

janhassel:
Quick question since I use v14 as default on my mac and regularly come across this: if it says "Node v10 and above" (or "Node v12 and above" with this change), shouldn't https://github.com/carbon-design-system/carbon/blob/master/package.json#L8 also reflect that and use `>=12.x` instead of `12.x` as the semver constraint?

joshblack:
@janhassel I think you're right, I went ahead and made this change over in: https://github.com/carbon-design-system/carbon/pull/7076

emyarod (approved):

PR #7065, opened 2020-10-17
fix(ClickListener): support Shadow DOM

dbrugger:
Closes

When using the OverflowMenu in a Shadow DOM, the it closes immediately.
The ClickListener wrongly determines that the clicked element (which is the OverflowMenu itself) is outside.
This is because in this scenario evt.target is the root element, which indeed lies outside.

#### Changelog

**Changed**

- Check if the event is [composed](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed) and in that case use the first one in the [composedPath](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath)

#### Testing / Reviewing

* verify that the behavior in normal circumstances (non Shadow DOM, e.g. in the Storybook) still works for components that use ClickListener:
  - OverflowMenu
  - ToolbarSearch
  - Tooltip

### DIFF
--- a/packages/react/src/components/OverflowMenu/OverflowMenu.js
+++ b/packages/react/src/components/OverflowMenu/OverflowMenu.js
@@ -409,7 +409,7 @@ class OverflowMenu extends Component {
         menuBody.ownerDocument,
         focusinEventName,
         (event) => {
-          const { target } = event;
+          const target = ClickListener.getEventTarget(event);
           const { current: triggerEl } = this._triggerRef;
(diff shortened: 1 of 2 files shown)


joshblack:
bump @tw15egan @andreancardona

tw15egan (approved):
Everything is still working as expected 👍 ✅

andreancardona (approved):
@dbrugger looks good to me! - feel free to merge when you're ready!

PR #7068, opened 2020-10-17
feat(Dropdown): remove selection for optional dropdowns

### LINKED ISSUE #6581: [Dropdown] Including a way to remove selection for optional dropdowns
### Summary

There is a need for users to have a way to clear the selection for **optional** dropdowns. 

**For example:** 
If a user made a selection on an optional field, but wanted to remove the selection and keep it blank.

We are looking to include a "Select" option in the dropdown to allow the user to revert back to a _null_ option, but would love the Carbon team's thoughts on this as well. 

Screen Shot 2020-07-29 at 2 40 41 PM

It would be helpful if this guidance was also included in the [dropdown component](https://www.carbondesignsystem.com/components/dropdown/usage) as well.

AlexanderLyon:
Closes #6581

#### Changelog

**New**

- Adds `isOptional` prop to `<Dropdown/>` component. This defaults to `false` (defaulting to the existing functionality)
- Adds the ability to "remove" a previously chosen selection from dropdown by dynamically adding either the label text or "Select" as a choice to the beginning of `items` array.
  - **NOTE**: this currently mutates the value of `items`, which may not be desirable

#### Testing / Reviewing

* If the `isOptional` (boolean) prop is `true` and `label` prop is set, the first choice in the dropdown should be the `label`'s text
* If the `isOptional` (boolean) prop is `true` and `label` prop is _NOT_ set, the first choice in the dropdown should be "Select"

The above should work correctly regardless of if `items` an array of **strings** or an array of **objects**.

image


emyarod:
cc @laurenmrice for visual review

emyarod (commented):
would you mind updating snapshots to pass CI? (`yarn test -u`)

AlexanderLyon:
@emyarod snapshots have been updated!

emyarod (commented):
thanks for updating your snapshots. did you have any updates for the storybook example with your new prop?

AlexanderLyon:
@emyarod hmm, if this prop defaults to `false` should it be shown in the storybook example? Or do you think the documentation is enough?

laurenmrice:
I think there should be a prop for the optional entry in the storybook example. @emyarod @AlexanderLyon

emyarod:
yes, it would be best if we could document the prop in Storybook examples as well. this would also allow for easier testing!

AlexanderLyon:
@laurenmrice @emyarod I've added a storybook example for an optional Dropdown (7e91929), as well as updated the component's documentation to explain the new prop a little bit better (906d767)

laurenmrice (commented):
- I see the `optional` prop now. Thank you!

- By default when the dropdown does not have the optional prop checked, the dropdown options should not include the prompt as an option in the menu. This should only happen when the optional prop is being used.

Screen Shot 2020-10-22 at 10 32 53 AM

aledavila:
Hey! Thank you for submitting this. At this time we will be closing the PR as requested changes haven’t been made. Feel free to open it again if new updates come up :)

AlexanderLyon:
@aledavila - I apologize for the delay in response on this pull request, unfortunately things has been very busy for me lately. 

I've opened up a new PR #7332 to continue this discussion, fixing the issue pointed out by @laurenmrice!

PR #7079, opened 2020-10-19
feat(react): add deferComponentRender HOC

### LINKED ISSUE #5901: [AccordionItem] extend to render contents when expanded/opened
### Summary

I'd like to offer up a contribution that we made recently in https://github.com/IBM/carbon-addons-iot-react/pull/1105 , if the team is interested. 

We've extended `AccordionItem` to include optional stateful logic to only render the contents on expand of the section.

### Justification

Some consuming applications with a ton of content (think entire dashboards) see a perf hit when contents is always rendered irrespective of if the `AccordionItem` is closed.

via a prop or composition, the accordion item can be configured to render content when expanded. Closing of the section should not unmount the contents, but preserve it.

n/a

### Available extra resources

I can submit a PR if y'all are good with this idea 👍

joshblack:
Closes #5901

This is a follow-up to #5901 that brings in a `deferComponentRender` HOC that can be used alongside components which teams may need to defer rendering for (Tabs, Accordion, Content Switcher) in pages with a large trees that are mounted/un-mounted (reference post for this issue is seen over in the [Twitter Lite](https://medium.com/@paularmstrong/twitter-lite-and-high-performance-react-progressive-web-apps-at-scale-d28a00e780a3) blog post)

#### Changelog

**New**

- Add `deferComponentRender` HOC that is derived from the implementation in the Twitter Lite blog post

**Changed**

- Updated Accordion docs to include performance section

**Removed**

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -7011,6 +7011,7 @@ Map {
     },
     "render": [Function],
   },
+  "deferComponentRender" => Object {},
   "unstable_PageSelector" => Object {
     "defaultProps": Object {
       "className": null,
--- a/packages/react/src/__tests__/index-test.js
+++ b/packages/react/src/__tests__/index-test.js
@@ -192,6 +192,7 @@ describe('Carbon Components React', () => {
         "TooltipDefinition",
         "TooltipIcon",
         "UnorderedList",
+        "deferComponentRender",
         "unstable_PageSelector",
         "unstable_Pagination",
         "unstable_TreeNode",
--- a/packages/react/src/components/Accordion/Accordion-story.js
+++ b/packages/react/src/components/Accordion/Accordion-story.js
@@ -22,6 +22,7 @@ import {
   AccordionSkeleton,
 } from '../Accordion';
 import Button from '../Button';
+import { deferComponentRender } from '../../tools/deferComponentRender';
 import mdx from './Accordion.mdx';
 
 export default {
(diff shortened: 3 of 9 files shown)


joshblack:
FYI @tay1orjones curious if this handles what you were thinking over in #5901 👀

tay1orjones:
@joshblack Yeah! Fantastic point of abstraction so it can be used with any component, not just the accordion items. Will definitely deprecate our solution once this is published and we update. Thanks! 👍

tay1orjones:
@joshblack I looked through our implementation again - one thing that is missing here that I mentioned in the original issue:

> Closing of the section should not unmount the contents, but preserve it.

This PR wouldn't meet that need directly, we'll still have to wrap AccordionItem to include some state so the contents isn't unmounted when it closes

emyarod (commented):
> Closing of the section should not unmount the contents, but preserve it.

should we extend the component to meet this? seems like a great enhancement if we can keep it at the library level

joshblack:
@tay1orjones ah, I get what you mean now. It seems like the defer for Accordion Item ends up doing two things:

- Defer initial render of `children` until it's been opened
- Persist `children` after it's been open once

joshblack:
@tay1orjones was trying to play around with different ways to build that but none really seemed to stick 🤔  The biggest challenge was honestly just trying to come up with something that was `AccordionItem` agnostic lol.

It sucks that the panels are _always_ rendered for accordion, otherwise the trick above would work out just fine 😞 

I'm going to go ahead and close this one for now, can definitely revisit it but not sure if there will be any progress on this (thanks so much for participating @tay1orjones!)

PR #7085, opened 2020-10-19
feat(breadcrumb): add overflow menu support

### LINKED ISSUE #7016: [breadcrumb] hover state bug
In Firefox, the second and third breadcrumbs flash when they receive a hover because they revert back to the enabled state color. This is happening in react and in the live demo. Chrome and Safari browsers seem to be fine.

In Firefox, the color I visually see for hover is the same as the enabled state. The token that should be used for hover is `hover-primary-text`. This is appearing correctly in Safari and Chrome.

Example:
breadcrumb

### LINKED ISSUE #7019: [breadcrumb] needs to have an overflow option
Breadcrumb needs to be able to have an overflow option.

Spec:
Artboard Copy 2

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7019
Closes https://github.com/carbon-design-system/carbon/issues/7016

Adds in support for `OverflowMenu` nested in `BreadcrumbItem` 

#### Changelog

**New**

- Story for the new variant
- Style overrides to achieve correct visual styles for `OverflowMenu` inside a `BreadcrumbItem`

**Changed**

- Modified `BreadcrumbItem.js` to pass down a `className` to the `FloatingMenu` placed by the `OverflowMenu`. This is needed to create the shark fin styles, as the `FloatingMenu` is placed outside of the entire `Breadcrumb` component and there was no way to scope the styles.
- Also passes in custom `menuOffset` props to align the menu to the middle of the overflow menu

#### Testing / Reviewing

Verify the `OverflowMenu` variant renders properly across all browsers


laurenmrice (changes requested):
Can we make these tweaks to the overflow icon and rule:
- Align the overflow dots to the baseline of the breadcrumb text.
- Have 2px space between the overflow dots and the rule.
- Make the rule 12px long so it aligns flush with the overflow dots.

breadcrumb1

---------------------------------------------------------------------------------------------

There is a break line showing up between the sharkfin triangle and the container of the overflow menu. They should be connected to eachother and not have the break.
breadcrumb2

---------------------------------------------------------------------------------------------
The sharkfin triangle point needs to align to the middle overflow menu dot. Right now it is a little to the left.
breadcrumb3

laurenmrice:
There is no more flashing of color for the breadcrumbs in Firefox for me. 👍🏻

tw15egan:
@laurenmrice updated 👍

laurenmrice (changes requested):
Safari specific bugs:
- The underline is really close to the the overflow menu dots on hover.
- There is this weird focus happening when you click on an overflow option. I am not sure if that is possible to get rid of?

Oct-21-2020 12-47-13


Everything else is looking great!

packages/components/src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -70,6 +74,73 @@
...
+  .#{$prefix}--breadcrumb-item
+    .#{$prefix}--overflow-menu.#{$prefix}--overflow-menu--open {
+    background: transparent;
+    box-shadow: none;
+  }
+
+  .#{$prefix}--breadcrumb-item .#{$prefix}--overflow-menu__icon {
+    position: relative;
+    margin-top: rem(8px);
+    transform: rotate(90deg);
```
emyarod:
```suggestion
    transform: rotate(90deg) translateX(rem(4px));
```

I think this should lead to consistent alignment across all browsers, replacing `margin-top` with a CSS transform

emyarod: it's kind of trippy using `translateX` to position it vertically, but I think that's because of the `rotate(90deg)` since we're just importing the overflow menu component directly, which uses the vertical overflow icon. we could also use overflow menu's `renderIcon` with the horizontal overflow icon and then do this maybe

```suggestion
    transform: translateY(rem(4px));
```

laurenmrice (approved):
Looks wonderful in Safari now! Thank you ! 🎉

tay1orjones (commented):
@tw15egan could [the docs pane](https://deploy-preview-7085--carbon-components-react.netlify.app/?path=/docs/breadcrumb--breadcrumb-with-overflow-menu) be updated to describe this feature/story? The 'show code' source view alone would be quite helpful imo

tw15egan:
@tay1orjones that's a great idea, let me add that to this PR 👍

lee-chase:
Hi, @tw15egan I have this responsive in-progress prototype https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/252 and have a couple of questions.

1. What positions can the overflow go in?
2. Can there be only overflow?

I have allowed that as I don't know what width my container will be.

Currently only visible here (see controls panel to adjust width)
- https://deploy-preview-252--ibm-cloud-cognitive.netlify.app/?path=/story/experimental-breadcrumbwithoverflow--breadcrumb-item-array

May arrive here soon 
- https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-breadcrumbwithoverflow--breadcrumb-item-array

lee-chase:
Hi, @tw15egan I have this responsive in-progress prototype https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/252 and have a couple of questions.

1. What positions can the overflow go in?
2. Can there be only overflow?

I have allowed that as I don't know what width my container will be.

Currently only visible here (see controls panel to adjust width)
- https://deploy-preview-252--ibm-cloud-cognitive.netlify.app/?path=/story/experimental-breadcrumbwithoverflow--breadcrumb-item-array

May arrive here soon 
- https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-breadcrumbwithoverflow--breadcrumb-item-array

BreadcrumbWithOverflow

tw15egan:
@lee-chase Nice! As far as your questions go:

1) I believe since it will only contain a list of BreadcrumbItem links, it would be placed between the first and last `BreadcrumbItem`. I think it could be confusing if the overflow menu appeared first or last, since those breadcrumbs are usually there to help with a user's context of where they are in an app. I'll defer to @laurenmrice though.

2) Oooo, I see that yours dynamically adds items to the overflow menu. That's a nice feature! This PR was originally just a quick fix to allow proper styling if an `OverflowMenu` was a child of `BreadcrumbItem`, but a more fully-featured implementation would be more of a benefit to our users. Wondering if we should port those changes here?

jnm2377 (approved):

tay1orjones (approved):

laurenmrice:
Can we merge this in?

PR #7097, opened 2020-10-21
feat(carbon): update Downshift to v6

dakahn:
Updates Downshift to latest. prerequisite pull request I figured we should get in before we refactor MultiSelect to use Downshift's fancy new `useMultipleSelection` hook.

Did a quick run through of the effected components and everything seemed a-okay, but check my work for sure.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "@carbon/icons-react": "^10.19.0",
     "classnames": "2.2.6",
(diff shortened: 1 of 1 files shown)


emyarod (commented):
I believe you will need to update more than `package.json` to get the updated version to install in the CI tests?

joshblack:
@dakahn are there any breaking changes that an end-user would observe through this update? (maybe through downshiftProps??)

PR #7108, opened 2020-10-21
fix(multi-select): update filterable multiselect default styles

### LINKED ISSUE #7041: Issue: Filterable Multiselect inline property not displaying correctly

- [ ] `carbon-components`

## Detailed description

The filterable Multiselect component does not look "inline" when you apply the inline property to it.

Screen Shot 2020-10-13 at 4 03 11 PM

> see changed?

It should look like an inline multiselect like:
image

Chrome

v7.20.0 - https://react.carbondesignsystem.com/?path=/story/multiselect--filterable

> should be aware of?
Cloud Pak for Data

> Go to https://react.carbondesignsystem.com/?path=/story/multiselect--filterable and select Inline knob

emyarod:
Closes #7041

This PR updates the filterable multiselect styles to match the updated component spec

#### Testing / Reviewing

Confirm the filterable multiselect variants are visually correct

### DIFF
--- a/packages/components/src/components/multi-select/_multi-select.scss
+++ b/packages/components/src/components/multi-select/_multi-select.scss
@@ -64,6 +64,13 @@
     }
   }
 
+  .#{$prefix}--multi-select--filterable.#{$prefix}--multi-select--inline,
+  .#{$prefix}--multi-select--filterable.#{$prefix}--multi-select--inline
+    .#{$prefix}--text-input {
+    background-color: transparent;
+    border-bottom: 0;
+  }
+
   .#{$prefix}--multi-select--selected .#{$prefix}--text-input {
     // this value will need to change based on the number of digits in
     // the number of items selected


joshblack (approved):

laurenmrice:
This change needs to only happen for the `inline` variant of the filterable multiselect. The default filterable multiselect needs to keep its field-01 background (and field-02 background for the light prop).

We just need to change the `inline` version to not have a field background color. When the menu is open, only the options should still get the field-01 (and field-02 for the light prop)

andreancardona (approved):
@emyarod looks good to me! 🎉

tw15egan (changes requested):
Should only apply to inline variant

tw15egan (approved):

laurenmrice (approved):
looks great! thanks ! 🙌🏻

PR #7109, opened 2020-10-21
fix(Tabs): remove nested tabs selector

emyarod:
related #6928

This PR removes an extra level of nesting in our scrollable tabs component, reverting the wrapper div back to a React Fragment just as it was prior to #6928

#### Changelog

**Changed**

- update test selectors

**Removed**

- remove nested selector in Tabs
- unnest styles

#### Testing / Reviewing

Ensure the tabs component styles are still namespaced and there are no regressions in scrollable tab styles


joshblack (approved):

tw15egan (approved):
looks good to me 👍 ✅

PR #7131, opened 2020-10-26
chore: add react and react-dom 17 version to peerDependencies

dbrugger:
When using carbon-components-react in a react 17 app, warnings appear about incorrect dependency version.

#### Changelog

**Changed**

- add react and react-dom version 17 in addition to 16 to peerDependencies

#### Testing / Reviewing

I've tried the storybook with react and react-dom version 17 and it seems to work (test was done by changing the devDependencies in packages/react so that the storybook build runs with the new version - note this is not part of this PR) 
Also I do realize that this might not be something for a quick PR but wanted to bring it up as folks will start using react 17 regardless, and it seems CRA is using it already.


joshblack (approved):

emyarod (approved):
looks good to me

tw15egan:
@all-contributors add @dbrugger for maintenance

PR #7133, opened 2020-10-26
fix(radio-button): margin bug fix

### LINKED ISSUE #7113: Radio button's focus state is truncated when in a container with a hidden overflow

- [ ] `carbon-components`

## Detailed description

When a radio button is used inside a container with a hidden or auto overflow set, the box-shadow used to show the focused state of the radio button is truncated

Radio button

> see changed?
Provide some padding around the control to account for the focus state's box-shadow

Chrome

latest 

> should be aware of?

1. Go to https://codesandbox.io/s/damp-grass-kjx1w?file=/index.js
2. Select the radio button to show a focused state
3. Observe that the box-shadow around the radio button is cut off below and on the left

>https://codesandbox.io/s/damp-grass-kjx1w?file=/index.js

## Additional information

- Screenshots or code
image

- Notes
A similar issue was recently raised and addressed with Checkbox here: https://github.com/carbon-design-system/carbon/issues/6480

andreancardona:
Closes #7113 

added a margin around radio button to fix truncation bug

#### Changelog

**New**

- `0.25 rem to margin-top & margin-bottom` around radio-button

#### Testing / Reviewing
 

- [ ] Go to: https://deploy-preview-7133--carbon-components-react.netlify.app/?path=/story/radiobutton--default

- [ ] add border & overflow around radio-button `bx--radio-button__label` as seen below `border: 10px solid red;` `overflow: hidden;`

- [ ] make sure radio-button's focus state is not cut off / truncated

Screen Shot 2020-10-26 at 11 46 41 AM


joshblack:
@andreancardona if I understand this issue, it's that the box-shadow bleeds over the box model of the radio button? Just wanted to confirm 👀

andreancardona:
> @andreancardona if I understand this issue, it's that the box-shadow bleeds over the box model of the radio button? Just wanted to confirm 👀

Yep! Let me know if there is a better way to solve for it

joshblack:
@andreancardona this makes me think about the focus style itself, I'm curious if the shadow should be inset then instead? Would require a sync with design but it seems like our focus styles should probably stay scoped inside of the container instead of growing otherwise we do have to add in some extra spacing to accommodate for it 🤔

andreancardona:
> kes me think about the focus style itself, I'm curious if the shadow should be inset then instead? Would require a sync with design but it seems like our focus styles should probably stay scoped inside of the contai

Yeah that makes a lot more sense. What is the best next step then in this situation?

packages/components/src/components/radio-button/_radio-button.scss
```diff
@@ -71,7 +71,7 @@
     flex-shrink: 0;
     width: rem(18px);
     height: rem(18px);
-    margin-right: $carbon--spacing-03;
+    margin: 0.2rem $carbon--spacing-03 0.2rem 0;
```
joshblack: Would this make sense as:

```suggestion
    margin: 2px $carbon--spacing-03 2px 0;
```

Which would account for the outline size? Or is the 3.2px value exact? Wasn't sure where it was coming from 👀

andreancardona: @joshblack that makes sense! I guess I just thought we always want to use `rem` - just in general what is the rule of thumb around that? Re: when to use `rem` v `px`?

joshblack: @andreancardona let me know if I'm missing something, but it seemed like this margin was being used to offset the outline width which was in px. If so, and we used rem, then we might go past the size of the outline if the font-size increases even though the outline width itself didn't change.

joshblack (approved):

joshblack:
bump @dakahn if you have a sec to review today

tw15egan:
Just curious, if we're adding a margin to the top and bottom to account for the focus, should we also add some to the left side?

Screen Shot 2020-11-05 at 11 47 52 AM

dakahn (approved):
seems cool to me, pending @tw15egan 's questions

andreancardona:
> Just curious, if we're adding a margin to the top and bottom to account for the focus, should we also add some to the left side?
> 
> Screen Shot 2020-11-05 at 11 47 52 AM

good catch! updated!

PR #7134, opened 2020-10-26
docs(Checkbox): disable checkbox group heading with knob

### LINKED ISSUE #7125: [checkbox] disable heading when using disabled prop
The checkbox heading needs to be disabled when the disabled prop is on. Right now the checkbox labels and icons are disabled, but the headers are not.

Screen Shot 2020-10-23 at 4 31 42 PM

emyarod:
Closes #7125

This PR updates the checkbox playground story so that the checkbox group heading label is also modified when the `disabled` knob is on


tw15egan (approved):

tw15egan:
Tried out locally by using the `FormGroup` component, as it handles the legend/fieldset classes OOTB, but it doesn't look like it handles any disabled styles. 

I made a note of it in the ***codename project next*** prop audit so that we can pass down `disabled` to the FormGroup component and apply disabled styles to the label

laurenmrice (approved):
Looks great ! thank you 🎉

aledavila (approved):

PR #7148, opened 2020-10-27
fix(Tabs): reduce scroll speed on overflow nav buttons

### LINKED ISSUE #7121: Tabs on narrow width may skip some tabs on arrow click
If I resize the browser that includes Tabs component implementation to show say 4 tabs wide, pressing on the > arrow may skip some tabs in between.

- [x] `carbon-components` 10.21.0
- [x] `carbon-components-react` 7.21.1

I have tried both Chrome and FF.

Make the screen narrow to show 4 tabs only.  After clearing cache and refresh the page, then
1. Select the first tab.  Press > to the next page. Curl - Java - Node - Python >
image

2. The first time it worked (shows < Python - Go - .Net >)
image

3. Press more to the end (shows < Ruby - Swift - Unity > )
image

4.  Click back to the first few tabs and select Curl again. Then repeat step 1.  Click the > arrow once.
This time it jumps directly to further away and bypass the Go and .Net tabs
image

I do not know if there is a minimum width required on each tab and do they need to be of the same width always.

Movie:
scrollTabsBug

emyarod:
Closes #7121

This PR adjusts the scroll speed for Tabs overflow nav buttons so that nav items are less likely to be skipped over


aagonzales:
Yes speed is much better!

aagonzales (approved):
Much better speed.

joshblack (approved):

tw15egan (approved):

PR #7156, opened 2020-10-28
fix(TextArea): remove placeholder attributes if values are falsy

### LINKED ISSUE #7132: Textarea shows both placeholder and aria-placeholder, and thats an issue
Using accessibility-checker we see that there is an issue with textarea - its adding both placeholer and aria-placeholder which is an issue.

As an aside, in my code I do not add placeholder at all
```
<TextArea
        id='description'
        name='description'
        {...description}
        labelText={t('addSecret.secretForm.description')}
        invalid={!!(descriptionMeta.touched && descriptionMeta.error)}
        invalidText={descriptionMeta.error}
      />
```

image

emyarod:
Closes #7132

This PR prevents the TextArea component from having empty `placeholder` or `aria-placeholder` attributes which would raise a11y errors

#### Testing / Reviewing

Confirm that `placeholder` and `aria-placeholder` are not added to the component when their values are falsy


dakahn (approved):

andreancardona (approved):
@emyarod 🥳

PR #7162, opened 2020-10-28
feat(themes): add helper function to determine current theme

scottdickerson:
Closes #

No existing issue opened, I needed a way to programmatically determine the current theme object in use because I need to pass those colors to a canvas programmatically.  This requires the css `emit-custom-properties` variable be used in the mixin to apply the appropriate carbon theme.

#### Changelog

**New**

- themes.js new themes.js file with a helper function getCurrentTheme to examine the current `--ui-01` cssVariable and return which theme object is in use

**Changed**

- index.js refactored to prevent a circular dependency eror

#### Testing / Reviewing

Verify that based on which theme is in use in your top-level mixin, the correct theme object is returned


tw15egan:
Conversation for reference: https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1603820525238200

packages/themes/src/themes.js
```diff
@@ -0,0 +1,66 @@
...
+    return null; // cannot determine current theme
+  }
+
+  const ui01Hex = rgbToHex(ui01ValueString);
+  if (!ui01Hex || ui01Hex === '#000000') {
+    return null; // cannot determine current theme
+  }
+
+  // Check the hex color of ui01 to determine the appropriate theme
+  switch (ui01Hex) {
```
emyarod, line 47: would this switch statement need to be updated if the theme color values change in the future? wondering if we could pull these hex values automatically somehow instead of hard coding it

emyarod (commented):
.

packages/themes/src/themes.js
theiliad: good idea

packages/themes/src/themes.js
```diff
@@ -0,0 +1,66 @@
...
+
+export { g10, g90, g100, white, v9 };
+/**
+ * Dynamically determines and returns the current theme token map
+ *
+ * @param {node} domNode the dom node to examine to determine the current theme
+ * @param {string} cssVariablePrefix the prefix of the carbon css variables that was set at mixin time
+ * @returns {object} object with current theme tokens, null if we can't determine the current theme
+ */
+export function determineCurrentTheme(domNode, cssVariablePrefix) {
```
theiliad:
```suggestion
export function getCurrentTheme(domNode, cssVariablePrefix) {
```
?

theiliad: seems like it's actually returning the full theme

theiliad (changes requested):

theiliad:
Btw this would be really useful in Carbon Charts. Lmk if you need any help with reviews 🙂

packages/themes/src/themes.js
scottdickerson: this is an easy fix I think, thanks for catching

joshblack:
@scottdickerson thanks for making a PR! Real quick, could you share more about this use-case in product? For determining the theme, had always thought that it would be something the application would try and coordinate through class names or context.

packages/themes/src/themes.js
scottdickerson: I fixed this @theiliad

packages/themes/src/themes.js
scottdickerson: ok ready for re-review @emyarod

scottdickerson:
@joshblack we're using canvas to draw some lines in a charting library @carbon/charts.  Canvas doesn't use css internally unfortunately, so we need to programmatically detect the selected carbon theme that's in use so we can use the javascript tokens from themes.  Since we're a library that is used by multiple external consumers, we can't set the theme ourselves, but need to support whatever theme the customer is currently using.

joshblack:
@scottdickerson totally makes sense on the canvas note. I guess the question was more-so the design decision around deriving the theme instead of having the theme be provided from the caller of the library.

This might sound arbitrary but in the context of this package the need to have access to each of the themes for this type of function to work would mean that consumers can't tree-shake themes that they aren't using. Definitely appreciate the problem here, though, since it seems like you don't know what theme you're using until you can derive it which would necessitate having each theme available.

One potential way around this could be using the token names directly and deriving the values from these names. This also has the added benefit of working with custom themes, potentially, but would have additional runtime cost, I guess? The other way would be structuring the product such that they could provide the theme to the library.

Curious what your thoughts are here, so sorry about the delay in responses too. I really appreciate your time/patience here 🙏

joshblack:
Going to close this out due to inactivity, feel free to leave a comment or reach out and I can re-open!

PR #7165, opened 2020-10-29
feat(Tooltip): add preventFocusAfterDismiss property

moczolaszlo:
Hi everyone,
We are facing a focus jumping problem with the Tooltip. I'm not sure how can I explain it so i made a video to show you:
https://video.ibm.com/recorded/128384160

To fix this I added a `preventFocusAfterDismiss` boolean property. With this the Tooltip will add back focus to the last active element (using `document.activeElement`) and reduce the wait time of debounce to zero.
You can see this here:
https://video.ibm.com/recorded/128384159

#### Changelog

**New**

- Add preventFocusAfterDismiss property to Tooltip component to prevent focus jumping.


emyarod:
will double check on this but if I'm not mistaken, the focus returning to the trigger button is related an a11y requirement

moczolaszlo:
> will double check on this but if I'm not mistaken, the focus returning to the trigger button is related an a11y requirement

Probably yes, but this jumping isn't good ux.

moczolaszlo:
I think the `allcontributors` bot removed a blank line from the README.md and that was the problem of prettier. I reverted this line but it should be fixed in the bot.

emyarod:
just wondering, in your first video are you pressing <kbd>Tab</kbd> or <kbd>Esc</kbd> or something else to dismiss the tooltip? this may be a bug that we can resolve without adding an additional prop

moczolaszlo:
> just wondering, in your first video are you pressing Tab or Esc or something else to dismiss the tooltip? this may be a bug that we can resolve without adding an additional prop

I only pressing <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd>.

emyarod:
> I only pressing <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd>.

ok got it! I think we should check whether or not the tooltip contains focusable content, and if it doesn't then we can preserve the existing tab order. otherwise if it does contain focusable content, we will place focus back on the trigger button like we do currently

moczolaszlo:
@emyarod What do you think? Could it be fixed? I tried another solutions inside the Tooltip, but only this one worked.

emyarod:
@moczolaszlo yes I think it can be fixed, and this is how I described the expected component behavior:

> I think we should check whether or not the tooltip contains focusable content, and if it doesn't then we can preserve the existing tab order. otherwise if it does contain focusable content, we will place focus back on the trigger button like we do currently

does that make sense?

moczolaszlo:
Honestly, I don't know if this could fix the problem. 😄

emyarod:
is there anything in particular that makes you think that? it would be similar to what the `preventFocusAfterDismiss` prop controls, except that the behavior would be automatically enabled/disabled based on the content of the tooltip, if that makes sense. if you'd like, you can create an issue for this and we can take a look, unless you're willing to update your PR to implement this behavior

moczolaszlo:
> is there anything in particular that makes you think that? it would be similar to what the `preventFocusAfterDismiss` prop controls, except that the behavior would be automatically enabled/disabled based on the content of the tooltip, if that makes sense. if you'd like, you can create an issue for this and we can take a look, unless you're willing to update your PR to implement this behavior

But I think it's not enough. The focus will jump back if something focusable inside the Tooltip - when the user want to jump to the next element. It's just a half solution, or I don't understand.
Maybe @FrivalszkyP could help. 🙏

emyarod:
> The focus will jump back if something focusable inside the Tooltip - when the user want to jump to the next element

right, when there is focusable content in the tooltip, this is the expected behavior for a11y purposes. the content of the tooltip would determine the behavior of the focus after dismissing the tooltip

moczolaszlo:
> > The focus will jump back if something focusable inside the Tooltip - when the user want to jump to the next element
> 
> right, when there is focusable content in the tooltip, this is the expected behavior for a11y purposes. the content of the tooltip would determine the behavior of the focus after dismissing the tooltip

Ok, it's better than now. 😄 But how can I check the content? Or should we check the `focusTrap` and `selectorPrimaryFocus ` properties?

emyarod:
I think you can just query the tooltip body for focusable content. This may be helpful as a reference https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/internal/FloatingMenu.js#L339-L345

moczolaszlo:
> I think you can just query the tooltip body for focusable content. This may be helpful as a reference https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/internal/FloatingMenu.js#L339-L345

Thank you, I try this and I will refresh this PR. Or it's better if I make a new?

moczolaszlo:
@emyarod 
I made a test site: https://pages.github.ibm.com/Laszlo-Moczo/carbon-test-preview/tooltip/?path=/story/tooltip--with-link (you need w3id, I hope you have 😄 ). But on this page the behaviour of Tooltip is not the same than in the first video. Because in the video `focusTrap` is false.

emyarod:
> Thank you, I try this and I will refresh this PR. Or it's better if I make a new?

@moczolaszlo that is up to you since either way will be fine, but could you also create a GitHub issue to help us better triage and track this bug?

regarding the test site example, it seems that the undesired focus behavior you originally described still exists. On a side note, `focusTrap` was only meant to be modified in fully uncontrolled tooltips (which is the reason why the knob is only shown in the uncontrolled example in our storybook)

moczolaszlo:
So the `focusTrap` is true as default. And if I don't change it to false I need to press <kbd>ESC</kbd> to close the Tooltip and move on. I talked with the QA guys and we decided that this is enough us for now. So I close this PR.
Thank you for your help and for your time! And sorry, It was a mistake from me I didn't know enough of the Tooltip and how it's works.

PR #7175, opened 2020-10-29
docs(data-table): switch checkboxes to uncontrolled

### LINKED ISSUE #7143: Data table with checkmark column example: checkboxes aren't interactive
Are the checkboxes supposed to be interactive in the [data table with checkmark column example](https://carbon-components-react.netlify.app/iframe.html?id=datatable--with-checkmark-columns)? I can't get them to work with mouse or keyboard.

joshblack:
Closes #7143

Update checkbox example in DataTable to be uncontrolled

#### Changelog

**New**

**Changed**

- Update checkbox example in DataTable to be uncontrolled

**Removed**

#### Testing / Reviewing

Verify checkbox story allows you to check/un-check the checkboxes in the table


carmacleod (approved):
Much better - thanks @joshblack !

Question: "checkbox" is a pretty terrible name for a checkbox, because in the screen reader it sounds like:
> checkbox checkbox checked

or
> checkbox checkbox not checked

Can the consumer specify the checkbox name? If not, then you probably want to at least make it the same as the column containing the checkboxes ("Enabled" in this case).

joshblack:
@carmacleod yeah, the name is generated for the story / isn't contextual. People who combine these components will supply a relevant name for the domain they're working in 👍

dakahn (approved):
great work sir

emyarod (approved):

PR #7176, opened 2020-10-29
chore(components): remove flag warning

joshblack:
We currently warn when folks diverge from the default feature flag set. At times this can be helpful, but unfortunately for most users this ends up creating noise because they're using it for the 16 column grid.

This PR removes the warning when feature flags diverge, in general the assumption would be that if you're using a feature flag that this is not related to semver.

Curious to see what people think of this change / if it's helpful 👀  Totally understand if we want to keep it in!

#### Changelog

**New**

**Changed**

**Removed**

- Warning for changed feature flags


andreancardona (approved):
@joshblack 🎉 looks good!

tw15egan (approved):
Sounds good to me 👍

PR #7188, opened 2020-10-30
fix(react): remove default prop for hideLabel

joshblack:
We received a support comment in Slack that `hideLabel` warns even if a user is not using that prop. This is because `hideLabel` is specified as a `defaultProp`.

To remove the warning, this PR removes the default prop and instead assigns `hideLabel` to `true` in `render` instead.

#### Changelog

**New**

**Changed**

- Remove `hideLabel` from `defaultProps`, default to `true` in render

**Removed**


emyarod (approved):
looks good to me but snapshots need to be updated to pass CI

tw15egan (approved):

PR #7194, opened 2020-10-31
fix: prevent feedback from appearing in the inline-code-snippet field

NBNARADHYA:
Fixes first part of carbon-design-system/carbon-website#1932

In the [live demo](https://www.carbondesignsystem.com/components/code-snippet/code#live-demo), the feedback text was appearing in the code-snippet field (for inline type). This has now been prevented from appearing.

#### Changelog

**Removed**

- Feedback text from inline code-snippet field

##### Before

97487907-e58a3a80-1933-11eb-928c-6c0b213051d3

##### After
Peek 2020-10-31 11-58


packages/react/src/components/Copy/Copy.js
```diff
@@ -68,7 +68,6 @@ export default function Copy({
         (!children && (animation ? feedback : other['aria-label'])) || null
       }>
       {children}
-      {animation ? feedback : other['aria-label']}
```
emyarod, line 71: I don't think we can remove this line since it announces the feedback to screenreaders IIRC

emyarod (commented):
I believe this issue is a website-only issue since it does not appear in the storybook. Seems a font size style rule is being overridden in the Carbon website

packages/react/src/components/Copy/Copy.js
NBNARADHYA: Could we change the `aria-label` to the button as follows ?
`<button ... aria-label={animation ? feedback : other['aria-label']}> ... </button>`

NBNARADHYA: Sorry for the late reply though. I was busy with the classes!

emyarod: I'm not sure if we need to change the component at all since the issue you reported is only found in the website. as I mentioned earlier I believe a font-size style rule is being overwritten in the website stylesheet but I haven't looked into it closely

joshblack:
Going to close this out due to @emyarod's feedback that this is most likely related to the website as the component does not reflect this behavior in the storybook.

Feel free to re-open if I've made a mistake here!

PR #7199, opened 2020-11-02
feat(Tile): add optional expand/collapse labels for chevron

### LINKED ISSUE #6802: Proposal: Allow "ExpandableTile" to accept a label for the expansion trigger
Hey, was wondering if we might be able to extend the `<ExpandableTile />` component a bit. What we’re looking to do is add the ability to label the expansion trigger, as well as set the positioning/color of the trigger (see image).

Happy to contribute this back if there are no objections.

image

ryanomackey:
Closes https://github.com/carbon-design-system/carbon/issues/6802

To help draw attention to the expand/collapse functionality, this PR adds two new props to the `ExpandableTile` component that allow the developer to add text labels next to the chevron.

#### Changelog

**New**

- Adds two new props to `ExpandableTile`:
  - `tileCollapsedLabel`: When "collapsed", a label to appear next to the chevron (e.g., "View more").
  - `tileExpandedLabel`: When "expanded", a label to appear next to the chevron (e.g., "View less").

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

Demo GIF:


packages/react/src/components/Tile/Tile.js
```diff
@@ -548,6 +560,8 @@ export class ExpandableTile extends Component {
       handleClick, // eslint-disable-line
       tileCollapsedIconText, // eslint-disable-line
       tileExpandedIconText, // eslint-disable-line
+      tileCollapsedLabel, // eslint-disable-line
+      tileExpandedLabel, // eslint-disable-line
```
emyarod:
```suggestion
      tileCollapsedIconText,
      tileExpandedIconText,
      tileCollapsedLabel,
      tileExpandedLabel,
```

are the eslint-disable comments required here? I actually don't think they're needed for `tileCollapsedIcontText` or `tileExpandedIconText` either since they're all referenced a few lines further down

tw15egan: Doesn't seem like they are, should be good to remove them 👍

packages/react/src/components/Tile/Tile.js
```diff
@@ -442,7 +452,9 @@ export class ExpandableTile extends Component {
     onBeforeClick: () => true,
     handleClick: () => {},
     tileCollapsedIconText: 'Interact to expand Tile',
+    tileCollapsedLabel: null,
```
tw15egan: Do we need to add them to the `defaultProps` if they are null?

aagonzales:
Not sure if this is related to this PR or would be a separate one but the padding for chevron (and now label) is off. It should be have that 16px padding on right and bottom.
image

tw15egan:
Hey @ryanomackey, looks like this proposal was accepted but has a couple of issues that need to be addressed. Do you have time to work on these?

ryanomackey:
Thanks, @tw15egan. Should be ready for another review.

packages/components/src/components/tile/_tile.scss
```diff
@@ -105,11 +105,15 @@
 
   .#{$prefix}--tile__chevron {
     position: absolute;
-    right: 0.5rem;
-    bottom: 0.5rem;
+    right: $carbon--spacing-05;
+    bottom: $carbon--spacing-05;
+    display: flex;
+    flex-direction: row;
```
tw15egan: This is the default property, so we can probably remove this line

tw15egan (approved):
Tiny style nit, but otherwise looks great. Thanks for contributing! 👍 ✅

laurenmrice (approved):
Props are working correctly and spacing looks good now. Thank you! 🙌🏻

packages/components/src/components/tile/_tile.scss
ryanomackey: Ah, yeah. Good call.

emyarod (approved):

Issue #7159, opened 2020-10-28
[code snippet] single line and multi-line bugs


laurenmrice:
There are a couple different bugs with the `single line` and `multi-line` code snippets.

-------------------------------------------------------------------------------------------------------------------------

**Single line:**
- The active state is not showing up when clicking on the copy button. It should be active-ui on click.
![code snippet active](https://user-images.githubusercontent.com/43969356/97491064-08b6e900-1938-11eb-85bf-e58eab973b89.gif)

- The horizontal scroll does not have a fade style. (This may be resolved with Andrews fix for horizontal tab scrolling.)
- When the light prop is active, the dark themes have incorrect hover states for the copy button and show more button. They should be using `hover-light-ui` for hover. (this is also happening for inline code snippet)
-------------------------------------------------------------------------------------------------------------------------

**Multi-line:**
- The active state is not showing up when clicking on the copy button (same problem with single line above^). 
- When the light prop is active, the dark themes have incorrect hover states for the copy button and show more button. They should be using `hover-light-ui` for hover.
- The copy button should have a 8px top and right padding within the field background so that it vertically aligns with the chevron icon in the show more button below it. 
![Artboard](https://user-images.githubusercontent.com/43969356/97491037-ffc61780-1937-11eb-94f4-e05246ce5469.png)

tw15egan:
I'm also wondering if we should investigate adding in copy functionality by default (https://github.com/carbon-design-system/carbon/issues/6995)

emyarod:
makes sense, if we're all on board with it then I can investigate that as well. I guess the biggest thing to look into would be browser and OS support

tw15egan:
It looks like `clipboard` has gained quite a bit of native browser support (at least for copy), we may even be able to just handle this natively without an external library: 

https://caniuse.com/clipboard

https://stackoverflow.com/questions/39501289/in-reactjs-how-to-copy-text-to-clipboard

emyarod:
for the light button hover styles, it looks like we are already using the `hover-ui-light` token https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/code-snippet/_code-snippet.scss#L373. Should the value of that token be changed? (draft PR here https://github.com/carbon-design-system/carbon/pull/7214)

and for the overflow indicators, what should the color tokens be? I have this ready to put in a separate PR but just need clarification on the colors

laurenmrice:
## **Hover issue:**

[...]

PR #7214, opened 2020-11-03
fix(code-snippet): update component to spec

emyarod:
Part of #7159

This PR adds an active style to the code snippet copy buttons and repositions the button to be aligned with the "show more" chevron in multiline snippets.

#### Changelog

**New**

- copy button active styles

**Changed**

- multiline code snippet copy button positioning
- light hover token for copy button

#### Testing / Reviewing

Confirm the code snippet matches the spec


joshblack (approved):

dakahn (approved):

laurenmrice (changes requested):
- The light prop needs to also have an active state change for the copying functionality for inline/single line/multi-line code snippets. It is not showing up. 

- I think the multi-line code snippet copy button is too large which is why it is misaligning with the ghost button below it. It needs to be a 32px icon button instead of 40px.
Artboard

laurenmrice:
So it was noticed that we need to make a token for `active-light-ui` for the light prop to achieve the correct colors.
$active-light-ui

Also for the current `hover-light-ui` token, it would be better to have calculated hover values for the dark themes like we do for the light themes.
$hover-light-ui

emyarod:
new active token and hover token updates will be here https://github.com/carbon-design-system/carbon/pull/7275

emyarod:
currently waiting on #7275 for dark theme changes

laurenmrice (changes requested):
When the light prop and hide copy functionalities are both on, the inline code snippet is getting a hover. There should be no hover state when hide copy is on.  
Nov-12-2020 11-31-32


Everything else looks great!

laurenmrice (approved):
Looks great! Thank you 🙌🏻

PR #7215, opened 2020-11-03
feat(numberinput): add support for "any" value for step prop

### LINKED ISSUE #6930: NumberInput component step prop did not support "any"
7.12.0 But issue exist on latest version too.

- [ ] `carbon-components`

## Detailed description

The type of prop 'step' should be string and pattern prop should be exposed.
1.HTML5 shows validation message if user put any value into number input which is not a multiple of step
For ex: If step=10 and value=10 and user enters 15 or 15.30 then on form submit it would shows validation message as
'Please enter a valid value. The two nearest valid values are 10 and 20'
2.So in this case if user do not want to show any validation message then we have pass step='any'.
But in carbon step prop is of number type. So the type should be updated to string so that user could pass 'any'
if does not want any client validation.
Currently pattern prop is hard coded and it could be exposed as a properties where user can set his own pattern.
As we have scenario where user can set decimal values like this
step="any" pattern="^\d*(\.\d{0,2})?$"

NumberInput

> see changed?
Step should accept "any" and "pattern" should be exposed as a prop. So that user can enter decimal value as per required format without any HTML5 violation.

Issue occurs on all browsers.

> should be aware of?

1. Enter a decimal value in numberinput and hover mouse over it browser will show invalid error.
2. Enter a non decimal number which should not be a multiple of step value, and hover mouse over it browser will show invalid error.

>
> - Style and vanilla JS:
> - React:

## Additional information

- Screenshots or code
image
image

- Notes

davidicus:
Closes #6930 

Currently Carbon does not allow for string values for the `step` prop. This is preventing users from using "any" as a value (without proptype warning) which in turn triggers validation when input value is submitted. This PR allows for both numbers and strings. 

#### Changelog

**New**

- added typeof check to see if value was a string and prevent rendering or step logic if so in NumberInput.js
- added test to ensure that step buttons were not rendered when the value of "any" was supplied

**Changed**

- changed proptype declaration of `step` to `oneOfType` in NumberInput.js
- changed NumberInput story to showcase the use of this prop value

#### Testing / Reviewing

- go to NumberInput story
- change knob to any of the different options
- see that buttons go away and no error message is present on hover when adding decimal to number input


packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -135,7 +135,7 @@ class NumberInput extends Component {
     /**
      * Specify how much the valus should increase/decrease upon clicking on up/down button
      */
-    step: PropTypes.number,
+    step: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
emyarod:
```suggestion
    step: PropTypes.oneOf([PropTypes.number, 'any']),
```

not 100% sure if this works but could we narrow down "string" to "any"? or are there other string values we should be accounting for?

emyarod (commented):
just to make sure I have a handle on what the original issue is, could we narrow down the expected type from "string" to "any"? or are there other string values we should be accounting for?

on a side note, can you update your PR title to match our [commit message guidelines](https://github.com/carbon-design-system/carbon/blob/master/.github/CONTRIBUTING.md#commit-tip)?

packages/react/src/components/NumberInput/NumberInput.js
davidicus: @emyarod I think we can do something like this:
```
PropTypes.oneOfType([
    PropTypes.number,
    PropTypes.oneOf(['all'])
])
```
if it is preferred.  I will update my PR title.

davidicus: This is the only string value allowed for this attribute.

davidicus: I have made the suggested change to the proptype declaration. It will now throw a warning if any other string value is entered beside "any"

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -244,7 +244,7 @@ class NumberInput extends Component {
         ? (min !== undefined && value > min) || min === undefined
         : (max !== undefined && value < max) || max === undefined;
 
-    if (!disabled && conditional) {
+    if (!disabled && conditional && typeof step !== 'string') {
```
joshblack: With `step` being either a number of `'any'` would it make sense to have the check be `step !== 'any'`?

joshblack (commented):
Definitely makes sense, thanks for updating the prop types. Left a small note but otherwise looks good.

For other reviewers, here are the docs for the `step` attribute which supports a special value of `any` that we don't currently handle: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step

packages/react/src/components/NumberInput/NumberInput.js
emyarod: thanks for the update! yeah if I'm not mistaken we can also change some of the type checks for strings to specifically look for "any" now

joshblack:
FYI @aagonzales the visual review bit is coming from the step of "any" causing the arrows to disappear (since there is no step size)

Steps to see:

- Visit Netlify link for storybook
- Go to NumberInput story
- Go to step knob and select `any`

aagonzales:
I guess if its saying there are no defined steps when using `any` then it makes sense to not show the up/down arrows because there are no steps defined. So I guess this is ok?


The storybook only allows 2 and 10 as steps but I'm guessing that's only for the purpose of demoing? In actuality a dev could set the step to be any number? 

I am curious why setting the step to `1` is less beneficial than `any`. The user would still have the ability to type any specific value if the step is `1` correct and then the step arrow would still appear.

tw15egan:
@aagonzales based on the original issue, you can bypass HTML5 validation by passing in `any` to the `step` attribute of a number input. Would still get validation if `step` is a number value, [see here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step)

specifically

> The default stepping value for number inputs is 1, allowing only integers to be entered—unless the stepping base is not an integer.

aagonzales:
_Looks up integer_

Ok I think I get it. Then yes its fine that the step arrows won't show when `any` is applied to the step.

aagonzales (commented):
I just want to make sure this isn't a problem in the component and just a storybook thing. But when i switch from a number step to `any` the size of the input changes. 

Nov-06-2020 14-26-22

guigueb:
I'm late to the party, but you are not reading "A string value of any means that no stepping is implied, and any value is allowed" correctly.

In this PR, 'any' is being read as "No stepping is going to be done, so we will remove the up/down buttons (oh - if so, we should stop up/down key actions too)".
But that is not what the mozilla definition of `step='any'` means.
Reading the definition of step we see that it refers to `the granularity that the value must adhere to`.
Even setting `step='any'` in the live example on the mozilla site does not stop stepping.
So, `step='any'` means, any value is valid.

The mozilla definition of step is for validating value (and for inc/dec the value... to the next **`valid`** value).
Examples:
step value of 1: value is valid only if it is an integer
step value of 10: value is valid only if it is an integer that is a multiple of 10
step value of 0.1: value must end in (.0, .1, .2, .3, .4, .5, .6, .7, .8, .9)
step value of 0.2: value must end in (.0, .2, .4, .6, .8)
[step is kinda like a modulus validator, value is only valid if `value/step = zero`]
[step is also used to inc/dec value... to the next **`valid`** value - given { value:"1.111", step:"1" } an increment will set value to "2" not "2.111"]

All that being said... I don't like the mozilla definition of step.
**`I think Carbon NumberInput should either change the behavior/definition of step or define a new prop.`**
**`Adding a newProp - stepIncDec - would allow those wanting the mozilla step behavior to still work`**

Maybe Carbon should adopt what we in Cognos Toolkit's NumberInput have re-defined step to mean...
- new prop stepIncDec
- be the number by which value is incremented/decremented
- set to zero disallows incremented/decremented (removing up/down button and key actions)
- if stepIncDec set, step is defaulted to 'any'
Examples:
stepIncDec set to 1.5; value 10 becomes 11.5, value 1 becomes 2.5, value 1.1 becomes 2.6, value 1.11 becomes 2.61

As long as we are here there should be another prop added to NumberInput... scale.
Which would, like min and max, be used to constrain value (like the mozilla granularity tried to do).
Examples:
scale 0 = valid values are integers
scale 1 = valid values only allow one digit to the right of the decimal =  valid 1.1, invalid 1.11
scale 2 = valid values only allow two digits to the right of the decimal =  valid 1.11, invalid 1.111

davidicus:
> I just want to make sure this isn't a problem in the component and just a storybook thing. But when i switch from a number step to `any` the size of the input changes.
> 
> Nov-06-2020 14-26-22

It's been a minute but finally getting back to this. Is this above an issue if users can define the width they want this input to take up? I honestly can not see why this size change is happening.

tw15egan:
@davidicus I believe the number input width changes based on the `max` property. I recently updated the component so that it is `width: 100%` to align with other form elements, so this issue may not be relevant anymore

davidicus:
Since the size change is no longer an issue can we merge this in? Or do you need anything else from me?

guigueb:
I left a comment earlier that I'm not sure has been read.... https://github.com/carbon-design-system/carbon/pull/7215#issuecomment-741053196

Short story:
Setting step to 'any' should NOT remove the inc/dec chevrons.
Settings step to 'any' effects onthe inc/dec validation, but should NOT stop inc/dec.

PaulJThompson:
I really think setting step="any" _should not_ remove the increment/decrement buttons. The current behaviour defined in MDN is that the increment/buttons are still available (and default to a stepping of 1) if step="any" is specified. We ought to have a very good reason before deviating from the standard behaviour and I don't see any such reason articulated in this thread.

davidicus:
The concern is with the validation so can totally add back the buttons and default to stepping by 1 when `step`'s value is a string.

guigueb:
To be clear, step should default to 1 when step is not given or is a string that does not represent only a number.
If step is a number or a string that represents only a number, step should be that number.

aagonzales (changes requested):
Seems fine expect when I do trigger an error (outside the range) the error icons is misaligned. 
image

tay1orjones:
I think it's worth mentioning that the MDN article isn't the target quite as much as the (more detailed) [html spec](https://html.spec.whatwg.org/multipage/input.html#the-step-attribute) is. 

> 3. ...if the attribute's value is an ASCII case-insensitive match for the string "any", then there is no allowed value step.

Having "no allowed value step" does not appear to impact the steppers, but does cause the `stepUp()` and `stepDown()` methods on the input to throw an ["InvalidStateError" DOMException](https://html.spec.whatwg.org/multipage/input.html#common-input-element-apis:concept-input-step). There might be an argument in there that browsers shouldn't provide steppers due to this, but that's not the browser behavior today.

tay1orjones (changes requested):
Echoing recent comments, at a high level I think this could get merged in with just a few changes: 

1. Steppers should not be removed when `step="any"`, so we match the current behavior of browsers.
   - If you really want to hide the steppers, you can still configure `hideSteppers` introduced in #8646
1. When `step="any"`, steppers should fall back to the default step as defined in the html spec
1. Correct the misalignment Anna identified above

PR #7216, opened 2020-11-03
fix(reset): fix reset specifying anchor color

### LINKED ISSUE #7206: bug(theme): reset causing issue in theme
The color for anchors is hard coded. It should point to $interactive-01.

This is causing an issue because even when updating $interactive-01 for custom themes the anchor link color is still set to blue because its hardcoded.

File:
https://github.com/carbon-design-system/carbon/blob/master/packages/type/scss/_reset.scss

Line: 78

Expect this to point to token also for maintainability sake

DusanMilko:
Closes: https://github.com/carbon-design-system/carbon/issues/7206

Currently reset is setting hard hex color for anchor that is not ideal.
Color is not a token so causes issues in regards to theme


andreancardona:
@DusanMilko hello! it seems the tests are failing - you will have to run `yarn test -u` to update the snapshot

packages/type/scss/_reset.scss
```diff
@@ -75,7 +75,7 @@
   }
 
   a {
-    color: #0062ff;
+    color: $interactive-01;
```
joshblack: If I remember correctly, I _think_ it's a hex value due to this token coming from the themes package. Since themes depends on type, if we had type depends on themes then we would have a cycle between the two.

DusanMilko: then should we remove this hard hex from the reset?

joshblack: @DusanMilko maybe we could swing something like:

```scss
  a {
    @if global-variable-exists($carbon--theme) and
      map-has-key($carbon--theme, 'link-01')
    {
      color: map-get($carbon--theme, 'link-01');
    } @else {
      color: #0062ff;
    }
  }
```

?

tw15egan: Do we even need to reset the color? It doesn't look like we reset any other color values?

joshblack: @tw15egan I think you're right that we don't need to reset colors, I assume that when the default type mixin was made it was just translating over from the v9 default styles which we could totally just drop in v11 👍

DusanMilko:
> @DusanMilko hello! it seems the tests are failing - you will have to run `yarn test -u` to update the snapshot

Got this error:

● Validation Error:

  Preset jest-config-carbon not found.

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html

DusanMilko:
@joshblack why is this failing?

tw15egan:
Screen Shot 2020-11-23 at 2 58 55 PM

> Error: argument `$name` of `global-variable-exists($name)` must be a string

Seems like the ` @if global-variable-exists($carbon--theme) and` is causing it, would it work if we switch it to `#{$carbon--theme}`, or would this always evaluate true?

DusanMilko:
> Screen Shot 2020-11-23 at 2 58 55 PM
> 
> > Error: argument `$name` of `global-variable-exists($name)` must be a string
> 
> Seems like the ` @if global-variable-exists($carbon--theme) and` is causing it, would it work if we switch it to `#{$carbon--theme}`, or would this always evaluate true?

Not sure, i know josh mentioned removing this possibly in carbon 11, is there a big reason we can't remove it now or atleast move out of reset and move into theme where it makes more sense? (I don't feel like its a breaking change but I differ to you)

joshblack:
Hey all! Sorry for the delay over break, `global-variable-exists` takes in a string which I think is causing it to fail. Instead of `global-variable-exists($carbon--theme)` it would be `global-variable-exists('carbon--theme')`

Reference: https://sass-lang.com/documentation/modules/meta#global-variable-exists

DusanMilko:
> Hey all! Sorry for the delay over break, `global-variable-exists` takes in a string which I think is causing it to fail. Instead of `global-variable-exists($carbon--theme)` it would be `global-variable-exists('carbon--theme')`
> 
> Reference: https://sass-lang.com/documentation/modules/meta#global-variable-exists

Still seems to be failing

packages/type/scss/_reset.scss
```diff
@@ -75,7 +75,13 @@
   }
 
   a {
-    color: #0062ff;
+    @if global-variable-exists('carbon--theme') and
+      map-has-key($carbon--theme, 'link-01')
+    {
+      color: map-get($carbon--theme, 'link-01');
+    } @else {
+      color: #0062ff;
```
tw15egan: Based on the snapshot I think we want this to be #0062fe 

```suggestion
      color: #0062fe;
```

Screen Shot 2020-12-08 at 4 20 34 PM

DusanMilko: thanks for the help guys

tw15egan (approved):

joshblack (approved):

PR #7218, opened 2020-11-03
feat(theme): adding optional utility classes

DusanMilko:
Proposing addition of optional utility classes to specify colors.

This helps with making theme support easier for components and instances of javascript component libraries.

Format: 
PROPERTY--TOKEN

To prevent generation of too many classes, tokens have appropriate properties assigned.

OPEN TO FEEDBACK 👍

### DIFF
--- a/packages/themes/scss/_utility-classes.scss
+++ b/packages/themes/scss/_utility-classes.scss
@@ -1,0 +1,121 @@
+$theme-class-map: (
+  'ui-background': (
+    'var': $ui-background,
+    'props': ('background')
+  ),
+  'interactive-01': (
+    'var': $interactive-01,
+    'props': ('background', 'color', 'fill')
+  ),
+  'interactive-02': (
+    'var': $interactive-02,
+    'props': ('background', 'color')
+  ),
+  'interactive-03': (
+    'var': $interactive-03,
+    'props': ('background', 'color')
+  ),
+  'interactive-04': (
+    'var': $interactive-04,
+    'props': ('background', 'color')
+  ),
+  'danger': (
+    'var': $danger,
+    'props': ('background', 'color')
+  ),
+  'ui-01': (
+    'var': $ui-01,
+    'props': ('background')
+  ),
+  'ui-02': (
+    'var': $ui-02,
+    'props': ('background')
+  ),
+  'ui-03': (
+    'var': $ui-03,
+    'props': ('background', 'border-color')
+  ),
+  'ui-04': (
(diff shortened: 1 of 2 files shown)


joshblack:
Love this, reminds me of some explorations I did for this and spacing: https://www.sassmeister.com/gist/552291162658f7054aa668d98bc291e4?token=a0ab2532571148490a9f6f5bf22712a17038cdd9&scope=gist,read:user

When thinking about color earlier, it seemed like Tailwind leaned on:

- `background-color` as `bg-{color}`
- `text-color` as `text-{color}`
- `border-color` as `border-{color}
- `fill` seems like they use `fill-current` to set it to the text color 🤔 

Curious what you think of those conventions, along with how they include variants for things like focus/hover or even responsive names 👀

DusanMilko:
> Love this, reminds me of some explorations I did for this and spacing: https://www.sassmeister.com/gist/552291162658f7054aa668d98bc291e4?token=a0ab2532571148490a9f6f5bf22712a17038cdd9&scope=gist,read:user
> 
> When thinking about color earlier, it seemed like Tailwind leaned on:
> 
> * `background-color` as `bg-{color}`
> * `text-color` as `text-{color}`
> * `border-color` as `border-{color}
> * `fill` seems like they use `fill-current` to set it to the text color 🤔
> 
> Curious what you think of those conventions, along with how they include variants for things like focus/hover or even responsive names 👀

I find using the property name as the class makes mapping the easiest and understandable for devs. These classes are used by devs so making it easier for devs to understand in my opinion is ideal. If we shorthand some stuff like background to bg or swap color to text it makes it harder for devs to pick up the utility classes. In css you never do text: HEX or bg: HEX

To change background color use class background--{TOKEN} = in css background: TOKEN
To change text color use class color--{TOKEN} = in css color: TOKEN
To change fill color use class fill--{TOKEN} = in css fill: TOKEN

The mapping is 1-to-1, super easy for devs to pick up and understand.

UPDATED:
updated to allow pseudo

joshblack:
@DusanMilko down for whatever naming makes sense, the structure above came from Tailwind as noted in the comment 👍 

I opened up an RFC as well for this use-case more broadly if you're interested: https://github.com/carbon-design-system/carbon/pull/7295

100% still a WIP but we've gotten a handful of requests for this and was curious what people would think if we just had a package for it and let folks configure what gets emitted/generated similar to Tailwind 👀

joshblack:
Gonna close this out for the broader RFC in #7295 👀

PR #7221, opened 2020-11-04
Fix issue in Jest setup, fix failing tests

### LINKED ISSUE #7186: Impossible to render <Tab> without deprecation warnings

- [ ] `carbon-components`

## Detailed description

`<Tab>`, `<Tabs>`

> see changed?

It should be possible to use any Carbon component in React without raising deprecation warnings. Our unit tests are rigged to fail if anything is output to the console.

`carbon-components-react@7.23.0` or thereabouts has modified the props on `<Tab>` to make it so that the props `tabIndex` and `href` are deprecated. However, the `<Tab>` component also specifies a default value of `'#'` for the `href` prop and a default value of `0` for the `tabIndex` prop. This means that if we specify the `href` or `tabIndex` prop, we see the deprecation warning, but if we *don't* specify the `href` or `tabIndex` prop, or set `href={null}` or `tabIndex={null}`, the default value is used instead and we still see the deprecation warning. There is no way to make the warnings go away.

Local unit testing outside of a web browser, using Enzyme.

10

> should be aware of?

IBM App Connect Enterprise. No urgent ship dates.

https://codesandbox.io/s/laughing-chaum-78qxi

Check the browser console.

## Additional information

I believe I have experienced a similar issue in Carbon before (I can't find the issue though), and I was able to work around it temporarily by deleting the default props:

```js
delete Tab.defaultProps.href
delete Tab.defaultProps.tabIndex
```

However, though this works for `href`, it doesn't work for `tabIndex`, because the parent component `<Tabs>` sets the `tabIndex` prop on all of its child `<Tab>`s no matter what.

joshblack:
Closes #7186 

While working through #7186 I noticed that our test suite should have been failing for the deprecation notice, but then it wasn't. When digging in I noticed a bug with how we error out on console statements that were swallowing errors. This PR fixes that issue and any corresponding tests that failed as a result.

#### Changelog

**New**

- Add codemod for some prop renames

**Changed**

- Update our Jest test setup to account for errors in `beforeEach`
- Update tab to not show deprecations by default
- Button size renames
- TileGroup test update
- TimePicker test updates
- Fix DropdownSkeleton deprecation with inline prop
- Search size update
- HeaderMenu test updates
- Toolbar test updates
- InlineLoading test updates


tw15egan (approved):
LGTM 👍 ✅

joshblack:
bump @aledavila @andreancardona if you all have a sec today 👀 

Also free to go over anything if there are any questions!

andreancardona (approved):
🙌  looks great @joshblack !!

PR #7224, opened 2020-11-04
chore(docs): update AccordionItem docs

### LINKED ISSUE #6951: AccordionItem onclick open state callback
### Summary

The onClick call back for AccordionItem should have an argument passed to the callback indicating whether the action was to open or close the AccordionItem

For development, no design changes.

### Justification

Currently, there is no way to tell (from the component it's self) whether a call to `onClick` was due to the user opening or closing the accordion, this info should be made available to the parent component.

### "Must have" functionality

`onClick` and `onHeadingClick` should have argument passed to callback indicating whether the click was to open or close the accordion.

```
onClick(isOpen: bool) => {
  // callback
}
```

Do you want this work within a specific time period? Is it related to an
upcoming release?
Sooner the better, if this is accepted I can contribute to the code.

### Available extra resources

I can contribute to the code if the proposal is accepted.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6951

Adds in a short blurb about obscure `AccordionItem` prop

#### Changelog

**New**

- docs about `onHeadingClick`

#### Testing / Reviewing

Ensure the docs help inform about the `onHeadingClick` prop

### DIFF
--- a/packages/react/src/components/Accordion/Accordion.mdx
+++ b/packages/react/src/components/Accordion/Accordion.mdx
@@ -106,6 +106,19 @@ for layout.
 </Accordion>
 ```
 
+### AccordionItem onHeadingClick
+
+You can use the `onHeadingClick` prop to determine the state of the
+`AccordionItem` when it is clicked. it will return
+
+```jsx
+<Accordion>
+  <AccordionItem onHeadingClick={(evt) => console.log(evt.isOpen)}>
+    Panel A
+  </AccordionItem>
+</Accordion>
+```
+
 ### AccordionItem title
 
 You can use the `title` prop to specify the accordion item's heading. The
--- a/packages/react/src/components/Accordion/Accordion.mdx
+++ b/packages/react/src/components/Accordion/Accordion.mdx
@@ -94,7 +94,9 @@ through to the underlying accordion header.
 
 The `className` prop passed into `AccordionItem` will be forwarded along to the
 underlying accordion header. This is useful for specifying a custom class name
-for layout.
+for layout. As a reminder, the callback signature will look like
+`{ isOpen: boolean, event: Event }` rather than the usual
+`event: Event, state: { isOpen: boolean }` until the next major version.
 
 ```jsx
 <Accordion>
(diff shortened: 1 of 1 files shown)


emyarod (commented):
I think we should also highlight that the callback signature is `{ event: Event, isOpen: boolean }` rather than the usual `event: Event, state: { isOpen: boolean }` until the next major version

joshblack (approved):

andreancardona (approved):

tw15egan:
@emyarod added 👍

packages/react/src/components/Accordion/Accordion.mdx
```diff
@@ -94,7 +94,9 @@ through to the underlying accordion header.
 
 The `className` prop passed into `AccordionItem` will be forwarded along to the
 underlying accordion header. This is useful for specifying a custom class name
-for layout.
+for layout. As a reminder, the callback signature will look like
+`{ isOpen: boolean, event: Event }` rather than the usual
+`event: Event, state: { isOpen: boolean }` until the next major version.
```
emyarod:
```suggestion
`({ isOpen: boolean, event: Event })` rather than the usual
`(event: Event, state: { isOpen: boolean })` until the next major version.
```

I guess I should've wrapped these with parentheses earlier in my comment

packages/react/src/components/Accordion/Accordion.mdx
```diff
@@ -106,6 +108,19 @@ for layout.
...
 ```
 
+### AccordionItem onHeadingClick
+
+You can use the `onHeadingClick` prop to determine the state of the
+`AccordionItem` when it is clicked. it will return
+
+```jsx
+<Accordion>
+  <AccordionItem onHeadingClick={(evt) => console.log(evt.isOpen)}>
```
emyarod:
```suggestion
  <AccordionItem onHeadingClick={({ isOpen }) => console.log(isOpen)}>
```

maybe something like this? or renaming `evt` since it isn't the Event object but is actually an object containing the Event object and the isOpen boolean

emyarod (approved):
looks good to me, just a couple comments

Issue #6539, opened 2020-07-24
Modal Overflow Indicator does not work with the enable-css-custom-properties feature flag 


kgibb:
## What package(s) are you using?

- [x] `carbon-components`
- [ ] `carbon-components-react`

## Detailed description

> Describe in detail the issue you're having.

The overflow gradient is not rendered when the `enable-css-custom-properties` feature flag is enabled. The CSS rule for the gradient has an invalid value.

```background-image: linear-gradient(to bottom, rgba(var(--cds-ui-01, #f4f4f4), 0), var(--cds-ui-01, #f4f4f4));```

`rgba` expects four arguments so will not work with the `var` syntax.

> Is this issue related to a specific component?

`Modal`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

As there is an issue using `transparent` as the start color in Safari, could another color custom property be added that has a `rbga` value? 

```
--cds-ui-01: #f4f4f4;
--cds-ui-01-transparent: rgba(244, 244, 244, 0);
```

```
-   background-image: linear-gradient(to bottom, $ui-01, $ui-01);
+   background-image: linear-gradient(to bottom, $ui-01-transparent, $ui-01);
```

> What version of the Carbon Design System are you using?

`10.14.0`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Watson Health - Application Framework

> Please create a reduced test case in CodeSandbox

Modal with scrolling content
https://the-carbon-components.netlify.app/?nav=modal

## Additional information

![image](https://user-images.githubusercontent.com/12196391/88381032-84cfec80-cd9d-11ea-92bb-48398997823b.png)

emyarod:
related #4345 https://github.com/carbon-design-system/carbon/issues/4426

yuriybelike:
Looks like this issue appears now in the docs site too: https://www.carbondesignsystem.com/components/modal/code/

The "With scrolling content" variant appears with no overflow indicator

![image](https://user-images.githubusercontent.com/60257988/151261506-6b3b6fa6-454c-485b-a489-5b69f2e10ef7.png)

sstrubberg:
Is this still an issue?

PR #7229, opened 2020-11-05
feat(CodeSnippet): add scroll overflow indicators

emyarod:
part of #7159

This PR implements scroll overflow indicators to the single line and multiline code snippets

#### Changelog

**New**

- overflow scroll indicators

#### Testing / Reviewing

Confirm that the overflow indicators appear on the correct sides only when scroll is possible

### DIFF
--- a/packages/components/src/components/code-snippet/_code-snippet.scss
+++ b/packages/components/src/components/code-snippet/_code-snippet.scss
@@ -447,6 +447,57 @@
     left: 50%;
   }
 
+  // overflow indicator
+  .#{$prefix}--snippet__overflow-indicator--left,
+  .#{$prefix}--snippet__overflow-indicator--right {
+    z-index: 1;
+    flex: 1 0 auto;
+    width: $carbon--spacing-03;
+  }
+
+  .#{$prefix}--snippet__overflow-indicator--left {
+    margin-right: -$carbon--spacing-03;
+    background-image: linear-gradient(to left, transparent, $ui-background);
+  }
+
+  .#{$prefix}--snippet__overflow-indicator--right {
+    margin-left: -$carbon--spacing-03;
+    background-image: linear-gradient(to right, transparent, $ui-background);
+  }
+
+  .#{$prefix}--snippet--light .#{$prefix}--snippet__overflow-indicator--left {
+    background-image: linear-gradient(to left, transparent, $ui-01);
+  }
+
+  .#{$prefix}--snippet--light .#{$prefix}--snippet__overflow-indicator--right {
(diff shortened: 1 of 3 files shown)


tw15egan:
Confirmed by running locally since the gradients don't render properly with css variables

Screen Shot 2020-11-09 at 1 03 30 PM
Screen Shot 2020-11-09 at 1 01 31 PM
Screen Shot 2020-11-09 at 1 00 52 PM

tw15egan (approved):
LGTM once the tokens are finalized 👍 ✅

emyarod:
the gradients should be rendering correctly with CSS variables, are you seeing otherwise? I am using the method I used for overflow tabs which gets around previous gradient issues we had with CSS variables

tw15egan:
@emyarod this is what I'm seeing in deploy preview: 

Screen Shot 2020-11-09 at 1 35 24 PM


I thought we had the same issue with Modal // Has scrolling content 

I thought this was just a storybook implementation issue but should work when using css variables in prod, so not a big deal just odd that it's not rendering for me 🤔 

Screen Shot 2020-11-09 at 1 36 35 PM

tw15egan:
I thought this was just a storybook implementation issue but should work when using css variables in prod

emyarod:
here are the deploy previews on my end on multiple platforms and browsers

image

image

modal (https://github.com/carbon-design-system/carbon/issues/6539) still has the issue because it isn't using the method I tried with tabs (and now this) so I will be backporting it to modal. storybook uses css variables under the hood, so for modal it will be broken in prod as well

my guess is for your single line deploy preview screenshot the fade area is probably not large enough to be visible, but it's there (so it would just be a style adjustment)

aledavila (approved):

laurenmrice (changes requested):
- When the text has the horizontal overflow with gradient and applying focus. The focus needs to be on top of the gradient, not underneath it
Screen Shot 2020-11-11 at 12 20 09 PM

- I can not get the gradient fade style when using Safari. It looks good in Firefox and Chrome though.

laurenmrice (changes requested):
- The scroll fade gradient doesn't go all the way to the edge on either side of the code snippet area. It should be right next to the edges so the white text is still underneath the gradient.

Screen Shot 2020-11-12 at 11 56 49 AM


- I still can't get the fade gradient style to show up in Safari.

emyarod:
the fade effects will not appear in Safari due to a browser bug (related https://github.com/carbon-design-system/carbon/issues/6539)

laurenmrice (approved):
Looks wonderful! Thank you 🙌🏻

PR #7029, opened 2020-10-12
style(button): fix various button bugs

tw15egan:
Closes #7021 
Closes #7025 
Closes #7026 

Various bug fixes for style issues found in `Button`.

#### Changelog

**New**
- Added `Danger` variant to Playground
- Explicitly set `:active` styles for `Tertiary Button` (#7026)
- Set icon-only tooltip to be hidden if `disabled` class is present (#7021)

**Changed**

- Moved `ButtonSet` styles to _after_ the `@include button-base;` mixin call that is applied to `bx--btn`. This was causing styling issues as things were not properly cascading down. (#7021)
- Replaced all `$disabled` tokens with `$disabled-03` (#7021)
- Changed box shadow inset from `$ui-02` to `$ui-background` (#7025)

**Removed**

- Redundant `disabled` style block that was causing wrong values to be used

#### Testing / Reviewing

- Check all variants of Button and ensure nothing has changed
- Enable `disabled` knob and ensure `disabled` styles are correct across all variants

Issue #7021, opened 2020-10-12
[button] disabled states have bugs


laurenmrice:
In the `playground story` there are a couple different kinds of bugs happening in the dark themes with the disabled state when you hover over the buttons.

This is happening for the variants: primary button, secondary button, icon button and set of buttons. I am assuming also for the `danger button` but the danger button is not currently in the story (but should be).

- Hovering over individual buttons (`primary`, `secondary`, `icon button`, check `danger`) causes the text/icon color to turn to a different shade of gray. The text/icon color should be `disabled-03`.
- Hovering over the divider line in the `set of buttons` also turns the divider color to a different shade of gray. This should also be `disabled-03`. (this is happening in all themes)
- The icon button gets a tooltip on hover for all variants. The tooltip should be not be coming up on hover when its disabled. The icon button should also get a disabled icon cursor change.

Example:
![button disabled](https://user-images.githubusercontent.com/43969356/95775106-c03cd180-0c8f-11eb-8212-3ab44307310c.gif)

PR #7253, opened 2020-11-09
fix: resolve icon-only button tooltip/label regression

### LINKED ISSUE #7233: Pagination buttons have no label element
go here: https://react.carbondesignsystem.com/?path=/story/pagination--multiple-pagination-components

Pagination "previous" and "next" buttons have no label element. This seems like regression as we didn't have that before.

image

emyarod:
Closes #7233

related #7021
related #7029

This PR resolves an icon only button tooltip regression which affected all consumers of tooltip styles including Pagination

#### Changelog

**Changed**

- hide icon only button tooltips when button is disabled (fix for #7021)

**Removed**

- icon only button specific styles from global tooltip mixins (fix regression from #7029)

#### Testing / Reviewing

Confirm that disabled icon-only buttons do not display tooltips on hover and confirm that pagination icon only buttons do not throw a11y errors for missing labels


laurenmrice (approved):
Looks good to me in all themes and firefox/chrome/safari 👍🏻

dakahn (approved):

tw15egan (approved):

PR #7254, opened 2020-11-09
fix(modal): fix wrapFocus issue with disabled primary button

### LINKED ISSUE #7249: Keyboard focus escapes Modal when OK button is disabled

- [x] `carbon-components`

## Detailed description

When the primary button in a Modal is disabled, shift-tab will move the focus outside the dialog (i.e., the focus trap fails).
The issues appears to be that wrapFocus looks for the a focusable element in the dialog with a selector that includes `*[tabindex]:not([tabindex='-1'])`. And the Modal sets the primary button to `disabled="true"` but leaves `tabIndex=0`. That makes it look focusable to the selector, but the browser won't allow it (since it's disabled). So the call to .focus() fails, leaving focus on a sentinel element. The next shift-tab is then focused somewhere on the page outside the modal.

See
https://codesandbox.io/s/gifted-panini-ow3pq

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7249

Fixes an issue where `wrapFocus` incorrectly views a disabled primary button as focusable. Uses the existing `primaryButtonDisabled` to change the `tabIndex` based on the disabled state.

#### Changelog

**New**

- `primaryButtonDisabled` knob in storybook

**Changed**

- `tabIndex` on the primary modal button changes based on the `disabled` state

#### Testing / Reviewing

Enable the `primaryButtonDisabled` knob and try to `Shift+Tab` through the modal


emyarod (commented):
I think we should just update the selector query string for the focus trap instead, rather than just the tabindex for modal primary buttons

tw15egan:
@emyarod makes sense, updated 👍

emyarod (approved):
looks good to me

aledavila (approved):

PR #7262, opened 2020-11-10
feat(tag): introducing custom icon

IgnacioBecerra:
Related: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/3669

This is a contribution for the `Tag` component in which allows the adopter to include any icon from `@carbon/icons` before the content text. This feature had been requested as part of the Expressive theme in Carbon for IBM.com. 

The new `renderIcon` prop requires an icon component as an argument, which will render it accordingly within the `Tag` component. `hasCustomIcon` must also be passed into the `Tag` component to activate the custom icon functionality.

Here are some examples:

Using `renderIcon: Tag16`
Screen Shot 2020-11-11 at 10 00 00 AM

Using `renderIcon: Compass16`
Screen Shot 2020-11-11 at 10 00 40 AM

#### Changelog

**New**

- added `renderIcon` prop to allow for a custom icon to be added before the content text

#### Testing / Reviewing

Enable/disable the new prop, experiment with various icon names to ensure they render properly.


jeanservaas:
Hey @IgnacioBecerra thanks for the contribution!

I was just looking at Shixie's spec I think I would tweak the spacing a bit 

She's actually modified our tag for digital (.com team) and is using: 

- 14/20pt type
- 32px tag height (6px of padding top and bottom)
- Gray 80 (your type is too light for contrast accessibility right now)
**- I changed her spec to make the space between the icon and the type 4px (instead of 8px)**

image

If you wanted a spec to add an icon to the Carbon tag, for product teams (which it looks like you're doing), here it is:

- [x] 12/16pt type
- [x] 24px tag height (6px of padding top and bottom)
- [ ] Gray 80 (your type is too light for contrast accessibility right now)
- [ ] Make the space between the icon and the type 4px (instead of 8px)

image

packages/react/src/components/Tag/Tag.js
```diff
@@ -9,7 +9,8 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
-import { Close16 } from '@carbon/icons-react';
+import * as Icons from '@carbon/icons-react';
```
tw15egan: I think it may be best to render icons using similar to the `renderIcon` prop found in other components

See Button and Tooltip for examples: 
https://github.com/carbon-design-system/carbon/blob/0049b31dfdfb54774eb49ab17f0df1b900cfa963/packages/react/src/components/Button/Button.js

https://github.com/carbon-design-system/carbon/blob/225f212a8f3e2bd86d4a23f374f205b25bf5b9f0/packages/react/src/components/Tooltip/Tooltip.js

mjabbink:
Icon cropping in the left corner


96F8F21B-C620-4A13-85DD-EC5A8017C8AB

IgnacioBecerra:
@jeanservaas Thank you for the specs! I modified the text color values to type 80 in `tokens.scss` to conform with contrast accessibility guidelines. 

@tw15egan Thank you for the examples! Using the `renderIcon` prop, an icon component will be passed as a prop into the `Tag` component instead of using strings. The Storybook knob for custom icons will be gone however, although it's not strictly necessary as the other components lack it as well.

@mjabbink Thanks for letting me know about that! I wasn't able to replicate the cropping, but I removed the css property that I suspect was the cause of the issue.

tw15egan:
@IgnacioBecerra Awesome! If you do want to allow a user to see that they can change the icon to whatever they want, you can add in a knob, like how we do for the button story. Here's [the link](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Button/Button-story.js#L18-L57) if you're interested

packages/react/src/components/Tag/Tag.js
```diff
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
-import { Close16 } from '@carbon/icons-react';
+import { Close16 }from '@carbon/icons-react';
```
tw15egan:
```suggestion
import { Close16 } from '@carbon/icons-react';
```

IgnacioBecerra:
@tw15egan Thanks! Added three examples in the Storybook using the `select` knob.

packages/react/src/components/Tag/Tag.js
```diff
@@ -34,6 +34,8 @@ const Tag = ({
   id,
   type,
   filter,
+  hasCustomIcon,
+  renderIcon: CustomIconElement,
```
tw15egan, line 37: Would we expect a user to always set their own custom icon? If so, do we just need the `renderIcon` prop, and base the rest of the logic on the presence of this prop?

IgnacioBecerra: Yes, users are expected to set their own custom icon if they choose to use this `Tag` variation, there's no "default" icon. Users would also need to include the `hasCustomIcon` in their `Tag` component, but yeah, I think needing two variables to activate this behavior may be redundant. 

Would you suggest changing it to only needing `renderIcon`/getting rid of `hasCustomIcon`?

tw15egan: Yeah, that's what I was thinking! That way by default there would be no icon and if a user wants to use one they would be expected to provide their own icon. Then we can just check if `renderIcon` exists

IgnacioBecerra: Got it, pushed the change, thanks!

packages/react/src/components/Tag/Tag.js
```diff
@@ -117,6 +139,12 @@ Tag.propTypes = {
    */
   onClose: PropTypes.func,
 
+    /**
+   * Optional prop to allow overriding the icon rendering.
```
tw15egan:
```suggestion
   * Optional prop to render a custom icon.
```

packages/react/src/components/Tag/Tag-test.js
```diff
@@ -64,6 +64,11 @@ describe('Tag', () => {
     expect(tag.text()).toEqual('New Version!');
   });
 
+  it('should allow for a custom icon', () => {
+    const tag = shallow(<Tag type="red" customIcon='Add16'>This is a tag</Tag>)
```
tw15egan:
```suggestion
    const tag = shallow(<Tag type="red" renderIcon={Add16}>This is a tag</Tag>)
```

tw15egan: Will also need to import the icon at the top of the test file

```
import { Add16 } from '@carbon/icons-react';
```

tw15egan (approved):
LGTM, thanks for taking the time to work on this! 👍 ✅ 🚀

packages/components/src/components/tag/_tag.scss
```diff
@@ -172,6 +172,22 @@
     }
   }
 
+  .#{$prefix}--tag__custom-icon {
+    flex-shrink: 0;
+    width: rem(16px);
+    height: rem(16px);
+    margin-right: rem(4px);
```
emyarod:
```suggestion
    margin-right: $spacing-02;
```

packages/react/src/components/Tag/Tag-story.js
```diff
@@ -36,6 +49,15 @@ const props = {
...
     };
   },
+  icon() {
+    let iconToUse = iconMap[select('Icon (icon)', icons, 'Tag16')];
+    return {
+      ...this.regular(),
+      onClick: action('onClick'),
+      onClose: action('onClose'),
+      renderIcon: iconToUse,
+    };
```
emyarod: I think we can just inline this since we aren't checking for falsiness or the svgData property

```suggestion
    return {
      ...this.regular(),
      onClick: action('onClick'),
      onClose: action('onClose'),
      renderIcon: iconMap[select('Icon (icon)', icons, 'Tag16')],
    };
```

packages/react/src/components/Tag/Tag.js
```diff
@@ -76,6 +78,25 @@ const Tag = ({
         <Close16 />
       </button>
     </div>
+  ) : CustomIconElement ? (
```
emyarod: can we refactor this to avoid nesting ternaries?

IgnacioBecerra:
@emyarod Thanks. for the suggestions! I refactored using if/else if for better readability, let me know if you have any thoughts about it!

packages/react/src/components/Tag/Tag.js
```diff
@@ -50,40 +51,67 @@ const Tag = ({
...
+          className={`${prefix}--tag__close-icon`}
+          onClick={handleClose}
+          disabled={disabled}
+          aria-labelledby={tagId}
+          title={title}>
+          <Close16 />
+        </button>
+      </div>
+    );
+  } else if(CustomIconElement) {
```
emyarod:
```suggestion
  }
  if (CustomIconElement) {
```

since we `return` in the block above, this following `else-if` block can just stand alone as an `if` block, if that makes sense

packages/react/src/components/Tag/Tag.js
```diff
@@ -50,40 +51,67 @@ const Tag = ({
...
+        </button>
+        <span
+          className={`${prefix}--tag__label`}
+          title={typeof children === 'string' ? children : null}>
+          {children !== null && children !== undefined ? children : TYPES[type]}
+        </span>
+      </div>
+    );
+  } else {
+    return (
```
emyarod:
```suggestion
  }
  return (
```

packages/react/src/components/Tag/Tag.js
```diff
@@ -50,40 +51,67 @@ const Tag = ({
...
-  ) : (
-    <span
-      className={tagClasses}
-      title={typeof children === 'string' ? children : null}
-      {...other}>
-      {children !== null && children !== undefined ? children : TYPES[type]}
-    </span>
-  );
+    );
+  }
```
emyarod:
```suggestion
```

removed because the final `return` statement does not need to be in an `else` block

emyarod (commented):
it looks like you are not using an autoformatter in your editor, can you try running `yarn format` in the repository root?

packages/react/src/components/Tag/Tag.js
```diff
@@ -50,40 +51,67 @@ const Tag = ({
...
+          title !== undefined
+            ? `${title} ${children}`
+            : `Clear filter ${children}`
+        }
+        id={tagId}
+        {...other}>
+        <button
+          className={`${prefix}--tag__custom-icon`}>
+          <CustomIconElement />
+        </button>
```
emyarod:
```suggestion
        <button
          type="button"
          className={`${prefix}--tag__custom-icon`}>
          <CustomIconElement />
        </button>
```

I believe this also needs a `type` attribute

jeanservaas (approved):
lgtm

packages/react/src/components/Tag/Tag.js
IgnacioBecerra: Hmm, thinking about it, the custom icon isn't a button in this context (it can't be clicked). Closing the `CustomIconElement` in a `<div>` would provide the same effect, so I'll be changing it to that instead.

johnbister (approved):

dakahn (approved):

packages/react/src/components/Tag/Tag.js
```diff
@@ -51,32 +52,60 @@ const Tag = ({
...
+      <div
+        className={tagClasses}
+        aria-label={
+          title !== undefined
+            ? `${title} ${children}`
+            : `Clear filter ${children}`
+        }
+        id={tagId}
+        {...other}>
+        <button className={`${prefix}--tag__custom-icon`}>
```
joshblack: Would this button need the same props as the other one?

On a related note, would it make sense to try and re-use the markup and conditionally render the custom icon?

packages/react/src/components/Tag/Tag.js
```diff
@@ -117,6 +146,12 @@ Tag.propTypes = {
    */
   onClose: PropTypes.func,
 
+  /**
+   * Optional prop to render a custom icon.
+   * Can be a React component class
+   */
+  renderIcon: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
```
joshblack, line 132: Would there be a reasonable way to just allow teams to render the icon by passing it as a child to the component? Wasn't sure but wanted to check 👀

IgnacioBecerra: It's possible to do what you described, although it would lack the styles required from the specs, spacing wise.

packages/react/src/components/Tag/Tag.js
tw15egan: Does this even need to be a button since this is just a container for the icon and isn't clickable?

IgnacioBecerra: Seems like I didn't push the change yesterday, but the custom icon shouldn't behave like a button. I had changed it to a `div` instead, but I'm currently looking into the conditional rendering instead of having three blocks of code in there

packages/react/src/components/Tag/Tag.js
```diff
@@ -77,12 +79,18 @@ const Tag = ({
...
-    </span>
+    <div className={tagClasses} id={tagId} {...other}>
+      {CustomIconElement ? (
+        <div className={`${prefix}--tag__custom-icon`}>
+          <CustomIconElement />
+        </div>
+      ) : (
+        ''
+      )}
+      <span title={typeof children === 'string' ? children : null} {...other}>
```
joshblack: Noticed that `{...other}` was being spread to both the outer `<div>` and the `<span>` here, I think we'll need to pick one and stick with it otherwise we'll end up in weird situations like both are receiving an `onClick` or `data-testid` or something 🤔 

I think just keeping `{...other}` on the outermost node would work, but curious what you think!

joshblack (changes requested):

packages/react/src/components/Tag/Tag.js
IgnacioBecerra: I agree, keeping it on the outermost node works better. Good catch, thanks @joshblack!

joshblack (approved):
Thanks for doing this!

PR #7268, opened 2020-11-11
fix: high-contrast tag when used as a filter should use inverse focus

lee-chase:
The high-contrast tag when used has a filter uses the standard $focus whereas the rest of Carbon uses the inverse focus on dark backgrounds.

This PR changes that focus color.

#### Changelog

m. packages/components/src/components/tag/_tag.scss


tw15egan (approved):
LGTM, thanks for the fix Lee! 👍 ✅

aledavila (approved):

PR #7094, opened 2020-10-20
feat(pictograms): update existing and add net new pictograms

emyarod:
ref #6923

This PR replaces the existing pictograms with newly exported assets and updates the names and aliases for a significant number of pictograms (the most recent asset revisions were made on Oct 20). For existing users this may be a breaking change as some name changes will mean the current asset names will not map to the same assets going forward. Several deprecations have been added to the list, and the current list of deprecated assets can be resolved in a future PR

#### Changelog

**New**

- various net new pictograms

**Changed**

- renamed assets which would have had naming collisions due to metadata changes. will be breaking changes for users of the current asset as the artwork that is mapped to the current name will be changing
- deprecated several assets

**Removed**

- all assets that would have caused name collisions

#### Testing / Reviewing

Ensure the updated pictogram metadata matches the requirements. This will require a manual process as there is no clear naming convention for the assets

PR #7272, opened 2020-11-11
chore: deprecate pictograms

### LINKED ISSUE #6923: Determine changes in pictogram assets
The brand team has delivered a set of new pictogram assets that have been redrawn to support `fill` instead of `stroke`. We need to determine the scope of these changes in order to verify the following:

- What pictogram assets have changed? (For example, what has been renamed)
- What pictogram assets have been removed? What are they to be replaced with? (if anything)
- What pictogram assets are staying the same?

[Pictogram_Breaking_Change.zip](https://github.com/carbon-design-system/carbon/files/5293550/Pictogram_Breaking_Change.zip)

[Master_naming_update_091020.xlsx](https://github.com/carbon-design-system/carbon/files/5293566/Master_naming_update_091020.xlsx)

Using this information, we can coordinate a migration path for the affected assets when we release our breaking change.

emyarod:
Closes #6923

This PR removes all of the pictograms which were marked for deprecation

#### Changelog

**Changed**

- pictogram metadata

**Removed**

- deprecated pictograms


joshblack:
@emyarod how are we anticipating folks using the deprecation warnings to move to the correct asset?

It seemed like in the PR that was merged that we added new assets and the deprecation warnings, but I'm not sure when those deprecation warnings will get seen by an end-user.

Is the hope that what is in master will get published and this change gets merged in? Or is the hope to merge this in and do something like a migration guide? Just curious 👀

emyarod:
we assumed since the pictogram updates are already considered breaking changes, the aliases in the new/updated pictogram metadata would point users to the right assets. Should we just be deprecating pictograms that weren't touched in #7094? The migration process is not really clear to us especially since #7094 is an irregular update

joshblack:
@emyarod I think removing the deprecated assets in the major change makes a lot of sense, I was just wondering how we were communicating these changes to users and helping them migrate from old to new. It seems like it is unrealistic for the deprecation messages to be included in a v10 release because the new assets wouldn't be available. On the other hand, including the deprecated assets with the message in v11 would just mean holding on to them for longer.

In this case, it could just be that there is a migration guide needed to help folks know what the changes are to help them update. We could also do a codemod as well that uses that list to automatically update codebases that use these assets.

Not sure what the right answer is here, just am trying to figure out what the migration experience will look like for teams.

emyarod:
I'm not familiar with the process when it comes to asset updates like this as typically @dudley-ibm handles these IIRC. I'm also not sure what the right answer is but I think it makes sense to involve brand here and how they want to communicate these changes

andreancardona:
The migration guide for Pictograms has been added to this PR 🎉

docs/migration/10.x-pictograms.md
```diff
@@ -0,0 +1,175 @@
+# Pictograms
```
joshblack: Would this be 11.x-pictograms.md? 👀

joshblack (commented):
Looking great! Should we have a guide for pictograms-react as well with the name changes?

docs/migration/10.x-pictograms.md
andreancardona: yes looks like @emyarod updated to `11.x-pictograms.md`!

andreancardona: @joshblack regarding a migration guide for Pictograms React - it looks like didn't do that for icons? 👀

joshblack: @andreancardona IIRC the icons migration guide was specifically for `carbon-icons`. I think `@carbon/icons-react` was published with v10 and so was using the new names by default 👍

mjabbink:
Looking good from what I can tell. I only see one pictogram missing (Folder) from this preview
https://deploy-preview-7272--carbon-elements.netlify.app/pictograms/examples/preview/

docs/migration/10.x-pictograms.md
andreancardona: done!

emyarod:
the file is still present in the repo but this may just be a bug with the pictogram preview

andreancardona:
@joshblack just added the migration guide for `Pictograms-React`

joshblack:
I believe the folder asset in the preview is using stroke still which is why it is not displaying 🔍

emyarod:
cc @dudley-ibm do you have an updated `Folder` pictogram that does not use stroke? The asset we have on hand still uses stroke

dudley-ibm:
[folder.svg.zip](https://github.com/carbon-design-system/carbon/files/5569660/folder.svg.zip)
Sent it in Slack, but here it is also.

andreancardona:
@joshblack @mjabbink `Folder.svg` has been updated

Thankssss @emyarod 🥳

emyarod:
the asset has been updated and is now visible in the pictograms preview

joshblack (approved):
Looks great! For the commit, would it make sense to do something like:

```git
feat(pictograms): remove deprecated assets

BREAKING CHANGE: deprecated assets have been removed
```

For conventional commits?

dudley-ibm:
approved

emyarod:
@joshblack makes sense, how would we implement that when the commits will be squashed when we merge?

joshblack:
@emyarod I think you can edit the merge commit when merging to match that format 👀

mjabbink:
@conradennis we only need one Prague right? Also only need one Weather

Screen Shot 2020-11-19 at 5 03 51 PM


Screen Shot 2020-11-19 at 5 06 23 PM

emyarod:
@joshblack got it, will make that change when merging. as for the asset comments, we will need to wait for them to be resolved before merging this PR (and merging into v10.25)

conradennis:
@mjabbink yup correct, the first one should be removed.

andreancardona:
> @mjabbink yup correct, the first one should be removed.

@conradennis the first one should be removed for `prague` and `weather`? Please clarify 👀 also please specify which code names.

dudley-ibm:
The artwork for the `weather` pictogram initially submitted in the Industry category was similar but slightly different than the `weather` pictogram submitted afterward in the Weather category. That's why I added the 01 & 02, but they're definitely close enough that we only need one.

dudley-ibm:
@emyarod @andreancardona it looks like I did not communicate that `prague.svg` should be deprecated in favor of `prague--charles-bridge-tower.svg`. Can that be updated before this merges tomorrow?

emyarod:
`prague.svg` has been removed

mjabbink:
👍

dudley-ibm:
Since `weather--01.svg` and `weather--02.svg` are so similar in artwork, let's kill the one in this PR slated for the Industry category (weather--01.svg) and only update the svg that appears in the Weather category. It will now just be called `weather.svg` again, and that file is attached.

[weather.svg.zip](https://github.com/carbon-design-system/carbon/files/5574447/weather.svg.zip)

emyarod:
`weather--01` and `weather--02` have been replaced by `weather`

andreancardona (approved):

mjabbink:
Go to the glass! Woohoo!

Issue #7088, opened 2020-10-20
[OverflowMenuItem] itemText cannot put Tooltip component


dawang021:
## What package(s) are you using?
```
"carbon-components": "10.22.0",
"carbon-components-react": "7.22.0",
```

- [x] `carbon-components`
- [x] `carbon-components-react`

## Detailed description

> Describe in detail the issue you're having.

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the Carbon Design System are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Use OverflowMenu, OverflowMenuItem component in DataTable TableCell.
2.  The whole OverflowMenu cannot open anymore.

Sandbox reproduce: https://codesandbox.io/s/overflow-menu-with-tooltip-8iykr?file=/src/index.js

## Additional information

- Screenshots or code
- Notes

tw15egan:
@dawang021 it is really helpful if you fill out the issue template so we can help you get sorted out properly. Do you have a link to a reproducible example on CodeSandbox? I am unable to reproduce the problem on our end: https://react.carbondesignsystem.com/?path=/story/datatable--with-overflow-menu

dawang021:
@tw15egan please take a look at this demo in sandbox: https://codesandbox.io/s/overflow-menu-with-tooltip-8iykr?file=/src/index.js

tw15egan:
I don't believe `Tooltip` is supported inside of an overflow menu. Overflow menu actions should be short, concise actions that do not need additional information. You could place the tooltip outside of the menu if more information is needed.

Please see this link for more info: https://www.carbondesignsystem.com/components/overflow-menu/usage#formatting

dawang021:
@tw15egan But it works in lower version https://codesandbox.io/s/overflow-menu-with-tooltip-working-well-hsdwo?file=/src/index.js.
![image](https://user-images.githubusercontent.com/42695523/96832429-4e277200-1471-11eb-8044-b4011ca93fc2.png)


The reason why I want to add Tooltip in OverflowMenuItem is OverflowMenuItem can be disabled, so I want to let users know why this OverflowMenuItem is disabled.

tw15egan:
[...]

PR #7277, opened 2020-11-11
feat(OverflowMenuItem): support `title` prop

emyarod:
Related #7088

This PR allows users to set a `title` prop in the event that the tooltip text should be different from the `itemText`

#### Changelog

**New**

- `title` attribute support in OverflowMenuItem

**Changed**

- update OverflowMenu docs to show both usages of `title` (when a custom value is provided/not provided)

#### Testing / Reviewing

Confirm that setting a `title` together with the existing `requireTitle` functions properly


tw15egan (approved):
LGTM 👍 ✅ 🚀

dakahn (approved):

PR #7290, opened 2020-11-13
chore(cli): update package.json repository config

tw15egan:
Noticed after https://github.com/carbon-design-system/carbon/commit/05649fbc7b920b4df3888a47b17dd29f9fb84a2b that the changes were overwritten by our sync tool. This changes the CLI tool to add the correct monorepo repository structure so relative links will work on NPM

#### Changelog

**Changed**

- Updates the CLI tool to render the `repository` field as an object so monorepo relative paths work correctly

**Removed**

- unneeded variable

#### Testing / Reviewing

@joshblack wasn't sure how to test this, let me know if there is something I can run to verify 👍


packages/cli/src/commands/sync/package.js
```diff
@@ -73,7 +70,11 @@ function run({ packagePaths }) {
   return Promise.all(
     packagePaths.map(
       async ({ packageJsonPath, packageJson, packageFolder }) => {
-        packageJson.repository = `${REPO_URL_BASE}/${packageFolder}`;
+        packageJson.repository = {
+          type: 'git',
+          url: 'https://github.com/carbon-design-system/carbon.git',
+          directory: `${packageFolder}`,
```
joshblack:
```suggestion
          directory: packageFolder,
```

joshblack (approved):
Looks great! Just had one style comment but seems perfect regardless

joshblack:
@tw15egan to test it locally you could run at the top-level:

```bash
yarn carbon-cli sync package
```

And it should work! Also totally feel free to commit those changes too if you want.

emyarod (approved):
looks good to me

PR #7295, opened 2020-11-16
[RFC] feat(project): add utility-css package

joshblack:
This is a proof-of-concept PR for Utility CSS, popularized by [Tachyons](https://tachyons.io/) and now [Tailwind](https://tailwindcss.com/).

It creates a package located at `package/utility-css` that is authored using Sass Modules and provides a plugin system based off of Tailwind's model. In particular, this model uses the concept of **plugins** along with **variants** to generate utility classes that target (typically) one property and one value based on our design system.

For example, we have a plugin for `margin`. This plugin will emit rules based on our spacing scale, creating classes like:

```css
.m-5 { margin: spacing(5); }
.mb-5 { margin-bottom: spacing(5); }
.ml-5 { margin-left: spacing(5); }
```

_Note: `spacing(5)` would resovle to `16px` in our spacing scale_

**Variants** allow plugin authors to emit styles built on top of the default rules. These can be based on pseudo states like hover or focus, or can emit per-breakpoint.

For example, we may have `background-color` classes based off of our color tokens:

```css
.bg-ui-01 {
  background-color: $ui-01;
}
```

We can use variants to specify that this selector should also be emitted with support for `hover`, `focus`, and `responsive` breakpoints:

```css
.focus:bg-ui-01:focus {
  background-color: $ui-01;
}

.hover:bg-ui-01:hover {
  background-color: $ui-01;
}

@include breakpoint(lg) {
  .lg:bg-ui-01 {
    background-color: $ui-01;
  }
}
```

This variant selectors can be used like `.focus:bg-ui-01`, `.hover:bg-ui-01`, `.lg:bg-ui-01`, etc.

---

This Pull Request is to get feedback from the team on if it'd be worth supporting this kind of approach, and also to try and create a home for teams requesting Utility CSS support from the system. What is currently being proposed in this PR can 100% still change, from naming to setup, but hopefully provides a good structure based on Tailwind if we wanted to continue building it. 

Curious what you all think!

#### Changelog

**New**

- Add utility-css package

**Changed**

**Removed**

#### Testing / Reviewing

- There is an example app located in this PR where you can play around with classes! Currently it is not built out completely for all the plugins, in the interim you can use Tailwind for docs 😅


sstrubberg:
Love this exploration!

metonym:
This is a great idea. It would greatly improve the DX for those that prefer using pre-compiled CSS and removing unused selectors at the build step (e.g., the popular TailwindCSS + PurgeCSS combo).

Would this package also ship JS tokens?

For example:

```js
const mb5 = `margin-bottom: 1rem`;
const mb5_var = `margin-bottom: var(--cds-spacing-05)`;
```

EDIT: To add some more context for my use case, I'm envisioning a Svelte preprocessor that could rewrite attributes as inline styles:

**Before**

```svelte
<Grid>
  <Row lb-7>
     ...
  </Row>
</Grid>
```

**After**

```svelte
<Grid>
  <Row style="margin-bottom: var(--cds-layout-07)">
     ...
  </Row>
</Grid>
```

DusanMilko:
@joshblack 

I prefer more consistency in the declaration names but I see you are trying to follow an established pattern from other libraries.

Would be good to be able to specify which utility classes get added to project as for my project I only need the color utility classes(bg, text, fill, border-color) and would not want to/need to pull in all the other classes or margin, padding, etc.

I see you did bg, text and fill with theme vars, can you also add **border-color**?

Maybe section these into separate optional imports that get enabled by scss vars being set to true.

**Color utility:** 
bg, text, fill, border-color

**Spacing utility:**
margin, padding

etc...

moczolaszlo:
So, what is the plan with this? I think we in the Watson media use this package, in some scenarios it's could be very useful.

joshblack:
Hey @moczolaszlo 👋  Right now it seems like we have two options:

- Try and move forward with this RFC, starting with spacing and colors at first and then expanding if folks want more support
- Try and integrate with something like Tailwind and provide a preset that teams can use to configure Tailwind with values from Carbon

Curious what folks on this thread think about either approach! 👀

metonym:
I'd prefer a dedicated `@carbon` package

bogy0:
hello @joshblack thanks for this RFC! This is something that I missing from carbon for a while.

Two questions:
> * Try and move forward with this RFC, starting with spacing and colours at first and then expanding if folks want more support

It means that we can import a `@carbon/utility-spacing` and `@carbon/utility-color` package if we want, and then we can use utility class like CSS class names in our HTML code?

> * Try and integrate with something like Tailwind and provide a preset that teams can use to configure Tailwind with values from Carbon

By preset you mean this kind of tailwind preset?
https://tailwindcss.com/docs/presets 

I am not familiar with the second approach, but for me, the first one sounds better, and this can be a solution which can be done small incremental steps I think. 👍

DusanMilko:
@joshblack also prefer a carbon solution. I think we can take inspiration from tailwind but a proprietary carbon solution I feel is better

wkeese:
I read Tailwind's [CSS manifesto](https://adamwathan.me/css-utility-classes-and-separation-of-concerns/).  It made a lot of sense to me until "Phase 4", which seems (to me) to go overboard in the other direction.  I liked "Phase 3" a lot though.

Similarly, for this proof-of-concept, I kind-of cringe at the idea of classes like `bg-ui-01`.  Are there no higher-level CSS classes like `main-background-color` and `alternate-background-color`?   Or the "content agnostic" CSS classes from Phase 3 like `card` which would imply a background color in addition to certain padding etc.?

joshblack:
Thanks everyone for weighing in! So sorry about the delay here in catching up with the comments on this PR.

Judging by everyone's opinions on this PR, it seems like going down the Tailwind route isn't preferred and it might be helpful to narrow down on what a Sass-based API might look like from a Carbon package (potentially `@carbon/utility-css`)

For those talking about the "higher level" classes, unfortunately part of this proposal comes from the requests of teams to have tailwind-esque/tachyons atomic CSS which would include those selectors. I do think it brings up a great point in the design goals for the package, namely:

- What are sensible defaults we could provide (along with strategies for mitigating cost of CSS that isn't used)
- How could we provide configuration for teams to easily turn off certain groups
- How could we provide configuration for teams to add rules that their team needs

One idea that came up when writing this comment was to mirror what Bootstrap is doing with their utilities API: https://getbootstrap.com/docs/5.0/utilities/api/ and I'm curious what people on this thread think of providing this sort of configuration (similar to what is proposed) that will allow us to provide defaults, enable folks to rename classes to their preference, as well as extend/modify with rules that their teams require.

joshblack:
For everyone interested in this PR, some of this work will spin off into a preset for Tailwind: https://github.com/carbon-design-system/tailwind-preset-carbon

Some of the other utilities (things like flex, type class helpers, etc) will find their way into the v11 release 👀 

Going to close this out for now, feel free to comment if you have any questions about this direction!

Issue #7266, opened 2020-11-10
Bidi support: Arabic numeric shaping is not supported in Carbon components


adir01:
In some Arabic countries, Arabic-Indic digits are in use instead of the regular Arabic-European numbers
More information can be found on https://w3-03.ibm.com/globalization/page/3651#digits

Numeric shaping property setting and processing are missing in Carbon components:

For instance:
Data with shaped digits - 

![image](https://user-images.githubusercontent.com/7374762/98743133-354b1600-23b8-11eb-9fba-af6187355a4f.png)

Examples for Carbon components:

1. Arabic numeric shaping in Time picker
(RTL style is applied on Carbon)

![image](https://user-images.githubusercontent.com/7374762/98742797-b5bd4700-23b7-11eb-8c1b-b9b388efb2cb.png)

2. Arabic numeric shaping in Slider

![image](https://user-images.githubusercontent.com/7374762/98742935-eac99980-23b7-11eb-812b-b8a4fff24123.png)

@mattrosno , @joshblack @wajnberg FYI

Issue #7265, opened 2020-11-10
Bidi support: Hijri calendar is not supported


adir01:
In some Arabic countries, usage in Hijri calendar is common, and even required according to law.
Hijri and Hebrew calendars are Lunar calendars, and therefore dates in these calendars are different from Gregorian used in European countries.

Examples:
1. Date Picker
Steps: 
Apply RTL style on Carbon
Set calendar = Hijri

Gregorian calendar is actually shown:

![image](https://user-images.githubusercontent.com/7374762/98740435-b3f18480-23b3-11eb-978e-f2ca62777e2f.png)

@mattrosno , @joshblack  FYI

tw15egan:
Seems like we are blocked on the flatpickr side of things, as they do not support this calendar:

https://github.com/flatpickr/flatpickr/pull/1526

https://github.com/carbon-design-system/carbon/issues/2585

adir01:
@wajnberg : I understand that the code on flatpickr/flatpickr#1526 was developed by you and I also understand that this code was written for Hijri calendar support in Watson Assistant which was completed. Can you please clarify the status of this PR on flatpickr ?

wajnberg:
Hello @adir01 

As mentioned in https://github.com/flatpickr/flatpickr/pull/1526#issuecomment-474847727 they want to refactor the Hijri code as a plugin but they suspect it would cause tremendous amount of work. I will try to follow up.

tay1orjones:
Maybe this could be part of https://github.com/carbon-design-system/carbon/discussions/13355

Issue #7264, opened 2020-11-10
BIDI - Support base text direction


adir01:
Base text direction is an integral part of the overall Bidi support and enables correct editing and readability of dynamic data for BIDI languages ( Arabic, Hebrew, Urdu)

Carbon code allows to specify following settings for the text direction, using textDir  property :
* auto
* ltr
* rtl
However most of Carbon components does not apply text direction processing expected to happen according to this property settings and as a result of that , actual text direction inherits the page direction (which is by default LTR and is RTL when RTL style is applied), independently on textDir property setting

Examples:
1. Base text direction is not applied on Rich text editor

Steps: 
Apply RTL style on Carbon
Set textDir property  = rtl
Text inside editor still has LTR text direction
![image](https://user-images.githubusercontent.com/7374762/98733769-b818a480-23a9-11eb-9009-6e3a35b99b36.png)

2. Base text direction is not applied on combo box
Steps: 
Apply RTL style on Carbon
Set textDir property  = auto
Text at combo has RTL text direction

![image](https://user-images.githubusercontent.com/7374762/98734333-7b997880-23aa-11eb-8244-bdf18f5352e1.png)

@mattrosno , @joshblack  FYI

adir01:
One could find more information about text direction on https://www.w3.org/International/articles/inline-bidi-markup/

wajnberg:
Just one precision.
The property `textdir` exists only in https://github.ibm.com/BusinessAnalytics/ba-ui-carbon-toolkit which is an IBM repo
It doesn't exist for this repo `carbon-design-system`
We would need to add this property here.

adir01:
@wajnberg 
1) PR that you submitted is not expected to cause closure of this issue. As we agreed it will be only the prototype which is expected to help Carbon developers provide similar feature support in other Carbon components.
2) Could you also please go through other components and ensure that similar solution is possible there ?

@joshblack : 
1) could you please estimate when all other Carbon components can be supported
2) will it be possible to supply testing framework for this feature support ?

adir01:
@joshblack I just noticed there is missing description of expected rendering of Hebrew text 
In Example #1 it would be 

![image](https://user-images.githubusercontent.com/7374762/107651483-3a09b980-6c88-11eb-8968-25f224759532.png)

Can be easily recreated in Notepad :

1) Paste following text
2) Right Mouse click and select Right-to-Left reading order

adir01:
[...]

PR #7297, opened 2020-11-16
chore(rtl): add RTL plugin to storybook

tw15egan:
Rel https://github.com/carbon-design-system/carbon/issues/7264 #7265 #7266 

Adds in an RTL plugin to showcase the components in RTL mode. WIll help with future RTL enhancements and bug fixes

#### Changelog

**New**

- `storybook-addon-rtl` plugin added to storybook

#### Testing / Reviewing

Ensure the RTL toggle works properly


joshblack (approved):

aledavila (approved):

tw15egan:
Not sure exactly what is the intended behavior, but here are the two options. With the RTL plugin, we are able to toggle it on and off. the `useRtl` var flips the styles entirely and there doesn't seem to be a way to toggle it without recompiling 

setting `useRtl = true` in `webpack.config.js`:
Screen Shot 2020-11-30 at 4 29 00 PM

Using the `RTL` plugin:
Screen Shot 2020-11-30 at 4 25 02 PM

(Default)
Screen Shot 2020-11-30 at 4 27 34 PM

tw15egan:
This plugin doesn't seem like it's going to do what we want it to do. running storybook with the command `CARBON_REACT_STORYBOOK_USE_RTL=true yarn storybook` properly enables RTL support, so I think we just need more guidance around how to compile styles for RTL 

Screen Shot 2020-12-01 at 11 00 16 AM

PR #5304, opened 2020-02-07
fix(tooltip): ensure tooltip is not hidden in storybook

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/5286

Centers `Tooltip` so that the pop up is not hidden under the side nav

Also adjusts Tooltip positioning logic for `left` and `right` directions

#### Changelog

**New**

- Storybook specific position styles

**Changed**

- Modified positioning for `left` and `right` variations

#### Testing / Reviewing

Ensure the tooltip no longer renders under the side nav

PR #7309, opened 2020-11-18
feat: add align to tooltip

### LINKED ISSUE #3837: Add alignment to the Carbon's tooltip component
### The problem
The tooltip overflows a parent modal when the tooltip label is near the edge of that modal.
This restricts the designer to place the label away from the edges of a parent container.

image

image

### The solution
The tooltip component should accept a default prop to align its tooltip to the start / centre / or end of the tooltip label (just like the tooltipDefinition component).
The caret should always point to the centre of the tooltip label.

image

image

### Additional information
This can be a partly fix for the above scenario, because the other part of it is when the label is in the end of a line within a container (when the default alignment is set to ‘start’).

lee-chase:
Closes #3837 

C&CS would like to use Tooltip as a basic overflow for tag components. However, this is problematic because overflow typically happens near the right side of the page and there is no option to align the tooltip

See https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-tagset--tag-array

NOTE: Width of the container is adjusted on the controls tab. 

#### Changelog

M       packages/components/src/components/tooltip/_tooltip.scss
M       packages/react/src/components/Tooltip/Tooltip-story.js
M       packages/react/src/components/Tooltip/Tooltip.js
M.      packages/react/src/components/Tooltip/Tooltip-test.js

#### Testing / Reviewing

Ran it up in storybook.


joshblack (approved):

tw15egan (approved):
LGTM, thanks for adding this in! 👍 ✅

tw15egan:
One thing to note for v11, but `Button` uses `TooltipDirection` and `TooltipAlignment`. We'll probably want to conform to a consistent name, I don't really have a preference either way. 

I've made a note in the v11 Prop Audit as well.

laurenmrice (changes requested):
Can we have a 16px padding before the tooltip sharkfin? This will align vertically with the text and match the padding on the opposite side.
Artboard

lee-chase:
> Can we have a 16px padding before the tooltip sharkfin? This will align vertically with the text and match the padding on the opposite side.
> Artboard

Sure can @laurenmrice also adjusted left and right to align with the text.

image
image
image
image

laurenmrice (changes requested):
**Bottom and Top directions**
-start 
-end

These need to have a 16px space between the edge of the shark fin and the container. Right now they still are not aligning vertically with the text.



---------------------------------------------------------------------------------------------

**Left and Right directions**
-center

The shark fin is not showing up at the center of the container. There is not an equal amount of space above or below it right now.

lee-chase:
@laurenmrice  Addressed the position for start and end when above and blow.

The center-left and center-right caret positions I have not changed as part of this PR. I do not know the reason they were set 9px above centre on the left and 3px below on the right. It has been like this since changes in #5304, perhaps @tw15egan could shed light on the reason for the offset.

tw15egan:
@lee-chase that offset was to get them centered in that specific [PR](https://github.com/carbon-design-system/carbon/pull/5304#pullrequestreview-355454729), but if they aren't needed any more just axe em 👍

lee-chase:
Unfortunately, I am not familiar with the floating menu code.  Could that be addressed in a separate issue as this issue has not changed it?

laurenmrice:
I am fine with addressing the centering problem in a separate pr 👍🏻. The position for start and end looks good now, thank you!

laurenmrice (approved):

PR #7312, opened 2020-11-18
fix(NumberInput): restore `defaultValue` prop functionality in uncontrolled number input

### LINKED ISSUE #7299: NumberInput: uncontrolled NumperInput not respecting `defaultValue`

- [ ] `carbon-components`

## Detailed description

- Using `NumberInput` as an uncontrolled component with `defaultValue` does not set the `defaultValue` correctly, and throws a warning in the console about using both `value` and `defaultValue` simultaneously

- `NumberInput`

> see changed?

- expected it to display the `defaultValue` correctly
- expected it to not throw a warning in the console about having `value` and `defaultValue` set simultaneously

- chrome 86.0.4240.198

- `carbon-components-react@7.23.2`

> should be aware of?

- IBM App Connect

1. Use `NumberInput` with a `defaultValue`

>
> - Style and vanilla JS:
> - React:

- https://codesandbox.io/s/nifty-framework-pxzh7

## Additional information

image
- probably due to always setting value here: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/NumberInput/NumberInput.js#L325. i think this should only set value if it's a controlled component.

emyarod:
Closes #7299

This PR restores functionality to the `defaultValue` prop in uncontrolled number inputs and removes the React warning about setting both `defaultValue` and `value` on the input

#### Changelog

**New**

- uncontrolled number input `defaultValue` test

**Changed**

- avoid spreading `defaultValue` into input
- add `defaultValue` typecheck
- remove default internal state value in `gDSFP`

#### Testing / Reviewing

Confirm that setting a `defaultValue` on an uncontrolled number input works as expected and the correct value is displayed on render

### DIFF
--- a/packages/react/src/components/NumberInput/NumberInput-test.js
+++ b/packages/react/src/components/NumberInput/NumberInput-test.js
@@ -155,6 +155,24 @@ describe('NumberInput', () => {
           );
         const getNumberInput = (wrapper) => wrapper.find('input');
 
+        it('should correctly set defaultValue on uncontrolled input', () => {
+          const wrapper = mount(
+            <NumberInput
+              min={-1}
+              max={100}
+              defaultValue={10}
+              id="test"
+              label="Number Input"
+              className="extra-class"
+            />
+          );
+          const numberInput = getNumberInput(wrapper);
+          expect(wrapper.find('NumberInput').instance().state.value).toEqual(
+            10
+          );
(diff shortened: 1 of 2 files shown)


joshblack (approved):

andreancardona (approved):
looks good! @emyarod

PR #7319, opened 2020-11-19
fix(Tooltip): prevent focus hijacking on tooltip dismissal

### LINKED ISSUE #7260: Tooltips disappearing when clicking different tooltips

- [x] `carbon-components`

## Detailed description

When there are multiple tooltips and one tooltip is open, clicking on another tooltip causes the second tooltip to briefly appear and then disappear.

Yes, the `Tooltip` component.

> see changed?

I expect the second tooltip to stay open after clicking on it.

Google Chrome Version 86.0.4240.193

`carbon-components`: 10.23.2
`carbon-components-react`: 7.23.2

> should be aware of?

See this sandbox: https://codesandbox.io/s/tooltips-kpcbk?file=/src/index.js

1. Click on the tooltip next to **Text input (tooltip with no id) 1**. Notice that the tooltip stays visible.
2. Click on the tooltip next to **Text input (tooltip with no id) 2**. Notice that the tooltip appears for a brief moment and then disappears.

Screen Recording 2020-11-10 at 9 39 23 AM

This was discussed in this [Slack thread](https://ibm-analytics.slack.com/archives/C2K6RFJ1G/p1603316421150500), where it was suggested to add unique `id` props to each `Tooltip`. I tried this and it _did_ fix the disappearing tooltip issue (play around with the **Text input (tooltip with id)** tooltips to see this), but it causes accessibility issues where you can no longer tab between text inputs (from tooltip to text input and so forth). In the case where tooltips do not have `id` props, tabbing is fine.

Screen Recording 2020-11-10 at 9 40 41 AM

emyarod:
Closes #7260
ref #7165

This PR prevents the tooltip from hijacking keyboard focus after dismissing the tooltip. We check whether or not the tooltip contains focusable content, and if it doesn't then we can preserve the existing tab order. otherwise if it does contain focusable content, we will place focus back on the trigger button (like we do currently)

#### Changelog

**Changed**

- consolidate internal tooltip ID and prevent ID prop from overwriting internal ID value
- tooltip focus behavior on tooltip dismissal

**Removed**

- negative tabindex on tooltip content container

#### Testing / Reviewing

Confirm the reported focus issues are no longer present, and that there are no regressions with tooltip focus management

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.js
+++ b/packages/react/src/components/Tooltip/Tooltip.js
@@ -244,6 +244,7 @@ class Tooltip extends Component {
    * @private
    */
   _tooltipId =
+    this.props.id ||
     this.props.tooltipId ||
     `__carbon-tooltip_${Math.random().toString(36).substr(2)}`;
 
@@ -506,9 +507,9 @@ class Tooltip extends Component {
               this._tooltipEl = node;
             }}>
             <div
-              id={this._tooltipId}
               className={tooltipClasses}
               {...other}
+              id={this._tooltipId}
               data-floating-menu-direction={direction}
               onMouseOver={this.handleMouse}
               onMouseOut={this.handleMouse}
--- a/packages/react/src/components/Tooltip/Tooltip.js
+++ b/packages/react/src/components/Tooltip/Tooltip.js
@@ -21,6 +21,10 @@ import FloatingMenu, {
 import ClickListener from '../../internal/ClickListener';
 import mergeRefs from '../../tools/mergeRefs';
 import { keys, matches as keyDownMatch } from '../../internal/keyboard';
+import {
+  selectorFocusable,
+  selectorTabbable,
+} from '../../internal/keyboard/navigation';
 import isRequiredOneOf from '../../prop-types/isRequiredOneOf';
 import requiredIfValueExists from '../../prop-types/requiredIfValueExists';
 import { useControlledStateWithValue } from '../../internal/FeatureFlags';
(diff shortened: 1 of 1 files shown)


emyarod:
added a temporary Storybook test to make testing easier (can be removed before merge)

tw15egan (approved):
Thanks for adding the quick story for testing, LGTM 👍 ✅

dakahn (approved):

carmacleod:
There's odd behavior when tabbing (or shift+tabbing) to a tooltip trigger: every _second_ time the trigger gets focus, the tooltip doesn't open. This only happens with tooltips _without_ focusable content.

To see this (I saw this in Windows Chrome/Firefox and macOS Safari - probably happens everywhere):
1. Run the [focus management test](https://deploy-preview-7319--carbon-components-react.netlify.app/iframe.html?id=tooltip--focus-management-test).
2. Start typing the <kbd>tab</kbd> key.
3. The first ℹ️ trigger opens a tooltip with focusable content, so type <kbd>Esc</kbd> to close it.
4. Type <kbd>tab</kbd> again, and the next ℹ️ trigger opens a tooltip with NO focusable content, so as expected, typing <kbd>tab</kbd> moves to the text input because it's next in the tab order. So far, so good.
5. Keep typing <kbd>tab</kbd> (10 more times) until you have tabbed through all of the content on the page. Still good.
6. Now, go back to the very first ℹ️ trigger - the one with focusable content (either keep typing <kbd>tab</kbd> until you loop all the way around, or simply click it with the mouse). Type <kbd>Esc</kbd> to close.
7. Ok, here's where it gets weird: <kbd>tab</kbd> to the 2nd ℹ️ trigger. The tooltip doesn't open? Keep typing <kbd>tab</kbd> and note that none of the tooltips open this time. Loop around to the top and keep typing <kbd>tab</kbd>... now the tooltips open again. Same behavior (tooltips only open every second time) when typing <kbd>shift + tab</kbd>.

I think it's important to understand what's going on here. :)

emyarod:
@carmacleod I believe the issue is resolved now

carmacleod:
Much better, @emyarod - thanks!
Only one more thing: right-clicking on the icon (i.e. to open a context menu) should not trigger the tooltip.
Let me know if you would prefer this in a separate issue.

Question: Do people usually put these info tooltips inside a form control's label element? Is that a thing? It kind of messes with the focus, because the label content is a click target for the form control, so clicking on either the label text or the info icon focuses the form control, but then focus immediately moves to the info "button", so there's a focus flash that looks positively weird. We really need to get together and seriously design out all of the Carbon tooltips and make sure all use cases are covered.

emyarod:
> Only one more thing: right-clicking on the icon (i.e. to open a context menu) should not trigger the tooltip.

yeah I think we can address that in a separate issue as it's existing behavior and not a regression caused by the current PR @carmacleod 

> Do people usually put these info tooltips inside a form control's label element?

I don't think it's a pattern that Carbon endorses but I have seen examples of it before

carmacleod (approved):
+1 Looks good!
Thanks @emyarod !

I opened https://github.com/carbon-design-system/carbon/issues/7423 for the right-click issue.

PR #7329, opened 2020-11-20
fix(accordion): increases active class selector specificity

### LINKED ISSUE #7328: fix[accordion]: nested Accordions not working as expected
# fix[accordion]: nested Accordions not working as expected

- [x] `carbon-components`

## Detailed description

Vanilla Carbon Accordions are not functional when placed inside of another Accordion. Nested Accordions do not expand and collapse.

Yes. Accordion.

> see changed?

That Carbon Accordions can be placed one within another and function normally.

Chromium Edge (but applicable to all!)

v10.24.0 (latest)

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

IBM Social Program Management

1. Update `carbon/packages/components/src/components/accordion/accordion.hbs` to include one Accordion component within another.
2. Run `yarn dev`.
3. Navigate to `http://localhost:3000/demo/accordion`.
4. Observe that nested Accordions.

## Additional information
| Example of current behaviour |
|-|
| nested-accordion-example |

jpbyrne:
Closes #7328 

Vanilla Carbon Accordions are not functional when placed inside of another Accordion. Nested Accordions do not expand and collapse.

#### Changelog

**Changed**

- Vanilla Accordions can now placed in one another.

#### Testing / Reviewing

Confirm that when an Accordion appears inside another Accordion, both Accordion buttons toggle independently.


tw15egan (approved):

aledavila (changes requested):
Waiting on design to review as this issue was changed to a proposal that has not been reviewed yet.

aagonzales:
We are advising against putting accordions inside of accordions. For nesting content we recommend a using a tree view or figuring out a better ux solutions where only one level of nesting content occurs. 

In the example shown in the issue the first image looks like dropdowns inside of an accordion, not nested accordions. People have mentioned wanting a large page section type of accordion that could include more bits of ui. I think that would be fine but I think that is a new variant to the standard accordion. This as is accordion is meant for one level of content and be exclusive to content like what you see here on the [cloud docs](https://cloud.ibm.com/docs?tab=faqs) page. 
image

And the second doesn't have real content so its hard to evaluate but I would say that the second layer of collapsed content was unnecessary. Why can't that content just be shown open already? Or why aren't those nested bits in two separated top level buckets. 

image


For the moment, I don't think we should allow nesting accordions in this basic accordion from the system level. Your team is welcomed to do it on your end if you see that its best but I still have some ux concerns about promoting it as an acceptable system pattern.

aledavila:
Going to close this as this isn’t a pattern we want in our core components. Thank you

PR #7332, opened 2020-11-23
feat(Dropdown): remove selection for optional dropdowns

### LINKED ISSUE #6581: [Dropdown] Including a way to remove selection for optional dropdowns

AlexanderLyon:
Closes #6581

#### Changelog

**New**

- Adds `isOptional` prop to `<Dropdown/>` component. This defaults to `false` (defaulting to the existing functionality)
- Adds the ability to "remove" a previously chosen selection from dropdown by dynamically adding either the label text or "Select" as a choice to the beginning of `items` array.
  - Since this new option is dynamically added to the `items` array, the component now creates a deep copy of the `items` prop in order to prevent mutation of the original data structure.

#### Testing / Reviewing

* If the `isOptional` (boolean) prop is `true` and `label` prop is set, the first choice in the dropdown should be the `label`'s text
* If the `isOptional` (boolean) prop is `true` and `label` prop is _NOT_ set, the first choice in the dropdown should be "Select"

The above should work correctly regardless of if `items` an array of **strings** or an array of **objects**.

image


laurenmrice (approved):
Looks good to me! Thank you 🙌🏻

joshblack:
So sorry for the delay looking at this!

@AlexanderLyon @laurenmrice is the change basically that the first item wouldn't have the selection style beside it with this option turned on? (But it would still be the value of the dropdown)

laurenmrice:
You can select the dropdown prompt option instead of selecting Option 1, 2 or 3 to populate the dropdown with an added  selection. When the dropdown first opens, nothing is selected but the prompt option is presented at the top inside of the dropdown menu.

andreancardona:
@AlexanderLyon @laurenmrice Out of curiosity in this use case `If the isOptional (boolean) prop is true and label prop is NOT set, the first choice in the dropdown should be "Select"` ----- Do we want Select to be a disabled as an option? / greyed out of some sort?

AlexanderLyon:
@andreancardona The "Select" option was meant to be a default in the event that `label` was missing. By greying out/disabling it, wouldn't that remove the ability to deselect an option unless `label` is set?

On second thought this default option wouldn't be very flexible for non-english use as-is. Should I change this functionality to only add the option if a `label` prop is set?

laurenmrice (changes requested):
The prompt option in the dropdown menu should not be disabled. It should just be taken out completely if the optional functionality is not available. @andreancardona 

@AlexanderLyon We have two minor design changes from the original spec in the issue. (it was an oversight on my part). I attached an image spec below. Once this is changed, we should be good to go from a visual perspective. 👍🏻

- Brackets need to be added around the text in the field. "[Choose an option]"
- Brackets need to also be added to the optional selection text in the menu and it needs to have a selected state by default. 

dropdown

AlexanderLyon:
@laurenmrice Those changes have been made! Now:
* Optional functionality will only be enabled if `label` is set (documentation has been updated to reflect this new requirement)
* The optional value is now surrounded by brackets
* The optional value is now automatically selected, _BUT_ can still be overridden by `initialSelectedItem` if a user wants a different element to be selected by default

laurenmrice:
Awesome @AlexanderLyon ! That looks great in the `Optional` story! The prop in the Playground story needs to be updated to reflect that, when checked it just doesn't have brackets in the top field and the option does not have the selected state right now.

AlexanderLyon:
@laurenmrice Oops, this has been adjusted in the Playground story - could you confirm it looks correct on your end?

laurenmrice (approved):
Looks great @AlexanderLyon . Thank you! 🎉

andreancardona (approved):
looks good to me!

andreancardona (approved):
looks good to me 🥳 @AlexanderLyon

joshblack:
Hey all! 👋  Was taking a look at this (sorry for the delay!) and was curious if this was do-able with the current implementation. I tried to create a [Codesandbox](https://codesandbox.io/s/optimistic-glitter-mpizz?file=/src/index.js) to try and replicate what this is hoping to accomplish, let me know if I'm missing some implementation detail or part of the proposal 👀

AlexanderLyon:
Hi @joshblack - that would certainly work, however I was under the impression that the proposal was to allow the option to be dynamically added & selected via prop (please correct me if I'm wrong @laurenmrice). 

This way this default option wouldn't need to be explicitly added to the `items` data structure when supplied to the component.

laurenmrice:
@AlexanderLyon yes, that was the original proposal indicated in the issue. It appears what Josh has implemented in the code sandbox will actually achieve this with our existing code, so we will no longer need to make a prop to do this.

I am sorry about the miscommunication on our end for this particular issue/pr, the work you did for this was great! However we are going to close this pr since it is already achievable in code.

PR #7359, opened 2020-11-30
fix(dataTable): search accessibility issue

### LINKED ISSUE #7358: [Search a11y] aria-hidden property is set to true on focusable search elements
## Environment

> Operating system

> Browser

## Detailed description

Carbon-components-react

aria-hidden should be false for focusable elements. 

> What happened instead?
aria-hidden is set to true.

Rule ID: aria-hidden-focus
Ruleset: axe-core 3.4
User Impact: Serious 
WCAG: 4.1.2

1. Find a persistent data table search
2. Inspect the search input element.
3. Check value for aria-hidden. 

[https://codesandbox.io/s/datatable-with-batch-actions-uolbz](https://codesandbox.io/s/datatable-with-batch-actions-uolbz
)
## Additional information

Using the aria-hidden="true" attribute on an element removes the element and ALL of its child nodes from the accessibility API making it completely inaccessible to screen readers and other assistive technologies. Aria-hidden may be used with extreme caution to hide visibly rendered content from assistive technologies only if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. If aria-hidden is used to hide visible content from screen readers, the identical or equivalent meaning and functionality must be exposed to assistive technologies.

aria-hidden should not be true for items that allow input or focus.

CassidyJensen:
Closes #7358

This PR removes the aria-hidden on the DataTable Toolbar Search Input to remove a11y violations. The search is an input and a focusable element, therefore aria-hidden should never be true. 

#### Changelog

**Changed**

 aria-hidden: {!expanded} [removed]
Aria-hidden was being set by whether the search was expanded. Because it is focusable either way, aria-hidden should always be false. It does not need to be set. 

#### Testing / Reviewing

Inspect the search element on a data table. The input element should have no value for aria-hidden


emyarod (commented):
should the attribute just be removed instead of set to false?

CassidyJensen:
> should the attribute just be removed instead of set to false?

A good point, yes it probably should just be removed. I will update it. Thank you!

emyarod (approved):
thanks for the update, looks good to me

emyarod:
you most likely will need to update snapshots to pass CI (run `yarn test -u`)

CassidyJensen:
> you most likely will need to update snapshots to pass CI (run `yarn test -u`)

@emyarod 
I seem to be running into an issue when I run 'yarn test -u' : 
image

Do you have any advice for how to fix it?

emyarod:
I'm not sure what the test failures are without being able to see the output but I went ahead and updated the snapshots and pushed them up

CassidyJensen:
Thank you!

joshblack (approved):

PR #7368, opened 2020-11-30
feat(date-picker): add warning styles, style audit fixes

### LINKED ISSUE #7334: [date picker] audit bugs
## Simple date picker:
- [x] Remove `min-width`. Field should be able to have only 16px on either side.
- [x] Missing invalid state? (I couldn't find it may be there).
image

## Single date calendar:
- [x] Invalid state missing error icon
image
- [x] Next/previous month should be 4px away from the field.
image
- [x] In Firefox and Safari: not showing the correct `today` date.
- [x] Year selector:active in G90 and G100 incorrect. Should be using color `interactive-04`
image

## Storybook:
- [x] Can only "play" around with the single date picker. There is no way to use the various knob on the different kinds of date pickers
   - [x] Missing: simple date picker + disabled
   - [x] Missing: simple date picker + sizes
   - [x] Missing: simple date picker + light variant 
   - [x] Missing: simple date picker + invalid
- [x] Missing: Calendar, disabled `day`

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7334

Adds in the `warn` prop to align with other form components. Also fixes various style bugs, as well as updates the Storybook to view all props on each variation of `DatePicker`

#### Changelog

**New**

- `warn` prop was added to `DatePicker`
- `invalid` and `warn` now use the proper icons when active
- Styles associated with the error states and the icons
- `bx--date-picker-input__wrapper--invalid` and `bx--date-picker-input__wrapper--warn` added as wrapper classes to allow better specificity in our `invalid` styles in `form.scss`

**Changed**

- Fixed an issue with `warning` icons that were targeting a `[data-icon-path='inner-path']` that no longer exists
- `simple` DatePicker no longer has a `min-width` but will have two preset widths to handle the default state and error state. 

**Removed**

- extraneous styles

#### Testing / Reviewing

Check the playground tab and ensure the 3 variations of `DatePicker` look correct in all states. 

Keep in mind, a lot of extra styles were added to accommodate `simple` variation with all the wrapper classes. `simple` `DatePicker` almost seems like it could just be deprecated, as it is essentially just a `TextInput`.


aagonzales (approved):
Looks good to me!

andreancardona (approved):
@tw15egan looks good! :)

aledavila (approved):

PR #7370, opened 2020-11-30
feat(filterable-multiselect): add onMenuChange event

### LINKED ISSUE #6434: Filterable Multiselect onClose
Is there any plans to have an onClose callback to the Filterable Multiselect? I have seen some cases where my team and others have wanted to make selections and then trigger some action after all selections have been made and dropdown has closed vs triggering the onChange after a single selection change.

bstncartwright:
Closes #6434 

This is very similar to #7288, but instead adds the `onMenuChange` event functionality to the [`FilterableMultiSelect`](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/MultiSelect/FilterableMultiSelect.js) component.

#### Changelog

**New**

- `onMenuChange` event on filterable multiselect

**Changed**

**Removed**

#### Testing / Reviewing

- Start storybook
- Open filterable multiselect story
- Open the multiselect menu
- See onMenuChange: (1) [true]
- Close the multiselect menu
- See onMenuChange: (1) [false]


tw15egan:
Thanks for contributing @munkurious!

Do we want `onMenuChange` to be called when the menu is clicked but is already open? Seems like it is also being called when text is entered, but the menu hasn't changed

filterable-multi

bstncartwright:
Great catch @tw15egan, thanks! I've fixed most of those issues but I am still having one that I am not sure how to resolve due to my inexperience with downshift.

It seems that when you click out of the multiselect, it calls `onMenuChange` with `false` twice. I've been able to diagnose which lines are causing them:
https://github.com/munkurious/carbon/blob/e8d02129305c39c897dfdc47e7d2975ffb601137/packages/react/src/components/MultiSelect/FilterableMultiSelect.js#L206
and 
https://github.com/munkurious/carbon/blob/e8d02129305c39c897dfdc47e7d2975ffb601137/packages/react/src/components/MultiSelect/FilterableMultiSelect.js#L228

When clicking out of the multi select, the `handleOnStateChange` function and the `handleOnOuterClick` function gets called. I'm not sure of the best way to handle this and am welcome to any ideas :)

emyarod (commented):
<details>
  <summary>expand to view GIF</summary>
  
</details>

the `onMenuChange` event seems to fire twice when collapsing the menu

edit: looks like you're aware of this already, I started recording this while you were writing your comment above

packages/react/src/components/MultiSelect/FilterableMultiSelect.js
```diff
@@ -184,17 +190,20 @@ export default class FilterableMultiSelect extends React.Component {
...
+      this.props.onMenuChange(isOpen);
+    }
   };
 
   handleOnOuterClick = () => {
     this.setState({
-      isOpen: false,
       inputValue: '',
     });
+    if (this.state.isOpen) this.handleOnMenuChange(false);
```
emyarod:
```suggestion
```

I believe you can eliminate this line to avoid the event firing twice on close, since you are already firing it on state change

bstncartwright: Yep, that worked perfectly, thanks!

bstncartwright:
Awesome, thanks for your help! It seems to do the job :+1:

emyarod (approved):
looks good to me

tw15egan (approved):
LGTM, thanks for contributing! 👍 ✅

PR #7373, opened 2020-12-01
fix(checkbox): correct line-height

janhassel:
For checkboxes, `line-height` was used to vertically center the label with the checkbox, resulting in a visual issue when the checkbox label would wrap. This PR removes the `line-height` override and adds a `padding-top` to achieve alignment.

#### Changelog

**Changed**

- Replaced `line-height` with `padding-top` implementation in `_checkbox.scss` to achieve vertical center

#### Testing / Reviewing

Pass a longer label to the Checkbox playground story and reduce the window's width until the checkbox label wraps. Compare netlify preview with current react storybook.

Before
image

After
image


tw15egan (approved):

laurenmrice (approved):
Thank you Jan! Looks good to me  🙌🏻

andreancardona (approved):
@janhassel looks good to merge! :)

PR #7380, opened 2020-12-01
feat(toggle): convert ToggleSmall to a variant of Toggle

### LINKED ISSUE #6653: ToggleSmall should be a variant of toggle
Ala button

motivation: This is what props are best used for. We should stick to the pattern of using kind/size prop rather than creating a whole new component.

### LINKED ISSUE #7327: React ToggleSmall params typing requires `aria-label` parameter
- [x] `carbon-components`

In a Typescript project using carbon-components-react, try creating a `<ToggleSmall>` element.

That element requires prop `aria-label`, which is not a valid JSX prop name. I don't know how that prop even *can* be provided. So ToggleSmall is unusable without disabling type checking.

I expect this would be fixed by https://github.com/carbon-design-system/carbon/issues/6653 , which I would love to see happen.

Using Typescript 4.0.5

toggle-small

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6653
Closes https://github.com/carbon-design-system/carbon/issues/7327
Refs https://github.com/IBM/carbon-components-svelte/pull/401

Deprecates `ToggleSmall` and adds `size` prop to `Toggle` since the two components are basically identical.

#### Changelog

**New**

- `size` prop for `Toggle` that accepts `sm` to enable the small toggle variant
- Tests for the new prop in `Toggle`

**Changed**

- Added deprecation warnings when using `ToggleSmall` in Dev mode
- Added deprecation notes for `ToggleSmall` to the storybook
- Workarounds to keep `ToggleSmall` tests from failing because of the deprecation warning

**Removed**

- Unnecessary stories 

#### Testing / Reviewing

Try changing the size of `Toggle` in the storybook and ensure all styles look correct


jeanservaas (approved):
cool

andreancardona (approved):
looks good!

packages/react/src/components/ToggleSmall/ToggleSmall-test.js
```diff
@@ -6,24 +6,48 @@
...
+    ToggleSmall = require('../ToggleSmall').default;
+
+    // const wrapper = mount(
+    //   <ToggleSmall
+    //     id="toggle-1"
+    //     aria-label="test label"
+    //     labelA="Off"
+    //     labelB="On"
+    //   />
+    // );
```
emyarod:
```suggestion
```

emyarod (approved):
looks good to me

PR #7383, opened 2020-12-01
fix(TableHeader): update sortable table header aria description for voice over

### LINKED ISSUE #7369: Data table sortable headers visual and programmatic labels do not match
The programmatic and visual labels in the DataTable headers do not match. This is causing an AC violation
image

The programmatic (aria) label needs to match the visual label (for example "Product/Offering Name") or you'll have a Accessibility Checker violation that looks like this

a way of preserving the current label format and having assistive technologies describe usage information that's implied visually to a sighted user would be having some situation like this

```html
<div style="display: none" id="usage-description-1">Click to sort</div>
<button aria-describedby="usage-description-1">Product/Offering Name</button>
```

That will read the label, element type, and it's intended effect nicely and avoid any AAT errors :thumbsup:

jnm2377:
Closes #7369 

#### Changelog
- removed aria-labels from icons
- added a hidden label for sortable button, which announces sorting for each header

When tabbing/entering a header this is announced: (description of how to sort, which way it will sort/by which header (name) you are sorting).
Screen Shot 2020-12-01 at 4 09 39 PM


packages/react/src/components/DataTable/TableHeader.js
```diff
@@ -18,24 +18,27 @@ import { sortStates } from './state/sorting';
 const { prefix } = settings;
 
 const translationKeys = {
-  iconDescription: 'carbon.table.header.icon.description',
+  buttonDescription: 'carbon.table.header.icon.description',
```
jnm2377, line 21: Note: Tried to change `carbon.table.header.icon.description` to `carbon.table.header.button.description` since this label is no longer the icon, but the button. However, I got a warning about the public api, so I left it as is.
Screen Shot 2020-12-01 at 4 34 44 PM

dakahn (approved):
Wonderful! Speaks really well in my tests with JAWS 2020 and NVDA 👏 👏 

Also got an all-clear from Accessibility Checker

tw15egan (approved):
This is great, nice work! 🎉 ✅ 👍

Issue #7360, opened 2020-11-30
Create a cli command for scaffolding out components


joshblack:
This issue falls under [this KR](https://app.zenhub.com/workspaces/okrs-5eef636e504e9d001694ba22/issues/carbon-design-system/okrs/93)

We would like to add a new command to `@carbon/cli` that our team, and others at IBM, could use to scaffold out a component. This scaffold should include a minimal set of conventions that we encourage for teams authoring components, in particular the folder scheme and test conventions.

In addition to generate a standalone folder as an option, this command should include a `--lib` option (or similarly named) for scaffolding out a component package. This option should be configurable for the following conventions:

- Testing (Jest)
- Building (Rollup)
- Development (Storybook)
- Linting (ESLint, StyleLint)
- Commit hooks

mattrosno:
@joshblack if a component package, can we also include a `carbon` object in the package.json? If a component folder, would that be a carbon.json or something? What goes in there? Not sure... but eventually information that would replace component index YML files? E.g. https://github.com/carbon-design-system/carbon-website/blob/master/src/data/index/dot-com/callout-media.yml

The `carbon` config would allow component maintainers to self-document that component's "completeness" (e.g. status, if it's the canonical implementation, how accessible it is, who maintains, etc. Anything of value that would help a user understand the component's history and if it's the right component for their "job".)

Some related concepts:

* From https://github.com/carbon-design-system/carbon-spec/pull/13, possible component status definitions: https://github.com/carbon-design-system/carbon-spec/blob/ffea894324da8f3a993319e98a4bd652e8145a15/docs/component-status.md#legend

mattrosno:
As for the folder scheme, we've discussed that containing specific markdown files (or structured data somehow) to codify component / pattern templates to assist documentation standardization across the ecosystem.

* https://www.carbondesignsystem.com/contributing/component
* https://www.carbondesignsystem.com/contributing/pattern#how-to-write-a-pattern

joshblack:
@mattrosno definitely could add that metadata and include it baked-in, makes a lot of sense to me 👍

PR #7384, opened 2020-12-01
feat(cli): add component command

joshblack:
Reference #7360 

This PR creates an initial command for creating a component. This does not include the package scaffold, instead it scaffolds out files inside of our react project following our current practices.

#### Changelog

**New**

- Add `component` command to `@carbon/cli`

**Changed**

**Removed**

#### Testing / Reviewing

- Try and create your own component with it! You can reference the CLI with the path `./packages/cli/bin/carbon-cli.js` and can call it directly from your shell
- See if tests and stories work 😬


dakahn:
Should there be output when the command is run?
image

joshblack:
@dakahn sorry, to be specific the command would be:

```bash
./packages/cli/bin/carbon-cli.js component
```

tw15egan:
cli

Screen Shot 2020-12-02 at 12 36 03 PM

Very cool!

dakahn:
maybe not a blocker, but there's a permission denied error on WSL
image

joshblack:
@dakahn I _think_ that's because the path you gave `/caverns` is at `/caverns` instead of a relative path (like `components/caverns`) and the profile trying to run the command doesn't have permissions to write to that path

dakahn (approved):
That worked! More WSL weirdness

tw15egan (approved):

PR #7392, opened 2020-12-03
fix(data-table): increase interactive area of expand button

### LINKED ISSUE #6589: [DataTable]: Expansion clickable area is not up to specification

- [x] `carbon-components`

## Detailed description

According to the [DataTable guidance on the Carbon website](https://www.carbondesignsystem.com/components/data-table/usage#interactions), the clickable / interactive area to expand a row is a square filling the entire height of the row, meaning with the default row height of 48px, the clickable area is 48px by 48px.

Currently in react implementation, this interactive area is 16px by 16px.

`DataTable.TableExpandRow`

Chrome 84

`carbon-components@10.16.0`
`carbon-components-react@7.16.0`

1. Visit react storybook (https://react.carbondesignsystem.com/?path=/story/datatable--with-expansion)
2. Try to click "around" the chevron

## Additional information

Click targets according to Carbon website:
image

Screen recording of react storybook:
Screen Recording 2020-07-30 at 14 44 03

emyarod:
Closes #6589

This PR adjusts the size of the interactive area for the data table expand button to match the spec

#### Changelog

**Changed**

- remove duplicate style rules
- increase table expand button size and styles

#### Testing / Reviewing

Confirm that the data table expansion button interactive area is increased and no style regressions were introduced

### DIFF
--- a/packages/components/src/components/data-table/_data-table-core.scss
+++ b/packages/components/src/components/data-table/_data-table-core.scss
@@ -292,8 +292,26 @@
 
   .#{$prefix}--data-table thead th.#{$prefix}--table-expand,
   .#{$prefix}--data-table tbody td.#{$prefix}--table-expand {
-    // 16px padding left + 16px checkbox width
+    width: rem(48px);
+    height: rem(48px);
+  }
+
+  .#{$prefix}--data-table--compact thead th.#{$prefix}--table-expand,
+  .#{$prefix}--data-table--compact tbody td.#{$prefix}--table-expand {
+    width: rem(24px);
+    height: rem(24px);
+  }
+
+  .#{$prefix}--data-table--short thead th.#{$prefix}--table-expand,
+  .#{$prefix}--data-table--short tbody td.#{$prefix}--table-expand {
     width: rem(32px);
+    height: rem(32px);
+  }
+
+  .#{$prefix}--data-table--tall thead th.#{$prefix}--table-expand,
+  .#{$prefix}--data-table--tall tbody td.#{$prefix}--table-expand {
+    width: rem(64px);
+    height: rem(64px);
   }
 
   .#{$prefix}--data-table--tall .#{$prefix}--table-column-checkbox {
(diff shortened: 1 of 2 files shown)


laurenmrice (changes requested):
For the expansion story, the rows should be 48px in height (It is currently showing 49px). The expansion does not offer the other sizes so I can only see the default `none` size option. The playground story seems to be doing this correctly, however for the `short` size its not showing a whole number for the height. It should be 32. https://deploy-preview-7392--carbon-components-react.netlify.app/?path=/story/datatable--playground.

Screen Shot 2020-12-04 at 12 28 25 PM


I think once the height of the row is fixed, the focus size around the chevron in each row will line up properly.

laurenmrice (changes requested):
-The height of the default rows for the Expansion story are 48px now 👍🏻

-The focus around the chevron for each row should be 48px x 48px. Right now the focus is missing 1px in height. 
Screen Shot 2020-12-09 at 1 57 35 PM

emyarod:
@laurenmrice the button is 47px tall because it's within the 48px tall row and the row has a 1px gray border at the bottom. in order to resolve this I think we would need to rewrite how we handle zebra style row hovers

laurenmrice:
Okay, that is good to know. This is why I bring this up, comparing dropdown with data table that have similar visual usage of rules, hovers and focus.

Artboard

tw15egan:
@emyarod looking at the `Dropdown` styles, it looks like we did add calc + 1px for the border. I think if we do

```css
.bx--table-expand__button { 
     height: calc(100% + 1px); 
}
```

 we'll get the intended result

emyarod:
@tw15egan yeah I think that hack is more viable with the listbox components since they are not stacked in rows on top of each other. adding that height rule with our current table row styles would work for the table body but I think we will need more one off rules for the header. if we are fine with that I can look into it more

tw15egan (approved):
LGTM 👍 ✅

laurenmrice (approved):
lgtm! 👍🏻

aledavila (approved):

Issue #5713, opened 2020-03-26
Pass down options to Flatpickr on the DatePicker component


enricoberti:
### Summary

It would be nice to be able to pass down to Flatpickr itself some options that are you can't set on the flatpickr instance itself (ie. `inline: true`)

In the React DatePicker component it's a matter of adding a prop and then spreading it when initing flatpickr

### Justification

We do have the need to have an inline/static picker on carbon-addons-iot for [this story](https://github.com/IBM/carbon-addons-iot-react/issues/854)

As it right now, we are reparenting the calendar DOM into our component but this is far from the best option to do it.

### Specific timeline issues / requests

For now we are going with the workaround but it would be nice to have soon :)

### Available extra resources

I can contribute back myself but I need some support from the Carbon team given that adding a new prop to DatePicker would break semver.

Thanks!

emyarod:
I believe the restriction on flatpickr options configuration is intentional, specifically around the `inline` prop

cc @carbon-design-system/design on whether or not inline date picker is a valid variant within the design system

aagonzales:
It was originally intentional to not include the inline format for datepicker. I think if we wanted to include it, it would need some visual exploration because of the calendar icon. This topic needs some investigation from the design perspective before we can determine if we should do it or not.

enricoberti:
hi @aagonzales!  design-wise you can also ping @lukefirth and @cynthiaz who worked on https://github.com/IBM/carbon-addons-iot-react/issues/854

jinhei:
Along the same lines, it would be nice to be able to pass down `disabled` and `enabled` flags to flatpickr, which we're currently unable to do. Our use case is that only certain days should be selectable based on availability.

See [flatpickr example documentation](https://flatpickr.js.org/examples/#disabling-specific-dates).

davidicus:
Any discussion still being had on this?

PR #7393, opened 2020-12-03
Component - React - DatePicker -Enable support for inline calendar 

Matthewvandenbussche:
Closes #

Within the Carbon React Datepicker component, the flatpicker `inline` option should be exposed.

#### Changelog

**New**

- Added inline prop to the Datepicker and passed the value when initializing flat picker.

**Changed**

- Update associated documentation

**Removed**

- none

#### Testing / Reviewing

Enable inline prop within the playground.

### DIFF
--- a/packages/react/src/components/DatePicker/DatePicker.js
+++ b/packages/react/src/components/DatePicker/DatePicker.js
@@ -269,6 +269,7 @@ export default class DatePicker extends Component {
       datePickerType,
       dateFormat,
       locale,
+      inline
       minDate,
       maxDate,
       value,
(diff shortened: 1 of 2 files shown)


emyarod (commented):
as far as inline date picker goes, it was intentionally left out of the flatpickr options due to a design decision (ref https://github.com/carbon-design-system/carbon/issues/5713#issuecomment-607416268) cc @aagonzales

joshblack:
Hey there! Going to close this out due to lack of activity, feel free to comment and I can re-open!

PR #7396, opened 2020-12-03
chore(project): remove react-hooks package

joshblack:
The `react-hooks` package unfortunately was an idea that never really moved forward. As a result, this PR removes this private package from our monorepo (and hopefully helps speed up builds by a bit 🤞 )

#### Changelog

**New**

**Changed**

**Removed**

- Removed `packages/react-hooks`


emyarod (approved):
looks good to me

tw15egan (approved):

PR #7400, opened 2020-12-04
style(code-snippet): increase expand button size

### LINKED ISSUE #7397: [code snippet] multi-line change 'show more' button size to 40px
In the multi-line code snippet, change the button size of the "show more/less" button to be 40px instead of 32px so that when you hover on it both the right and bottom edges are flush with the edge.

## As-is:
image

## To-be:
image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7397

Increases size of ` show more ` button in CodeSnippet

#### Changelog

**Changed**

- set `bottom: 0`

**Removed**

- `size="small"` from expand button

#### Testing / Reviewing

Ensure the `show more` button looks correct


laurenmrice (changes requested):
- The alignment of the show/less more button looks good, but the container size needs to be `40px` height instead of 48px.

- When using the light prop, the button needs to have the light tokens applied to it (the copy button is already doing this). 
code snippet button

tw15egan:
@laurenmrice updated!

aledavila (approved):
Lgtm

laurenmrice (approved):
Looks great. Thank you! 🎉

emyarod (approved):

PR #7403, opened 2020-12-04
ci(actions): add workflow to replace circle ci

joshblack:
Adds in a GitHub workflow to replace our CircleCI config. This PR only adds in the workflow to make sure that it works on merge and for PRs afterwards, after which we can remove the CircleCI config 👀 

#### Changelog

**New**

- Add GitHub workflow to replace CircleCI

**Changed**

**Removed**

#### Testing / Reviewing

Unfortunately, GitHub Actions don't run from forks 😞  Really the only thing we can do is double-check that the syntax is valid and that it does something similar to CircleCI for our main CI process.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  ci-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 12.x
```
tw15egan: Do we want to bump this to 14?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
```
joshblack:
```suggestion
      - name: Use Node.js 14.x
        uses: actions/setup-node@v1
        with:
          node-version: '14.x'
```

tw15egan (approved):

emyarod (approved):

PR #7409, opened 2020-12-07
style(modal): remove type enforcement inside Modal body

### LINKED ISSUE #7404: Modal overrides the font-size of all its content
Modal overrides the font size of its content using the  `.bx--modal-content` class.

image

- [ ] `carbon-components`

## Detailed description

I was trying to use a `h3` heading inside a Modal body, but hit an issue where the HTML heading doesn't change the font size. It looks like the `.bx--modal-content` class. is overriding all the font properties for anything in the Modal body. Is this by design? If so, what is the recommended way to have different heading sizes in a Modal body?

https://codesandbox.io/s/carbon-headings-ym43d

Chrome 87.0.4280.88 (Official Build) (64-bit)

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7404

We were setting all text inside Modal body to `body-long-01`. This removes that enforcement

#### Changelog

**Removed**

- Typography styles being set to all sub-elements of Modal content

#### Testing / Reviewing

Place an `h1` inside `Modal` and ensure the text is the correct size


aagonzales (approved):
Looks good!

joshblack (approved):

aledavila (approved):

PR #7414, opened 2020-12-08
fix(tooltip): correct scroll width caused by #7309

lee-chase:
The alignment of the tooltip created in #7309 moved the inner (content) part of the tip. This meant that the outer part caused the width of the component to include space outside the display area of the tip, even when the tip was shown with the direction of the setting: bottom align: end. This could cause a scroll bar in circumstances where the tip appeared to be displayed entirely within the available space.

In addition #7309 appeared to square off the top corners of the tip.

This PR reverts the SCSS changes in #7309 and instead styles the outer tooltip element and carrot to the correct positions.

#### Changelog

**Changed**

packages/components/src/components/tooltip/_tooltip.scss 

#### Testing / Reviewing

Visually inspected the 8 off center cases.


dakahn (approved):
Seems great!

tw15egan (approved):
LGTM 👍 ✅

jeanservaas (approved):

PR #7418, opened 2020-12-08
fix(Tab): implement renderButton support

### LINKED ISSUE #7416: Tab: renderAnchor/renderButton no longer runs cleanly
- [x] `carbon-components`

## Detailed description
The following PR, https://github.com/carbon-design-system/carbon/pull/6989, deprecated `renderAnchor`, suggesting an alternative `renderButton` to achieve the same result. However, the current state of the Tab component means that it is no longer possible for it to run cleanly with either prop without a warning.

Using renderAnchor produces the deprecated warning.
https://codesandbox.io/s/festive-jones-i6x5k?file=/src/index.js

Using renderButton produces an unrecognised prop warning and also doesn't work
https://codesandbox.io/s/pensive-chatterjee-9mxnm?file=/src/index.js

> see changed?

If the intention is to rename renderAnchor to renderButton then the new prop should be available to use in PropTypes.

emyarod:
Closes #7416

This PR adds support for `renderButton` due to `renderAnchor` deprecation in the Tabs component

#### Changelog

**New**

- `renderButton` support

**Changed**

- avoid spreading `renderButton` into `li`

#### Testing / Reviewing

Confirm that using `renderButton` no longer emits an unrecognized prop warning and using `renderAnchor` continues to display deprecation warnings in Tabs


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #7421, opened 2020-12-08
fix(text-area): remove aria-placeholder prop

### LINKED ISSUE #7132: Textarea shows both placeholder and aria-placeholder, and thats an issue

joshblack:
Closes #7132 

Based on the discussion in #7132, it seems like we can safely remove `aria-placeholder` and resolve the violation.

#### Changelog

**New**

**Changed**

- We no longer pass in `placeholder` to `aria-placeholder` in `TextArea`

**Removed**


tw15egan (approved):

emyarod (approved):
not sure why I was unassigned as I was investigating this along with @carmacleod, but this solution seems to be fine as the JAWS issue that prompted the original change could not be replicated

joshblack:
@emyarod just threw something up for it, didn't mean anything by it!

PR #7434, opened 2020-12-10
fix(Tooltip): prevent right click from triggering tooltip

### LINKED ISSUE #7423: right-clicking on a tooltip button (i.e. to open a context menu) should not trigger the tooltip
Steps:
1. [Carbon Tooltip example](https://carbon-components-react.netlify.app/?path=/story/tooltip--default-bottom)
2. Right-click on the "tooltip" info (ℹ️) button
3. Tooltip opens and context menu opens

Should only open context menu.

emyarod:
Closes #7423

This PR modifies the `handleMouse` method to include right clicks and prevent the tooltip from opening when the trigger button is right clicked

#### Testing / Reviewing

Confirm that the tooltips are not opened by right click events and that no behavioral regressions were introduced


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #7435, opened 2020-12-10
fix(Tooltip): restore text selection functionality

emyarod:
This PR resolves a regression where noninteractive tooltip content could not be selected or highlighted. The `blur` event is now ignored if a user clicks within the tooltip element

#### Testing / Reviewing

Confirm that noninteractive tooltip content can be selected/highlighted without the tooltip collapsing


tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

PR #7437, opened 2020-12-10
fix(TooltipDefinition): update `children` PropType

### LINKED ISSUE #6484: [Tooltip/TooltipDefinition] Support for hover or truncation with ellipses

emyarod:
Closes #6484

This PR changes the TooltipDefinition's `children` prop type to support nodes instead of restricting it to strings (alternatively, we can insert a span that wraps the `children` if we want to keep the proptype definition as `string`). This will allow users to truncate the tooltip trigger text if necessary with CSS, and our existing guidelines dictate that the definition tooltip should be used for small amounts of text

#### Changelog

**Changed**

- TooltipDefinition `children` proptype definition


joshblack:
If I understand correctly, this is allowing truncation of the text that triggers the tooltip? If so, do we need to worry about how someone could see what the truncated text itself is? Seems like we solved that in the past with tooltips or (unfortunately) title, but if this itself has a tooltip it seems like it would conflict.

Let me know if I'm misunderstanding, just was curious 🤔

emyarod:
we wouldn't be able to control the truncation logic since that would all be on the application's side. this change is just for the proptype definition since the user is technically already able to truncate the tooltip trigger

andreancardona (approved):
looks good!

wait.. 

@emyarod is this still pending a question?

joshblack (approved):

PR #7447, opened 2020-12-11
fix(Loading): remove extra padding on loading spinners

### LINKED ISSUE #6814: [Large loading component]: Loading has different size to defined on carbondesignsystem website

- [X] `carbon-components`

## Detailed description

The Carbon Design Website Defines the .bx--loading class as having a width/height of 88px/5.5rem.

The value defined for .bx--loading in the file packages/components/src/components/loading/_vars.scss is 10.5rem

It looks like the ultimate displayed spinner is the correct size as there is a padding of 40px (which times 2) reduces the spinner size down to 88px.

However, it does not explain the discrepancy between the .bx-loading class value defined on the website and the value for the same class used in the code.

emyarod:
Closes #6814

This PR removes the extra padding on our loading spinners so that it reflects the spec more accurately

#### Testing / Reviewing

Confirm the regular and small loading spinners more closely match the spec


aledavila (approved):

laurenmrice (approved):
Looks good. Thank you ! 👍🏻

tw15egan (approved):

joshblack (approved):

PR #7456, opened 2020-12-14
fix(OverflowMenu): check for custom closeMenu method

### LINKED ISSUE #6412: If I use Dropddown or ComboBox in TableToolbarMenu, the popup menu will be closed upon selecting an option

- [x] `@carbon/icons-react - 10.13.0`
- [x] `carbon-components - 10.15.0`
- [x] `carbon-components-react - 7.15.0`

## Detailed description

Basically whenever I select an option in `Dropdown` or `Combobox`, the `TableToolbarMenu` popup will be closed. This behavior is not seen with `Select`

> see changed?  
  
I expect the behavior to be the same with `Select`  
  
Chrome

10.15.0

>
> - React:
>   https://codesandbox.io/s/carbon-components-react-7uqsd?file=/src/index.js

emyarod:
Closes #6412

This PR prevents custom `closeMenu` methods from being overwritten by the default method in order to support a wider range of interactions on the overflow menu

#### Changelog

**Changed**

- check for OverflowMenuItem `closeMenu` before cloning the element and assigning a default value for the function

#### Testing / Reviewing

confirm that adding a custom `closeMenu` method on an `OverflowMenuItem` will not be overwritten


packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -494,7 +494,7 @@ class OverflowMenu extends Component {
     const childrenWithProps = React.Children.toArray(children).map(
       (child, index) =>
         React.cloneElement(child, {
-          closeMenu: this.closeMenu,
+          closeMenu: child.props.closeMenu || this.closeMenu,
```
joshblack, line 497: Would we need to run both functions instead of deferring to a given `child` `closeMenu`? Seems like `this.closeMenu` has some state logic that we want to run but I might be misunderstanding 👀

emyarod: the main intention is to allow a function that is different from or replaces the default `closeMenu` (without expanding the component API). if the user still wants to run the existing logic they can use the existing `onClose` prop. maybe it would be more preferable to just add a boolean prop that controls whether or not the menu will close on selection of an item?

joshblack (approved):

aledavila (approved):

PR #7466, opened 2020-12-16
fix(combo-box): state disabled background color

epodgaetskiy:
This PR makes compliance with [the guide](https://www.carbondesignsystem.com/components/dropdown/style#interactive-states)

#### Changelog
**Changed**
- Indicated a specific color for the disabled state. The error was not visible because the colors in the theme correspond to `field-01` and` disabled-01`.

#### Testing / Reviewing

- Pull PR
- Look fix at netlify storybook for combobox :disabled

### DIFF
--- a/packages/components/src/components/list-box/_list-box.scss
+++ b/packages/components/src/components/list-box/_list-box.scss
@@ -163,7 +163,7 @@ $list-box-menu-width: rem(300px);
 
   // Disabled state for `list-box`
   .#{$prefix}--list-box--disabled:hover {
-    background-color: $field-01;
+    background-color: $disabled-01;
   }
 
   .#{$prefix}--list-box--light.#{$prefix}--list-box--disabled {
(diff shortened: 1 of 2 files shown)


packages/components/src/components/list-box/_list-box.scss
```diff
@@ -173,6 +173,7 @@ $list-box-menu-width: rem(300px);
   .#{$prefix}--list-box--disabled,
   .#{$prefix}--list-box--disabled .#{$prefix}--list-box__field,
   .#{$prefix}--list-box--disabled .#{$prefix}--list-box__field:focus {
+    background-color: $disabled-01;
```
tw15egan, line 176: I don't believe we want `light` variants having a `disabled` background.

tw15egan:
@epodgaetskiy I'd create issues for any more inconsistencies between the website and components, as there is an ongoing component audit. In some cases the website is incorrect and needs to be updated rather than the component itself

epodgaetskiy:
@tw15egan Yes of course, but logically, the colors of the states combo-box must match with[dropdown](https://github.com/carbon-design-system/carbon/pull/7461).

tw15egan:
I agree, I think we need to revert that PR then, as I was unaware that that was targeting the `light` variant. This is why working off of issues would be really helpful, so we can get clarification from design before making any PR's

tw15egan:
You can see the hover states with `light` and `disabled` are now broken: https://deploy-preview-7461--carbon-components-react.netlify.app/?path=/story/dropdown--playground

tw15egan:
Closing this because this is currently working as expected

PR #7468, opened 2020-12-16
fix(select): update light disabled state

### LINKED ISSUE #7464: Select light disabled not light

- [ ] `carbon-components`

## Detailed description

The `Select` component in the light variation does not render as a disabled light input

Yes, the `Select` component

> see changed?

I expected the disabled light `Select` to have the same colors as other disabled text inputs for consistency.

Chrome

7.25.0

> should be aware of?

1. Go to the `Select` component storybook
2. Set `light` and `disabled` knobs to true
3. Toggle the `light` knob and notice that there is no difference
4. To confirm the expected behavior, do the same steps on the `TextInput` component
5. Notice that the `TextInput` component had different disabled colors for light and non-light variations

## Additional information

image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7464

Uses `field-02` as the background color when `light` prop is enabled, to stay consistent with other `light` form elements

#### Changelog

**Changed**

- The background is `field-02` when disabled

#### Testing / Reviewing

Toggle the `light` prop, then `disabled` for `TextInput`. Then do the same for `Select`. They should match.

FYI @diego-codes


emyarod (approved):

jeanservaas (approved):

andreancardona (approved):
@tw15egan looks good to me 🌮

Issue #3216, opened 2019-06-28
[Notification] Dark theme icon color wrong


shixiedesign:
In warning notifications, center exclamation mark should be **black in all themes**. We do not have a token with such value. This is okay to be hard coded in this situation. 

![image](https://user-images.githubusercontent.com/15144993/60311505-2546a980-991d-11e9-8743-590176661ed7.png)

PR #7474, opened 2020-12-16
fix(FileUploader): resolve component audit issues

### LINKED ISSUE #7448: [file uploader] component audit bugs
## General fixes

**Content**
- [x] Header and description of the component should say the following:

----
**File focus state**
- [x] Can we make the focus around the close focus border `24px`?
focus
----
**File invalid state**

- [x] There is a slight shift down of the file label when the invalid state is applied.
error

- [x] Invalid state does not adjust with different heights of the file.
error

- [x] Error message text should be `caption-01`.
- [ ] Alert icon needs a `black` inner fill for dark themes.

----

## File uploader button story
-Why do we have a `file uploader button` story? We do not want to encourage people to upload a file inside of a button container. Is this something we can deprecate?

## File uploader and File uploader item story
- [x] For the demo of this story, the width of the file should be 288px. Right now its changing dynamically based on the length of the word. People can edit the size of this if they want to on their end.
size

## Drag and drop container example application story

The disabled state is not correct.
- [x] The header and description need to be `disabled-02`.
- [x] The container needs to be `disabled-02`.

emyarod:
Closes #7448

This PR updates the file uploader examples to resolve component audit bugs. A few questions about the audit points:

>  Alert icon needs a `black` inner fill for dark themes.

for the black fill within alert icons on dark themes, we run into the same issue with notifications from before (https://github.com/carbon-design-system/carbon/issues/3216) except we're not able to force file uploaders to be on dark backgrounds. alternatively we can just apply the black fill across the board on all themes if that's fine

>  The container needs to be `disabled-02`.

which container is this referring to?

#### Testing / Reviewing

Confirm the file uploaders match the updated spec

### DIFF
--- a/packages/components/src/components/file-uploader/_file-uploader.scss
+++ b/packages/components/src/components/file-uploader/_file-uploader.scss
@@ -185,7 +185,6 @@
     max-width: rem(320px);
     margin-bottom: $carbon--spacing-03;
     background-color: $field-01;
-
     outline-width: 1px;
   }
 
@@ -286,7 +285,6 @@
   .#{$prefix}--file__state-container .#{$prefix}--file-invalid {
     width: $carbon--spacing-05;
     height: $carbon--spacing-05;
-    margin-right: $carbon--spacing-03;
     fill: $support-01;
   }
 
--- a/packages/components/src/components/file-uploader/_file-uploader.scss
+++ b/packages/components/src/components/file-uploader/_file-uploader.scss
@@ -119,8 +119,8 @@
   .#{$prefix}--file__selected-file {
     display: grid;
     grid-auto-rows: auto;
-    grid-gap: $carbon--spacing-05;
     grid-template-columns: 1fr auto;
+    gap: rem(14px) $carbon--spacing-05;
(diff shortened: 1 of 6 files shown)


tw15egan (approved):
LGTM 👍 ✅

laurenmrice (changes requested):
**Default file uploader**
- Header needs to say “Upload files”
- Button text needs to say “Add file"


**Uploaded File**
Invalid state
- There is still a shift down when the invalid state is applied. The background of the field should increase in height to accommodate the error message below it. But the text and icons in the field should stay in the same place and not be jumping down.
invalid state


- There needs to be a 8px space between the alert icon and the close icon.
Artboard


**File Uploader story**
- The length of the file changes in width depending on the length of the text inside of it. Can we fix this so the file always has a width of 288px like it does in the other examples we show?

laurenmrice (commented):
The invalid state looks perfect for the `default` size file! It is still jumping down for the `field` and `small` sizes.

laurenmrice (approved):
Looks great! 👏

PR #7477, opened 2020-12-17
refactor(select): remove unused disabled option styles

### LINKED ISSUE #7475: CSS stlying is not property for disabled select item in firefox
firefix does not support to set the opacity for select item, https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/select/_select.scss#L164

it's not a grey out styling for the disabled select item in firefox.

https://stackoverflow.com/questions/58437806/why-wont-firefox-grey-out-my-disabled-select-options

emyarod:
Closes #7475

This PR removes unused styles on the Select component for disabled options. The disabled component already matches the component spec, and opacity rules on disabled `<select>` options are not rendered in the browsers we support

#### Testing / Reviewing

Confirm there are no visual component regressions

### DIFF
--- a/packages/components/src/components/select/_select.scss
+++ b/packages/components/src/components/select/_select.scss
@@ -51,6 +51,8 @@
     border-bottom: 1px solid $ui-04;
     border-radius: 0;
     cursor: pointer;
+    // reset disabled <select> opacity
+    opacity: 1;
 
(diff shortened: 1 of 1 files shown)


laurenmrice (approved):
Looks great! Thank you 🙌🏻

andreancardona (approved):
@emyarod tested & reviewed 👍 lgtm.

aledavila (approved):

PR #7493, opened 2020-12-22
feat(translations): pass reference to default translate function (#7492)

### LINKED ISSUE #7492: Easier access to default translations for component labels to use as a fallback.
### Summary

Asking for: development

The issue: 
It would be helpful to have access to the default `translationWithId` function carbon defines in many component's `defaultProps` as a fallback to use.
Translations for keys may not exist for a couple of reasons:
1. Key(s) not yet localized for a particular language.
2. Carbon package was upgraded in project that brought new translation keys (circling back to 1)

Passing a custom `translateWithId` function loses the default translation values.

### Justification

To effectively provide translations where possible without needing to maintain copies of the default carbon english translations. It will more easily allow upgrading carbon and filling in the translation gaps at later times.

My desire is to have the implementation for our custom `translateWithId` to look like this:

```typescript
// NEW: defaultTranslateWithId
const customTranslateWithId = (messageId: string, args: { [key: string]: string } | undefined, defaultTranslateWithId: (key: string) => string) => {
  if (i18nextInstance.hasKey(messageId)) {
    return i18nextInstance.get(messageId);
  }
  return defaultTranslateWithId(messageId);
}
```

If every `translateWithId` function in carbon has this or a similar signature then I can write a single function that accepts an i18next instance an returns the above function to pass to each carbon component. The function only needs to be defined once and will work across all carbon components because each individual component will pass its `defaultTranslateWithId` dynamically.

~~An alternative solution: allow returning `null` in a custom `translateWithId` function. When `null` is returned, carbon internally falls back to the default `translateWithId`.~~

No timeline. I have a workaround: manually composing all the component's `Component.defaultProps.translateWithId` into a single function but this needs to be manually maintained and updated.

I may be able to submit a PR for this time permitting.

kalbert312:
Closes #7492

## Changelog

Passes the default function to the translate prop callback as the 3rd parameter to use optionally as a fallback in a custom `translateWithId`. The 2nd parameter due to the usage in DataTable is for args to the translation so for the calls with no existing args, `undefined` is passed as the 2nd parameter. This should introduce no breaking changes.

### DIFF
--- a/packages/react/src/components/DataTable/DataTable.js
+++ b/packages/react/src/components/DataTable/DataTable.js
@@ -41,7 +41,7 @@ const defaultTranslations = {
   [translationKeys.unselectRow]: 'Unselect row',
 };
 
-const translateWithId = (id) => defaultTranslations[id];
+const defaultTranslateWithId = (id) => defaultTranslations[id];
 
 /**
  * Data Tables are used to represent a collection of resources, displaying a
@@ -154,7 +154,7 @@ export default class DataTable extends React.Component {
     locale: 'en',
     size: 'normal',
     overflowMenuOnHover: true,
-    translateWithId,
+    translateWithId: defaultTranslateWithId,
   };
 
   static translationKeys = Object.values(translationKeys);
@@ -248,7 +248,7 @@ export default class DataTable extends React.Component {
       : translationKeys.expandAll;
     return {
       ...rest,
-      ariaLabel: t(translationKey),
+      ariaLabel: t(translationKey, undefined, defaultTranslateWithId),
       isExpanded,
       // Compose the event handlers so we don't overwrite a consumer's `onClick`
       // handler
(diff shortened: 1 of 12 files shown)


kalbert312:
Could someone help me get `yarn test` to work locally? I ran `yarn install` in both the root and `packages/react`, yet when running tests, I received `cannot find @carbon/icons-react`. I'm using yarn v2 and I noticed the lockfiles say v1 so maybe that's why.

emyarod (commented):
would you mind updating `docs/g11n.md` or the component proptypes to describe the updates to the default function? also did you mean to reference #1 in the original ticket?

kalbert312:
Looking at `docs/g11n.md` now makes me doubt that the current version of these changes is good to implement because it advertises the `translateWithId` prop as "API compatible" with react-i18next. Users may be passing the API method directly to the prop which could break those usages. Looks like it makes more sense to just export the default translations or just use `Component.defaultProps.translateWithId` directly as a fallback.

And no on the 1 reference. I edited the issue. I meant to reference 1 in the list.

kalbert312:
Yeah I'm going

PR #7499, opened 2021-01-04
fix(Loading): increase stroke-dashoffset

### LINKED ISSUE #7497: [inline loading] audit bugs
- [ ] `stroke-dashoffset` should be set to 160, not 99. The blue active stroke should be about 50% of the spinner
   - As-is:
image
   - To-be:
image

image

emyarod:
Closes #7497

This PR adjusts the small loading spinner `stroke-dashoffset` values according to audit feedback. Not sure if the suggested value achieves the desired effect though (requires visual review), since the referenced style rules were different from the example screenshots


joshblack (approved):

aledavila (approved):

andreancardona (approved):
@emyarod looks good 🍔  !

aagonzales (approved):
Looks good!

Issue #7325, opened 2020-11-20
[dropdown] component audit bugs


laurenmrice:
Assorted bugs that were found from the component audit for dropdown, combobox and mutli-select.

## Dropdown
- [x] **Light prop:** `hover-light-ui` color token needs to be used for the hover states.
- [ ] **Warning state:** warning icon needs a `black` inner fill for all themes.
- [ ] **Invalid state:** alert icon needs a `black` inner fill for dark themes.
- [x] **Inline variant:** is not available to interact with in the playground story. 

## Combobox
- [x] **Light prop:** `hover-light-ui` color token needs to be used for the hover states of the options in the menu.
- [ ] **Warning state:** there is currently no warning state for combobox, but it should be added.
- [ ] **Invalid state:** alert icon needs a `black` inner fill for dark themes.
- [x] The top input field should not have a hover color change, only a cursor change to type.
- [x] The close icon should be `icon-01`.

## Multi-select
- [x] **Light prop:** `hover-light-ui` color token needs to be used for the hover states.
- [ ] **Warning state:** warning icon needs a `black` inner fill for all themes.
- [ ] **Invalid state:** alert icon needs a `black` inner fill for dark themes.
- [x] The entire tag right now is clickable which clear the selection. Only the close icon in the tag should clear the items selected.
- [x] Tag hover color for the close icon needs to be `inverse-hover-ui` color token.
- [x] Tag focus border needs to be around the close icon, not the entire tag. The focus color should be `inverse-focus-ui` color token.

tw15egan:
Seems like the outstanding issues are resolved with https://github.com/carbon-design-system/carbon/pull/8092

PR #7518, opened 2021-01-07
fix(ListBox): update listbox components to match spec

emyarod:
Ref #7325

This PR updates the Dropdown, Combobox, and MultiSelect components to match the component spec based on the recent component audit notes. The SVG changes will be related to the current ongoing issue with inner fills in icon SVGs (related https://github.com/carbon-design-system/carbon/issues/3216)

#### Changelog

**New**

- add inline dropdown to storybook example

**Changed**

- update menu option hover colors in light variant
- multiselect selection count tag matches Tag component spec

**Removed**

- input hover color on combobox

#### Testing / Reviewing

Confirm the 3 listbox components matches the expected appearance and behavior. Ensure that the new multiselect selection tags are accessible


joshblack (approved):

aledavila (approved):

laurenmrice (changes requested):
### Dropdown
- [x] The inline variant needs to have 24px space between the label and the input
inline dropdown spec



----

### Combobox
- [x] When the light prop is on, there is a hover color change when you hover on the top field. There should not be a hover color change. 
giph hover combo
 
- [x] The close icon should be `icon-01`
- [ ] There needs to be a warning state added in for combobox

----
### Multi-select

**Filterable:**
- [x] Needs to have 16px space before the text in the input field.
multi-select filterable spec

- [x] There is a bug when a tag appears in the input field. The tag gets cut off. 
Jan-13-2021 17-32-11

- [x] There is a hover color change when you hover on the top field. There should not be a hover color change. 

- [x] The tag hover is not the right size. Here are the specs for the tag. (This is also wrong for the regular multi-select)
multi-select tag spec

emyarod:
@laurenmrice for the last point about tag styles, these are lifted directly from the Filter Tag, so do those need to be adjusted as well?

laurenmrice:
@emyarod yes!

emyarod:
multiselect tags should now be inheriting the changes from #7664 and I will add the combobox warning state separately as it is not currently in the library (https://github.com/carbon-design-system/carbon/issues/7685)

laurenmrice (approved):
Looks good to me! 🙌🏻

PR #7519, opened 2021-01-07
feat(Tab): autogenerate IDs

### LINKED ISSUE #7511: Tab without id prop fails accessibility

- [ ] `carbon-components`

## Detailed description

The documentation does not mark the id prop as required, however if this prop is not provided then the rendered html will look like this: 
```html
<button 
  type="button" 
  role="tab" 
  aria-selected="true" 
  aria-controls="undefined__panel" 
  class="bx--tabs--scrollable__nav-link" 
  href="#" tabindex="0"
>
  asdf
</button>
```

note the `undefined__panel` as aria-control

Tab

> see changed?

Either the id prop needs to be marked as required, or if there is no id passed in then use some sort of default id.

Chrome, firefox

x.25

> should be aware of?

IBM Cloud Docs

1. Render Tabs with a Tab element without passing any props
2. Check html tree

https://codesandbox.io/s/gracious-perlman-3v6iw

## Additional information

- Screenshots or code
- Notes

emyarod:
Closes #7511

This PR autogenerates tab IDs so that the prop remains optional but the functionality does not break if a value is not provided by the user

#### Changelog

**Changed**

- autogenerate tab IDs by default

**Removed**

- remove hard coded IDs in storybook examples

#### Testing / Reviewing

Confirm that tabs function as expected when an ID is not provided


joshblack (approved):

packages/react/src/components/Tab/Tab.js
```diff
@@ -105,9 +106,11 @@ export default class Tab extends React.Component {
     onKeyDown: () => {},
   };
 
+  tabId = this.props.id || `tab-${getInstanceId()}`;
```
joshblack, line 109: One quick question, we've gotten some feedback around DOM mismatches when using auto-ids and server-side rendering. Curious if there is something we could do here, or if we're just saying if you want to SSR then you should pass in an `id` 👀 

Reference for the `useId` hook for background (totally get we can't use the hook in a class component): https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/internal/useId.js#L8

emyarod: > we're just saying if you want to SSR then you should pass in an `id`

yeah this sounds logical to me, but how have we handled it previously? the auto-id generators can be found throughout the library currently so I thought it was a viable solution here

dakahn (approved):

packages/react/src/components/Tab/Tab.js
joshblack: @emyarod sounds great to me, especially for the class components 👍

PR #7410, opened 2020-12-07
feat(Tabs): add control for autoscroll on component rerender

emyarod:
Closes #7399

This PR adds a boolean control for controlling tab autoscroll behavior on component rerenders only

#### Changelog

**New**

- `scrollIntoView` prop

#### Testing / Reviewing

Confirm that the newly added control only impacts auto scroll behavior on component rerenders, particularly from prop updates

PR #7543, opened 2021-01-12
fix(Tabs): apply `scrollIntoView` prop on mount

emyarod:
This PR is a follow up to #7410 which added a `scrollIntoView` prop to allow more granular control over autoscroll behavior. The prop value should now be respected on component mount as well as updates

#### Testing / Reviewing

Confirm that autoscroll is dictated by the prop value on component mount and on component updates


joshblack (approved):

andreancardona (approved):
@emyarod looks good 🌯

Issue #7579, opened 2021-01-15
CodeSnippet: there should be a way to set the number of rows when in closed and expanded mode


guigueb:
This request got interpreted as a way to set the maximum number of rows.
Its intention was to set the number of rows, both min and max, when closed and expanded.
I have created [#8056](https://github.com/carbon-design-system/carbon/issues/8056) for the minimum number of rows.
Please when adding props for this keep in mind they are for the maximum number of rows.
So use props like  maxClosedNumberOfRows and maxExpandeNumberOfRows.

### Summary

CodeSnippet: there should be a way to set the maximum number of rows when in closed and expanded mode

Currently in closed mode the CodeSnippet shows a maximum of 15 rows.
Currently in expanded mode the CodeSnippet shows a maximum of unlimited, it grows to fit the rows in the data.

### Justification

If the client wants to use the component in a spot where space is limited they will want/need to control the number of rows shown.
If the client does not want to show all the rows in expanded mode there is no way (other than controlling the container it is in) to control the expanded displayed rows.

### Desired UX and success metrics

When not given these are the default values to match the current values:
maxClosedNumberOfRows: 15
maxExpandedNumberOfRows: undefined (this is unlimited; which has unpredictable results - it should have a cap)

New prop maxClosedNumberOfRows - to control the maximum number of rows shown when closed:
- defaults to 15 (existing behavior)
- if < 1 then maxClosedNumberOfRows === 100%

New prop maxExpandedNumberOfRows - to control the maximum number of rows shown when expanded:
- if undefined, 100% (existing behavior, [TBD] we may want to set this)
- if <= maxClosedNumberOfRows then the number of rows shown for both closed and expanded are the same
-- the showMore/showLess buttons are not shown
-- navigation for overflows are controlled via scrolling both horizontally and vertically

`0` Zero below means any value less then 1

Testcase - closed maximum of 15, expanded maximum of unlimited (the default):
{  maxClosedNumberOfRows: (15 || undefined), maxExpandedNumberOfRows: (`0` || undefined) }
The 'Show more'/'Show less' button is available.

Testcase - closed maximum of 10, expanded maximum of unlimited:
{  maxClosedNumberOfRows: 10, maxExpandedNumberOfRows: (`0` || undefined) }
The 'Show more'/'Show less' button is available.

Testcase - closed maximum of 15, expanded maximum of 20:
{  maxClosedNumberOfRows: (15 || undefined), maxExpandedNumberOfRows: 20 }
[...]

Issue #7572, opened 2021-01-14
CodeSnippet: showMore should only be displayed when there are more then 15 rows


guigueb:
## What package(s) are you using?

- [ ] `carbon-components`
- [x] `carbon-components-react`

## Detailed description

When using <CodeSnippet type={'multi'}>{codeSnippet}</CodeSnippet>,...

If {codeSnippet} contains less than 14 rows...
- the rows are shown and the showMore is not shown
- this is good

If {codeSnippet} contains more then 15 rows...
- 15 rows and the showMore button is shown
- clicking the button shows all the rows and the show Less button
- this is good

If {codeSnippet} contains exactly 15 rows...
- 15 rows and the showMore button is shown
- clicking the button shows the same 15 rows and the show Less button
- this is not good

This shows in the latest Carbon Design System using any browser.

## Steps to reproduce the issue

1. View CodeSnippet with type='multi' in the Carbon Design System
2. Test with 14, 15, and 16 rows
3. Observe and click the showMore button

guigueb:
This bug is described in https://github.com/carbon-design-system/carbon/issues/7551

guigueb:
PR https://github.com/carbon-design-system/carbon/pull/7582

guigueb:
@tw15egan 
I'm sorry to play the new guy card...
This has a state of `waiting for maintainer response`, who is that and what type of response is required? 
Actually, is there a flowchart of the Carbon approval process?
How can I tell what stages there are, where I am in the process, and who to contact when in those stages?
Maybe I should post this in a slack chat...

guigueb:
PR merged - this issue is resolved.
On tohttps://github.com/carbon-design-system/carbon/issues/7579

guigueb:
PR merged - this issue is resolved.
On to https://github.com//issues/7579

PR #7582, opened 2021-01-15
fix(react CodeSnippet): only display showMore if more then 15 rows

guigueb:
Closes # https://github.com/carbon-design-system/carbon/issues/7572

Only display showMore Button if there are more than 15 rows of data being displayed.

#### Changelog

**Changed**

Replaced hard coded px count with a calculation.
The height of the closed CodeSnippet = 15 rows * 18xp per row.

Hopefully, in the near future this calculation can be based on a user set closed row prop.
See: https://github.com/carbon-design-system/carbon/issues/7579

#### Testing / Reviewing

1. In the Carbon Design System storybook create a CodeSnippet with type='multi'
2. Test the CodeSnippet with 14, 15, and 16 rows of data (ie: its children)
3. Observe the showMore button

The showMore button should only show when there are more then 15 rows of data.


guigueb:
I made this PR but need help with the unit test I included.
The 'when more than 15 rows' test fails.
If I modify the story with > 15 rows of data and include a `expect(wrapper).toMatchSnapshot();` before line 97 the snapshot shows there are > 15 row, but there is no showMore Button in the snapshot.

tw15egan:
This makes sense to me. I also like the idea of being able to pass in a number to the component to be able to have the `show more` button display based on the apps requirement, instead of always being 15 rows.

Thoughts @carbon-design-system/design @carbon-design-system/developers ?

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
```diff
@@ -63,4 +64,36 @@ describe('Code Snippet', () => {
...
+
+    it('when exactly 15 rows', () => {
+      wrapper.setProps({
+        children: '1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15',
+      });
+
+      expect(wrapper.find(Button).length).toBe(0);
+    });
+
+    it.skip('when more then 15 rows', () => {
```
joshblack, line 96:
```suggestion
    it('when more then 15 rows', () => {
```

joshblack (commented):
Looks great! Thanks for adding this PR! Just left a small note 👀

laurenmrice:
@tw15egan That makes sense to me. This actually came up in a support question recently and I can add it to our usage guidance.

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
guigueb: I would love to have this unit test un-skipped.
However, when I do so it fails, and I am not sure why.

If I log the snapshot expect(wrapper).toMatchSnapshot() - the snapshot has 16 rows of data but the button is not there.
I can force the button render in CodeSnippet.js and the test will work - line 182: `{(true || shouldShowMoreLessBtn) && (`.

I wrote the testcase using (setProps(children)) - thinking I need to invoke a resize, but it still doesn't work.
Then, I debugged the setShouldShowMoreLessBtn react hook in onResize, it gets hit via storybook, but not running the unit tests.

So, I'm at a loss how to trigger the setShouldShowMoreLessBtn react hook to get the button to render.

That said, the tests did not exist before - so the testing isn't getting any worse.
I will need to figure out hook testing for #7579 which will change the closed/expanded size of the CodeSnippet.

tw15egan:
@guigueb it looks like design has signed off on the proposal to add a new prop, perhaps `numberOfRowsClosed` and `numberOfRowsOpen`, to allow an end-user to customize when the `show more` button should display based on their applications needs, and then the height of the `expanded` CodeSnippet which would allow vertical scrolling.

Do you want to update this PR to accommodate that? It seems like that would resolve both #7572 and #7579

guigueb:
If you wouldn't mind, even though this is a very simple change and the addressing #7579 would address this one, I'd like to deliver in two parts.
A few reasons...
- this was a quick shot - my first contribution - to get my feet wet, I'd kinda like to complete it
- the next change will cross packages, both the react and components need changes
- I will be doing the change but I'm not 100% sure of my timeframe

I'm hoping - once I'm a contributor - I can take ownership of the other CodeSnippet issues I logged.

tw15egan:
@guigueb sure thing! We'll keep them separated to make it easier and get your first contribution into the library 🙂

tw15egan (commented):
Looks great! One question for @laurenmrice : do we want to remove the bottom padding if the `show more` button isn't present? 
Screen Shot 2021-01-25 at 1 39 35 PM

laurenmrice:
@tw15egan yes, we should remove the extra bottom padding if the `show more` button is not present. It should just have a 16px bottom padding to match the other sides. 👍🏻

guigueb:
About - "do we want to remove the bottom padding if the show more button isn't present?"
I was - and will - going bring this up when I do #7579.

The 16px padding on all four sides that Lauren is talking about is applied to the outer most CodeSnippet `<div>` and is always there.

The padding TJ is talking about is applied to the `<pre>` tag within the `<code>` tag all the time.
I think its goal was to leave space to not jumble the codeSnippet data with the copy and more/less buttons.
But as the `<code>` tag is fully sized - with its overflow contents, including the padding, not shown - this will not work.

If the goal is to not crowd the buttons then maybe the `<div>` containing `<code>` should be padded,
or, better yet, position the snippet-container `<div>` to not overlap the buttons.
_In my opinion the best approach would be to removing the `<pre>` padding altogether and have the codeSnippet contents render(hidden) under the buttons._

Let us address this in #7579.

This image shows the 16px CodeSnippet border with the `<pre>` padding removed.

image

guigueb:
@tw15egan 
@joshblack 

Hey folks, I'm not sure the procedures here.
Am I missing an action on this?

tw15egan:
@laurenmrice do you want the style changes made here in this PR, or would a follow-up PR be acceptable?

guigueb:
The style changes should be made in the follow-up PR #7579.
The `<pre>` padding will be better understood when we start to adjust the closed/expanded sizes.
Also the ability to not show the More button is part of setting the closed/expanded sizes.

laurenmrice:
@guigueb @tw15egan Yes, I am fine with having the style changes addressed in the follow up PR.

tw15egan (approved):

guigueb:
@tw15egan
@joshblack
@dakahn 
I feel it is so close to being able to merge.
It looks to be needing the mergeConflict tag lifted... and a 2nd approver?
It will be nice to not have to merge the contributors file in the future PRs :-)

laurenmrice (approved):

dakahn (approved):

Issue #7571, opened 2021-01-14
A "hideLabel" prop for Dropdowns (specifically inline Dropdowns)


molyholy:
## Environment

> Browser
Firefox

> Assistive technology used to verify
IBM Equal Access Accessibility Checker

## Detailed description

> What WCAG 2.1 checkpoint does the issue violate?
IBM CI 162 Checkpoint Violation: 1.3.1 Info and Relationships
https://www.ibm.com/able/guidelines/ci162/info_and_relationships.html
https://www.ibm.com/able/guidelines/ci162/name_role_value.html

## Steps to reproduce the issue

1. Log into https://test.cloud.ibm.com/billing-nouveau/invoices
2. Run the Accessibility Checker

## Additional information

- Screenshots or code
<img width="1790" alt="image" src="https://user-images.githubusercontent.com/23204596/104644569-e8b3dd80-567b-11eb-83d3-2d79438ae006.png">

- Notes
- The inline Dropdown in the top right does not have a "Label" component, which creates an a11y issue. However, adding a label seems to defeat the purpose of the inline Dropdown (and if nothing else, looks kind of weird). The Select component already has a hideLabel prop which hides the label visually but not programatically, and I think the inline Dropdown could really benefit from something similar.

@dakahn @ryanomackey

dakahn:
I'm cool with this. We have these on various Textinputs and the Select component.

PR #7594, opened 2021-01-19
feat(Dropdown): add hideTitleText prop

molyholy1:
Closes #

Addresses this issue https://github.com/carbon-design-system/carbon/issues/7571

#### Changelog

**New**

- Add a `hideTitleText` prop for the Dropdown. The `titleText` will still exist programatically but hiding it visually lends itself to smoother design (especially with inline Dropdowns). This sort of prop also exists in the Select component.

#### Testing / Reviewing
1. Start up the React Storybook and navigate to the Dropdown component
2. In the code, pass in the `hideTitleText` prop
3. (Optional but highly recommended) Run the [IBM Equal Accessibility Checker](https://www.ibm.com/able/toolkit/) with `hideTitleText` being true. There should not be an error for "The WAI-ARIA property must reference a non-empty unique id of an existing element that is visible"  which normally occurs when the `titleText` is missing entirely. However, because we have hidden it visually but not programatically, the error shouldn't appear.


emyarod (commented):
would you mind renaming the prop to `hideLabel` to match the other Carbon components? also it looks like you will need to update snapshots in order to pass CI

edit: yeah it looks like only the listbox components use "title" instead of "label" but this may need to be updated in the future to match the existing components like TextInput, Select, etc.

brawl1111:
> would you mind renaming the prop to `hideLabel` to match the other Carbon components? also it looks like you will need to update snapshots in order to pass CI

Yep! I only changed it to `hideLabel` cuz there isn't a `label` prop for Dropdowns but yeah I can change it 👍

tw15egan (approved):
Not seeing any a11y errors with the `hideLabel` prop enabled. Thanks for contributing! LGTM 👍 ✅

kingtraceyj (approved):
looks good to me

emyarod (approved):

PR #7609, opened 2021-01-20
fix(ComposedModal): add accessible name and aria-label prop

### LINKED ISSUE #6543: Carbon Compose Modal accessibility violation
## Detailed description

experiencing same/similar issue that was in ticket https://github.com/carbon-design-system/carbon/issues/4587. `An interactive element/widget must have an accessible name.` violation is thrown on top level of compose modal component. can be seen in both sample application below and in story book here https://react.carbondesignsystem.com/?path=/story/composedmodal--example-usage-with-trigger-button

 - versions, also found in codesandbox: 
    "carbon-icons": "7.0.7",
    "react": "16.3.1",
    "react-dom": "16.3.1",

- no accessibility violations

https://aat.w3ibm.mybluemix.net/token/4950d599-abf0-4c72-b249-9eed7e3bf02e/82ec2f72-e53b-4c22-98b3-faab6bd311db/archives/2020JanDeploy//doc/w3/help/en-US/idhi_accessibility_check_g1156.html

https://codesandbox.io/s/carbon-components-react-du1ip?file=/src/index.js
## Additional information

Screen Shot 2020-07-24 at 9 21 44 AM

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6543

Fixes a11y issues with `ComposedModal`

#### Changelog

**New**

- `ariaLabel` prop for `ComposedModal`
-  `aria-modal` attribute added

**Changed**

- `ModalHeader` now uses `h2` and `h3` instead of `p` elements

#### Testing / Reviewing

Ensure there are no a11y violations with Composed Modal


packages/react/src/components/ComposedModal/ComposedModal.js
```diff
@@ -33,6 +34,11 @@ export default class ComposedModal extends Component {
   endSentinel = React.createRef();
 
   static propTypes = {
+    /**
+     * Specify the aria-label for bx--modal-container
+     */
+    ariaLabel: PropTypes.string,
```
joshblack: This is definitely a point where the codebase is split, but would you want to do `aria-label` if doing the DOM attribute and `label` if we're setting it automatically for folks?

In other words, if someone is explicitly setting an ARIA label then they use `aria-label` as the prop.

packages/react/src/components/ComposedModal/ComposedModal.js
```diff
@@ -22,6 +22,7 @@ export default class ComposedModal extends Component {
   state = {};
 
   static defaultProps = {
+    ariaLabel: 'modal-container',
```
joshblack: Definitely a bummer that we can't make this required without breaking people, would there be a good prop that we could derive a meaningful label from, potentially?

tw15egan: I originally didn't think we'd be able to, but after taking a deeper look I've modified the code a bit so that it will only generate one if a user does not specify their own `aria-label` and will be based on the `label`, matching `Modal` functionality

tw15egan:
@joshblack updated!

dakahn (approved):
Great work. Reads well in testing with NVDA on Firefox latest Windows 10

tw15egan:
@joshblack let me know if this solution works 👍

joshblack (approved):

PR #7612, opened 2021-01-20
fix(TableToolbarSearch): ensure search cannot be expanded when disabled

### LINKED ISSUE #7522: TableToolbarSearch isDisabled doesn't disable the expand

- [x] `carbon-components`

## Detailed description

When `disabled` is true in TableToolbarSearch, the icon button is still clickable and it will still expand the Search, although the Search is disabled

TableToolbarSearch

> see changed?

I expect the icon button for searching to be disabled as well

Chrome

```
"carbon-components": "10.17.0",
"carbon-components-react": "7.17.0",
```

> should be aware of?

IoT Addons. No specific dates

1. Set `disabled` to true in TableToolbarSearch
2. See that the icon button is still clickable

https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-tablecard--editable

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7522

Disabled styles will now prevent the search from expanding if `disabled` is passed to `TableToolbarSearch`

#### Changelog

**New**

- Disabled styles for non-expanded disabled Search button
- Logic to prevent expansion if `disabled` is passed to `TableToolbarSearch`
- Removes `TableToolbarSearch` from the tab order if it is disabled

#### Testing / Reviewing

Pass `disabled` to `TableToolbarSearch` in the `DataTable-filtering.js` file and make sure the search is not able to interact with

### DIFF
--- a/packages/components/src/components/data-table/_data-table-action.scss
+++ b/packages/components/src/components/data-table/_data-table-action.scss
@@ -89,7 +89,7 @@
     .#{$prefix}--search-magnifier {
     background: $disabled-01;
     cursor: not-allowed;
-    transition: none;
+    transition: background-color none;
   }
 
   .#{$prefix}--toolbar-search-container-expandable
@@ -113,7 +113,7 @@
     .#{$prefix}--search-magnifier:hover {
     background: $disabled-01;
     cursor: not-allowed;
-    transition: none;
+    transition: background-color none;
   }
 
   .#{$prefix}--toolbar-search-container-expandable
--- a/packages/components/src/components/data-table/_data-table-action.scss
+++ b/packages/components/src/components/data-table/_data-table-action.scss
@@ -85,6 +85,14 @@
   }
 
   .#{$prefix}--toolbar-search-container-expandable
+    .#{$prefix}--search--disabled
+    .#{$prefix}--search-magnifier {
+    background: $disabled-01;
+    cursor: not-allowed;
+    transition: none;
+  }
+
+  .#{$prefix}--toolbar-search-container-expandable
     .#{$prefix}--search
     .#{$prefix}--search-magnifier:focus {
     @include focus-outline('outline');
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Search/Search.js
```diff
@@ -188,7 +188,10 @@ export default class Search extends Component {
 
     return (
       <div role="search" aria-labelledby={searchId} className={searchClasses}>
-        <Search16 className={`${prefix}--search-magnifier`} />
+        <Search16
+          disabled={disabled}
```
emyarod: does the SVG need `disabled=""` on it?

tw15egan: Nope! Just removed it

packages/components/src/components/data-table/_data-table-action.scss
```diff
@@ -84,6 +84,14 @@
     pointer-events: all;
   }
 
+  .#{$prefix}--toolbar-search-container-expandable
+    .#{$prefix}--search--disabled
+    .#{$prefix}--search-magnifier {
+    background: $disabled-01;
```
emyarod, line 90:
```suggestion
    background-color: $disabled-01;
```

to avoid changing more properties than we need to

joshblack (approved):

tw15egan:
@emyarod let me know if there are any other issues with this 👍

packages/components/src/components/data-table/_data-table-action.scss
```diff
@@ -96,6 +104,18 @@
...
   }
 
+  .#{$prefix}--toolbar-action.#{$prefix}--toolbar-search-container-disabled {
+    cursor: not-allowed;
+  }
+
+  .#{$prefix}--toolbar-search-container-expandable
+    .#{$prefix}--search--disabled
+    .#{$prefix}--search-magnifier:hover {
+    background: $disabled-01;
```
emyarod, line 114:
```suggestion
    background-color: $disabled-01;
```

packages/components/src/components/data-table/_data-table-action.scss
```diff
@@ -579,6 +599,10 @@
     }
   }
 
+  .#{$prefix}--search--disabled .#{$prefix}--search-magnifier:hover {
+    background: transparent;
```
emyarod, line 603:
```suggestion
    background-color: transparent;
```

tw15egan:
@emyarod updated

emyarod (approved):
should we do the same for the transition overrides as well? otherwise looks good to me

PR #7621, opened 2021-01-21
fix(CodeSnippet): use field-01 instead of internal variable

### LINKED ISSUE #6994: Internal Code snippet variable is causing issues with custom themes
Based on convo in Slack https://ibm-studios.slack.com/archives/C046Y0YUD/p1601663658098800

For some reason, when a user tried to set a custom `$ui-01` using css custom properties, the value was not updated. We use an internal token in the code snippet mixin, which is set to `$ui-01` by default. The inline code snippet updated properly but does not use the internal mixin. 

Wondering if we should just update the mixin the use `$field-01` directly (the default), and not `$snippet-background-color;`, which is not used anywhere else. 

https://github.com/carbon-design-system/carbon/blob/ad4d9a2981b31170b4fd2320e61bf3e4e26ca7d3/packages/components/src/components/code-snippet/_mixins.scss#L19

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6994

DIscontinues use of `$snippet-background-color` and instead uses `$field-01` directly, to prevent issues when users try to use custom themes.

#### Changelog

**Changed**

- `$snippet-background-color` becomes `$field-01`
- Deprecates `$snippet-background-color`

#### Testing / Reviewing

Ensure there are no style issues with `CodeSnippet`


joshblack (approved):

andreancardona (approved):
@tw15egan looks good!

PR #7635, opened 2021-01-22
fix(elements): add back in missing exports

joshblack:
The `@carbon/elements` package is missing some exports (namely layout and spacing) and this PR adds them back in.

#### Changelog

**New**

**Changed**

- Add back-in missing exports from elements package

**Removed**


andreancardona (approved):
🙌🏽

aledavila (approved):

PR #7646, opened 2021-01-25
fix(SideNavLink): remove fixed height on SideNavLink

### LINKED ISSUE #7195: Sidebar spacing changes when collapsed and hovered over

- [ ] `carbon-components`

## Detailed description

When the sidebar collapses and is hovered over. the spacing changes and makes the icons move about.

Yes, `SideNav` and `SideNavItems`, `SideNavLink`

> see changed?

Stop the sidebar spacing changing on hover.

Safari and Chrome on macOS.

Latest.

> should be aware of?

Needs to be fixed for a Q1 2021 release but ideally before the end of Q4 2020.

1. Create SideNav with Items inside
2. Make it collapsible
3. Move mouse into sidebar area and notice how the icons move as their spacing changes.

https://codesandbox.io/s/wizardly-thunder-y0hl5?file=/src/index.js

## Additional information

In Slack, Andrew Wang suggested the issue seems to be that it looks like there’s an inline height and width style being set on li s when the nav is hovered over.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7195

Removes fixed height on `SideNavLink` so that there is no jitter when a Side Nav item is hovered over ([see here](https://codesandbox.io/s/small-hill-yxeie?file=/src/index.js)) 

#### Changelog

**Changed**

- Switched from a fixed height/width to auto

**Removed**

- Outdated styles

#### Testing / Reviewing

Ensure SideNav conforms to specs detailed here: https://www.carbondesignsystem.com/components/UI-shell-left-panel/style


dakahn (approved):

emyarod (approved):
looks good to me

Issue #7331, opened 2020-11-20
Combobox: `role="combobox"` attribute is on the wrong element causing Accessibility Checker error


dakahn:
background https://github.com/IBMa/equal-access/issues/282#issue-747729478

Our Combobox has a wrapper div with the attribute `role="combobox"`. This attribute should be on the input itself. This is causing the below AC violation. 
![image](https://user-images.githubusercontent.com/40970507/99859640-39b0c500-2b56-11eb-9624-4b85bb6edb68.png)

Sort of a weird error but essentially since AC expects -- as per WCAG recommendation -- for the combobox role to be on the input and it expects that input to receive focus first in the DOM AND instead something else gets focus -- we get this error. 

Moving that attribute to the input itself should fix this issue.

JensenAlexander:
Is there a workaround for this?

tw15egan:
Seem to also be related to issues seen in `Filterable Multiselect` when the menu is open and items are selected (#7296 and #6570)

- Initial DOM focus on the combobox is not set on the text input when the combobox receives focus
- The element using 'aria-activedescendant' set to "carbon-multiselect-example-item-1" does not have its 'tabindex' attribute value set to 0 or -1

dakahn:
closing as a dupe of https://github.com/carbon-design-system/carbon/issues/6938

PR #7659, opened 2021-01-26
fix(combobox): remove extra or unnecessary ARIA and refactor interactive elements

### LINKED ISSUE #6938: Fix Combobox aria roles to match aria 1.2 spec.
The Combobox currently has strange aria roles, especially where the `<input>` is inside a `role=button`. 

The template is also outdated in the sense that it has the `role=combobox` on a wrapper node rather than the `<input>`.  As of the Aria 1.2 spec (https://w3c.github.io/aria-practices/#examples-2), the `role=combobox` is back on the `<input>`.

So, it should follow the patterns in https://w3c.github.io/aria-practices/#examples-2.   Note that the pattern is a bit different depending on whether or not the combobox is filterable.

joshblack:
Closes #6938

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


dakahn:
related and potentially could close #7331

joshblack:
Superseded by https://github.com/carbon-design-system/carbon/pull/7777

PR #7664, opened 2021-01-27
fix(Tag): update filter tags to match updated spec

### LINKED ISSUE #7630: [tag] fix hover and focus for filterable tag
- The current hover and focus states are slightly off for our filterable tag component.
- Since we are also adding a `small` tag variation, we need to make an adjustment to these states so they are consistent across both sizes. 

Artboard Copy 4

------
Artboard Copy 5

emyarod:
Closes #7630
related #7619

This PR updates the filter tag and the new small filter tags to match the updated component spec. Unblocks #7518

#### Testing / Reviewing

Ensure the filter tag variants match the component spec

### DIFF
--- a/packages/components/src/components/tag/_tag.scss
+++ b/packages/components/src/components/tag/_tag.scss
@@ -230,7 +230,7 @@
   }
 
   .#{$prefix}--tag--sm.#{$prefix}--tag--filter {
-    padding-right: rem(2px);
+    padding-right: 0;
   }
 
   // Skeleton state
--- a/packages/components/src/components/tag/_tag.scss
+++ b/packages/components/src/components/tag/_tag.scss
@@ -147,7 +147,9 @@
 
   // tags used for filtering
   .#{$prefix}--tag--filter {
-    padding-right: rem(2px);
+    padding-top: 0;
+    padding-right: 0;
+    padding-bottom: 0;
     cursor: pointer;
 
     &:hover {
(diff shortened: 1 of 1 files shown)


laurenmrice (approved):
Looks great! Thank you 🙌🏻

packages/components/src/components/tag/_tag.scss
```diff
@@ -195,7 +200,7 @@
   .#{$prefix}--tag__close-icon:focus {
     border-radius: 50%;
     outline: none;
-    box-shadow: inset 0 0 0 2px $focus;
+    box-shadow: inset 0 0 0 1px $focus;
```
tw15egan, line 204: Do we also need to change this on L214?

packages/components/src/components/tag/_tag.scss
```diff
@@ -156,10 +158,13 @@
...
+    display: flex;
     flex-shrink: 0;
-    width: rem(16px);
-    height: rem(16px);
-    margin: 0 0 0 rem(4px);
+    align-items: center;
+    justify-content: center;
+    width: rem(24px);
+    height: rem(24px);
+    margin: 0 0 0 rem(2px);
```
tw15egan, line 167: Can we add in our transition variables here?

```scss
transition: background-color $duration--fast-01 motion(standard, productive),
      box-shadow $duration--fast-01 motion(standard, productive);
```

packages/components/src/components/tag/_tag.scss
emyarod: I kept it since it was high contrast specific but I'll reduce it for consistency

tw15egan: Yeah just wasn't sure if it was intentional or not, just wanted to bring it up 👍

tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

PR #7666, opened 2021-01-27
fix(DatePicker): add more locales to valid locales

tw15egan:
Adds more locales to the list of valid locales from Flatpickr

#### Changelog

**New**

- A few new locales have been added to the `locale` PropType

#### Testing / Reviewing

Ensure the new `locale` values can be used without emitting a warning


guigueb:
Hello there, in our locales we also have `kk`, `nb`, `pt-br`, `zh-cn`, and `zh-tw`.
I see `zh-tw` has been added.
I'm checking if we map `zh-cn` to `zh`.
And I'm checking into the others as well, but as they don't show in flatpickr I'm not sure what we'll do if they are languages we still support.

emyarod (approved):

joshblack (approved):

PR #7667, opened 2021-01-27
fix(breadcrumb): fix hover bug

### LINKED ISSUE #7016: [breadcrumb] hover state bug

aledavila:
Closes #7016 

Thanks @tw15egan from his other PR

#### Changelog

**New**

- added hover to visited

#### Testing / Reviewing

Primary bug was on firefox. make sure to check it works fine on all themes and it doesn't flash. (check issue for visual)

### DIFF
--- a/packages/components/src/components/breadcrumb/_breadcrumb.scss
+++ b/packages/components/src/components/breadcrumb/_breadcrumb.scss
@@ -37,6 +37,10 @@
 
   .#{$prefix}--breadcrumb-item .#{$prefix}--link:visited {
     color: $link-01;
+
+    &:hover {
+      color: $hover-primary-text;
+    }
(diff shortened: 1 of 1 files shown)


tw15egan (approved):
LGTM 👍 ✅

laurenmrice (approved):
Looks good! 🙌🏻

PR #7677, opened 2021-01-27
Update TooltipDefinition.js

harshitshah436:
To support accessibility by adding a title or area-label

Closes #

This will resolve accessibility defect related to the tooltip button has no labels associated with hit.

#### Changelog

**New**

- Add ` aria-labelledby` property.

**Changed**

- It's an addition.

#### Testing / Reviewing

- Can verify if the button component has an additional property called `area-labelledby` with the same id.

#### Sample accessibility error for the tooltip component

tooltip-definition_error

### DIFF
--- a/packages/react/src/components/TooltipDefinition/TooltipDefinition.js
+++ b/packages/react/src/components/TooltipDefinition/TooltipDefinition.js
@@ -77,6 +77,7 @@ const TooltipDefinition = ({
         type="button"
         className={tooltipTriggerClasses}
         aria-describedby={tooltipId}
+        aria-labelledby={tooltipId}
         onFocus={composeEventHandlers([onFocus, handleFocus])}>
         {children}
       </button>


tw15egan:
Hey there @harshitshah436, is there an issue associated with this PR? I ran the accessibility checker and I'm not seeing any violations

Screen Shot 2021-01-27 at 12 41 54 PM

harshitshah436:
@tw15egan Not sure why it's not showing. But we define the tooltip as below and getting the accessibility error as attached in the issue description screenshot.

```
<TooltipDefinition
        id={i18next.t("common:accessibility_labels.tooltip_definition")}
        direction={tooltipDirection}
        align="start"
        onClick={e => {
          e.preventDefault();
        }}
        tooltipText={tooltipText}
      >
        {this.props.tooltipIcon ? this.props.tooltipIcon : tooltipIcon}
      </TooltipDefinition>
```

tw15egan:
In the source code, it looks like the `aria-describedby={tooltipId}` is generated by the `id` prop. Can you verify in your rendered definition tooltip that the `aria-describedby` value is being properly set by the supplied `id`? I think that value should fix the violation, but wondering if this is an issue with the i18n functionality

harshitshah436:
> In the source code, it looks like the `aria-describedby={tooltipId}` is generated by the `id` prop. Can you verify in your rendered definition tooltip that the `aria-describedby` value is being properly set by the supplied `id`? I think that value should fix the violation, but wondering if this is an issue with the i18n functionality

that's parsed correctly. Here's the attached screenshot. But when we run accessibility test, it complaints about this button.
profile-tooltip

dakahn:
@harshitshah436 can you open a ticket for this issue and link it in this PR. we need to triage and verify this problem for ourselves. In the issue template you'll find a link to a Codepen where you can recreate so we can verify and then we can move forward here.

harshitshah436:
Here's the ticket for this issue: https://github.com/carbon-design-system/carbon/issues/7688

dakahn:
Since we can't reproduce on the issue are we good to close this PR, friend?

harshitshah436:
yeah, I think there's something wrong with the electron browser.

PR #7679, opened 2021-01-27
fix(Tag): fix issues with skeleton in safari

### LINKED ISSUE #7672: Component Tag Skeleton does not work well in Safari.
I was testing my project in Safari and the Skeleton loading of Tag seems a bit different than in Firefox or Chrome.
My Safari Version is Versão 13.1.2 

    "@carbon/icons-react": "^10.9.1",
    "@carbon/themes": "^10.10.1", 
    "carbon-components": "^10.23.1",
    "carbon-components-react": "^7.23.1", 

1. I add the tag component to my code like this:
`<div className="bx--col-md-3 bx--col-lg-4 bx--col-lg-3 bx--p-b-1 bx--p-t-1">
            <TagSkeleton className="bx--tag-sev" /> <TagSkeleton />
  </div>`
2. Then check in Firefox and Safari 
 
## Additional information
 
image (2)

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7672

Fixes an issue with how Safari handles overflow content and pseudoelements. 

#### Changelog

**New**

- Safari specific media query to fix `overflow: hidden` not being respected on the Tag Skeleton

**Changed**

- The tag skeleton background colors were being overridden by `$ui-03` and causing issues on the dark theme

#### Testing / Reviewing

Ensure the tag skeleton renders correctly in Safari / Chrome / Firefox in all themes


emyarod (approved):

aledavila (commented):
Hey looks good so far. I was wondering if the skeleton on g100 is supposed to look this dark 

https://user-images.githubusercontent.com/12533409/106184115-bd92b900-6166-11eb-9b8c-9611e6c34c37.mov

tw15egan:
@aledavila I updated it to use `skeleton-01` which is a little bit better, but the `ui-03` had an even more imperceptible change in color. @carbon-design-system/design any idea if we need to update the `skeleton` tokens in `g90`/`g100`?

aledavila (commented):
I'll approve this since an issue has been opened to investigate tokens

Yay thanks TJ 🎉

aledavila (approved):

aagonzales:
I think you're right it needs to be a different color. The `skeleton-02` in the gray 100 theme should use `#525252` Gray70

aagonzales (approved):

PR #7683, opened 2021-01-28
feat: add textDir property to FormLabel

### LINKED ISSUE #7264: BIDI - Support base text direction

wajnberg:
Ref #7264

Add the textDir property for FormLabel. This is just a sample. This property should be applied to all relevant react components

#### Changelog

**New**
bidiUtils

**Changed**

FormLabel

#### Testing / Reviewing

Add the new story WithTextDir in the FormLabel stroy

### DIFF
--- a/packages/react/src/components/FormLabel/FormLabel.js
+++ b/packages/react/src/components/FormLabel/FormLabel.js
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2018
+ * Copyright IBM Corp. 2016, 2021
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -9,14 +9,19 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
 import { settings } from 'carbon-components';
+import resolveBaseTextDir from '../../tools/bidiUtils';
 
 const { prefix } = settings;
 
-const FormLabel = ({ className, children, id, ...other }) => {
+const FormLabel = ({ className, children, id, dir, ...other }) => {
   const classNames = classnames(`${prefix}--label`, className);
 
   return (
-    <label htmlFor={id} className={classNames} {...other}>
+    <label
+      htmlFor={id}
+      className={classNames}
+      dir={resolveBaseTextDir(children, dir, false)}
+      {...other}>
       {children}
     </label>
   );
@@ -32,6 +37,11 @@ FormLabel.propTypes = {
    * Provide a custom className to be applied to the containing <label> node
    */
   className: PropTypes.string,
+
+  /**
+   * Provide the text direction for the given <FormLabel>
+   */
+  dir: PropTypes.oneOf(['', 'ltr', 'rtl', 'auto']),
 
   /**
    * Provide a unique id for the given <FormLabel>
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -304,6 +304,17 @@ Map {
       "defaultChecked": Object {
         "type": "bool",
       },
+      "dir": Object {
+        "args": Array [
+          Array [
+            "",
+            "ltr",
+            "rtl",
+            "auto",
+          ],
+        ],
+        "type": "oneOf",
+      },
       "disabled": Object {
         "type": "bool",
       },
@@ -2836,6 +2847,17 @@ Map {
       },
       "className": Object {
         "type": "string",
+      },
+      "dir": Object {
+        "args": Array [
+          Array [
+            "",
+            "ltr",
+            "rtl",
+            "auto",
+          ],
+        ],
+        "type": "oneOf",
       },
       "id": Object {
         "type": "string",
--- a/packages/react/src/components/Checkbox/Checkbox-story.js
+++ b/packages/react/src/components/Checkbox/Checkbox-story.js
@@ -74,3 +74,12 @@ export const playground = () => (
     <Checkbox {...props()} id="checkbox-label-2" />
   </fieldset>
 );
+
+export const checkboxWithTextDir = () => {
+  return (
+    <fieldset className={`${prefix}--fieldset`}>
+      <legend className={`${prefix}--label`}>Checkbox with textDir</legend>
+      <Checkbox labelText={`שלום Hello !!`} dir="auto" id="checkbox-label-1" />
+    </fieldset>
+  );
+};
--- a/packages/react/src/components/Checkbox/Checkbox.js
+++ b/packages/react/src/components/Checkbox/Checkbox.js
@@ -9,6 +9,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
+import resolveBaseTextDir from '../../tools/bidiUtils';
 
 const { prefix } = settings;
 
@@ -22,6 +23,7 @@ const Checkbox = React.forwardRef(function Checkbox(
     hideLabel,
     wrapperClassName,
     title = '',
+    dir,
     ...other
   },
   ref
@@ -58,7 +60,11 @@ const Checkbox = React.forwardRef(function Checkbox(
         }}
       />
       <label htmlFor={id} className={labelClasses} title={title || null}>
-        <span className={innerLabelClasses}>{labelText}</span>
+        <span
+          className={innerLabelClasses}
+          dir={resolveBaseTextDir(labelText, dir, false)}>
+          {labelText}
+        </span>
       </label>
     </div>
   );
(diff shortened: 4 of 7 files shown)


wajnberg:
@sadir@il.ibm.com. FYI

packages/react/src/components/FormLabel/FormLabel.js
```diff
@@ -9,14 +9,19 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
 import { settings } from 'carbon-components';
+import resolveBaseTextDir from '../../tools/bidiUtils';
 
 const { prefix } = settings;
 
-const FormLabel = ({ className, children, id, ...other }) => {
+const FormLabel = ({ className, children, id, textDir, ...other }) => {
```
joshblack: Would it be possible for users to use the `dir` prop directly, e.g. 

```jsx
<FormLabel dir={resolveBaseTextDir(text)} />
```

Instead of adding in a `textDir` prop?

Also, was curious when would someone apply the `dir` attribute to specific components versus regions of the page. Would love any insights that you could share!

joshblack (commented):
Thanks so much for putting this together! Left a quick question 👀

packages/react/src/components/FormLabel/FormLabel.js
wajnberg: OK I fixed this

wajnberg:
Hello @joshblack 

Could you please review again ?

Thank you

joshblack:
@wajnberg so sorry for the delay!

Taking a look at the PR, something I'm curious about is if what is being shared is able to be accomplished in user-land versus needing to be embedded in the components. In the case of `FormLabel`, it seems like it can accept the `dir` prop and the `resolveBaseTextDir` could be the helper used to supply that instead of the component including this behavior.

For example:

```jsx
<FormLabel dir={resolveBaseTextDir(text, dir, false)}>Label text</ForMLabel>
```

wajnberg:
Hello @joshblack ,

For the `FormLabel` sample you are right. One can write
<FormLabel dir={resolveBaseTextDir(text, dir, false)}>Label text</ForMLabel>
and get the expected result
although I personally think that the end user doesn't have to deal with calling methods such `resolveBaseTextDir`. I think that the syntax for the end user should be as simple as possible

But for `Checkbox` the solution you propose is not possible.
Actually if you write `<Checkbox labelText={`שלום Hello !!`} dir="rtl" id="checkbox-label-1" /> you stiil get
Carbon1

Hence the need to add the textDir support directly in the component as I propose in this PR. With this change you get the expected result:
Carbon2

tw15egan:
@wajnberg I'd ensure you are running the storybook with `rtlcss` compilation mode enabled as well (`export CARBON_REACT_STORYBOOK_USE_RTL=true` and then `yarn storybook`). Testing this, I get this behavior when viewing the checkbox, which differs from your screenshot:

Screen Shot 2021-02-11 at 12 05 39 PM

wajnberg:
Hello @tw15egan 

I think there is a confusion here. You are talking about `mirroring` whereas I am talking about text direction which are 2 different things.

Of course when you enable CARBON_REACT_STORYBOOK_USE_RTL you get all the GUI mirrored and therefore the text direction is RTL.

But I want to get text direction = RTL even the GUI is not mirrored. 

I hope I was clear

tw15egan:
Is it a typical use-case to have a non-mirrored UI but have text render in RTL mode? I was under the assumption that mirroring and text direction went hand in hand, which I seem to have assumed incorrectly.

Is this something that would need to be added to every single node of text that is rendered via the components?

joshblack:
@wajnberg for the checkbox example, I think it's a case of additional props being based onto the underlying `<input>` and therefore the `<label>` is not a descendant of a parent that has `dir="rtl"` set. We could definitely allow folks to pass in a `dir` prop and set it on the outer `<div>` 👍

wajnberg:
> 
> 
> Is it a typical use-case to have a non-mirrored UI but have text render in RTL mode? I was under the assumption that mirroring and text direction went hand in hand, which I seem to have assumed incorrectly.
> 
> Is this something that would need to be added to every single node of text that is rendered via the components?

Actually text direction and GUI mirroring are 2 independent features. One typical case is an Hebrew user typing/viewing Hebrew words in a LTR GUI (English GUI). This user would like to see these Hebrew words in RTL direction

wajnberg:
> 
> 
> @wajnberg for the checkbox example, I think it's a case of additional props being based onto the underlying `<input>` and therefore the `<label>` is not a descendant of a parent that has `dir="rtl"` set. We could definitely allow folks to pass in a `dir` prop and set it on the outer `<div>` 👍

I agree. But anyway to support text direction for Checkbox we should make some changes in the Checkbox implementation (namely set the `dir` attribute in the outer `<div>`. Do you agree ?

joshblack:
@wajnberg thanks so much for this insight:

> Actually text direction and GUI mirroring are 2 independent features. One typical case is an Hebrew user typing/viewing Hebrew words in a LTR GUI (English GUI). This user would like to see these Hebrew words in RTL direction

I think this really clearly explained some of our difficulties in understanding this topic and I really appreciate it! 🙏 

> for Checkbox we should make some changes in the Checkbox implementation

Agreed, it seems like for that use case of setting _only_ text direction that the current prop API is insufficient. I tried to make a quick prototype over in this [Codesandbox](https://codesandbox.io/s/tender-voice-04ftb?file=/src/App.js) to play around with some ideas that we could do in the library, curious what you think. A big focus in the demo is to separate out the ideas of layout and text direction and make it simple for a developer to toggle either through React context for an entire application or a specific component / sub-tree.

wajnberg:
Hello @joshblack 

Thank you for putting this Codesandbox together. I have 2 main remarks:

1. I don't think we need an inheritance mechanism for text direction. Actually text direction is only needed for user originated data  (not for PII: product fixed strings). So the classic case is to support text direction in a very limited number of GUI elements in a given page.
This is also true for mirroring. I mean the whole page is mirrored or not. So something like `<html dir="rtl"> should be sufficient.
2. The writing of `<TextDirection dir="rtl">
        <Checkbox>שלום Hello</Checkbox>
      </TextDirection>` is not easy for the end user. I'd much prefer the simple `<Checkbox dir="rtl">שלום Hello</Checkbox>`.

joshblack:
@wajnberg I think the big thing was trying to figure out how to tackle these challenges at the system level. Totally understand that this specific case is the direction in the checkbox, but was thinking in terms of:

- How can I set the text direction of:
  - The entire app
  - A specific part of a page
  - A specific component
- How can I set the layout direction of:
  - The entire app
  - A specific part of a page
  - A specific component

In the case of a checkbox, the direction could be set in terms of an entire page, an entire form, or at the component level. The `TextDirection` component was one attempt to try and satisfy these use-cases in a way that could be re-used from component-to-component.

This could totally be me misunderstanding requirements though, so definitely let me know! I definitely don't want to make this harder than it needs to be for folks trying to tackle these use-cases 👍 

Just curious, is there a component library in the React ecosystem that we could look towards as exemplars in this space?

wajnberg:
Hello @joshblack 

In my opinion the only requirement is to set the text direction at the component level.

At: https://github.ibm.com/BDL/BidiMCSforReact you can find some utilities functions. But it's not really a library of components.

joshblack:
@wajnberg definitely agreed, it does seem important to be able to set things like `<Checkbox dir="rtl" />`. The next question that came to mind, personally, was what if I have a group of components for which I would like to set `dir="rtl"`. Just was playing around with ideas to make this easier for this use-case and other translation use-cases that we might see.

Also just wanted to say thanks again for the back-and-forth, really appreciate it. Just to get a handle on what is being proposed, I had a couple of follow-up questions:

- Would the proposal here be that all components expose a `textDir` prop to configure the direction of text in a component?
  - If it's only a handful, what would you say is the rule to determine whether a component should have a `textDir` prop?
- Would `textDir` need to passed through to child components?
- If a component has multiple text nodes, would it need to expose a prop to configure these nodes independently?

Side-note, seems like that repo may be private but would love to take a cool if it's possible to get access!

wajnberg:
Hello @joshblack ,

To answer your questions:

1. `textDir` attribute is not needed by every component. For instance `Slider` contains only numeric values. So textDir is not relevant there. Actually `textDir` support is only needed where user originated data is displayed (by opposition to fixed labels of a given product).
2. Would textDir need to passed through to child components? My 1st guess would be to say no. But could you please give some samples ?
3. If a component has multiple text nodes, would it need to expose a prop to configure these nodes independently? No. In a real product user sets his own text direction preference. So the same text direction would be applied to every text node.  But could you please also give here some samples ?

You should now have access to  https://github.ibm.com/BDL/BidiMCSforReact

PR #7684, opened 2021-01-28
feat(popover): add popover component and initial styles

### LINKED ISSUE #7713: [popover] basic component spec
#### Container
- Color and dropshadow layer

#### Variants
- No tip
- With tip
- Light prop
- High contrast for tooltips

#### Directions
- Bottom (center/left/right)
- Top (center/left/right)
- Right (center/top/bottom)
- Left (center/top/bottom)

#### Dimensions
- Padding
- Height and width (max)

#### Triggers
- On click
- Hover
- Focus
- Dismissable

#### Themes
- White theme
- Gray 10 theme
- Gray 90 theme
- Gray 100 theme

Ref #7649 and #6633

---------------
Popover - basic specs Copy

joshblack:
Closes #7713 

This PR implements the initial spec for the Popover component, a utility for rendering UI on a layer that sits above the current UI. Currently, it handles the follow situations from the spec in the issue:

- Variants
  - [x] With caret
  - [x] Without caret
  - [x] Light 
  - [x] High contrast
- Alignment (to the trigger element)
  - [x] Bottom (center, left, right)
  - [x] Top (center, left, right)
  - [x] Right (center, top, bottom)
  - [x] Left (center, top, bottom)
- Dimensions
  - [x] Width (with max)
  - [ ] ~Padding~ This will be implement for layouts that use it since the component can both render text and things like menus

#### Changelog

**New**

- `_popover.scss` in `packages/components`
- `src/components/Popover` in `packages/react`

**Changed**

- Update our `useId` helper to reference `canUseDOM` from a shared `environment.js` file

**Removed**


laurenmrice (changes requested):
Lookin good! 🎉 I have a couple different comments around alignment and spacing:

----------

- When the caret is positioned top/bottom and left/right. The caret needs to be 16px from that  specified edge. 

16px caret edge

----------

**Examples story:**
- There is a tooltip that comes up as well as a popover. For these examples shouldn’t we only have the popover appear here?

Screen Shot 2021-02-24 at 11 28 39 AM

- For the 48px icon buttons, for demoing purposes on this story we need to move the popover over so the caret center aligns with the button, instead of aligning to the container. I can show other different directions with specs if that would be helpful.

for demo show as intended

----------

**Playground story:**
- Can we bring the popover more into view on the page, so when people try to play with it they can see the popover better?

Screen Shot 2021-02-24 at 11 29 56 AM

joshblack:
@laurenmrice updated! I fixed the caret positioning so it should be good, I also excluded the dev-only stories like I mentioned in Slack so it should just be the playground now 👀

packages/components/src/components/popover/_popover.scss
```diff
@@ -0,0 +1,233 @@
...
+/// @group popover
+@mixin popover {
+  $popover-text-color: $text-01;
+  $popover-caret-offset: 1rem;
+  $popover-box-shadow: 2px 2px 6px 0 rgba(0, 0, 0, 0.2);
+  $popover-offset: 8px;
+
+  .#{$prefix}--popover {
+    position: absolute;
+    z-index: z('floating');
```
tay1orjones, line 23: Will this cause Popovers to render behind a modal by default?

joshblack: @tay1orjones I _think_ so based off of: https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/globals/scss/_layout.scss#L14

laurenmrice (approved):
Looks good to me ! 🙌🏻

joshblack:
Let me know if you have any questions @tay1orjones @emyarod @tw15egan !

packages/react/src/components/Popover/Popover-story.js
```diff
@@ -0,0 +1,442 @@
...
+  return (
+    <Popover
+      align={select('Specify the caret alignment', options, 'top')}
+      caret={boolean('Specify whether the caret should be visible', true)}
+      highContrast={boolean(
+        'Specify whether the high-contrast variant should render',
+        false
+      )}
+      light={boolean('Specify whether the light variant should render', false)}
+      open={boolean('Specify whether the popover is open or closed', true)}>
```
emyarod: we typically include the prop names in the label right?

```suggestion
      align={select('Specify the caret alignment (align)', options, 'top')}
      caret={boolean('Specify whether the caret should be visible (caret)', true)}
      highContrast={boolean(
        'Specify whether the high-contrast variant should render (highContrast)',
        false
      )}
      light={boolean('Specify whether the light variant should render (light)', false)}
      open={boolean('Specify whether the popover is open or closed (open)', true)}>
```

packages/components/src/components/popover/_popover.scss
```diff
@@ -0,0 +1,233 @@
...
+
+  .#{$prefix}--popover--open {
+    display: block;
+  }
+
+  .#{$prefix}--popover-contents {
+    @include box-shadow;
+
+    position: relative;
+    width: max-content;
```
emyarod, line 39: we've been avoiding this but have we already dropped support for IE?

emyarod (commented):
are some of the stories missing from the storybook preview? Default and Examples aren't appearing

packages/react/src/components/Popover/Popover-story.js
```diff
@@ -0,0 +1,442 @@
...
+} from '@carbon/icons-react';
+import { withKnobs, boolean, select } from '@storybook/addon-knobs';
+import React from 'react';
+import { Popover, PopoverContent } from '../Popover';
+import Button from '../Button';
+import mdx from './Popover.mdx';
+import { useOutsideClick } from '../../internal/useOutsideClick';
+
+export default {
+  title: 'unstable_Popover',
```
tw15egan:
```suggestion
  title: ''Experimental/unstable_Popover',
```

packages/react/src/components/Popover/Popover-story.js
```diff
@@ -0,0 +1,442 @@
...
+          size={size}
+        />
+        <Popover align={align} open={open} relative={relative}>
+          <PopoverContent className="p-3">{children}</PopoverContent>
+        </Popover>
+      </div>
+    );
+  }
+
+  return (
```
tw15egan, line 224: Can all this extra stuff in the story file be removed?

packages/components/src/components/popover/_popover.scss
joshblack: @emyarod the `Popover` work is targetting post-IE11 (v11) stuff 👀

packages/react/src/components/Popover/Popover-story.js
joshblack: I figured we could keep it for now until we figure out a better alignment story that we can use to verify (or only enable ti for development but not for production kind of thing)

joshblack:
@emyarod correct, some of them are hidden since they are more-so for development working on it versus intended to be used by the end-user. I just don't think we have a good example of a story to use for alignment yet so it's just hidden for now while we figure that out 👀

emyarod (approved):

tw15egan (approved):
LGTM 👍 ✅

Issue #4607, opened 2019-11-07
[Modal] Design enhancements - Ready


aagonzales:
Introducing new modal enhancements and variations.
Check indicates dev delivered. 

- Sizes: https://github.com/carbon-design-system/carbon/issues/4607#issuecomment-551198387
   - [x] 4 sizes
   - [x] Responsive widths
   - [x] Max heights
   - [x] Overflow content
   - [x] Margin-right
- Types: https://github.com/carbon-design-system/carbon/issues/4607#issuecomment-551959471
   - [x] Passive
   - [x] Transactional
   - [x] Danger
   - [x] Input
   - [ ] Progress (formerly Linear) 
- Button groups: https://github.com/carbon-design-system/carbon/issues/4607#issuecomment-551996283
   - [x] 1 button
   - [x] 2 buttons
   - [x] 3 buttons
- Examples: https://github.com/carbon-design-system/carbon/issues/4607#issuecomment-552002580
- Docs

**Sketch file:**
https://ibm.box.com/s/38r605b633kvmpt5qcv67owqn6imfmhk

aagonzales:
# Modal sizes
- XS
- Small
- Default
- Large

## Widths 
![image](https://user-images.githubusercontent.com/11670886/68413537-13d79e80-0154-11ea-9708-53cb3a99fe5a.png)

The width's will vary based on breakpoint

Breakpoint | XS  | Small | Default | Large 
-- | -- | -- | -- | --
1584 | 24% | 36% | 48% | 72%
1312 | 24% | 36% | 48% | 72%
1056 | 32% | 42% | 60% | 84%
672 | 48% | 60% | 84% | 96%
320 | 100% | 100% | 100% | 100%

% corollate to an approximate column count

% | ~# of columns
-- | --
72% | 12
60% | 10
48% | 8
42% | 7
36% | 6
24% | 4

## Heights 

![image](https://user-images.githubusercontent.com/11670886/68435333-b0b03100-0180-11ea-9666-9c1b0a691eab.png)

Modal size | Max-height
-- | --
XS | 48%
Small | 72%
Default | 84%
Large | 96%

### Overflow content
Use a fade to indicate there is content below the fold

![image](https://user-images.githubusercontent.com/11670886/68436605-194cdd00-0184-11ea-847c-8cf34919783d.png)


## Margin-right

- If modal is **smaller** than 32% then the margin-right should be **16px/1rem**
[...]

PR #7690, opened 2021-01-29
fix(Modal): clean up Modal, ComposedModal stories

### LINKED ISSUE #7671: [audit] modal storybook bugs
- Storybook needs to be update to new structure with playground
- Playgrounds and example for `modal` and `composedModal` should include the same content. Right now its featuring different content and examples.
- Need a knob for different footer options: 1, 2 and 3 button options

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7671

Syncs the `Modal` and `ComposedModal` storybooks so that they show similar examples. This also adds in single-button functionality to `Modal` instead of users needing to use `ComposedModal` for that variant.

@aagonzales let me know if there are any other stories we want to display. I don't see anything for 3 buttons, so I'm wondering if we should split that into a separate issue and I can add that in separately.

#### Changelog

**New**

- A new knob called `Number of buttons` allows you to dynamically choose how many buttons should display. Can be found in the `Playground`

**Changed**

- There are now 4 stories, `Default`, `Passive`, a state manager example to handle `open`/`close`, and a `Playground` that has all knobs enabled
- The presence of the `secondaryButtonText` prop will now handle displaying the secondary button in `Modal`

**Removed**

- Old examples

#### Testing / Reviewing

Ensure all knobs work, there are no prop warnings in the console

Are there any additional stories we would like to see here?


emyarod (commented):
it looks like there are a few failing CI tests currently, otherwise looks good to me after design review. on a side note I have some leftover local modal enhancements from #4607 if we want to reprioritize it

tw15egan:
@emyarod awesome! Yeah I think we should definitely prioritize getting that into the next sprint 👍

aagonzales:
Looks good!

@tw15egan For additional stories, does the modal component have built in [validation](https://carbondesignsystem.com/components/modal/usage/#validation) or [loading](https://carbondesignsystem.com/components/modal/usage/#loading)? Or are those applied later?

tw15egan:
@aagonzales Nothing built-in, but a user could use those individual components (Form elements, Inline Loading) to create those experiences

aagonzales (approved):

emyarod (approved):
I guess I left my previous review on "comment", was waiting for this to continue #4607

aledavila (approved):
lgtm 🎉

PR #7695, opened 2021-01-29
fix(inline-notification): remove ghost button color override

janhassel:
A ghost button's color in low-contrast inline notifications was previously overwritten with `link-01`, causing contrast issues in dark themes (`g90` and `g100`). This PR removes that override to ensure the correct shade of blue is used.

#### Changelog

**Removed**

- ghost button color override in low-contrast inline notifications

#### Testing / Reviewing

- Make sure the color of the ghost (action) button has sufficient contrast on all types of low-contrast inline notifications across all four themes.

#### Additional information

Inline notification component, kind: info, lowContrast: enabled, theme: g100

Before:
image

After:
image


emyarod (approved):

laurenmrice (approved):
Looks good to me 🙌🏻

tw15egan (approved):

PR #7698, opened 2021-01-29
fix(modal): set modal content position rule

### LINKED ISSUE #7682: Modal with long selectable data-table in modal body : selecting a row causes modal to content overflow

- [x] @console/pal v1.337.5

## Detailed description
Creating a modal with a table inside with selectable rows .
if you have a lot of selectable rows selecting the last row will cause the content to overflow outside of the modal.
causing a blank modal content

- create a modal with long table content with selectable rows
- scroll to the last element
- select last element
- content disappears

 sandbnox replication of the issue : https://codesandbox.io/s/dark-bush-dedz7?file=/index.js

## Additional information

image

## Suggestions for fixing this
- it appears that this can be fixed by adding position: relative to bx--modal-content currently its inheriting a static position in css

emyarod:
Closes #7682

This PR sets the `position` rule for the modal content container so that overflow content will not be inaccessible

#### Testing / Reviewing

Confirm that overflow content is accessible and viewable within modals (refer to original ticket for example code) and that there are no regressions


tw15egan (approved):
Confirmed the bug is no longer occurring 👍 ✅

joshblack (approved):

dakahn (approved):

wkeese:
Unfortunately this change causes a regression where dropdowns/popups are now clipped by the edge of the dialog, see #9524.

PR #7699, opened 2021-01-29
style(Themes): update skeleton-02 value in g100 theme

### LINKED ISSUE #3335: [Tokens]: investigate skeleton values
From @jeanservaas:

I read @shixiedesign's issue and it does seem like the token change is to spec. However, I'm noticing some discrepancies in the values for the skeleton tokens.

- I searched Git and I can't find Shixie's design/animation spec for the skeleton state. I wanted to double check it because I noticed that incorporating the @skeleton-02 token in the light themes seems very dark. I wanted to make sure this was intentional.

- This may not be a token issue... it may just be a value issue, in which case we could merge this PR and open another issue... but I want to make sure. I will ask @laurenmrice when she's in too, because she probably has more insight.

**$skeleton-02 value is very dark in light themes**
image

**skeleton token values (at least for G90) are incorrect on the Carbon site**
image

**Netlify elements preview is showing different values for G90 skeleton tokens then the website, which one is right?**
image

### LINKED ISSUE #7348: Bug - Checkbox component loading skeleton appears to be static
### Affected component:
https://ibm-security.carbondesignsystem.com/?path=/story/components-checkbox–skeleton

We inherit the skeleton styling directly from Carbon. It appears to be the same color across all components for the Gray 100 theme.

### Expected behavior -
Clearly see the skeleton loading when page is being loaded

### Actual behavior -
Contrast on the skeleton component is not enough, so that you can only tell it is loading after zooming in multiple times

### Steps for reproducing

1. View the component here: https://ibm-security.carbondesignsystem.com/?path=/story/components-checkbox–skeleton
2. Check does it look like it's actually loading dynamically when you refresh the brower page

image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/3335
Closes https://github.com/carbon-design-system/carbon/issues/7348
Rel #7679 

Updates `$skeleton-02` value in g100 theme

#### Changelog

**Changed**

- Bumped `$skeleton-02` value from `Gray80` to `Gray70` to improve contrast

#### Testing / Reviewing

Ensure `$skeleton-02` value is correct


aagonzales (approved):
Yeah that's better!

emyarod (commented):
looks good to me

emyarod (approved):

aledavila (approved):
Lgtm

PR #7703, opened 2021-01-29
feat(themes): add experimental sass modules for themes

joshblack:
Adds experimental Sass Module support to `@carbon/themes`.

#### Changelog

**New**

- Add `sass-module` example to `themes/examples`
- Add `index.scss` entrypoint to `packages/themes`

**Changed**

**Removed**

#### Testing / Reviewing

- Run the `sass-modules` example locally and verify that custom properties were emitted for each of the sections and the `body` element


tay1orjones (approved):
🎉   I see the same amount of custom properties on the body and on each section 👍

dakahn (approved):

PR #7710, opened 2021-02-01
chore(project): remove sassdoc commands

joshblack:
We currently generate SassDoc files under `docs/sass.md`. However, there have been questions as to whether these artifacts are helpful in their current format. After a quick discussion on Slack, we talked about potentially removing these artifacts and revisiting them if teams do in fact rely on them heavily.

Since folks weren't leaning too strongly on this one way or another, we'll defer to removal since it adds to the codebase in terms of file changes and build times (albeit not substantially)

#### Changelog

**New**

**Changed**

**Removed**

- Remove SassDoc commands and generated artifacts


jnm2377 (approved):
LGTM. Totally agree, seemed like removing was a good move if people don't really use it. Thanks for updating this!

tw15egan (approved):
LGTM 👍 ✅

tw15egan:
@joshblack just has some merge conflicts but G2G afterwards

joshblack:
Ugh I'll just wait till EOD when all the PRs are merged in since each one will cause a merge conflict 😂

PR #7726, opened 2021-02-02
style(DataTable): fix sortable table header padding

### LINKED ISSUE #7724: [data table] style regression: sort hover is rendering incorrectly
Hover is ending right at the edge of the text.
image

It should have 16px padding on the left edge
image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7724

Scopes the padding to the proper element when using the `sortable` version of `DataTable`

#### Changelog

**Changed**

- Removed padding from the `th` and applied it to the button that receives hover styles

#### Testing / Reviewing

Ensure hover styles are correct on the Sortable version of DataTable, and that no other variants have changed


aagonzales (approved):
That was fast!

dakahn (approved):

joshblack (approved):

dabrad26:
If using a table that has a mix of sortable headers and not sortable headers this causes the non-sortable headers to have no padding.  

I assume you would need to do something like as well to compensate for that use case?

```
.#{$prefix}--data-table.#{$prefix}--data-table--sort th > .#{$prefix}--table-header-label {
    padding-right: $spacing-05;
    padding-left: $spacing-05;
}
```

PR #7728, opened 2021-02-02
fix(CodeSnippet): preserve whitespace in single, inline variants

### LINKED ISSUE #7727: Whitespace is removed from CodeSnippet in single and inline variants
For context: https://codesandbox.io/s/restless-surf-611lw?file=/index.js

## Detailed description

`white-space: nowrap` removes whitespace from strings in `CodeSnippet`. We should preserve text and whitespace, but restrict wrapping. We can do this by using the `pre` value.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7727

Switches `white-space: nowrap` to `white-space: pre` 

`nowrap`:
Collapses white space as for normal, but suppresses line breaks (text wrapping) within the source.

`pre`: 
Sequences of white space are preserved. Lines are only broken at newline characters in the source and at <br> elements.

#### Changelog

**Changed**

- Switches `white-space: nowrap` to `white-space: pre`
- Moves `single` white space style to new block with `inline`

#### Testing / Reviewing

Ensure a string like `<div>   Hello World    </div>` does not have its whitespace stripped in `single` and `inline` variants.

For example https://codesandbox.io/s/restless-surf-611lw?file=/index.js


emyarod (approved):
looks good to me

andreancardona (approved):
looks good 🌮

Issue #7346, opened 2020-11-25
Create add-on package for animated icons 


kristastarr:
I have created animations of some of the Carbon Icons, to be used in the API Connect landing page. We would like to start an animation library to make the existing animations available for others to use, and to enable others to design and contribute additional animations.

Prototype of animations: https://pages.github.ibm.com/Krista-Starr/motion/#/landing-page
(hover on the tiles)

Animations are created using CSS and are of the form:

.trigger-selector:hover .icon-element-selector {
  ... some CSS
}
The user of the package could either use a pre-canned trigger selector e.g. .carbon-animate-icon:hover, or use an SCSS mix-in to apply the animation to a trigger-selector of their choice.

We would need to have classes added on svg elements, perhaps optionally based on a component switch, and ideally added by the icon builder code.

Credit to @lee-chase for the above implementation ideas.

lee-chase:
Possible mixin implentation.

```scss
// library
@mixin carbon-animation--icon-develop() {
    .icon--develop {
      transition: 0.2s all;
      transform-origin: 50% 50%;
      height: 36px;
      width: 36px;
      transform: scale(1.02);
    }
    .pencil {
      will-change: transform;
    }
    &:hover .pencil {
      animation: carbon-animation--icon-develop 2.5s infinite cubic-bezier(0.4, 0.14, 0.3, 1);
    }
};
@keyframes carbon-animation--icon-develop {
  0% {
    transform: translateX(0);
  }
  10% {
    transform: translateX(10px);
  }
  20% {
    transform: translateX(0px);
  }
  100% {
    transform: translateX(0px);
  }
}
// Usage
.info-tile {
  @include carbon-animation--icon-develop();   
}
```

lee-chase:
NOTE: this also works for pictograms built using the icon builders

lee-chase:
@kristastarr thought you might be interested in this. I have only updated the React builder to test out the interest in progressing this with Carbon.

kristastarr:
@johnbister Here's a link to the prototype: 
https://pages.github.ibm.com/Krista-Starr/carbon-icons/#/
https://github.com/kristastarr/carbon-icons-demo

sstrubberg:
Addressing this issue at [carbon-design-system/icons-motion](https://github.com/carbon-design-system/icons-motion)

PR #7729, opened 2021-02-03
Add icon element class names for animation

lee-chase:
Contributes to #7346 

Currently, SVG Icon elements do not have classes to allow targetted CSS animation.

This PR adds the ability to utilise a className as part of the SVG metadata. It also adds a property to the Icon component to optionally add an automatically generated class based on the element type and index in the metadata.

NOTE 1: A slightly more robust auto generated class could be created, based on a selection of attribute values in the element.
NOTE 2: We could go a little further here and split `Path` elements on a 'z' and/or capitalized 'M' to make the animations more versatile. The same strategy used for differentiating auto generated classes could be used here.

#### Changelog

**Changed**

M       packages/icon-build-helpers/src/builders/react/builder.js
M       packages/icon-build-helpers/src/builders/react/components/Icon.js

#### Testing / Reviewing

Tried it out in a test app with and without prop 'addElementClasses' also with and without a className attribute set in metadata.json.


joshblack:
@lee-chase thanks so much for putting this together! It's awesome to see the generation logic in there.

I'm a little hesitant about is adding this to the base packages, are there any other strategies we could use for enabling these kinds of transforms for adding classes to the paths?

One thing that came to mind is a dedicated package that uses the icon data from `@carbon/icons` and provides bespoke class names. It definitely isn't auto-generated, however, but may provide more control if that is needed.

lee-chase:
@joshblack  Another option would be to provide a wrapper to the Icon component, or perhaps just an alternate that walks the children and creates a component with classes added on the fly. The performance hit would be tiny on small SVGs.

lee-chase:
Perhaps we can get together to discuss how #7751 / #7346 are progressed. Are there others you would suggest joining such a conversation @joshblack ?

joshblack:
@lee-chase I think moving forward @johnbister will be great on the design side. 

On the technical side, I think it'd be great to have a dedicated package for this kind of stuff that uses the icon data from `@carbon/icons` to generate the icons with the preferred class names for the icons that are being worked on.

I think from that experiment we can get a feel for how widespread this use-case is and how folks will want to leverage it before folding it into `@carbon/icons-react`, if that makes sense.

Let me know what you think!

lee-chase:
> @lee-chase I think moving forward @johnbister will be great on the design side.
> 
> On the technical side, I think it'd be great to have a dedicated package for this kind of stuff that uses the icon data from `@carbon/icons` to generate the icons with the preferred class names for the icons that are being worked on.
> 
> I think from that experiment we can get a feel for how widespread this use-case is and how folks will want to leverage it before folding it into `@carbon/icons-react`, if that makes sense.
> 
> Let me know what you think!

Will try and sort this at some point Josh, but struggling for time.

joshblack:
Hey @lee-chase! 👋  Going to close this out since I think the group is going for a smaller approach first 👀  Let me know if this is not the case!

PR #7730, opened 2021-02-03
fix(react): update table header unique id to use stable id

### LINKED ISSUE #7721: Is uniqueId in TableHeader for aria-describedby and id necessary?
Hi, 
First of all, let me apologize if this is neither the place nor the issue type or label for this kind of thing. I tried to place this question in the Discord server but I think this is not available yet.

Whenever you test a snapshot of a mounted component and that component has `TableHeader` as a child, the test will failed as there is a `Math.random()` call in `TableHeader.js` for some props, specifically `id` and `aria-describedby`. Obviously you can mock `Math.random()` and the tests will passed (that's also why I wouldn't consider this a bug), but I wonder if we could add some props for these fields.
The code in question in `TableHeader.js` is the following:

Screen Shot 2021-02-02 at 15 10 23

Thanks in advance!

joshblack:
Closes #7721

Updates `TableHeader` to leverage `useId` versus `Math.random` for an internal unique id. While this doesn't prevent all randomness, there is a deterministic order that should make it easier for teams that use snapshot testing to snapshot components that contain `TableHeader`.

#### Changelog

**New**

**Changed**

- Update `TableHeader` to use `useId` versus `Math.random`

**Removed**

#### Testing / Reviewing

- Verify that the column headers in data table have the correct `id`s generated and applied


dakahn (approved):

tw15egan (approved):
LGTM 👍 ✅

Issue #7649, opened 2021-01-25
Popover Component


joshblack:
- [x] Design research around use-cases, competitors, etc. 
- [x] [Design specs ](https://github.com/carbon-design-system/carbon/issues/7713)
- [ ] Development
  - Make an experimental version of Popover with minimal styles (until design specs) that supports different directions for carets and arbitrary content
  - [Stretch] Make an experimental version of Disclosure
- [ ] Guidance
  - [x] Design docs
  - [ ] Dev docs

Links & Resources

- [Tooltips: Investigation into four parts ](https://www.youtube.com/watch?v=lb0_v7D4kbs)
- [Disclosure Component](https://github.com/carbon-design-system/carbon/issues/6633)
- [Primer Popover Component](https://primer.style/components/Popover#caret-position)

tw15egan:
@joshblack would this possibly include enhancement to the current `Icon only Button`? 

It looks like we've got some other issues pertaining to:
- customization of the tooltip delay (https://github.com/carbon-design-system/carbon/issues/6718 & https://github.com/carbon-design-system/carbon/issues/6178)
- icon button retaining the tooltip when focused and another Tooltip is hovered (https://github.com/carbon-design-system/carbon/issues/4560 & https://github.com/carbon-design-system/carbon/issues/7040)

joshblack:
@tw15egan I think it definitely could be related, when talking to @laurenmrice we were trying to see if we could get the dark/high-contrast "tooltip" styles included so that we can build out those kinds of components as well.

PR #7736, opened 2021-02-03
fix(HeaderGlobalAction): provide visual text for link via title attribute

### LINKED ISSUE #7731: Action icons in the UI Shell Header do not provide visual text to describe the link purpose
## [Action icons in the UI Shell Header](https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-actions) do not provide visual text to describe the link purpose
This is a failure of [WCAG 2.1 Success Criterion 2.4.4 Link purpose (In context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context#techniques) The intent of this Success Criterion is to help users understand the purpose of each link so they can decide whether they want to follow the link. 
If a user does not know what the icon means, there should be visible text to describe the link's purpose such as a tooltip.

tay1orjones:
Closes #7731

Adds a `title` attribute to the HeaderGlobalAction in the same fashion that HeaderMenuButton has to provide visual text for the link.

#### Changelog

**New**

- add `title`

#### Testing / Reviewing

Title text should be visible on hover of a HeaderGlobalAction


tw15egan:
I'll defer to DA if a `title` will be sufficient enough, but if not, I think we can just change this from a `button` to `Button` and add a few props:

```jsx
<Button
      {...rest}
      {...accessibilityLabel}
      className={className}
      onClick={onClick}
      type="button"
      hasIconOnly
      iconDescription={ariaLabel}
      tooltipPosition="bottom"
      tooltipAlignment="end"
      ref={ref}>
      {children}
    </Button>
```

which renders as 
Screen Shot 2021-02-03 at 1 27 08 PM



Looks like we'd just need to add a `justify-content: center` to `.bx--btn.bx--btn--icon-only.bx--tooltip__trigger` to fix the alignment issues.


We'd also probably want to pass down the `tooltipAlignment` from `<HeaderGlobalAction>`, or figure out a way to determine if it is the last item.




This also then ties into the problem discovered here, but that can be fixed in a separate PR: https://github.com/carbon-design-system/carbon/issues/7040

Screen Shot 2021-02-03 at 1 33 42 PM

tay1orjones:
@dakahn for context here, I _think_ the `title` is enough based on [the WCAG link](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context#techniques) provided in the initial issue; `Sufficient technique #6, H33`, but let me know if that's off base. I'm happy to take the approach TJ outlined if it makes more sense

tw15egan:
@tay1orjones for sure, this may be our best bet until we get the popover component (https://github.com/carbon-design-system/carbon/issues/7649) to avoid issues a few people have pointed out (#4560, #7040)

aledavila:
I wonder the issues this would cause considering theyre not links but buttons that would open up the panels that are in the pipeline for v11. While I understand the need to descirbe a link they're not and shouldn't be used as such.

dakahn:
**Sorry, missed this in my notifications.** `title` is pretty well supported with NVDA and JAWS, but gets sketchier on macOS and iOS using VoiceOver as seen [here in powermapper](https://www.powermapper.com/tests/screen-readers/attributes/#title). I think the better way to go moving forward when programmatically labeling our icon buttons is to have a visually hidden label and reference the hidden label via `aria-labelledby`
```html
<button aria-labelledby="button-label"> 
    <span id="button-label" hidden>Menu</span>
    <svg aria-hidden="true" focusable="false" width="24" height="28" viewBox="0 0 24 28">
        <!-- svg content -->
    </svg>
</button> 
```

dakahn (dismissed):
**Sorry, missed this in my notifications.** `title` is pretty well supported with NVDA and JAWS, but gets sketchier on macOS and iOS using VoiceOver as seen [here in powermapper](https://www.powermapper.com/tests/screen-readers/attributes/#title). I think the better way to go moving forward when programmatically labeling our icon buttons is to have a visually hidden label and reference the hidden label via `aria-labelledby`
```html
<button aria-labelledby="button-label"> 
    <span id="button-label" hidden>Menu</span>
    <svg aria-hidden="true" focusable="false" width="24" height="28" viewBox="0 0 24 28">
        <!-- svg content -->
    </svg>
</button> 
```

tay1orjones:
I just pushed an update to use `Button` and pass through the tooltip alignment 👍

tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):
Very cool thanks so much

PR #7741, opened 2021-02-04
[RFC] feat(CodeSnippet): add language prop to work with PrismJS syntax highlighting

jnm2377:
PrismJS is a popular package used to add syntax highlighting to code snippets. However, the way it works is by adding a className (something like `language-javascript`) to the `<code>` element in a code snippet. Because we already include `<pre><code>` inside the code snippet, users don't have access to these to add highlighting. 

This RFC proposes adding a `language` prop, which adds the className to the `<code>` element in the CodeSnippet. This seemed like the best way to be able to provide this support for users (as opposed to perhaps adding a ref prop that will allow users to target the `<code>` element and then add the className).

The proposal simply adds the option to add that language className. Users can then create a custom component in their own repo using the Prism dependency. Something like:

```js
import React, {useEffect} from 'react';
import Prism from 'prismjs';
// load prism theme 
import 'prismjs/themes/prism.css'; //themes that work with carbon styles: default (light theme), solarizedlight (light theme), okaidia (dark theme), tomorrow (dark theme)
// load any additional languages to highlight
import 'prismjs/components/prism-scss'; // default languages loaded are markup, css, javascript, clike

const HighlightedCodeSnippet = ({ children, language, type, feedback }) => {

  useEffect(() => {
    Prism.highlightAll();
  }, []);

return (
      <CodeSnippet
        type={type}
        language={language}
        feedback={feedback}>
        {children}
      </CodeSnippet>
);
}

export default HighlightedCodeSnippet;
```

The highlighting would look like (using default theme):
Screen Shot 2021-02-04 at 10 43 00 AM

#### Changelog

**New**

- adds `language` prop to CodeSnippet to work with PrismJS syntax highlighting;

**Changed**

- wraps `<pre>` around `<code>` instead of otherway around to be able to apply prism js styles correctly. This does not affect carbon styles if prism js highlighting isn't applied.
- temporarily adds prism js dependency for storybook example; can remove this before merging, if it's something we're interested in.

#### Testing / Reviewing

- in the Storybook, go to CodeSnippet > Multiline Highlighted to view example


emyarod (commented):
should we also try exploring some less opinionated implementations as well? for example being able to support other libraries such as highlight.js

joshblack:
Very nice @jnm2377! 🎉 

For syntax highlighting, this might be a great use-case for an example/pattern that we document versus embedding it in the component directly. That could give us as a team a resource to point to when people ask about this use-case, and also keep the component flexible enough if people want o use other solutions (I think highlight.js is the other big one??)

Anyways, just some initial thoughts. Curious what you think! Super awesome stuff either way 🥳

jnm2377:
Thanks for the feedback, y'all. I'll close this for now, and open a discussion for further feedback! :)

PR #7756, opened 2021-02-08
feat(ui-shell): add side-nav-divider component

janhassel:
Introduces a new component for the SideNav: `<SideNavDivider />`.

It renders a diving line with the styling used on carbondesignsystem.com

#### Changelog

**New**

- Add `SideNavDivider` component

**Changed**

- Updated relevant snapshots

#### Testing / Reviewing

- Verify styling in new UI Shell story: "Fixed SideNav w/ Divider" (https://deploy-preview-7756--carbon-components-react.netlify.app/?path=/story/ui-shell--fixed-side-nav-w-divider)

#### Additional information

I opted for an `<li>` with the role `separator` instead of an `<hr>` element since according to the HTML spec, children of a `<ul>` must be `<li>` elements (https://html.spec.whatwg.org/multipage/grouping-content.html#the-hr-element) and the `<hr>` element has the implicit role `separator` (https://w3c.github.io/html-aam/#el-hr).


emyarod (approved):
looks good to me, requesting visual review @carbon-design-system/design

packages/components/src/components/ui-shell/_side-nav.scss
```diff
@@ -344,6 +344,15 @@
     height: mini-units(6);
   }
 
+  //----------------------------------------------------------------------------
+  // Side-nav > Navigation > Divider
+  //----------------------------------------------------------------------------
+  .#{$prefix}--side-nav__divider {
+    height: 1px;
+    margin: $spacing-03 $spacing-05;
+    background-color: $ui-03;
```
aagonzales: So we don't use the component color tokens in the UI Shell right now (hopefully correcting this in the future). We actually just use the language colors and it will look the same in each theme. Call the `Gray20` element token instead here.

aagonzales (changes requested):

packages/components/src/components/ui-shell/_side-nav.scss
janhassel: Ah, right, good catch!

aagonzales (approved):
🎉

packages/react/src/components/UIShell/SideNavDivider.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { settings } from 'carbon-components';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const { prefix } = settings;
+
+const SideNavDivider = ({ className }) => {
```
joshblack: Small style-nit, we try to use the function declaration form based on the [styled guide](https://github.com/carbon-design-system/carbon/blob/master/docs/style.md#writing-a-component) 👀

joshblack (approved):

packages/react/src/components/UIShell/SideNavDivider.js
janhassel: 👍  Updated

joshblack (approved):

PR #5744, opened 2020-03-30
feat(sketch): create shared styles for icon stroke

emyarod:
Closes #5692

This PR renames the fill shared style layers and lays the groundwork for specifying shared style type (currently "fill" or "border") to prepare for #4130/#5567 (branched off of #5569)

#### Changelog

**Changed**

- change shared sync command and generate commands for colors, themes, and icons to use new naming scheme
- fix symbol sync test

#### Testing / Reviewing

Ensure the sketch plugin works as expected and the new shared style names are correct

Plugin can be downloaded and tested here:

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/4404430/carbon-elements.sketchplugin.zip)

Issue #5692, opened 2020-03-24
Create shared styles for icon stroke


joshblack:
We need to author shared styles similar to how we do now for IDL and our tokens but specifically targeting icon stroke. This issue is for creating the command and code necessary for generating these shared styles.

The naming convention for this feature is: `color / border / <color> / <grade>`

We have a couple of options when it comes to commands, but we could add it to existing color sync commands or do a standalone command. Either works!

laurenmrice:
here is a sketch file with an example of how the color styles should be:
https://ibm.box.com/s/uz7nrdlt1o9szlzb6rz96kti1q8u4fnp

-------------------------------------------------------------------------------------------------

Trying to decide which naming convention makes more sense from a code and sketch perspective. `border` or `stroke`. From a sketch perspective, border makes more sense, is this the terminology you are using in code?

emyarod:
in code it's a stroke but I named it border in my draft PR and will probably keep that since that's what it's called in Sketch

laurenmrice:
ok cool lets go with border

PR #5664, opened 2020-03-19
fix(sketch): update shared color names to include space

joshblack:
Closes https://github.com/carbon-design-system/carbon-design-kit/issues/267

Update shared style name to include spaces around the dividing `/`.

image

#### Changelog

**New**

**Changed**

- Include space around dividing `/` in shared color names

**Removed**

PR #7762, opened 2021-02-08
fix(sketch): resolve sketch plugin color layer style duplication

### LINKED ISSUE #7734: [icons] color layer style duplication
## Color layer styles

Color layer styles are being duplicated. There are 2 separate sets of the ibm color palette. We only need one copy of the ibm color palette in the color layer styles.

 - [ ] We need to remove the word `fill` from the naming convention and also add spaces inbetween the slashes so it looks as follows: `color / blue / 60`
 - [ ] Remove any duplicate colors that show up

------

 **Example of what currently is happening with the plugin:**
- One set has a naming convention of `color / fill / blue / 60` with spaces inbetween the slashes.
Screen Shot 2021-02-03 at 1 55 05 PM

- The other set has a naming convention of `color/blue/60` without spaces inbetween the slashes.
Screen Shot 2021-02-03 at 1 54 49 PM

emyarod:
Closes #7734

related #5664 #5692 #5744

This PR resolves the color layer style duplication issue. Since we have changed direction in how we expect shared styles to be named, we can undo some changes initially introduced in #5744 to address #5692. We also add some logic to check for the breaking change in #5664 and rename the style layers in place

#### Changelog

**Changed**

- remove style type from style name
- use param objects instead of lists of params
- rename `formatSharedStyleName` to be more descriptive

**Removed**

- Color style "type" references

#### Testing / Reviewing

color style names should no longer contain style type, and any references to color style type should no longer be in the plugin codebase

image

image

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/5946503/carbon-elements.sketchplugin.zip)


joshblack (approved):

andreancardona (approved):
🎉

laurenmrice (changes requested):
After syncing the plugin with the icon sketch file, the new color palette structure is coming through and is being used by the icons, but the old color structure also remains which causes a duplication of the same colors.

Example:
Screen Shot 2021-02-09 at 3 54 02 PM
Screen Shot 2021-02-09 at 4 04 21 PM

emyarod:
here is the updated plugin which should rename layers in place. the layers should now have the correct name format and no duplicate layers will be generated

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/5968422/carbon-elements.sketchplugin.zip)

laurenmrice (approved):
I see the correct color styles and they are maintaining overrides when I tested it 🎉

PR #7767, opened 2021-02-09
style(rtlcss): update rtl blocks for better compatibility

### LINKED ISSUE #7476: According arrow up/down position on RTL mode

- [x] `carbon-components`

## Detailed description

I am trying to implement RTL feature on IBM pages using `carbon-for-ibm.com` with rtlcss and dir attr and got success in all cases, but just on Footer mobile the arrow position changed… I know the rtlcss package change the top-bottom and right-left position, however on masthead nothing was changed (worked well even on mobile). I mean the arrow on mobile needed to be facing down, but this is the current behaviour right below. 

Screen Shot 2020-12-16 at 18 01 56

Accordion

> see changed?

Would like to see the arrow to be facing down... to show there is a content there.

Chrome

10.25.0

1. Step one

Stating url: https://pages.github.ibm.com/Diego-Marcuz/testing-rtl/
Needs W3 id to log in.

2. Step two

Go to mobile screen and see the accordion behaviour on footer.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7476

Fixed an issue with Accordion chevrons not displaying in the correct direction when `rtl` mode is enabled. While working on this I noticed the Checkmark styles were not being interpreted properly and came across this not on the rtlcss website:

Screen Shot 2021-02-09 at 11 30 48 AM

Once I applied this, the Checkmark styles were not displaying properly and I had to use a funky 3d transform to get them to display correctly. If anyone has a better suggestion I'm all ears!

#### Changelog

**New**

- Updated rtlcss blocks to use the sass interpolation method so they are properly rendered `#{'/*rtl:...'}`

**Changed**

- Changed a few values to render checkmark as closely as possible to non-rtl variant

#### Testing / Reviewing

In your terminal run `export CARBON_REACT_STORYBOOK_USE_RTL=true` and ensure Accordion and Checkbox render as expected (I removed the Hebrew but should retain the same structure).

Screen Shot 2021-02-09 at 11 10 19 AM

Screen Shot 2021-02-09 at 11 08 43 AM


joshblack (approved):
Ugh this is such a bummer of a bug, thanks for digging into this!

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -113,8 +113,7 @@
     width: 1rem;
     height: 1rem;
     margin: $accordion-arrow-margin;
-    // TODO: RTL rotate(180deg);
-    transform: rotate(90deg);
+    transform: rotate(90deg) #{'/*rtl:rotate(90deg)*/'};
```
asudoh:
```suggestion
    transform: rotate(90deg) #{'/*rtl:ignore*/'};
```

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -197,8 +196,7 @@
     }
 
     .#{$prefix}--accordion__arrow {
-      /* rtl:ignore */
-      transform: rotate(-90deg);
+      transform: rotate(-90deg) #{'/*rtl:rotate(-90deg)*/'};
```
asudoh:
```suggestion
      transform: rotate(-90deg) #{'/*rtl:ignore*/'};
```

asudoh (commented):
Weighing in - Just in case the suggestions match the intent. Thanks @tw15egan!

packages/components/src/components/accordion/_accordion.scss
tw15egan: Ah, right! I think I had tried that before realizing Sass was not reading the rtl comments unless interpolated. Thanks for catching that @asudoh 😄

andreancardona (approved):
looks good! 🎉

PR #7774, opened 2021-02-10
fix(components): include side effects from src for rollup resolution

### LINKED ISSUE #7773: Unable to use components from CDN using vanilla JavaScript
We received a couple of reports on Slack about the vanilla JavaScript from `carbon-components` failing to correctly initialize when using the latest versions of our packages.

To reproduce, here is a minimal HTML file from one of the individuals who reported this issue:
[html.zip](https://github.com/carbon-design-system/carbon/files/5960550/html.zip)

When diving into this issue, it seems like Rollup is incorrectly stripping away our `src/globals/js/boot.js` import from `src/bundle.js`. It seems like in the latest versions of `@rollup/plugin-node-resolve`, the `sideEffects` field from the package's `package.json` is used. Since the files from `src` weren't listed, they would not be included in the file output because rollup assumed they had no side effects.

Updaing the `sideEffects` array in `packages/components/package.json` will address this problem.

joshblack:
Closes #7773 

Adds in the relevant src files to the `sideEffects` array to correctly emit our document event listeners on boot.

#### Changelog

**New**

**Changed**

- Update `sideEffects` in `components/package.json` to include source files so that rollup's node-resolve plugin correctly emits imports that have side effects

**Removed**

#### Testing / Reviewing

- Checkout the PR
- Build the components package locally (`yarn gulp scripts:rollup` is a nice shortcut in `packages/components`)
- Check out `scripts/components.js` and verify that the following block of code appears at the end:

```js
  if (document.readyState === 'loading') {
    document.addEventListener('DOMContentLoaded', init);
  } else {
    // DOMContentLoaded has been fired already
    // Let consumer have chance to see if it wants automatic instantiation disabled, and then run automatic instantiation otherwise
    setTimeout(init, 0);
  }
```


tw15egan (approved):
LGTM 👍 ✅ 

Screen Shot 2021-02-10 at 11 21 31 AM

dakahn (approved):
image

PR #7775, opened 2021-02-10
feat(Modal): implement support for 3 button modals

emyarod:
(tests will be updated when we have confirmed if the API changes are good to go)

--

Related #4607

This PR adds support for 3 button modals according to the latest modal spec. Since our existing Modal API expects only 1 primary and 1 secondary button, a `secondaryButtons` prop is introduced which can be used in place of the `secondaryButtonText` and `onSecondarySubmit` props. The `secondaryButtons` prop expects an array of up to 2 objects with the following shape:

```js
{
  buttonText: PropTypes.string,
  onClick: PropTypes.func,
}
```

if an `onClick` handler is not supplied to the secondary buttons in a `ComposedModal` it will fall back to the internal `handleRequestClose` method by default

#### Changelog

**New**

- `secondaryButtons` prop
- 3 button modal styles (not including linear directional modals)

#### Testing / Reviewing

Confirm the 3 button modal styles appear correct and there are no visual regressions with existing supported modal styles


aagonzales:
The primary button is reflowing weird. It should keep a consistent top margin. Is this just a button bug unrealted to this PR?
image

emyarod:
that's just directly related to the width change, since instead of 50% width they're now 25% width. if you increase the size of your viewport it would be what you'd normally expect I think

image

emyarod:
modified button text vertical alignment for 3 button modals only after discussion with @aagonzales

aagonzales (approved):
Thanks for fixing the button text. Looks good!

packages/react/src/components/ComposedModal/ComposedModal-story.js
```diff
@@ -28,14 +35,17 @@ const sizes = {
 };
 
 const buttons = {
-  'None (0)': 0,
-  'One (1)': 1,
-  'Two (2)': 2,
+  'None (0)': '0',
```
tw15egan, line 38: Seems like `0` doesn't do anything unless we specify the `passiveModal` prop, so should we just allow choosing 1, 2, or 3 buttons, and controlling the 0 option via the `passiveModal` knob?

packages/react/src/components/Modal/Modal-story.js
```diff
@@ -24,14 +31,17 @@ const sizes = {
 };
 
 const buttons = {
-  'None (0)': 0,
-  'One (1)': 1,
-  'Two (2)': 2,
+  'None (0)': '0',
```
tw15egan, line 34: Seems like `0` doesn't do anything unless we specify the `passiveModal` prop, so should we just allow choosing 1, 2, or 3 buttons, and controlling the 0 option via the `passiveModal` knob?

tw15egan:
Looks really great! Do you think we can add a quick blurb to the `.mdx` [files](https://react.carbondesignsystem.com/?path=/docs/modal--default#modal-button-variants) to showcase how to use the 3 button variant of `Modal` and `ComposedModal`? The doc table doesn't seem to showcase the expected config objects / props

Screen Shot 2021-02-12 at 12 39 51 PM

packages/react/src/components/Modal/Modal-story.js
emyarod: selecting 0 should be fixed now

tw15egan (approved):
LGTM 👍 ✅ 🎉

joshblack (approved):

martijn-cloudnatix:
I try to use the `secondaryButtons` props but in the typescript definition this field is marked as `readonly`.

PR #7777, opened 2021-02-10
fix(combobox): update to match ARIA 1.2 criteria

### LINKED ISSUE #6938: Fix Combobox aria roles to match aria 1.2 spec.

joshblack:
Closes #6938 

Updates the combobox implementation to match the [authoring practices reference](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-list.html) and includes updates from #6938 as suggested by @carmacleod 

#### Changelog

**New**

- Add new versions of components for ListBox that more closely align with where we want these components to go, this helps us add the required behavior into combobox without breaking existing components

**Changed**

- Update ComboBox so that the wrapper node no longer has `getRootProps` applied to it, these attributes now are on the `input`
- Convert both the trigger and clear icons to buttons and remove them from tab order (the equivalent functionality already exists through keyboard)

**Removed**

#### Testing / Reviewing

- Verify only combobox is impacted by these changes (splash zone includes dropdown, multiselect, filterable multiselect)
- Verify AC has no violations with both closed and open menus
- Verify Combobox can be operated by VO, JAWS, and NVDA
- Verify the following changes have occurred to the markup:
  - [ ] Outermost node no longer has combobox role and is now a wrapper div
  - [ ] The list box field is no longer interactive and has no additional attributes like ARIA labels
  - [ ] The list box input, menu icon, and selection are now siblings instead of descendants of a field which is a button
  - [ ] Focus shifts to the input when the menu is open or closed
  - [ ] Focus shifts to the input when the input is cleared
  - [ ] Both the menu trigger and clear button are not in tab order


dakahn (changes requested):
Looks great. Such awesome work. Checked on Windows 10 in Firefox with Accessibility Checker, NVDA, and JAWS latest

The only thing I noticed is that selected options aren't read as 'selected'. WCAG spec allows `aria-describedby` to reference `hidden` html elements. So maybe giving the svg the name 'selected' and then setting it to `hidden` and referencing it's name with `aria-describedby` on the menu item might be a cool option.

joshblack:
Issue for the enhancement: https://github.com/carbon-design-system/carbon/issues/7787

joshblack:
@laurenmrice just a heads up, with this change the button now receives a focus style and has a click area. It seems like the dimensions of this might need to be updated since it is too narrow, here is a video and an image that I hope will help out with explaining 👀 

Screen Shot 2021-02-11 at 14 45 22


https://user-images.githubusercontent.com/3901764/107696763-fa3ad600-6c77-11eb-9258-7dee2515f06d.mov

With these changes, what do you think would be the ideal spec for the menu trigger? 👀

dakahn (approved):

laurenmrice:
@joshblack We should make both focus squares for the close icon and chevron icon 24px. We also do the same thing right now for file uploader.
Artboard Copy

joshblack:
Thanks @laurenmrice !

@andreancardona would you want to tackle this or should I? 👀

mashenka123:
@joshblack This is blocking Code Engine eGA in Mar, could you please finish this bug fix this month? Thank you very much!
https://github.ibm.com/coligo/ui/issues/976
https://github.ibm.com/coligo/ui/issues/940

mashenka123:
@joshblack This is blocking Satellite eGA on 3/1, could you please fix this asap? Thank you very much! https://github.ibm.com/alchemy-containers/satellite-ui/issues/469

carmacleod:
Accessibility review

- [x]  Please delete `aria-owns="downshift-0-menu"` from the input (because an input cannot have a listbox child - instead, it _controls_ the listbox (and you already have `aria-controls="downshift-0-menu"`).

 

- [x] [FIXED IN ANOTHER PR] Please don't use the `title` attribute on an item unless the item text is too long and is truncated. Chrome and Safari both use title as the menuitem's description, which results in screen readers saying the menuitem name twice. (Firefox wisely prunes out the description when it is identical to the accessible name, so this doesn't happen in Firefox).

- [x]  You don't need both `<label for>` and `aria-labelledby` to label the input, so go with the HTML labelling method, and delete some excess ARIA. :)

   image

- [ ] The live region "help/hint text" that says, "6 results are available, use up and down arrow keys to navigate. Press Enter key to select." should just be deleted. Screen reader users usually turn off the help/hint text for their screen reader (as soon as they become proficient with it) in order to reduce verbosity, and since they wouldn't be able to turn this off, it would get annoying pretty quickly. :)

- [x]  Please don't use `aria-selected="false"` on the options. Instead, just remove the `aria-selected` attribute from options that are not selected. (Some screen readers are saying "unselected", which isn't useful).

Other things I noticed:

- [x]  you might want to give the open button `outline: none;` because it looks a bit unexpected to see the focus outline (very apparent in Chrome/Win) when this element does not take focus.

- [ ]  the styling for a selected option (light grey background) and the current option (checkmark plus similar light gray background) are confusing when seen side-by-side. To see this, type 2 and then Enter (Option 2 is selected). Now type down arrow. Option 3 currently has focus, but it looks almost the same as Option 2 except for the checkmark. Users may get confused.

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -382,126 +386,156 @@ export default class ComboBox extends React.Component {
...
+            {
+              suppressRefError: true,
+            }
+          );
+          const labelProps = getLabelProps();
+          const buttonProps = getToggleButtonProps({
+            disabled,
+            onClick: this.onToggleClick(isOpen),
+            // When we moved the "root node" of Downshift to the <input> for
+            // WCAG 2.1 compliance, we unfortunately hit this branch for the
```
carmacleod:
```suggestion
            // ARIA 1.2 compliance, we unfortunately hit this branch for the
```

andreancardona:
@carrenelloyd @alisonjoseph Just an update I am working to address all these issues! :)

andreancardona:
@laurenmrice Hey Lauren! I wanted to address Carolyn's last comment regarding the similar shades if gray as seen here. Wanted to know if you had any thoughts? And / or if this should be a broader discussion extracted into a different issue? 

Let me know! 

Screen Shot 2021-02-24 at 1 22 20 PM

laurenmrice:
@andreancardona Is this currently breaking an accessibility rule? For the selected state, we have the same word presented in the top field of the combobox as well as inside of the menu. The selected state also gets a background color change and the addition of an icon. 

Artboard

The selected state background (#e0e0e0) and the hover state background (#e5e5e5) are different hex values but the colors do look similar. Changing the values of one of these colors would need to be looked into by the idl/brand team.

andreancardona:
> @andreancardona Is this currently breaking an accessibility rule? For the selected state, we have the same word presented in the top field of the combobox as well as inside of the menu. The selected state also gets a background color change and the addition of an icon.
> 
> Artboard
> 
> The selected state background (#e0e0e0) and the hover state background (#e5e5e5) are different hex values but the colors do look similar. Changing the values of one of these colors would need to be looked into by the idl/brand team.

@carmacleod can you please clarify if it is an accessibility violation? :) Thank you!

carmacleod:
I'm not sure. :)

It may fail https://w3c.github.io/wcag21/guidelines/#non-text-contrast because the colors are so similar, and the focus/hover item can be right beside the "selected" item.

@mbgower suggestions?
To see this in context:
- Go to https://deploy-preview-7777--carbon-components-react.netlify.app/iframe.html?id=combobox--combobox
- Focus the combobox
- Type 2 and then Enter (Option 2 is selected).
- Now type down arrow. Option 3 currently has focus, but it looks almost the same as Option 2 except for the checkmark.
- Similar potential confusion with hover state.

I'm wondering if simply removing the background color from the selected item would work, i.e. only change background for focused (or hovered) item.

mbgower:
Hi all, just absorbing all this.
In regards to my take on the alternating shading on the options, I have a suggestion, but it's a little unexpected, perhaps.

Our findings from a low vision study we did a couple years back was that menus (button menus in particular) are one of the biggest challenges for LV users. The reason for this is that an open menu does not have well defined edges in most implementations. A lot of cases add a very light grey shadow, which utterly fails minimum for non-text contrast.

The result for LV users is that they have difficulty both locating the open menu and understanding where the menu items end and surrounding text begins.

What does all this have to do with alternating colours in your options?

I'm going to suggest **_you don't need alternating shading, and that the experience would be better served with a clear outline_** around the whole menu item area.

Some examples of this...
Here's a stock Select on a few browsers.
Chrome on MacOS
image

Safari on MacOS
image


Here's a stock menu of any app on MacOS (the Safari menu, in this case)
image
And Chrome
image

You'll notice in all cases there is no alternating colour on the items, and that all have a clear border around the whole menu. You'll also notice that the selected item (in the selects) and hovered item (in the menu) share a very clear contrast to the unselected/unhovered items. I'm not advocating you have to use that strong blue of MacOS. Just noting. 

So that's my suggestion: strip out the alternating shading and add an outline on the whole area that achieves at least 3:1 against item background and page background.

andreancardona:
@mbgower Thank you so much for your input! It all makes sense :) I think since it is a bigger chance that will affect other components I will create an issue to address this.

Let me know how that sounds @laurenmrice?

laurenmrice:
@mbgower @carmacleod Thanks for all of your inputs on this! @andreancardona Yes, lets make a separate issue for this topic because it will affect all other menus we have in the system and may need further discussion/design investigation.

carmacleod:
A bit more on why to delete that live region... :)
"6 results are available, use up and down arrow keys to navigate. Press Enter key to select."

Regarding "use up and down arrow keys to navigate. Press Enter key to select."
- The role="combobox" communicates this fully without adding any words, i.e. as soon as the user hears "combobox", they already know to use up/down arrow and enter (plus some other keys as well, like home, end, space).

Regarding "6 results are available, "
- Browsers are [required to calculate aria-setsize and aria-posinset](https://w3c.github.io/core-aam/#mapping_additional_position) for things like options in a listbox. This allows AT to say things like "Option 2, 3 of 6", "Option 3, 4 of 6". So this info is available if the screen reader user wants it.

(By the way, note that the strings in your example are not the greatest, particularly as the numbers don't line up with what a screen reader would say. Maybe move the long item to the end of the list for now? And someday, maybe use real data, like a list of fruits or animals to choose from, with corresponding nice labels and descriptions?)

> [FIXED IN ANOTHER PR] Please don't use the title attribute on an item unless the item text is too long and is truncated.

Can you please link to the other PR? Thanks!

andreancardona:
> * already know to use up/down arrow and enter (

@carmacleod 

I will definitely make sure to get this fixed. It wasn't forgotten about just ran out of time last week. Working on it today. :)

carrenelloyd:
@andreancardona Just wanted to give you a heads up that you may have at mentioned the wrong person.

andreancardona:
> @andreancardona Just wanted to give you a heads up that you may have at mentioned the wrong person.

oops thank you!

andreancardona:
> A bit more on why to delete that live region... :)
> "6 results are available, use up and down arrow keys to navigate. Press Enter key to select."
> 
> Regarding "use up and down arrow keys to navigate. Press Enter key to select."
> 
> * The role="combobox" communicates this fully without adding any words, i.e. as soon as the user hears "combobox", they already know to use up/down arrow and enter (plus some other keys as well, like home, end, space).
> 
> Regarding "6 results are available, "
> 
> * Browsers are [required to calculate aria-setsize and aria-posinset](https://w3c.github.io/core-aam/#mapping_additional_position) for things like options in a listbox. This allows AT to say things like "Option 2, 3 of 6", "Option 3, 4 of 6". So this info is available if the screen reader user wants it.
> 
> (By the way, note that the strings in your example are not the greatest, particularly as the numbers don't line up with what a screen reader would say. Maybe move the long item to the end of the list for now? And someday, maybe use real data, like a list of fruits or animals to choose from, with corresponding nice labels and descriptions?)
> 
> > [FIXED IN ANOTHER PR] Please don't use the title attribute on an item unless the item text is too long and is truncated.
> 
> Can you please link to the other PR? Thanks!

@carmacleod 

Hey Carolyn! Just testing this with @dakahn and we aren't getting this verbosity while testing with JAWS (did not speak the number of list items with the menu was open but did speak the controls to operate) nor with NVDA (did not mention the number of list items nor the controls to operate it).

- Were you testing with voice over? What are you testing with?

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -332,7 +336,7 @@ export default class ComboBox extends React.Component {
       placeholder,
       initialSelectedItem,
       selectedItem,
-      ariaLabel,
+      ariaLabel, // eslint-disable-line no-unused-vars
```
joshblack: I would imagine we should still be passing along the `ariaLabel` prop to downshift, right? @andreancardona

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -382,126 +386,175 @@ export default class ComboBox extends React.Component {
...
+              }
+
+              if (match(event, keys.Enter)) {
+                toggleMenu();
+              }
+            },
+          });
+
+          // Remove excess aria `aria-labelledby`. HTML <label for> provides this aria information.
+          const inputProperties = Object.keys(inputProps).reduce(
```
joshblack: @andreancardona  Could this be accomplished above by specifying:

```js
const inputProps = getInputProps({
  'aria-labelledby': null,
});
```

tay1orjones (commented):
~I see one violation in AC when combobox is open~

Edit - ignore, I must've been using an outdated/incorrect deploy preview

packages/react/src/components/ComboBox/ComboBox.js
andreancardona: Yes that was deleted on accident 🤦🏽‍♀️

dakahn (changes requested):
I'm not seeing the `aria-selected` violation here on my machine, but I do see a different violation for not having the first list item active. Compare opening our listbox with opening the example [seen here](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html)
image



I'll sync with yall on this

carmacleod:
@andreancardona 
> Hey Carolyn! Just testing this with @dakahn and we aren't getting this verbosity... What are you testing with?

Hi, Andrea and D.A.
I was using:
- https://deploy-preview-7777--carbon-components-react.netlify.app/iframe.html?id=components-combobox--combobox
- focus the combobox, type down arrow, and **_wait_** - i.e. don't keep typing down arrow, just keep listening (if the first item was shorter, you would have noticed it... most comboboxes don't have long items like that as the first item. 😄 )
- JAWS 2021.2102.34 / Chrome 88.0.4324.190
- JAWS 2021.2102.34 / FF 86.0
- NVDA 2020.4 / Chrome 88.0.4324.190
- NVDA 2020.4 / FF 86.0
- VoiceOver on macOS Big Sur 11.2.2 / Chrome 88.0.4324.192
- VoiceOver on macOS Big Sur 11.2.2 / Safari 14.0.3

Here's the output from JAWS Speech history (INS+Space, H) in Chrome:
```
JAWS
ComboBox title edit combo collapsed 
Filter...
To set the value use the Arrow keys or type the value.
[I typed down arrow here]
ComboBox title List box  
An example option that is really long to show what should be done to handle long text  
1 of 6
To move to an item press the Arrow keys.
6 results are available, use up and down arrow keys to navigate. Press Enter key to select.
```

Here's the output from NVDA Speech viewer (INS+N -> Tools -> Speech viewer) in Chrome:
```
ComboBox title  combo box  collapsed  has auto complete  editable  Filter...  blank
[I typed down arrow here]
ComboBox title  list
An example option that is really long to show what should be done to handle long text 
An example option that is really long to show what should be done to handle long text  1 of 6
6 results are available, use up and down arrow keys to navigate. Press Enter key to select. 
```

Here's what VoiceOver says in Chrome:
```
Combobox title Filter..., list box pop up, Menu pop-up combobox
[I typed down arrow here]
An example option that is really long to show what should be done to handle long text, selected (1 of 6)
6 results are available, use up and down arrow keys to navigate. Press Enter key to select.
[and then after a pause, it reads the title attribute (you mentioned this is fixed in another PR - which one? Thx!)]
An example option that is really long to show what should be done to handle long text
```

(Unfortunately, the combobox doesn't work very well in VoiceOver/Safari. The only thing VoiceOver says after dropping down the combo is the live region text. Hoping that removing that will let the combo work better in VO/Safari...)

----

You don't have to run a screen reader to see that a live region is being updated when the combo drops down.
If you go to DevTools and Find (Ctrl+F) "aria-live" you will find the live region.
image

Select this element and right-click and choose Break on -> Subtree modifications
Then drop down the combobox.
Hopefully, you will be able to debug into whatever code is setting this. :)

dakahn:
@carmacleod That live region is generated by Downshift and there is no way to deactivate it. A potential work around to disable it is referenced here
https://github.com/downshift-js/downshift/issues/1238

Seems hacky though

dakahn (approved):
ignore my previous review -- WRONG PR

carmacleod:
@dakahn 
> That live region is generated by Downshift and there is no way to deactivate it. 

There may be a way... https://github.com/downshift-js/downshift/issues/1244  :)

carmacleod:
@tay1orjones You were not imagining things - the violation you saw is still there. Maybe you didn't have anything selected the second time you tried it. :)

@andreancardona I'm sorry I wasn't clearer earlier about where to put `aria-selected="true"` - it should follow focus in a single-select listbox, i.e. it should follow `aria-activedescendant` in a combobox. If you want to mark up "the item that is currently 'selected' in the combobox" (i.e. the listbox option that has the checkmark), then you can use `aria-current="true"` (aria-current is not required in a combobox because the input contains the value, but since this UI shows a checkmark it might be a nice way to tie the visual and the semantic together (please do not use aria-checked for this because that indicates something that can be checked/unchecked directly)). Hopefully this screen snapshot is helpful:

image

tay1orjones (approved):
Sounds like the AC violation may not be valid, but I figured out how to reproduce it.

https://user-images.githubusercontent.com/3360588/109669126-70c04a80-7b37-11eb-8cb8-bb6e922e9533.mov

carmacleod:
@tay1orjones 
> Sounds like the AC violation may not be valid, but I figured out how to reproduce it.

No worries - it's valid. Please see my [previous comment](https://github.com/carbon-design-system/carbon/pull/7777#issuecomment-788973856).

@dakahn 
> I'm not seeing the aria-selected violation here on my machine, but I do see a different violation for not having the first list item active. Compare opening our listbox with opening the example seen here...

It depends on how you open it. :)
If you use <kbd>down/up arrow</kbd>, then yes, the first/last item should be selected.
If you use <kbd>alt+down arrow</kbd>, then no items should be selected.

(i.e. so Accessibility Checker is being a little bit overly-zealous here - I'll open an issue.)
[**Edit:** D.A., I'm not seeing the violation for not having the first list item active. How did you get it?]

@andreancardona Please note that if the user types <kbd>alt+down arrow</kbd>, then the listbox should open without any items selected.
Also, consider changing the <kbd>alt+up arrow</kbd> behavior to close the popup, as described here: https://w3c.github.io/aria-practices/#combobox-keyboard-interaction

laurenmrice (changes requested):
The selected state of a menu option should have a background of `selected-ui` color token. Right now it is the wrong color.
Artboard

dakahn:
@carmacleod That's great! I didn't know about the varying requirements on initial focus given the method of how a keyboard user opens the listbox. I'm honestly not sure how to recreate my AAT error. Let's disregard until I figure out how I did it 😓 

As always thank you so much for your time and energy 🙏🏽

carmacleod:
@joshblack Would you be able to change the title of this PR to:
> update to match ARIA 1.2 criteria

... because this PR doesn't really have anything to do with WCAG 2.1 AA (other than [Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html), but that mostly defers to ARIA). :)  Thanks!

Note that [ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/) reached W3C Candidate Recommendation status today, so this update is timely. :)

Note also that at some point, [downshift will update to ARIA 1.2](https://github.com/downshift-js/downshift/issues/1239), and that will be a breaking change, so you may need to update again... (just a heads-up... not sure if you contribute back to downshift, but if so, this might be one to consider... might save you some pain 😄 ).

packages/components/src/components/list-box/_list-box.scss
```diff
@@ -751,7 +758,7 @@ $list-box-menu-width: rem(300px);
 
   .#{$prefix}--list-box__menu-item--active:hover,
   .#{$prefix}--list-box__menu-item--active.#{$prefix}--list-box__menu-item--highlighted {
-    background-color: $hover-selected-ui;
+    background-color: $selected-ui;
     border-bottom-color: $hover-selected-ui;
```
carmacleod: Should border-bottom-color change, too?
I don't know, but it just looks funny to have 2 different colors here?
```suggestion
    background-color: $selected-ui;
    border-bottom-color: $selected-ui;
```

andreancardona: Updated :) also I just your comment about `aria-current="true"` I will be updating that shortly @carmacleod

carmacleod: Thanks, @andreancardona. I'm not 100% sure how well `aria-current="true"` is supported in comboboxes, so after you put it in, and fix up `aria-selected="true"` to match the aria-activedescendant, please ping me and I will try it again in all the places to see if the `aria-current="true"` is helpful or just confusing.

andreancardona:
> @tay1orjones You were not imagining things - the violation you saw is still there. Maybe you didn't have anything selected the second time you tried it. :)
> 
> @andreancardona I'm sorry I wasn't clearer earlier about where to put `aria-selected="true"` - it should follow focus in a single-select listbox, i.e. it should follow `aria-activedescendant` in a combobox. If you want to mark up "the item that is currently 'selected' in the combobox" (i.e. the listbox option that has the checkmark), then you can use `aria-current="true"` (aria-current is not required in a combobox because the input contains the value, but since this UI shows a checkmark it might be a nice way to tie the visual and the semantic together (please do not use aria-checked for this because that indicates something that can be checked/unchecked directly)). Hopefully this screen snapshot is helpful:
> 
> image

Done! :) @carmacleod Thank you so much for catching that!

laurenmrice (approved):

oartigue:
In version 11, if I add the aria-label property on Combobox, the property is ignored and the aria-labelledby property is automatically generated. When I use the IBM accessibility Checker, I get this error:
`The 'id' "downshift-86-label" specified for the ARIA property 'aria-labelledby' value is not valid for combobox`

oartigue:
In version 11, if I add the aria-label property on Combobox, the property is ignored and the aria-labelledby property is automatically generated. When I use the IBM accessibility Checker, I get this error:
`The 'id' "downshift-86-label" specified for the ARIA property 'aria-labelledby' value is not valid for combobox`

PR #7782, opened 2021-02-11
feat(motion): add duration tokens to motion

### LINKED ISSUE #7738: @carbon/motion should include motion duration tokens
### Summary

When using the `@carbon/motion` package, the developer does not have access to the motion duration tokens[ displayed here](https://carbondesignsystem.com/guidelines/motion/overview#duration-tokens).

An example of [using it in JS for our project](https://github.ibm.com/CDAI-UX-Engineering/motion-dashboard/blob/6e8900aa4ac2a276ff8327f29fc68d6faaf39e17/src/constants.js#L11). Would be great if we can import these from the `@carbon/motion` package itself.

### Justification

It's kind of redundant to include the `@carbon/motion` package if you need to include the duration tokens from the main library. This leads to a double motion import in sass files.

joshblack:
Closes #7738 

Adds support for duration tokens to the motion package.

#### Changelog

**New**

- Add tests for public API of motion package

**Changed**

- Update module style and JS entrypoints to include duration tokens

**Removed**

#### Testing / Reviewing

- Verify token values are correct
- Verify tokens appear correctly in snapshots for the motion package


tw15egan (approved):
LGTM 👍 ✅

tw15egan:
@joshblack looks like snapshots just need to be updated

dakahn (approved):
👍🏽

Issue #7757, opened 2021-02-08
Toggle buttons (active/pressed state)


janhassel:
### Summary

Buttons should allow for an active/pressed state to indicate a feature/function/option is turned on or off, effectively working as a toggle.

### Justification

A big use case for this is for icon-only buttons, for example to indicate text formatting, sidepanel statuses or options:

**Examples**

Keynote
![image](https://user-images.githubusercontent.com/28265588/107227693-01fe3e80-6a1c-11eb-9f88-f5bd247c54a6.png)

Sketch
![image](https://user-images.githubusercontent.com/28265588/107227726-0c203d00-6a1c-11eb-9165-feba958e09a0.png)

VSCode
![image](https://user-images.githubusercontent.com/28265588/107227849-3245dd00-6a1c-11eb-8082-c6fde34940c6.png)

Slack
![image](https://user-images.githubusercontent.com/28265588/107228051-7a64ff80-6a1c-11eb-9f32-d645ec4c2751.png)

Spotify
![image](https://user-images.githubusercontent.com/28265588/107228218-b009e880-6a1c-11eb-8b77-cc39b6b18ec5.png)

### "Must have" functionality

A prop for the button component (like `active`, `toggled`, `pressed` or something like this) to enable and disable the respective styling.
The `aria-pressed` attribute must be set accordingly (https://www.w3.org/TR/wai-aria-practices-1.1/#button, https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#toggle_buttons)

### Available extra resources

Quick exploration with the [text toolbar pattern](https://www.carbondesignsystem.com/patterns/text-toolbar-pattern):
![image](https://user-images.githubusercontent.com/28265588/107228727-4938ff00-6a1d-11eb-9354-5659aeef8b61.png)

From other libraries:
- [Toggle button on Material Design](https://material.io/components/buttons#toggle-button)
- [Toggle button on FluentUI](https://developer.microsoft.com/de-de/fluentui#/controls/web/button) (last example)
- [Selected button on Atlassian Design System](https://atlassian.design/components/button/examples#selected)
- [Dual stateful button on Lightning Design System](https://www.lightningdesignsystem.com/components/buttons/#Dual-Stateful-Button)
- [Toggle button on Bootstrap](https://getbootstrap.com/docs/5.0/components/buttons/#toggle-states)

I could submit a PR for this if the proposal is approved and the visual specification / color token to use is defined.

[...]

PR #7784, opened 2021-02-11
feat(button): add selected state to icon only ghost button

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/7757

Initial PR adding in `selected` support to icon only ghost buttons. More variants could be added later if needed.

#### Changelog

**New**

- `isSelected` prop to `Button`, will only render `isSelected` styles if `kind="ghost"` and `hasIconOnly` props are supplied

#### Testing / Reviewing

Go to `Button` --> `Playground` and check the `isSelected` knob. Nothing should change. Now change the `kind` knob to `ghost` and the icon only button should receive `selected` styles.


aagonzales (approved):
🎉 nice! thanks for adding it so fast.

joshblack (approved):

tay1orjones (approved):

PR #7785, opened 2021-02-11
fix: update listbox titleText proptypes to match Dropdown

### LINKED ISSUE #7779: ComboBox: Support node in titleText attribute
### Summary
In [Dropdown component](https://react.carbondesignsystem.com/?path=/docs/dropdown--default#component-api) titleText  can be `string | node` . But in [combobox component](https://react.carbondesignsystem.com/?path=/docs/combobox--combobox#component-api), titleText  only accepts a `string`.

Can you please add support for node object in titleText attribute?

### Justification
titleText is used as the label for ComboBox. By supporting node object, we can also add tooltips in the label.

emyarod:
Closes #7779

This PR updates the ComboBox titleText prop definition to support nodes in addition to strings. This increases the parity between the listbox components

#### Testing / Reviewing

Confirm no proptype warnings are emitted when adding rich content to a ComboBox title

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.js
+++ b/packages/react/src/components/ComboBox/ComboBox.js
@@ -191,7 +191,7 @@ export default class ComboBox extends React.Component {
      * Provide text to be used in a `<label>` element that is tied to the
      * combobox via ARIA attributes.
      */
-    titleText: PropTypes.string,
+    titleText: PropTypes.node,
 
     /**
      * Specify a custom translation function that takes in a message identifier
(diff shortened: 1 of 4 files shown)


packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -191,7 +191,7 @@ export default class ComboBox extends React.Component {
      * Provide text to be used in a `<label>` element that is tied to the
      * combobox via ARIA attributes.
      */
-    titleText: PropTypes.string,
+    titleText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
tw15egan: I believe a string will still be considered a `node` and not throw any warnings

```suggestion
    titleText: PropTypes.node,
```

emyarod: it was lifted directly from the dropdown but I updated them all across the board

tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):

PR #7801, opened 2021-02-15
fix(component-tokens): only emit static values in emit-component-tokens

### LINKED ISSUE #7800: Tag component with css variables
Hi!

As I see, this example not working: https://www.carbondesignsystem.com/components/tag/code
If I change the theme, the colors don't react, because the css custom variable is undefined and always the fallback is active.
We are facing the same problem, because if we add this to our root scss, it's don't contains the vars for tag:
```scss
@include carbon--theme($carbon--theme--g100, true);
```

Is this intentional? Or is there a mixin/function to get this variables?

For the quick fix, I made a small mixin:
```scss
@mixin get-variables($globalTheme, $tokens) {
	@each $key, $options in $tokens {
		$values: map-get($options, 'values');
		@each $valueObject in $values {
			$theme: map-get($valueObject, 'theme');
			$value: map-get($valueObject, 'value');
			@if $theme == $globalTheme {
				@include custom-property($key, $value);
			}
		}
	}
}
```
Before this we include the tokens of tag (https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/tag/_tokens.scss), and we include our mixin like this:
```scss
.root-element {
  @include get-variables($carbon--theme--g100, $tag-colors);
}
```

And it's will generate:
```css
.root-element {
  --cds-tag-background-red: #ffd7d9;
  --cds-tag-color-red: #750e13;
  --cds-tag-hover-red: #ffb3b8;
...
}
```
What do you think about it?

janhassel:
Closes #7800

The `emit-component-tokens` mixin currently emits all tokens as css custom properties if the feature flag is enabled, resulting in the actual css variable never declared and the colors always falling back to the default, white theme.

The workaround so far was to include the component's styles again, like so:
```scss
:root {
  @include carbon--theme($carbon--theme--g100, true) {
    @include emit-component-tokens($tag-colors);
    @include tags;
  }
}
```

This results in a lot of unnecessary css being generated and can create specificity issues.

#### Changelog

**New**

- Added an optional parameter `$force-static-values` to `get-token-value` mixin.

**Changed**

- Pass `true` to `$force-static-values` from `emit-component-tokens` mixin.

#### Testing / Reviewing

Before, this code should **not** result in the correct tag colors:
```scss
:root {
    @include carbon--theme($carbon--theme--g100, true) {
      @include emit-component-tokens($tag-colors);
    }
  }
```

With the changes in this PR, it should result in the correct tag colors.

In the inspector, all tag colors should be shown as static values: 
image

compared to before
image

### DIFF
--- a/packages/components/src/globals/scss/_component-tokens.scss
+++ b/packages/components/src/globals/scss/_component-tokens.scss
@@ -20,7 +20,10 @@ $enable-css-custom-properties: feature-flag-enabled(
 /// @example get-token-value($component-tokens, 'tag-background-red');
 /// @returns {String} Token value
 /// @group component-tokens
-@function get-token-value($tokens, $name) {
+@function get-token-value($tokens, $name, $force-static-values: false) {
+  $emit-as-css-custom-property: $force-static-values == false and
+    feature-flag-enabled('enable-css-custom-properties');
+
   @if map-has-key($tokens, $name) {
     $config: map-get($tokens, $name);
     $fallback: map-get($config, 'fallback');
@@ -33,7 +36,7 @@ $enable-css-custom-properties: feature-flag-enabled(
         $value: map-get($option, 'value');
 
         @return if(
-          feature-flag-enabled('enable-css-custom-properties'),
+          $emit-as-css-custom-property,
           var(--#{$custom-property-prefix}-#{$name}, #{$value}),
           $value
         );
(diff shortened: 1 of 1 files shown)


joshblack:
@janhassel would we want this to be the default behavior (forcing static values)? This specific mixin is private IIRC so we can totally update it if it should be that behavior by default.

janhassel:
@joshblack I was thinking since the mixin is called far more often without the `force-static-values` parameter than with it, it would make more sense to have it default to `false`.

https://github.com/carbon-design-system/carbon/blob/1de98d0d5887e7a9aaa43b5118a6dad40baa7482/packages/components/src/components/tag/_tag.scss#L48-L126

Open to change it though if you think it makes more sens to have the default set as `true`.

joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #7810, opened 2021-02-16
fix(storybook): filter enabled addons

joshblack:
By default, the feature flag for including our custom flag is false. This leads to the following warning in the console when building storybook:

image

This PR updates this object to filter out false-y values to resolve the error.

#### Changelog

**New**

**Changed**

- Update storybook config to filter out false-y values

**Removed**

#### Testing / Reviewing

- Run yarn storybook in packages/react and verify the error does not occur


tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):
🎉

PR #7818, opened 2021-02-16
fix(loading) small loader not spinning on fixed position

### LINKED ISSUE #2106: [live demo] loading audit bugs
Live demo loading:
- [ ] missing the toggle to turn on an overlay
- [x] in the demo the small loading is wobbling around and not staying in place.

peter310:
Closes https://github.com/carbon-design-system/carbon-website/issues/2106 

change the height and width on small loader class to fix the issue

### DIFF
--- a/packages/components/src/components/loading/_loading.scss
+++ b/packages/components/src/components/loading/_loading.scss
@@ -49,8 +49,8 @@
   }
 
   .#{$prefix}--loading--small {
-    width: rem(16px);
-    height: rem(16px);
+    width: 2rem;
+    height: 2rem;
 
     circle {
       stroke-width: 16;


tw15egan:
Hey there @peter310, this is doubling the overall height/width of the small loading, which I don't think is intended. 

Screen Shot 2021-02-17 at 11 17 09 AM


If you are trying to resolve the comment about
> in the demo the small loading is wobbling around and not staying in place.

I believe we just need to add `position: absolute` to `.bx--loading__svg` as an override on the Carbon website repo, as this is a website-specific bug. Specifically, we can add it here https://github.com/carbon-design-system/carbon-website/blob/master/src/components/ComponentDemo/ComponentDemo.module.scss

Specifically, 

```scss

  [class*='--loading--small'] [class*='--loading__svg'] {
    position: absolute;
  }
```

I'm going to close this out, but feel free to make a PR on the website repo with the suggested change. Thanks!

PR #7821, opened 2021-02-17
fix(guidelines): guidelines-icons-button-label-audit

andreancardona:
Closes (Carbon Website): https://github.com/carbon-design-system/carbon-website/issues/1498
- Adds an aria-label in the Tooltip to fix a11y violation 

- Adds: `ariaLabel: {tooltipText}` to TooltipDefinition component


packages/react/src/components/TooltipDefinition/TooltipDefinition.js
```diff
@@ -77,6 +77,7 @@ const TooltipDefinition = ({
         type="button"
         className={tooltipTriggerClasses}
         aria-describedby={tooltipId}
+        aria-label={tooltipText}
```
joshblack, line 80: Would this end up causes the same text (`tooltipText`) to be announced twice from both `aria-label` and `aria-describedby`

tw15egan:
I just did some further research into this issue, and it seems like the root of the issue is that we are trying to use `TooltipDefinition` as an icon-only button. But changing these to `Button` and not `TooltipDefinition` in `ActionBar.js` in the website repo, I was able to get rid of the errors

```jsx
<Button
        kind="ghost"
        hasIconOnly
        onFocus={() => setIsActionBarVisible(true)}
        onClick={handleDownload}
        tooltipAlignment="center"
        tooltipPosition="top"
        iconDescription="Download SVG"
        className={styles.tooltip}
        triggerClassName={styles.trigger}
        renderIcon={Download16}
      />
```

Screen Shot 2021-02-17 at 12 29 31 PM


@andreancardona I'll throw up a PR so you can take a look and verify if this will close out the linked issue 👍🏻 

PR: https://github.com/carbon-design-system/carbon-website/pull/2144

packages/react/src/components/TooltipDefinition/TooltipDefinition.js
andreancardona: @joshblack good point! `aria-describedby` announces the `tooltipId` which I think should be different from the text

joshblack: @andreancardona I believe the way `aria-describedby` works is that it uses the `id` to find the DOM node from which it computes the description. In this case, using `aria-label` with the `tooltipText` would give the node the accessible name of `tooltipText`. It would also give the node the description of `tooltipText` from `aria-describedby` as it points to a node that contains that as its accessible label.

In the preview, we can see this in the accessibility tab: 

image

Another issue that this can cause is that now the `button` no longer uses the underlying text as the label, so the user would only hear the tooltip text but would not know what that text it is defining.

andreancardona:
Closing this issue with @tw15egan fix :) carbon-design-system/carbon-website#2144

PR #7832, opened 2021-02-17
fix(list-box): update hover active item background color

### LINKED ISSUE #7490: Select/Multiselect active descendent is the same colour as already selected items, affecting Keyboard users
## Environment

> Operating system: Windows10

> Browser: Firefox 84, Chrome 87.0.4280.88, Edge 87.0.664.66

## Detailed description
The active descendent in the multiselect uses the same colour as a selected item. This results in it being difficult for a keyboard user to know where they are in the list (without having to either count key presses or toggling the selection to find where they are).
See the screenshot.

Vanilla Components version
React Components version ^7.25.0
Last updated 11 December 2020

It would be easy to tell the difference between selected items and the currently active descendent.

> What happened instead?
No colour change

2.4.7: Focus Visible
Specifically
> Benefits
    This Success Criterion helps anyone who relies on the keyboard to operate the page, by letting them visually determine the component on which keyboard operations will interact at any point in time.
    People with attention limitations, short term memory limitations, or limitations in executive processes benefit by being able to discover where the focus is located.

While the input does have a focus indicator and is ready to accept text when the user types, the options list part of the component doesn't satisfy this.

1. Navigate a browser to: https://www.carbondesignsystem.com/components/dropdown/usage#live-demo
2. Change the Variant selector to either the Multiselect or Filterable Multiselect
NOTE: the Select and Drop Down variants have this problem too, but because there's only one item selectable it's not as intrusive.
3. Using only the keyboard, open the drop down list from the multiselect
4. Using the arrow keys, make an option active and select it with Enter, repeat this to select multiple items.
5. Using the arrow keys again, make any of the selected items active.

## Additional information

- Screenshots or code
- Notes

emyarod:
Closes #7490

This PR updates the hover state background color for active/selected items in list boxes per new spec guidelines

#### Testing / Reviewing

Select and hover over a multiselect item in the storybook and confirm the correct color tokens are used


aagonzales (changes requested):
The selected+hover looks correct now but the original problem in the issue is still there. When using the keyboard and arrowing through the selected options there is no visual indicator. It should look the same keyboard arrowing through the items as it does when you hover over them.

aagonzales (approved):

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #7838, opened 2021-02-18
fix(OverflowMenu): resolve child.props NPE

### LINKED ISSUE #7837: Having text wrapped inside OverflowMenu causes rendering to fail when minifying due to child props mapping

- [ ] `carbon-components`

## Detailed description

If you have the following code then there is a rendering bug which causes the component to stop working. The text inside the OverflowMenu seems to be causing React to add the text as a child node, and this doesn't have any props and there is no check that is has no props. 

The exception states: `TypeError: Cannot read property 'closeMenu' of undefined`

```
<OverflowMenu> 
Some text here
</OverflowMenu>
```

Yes, specifically `OverflowMenu` component, on line [497](https://github.com/carbon-design-system/carbon/commit/56bbffeff996334b80a2a6fee91e5025cdad164a)

> see changed?

A null check on the `child.props` should ensure the bug is resolved.

Chrome Version 88.0.4324.150 (Official Build) (x86_64)

Bug is from 7.27.0 and was introduced in PR #7456, reverting to 7.25.0 fixes the issue.

> should be aware of?

Bug was discovered by the My Invenio team, an external company working with IAF on integrating their product into IAF.

1. The code block above describes how to recreate the bug locally.

Sandbox is located here: https://codesandbox.io/s/modern-tree-3qt0i

## Additional information

- Screenshots or code
- Notes

emyarod:
Closes #7837

This PR adds a check for overflow menu children props before calling `child.props.closeMenu` to avoid type errors when the props object is not found

#### Testing / Reviewing

Confirm that the component will still render even when its children do not contain props

### DIFF
--- a/packages/react/src/components/OverflowMenu/OverflowMenu.js
+++ b/packages/react/src/components/OverflowMenu/OverflowMenu.js
@@ -494,7 +494,7 @@ class OverflowMenu extends Component {
     const childrenWithProps = React.Children.toArray(children).map(
       (child, index) =>
         React.cloneElement(child, {
-          closeMenu: child.props.closeMenu || this.closeMenu,
+          closeMenu: child?.props?.closeMenu || this.closeMenu,
           handleOverflowMenuItemFocus: this.handleOverflowMenuItemFocus,
           ref: (e) => {
             this[`overflowMenuItem${index}`] = e;


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #7839, opened 2021-02-18
feat(Tag): add pointer on hover if component is interactive

### LINKED ISSUE #7592: Tag: when a Tag is actionable(has an onClick), the pointer icon does not change

- [ ] `carbon-components`

## Detailed description

When a Tag has a filter property, the mouse icon will change when hovering over the Tag.
  `<Tag type="red" title="Clear Filter" filter> Red </Tag>`
If a Tag does not have the filter property but does have an actionable property(onClick), the mouse will NOT change when hovering over the Tag.
`  <Tag type="red" title="Clear Filter" onClick={()=>alert("red")}> Red </Tag>`

I expected the mouse icon to change when actionable in the same way as when filter property is set.

This happens in any browser.
This happens in any version of the Carbon Design System.

1. Create a tag like `<Tag type="red" title="Clear Filter" filter> Red </Tag>`
2. Create a tag like `<Tag type="red" title="Clear Filter" onClick={()=>alert("red")}> Red </Tag>`
3. Hover over each and observe the icon changing(or not).

## Use case

The product has a search input.
The product has three suggestions for searching "Platform", "Runtimes", and "Services".
It shows these as clickable tags for quick filtering.
This is on the Carbon Design System Tag page just above the Live Demo.
https://www.carbondesignsystem.com/components/tag/usage#live-demo

## Work arounds

The child of Tag is a node, so it could be coded as an input.
Styling could be applied, to account for the hover.
Always create a filter tag - and apply display: none; to the filter button when not a filter tag.
All of these are client work and they come with their own problems.

emyarod:
Closes #7592

This PR adds a pointer cursor on hover when an `onClick` handler is supplied to the Tag component

#### Testing / Reviewing

Ensure that the interactive tag example changes cursor on hover

### DIFF
--- a/packages/components/src/components/tag/_mixins.scss
+++ b/packages/components/src/components/tag/_mixins.scss
@@ -11,6 +11,7 @@
   color: $text-color;
   background-color: $bg-color;
 
+  &.#{$prefix}--tag--interactive,
   .#{$prefix}--tag__close-icon {
     &:hover {
       background-color: $filter-hover-color;
--- a/packages/components/src/components/tag/_tag.scss
+++ b/packages/components/src/components/tag/_tag.scss
@@ -168,6 +168,10 @@
     }
   }
 
+  .#{$prefix}--tag--interactive {
+    transition: background-color $duration--fast-01 motion(entrance, productive);
+  }
+
   .#{$prefix}--tag__close-icon {
     display: flex;
     flex-shrink: 0;
--- a/packages/components/src/components/tag/_tag.scss
+++ b/packages/components/src/components/tag/_tag.scss
@@ -147,6 +147,11 @@
     text-overflow: ellipsis;
   }
 
+  .#{$prefix}--tag--interactive:focus {
+    outline: none;
+    box-shadow: inset 0 0 0 1px $focus;
+  }
+
   .#{$prefix}--tag--interactive:hover {
     cursor: pointer;
   }
--- a/packages/components/src/components/tag/_tag.scss
+++ b/packages/components/src/components/tag/_tag.scss
@@ -131,7 +131,8 @@
   }
 
   .#{$prefix}--tag--disabled,
-  .#{$prefix}--tag--filter.#{$prefix}--tag--disabled {
+  .#{$prefix}--tag--filter.#{$prefix}--tag--disabled,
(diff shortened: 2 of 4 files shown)


dakahn (approved):

jnm2377 (changes requested):
Should we also add a bg hover color, and maybe focus state for interactive as well? Since it'll be clickable/keyboard accessible. We have a bg hover color for the `x` in the filterable tag, I would expect it to be the same color for interactive tag.

emyarod:
this is only restoring a reported regression in pointer styles (the tags remain keyboard inaccessible, keeping the current behavior). I don't have the context from the Slack conversation referenced in the original ticket but from what I understand, those style changes have not been specced out yet. but if we want to roll in the regression fix and the interactive style updates together I will need a spec from design. is that how we should proceed?

tw15egan:
cc @aagonzales 

 https://github.com/carbon-design-system/carbon/issues/7592#issuecomment-785248474

emyarod:
yeah I reached out to her about this and I'm watching the thread. not sure if this means this regression fix needs to be blocked

tw15egan:
@emyarod Looks like @aagonzales has given this the go-ahead. Hover looks good, but doesn't seem to be focusable via keyboard. Should we forward any `onClick` handlers to an internal `onKeyDown` / add a `tabindex`?

emyarod:
yeah I'm currently discussing potential hover/active states for interactive tags with her. this PR was initially to restore the reported regression for hover state but depending on the discussions then more style and functionality will need to be introduced

aagonzales (approved):
Perfect, I think the hovers are a nice addition. Thanks for adding those!

packages/components/src/components/tag/_mixins.scss
```diff
@@ -11,6 +11,7 @@
   color: $text-color;
   background-color: $bg-color;
```
tw15egan, line 12: Should we add `transition: background-color $duration--fast-01 motion(entrance, productive);`?

aagonzales: Oh yeah that would be nice. I always forget about motion.

jnm2377 (approved):
Looks great! 🔥

tw15egan (approved):
LGTM 👍 ✅

PR #7840, opened 2021-02-18
fix(elements): use explicit exports from motion to prevent duplicates

joshblack:
This came up during our 10.29 release. The motion package and type package both share `unstable_tokens` exports. Re-exporting all members from the `motion` package caused duplicates to appear in the ESM build. This PR addresses that by using explicit exports from the motion package

#### Changelog

**New**

**Changed**

- Update elements package to re-export explicit bindings from the motion package

**Removed**

### DIFF
--- a/packages/elements/src/index.js
+++ b/packages/elements/src/index.js
@@ -162,4 +162,13 @@ export {
   spacing,
 } from '@carbon/layout';
 export * from '@carbon/themes';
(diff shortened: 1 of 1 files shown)


joshblack:
@tw15egan @andreancardona just a heads up, going to force merge this for release this afternoon. Let me know if this is a problem, happy to revert if something comes up 👀

Issue #4555, opened 2019-11-04
[condensed grid] wrapping columns collapse


photodow:
## What package(s) are you using?

- [ ] `@carbon/grid`
- [ ] `carbon-components`

## Detailed description

When using the condensed grid the vertical spacing is put on the `bx--row` class. This doesn't allow for wrapping of columns. For instance when you have `16` columns setup at `(1, 16)`, and on the medium breakpoint they fall back to `(1, 8)`. To get this the columns begin to wrap. You now have 2 rows of 8, but because the `bx--row` doesn't wrap with it the spacing in between the columns collapse and butt up against each other.

This is actually two rows of wrapped columns that demonstrates the problem.
![image](https://user-images.githubusercontent.com/3793636/68149716-04faad00-ff04-11e9-8ef9-642944b5108e.png)

> Is this issue related to a specific component?

condensed grid

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I expect each column to manage the vertical spacing so when they wrap, they get the proper spacing.

Thinking out loud here we could replace [L323-L329](https://github.com/carbon-design-system/carbon/blob/master/packages/grid/scss/_mixins.scss#L323-L329) with something like this...

```scss
.#{$prefix}--grid--condensed [class*='#{$prefix}--col-'] {
    padding-top: $condensed-gutter / 2;
    padding-bottom: $condensed-gutter / 2;
}
```

> What version of the Carbon Design System are you using?
10.6.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
IBM Dotcom

## codesandbox example

https://codesandbox.io/s/unruffled-raman-6rl7u?fontsize=14

tay1orjones:
Looks like this was resolved back in https://github.com/carbon-design-system/carbon/pull/4960

https://codesandbox.io/s/serene-williams-s6pgw?file=/src/index.js

PR #7851, opened 2021-02-19
feat(grid): add CSS Grid Experiment

joshblack:
This is currently a draft / WIP Pull Request to see what we could pull off with CSS Grid in terms of grid primitives, subgrid support, and layouts.


joshblack:
@alisonjoseph @tay1orjones thought you all might be interested in this, in particular:

- [Example page](https://github.com/carbon-design-system/carbon/pull/7851/files#diff-6edf42ead30bf5c696e1d185ada777b5d7d9db2a880a16ed6c2e4aeb04a5d367)
- [CSS Grid styles](https://github.com/carbon-design-system/carbon/pull/7851/files#diff-91b6db26ff67f780f41d81bef280c5bc084ece7f1589234b7806271878199526)

I think the notable exception is that, since we use `column-gap` we can't support the `narrow` grid mode. We can however drop all the gutter and offer classes that add back in the missing gutter from the custom property.

tay1orjones:
This is awesome! 🎉  After taking a look, could we alternatively keep the wide mode/normal gutter but incorporate a narrow grid mode modifier that applies a negative margin to each child? In addition we could allow this at the per-column level to allow mixing of modes.

```css
.bx--css-grid--narrow > *,
.bx--col-narrow {
    margin-left: calc(var(--cds-grid-gutter)/2*-1);
}
```

In both cases the gutter to the right of the column would remain unaffected per the spec. It would be an intentional violation of the css grid and would be visible in the grid mode explorer overlay in dev tools (see `span 2` below), but it would conform to the design spec.

image

tay1orjones (approved):
I added the following here:

* Direct support for narrow mode with the "hang by default" negative left margin. 
* 12 column grid modifier/helper class
* Proper support for mixing modes via subgrid
  * Condensed subgrid has proper gutters when placed in a wide or narrow parent grid
  * Narrow subgrid has proper gutters when placed in a wide or condensed parent grid
  * Wide subgrid has proper gutters when placed in a narrow or condensed parent grid
* Additional simple layout exploration/examples on `/layouts` page

tay1orjones:
I also just added row-gap to the condensed grid so that consistent vertical guttering is applied when columns wrap as referenced in https://github.com/carbon-design-system/carbon/issues/4555

joshblack:
@andreancardona @jnm2377 let me know if you have any questions!

jnm2377 (approved):

andreancardona:
@tay1orjones Do we care at all about mobile?  Span 1's are getting cut off a bit
Screen Shot 2021-03-17 at 4 57 41 PM

tay1orjones:
@andreancardona great catch, thanks! I just pushed up a fix.

PR #7854, opened 2021-02-19
feat(project): add styles experiment

joshblack:
This PR adds in a private package at `packages/styles` called `@carbon/styles`. This package is not meant to be published unless we confirm as a team that we want to move our styles to a scoped package.

As an experiment, this package is paired alongside our migration to Sass Modules to help influence/guide how that transition will happen. It is designed in a way that we could rename the package, merge it into `carbon-components`, etc. at the end of the experiment.

This initial PR sets up the first thoughts around what our first-class Sass API should look like, including bringing in previous Sass Modules migrations for our elements packages, along with introducing a migration document for `carbon-components` to assist with documenting public files and their exports.

#### Changelog

**New**

- Add `packages/styles` package with initial files and tests
- Add initial migration document for `carbon-components` to document public API in terms of files and module members to make sure we take care of them in the migration

**Changed**

**Removed**


joshblack:
FYI @carbon-design-system/developers-system if you're interested 👀

joshblack:
Let me know if you have any questions @emyarod @tw15egan !

tw15egan (approved):

emyarod (approved):

PR #7866, opened 2021-02-22
docs(Storybook): update storybook to use carbon styles

tw15egan:
Updates our Storybook environment to use Carbon styles to better match some of the other Carbon storybook environments out there. Wonder if we also want to add some sort of intro page like [Carbon Charts](https://carbon-design-system.github.io/carbon-charts/?path=/story/intro--welcome)?

#### Changelog

**New**

- Added Carbon favicon
- Added Carbon theme file

#### Testing / Reviewing

Everything should work the same, let me know if we want to tweak any of the colors/styles, more info can be found in [the storybook documentation](https://storybook.js.org/docs/react/configure/theming#create-a-theme-quickstart)


joshblack (approved):
Looks awesome!

jnm2377 (approved):
So cool! I think an intro page would be really dope, but not a blocker!

aagonzales:
This looks nice! I like it reflecting the other storybooks and the carbon brand more. A intro page would be nice too so we don't have to link to accordion or another component when sharing a link to it as well. But I agree that's not a stop ship.

Some small visual updates if you're looking for recommendations:
- Change the tab label's and knob label's font-weight to `semiBold` / 600.
- Change the selected tab text color to Gray100 `#161616` to closer match the style of our [component tabs](https://react.carbondesignsystem.com/?path=/story/tabs--default) 
- Change the border dividers to use Gray20 `#e0e0e0`
- Change the selected page background in the left panel to be Gray20 `#e0e0e0` and use Gray100 `#161616` text.

tw15egan:
@aagonzales Thanks! I'll make those updates. If you have any quick mocks for an intro page I'd be happy to add that in, but we can add that in later if you're busy 🙂

tw15egan:
@joshblack @jnm2377 I grouped the components into categories, so you may want to take another look and see if this is something we would want to do. We could probably group a few more components together (Tooltip variants stand out), but figured this was a quick easy start to see if we wanted to stick with this

@aagonzales style changes made and welcome page added 🎉

jnm2377:
Oooh, I really like the groups. It feels a lot more organized. Esp having deprecated and experimental in their own sections. This is great!!! Thanks for doing this!  🔥 🎉 💯

aagonzales (approved):
Yes! That looks great

joshblack:
@tw15egan love it 🔥

Issue #7791, opened 2021-02-12
Table headers hover state is too narrow


ConradSchmidt:
## What package(s) are you using?

- [x] `carbon-components`
- [x] `carbon-components-react`
- [x] `carbon-components-angular`

Not affected:
- [x] `@carbon/vue`

## Detailed description

> Describe in detail the issue you're having.
Table headers having no or not enough spacing to the left when hovering. It is quite different over all the versions and use cases

### [Vanilla JS](https://the-carbon-components.netlify.app/?nav=data-table) (Select)
![image](https://user-images.githubusercontent.com/3808948/107754854-5b1ee880-6d22-11eb-8f40-959f13181657.png)
**8px instead of 16px**

### [React](https://react.carbondesignsystem.com/?path=/story/datatable--usage)
![image](https://user-images.githubusercontent.com/3808948/107755051-99b4a300-6d22-11eb-90ef-121771d8d762.png)
**0px instead of 16px**

### [Angular](https://angular.carbondesignsystem.com/?path=/story/components-table--basic)
![image](https://user-images.githubusercontent.com/3808948/107755190-c9fc4180-6d22-11eb-8b75-639a45a9a871.png)
this is good, but in [other uses cases](https://angular.carbondesignsystem.com/?path=/story/components-table--with-expansion) it is still broken:
![image](https://user-images.githubusercontent.com/3808948/107755405-0b8cec80-6d23-11eb-9b14-5f6797a63450.png)
**0px instead of 16px** - this is a custom header, but in the [next use case](https://angular.carbondesignsystem.com/?path=/story/components-table--with-dynamic-content) with the same "custom header" is fine again

#### [Vue](http://vue.carbondesignsystem.com/?path=/story/components-cvdatatable--name-only-sortable)
![image](https://user-images.githubusercontent.com/3808948/107755935-c0270e00-6d23-11eb-9c0e-61c099cf5946.png)
**this is good**

> Is this issue related to a specific component?

- [x] `DataTable`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Table headers should have 16px to left as described here:
https://www.carbondesignsystem.com/components/data-table/style#columns
and shown here:
https://www.carbondesignsystem.com/components/data-table/usage#sorting
![Hover spec](https://www.carbondesignsystem.com/static/fe75eed08eed65bfef928f8108c26bea/3cbba/datatable-sorting.png)

> What browser are you working in?

Chrome

[...]

PR #7888, opened 2021-02-24
fix(DataTable): fix padding issue when only some headers are sortable

tw15egan:
Ref https://github.com/carbon-design-system/carbon/issues/7791
Ref https://github.com/carbon-design-system/carbon/pull/7726#issuecomment-784704247

#### Changelog

**Added**

- Specific override for when `isSortable` is set on `DataTable` but some specific Headers are not sortable

**Changed**

- Moved previous sort styles from `core` to `sort` scss file

#### Testing / Reviewing

You can test this by changing the header map inside `TableHead --> TableRow` in the `DataTable-sorting.story` to the code below. Also, try removing `isSortable` from `DataTable` to test different variations. 

```jsx
              {headers.map((header, i) => {
                const isSort = i === 2;
                return (
                  <TableHeader
                    key={header.key}
                    {...getHeaderProps({ header, isSortable: isSort })}>
                    {header.header}
                  </TableHeader>
                );
              })}
```


packages/components/src/components/data-table/_data-table-sort.scss
```diff
@@ -56,10 +58,19 @@
...
+    padding-right: $spacing-05;
+    padding-left: $spacing-05;
+  }
+
   // -------------------------------------
   // Th > Button > Span (span required for flex bugs in Safari)
   // -------------------------------------
-  .#{$prefix}--data-table--sort th .#{$prefix}--table-sort__flex {
+  .#{$prefix}--data-table--sort th .#{$prefix}--table-sort__flex,
+  th .#{$prefix}--table-sort__flex {
```
emyarod, line 72: is this extra selector intentional?

emyarod (commented):
just to make sure I have the context, this reverts/updates #7726 is that right?

packages/components/src/components/data-table/_data-table-sort.scss
tw15egan: It updates #7726, I put those styles in the `core` scss file, this PR moves them to the `sort` file with the preexisting selectors.

This selector is needed because in some instances you can pass `isSortable`  to the `TableHeader`, which does not add a global `.#{$prefix}--data-table--sort` class. Passing `isSortable` to `DataTable` gives us that class.

emyarod: I see so if L73 is needed then L72 would be redundant and we should be able to just simplify it down to `.#{$prefix}--table-sort__flex` right?

tw15egan: Yeah most likely, let me just verify that the specificity is sufficient in all variants

tw15egan: Updated 👍🏻

tay1orjones (approved):

emyarod (approved):
looks good to me

PR #7893, opened 2021-02-24
chore(social): update social share link

tw15egan:
Updates the image that appears when a link to Storybook is shared

#### Testing / Reviewing

See if the new image shows up properly


joshblack:
@tw15egan what's a good way to test the social link? I tried dropping the link in Slack but it didn't seem to render 🤔

emyarod (approved):
looks good to me

image

do we want to use a different image host/self host these images?

tw15egan:
@emyarod I tried with a relative link to the assets folder inside storybook and it didn't work properly, open to ideas

emyarod:
not sure where these assets are from but I assumed these assets were already hosted online by Carbon/IBM somewhere. if anything we could also continue hosting on IBM GitHub like the old image

tw15egan:
@emyarod updated, good catch

@joshblack the deploy preview should have the updated image

joshblack (approved):

PR #7894, opened 2021-02-25
fix(tab): tabpanel is tabbable when content has no tabbable content

### LINKED ISSUE #7078: Tabs control: tabpanel needs to be in tab order if there's no focusable tabpanel content
Steps:
- Open the [Tabs control example](https://carbon-components-react.netlify.app/iframe.html?id=tabs--default)
- Type the <kbd>Tab</kbd> key - focus moves to "Tab Label 2"
- Type the <kbd>Tab</kbd> key again - focus should move into the tabpanel (i.e. the div containing "Content for second tab...") but instead, it goes to the browser's location bar.

The [APG Guidance for Tabs component](https://w3c.github.io/aria-practices/#keyboard-interaction-20) says that if focus is in the `tablist` and the user types the <kbd>Tab</kbd> key:
> ... move focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable.

The Carbon Tabs component doesn't need to know what focusable element inside the tabpanel contains meaningful content. :)
It only needs to know whether or not _any_ of the `tabpanel` content is focusable. If none are, then the `tabpanel` itself needs to be focusable (i.e. have `tabindex="0"`).

Here's how one Tabs component decided on a prop for this: point 2 in https://github.com/juanca/react-aria-components/pull/6
based on this comment in issue https://github.com/w3c/aria-practices/issues/323#issuecomment-400011586.

Happy to talk this through before anything is coded.

tay1orjones:
Closes #7078

~Add `hasNoFocusableContent` prop to Tab to inform if the tabpanel container should be focusable. This preserves existing functionality and allows users to opt-in to the new focus behavior when their tab content is text only or has no focusable elements/content.~

Add `useTabbableContent` to TabContent to automatically determine if the contents contains a tabbable element.

#### Changelog

**New**

- Add `useTabbableContent` hook to TabContent

**Changed**

- Update Tabs story to have a tabbable button in the contents of one tab
- Refactor TabContent tests to use testing-library and test for tabindex/hook functionality

#### Testing / Reviewing

1. Open the [default Tabs story](https://deploy-preview-7894--carbon-components-react.netlify.app/?path=/story/components-tabs--default)
1. Type the <kbd>Tab</kbd> key - focus should move to the tabpanel/content container.
1. Type <kbd>Shift + Tab</kbd> - focus moves back to "Tab Label 1"
1. Type <kbd>→</kbd> - focus should move to "Tab Label 2"
1. Type the <kbd>Tab</kbd> key again - focus should move to the button inside the content (the tabpanel should not be focused as the content contains a focusable element)


joshblack (commented):
Would we want to do this automatically for users, potentially? Like in an effect. Something like:

```jsx
function TabContent(props) {
  // ...
  const ref = useRef(null);
  const hasFocusableContent = useFocusableContent(ref);

  // ...
  return <div role="tabpanel" tabIndex={hasFocusableContent ? 0 : undefined}>...</div>
}

function useFocusableContent(ref) {
  const [hasFocusableContent, setFocusableContent] = useState(false);

  useEffect(() => {
    const elements = getFocusableElements(ref.current);
    if (elements.length > 0) {
      setFocusableContent(true);
    }
  }, [ref]);

  return hasFocusableContent;
}
```

tay1orjones:
I dig the magic there for sure. It would also work if content is passed in via `renderContent`. 

My only concern would be performance of `getFocusableElements(ref.current);`. Iterating over every element recursively against a known list of focusable elements on every render feels ... expensive 😬

tay1orjones:
You know what... I forgot about `querySelectorAll` 😅  Yeah this sounds good, I'll update the PR.

emyarod:
this might be helpful for your querySelectorAll https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/internal/keyboard/navigation.js#L64-L72

tay1orjones:
Thanks @emyarod, I hadn't seen that. I think it's `selectorTabbable` that I really wanted, "focusable elements" was a misnomer. 

I modified the PR body to reflect the new approach. Also added some tests and refactored the tests off enzyme.

packages/react/src/components/TabContent/TabContent.js
```diff
@@ -6,24 +6,46 @@
...
+ */
+function useTabbableContent() {
+  const [hasTabbableContent, setHasTabbableContent] = useState(false);
+
+  const ref = useCallback((node) => {
+    if (node !== null) {
+      // we don't need to know which element or how many are tabbable, just that there is _something_ tabbable
+      node.querySelector(selectorTabbable)
+        ? setHasTabbableContent(true)
+        : setHasTabbableContent(false);
```
emyarod:
```suggestion
      setHasTabbableContent(node.querySelector(selectorTabbable));
```

maybe something like this? seems simpler than a ternary statement

jnm2377 (approved):
LGTM! Tab content is now tabbable if there's no interactive elements. Thanks for adding the comments to the code too! Super helpful! 🔥

joshblack:
@tay1orjones definitely get what you mean, it might be that this abstraction doesn't make see if the tabpanel also needs to be explicitly labeled so would totally understand moving forward with the prop approach if this wouldn't be a good fit to do automatically.

tay1orjones:
@joshblack It works quite nicely. Explicitly labelling the tabpanel feels like a very unlikely use case, but we could allow it by placing the automatic tabindex above `{...others}` so it's overridable?

packages/react/src/components/TabContent/TabContent.js
```diff
@@ -6,24 +6,44 @@
...
+
+  return [hasTabbableContent, ref];
+}
+
 const TabContent = (props) => {
   const { className, selected, children, ...other } = props;
   const tabContentClasses = classNames(`${prefix}--tab-content`, {
     [className]: className,
   });
+  const [hasTabbableContent, ref] = useTabbableContent();
```
joshblack: Love this callback pattern, seems super clever for running code when a component "mounts". I wonder if there would there be an easy way to have this hook accept a ref just to make composition with hooks that use refs a little easier?

Something like:

```js
function useTabbableContent(ref) {
  const [hasTabbableContent, setHasTabbableContent] = useState(false);

  // useEffect on the server, useLayoutEffect on the client
  useIsomorphicEffect(() => {
    if (ref.current) {
      setHasTabbableContent(ref.current.querySelector(selectorTabbable));
    }
  });

  return hasTabbableContent;
}
```

I _think_ these would run at similar times but wasn't sure 🤷

tay1orjones: Great idea! Updated

andreancardona (approved):
@tay1orjones  this looks awesome!

I'm so so sorry for the review delay... 👀

PR #7900, opened 2021-02-25
fix(button): set aria-pressed to props.isSelected for icon-only ghost

janhassel:
Related #7757, #7784

Following the [WAI-ARIA Authoring Practices for toggle buttons](https://www.w3.org/TR/wai-aria-practices-1.1/#button), this PR sets the button's `aria-pressed` attribute according to `props.isSelected` (if it is icon-only and ghost).

#### Changelog

**New**

- added `aria-pressed` attribute to button icon-only ghost buttons to reflect `props.isSelected`

**Changed**

- Updated jest snapshots that use icon-only ghost buttons

#### Testing / Reviewing

- When `props.isSelected` is not passed, the button should not have the `aria-pressed` attribute.
- When `props.isSelected = false` is passed, the button should have `aria-pressed="false"`
- When `props.isSelected = true` is passed, the button should have `aria-pressed="true"`
- When `props.isSelected = true` is passed but the button is not icon-only and ghost, the butto should not have the `aria-pressed`attribute


joshblack (approved):

emyarod (approved):
looks good to me

tw15egan (approved):
LGTM 👍 ✅

PR #7901, opened 2021-02-25
fix(toast-notification): remove default value from optional caption prop

janhassel:
According to the toast notification's [component guidelines](https://www.carbondesignsystem.com/components/notification/usage/#toast-notifications) and the propTypes, the `caption` prop (intended for timestamp) is optional. However, currently it defaults to "provide a caption" which
a) makes it seem to be required (as other component usually do this for missing, required props such as titles), and
b) require the developer to pass an empty caption to the component in order to remove it, which is unusual for an optional prop

#### Changelog

**Removed**

- removed default value ("provide a caption") for `props.caption` in `ToastNotification`

#### Testing / Reviewing

- If no caption is passed, there should be no rendered caption
- If a caption is passed, it should be rendered


packages/react/src/components/Notification/Notification.js
```diff
@@ -399,7 +399,7 @@ ToastNotification.propTypes = {
 ToastNotification.defaultProps = {
   kind: 'error',
   title: 'provide a title',
-  caption: 'provide a caption',
+  caption: '',
```
joshblack: Would it make sense to remove the `caption` as a default prop?

janhassel: Yeah, good point!

tw15egan:
Everything looks good code-wise, just want to get some design input if the padding is correct when there is no caption 

Screen Shot 2021-02-25 at 9 56 13 AM

jnm2377 (approved):

tay1orjones (approved):

joshblack (approved):

kingtraceyj (changes requested):
the padding from the bottom is looking a little large. it should be 16px.
current:
image

updated:
image

janhassel:
@kingtraceyj updated 👍

tw15egan (approved):
LGTM 👍 ✅

kingtraceyj (approved):
looks good to me. thanks for the update!

PR #7904, opened 2021-02-25
fix(RadioButton): remove empty span if label text is falsey

emyarod:
This PR removes an empty span element when the radio button labelText is falsey. There was a case where a falsey labelText value together with `hideLabel="true"` would render an empty span with `.bx--visually-hidden` styles

image

### Testing/Reviewing

Clear the label in the radio button storybook and confirm that the empty span is no longer rendered


andreancardona (approved):
Looks good!

tw15egan (approved):

PR #7905, opened 2021-02-25
fix(Tabs): prevent tablist role from stealing focus

### LINKED ISSUE #7892: Tabs highlight on Tab click in safari, but none in chrome/firefox

- [ ] `carbon-components`

## Detailed description

Tabs component will highlight entire container when clicking on a tab in safari

Tabs, Tab

> see changed?

It should be consistent across browsers, assuming this means no highlight

Chrome, FF, Safari

Repro in Safari

10.27, repro in latest

> should be aware of?

IBM cloud docs

1. Go to https://react.carbondesignsystem.com/?path=/story/tabs--default on Safari
2. Click on any tab
3. 
image

emyarod:
Closes #7892

when the `tablist` role is applied to an element, it becomes focusable on click. In Safari, clicking on a Carbon tab within the tab container was stealing focus away from the tab and adding browser default outlines to the element with `role="tablist"`. This PR overrides the browser styles and explicitly calls `focus()` on the selected tab to ensure focus remains on the selected tab (non-Safari browsers did not need this to keep focus on the selected tab)

#### Testing / Reviewing

Ensure the correct element is focused in Safari after selecting tabs via mouse and ensure no behavioral regressions appear in non-Safari browsers


tw15egan (approved):
LGTM, no longer seeing (incorrect) focus in Safari 👍 ✅

joshblack (approved):

PR #7917, opened 2021-02-26
fix(tabs): remove Safari specific overflow indicator margins

### LINKED ISSUE #7906: Tab overflow indicator positioning is misaligned in Safari
It seems that some time in the last 5 months, Safari updated how it calculates margins or positions elements with negative margin so that the Safari-specific rules put in place for tab overflow indicators are outdated

image

emyarod:
Closes #7906

This PR removes previously needed margins for the tab overflow indicators in Safari

#### Testing / Reviewing

Reduce the width of the viewport so that scrolling tabs are enabled and the indicators appear. This will not be testable in the storybook due to custom properties support, so this will need to be viewed with custom properties disabled locally


tw15egan (approved):
LGTM 👍 ✅

joshblack (approved):

PR #7918, opened 2021-02-26
feat(Notification): add onClose support

### LINKED ISSUE #7859: carbon-components-react/ToastNotification: Add `onClose` prop
- [x] `carbon-components`

## Detailed description

Will you please add an `onClose: () => void` prop to the `ToastNotification` react component?

It should replace the inner setState so that it's possible to implement the animation behaviour defined in your design guidelines without ugly hacks that hijack the close button handler.

I've solved this for now by removing the close button & hooking our handler to `onClick`, but our UX designer is not amused and would like to keep the close button instead.

emyarod:
Closes #7859

This PR adds support for the `onClose` method on ToastNotification and InlineNotification. Similar to the ComposedModal the default close behavior can be avoided if the method returns `false`

#### Changelog

**New**

- `onClose` support in ToastNotification and InlineNotification

#### Testing / Reviewing

Confirm that the `onClose` action appears in the storybook action logger. When editing the return value of the method locally, returning `false` should prevent the notification from closing


tw15egan (approved):
LGTM 👍 ✅

packages/react/src/components/Notification/Notification.js
```diff
@@ -266,25 +267,33 @@ export function ToastNotification({
     [`${prefix}--toast-notification--${kind}`]: kind,
   });
 
+  const handleClose = useCallback(
+    (evt) => {
+      if (!onClose || onClose(evt) !== false) {
+        setIsOpen(false);
+      }
+    },
+    [onClose]
```
joshblack: I think we could have this be a function declaration versus `useCallback`. The rationale would be that the identify of `onClose` is likely to change between renders (if it's inline as an arrow function or a function declaration in a functional component). Since it is likely to change, then we are doing the work of equality check and re-creating the closure/function every render instead of just allocating the closure/function. Curious what your thoughts are 🤔

packages/react/src/components/Notification/Notification.js
```diff
@@ -266,25 +267,33 @@ export function ToastNotification({
     [`${prefix}--toast-notification--${kind}`]: kind,
   });
 
+  const handleClose = useCallback(
+    (evt) => {
+      if (!onClose || onClose(evt) !== false) {
```
joshblack: For the `onClose(evt) !== false` branch, would we more so want people to be able to configure visibility of the close button if they aren't able to dismiss/close the notification?

packages/react/src/components/Notification/Notification.js
```diff
@@ -266,25 +267,33 @@ export function ToastNotification({
...
-      onCloseButtonClick(event);
+    const timeoutId = window.setTimeout((evt) => {
+      handleClose(evt);
     }, timeout);
 
     return () => {
       window.clearTimeout(timeoutId);
     };
-  }, [onCloseButtonClick, timeout]);
+  }, [handleClose, timeout]);
```
joshblack: This might be related to the `useCallback` note above, but if we save the `onClose` prop in a ref we might be able to do:

```js
function ToastNotification({
  // ...
  onClose
}) {
  const savedOnClose = useRef(onClose);

  useEffect(() => {
    savedOnClose.current = onClose;
  });

  useEffect(() => {
    if (!timeout) {
      return;
    }

    const timeoutId = window.setTimeout((event) => {
      setIsOpen(false);
      if (savedOnClose.current) {
        savedOnClose.current(event);
      }
    });
    return () => {
      window.clearTimeout(timeoutId);
    };
  }, [timeout]);
}
```

packages/react/src/components/Notification/Notification.js
emyarod: this is for the visibility of the notification itself, similar to how the visibility of modals is controlled rather than the visibility of the modal close button

packages/react/src/components/Notification/Notification.js
emyarod: makes sense together with the previous comment, updated

packages/react/src/components/Notification/Notification.js
joshblack: @emyarod right, the way I interpreted it was that if the `onClose` prop is called and returns `false` then it will prevent the close from happening. Is that correct?

If so, then I was just asking that if the notification is unable to be dismissed (the `onClose` prop would return `false`) it seems like there would need to be some indicator of that. I think from my comment that was hiding the button if someone couldn't dismiss the notification but it could be something else, for sure.

emyarod: wouldn't this be left up to the consumer to resolve? similar to how we provide this API in our modals

joshblack (approved):

kubijo:
Hi. I've just upgraded to `carbon-components-react@7.31.0` and this doesn't seem to be included. Do you have a clue which release might get this?

Edit... I might have misread the state of affaires… I thought it's already merged, but now I see that it's just approved… the question holds though :-)

tw15egan:
@emyarod is this good to go?

PR #7928, opened 2021-03-01
feat: add FloatingMenu to dropdown

### LINKED ISSUE #7565: Dropdown menus should overlay scrollable areas
### Summary

If a dropdown is used within a scrollable area they will be cut off by the surrounding container. This leads to broken layouts or even actions impossible to access.
Example (dropdown hidden by scrollable area and actions can't be accessed):
dropdown-scroll-issue
Try it [here (IBM internal)](https://pages.github.ibm.com/conrad-schmidt/wkc-patterns-showcase/#/user-modal)

Currently only the DatePicker supports an `appendTo` option, which would solve this issue.
This issue can't be resolved from the `outside` (css, add-ons, js, ...) - only the Carbon components itself can fix this issue, that is why it is so hard to deal with. 

### Justification

This issue occurs in several products:
- Watson Knowledge Catalog (WKC)
- Cloud PAK for Data (CPD)
- Also covered in this [Slack discussion](https://ibm-analytics.slack.com/archives/CPZH52RDE/p1607629066050400)
- And other (see https://github.com/carbon-design-system/carbon/issues/6288) 

The dropdown should overlay the scrollable container.
image

### "Must have" functionality

The dropdown should overlay the scrollable container (see above).

As soon as possible, because there features which should be shipped by **begin of February**. Currently the design is **asked to design around this issue**, which shouldn't be the case.

_NB: The Carbon team will try to work with your timeline, but it's not
guaranteed. The earlier you make a request in advance of a desired delivery
date, the better!_

### Available extra resources

[Modal with dropdown issue(IBM internal)](https://pages.github.ibm.com/conrad-schmidt/wkc-patterns-showcase/#/user-modal)
[Dialog appendTo plugin](https://github.com/carbon-design-system/carbon/blob/3361b8046013eb95d4fc88107384d8202f6c2468/packages/react/src/components/DatePicker/plugins/appendToPlugin.js)
[Existing issue](https://github.com/carbon-design-system/carbon/issues/6288)

ConradSchmidt:
Closes #7565

Adds a FloatingMenu to the dropdown component, to use it within scrollable areas

#### Changelog

**New**

**Changed**

- Dropdown with FloatingMenu

**Removed**

#### Testing / Reviewing

- open dropdown story "Overflow" and "Overflow in Modal" to check if the dropdown menu is shown properly (not cut off by the overflow area)

#### Open points
1. Should this be a default or triggered by a property?
2. Should this be added to multi select?

3. This is required by the FloatingMenu, should these properties optional? ([Code](https://github.com/carbon-design-system/carbon/pull/7928/files#diff-b645c78914204385341704484c6145f99bbfa3765701ebc9e2c7bf1154212133R189))
```js
          menuRef={() => {}}
          menuOffset={() => {}}
```
4. This is needed for the modal to work (otherwise it closes on select) - any other ideas to fix this? ([Code](https://github.com/carbon-design-system/carbon/pull/7928/files#diff-b645c78914204385341704484c6145f99bbfa3765701ebc9e2c7bf1154212133R193))
```js
          <ListBox.Menu
            onMouseDown={e => e.stopPropagation()}
```

5. The FloatingMenu works fine except that it is not taking scrolling into account and should be addressed in another issue
carbon-dropdown-scroll-issue
A simple solution for the dropdown would be to close it on scroll, to avoid updating the menu via JS (because this is laggy). For other components like the Tooltip this could work as well, but I guess this needs to be discussed in the separate issue.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown-story.js
+++ b/packages/react/src/components/Dropdown/Dropdown-story.js
@@ -5,12 +5,14 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { action } from '@storybook/addon-actions';
 import { withKnobs, boolean, select, text } from '@storybook/addon-knobs';
 import Dropdown from '../Dropdown';
 import DropdownSkeleton from './Dropdown.Skeleton';
 import mdx from './Dropdown.mdx';
+import Modal from '../Modal';
+import Button from '../Button';
 
 const items = [
   {
@@ -141,3 +143,34 @@ export const Skeleton = () => (
     <DropdownSkeleton />
   </div>
 );
+
+export const Overflow = () => {
+  const propsObj = props();
+  return (
+    <div style={{ width: 300, overflow: 'auto', resize: 'both' }}>
+      <Dropdown
+        {...propsObj}
+        items={items}
+        itemToString={(item) => (item ? item.text : '')}
+      />
+    </div>
+  );
+};
+export const OverflowInModal = () => {
+  const [isOpen, setIsOpen] = useState(true);
+  const propsObj = props();
+  return (
+    <div>
+      <Button onClick={() => setIsOpen(!isOpen)}>Open</Button>
+      <Modal open={isOpen} onRequestClose={() => setIsOpen(false)}>
+        <div style={{ width: 300, overflow: 'auto' }}>
+          <Dropdown
+            {...propsObj}
+            items={items}
+            itemToString={(item) => (item ? item.text : '')}
+          />
+        </div>
+      </Modal>
+    </div>
+  );
+};
--- a/packages/react/src/components/Dropdown/Dropdown-story.js
+++ b/packages/react/src/components/Dropdown/Dropdown-story.js
@@ -148,9 +148,13 @@ export const Skeleton = () => (
 export const Overflow = () => (
   <div style={{ width: 300, height: 100, overflow: 'auto', resize: 'both' }}>
     <Dropdown
-      {...{ detachMenu: boolean('Detach menu (detachMenu)', true) }}
+      id="default"
+      titleText="Dropdown label"
+      helperText="This is some helper text"
+      label="Dropdown menu options"
       items={items}
       itemToString={(item) => (item ? item.text : '')}
+      {...{ detachMenu: boolean('Detach menu (detachMenu)', true) }}
     />
   </div>
 );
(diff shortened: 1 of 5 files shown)


emyarod (commented):
thanks for jumping in on this issue, the menu placement looks good so far. currently keyboard navigation is not supported in the floating menu though, which is a regression

to answer some of your questions:


> * Should this be a default or triggered by a property?

I'm leaning towards this being triggered by a property because of the scrolling issue you mention below

> * Should this be added to multi select?

yes I believe eventually this could make its way to the other listbox components but we can focus on dropdown for the time being

> * This is required by the FloatingMenu, should these properties optional? ([Code](https://github.com/carbon-design-system/carbon/pull/7928/files#diff-b645c78914204385341704484c6145f99bbfa3765701ebc9e2c7bf1154212133R189))
> 
> ```js
>           menuRef={() => {}}
>           menuOffset={() => {}}
> ```
> 

I believe `menuRef` will be required to find the menu body and focus on the proper element, but `menuOffset` probably won't be needed

> * This is needed for the modal to work (otherwise it closes on select) - any other ideas to fix this? ([Code](https://github.com/carbon-design-system/carbon/pull/7928/files#diff-b645c78914204385341704484c6145f99bbfa3765701ebc9e2c7bf1154212133R193))
> 
> ```js
>           <ListBox.Menu
>             onMouseDown={e => e.stopPropagation()}
> ```

I don't think there's a way around this, since the floating menu is no longer a child of the modal, it views clicking inside the menu as clicking outside of the modal, which closes the modal by default. Probably another reason to go with prop vs default behavior (referring to question 1)

ConradSchmidt:
1. I added a `detachMenu` property, but I'm not happy with this name. `appendTo` would be better, but that would imply to provide an elementRef, but I didn't got this to work
2. I ported the fix to the multiselect dropdown, but it is not working (visually) 100% - I guess its easy to fix, but first I want to have an approval for the general fix, otherwise I would have to rework the whole multiselect
3. `menuRef` is just a [callback](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/internal/FloatingMenu.js#L179) - there is some logic going on which triggers the proper placing of the menu - somehow providing a function works here. maybe @asudoh can help out here?
4. okay 👍 
5. ...nothing new...

> the menu placement looks good so far. currently keyboard navigation is not supported in the floating menu though, which is a regression

6. The problem was that the detached menu wasn't focused and didn't received any keyboard inputs. I fixed it by just attaching the `onKeyDown` event on the button itself (most easy solution). I see 4 other solutions here:
6.1 forward all events to menu (you can “retrigger” events on the menu - but quite “complex”)
6.2 render an attached “hidden” menu to still trigger the events
6.3 somehow focus the detached menu (didn’t get it to work)
I also had to add a [ForwardRef](https://github.com/carbon-design-system/carbon/pull/7928/files#diff-f8a03419850a9fef647796282cfc052ad678b35c5d0000518c1b5c15a94efcf5R469) to the FloatingMenu, otherwise the `onSelect` event wasn't fired

Working keyboard navigation:
carbon-dropdown-keynav-issue

ConradSchmidt:
6. Updated the solution to use a wrapper element to get the `activeElement` state on the detached menu (solution 6.3)

Also:
1. Solved the styling issue for the MultiSelect
6. Removed the ForwardRef from FloatingMenu

ConradSchmidt:
@emyarod are there any updates on this PR? Is this solution still viable or are there any other solutions to this issue?

emyarod:
there are still some open questions from our last conversation, mainly looking for any downshift examples where a floating menu is used. I still have concerns about this use case and its compatibility with downshift

tay1orjones:
I'm going to close this due inactivity, but feel free to reopen if the use case concerns and downshift compatibility can be addressed

Issue #5367, opened 2020-02-17
Clarify default component spacings


janhassel:
### Summary

Right now, most carbon components do not have any margin around them to allow the designer and developer to use them freely in their layout. Some components though (listed below) actually **do** have margins. Sometimes all around, sometimes only on specific sides.

These margins are not represented in the Sketch Kit, causing designers to build their layouts as if there were no spacings and developer struggling to implement the designs because they have to either (1) override the margins provided by carbon (and maybe shoot themselves in the foot in the long run) or (2) reject / adjust the designs (which can have serious issues on visual hierarchy and obviously makes it hard for the design team to review the implementation).

### List of carbon components with default spacings
- [ ] Fieldset (affects groups of checkboxes and radio buttons)
  - [ ] Bottom
- [ ] File Uploader Drop Container (https://react.carbondesignsystem.com/?path=/story/fileuploader--fileuploaderdropcontainer)
  - [ ] Bottom
- [ ] FormLabel
  - [ ] Bottom
- [ ] InlineLoading
  - [ ] Top
  - [ ] Bottom
  - [ ] Left
  - The spacings are not defined as margin, but the size of the svg is set to `2rem` despite only being 16px
![image](https://user-images.githubusercontent.com/28265588/74665664-6b989380-51a0-11ea-8225-1c490d0084a0.png)
- [ ] Loading
  - [ ] Top
  - [ ] Right
  - [ ] Bottom
  - [ ] Left
  - Spacings are also not defined as margin, same behaviour as with InlineLoading
- [ ] Toast Notification
  - [ ] Top
  - [ ] Right
  - [ ] Bottom
- [ ] Inline Notification
  - [ ] Top
  - [ ] Bottom
- [ ] Structured list
  - [ ] Bottom
- [ ] Toggle
  - [ ] Top
  - [ ] Bottom
- [ ] Small Toggle
  - [ ] Top
  - [ ] Bottom

Issue #2549, opened 2018-12-21
Favor explicit usage of <FormItem> component


emyarod:
Currently many form item components contain a wrapper div within them with the `.bx--form-item` class. When composing forms, these wrapper divs are often duplicated or erroneously added, causing components to diverge from the spec. Examples of this can be seen in the current `<NumberInput>`, `<ComboBox>, `<MultiSelect>`, <FilterableMultiSelect>`, and `<Slider>` stories

vpicone:
I'm not sure forcing devs to wrap all of our form inputs with a `FormItem` wrapper is a great idea it gets really gnarley for complex forms. Using the FormItem to combine label logic though could be a good move.

As of now, the FormItem component doesn't really serve a purpose as a public since it's baked into our other form inputs. I wonder if we should just delete the story?

emyarod:
I'm unsure about the origins of exposing this component as a public and the discussion around the original PR is now missing because of the monorepo move. but to me it seemed like it was a way to enforce spacing between form elements. right now we already wrap our form inputs in `.bx--form-item` so this wrapper component doesn't have much use

IIRC the reason the idea of explicitly requiring `FormItem` when building forms came up is so that we wouldn't have to constantly override the default form spacing/margin in components (like `Slider`, `Pagination` which consume other components like `NumberInput`, `Select`)

tw15egan:
@emyarod is there any action we want to take here or do we want to close this out?

emyarod:
this can be closed since the discussion went nowhere

PR #7936, opened 2021-03-01
feat(FormGroup): add hasMargin, add story back in

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/5367

After talking with @janhassel, `FormGroup` is still used but the default margin is problematic. This adds in a new prop `hasMargin`, which is `true` by default, that can be toggled off to remove the default bottom margin. 

This also adds back in the `FormGroup` story, which showcases the new prop. 

I also realized `disabled` can be set on the `fieldset` element itself, which disabled all form elements nested inside. This is a useful feature that should be showcased, so I've also added this to the playground. 

#### Changelog

**New**

- `FormGroup` story added back in
- `hasMargin` prop to toggle `margin-bottom` on the `fieldset` element rendered by `FormGroup`
-  Styles to disabled label, helper text nested inside `FormGroup`

#### Testing / Reviewing

Check out `FormGroup` story and ensure it looks good 👍🏻


emyarod (commented):
do we have any clarity on form spacing (and/or component spacing in general) from design? I think the `hasMargin` addition to the FormGroup API is kind of a band aid solution to a larger problem highlighted in #5367

emyarod (commented):
do we have any clarity on form spacing (and/or component spacing in general) from design? I think the `hasMargin` addition to the FormGroup API is kind of a band aid solution to a larger problem highlighted in #5367

tw15egan:
Yeah, I agree it's definitely a band-aid and this needs to be addressed in our next major, but any changes to the current spacing would be considered a breaking change. 

Toggling our default spacing via a prop could also be helpful in our next major on a case-by-case basis in the context of a form. Do you have any other recommendations?

emyarod:
I don't think there's any immediate action to take until we know what the plan for FormGroup/component spacing in general is. for FormGroup specifically, if I understand correctly, consumers would be setting a `hasMargin` prop rather than adding a style rule, is that right?

tw15egan:
Yes, instead of having to set style rules on a class-by-class basis (or have a global class they could add) this would just remove the margin via a prop. 

This also adds the functionality of disabling an entire `fieldset` by passing down `disabled` to `FormGroup` instead of each individual component. 

If the main issue is adding in the margin toggle, we can remove that, but in chatting with @janhassel it seems like it would be helpful as teams are currently using FormGroup

emyarod:
my only concern is with the API change, especially if it can become obsolete as early as v11 launch pending a larger design decision (and in this case specifically workarounds are already possible via CSS as you mentioned). for this specific component, if we're not able to get design input about FormGroup, I would lean more towards just adding the `fieldset--no-margin` class over expanding the component API for the time being, curious to hear others' thoughts on this. aside from that the other changes are good to go

that being said, I'm not sure how soon we'd be reaching a decision about form spacing so maybe adding `hasMargin` to the form group API is a worthwhile quality of life enhancement, related https://github.com/carbon-design-system/carbon/issues/2549

emyarod (approved):

joshblack (approved):

PR #7937, opened 2021-03-01
docs(mdx): fix preview paths

tw15egan:
Fixes an issue with previews when on the `Docs` tab after sorting components into groups.

To test, ensure previews showup on the Docs tab


andreancardona (approved):
nicee 🎉

jnm2377:
I'm seeing this in the ComposedModal `Docs` tab. When I check our live storybook it also doesn't have a preview. Is this expected or was it supposed to show something here?
Screen Shot 2021-03-01 at 8 24 31 PM

tw15egan:
@jnm2377 good catch! Updated

dakahn (approved):

PR #7939, opened 2021-03-01
chore(license): updates license files to current year

tw15egan:
Updates license files to current year

@joshblack I fixed some [linting errors](https://github.com/carbon-design-system/carbon/compare/master...tw15egan:update-license?expand=1#diff-5244f114ca1f6ac65461d0b02e52494c45abc564db68e18587ae8b8674976a3fL2-R136) in `tasks/sync.js` to remove some unused commands, not sure what it's used for but wanted to make sure I don't break anything.

### DIFF
--- a/actions/add-review-labels/index.js
+++ b/actions/add-review-labels/index.js
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2020, 2020
+ * Copyright IBM Corp. 2016, 2021
  *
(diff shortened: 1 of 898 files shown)


tw15egan:
Actually, upon further reading, it doesn't seem like we need to update this to be the current year, "date of publication" should be sufficient. Put this PR up so we can update if needed, but going to close it for now.

PR #7960, opened 2021-03-03
fix(SideNav): pass additional props provided to SideNav to the rendered nav element

### LINKED ISSUE #7903: Sidenav should allow extra props to be passed to the rendered component
Use this template if you want to request a new feature, or a change to an
existing feature.

If you are reporting a bug or problem, please use the bug template instead.

### Summary

Sidenav should pass additional props to the rendered component, similar to what some of the other components are doing , for example Tab: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Tab/Tab.js#L165

Development effort.

### Justification

Allows for additional attributes to be passed to the component for greater control/accessibility compliance.

Any time

### Available extra resources

I can open a PR if the proposal is accepted.

gptt916:
Closes #7903

pass additional props provided to `SideNav` to the rendered `nav` element.

#### Changelog

**New**

- SideNav pass additional props to rendered element

#### Testing / Reviewing

- Create `SideNav` element
- in it's props, pass an arbitrary prop, like `data-loating-menu-container={true}`
- verify that this attribute is rendered in the `nav` element


packages/react/src/components/UIShell/SideNav.js
```diff
@@ -122,6 +123,7 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
         {...accessibilityLabel}
         {...eventHandlers}>
```
tw15egan:
```suggestion
        {...eventHandlers}
        {...other}>
```

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -122,6 +123,7 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
         {childrenToRender}
+        {...other}
```
tw15egan:
```suggestion

```

tw15egan:
@gptt916 I just pushed up a change that comments out some old code that was causing console warnings since it was now being passed down to `SideNav`.

Everything else looks good 👍🏻

tw15egan (approved):

dakahn (approved):

PR #7980, opened 2021-03-04
feat(react): add useControllableState

joshblack:
We have several components that can be controlled or uncontrolled and have built up a couple of patterns for implementing this throughout the codebase. This PR introduces a custom `useControllableState` hook to start the discussion around what it would look like to standardize this pattern across the codebase.

For controlled versus uncontrolled components, a basic example of this is an `<input>`. This primitive can be uncontrolled or controlled:

```jsx
function Uncontrolled() {
  return <input type="text" />
}

function Controlled() {
  const [value, setValue] = useState('');

  function onChange(event) {
    setValue(event.target.value);
  }

  return <input type="text" onChange={onChange} value={value} />;
}
```

One important note with this: switching from controlled to uncontrolled, or vice versa, is **not supported**.

To replicate this behavior for components we build, we would use `useControllableState`:

```jsx
function TextInput({
  onChange: controlledOnChange,
  value: controlledValue,
}) {
  // Provide the controlled value and onChange along with the default value
  const [value, setValue] = useControllableState(controlledValue, controlledOnChange, '');

  function onChange(event) {
    setValue(event.target.value);
  }

  return (
    <input
      type="text"
      onChange={onChange}
      value={value}
    />
  );
}
```

We can then use the same component, `TextInput`, for both controlled and uncontrolled scenarios:

```jsx
function Uncontrolled() {
  return <TextInput />;
}

function Controlled() {
  const [value, setValue] = useState('');
  
  function onChange(value) {
    setValue(value);
  }

  return <TextInput onChange={onChange} value={value} />;
}
```

Note: this was heavily inspired by a similar hook in [Chakra UI](https://chakra-ui.com/docs/hooks/use-controllable

#### Changelog

**New**

- Add `useControllableState` hook for components that can be controlled or uncontrolled

**Changed**


packages/react/src/internal/useControllableState.js
```diff
@@ -0,0 +1,75 @@
...
+          'This is likely caused by the value changing to a defined value ' +
+          'from undefined. Decide between using a controlled or uncontrolled ' +
+          'value for the lifetime of the component. ' +
+          'More info: https://reactjs.org/link/controlled-components'
+      );
+    }
+
+    // Controlled -> Uncontrolled
+    // If the component prop is controlled, the prop value should be defined
+    if (controlled.current == true && controlledState === undefined) {
```
joshblack:
```suggestion
    if (controlled.current === true && controlledState === undefined) {
```

dakahn (approved):

dakahn:
Thanks for walking me through this!

andreancardona (approved):

PR #8020, opened 2021-03-08
fix(feature-flags): update enabled to check for process

### LINKED ISSUE #8007: 'process is not defined' error in feature-flags package with 'enable-css-custom-properties'
Hi,
We use Carbon with true `enable-css-custom-properties` variable, and with the latest version get this error after build:
```
Uncaught ReferenceError: process is not defined
    at eval (index.js?9f8b:11)
    at Module../node_modules/@carbon/feature-flags/es/index.js
```
Package versions:
```
    "carbon-components": "10.30.0",
    "carbon-components-react": "7.30.0",
```

If I downgrade back versions (10.29.0, 7.29.0), I don't get the error.
Probably this PR: https://github.com/carbon-design-system/carbon/pull/7802 ?

joshblack:
Closes #8007

This PR updates our `feature-flags` package to work correctly in environments that don't have `process.env` available. It adds in a try/catch block that will read from `process.env` and fallback to default values if it is undefined.

#### Changelog

**New**

**Changed**

- Update build task for `feature-flags` package to safely read from `process.env`

**Removed**

#### Testing / Reviewing

- Run build on the feature flags package 
- Verify output uses a try/catch around `process` statements/expressions
- Verify that the `enabled` setting for a flag correctly reads off `process.env` if it is defined, otherwise it should fall back to the default value
- Verify that the catch block uses the default enabled value for the flag
- Verify that the feature flags read the `enabled` value with the correct key


joshblack:
bump @andreancardona @jnm2377 let me know if you have any questions!

tay1orjones (approved):
Output looks good 👍

andreancardona (approved):
👍🏽

jnm2377 (approved):

PR #8021, opened 2021-03-08
feat(react): add warning helper

### LINKED ISSUE #8017: Convert `warning` usage to `console.warn`
We currently use the `warning` module to emit warnings to developers when a code path they have encountered is deprecated. However, the `warning` module uses `console.error` under the hood, instead of `console.warn`, which has caused some issues that both our team and those in the React ecosystem has faced in terms of confusion.

Instead, it would be great if we could transition to `console.warn` usage instead of `console.error`, most likely through our own implementation of warning that we could then use in the codebase.

- [x] Make our custom warning function
- [x] Update where our code uses `warning` to use our custom `warning` function
- [x] Remove the `warning` dependency from `carbon-components-react`

andreancardona:
Closes #8017 

- Added custom warning helper function

#### Testing / Reviewing

- [ ] Make sure all tests pass for warning helper

- [ ] Make sure warning helper functions just as properly as our current warn function

### DIFF
--- a/packages/react/src/internal/__tests__/warning-test.js
+++ b/packages/react/src/internal/__tests__/warning-test.js
@@ -1,0 +1,42 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { warning } from '../warning';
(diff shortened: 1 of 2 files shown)


joshblack (approved):

tay1orjones (approved):
Love it! 🔥

PR #8048, opened 2021-03-09
fix(DataTable): return correct sort direction in sort action callback

### LINKED ISSUE #4704: Send correct value of `sortDirection` on the `handleSortBy` callback

- [x] `carbon-components` (10.5.1)
- [x] `carbon-components-react` (7.3.2)

## Detailed description

The provided `handleSortBy` callback provides the old value for `sortDirection` as an input argument to the function because it does not wait for the state to change.

```
onClick: composeEventHandlers([
  this.handleSortBy(header.key),
  onClick
    ? this.handleOnHeaderClick(onClick, {
      sortHeaderKey: header.key,
      sortDirection,
     })
    : null,
]),
```
 Over here, the `sortDirection` comes straight from the state and does not wait for the state change.

Would it be possible to move the callback call in the `setState` callback of the `handleSortBy` method so that it receives the correct value? Or call the callback with the `getNextSortState(sortDirection)` so that we get the correct value.

Right now I have to make a weird mapping knowing that when I get the `"NONE"` direction, the user actually clicked the sort and it should be done in ascending order (`"ASC"`). The mapping looks like this:
- NONE -> ASC
- ASC -> DESC
- DESC -> NONE

This is pretty weird and unintuitive. I would like to refrain from keeping a separate manager in the app for the sort logic.

DataTable, TableHeader

> see changed?

I expected to get the real value of `sortDirection` in the `handleSortBy` callback instead of the old one.

1. Use the DataTable + TableHeader (+ the other needed components) for a simple table
2. Provide the `onClick` property on the TableHeader
```
// somewhere in the component
sortConfig = {
  ...,
  onClick: (e, params) => {
    console.log(params); // on first click `params.sortDirection` will be `NONE` even though the UI displays the `ASC` state (which is the correct one)
  }
}
// in render
<IlTableHeader
  {...getHeaderProps({ header: headerObj, ...sortConfig })} // sortConfig has the `onClick` function
>
  {headerObj.header}
</IlTableHeader>
```

>
> - Style and vanilla JS:
> - React:

Example: 
https://codesandbox.io/s/priceless-waterfall-siw97
Notice the console.log output

emyarod:
Closes #4704

This PR returns the correct sort direction to the `handleSortBy` callback function in the DataTable's built in state manager

#### Testing / Reviewing

Confirm that the table sort parameters passed down to a user-provided `handleSortBy` callback accurately reflects the sort state now

### DIFF
--- a/packages/react/src/components/DataTable/DataTable.js
+++ b/packages/react/src/components/DataTable/DataTable.js
@@ -217,17 +217,18 @@ export default class DataTable extends React.Component {
       sortDirection,
       isSortable,
       isSortHeader: sortHeaderKey === header.key,
-      // Compose the event handlers so we don't overwrite a consumer's `onClick`
-      // handler
-      onClick: composeEventHandlers([
-        this.handleSortBy(header.key),
-        onClick
-          ? this.handleOnHeaderClick(onClick, {
+      onClick: (event) => {
+        const nextSortState = getNextSortState(this.props, this.state, {
+          key: header.key,
+        });
+        this.setState(nextSortState, () => {
+          onClick &&
+            this.handleOnHeaderClick(onClick, {
               sortHeaderKey: header.key,
-              sortDirection,
-            })
-          : null,
-      ]),
+              sortDirection: nextSortState.sortDirection,
+            })(event);
+        });
+      },
     };
   };
 


tw15egan (approved):
LGTM 👍 ✅

Screen Shot 2021-03-09 at 1 33 37 PM

dakahn (approved):

PR #8058, opened 2021-03-10
feat(tile): add support for disabled state for radio and selectable tiles

janhassel:
The `RadioTile` and `SelectableTile` components are used as alternatives to `RadioButton` and `Checkbox` if more description or a graphical content is needed, hence they should support a `disabled` state.

This PR adds in support for `props.disabled` on both, the radio and selectable tile.

#### Changelog

**New**

- `props.disabled` for `RadioTile` and `SelectableTIle`
- `disabled` knob in both stories

**Changed**

- Updated api snapshot to reflect the addition of this prop
- Lowered the specificty of the checkmark icon svg selector

**Removed**

- nothing

#### Testing / Reviewing

- Automated
  - Unit tests should ensure no functionality is broken
- Manual
  - The cursor should be `not-allowed` when hovering over a disabled tile
  - No hover effect should be shown
  - The tiles are excluded from the tab order when disabled
  - No event listeners should report any interaction while a tile is disabled
- Visual
  - I used `disabled-02` for the text color and `disabled-03` for the selected outline and checkmark. Please let me know what you think and if you'd prefer different colors to be used.


tw15egan:
Hey, @janhassel thanks for taking the time to add this in! I'm wondering if a `SelectableTile` or `RadioTile` could ever have a `Link` or `Tooltip` inside of them? If so, we would probably also want to somehow apply disabled states to these subelements. These are the first two that come to mind, but there may be others. 

Screen Shot 2021-03-10 at 12 43 44 PM

laurenmrice (changes requested):
For the disabled selected state, the icon and border need to be `disabled-02` so they don't look as clickable/active.

Artboard

janhassel:
@laurenmrice Updated, thanks for the guidance! 👍 

@tw15egan That's a very good question! I know some teams (including ours) use links or tooltips next to disabled items to explain _why_ they're disabled / how a user can enable them, or give a pointer to "Learn more" pages.

An example could be a pricing plan selection, where one of the plans is disabled and a tooltip inside explains that it's not possible for the user to switch to this plan because they might have resources running that would no longer be supported in a lower plan or something like that.

If a team uses interactive elements inside which are only valid if the tile is enabled, they can of course disable them manually; the question is if that's the encouraged / default way in which case we'd need to implement some logic in the Tile component. Curious what you think!

tw15egan:
@janhassel That makes a lot of sense, I don't think that needs to block this PR then because, as you said, teams will be able to also toggle `disabled` on those elements when they set the Radio / Selectable Tile to disabled. Thanks for giving me a bit more insight into some use-cases. We can always come back and add more logic to these components if needed.

tw15egan (approved):
LGTM 👍 ✅

laurenmrice (approved):
Looks good! 🎉

tay1orjones (approved):

PR #6008, opened 2020-05-04
feat(TreeView): add experimental TreeView component

emyarod:
ref #5180

This PR adds the first iteration of the experimental (unstable) Carbon TreeView component. The initial feature list for this release includes:

* Default and compact tree sizing
* Support for node icons and rendering custom content for node labels
* Node activation
* Node expansion
* Node selection (single and multiple node selection)
* Keyboard controls for both all three actions

- <kbd>Enter</kbd> will "activate" and select the currently focused node, but activation only occurs when the tree is a single selection tree, or if there is no other selected node in a multiselect tree
- <kbd>Space</kbd> will make a node selection, and in a multiselect tree it will toggle selection
- In a multiselect tree, holding <kbd>Ctrl</kbd> on Windows, <kbd>Cmd</kbd> on macOS, or <kbd>Meta</kbd> on Linux while making a selection will toggle the selection state of the currently focused node while preserving the currently selected range of nodes
- up and down arrows will navigate through the tree

> - Right arrow:
>   - When focus is on a closed node, opens the node; focus does not move.
>   - _When focus is on a open node, moves focus to the first child node._
>   - When focus is on an end node, does nothing.
> - Left arrow:
>   - When focus is on an open node, closes the node.
>   - _When focus is on a child node that is also either an end node or a closed node, moves focus to its parent node._
>   - When focus is on a root node that is also either an end node or a closed node, does nothing.

PR #8075, opened 2021-03-11
fix(treeview): update text and color tokens to match spec

### LINKED ISSUE #8000: [tree view] code bugs
Following up on the experimental tree view component. I have a couple different bug items to address:

----

- [x] **Width:** Lets set the width by default to be 256 px/16 rem. Teams can adjust as needed.

width
 
- [x] **Enabled state:** icons need to be `icon-02`

enabled

- [x] **Hover state:** the text needs to be `text-01`. The icons need to be `icon-01`

hover

- [x] **Selected state:** Blue side bar should be `interactive-04`

selected

emyarod:
Closes #8000

This PR updates the text and color tokens for the treeview hover and enabled state

#### Changelog

**New**

- story specific tree width rule

**Changed**

- hover and enabled state text and color tokens


laurenmrice (changes requested):
**Hover and Selected states:** Icons need to be `icon-01` to match the text color.
hover

laurenmrice (changes requested):
When a top level parent is selected and open, the nested items that have a caret icon are currently getting an `icon-01` color. The unselected items with caret icons should be `icon-02` unless they are selected or hovered on.
Screen Shot 2021-03-15 at 9 48 59 AM

tay1orjones (approved):

laurenmrice (changes requested):
Ok last thing, the disabled nodes just need to get a disabled cursor applied on hover. Everything else looks great !

emyarod:
I believe the cursor style for disabled items was [intentional](https://github.com/carbon-design-system/carbon/pull/6008/commits/75b57b30dadc942fb98dd18ff79a085e773780af) to match behavior in native context menus, but since we now have a context menu component and use the disabled cursor there I can make that change here as well

jnm2377 (approved):

laurenmrice (approved):
Looks good to me 🎉

PR #8080, opened 2021-03-12
fix(DataTable): expandable table size variations alignment issues

### LINKED ISSUE #4796: Data table row sizes not applied when expanded content used
Setting the row size to compact or small using bx--data-table--short or bx--data-table--compact only shrinks the head row.

- [x] `carbon-components`

I would assume all are affected

## Detailed description

Compact size produces
image

CvDataTable

> see changed?
All rows including expanded content respect the row size.

Chrome

10.7

> should be aware of?
None

1. Go here http://react.carbondesignsystem.com/?path=/story/datatable--with-expansion
2.Set row height to compact

## Additional information

- Screenshots or code
- Notes

jnm2377:
Closes #4796

#### Changelog

**New**
- adds stories for size variation expandable tables (compact, short, tall)
- updates `padding-left` for expandable child row size variations

**Removed**
- removed unneeded code

**Testing**
- Review expandable table size variations to make sure row height changes for each variation and everything aligns correctly.


tw15egan (commented):
All the header alignment looks correct (I even tested with sortable variants since that has been giving us issues 😆 ), but it looks like the top/bottom padding has been changed for the normal variant. Just wanted to make sure this was an intentional change:

**This PR**
Screen Shot 2021-03-12 at 11 25 08 AM

**Deployed Storybook**
Screen Shot 2021-03-12 at 11 25 18 AM

aagonzales (changes requested):
The left-padding of the icon and the focus border size should stay the same across row sizes so the expansion icon is always in the same alignment no matter the row size.

image

dakahn (approved):
Looks good to me once design is satisfied 👍🏽

jnm2377:
@aagonzales sorry it's taken so long for me to push these changes 😓  should be fixed now 👍🏽

aagonzales (approved):
Yeah that looks good! thanks for updating.

Issue #1937, opened 2019-02-27
AVT 3 - Carbon X Structured List does not announce checkbox or checkbox status using a screen reader


snidersd:
## Environment
macOS Mojave version 10.14.2
Chrome version 72.0.3626.96 (Official Build) (64-bit)

Windows 7
Chrome version 72.0.3626.96 (Official Build) (64-bit)
JAWS

## Description
AVT 3 - Structured List does not announce that there is a checkbox or the check box state.

1. Navigate to the first row in the list by pressing the tab key.
2. Press Ctrl-Option Right Arrow and navigate across the row.

Actual result: The checkbox is skipped and the status is not announced.
Expected result: VO should announce the checkbox and the state of checked or unchecked.

Note: Add the ARIA role="checkbox" and aria-checked="true"(or false when unchecked).

dakahn:
Great catch! Will need to do a little further research and see if this fails for VO and NVDA as well.

snidersd:
This issue should remain open until it is fixed and retested.

snidersd:
Retested and this is still an issue in Structured List with Selection. 
For React: Tab to selection and keyboard does not allow a selection or announce which one is selected. 
For Vanilla: Tab to selection and hit Enter to make a selection, the check box displays but does not hold the setting or announce that it is checked. 

The keyboard issues are present with and without of screenreader for both Vanilla and React.

joshblack:
Resolution for this is to use the `visually-hidden` class on the `<input>` instead of `display: none` as this removes it from the accessibility tree.

emyarod:
I believe this is blocked similar to #1882 esp if the markup is to be updated. substituting `display: none` with the `visually-hidden` classes introduces new DAP errors which don't seem to be false positives

joshblack:
Following up on this now, our first step was to update the markup to contain the relevant ARIA roles for tabular data so that the control could be navigable.

Marking as blocked as we are unsure how to proceed with implementing selection. In particular, how do we communicate that a row has been selected to a screen reader.

dakahn:
Tested on Windows 10 latest in Firefox and cannot make a selection with the keyboard :hurtrealbad:

PR #8084, opened 2021-03-13
feat(StructureList): refactor class components, add aria table roles

dakahn:
part of #1937 

We're working on #1937 and taking the opportunity to update StructuredList from an es6 class components to functional components w/hooks. StructuredList also seemed to to have no table markup whatsoever so i've added some basic aria roles making the component identifiable as a table to screen readers. 

#### Testing / Reviewing

This PR shouldn't have any visible changes or functionality changes whatsoever.


dakahn:
@tw15egan can't figure out why the unit test for unique IDs is failing. `getInstanceId()` seems to be working as intended -- at least I think it is 😕 

Don't have a whole lot of enzyme experience but is that failing test _looking for a `uid` prop_?

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -5,321 +5,309 @@
...
+   * Specify an optional className to be applied to the node
+   */
+  className: PropTypes.string,
+};
+
+export function StructuredListBody(props) {
+  const { children, className, ...other } = props;
+  const classes = classNames(`${prefix}--structured-list-tbody`, className);
+  const { labelRows, setLabelRows } = useState(null);
+  const { rowSelected, setRowSelected } = useState(0);
```
dakahn: this is kinda weird, probably something i'm missing, but `setLabel` and `rowSelected` were tracked in state, but unused?

tw15egan: AFAIK `StructuredList` hasn't been touched in quite some time and could definitely have some leftover unused hooks that may need to be refactored

andreancardona:
@dakahn I think we need to update tests!

packages/react/src/components/StructuredList/StructuredList.js
dakahn: Totally the case -- just got rid of em

dakahn:
`wrapper.instance()` checks against a class instance for a prop called `uid`. I updated the test to instead just check for unique id attributes being assigned correclty 👍🏽 @tw15egan

andreancardona (approved):
yay!

tw15egan (approved):
LGTM 👍 ✅

tw15egan:
@dakahn just confirming, is selection via keyboard the next step as part of #1937?

dakahn:
@tw15egan yep! This refactor is preliminary work _before_ we start hacking on 1937 👍🏽

Issue #8059, opened 2021-03-10
CodeSnippet: Remove or improve the overflow indicators


guigueb:
### Summary

The CodeSnippet component has grey shading zones on the left and right zones of scrollable content.
There needs to be a better way to indicate scrollable (see [#8057](https://github.com/carbon-design-system/carbon/issues/8057) and [#7574](https://github.com/carbon-design-system/carbon/issues/7574)).
And it needs to be applicable for both horizontal and vertical scrollable content.

### Justification

The shading zones are of little (and sometimes no) use.
They are very subtle and easily overlooked.
If there is no content in the shading zones they have no effect at all (there is nothing to shade).
On all platforms besides apple scrollbars are always visible (indicating scrollable content) and the shading zones are not needed.

Apple users are aware/used to hidden scrollbars and have the option of turning them on (https://www.matuzo.at/blog/css-pro-tip-for-mac-users-always-show-scroll-bars-in-macos/)

The overflow indicators should be removed, and replaced with a more stable/noticeable option.
One solution would be to always force CodeSnippet to show scrollbars.
This would make it consistent across platforms (apple and others) and clearly show what scrolling is available.

### Desired UX and success metrics

Testcase:
Given a CodeSnippet with data that overflows its container; it would be nice to be able to identify that scrolling to view more data is possible.

### "Must have" functionality

No matter what the data is there needs to be some way to know that scrolling is an option.

### Specific timeline issues / requests

nope

### Available extra resources

What resources do you have to assist this effort?

Maybe

sstrubberg:
Hey @guigueb, thanks for opening the request. Here are some notes from our backlog cleaning meeting today.

- If a consumer enables show scroll bars within their OS preferences, this is mostly taken care of.<img width="827" alt="image" src="https://user-images.githubusercontent.com/15822070/207113719-dc150795-28d5-4cea-b7fe-287b03e4f100.png">

- The shading zones you are referring to are more of a secondary indicator of scrollable content.

- Adjacently, @alisonjoseph made a [fix](https://github.com/carbon-design-system/carbon/pull/12605) to address the overflow which improves the usability of the scrollbars, which supplant the “secondary indicator” fade.

- Lastly, if you or a designer had a more concrete spec on how we could improve the overflow indicators, we would love to consider it.

Issue #8057, opened 2021-03-10
CodeSnippet: the scrollable content needs to be WCAG 1.4.10 compliant


guigueb:
CodeSnippet needs to deal with scrolling its data.
This issue should wait until [#7579](https://github.com/carbon-design-system/carbon/issues/7579) and [#8056](https://github.com/carbon-design-system/carbon/issues/8056) are complete as the circumstances around scrolling will become more apparent and predictable.

For reference... https://www.w3.org/WAI/WCAG21/Understanding/reflow.html

### Summary

In order for the CodeSnippet to be WCAG 1.4.10 compliant the component will need to know more information about what the data it is displaying.

About data for a horizontal language....
- typically the data will be wrapped and there will be a vertical scrollbar only
- wrapText will need to be smarter (where some whitespace is non-breaking and some non-whitespace is breaking)
- if the user sets wrapped=false...
-- we need to assume the data will overflow the horizontal axis and show the horizontal only
-- we need to assume the number of rows will show without needing a vertical scrollbar
-- we need to know if the language is RTL or LTR and justify and initially display the data accordingly

About data for a vertical language....
- typically the data will be wrapped and there will be a horizontal scrollbar only
- wrapText will need to be smarter (where some whitespace is non-breaking and some non-whitespace is breaking)
- if the user sets wrapped=false...
-- we need to assume the data will overflow the vertical axis and show the vertical only
-- we need to assume the number of rows will show without needing a horizontal scrollbar

### Justification

CodeSnippet to be WCAG 1.4.10 compliant

### Desired UX and success metrics

Testcase - horizontal language with wrapText=on (the default):
The client supplies data for a horizontal language.
Long lines will be wrapped.
A vertical scrollbar will show when data exceeds the visible (closed or expanded) area vertically.
A horizontal scrollbar will never show.

Testcase - horizontal language with wrapText=off:
The client supplies data for a horizontal language.
Long lines will NOT be wrapped.
A horizontal scrollbar will show when data exceeds the visible (closed or expanded) area horizontally. 
A vertical scrollbar will never show.

Testcase - horizontal language with LTR BiDi:
The client supplies data for a horizontal LTR language.
The data shows left justified.
The scrollbar behavior is described in other stories.

Testcase - horizontal language with RTL BiDi:
The client supplies data for a horizontal RTL language.
The data shows right justified.
The scrollbar behavior is described in other stories.

Testcase - vertical language with wrapText=on:
The client supplies data for a vertical language.
[...]

Issue #8056, opened 2021-03-10
CodeSnippet: there should be a way to set the min number of rows when in closed and expanded mode


guigueb:
I have created #[7579](https://github.com/carbon-design-system/carbon/issues/7579)  for the maximum number of rows.
Please when adding props for this keep in mind they are for the minimum number of rows.
So use props like minClosedNumberOfRows and minExpandeNumberOfRows.

### Summary

CodeSnippet: there should be a way to set the min number of rows when in closed and expanded mode

Currently in closed mode the CodeSnippet initially shows up 3 rows.
Currently in expanded mode the CodeSnippet shows as a minimum one more row than when the maximum when in closed mode.

### Justification

Clients should have control over the minimum rows shown in closed and expanded states.

### Desired UX and success metrics

When not given these are the default values to match the current values:
minClosedNumberOfRows: 3
minExpandedNumberOfRows: 16
From #[7579](https://github.com/carbon-design-system/carbon/issues/7579)
maxClosedNumberOfRows: 15
maxExpandedNumberOfRows: undefined (this is unlimited; which has unpredictable results - it should have a cap)

Testcase (props that would match current defaults):
Client wants...
- a closed mode with a minimum of 3 lines which grows with the data to 15
- an expanded mode with a minimum of 16 lines which grows with the data to forever, ever, ever....
props={minClosed: 3, maxClosed: 15, minExpanded: 16, maxExpanded: undefined} 

Testcase:
Client wants...
- a closed mode with a minimum of 1 line which grows with the data to 5
- an expanded mode with a minimum of 15 lines which grows with the data to 20 rows
props={minClosed: 1, maxClosed: 5, minExpanded: 15, maxExpanded: 20} 

Testcase:
Client wants...
- a closed mode with a minimum of 10 line which does not grow but stays at 10
- an expanded mode with a minimum of 11 lines which grows with the data to 20 rows
props={minClosed: 10, maxClosed: 10, minExpanded: 11, maxExpanded: 20} 

Testcase:
Client wants...
- a closed mode with a minimum of 10 line which does not grow but stays at 10
- an expanded mode with a minimum of 20 lines which grows with the data to 20 rows
props={minClosed: 10, maxClosed: 10, minExpanded: 20, maxExpanded: 20} 

### "Must have" functionality

Pretty much need it all

### Specific timeline issues / requests

Without the control over the layout on the screen IBM Cognos can not use CodeSnippet.
[...]

PR #8085, opened 2021-03-13
feat(CodeSnippet): set closed and expanded max number of rows

guigueb:
defaulted the height to show everything (100%)
added two new props maxClosedNumberOfRows and maxExpandedNumberOfRows
added logic using the new props to determine when to show moreLessBtn
added maxHeight style based on new props

Closes #

https://github.com/carbon-design-system/carbon/issues/7579

#### Testing / Reviewing

See [7579](https://github.com/carbon-design-system/carbon/issues/7579) for the tests to verify this PR.

Tested:

MacOS - Chrome, FireFox, and Safari
Win10 - Chrome, FireFox, and Edge

The margin fix described in [7582](https://github.com/carbon-design-system/carbon/pull/7582) was not done based on comments in [7826](https://github.com/carbon-design-system/carbon/pull/7826).

Any issues with (scrollbar, WCAG 1.4.10 compliance, minHeight, and overFlow indicators) have been discussed in [7826](https://github.com/carbon-design-system/carbon/pull/7826) will be addressed in these future PRs.
[8056](https://github.com/carbon-design-system/carbon/issues/8056) CodeSnippet: there should be a way to set the min number of rows when in closed and expanded mode
[7574](https://github.com/carbon-design-system/carbon/issues/7574) CodeSnippet: missing horizontal scrollbar when type is multi and there is a long row of data
[8057](https://github.com/carbon-design-system/carbon/issues/8057) CodeSnippet: the scrollable content needs to be WCAG 1.4.10 compliant
[8059](https://github.com/carbon-design-system/carbon/issues/8059) CodeSnippet: Remove or improve the overflow indicators


guigueb:
Where can I see the conformance list for the platforms and browsers Carbon supports?
I likely missed it reading the Carbon Design System pages.

The reason being... I viewed CodeSnippet on my Samsung Tablet and the number of lines displayed does not match the max values given.
Although, if this is a problem its beyond this PR as it wasn't displaying 15 before this PR.

emyarod (commented):
Carbon targets evergreen browsers, but that would include the mobile variants so I think the issue with the incorrect number of lines being shown on mobile browsers should be addressed together

guigueb:
@emyarod
I just retested and I am not able to reproduce the issue.
I believe the pixel count per row was off, and when I adjusted rowHeightInPixels from 18 to 16 it addressed the issue.

You mention evergreen browsers... can you point me to the conformance text on the Carbon Design System pages that mentions this?

The list of evergreen browsers (I learnt something) is huge and my ability to test them is limited (I can do mac, win, Samsung Tab S2, and a very old chromebook) so hard to tell if it works everywhere.

emyarod:
@guigueb [here](https://github.com/carbon-design-system/carbon/tree/main/config/browserslist-config-carbon) is the Carbon browserslist

packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -147,6 +162,23 @@ function CodeSnippet({
...
+  let containerStyle = {};
+  if (type === 'multi') {
+    if (expandedCode) {
+      if (maxExpandedNumberOfRows > 0) {
+        containerStyle.style = {
+          maxHeight: maxExpandedNumberOfRows * rowHeightInPixels,
+        };
+      }
+    } else {
+      if (maxClosedNumberOfRows > 0) {
```
emyarod:
```suggestion
    } else  if (maxClosedNumberOfRows > 0) {
```

this does not need to be nested right?

emyarod (commented):
also requesting ux/visual review for this feature

tw15egan:
Since this PR introduces the ability to set the max number of rows when expanded to _less_ than the total number of lines, I think we'll also need to add a selector that allows vertical scrolling, something like

```scss
  .#{$prefix}--snippet--multi.#{$prefix}--snippet--expand
    .#{$prefix}--snippet-container {
    overflow-y: auto;
  }
```

packages/react/src/components/CodeSnippet/CodeSnippet.js
guigueb: It does not, for this PR.
The code is a set up for the next PR for #8056 dealing with minHeight, it will need to be nested (or at least reworked).

guigueb:
I completely agree that overflow(scrollbars) should be taken into account for this.
That was attempted in #7826 but scrollbars are not directly related to adding the max props (although related/exacerbated, scrolling is an existing issue with large 100% height CodeSnippets - which is what I'm working on fixing) and introduces issues with WCAG 1.4.10 compliance.
As such they should be dealt with in #7574 and #8057.

tw15egan:
@guigueb in the current implementation, when the content is expanded, it shows every line of code. If this PR introduces the ability to show less than the max number, you need to be able to scroll to the bottom to view all the code, and should not be handled in a separate PR. 

Example: In the preview, set the max expanded to `16`. When you hit expand, it only expands 1 line, and the rest of the code cannot be scrolled to

guigueb:
I aware of the issue and agree, I tried to address this with my other PR - and was told to stick to the props.
I'll add changes for overflow back into this one, keep in mind it will be non-WCAG 1.4.10 compliant.

With current CodeSnippet showing 100% if there are 20000 lines you see only what you can... and the rest of the code cannot be scrolled to (same issue different number of max rows).
I suppose you can work around this by putting it in a scrollable container but that only causes other issues.
The vertical scroll issue isn't new but it is easer to see, with the limit.

tw15egan:
@guigueb it seems like we just need to add the line I suggested to fix that issue. The current implementation will push all other content down, no matter the number of lines. I'm not seeing any "unreachable" code

scroll

packages/components/src/components/code-snippet/_code-snippet.scss
```diff
@@ -202,16 +202,16 @@
...
     overflow: hidden;
     transition: max-height $duration--moderate-01 motion(standard, productive);
   }
 
   // expanded snippet container
   .#{$prefix}--snippet--multi.#{$prefix}--snippet--expand
     .#{$prefix}--snippet-container {
-    max-height: 100%;
     padding-bottom: $spacing-05;
+    overflow-y: auto;
```
tw15egan: Can we move this to L206?

guigueb: Sure can - apply it for both closed and expanded modes

tw15egan (approved):
Thanks for working on this! LGTM 👍 ✅

packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -253,6 +286,16 @@ CodeSnippet.propTypes = {
    */
   light: PropTypes.bool,
 
+  /**
+   * Specify the maximum number of rows to be shown when in closed view
+   */
+  maxClosedNumberOfRows: PropTypes.number,
```
tay1orjones: Would it make sense to use to "collapsed" instead of "closed'? I think there's precedent elsewhere in the component for the concept of it "expand"ing which would point to using "collapse"

```suggestion
  maxCollapsedNumberOfRows: PropTypes.number,
```

tay1orjones (commented):
Just one small piece of feedback on prop naming. Looks great otherwise!

packages/react/src/components/CodeSnippet/CodeSnippet.js
guigueb: The name can be changed if you want.
I went with closed, because comments in the scss used closed.
Let me know.

guigueb: changed

packages/react/src/components/CodeSnippet/CodeSnippet-story.js
```diff
@@ -43,6 +49,14 @@ const props = () => ({
   copyButtonDescription: text('Copy button title', 'Copy code snippet'),
   ariaLabel: text('ARIA label', 'Container label'),
   wrapText: boolean('Wrap text (wrapText)', true),
+  maxClosedNumberOfRows: number(
+    'maxClosedNumberOfRows: Specify the maximum number of rows to be shown when in closed view',
```
tw15egan:
```suggestion
  maxCollapsedNumberOfRows: number(
    'maxCollapsedNumberOfRows: Specify the maximum number of rows to be shown when in closed view',
```

guigueb: Crap - missed the story.
Done now.

aagonzales (approved):
Looks like it's working and looking correct to me!

tay1orjones (approved):

PR #8090, opened 2021-03-15
fix(NumberInput): set negative tabindex on number controls

### LINKED ISSUE #4185: AVT 3 NumberInput  Components - VO does not announce value or increment when the up and down arrows used to input the number
## Environment
macOS Mojave version 10.14.2
Chrome version 72.0.3626.96 (Official Build) (64-bit)

## Detailed Description
With VoiceOver enabled Tab to number input. (Focus is on the up arrow)
Press the up arrow. Voice over does not announce the value or the increment number.
Tab to the down arrow is the same.
Tab to the Number input and the value is announced.
Enter an odd number directly into the Number field.
Shift-Control-Option to enter the stepper
Control Option Up or Down Arrow keys to change the number.
The number is not announced correctly by VoiceOver.

## Expected result:
When on an arrow that increments the number up or down the user should be advised that the number is incrementing by 10 as well as the number that is displayed in the number input field.
If a user inputs a number directly that is in the wrong increments they should get a warning.
If they change the number using the VO Control-Option-Up or Down arrow keys the number should be announced correctly.

emyarod:
Closes #4185

This PR sets a negative tabindex on the increment/decrement controls on the number input so that it is not tabbable to match the vanilla number input and w3 examples. The component currently reads correctly in VO and this change should have no effect on the VO behavior

#### Testing / Reviewing

Confirm the number input controls are not tabbable and that VO still correctly reads number input interactions


dakahn (approved):
Tested on Windows 10 with JAWS and NVDA and works as intended 👍🏽

jnm2377 (approved):
Working as expected for me 👍🏽

Issue #8091, opened 2021-03-15
[List] Dropdown control has a hard-coded "Open menu" and "Close menu" tooltip and not compatible with globalization 


cheer021:
## What package(s) are you using?

- [x] `carbon-components`
- [ ] `carbon-components-react`

## Detailed description

> Describe in detail the issue you're having.

[List] List control has a hard-coded "Open menu" and "Close menu" tooltip and not compatible with globalization 

> Is this issue related to a specific component?

Dropdown control 
https://www.carbondesignsystem.com/components/dropdown/usage

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Either 
- remove the tooltips completely so consumers can pass their own string (MORE preferred) 
Or
- Read the user locale and translate the string (LESS preferred - because consumers should decide where to read the locale which can be different from browser locales) 

> What browser are you working in?

Chrome - this is a cross-browser issue

> What version of the Carbon Design System are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM Cognos Analytics: This issue is not gating our release 

## Steps to reproduce the issue

1. Open the List control link
2. Hover over the arrow svg
3. See "Open menu"

## Additional information

![Screen Shot 2021-03-15 at 12 24 05 PM](https://user-images.githubusercontent.com/2319230/111202330-5a7bab00-8589-11eb-93cd-febe635e3ded.png)

tw15egan:
The `ListBox` components should all accept a `translateWithId` prop that can be used to implement globalization. Here is a quick example:


```jsx
const listBoxMenuIconTranslationIds = {
  'close.menu': 'Translated Close text ',
  'open.menu': 'Translated Open text ',
  'clear.all': 'Clear all',
  'clear.selection': 'Clear selection',
};

<Dropdown
   ...
   translateWithId={(id) => listBoxMenuIconTranslationIds[id]}
/>
```

![Screen Shot 2021-03-15 at 11 36 01 AM](https://user-images.githubusercontent.com/11928039/111204157-17b6d480-8583-11eb-9230-7c5e00fa3d0b.png)

![Screen Shot 2021-03-15 at 11 40 37 AM](https://user-images.githubusercontent.com/11928039/111204695-a592bf80-8583-11eb-86b9-5ef203652fe9.png)

emyarod:
[...]

PR #8094, opened 2021-03-15
fix(ListBoxTrigger): support `title` on ListBoxTrigger

emyarod:
related #8091

This PR adds `title` support to the ComboBox `ListBoxTrigger` and adds Storybook knobs to Dropdown and ComboBox for using `translateWithId` (matching the multiselect story)

#### Changelog

**New**

- `title` on ListBoxTrigger
- storybook knobs for dropdown and combobox

#### Testing / Reviewing

Ensure the `translateWithId` storybook knobs work correctly for the listbox components


tw15egan (approved):
Thanks for adding that in so quick 🚀 👍🏻

packages/react/src/components/ListBox/next/ListBoxTrigger.js
```diff
@@ -36,6 +36,7 @@ const ListBoxTrigger = ({ isOpen, translateWithId: t, ...rest }) => {
     <button
       {...rest}
       aria-label={description}
+      title={description}
```
joshblack, line 39: I think for the trigger we removed `title` based on one of Carolyn's suggestions. It seems like having both would basically repeat the same information twice for a screen reader user. I think an alternative she suggested was to use only `title` if we need to offer it (remove the `aria-label in this case). This was the comment IIRC: https://github.com/carbon-design-system/carbon/issues/6938#issuecomment-706717135

emyarod: > it is ok to keep the `<title>Close menu</title>` for sighted mouse users, but please change it to `<title>Close</title>` (which is less verbose, and removes the confusing word "menu")

are you referring to this point?

on a side note, she and I briefly discussed this for #4185 / #8090 and from testing the combobox, multiselect, number input, etc it doesn't seem to be double speaking (in VO at least)

image

the current multiselect places the title under the SVG rather than the button itself, so that may be an option as well but right now it doesn't seem like either approach is causing VO to read twice

joshblack: @emyarod I'm not sure which screen reader it shows up in, just was going off of what Carolyn shared. Just was sharing that if we needed the `title` then we could just drop the `aria-label` altogether based on what she shared.

jnm2377 (approved):

andreancardona (approved):

PR #8095, opened 2021-03-15
docs(Welcome): update welcome page docs tab

tw15egan:
Updated the `Docs` tab for the Welcome page, so that it does not display a scrollable container.

Added in some helpful links, let me know if we want to add anything else

### DIFF
--- a/packages/react/.storybook/Welcome/Welcome.mdx
+++ b/packages/react/.storybook/Welcome/Welcome.mdx
@@ -1,0 +1,15 @@
+# Welcome
+
+- [Carbon Website](https://www.carbondesignsystem.com/)
+- [IBM Design Language](https://www.ibm.com/design/language/)
+
+### Other Frameworks
+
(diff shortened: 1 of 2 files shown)


joshblack (approved):

packages/react/.storybook/Welcome/Welcome.mdx
```diff
@@ -0,0 +1,15 @@
+# Welcome
+
+- [Carbon Website](https://www.carbondesignsystem.com/)
+- [IBM Design Language](https://www.ibm.com/design/language/)
+
+### Other Frameworks
```
aagonzales: Can we make "Other Frameworks" an `h2` as well so they're aren't so many type sizes on a short page.

aagonzales (approved):
🎉 Thanks for fixing that page!

tay1orjones (approved):

PR #8099, opened 2021-03-16
fix(context-menu): prevent immediate closing in Safari

janhassel:
Ref https://github.com/carbon-design-system/carbon/pull/7350#pullrequestreview-573706411

In Safari, when the context menu is triggered by ctrl+click, it closes immediately after appearing. This is caused by Safari still emitting the `click` event when the `contextmenu` event's default was prevented. The `click` event is being listened to by the internal `ClickListener` component to close the menu when it registers an outside click.

There are two ways to address this:
1. wait for a short amount of time after `mouseup`was fired before allowing the menu to close.
2. wait for `click` event to be fired before allowin the menu to close. The `click` event is only fired in Safari though, meaning that this technique would require to check which browser the user is using.

Currently, I went with technique 1 to avoid browser detection. Let me know if you think technique 2 would be better.

#### Changelog

**Changed**

- After opening context menu, added short delay before it can be closed again.

#### Testing / Reviewing

In multiple browsers, try triggering the context menu with both, right-clicking and ctrl+clicking (mac). Menu should open and not immediately close.


tay1orjones (commented):
Hi Jan! I can confirm this works as expected for me, though I'm not sure how much of a guarantee we have that the `click` event (or the setTimeout callback) will be fired in <25ms. Non performant devices or apps/environments with a strained/full event queue, for instance.

On the other hand, if we only apply logic in Safari to explicitly wait for a `click` event and safari changes/patches the behavior to no longer emit a `click` event then we'd also have a bug.

Maybe this needs both? If Safari, wait for a click event and if there isn't one in 25ms, just go ahead and `setCanBeClosed(true)`. For all other browsers it can be called immediately on `mouseup`.

I definitely could be over thinking this though, let me know what you think.

janhassel:
@tay1orjones Yeah I agree it could be error-prone. 😕

I just pushed an update combining the two:
- on `click` it immediately calls `setCanBeClosed(true)`
- on `mouseup` it calls it after a 50ms delay

Maybe browser detection would be the better way to go after all to not have those timing assumptions? Do you know if any other Carbon components currently do browser detection already or if this would be something new?

tay1orjones (approved):
@janhassel I think it would be new. The only browser-aware code I'm aware of is in styles, like [this media query targeting Safari](https://github.com/carbon-design-system/carbon/pull/8120/files).

I hadn't thought about just calling it twice! Honestly I think that's a good patch. If a need surfaces in the future for something more precise we can look into browser detection.

andreancardona:
@janhassel sorry for the delay here - it looks awesome and also works as expected! :)

andreancardona (approved):

andreancardona:

janhassel:

PR #8101, opened 2021-03-16
feat(search): add support for expandable variant

### LINKED ISSUE #6360: [Search] Add support for expandable kind
### Summary

It'd be great if the default Search component could be used in the same fashion as both the DataTableSearch and the GlobalSearch (from the Gatsby-Theme), meaning it could be in a collapsed state, showing only the magnifying glass icon. Clicking it would expand to the full input, blurring it would bring it back into collapsed state (unless there is a search term inside) – same as DataTableSearch.

### Justification

This would make it easier to provide consistent experiences without the need to "hack" around the default search component or using the DataTableSearch component out-of-context (actually didn't verify if that would work without any problems).

### "Must have" functionality

**Must have:** Same behaviour as DataTableSearch (as described above) as a prop

**Ideal:** Making all three searches consistent (Search, DataTableSearch, GlobalSearch) by bringing over the "close" button from the GlobalSearch into the other two.

janhassel:
Closes #6360

Adds an expandable variant to the search component.

#### Changelog

**New**

- Expandable variant / kind for search component

**Changed**

- Renamed previous `bx--search-magnifier` classname to `bx--search-magnifier-icon`
  - If changing classnames is considered a breaking change, I can also revert this and change the new `bx--search-magnifier` to `bx--search-magnifier-container` – just let me know!
- Updated API snapshot to reflect new prop

#### Testing / Reviewing

When `props.kind = 'expandable'`, the search component should
- be collapsed by default
- open by either
  - clicking the search icon
  - tabbing into it / receiving focus
- close when it loses focus and has no content
- stay open if it loses focus and has content


packages/components/src/components/search/_search.scss
```diff
@@ -276,6 +299,43 @@
       color: transparent;
     }
   }
+
+  .#{$prefix}--search--expandable .#{$prefix}--search-input {
+    width: 0;
+    padding: 0;
+    transition-timing-function: motion(standard, productive);
+    transition-duration: $duration--fast-01;
+    transition-property: width, padding;
```
tay1orjones, line 308: Cool, so it looks like this will inherently support animating leftwards, rightwards, or both from the center 👍

tay1orjones (changes requested):
The alignment is off in DataTable in the button/hover icon alignment, the alignment of the button when it's expanded, and the interior left padding of the input when expanded.

search alignment

packages/react/src/components/Search/Search.js
```diff
@@ -43,6 +43,11 @@ export default class Search extends Component {
      */
     id: PropTypes.string,
 
+    /**
+     * Optional prop to specify the kind of this component
+     */
+    kind: PropTypes.oneOf(['default', 'expandable']),
```
joshblack: One quick question, curious what you think between using `kind` props versus named variants of a component.

e.g. we have `<Search kind="expandable" />` or `<ExpandableSearch />`

I've definitely leaned towards the latter since the `kind` prop makes it harder to group related functionality around a variant but I'm super curious what you think @janhassel

tw15egan: Could the prop name also be something like `isExpandable`? 

Or we could mimic the `TableToolbarSearch` prop, and use `persistent` and default that to `true`? (perhaps renamed in both places to `isPersistent` in v11)

janhassel:
@tay1orjones Good catch! This was caused by the change class name change. Just pushed an update!

packages/react/src/components/Search/Search.js
janhassel: I initially thought this variant would not be big enough to make into its own component but after looking further into some other components, maybe the guidance could be to use `kind` whenever the appearance is the main differentiator and creating a new component whenever the behaviour is affected. In that case, this should be an `<ExpandableSearch />` as @joshblack mentioned.

This is based on the following:
| Component | `kind` | wrapper | Effect |
| :- | :- | :- | :- |
| Button | X | | Changes appearance: primary, secondary, ghost, danger |
| Notification | X | | Changes apperance: success, error, warning, information |
| Tile | | X | Changes behaviour: ClickableTile, SelectableTile, RadioTile |
| ContextMenu | | X | Changes behaviour: ContextMenuItem, ContextMenuSelectableItem, ContextMenu |

The (somewhat) recent change in `Toggle` / `ToggleSmall` also reflects this rule: https://github.com/carbon-design-system/carbon/pull/7380

Let me know what you think about this approach and I'll update the PR accordingly.

joshblack: @janhassel that's a great breakdown. When the prop controls _behavior_ (or even something like markup potentially) the wrapper approach works great. If it's a visual change only (like classes) then inlining in the component makes more sense.

Following that chart, does it make sense for `ExpandableSearch` to be a wrapper, or am I misunderstanding? 👀

janhassel: @joshblack Yes, I'd say this should be a wrapper component called `ExpandableSearch` following above's approach. If you agree, @tw15egan then I'll go ahead and update the PR.

tw15egan: Let's do it! 👍🏻

janhassel: @joshblack @tw15egan Just pushed an update. I tried syncing the prop types with `ExpandableSearch.propTypes = Search.propTypes` which seems to work fine for validation but the storybook docs don't pick them up. Do you guys have any idea on how to solve this?

dakahn:
@carbon-design-system/design just wanted to tap yall in for a sign off here
This PR adds a collapsible global search variant like the search seen in DataTable (where it expands from just a magnifying icon).

Specifically want to call attention to the animation behavior where the search expands on focus -- which seems different than what we do on carbondesignsystem.com (expand on focus is better 👍🏽)

laurenmrice (changes requested):
- In Safari: The word `Search` is persistent by default next to the search icon. It should only be the icon.

Screen Shot 2021-04-06 at 4 57 52 PM

janhassel:
@laurenmrice Should be fixed now!

laurenmrice (approved):
looks good 👍

packages/react/src/components/ExpandableSearch/ExpandableSearch.js
```diff
@@ -0,0 +1,79 @@
...
+  useEffect(() => {
+    if (searchRef.current?.magnifier) {
+      const { magnifier } = searchRef.current;
+      magnifier.addEventListener('click', focusInput);
+
+      return () => {
+        magnifier.removeEventListener('click', focusInput);
+      };
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
joshblack: Feel free to inline that `focusInput` helper to the `useEffect` since it seems like we're only using it there 👀  Should help with the eslint violation without much cost.

joshblack (approved):
Looking great!

joshblack:
bump @tay1orjones @tw15egan when you have a sec to re-review 👀

tay1orjones (approved):
Sorry, thought I had already put an approve on here. 

The storybook prop table is a bummer but I don't think there's a workaround since it's probably blowing up react-docgen. The easiest way I know of to fix is to just copy/paste the proptypes 😞

packages/react/src/components/ExpandableSearch/ExpandableSearch.js
janhassel: Thanks for the tip!

tw15egan (approved):

PR #8103, opened 2021-03-16
feat(CodeSnippet): add copy functionality by default

### LINKED ISSUE #4448: React Code Snippet copy button not working

- [ ] `carbon-components`

## Detailed description

> Describe in detail the issue you're having

- go to http://react.carbondesignsystem.com/?path=/story/codesnippet--single-line
- click the copy button
- paste it somewhere else

- `CodeSnippet`, but may affect other components that are using the copy button

> see changed?

- I expected it to copy the text snippet. It didn't.

- Chrome Version 77.0.3865.120 (Official Build) (64-bit)
- MacOS Catalina

- Whatever's on your demo page

> should be aware of?

### LINKED ISSUE #6995: Investigate adding copy functionality by default to CodeSnippet
Wondering if there is any downside to including something like [react copy to clipboard](https://www.npmjs.com/package/react-copy-to-clipboard) as a dependency so users do not have to add in their own copy functionality.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/6995
Closes https://github.com/carbon-design-system/carbon/issues/4448

[Comment](https://github.com/carbon-design-system/carbon/issues/4448#issuecomment-580514714) in #4448 suggest `doc.execCommand`, but it seems like this is [deprecated](https://caniuse.com/?search=execCommand) and is not suggested for use. This uses `navigator.clipboard.writeText` which is fully supported outside of IE11. 

I believe if we want to handle IE11 support, we'd need to introduce an external library (like `copy-to-clipboard`). It seems lightweight enough. https://www.npmjs.com/package/copy-to-clipboard

If we are dropping IE11 in the next major, we can always just hold off on adding the external library (and perhaps waiting to add this until the next major), but I wanted to see what everyone's thoughts were

#### Changelog

**New**

- `onClick` of copy button now calls internal `handleCopyClick` function
    - This function checks first to see if an `onClick` is provided, and if so, returns. This is to prevent duplicate copies.
    - If not, we run the copy command `navigator?.clipboard?.writeText(children);`

#### Testing / Reviewing

Ensure `inline` and `single` copy the content as expected. `playground` and `multi` should not since it provides an `onClick`

### DIFF
--- a/packages/react/src/components/CodeSnippet/CodeSnippet.js
+++ b/packages/react/src/components/CodeSnippet/CodeSnippet.js
@@ -114,6 +114,15 @@ function CodeSnippet({
     handleScroll();
   }, [handleScroll]);
 
+  const handleCopyClick = (evt) => {
+    if (onClick) {
+      onClick(evt);
+      return;
+    }
+
+    navigator?.clipboard?.writeText(children);
+  };
+
   const codeSnippetClasses = classNames(className, `${prefix}--snippet`, {
     [`${prefix}--snippet--${type}`]: type,
     [`${prefix}--snippet--disabled`]: type !== 'inline' && disabled,
@@ -137,7 +146,7 @@ function CodeSnippet({
     return (
       <Copy
         {...rest}
-        onClick={onClick}
+        onClick={handleCopyClick}
         aria-label={copyLabel || ariaLabel}
         aria-describedby={uid}
         className={codeSnippetClasses}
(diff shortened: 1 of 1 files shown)


jnm2377 (approved):
sweet!!! inline and single are copying the text! 🔥

tw15egan:
@jnm2377 it looks like we're going to go with #8105 (the functionality should remain the same). Thanks for reviewing though 😄

PR #8105, opened 2021-03-16
feat(CodeSnippet): add copy functionality with external package

### LINKED ISSUE #4448: React Code Snippet copy button not working

### LINKED ISSUE #6995: Investigate adding copy functionality by default to CodeSnippet

tw15egan:
Alternate to https://github.com/carbon-design-system/carbon/pull/8103
Closes #6995
Closes #4448

This PR adds in the same functionality as #8103, but uses the external `copy-to-clipboard` package to handle copying for better browser support. This will still prevent copying if a user has already provided an `onClick`

#### Changelog

**New**

- `copy-to-clipboard` added as a dependency 
- copy functionality is added if an `onClick` is not provided to `CodeSnippet`

#### Testing / Reviewing

Test `inline` and `single` components, the code should be copied. Test `multi` and `playground`, the code should *not* be copied


jnm2377 (approved):

guigueb:
If I read handleCopyClick  correctly it will either...
- call onClick
- or perform a copy-to-clipboard

Please consider our requirement that the copy-to-clipboard is done AND the onClick is called for additional processing.

packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -114,6 +115,15 @@ function CodeSnippet({
     handleScroll();
   }, [handleScroll]);
 
+  const handleCopyClick = (evt) => {
```
guigueb, line 118: @tw15egan
@jnm2377

If I read handleCopyClick  correctly it will either...
- call onClick
- or perform a copy-to-clipboard

Please consider our requirement that the copy-to-clipboard is done **`AND`** the onClick is called for additional processing.

My comment in #4448

I thank you two for continuing to address this.
We - IBM Cognos - have also ran into this issue and will benefit from the fix.

I would like to make a request.
In @repjarms's comment on Jan 30 the it was mentioned to override the onClick - and I agree this is needed.
But I think there's also a case for not overriding but to propagate the onClick.

Usecase (we have this situation):

client calls CodeSnippet with an onClick
client wants the copy-to-clipboard to happen and also receive an onClick event to do some other task
I'm not sure the best way to accomplish this.

Maybe an additional prop(copyToClipBoard) to indicate if the copy-to-clipboard should happen (or not happen).

 ```
const copyToClipboard = this.props.copyToClipboard ?? this.props.onClick ? false : true;

const handleCopyClick = (evt) => {
    if( copyToClipBoard) {
        copy(children);
    }
    onClick?.(evt);
};
```

guigueb (changes requested):

packages/react/src/components/CodeSnippet/CodeSnippet.js
tw15egan: `onClick` will now be propagated after the copy function is executed

guigueb: I hate to bring it up but....
What if we don't want it to copy when supplying an onClick?
The client wants to handle it themself and/or does not want to pollute/overwrite what is in the clipboard.

guigueb: Discussed with TJ.
There may be a need to deal with actions besides copy.
This is ok for now.

tay1orjones (approved):

PR #2660, opened 2019-05-09
fix(dropdown): set z-index to above the modal

cal-smith:
Closes #1064

z-indexes should be totally reordered at some point (v11?) - to prevent breakages this just bumps the z-index to `9100` for dropdown - Everything else should work as expected, and the dropdown menu should show up in modals!

PR #8107, opened 2021-03-16
fix(ListBox): change ListBox z-index higher than modal

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/2660
Refs https://github.com/carbon-design-system/carbon/pull/6365

Screen Shot 2021-03-16 at 1 12 03 PM

This stems from a conversation on Slack from @carbon-design-system/angular-maintainers and you can see the issue occurring [here](https://angular.carbondesignsystem.com/?path=/story/components-combobox--in-modal). This was previously fixed by the Angular team in #2660 but reverted in #6365 

Updates `ListBox` elements z-index values to above `Modal`, so that a Dropdown rendered _outside_ of the Modal content (to prevent Dropdown causing scrolling inside Modal) will appear above the Modal itself. This should not cause any issues when triggering a Modal when a `ListBox` element is open, since the menu's all close when the focus is removed. Let me know if I am missing a particular use-case.

#### Changelog

**Changed**

- Bumps `dropdown` value from `6000` to `9100`
- Rearranged values in descending order

#### Testing / Reviewing

`z('dropdown')` is only used in `Dropdown` and `Listbox` components

### DIFF
--- a/packages/components/src/globals/scss/_layout.scss
+++ b/packages/components/src/globals/scss/_layout.scss
@@ -17,12 +17,12 @@ $z-indexes: (
   // having a higher z-index *should* not cause issues.
   dropdown: 9100,
   modal: 9000,
-  overlay: 6000,
   header: 8000,
+  overlay: 6000,
+  floating: 6000,
   footer: 5000,
   hidden: - 1,
-  overflowHidden: - 1,
-  floating: 6000
+  overflowHidden: - 1
 );
 
 /// @access public
--- a/packages/components/src/globals/scss/_layout.scss
+++ b/packages/components/src/globals/scss/_layout.scss
@@ -12,9 +12,9 @@
 /// @type Map
 /// @group global-layout
 $z-indexes: (
+  dropdown: 9100,
   modal: 9000,
   overlay: 6000,
-  dropdown: 6000,
   header: 8000,
   footer: 5000,
   hidden: - 1,
--- a/packages/components/src/globals/scss/_layout.scss
+++ b/packages/components/src/globals/scss/_layout.scss
@@ -12,6 +12,9 @@
 /// @type Map
 /// @group global-layout
 $z-indexes: (
+  // Dropdowns that render outside of a Modal should render above a Modal.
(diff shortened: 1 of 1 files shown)


dakahn (approved):

emyarod (commented):
since this issue has swung back and forth it might be worth specifically targeting components within a modal for a z-index bump rather than re-reverting a global variable (i.e. modal z-index > dropdown z-index but dropdown z-index within a modal > modal?). it looks like this same concern was raised in #2660 as well

cal-smith:
> since this issue has swung back and forth it might be worth specifically targeting components within a modal for a z-index bump rather than re-reverting a global variable (i.e. modal z-index > dropdown z-index but dropdown z-index within a modal > modal?). it looks like this same concern was raised in #2660 as well

I'd be curious why the cleanup reverted the list-box to below the modal in the first place - It seems logical to me that floating menus and lists should occupy the top of the z-index range, with modals and other "windows" sitting somewhere below. As far as I understand menus and dropdowns should close when a modal opens (even just by virtue of it being an outside click), so that shouldn't cause a layering conflict.

This particular issue occurs because of what's required to break out of a modal - the list-box has to be rendered into the body. I'm not sure how you would specifically target the list-box when it's not a sibling of the dropdown, _or_ a child of the modal?

As mentioned in #2660, the whole z-index order probably should be reworked a bit anyway - setting it to 9100 was a minimally breaking hack.

cal-smith:
Perhaps commenting why the z-indices are what they are in the code would help prevent this from regressing unnecessarily?

tw15egan:
@cal-smith agreed, z-index's should probably be revisited as a whole, but added a comment for now. I didn't change `floating` as this can already be achieved by adding `data-floating-menu-container` to the modal. 

@emyarod due to the way the Angular team is rendering Dropdowns (child of `body` and not `modal`) this change is needed. Since menus close on loss of focus, there shouldn't be any issues with them rendering on top of, say, a modal when it opens

theiliad (approved):

emyarod (approved):

PR #8114, opened 2021-03-17
feat(react): remove warning dependency in favor of internal module

### LINKED ISSUE #8017: Convert `warning` usage to `console.warn`

joshblack:
Closes #8017

This PR finishes up #8017 and removes the `warning` dependency from `packages/react`. It updates code to use our internal `warning.js` module instead.

This PR also includes updates to test files that were mocking the `warning` module. Now, we have a custom mock at `internal/__mocks__/warning.js` and test files can use that by writing `jest.mock('../path/to/internal/warning');`

#### Changelog

**New**

**Changed**

- Update components and test files to use the new `internal/warning` module

**Removed**

- `packages/react` no longer depends on the `warning` module

#### Testing / Reviewing

- Verify that warnings for deprecated components still show up in the console
- Verify that `warning` acts as intended (when the condition is false the message is displayed)


tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #8135, opened 2021-03-19
fix(UIShell): avoid spreading `isSideNavExpanded` to non Carbon elements

### LINKED ISSUE #3970: React does not recognize the `isSideNavExpanded` prop on a DOM element

- [x] `carbon-components` - 10.6.1 (latest)
- [x] `carbon-components-react` - 7.6.1 (latest)

## Detailed description

It looks like https://github.com/carbon-design-system/carbon/pull/3626 introduced a warning about unrecognized props being applied to DOM elements.

```
Warning: React does not recognize the `isSideNavExpanded` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `issidenavexpanded` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in a (created by Context.Consumer)
    in Link (created by Context.Consumer)
    in NavLink (created by Link)
    in Link (created by ForwardRef(SideNavMenuItem))
```

here, the props (including `isSideNavExpanded`) are being spread onto `element`: https://github.com/carbon-design-system/carbon/blob/99f12f98e226b2ab1872b7379167d827e4fe7bd5/packages/react/src/components/UIShell/Link.js#L19

> see changed?

The Carbon components should not be passing through all props like this, instead they should filter out those that are not supported / relevant for the target component / element, or explicitly pass only the expected props to their children.

Having these warnings causes a lot of noise in logs, both in the browser and in tests, making it harder to spot / debug genuine problems.

This can be reproduced by running the storybook in dev mode
1. from packages/react: `npm run storybook`
2. navigate to http://localhost:9000/?path=/story/ui-shell--fixed-sidenav
3. see the warning in the browser console

image

>
> - Style and vanilla JS:
> - React:

## Additional information

- Screenshots or code
- Notes

emyarod:
Closes #3970

This PR checks the display name of Carbon SideNav children before applying UI Shell specific props to avoid unrecognized prop errors. This is likely a stop gap solution until the UI Shell is refactored

#### Testing / Reviewing

Confirm that `isSideNavExpanded` is only found on components that are expecting this prop


tw15egan (approved):
Not seeing any warnings when running locally
LGTM 👍 ✅

andreancardona (approved):
same - looks good!

PR #8159, opened 2021-03-22
fix(ClickableTile): ensure links are correct color when link has been visited

### LINKED ISSUE #8158: [ClickableTile]: All text looks like link if a real href is used.

- [x] `carbon-components`

## Detailed description

In `ClickableTile` if a real href is specified, then all of the text becomes blue and looks like a link. This might be intended but isn't by default what shows up with the empty href in the storybook so I wanted to clarify. 

This doesn't look so bad in the storybook because it's just two words but real use cases have more content in the tiles and it looks a bit odd if it all looks like a link. 

1. Go to https://carbon-components-react.netlify.app/?path=/story/components-tile--clickable 
2. Change the `href` to a real link, say https://www.google.com/
3. See that `Clickable Tile` looks like a link

Also can see this in the sandbox below. 

 https://codesandbox.io/s/intelligent-platform-ji1xi?file=/src/index.js

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8158

Overrides `:visited` styles on `ClickableTile` so that text always renders `text-01`

#### Changelog

**New**

- `Link` added to the normal Tile story to ensure `Link` styles are correctly rendered in that variant

**Changed**

- `:visited` style overrides in the `ClickableTile` component

#### Testing / Reviewing

Ensure that text renders correctly (not blue) when an `href` is provided to `ClickableTile` (added one by default to the story).

### DIFF
--- a/packages/components/src/components/tile/_tile.scss
+++ b/packages/components/src/components/tile/_tile.scss
@@ -76,7 +76,9 @@
   }
 
   .#{$prefix}--tile--clickable:hover,
-  .#{$prefix}--tile--clickable:active {
+  .#{$prefix}--tile--clickable:active,
+  .#{$prefix}--tile--clickable:visited,
+  .#{$prefix}--tile--clickable:visited:hover {
     color: $text-01;
     text-decoration: none;
   }
--- a/packages/react/src/components/Tile/Tile.js
+++ b/packages/react/src/components/Tile/Tile.js
@@ -9,6 +9,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
+import Link from '../Link';
 import {
   CheckmarkFilled16 as CheckmarkFilled,
   ChevronDown16,
@@ -168,14 +169,14 @@ export class ClickableTile extends Component {
     );
 
     return (
-      <a
+      <Link
         href={href}
         className={classes}
         {...other}
         onClick={this.handleClick}
         onKeyDown={this.handleKeyDown}>
         {children}
-      </a>
+      </Link>
     );
   }
 }
--- a/packages/react/src/components/Tile/Tile.js
+++ b/packages/react/src/components/Tile/Tile.js
@@ -9,7 +9,6 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { settings } from 'carbon-components';
-import Link from '../Link';
 import {
   CheckmarkFilled16 as CheckmarkFilled,
   ChevronDown16,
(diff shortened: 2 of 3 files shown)


dakahn (approved):

packages/react/src/components/Tile/Tile.js
```diff
@@ -169,14 +168,14 @@ export class ClickableTile extends Component {
     );
 
     return (
-      <Link
+      <a
```
emyarod: I think this would be a breaking change by removing features previously supported by the Link component

emyarod (commented):
on a separate note I brought up the cause of the underlying problem in the issue ticket but I'm not sure if anyone else confirmed it outside of #8177

packages/react/src/components/Tile/Tile.js
tw15egan: ~It seems like the only prop that causes a change is the `visited` prop, but we always want the text to be `#161616` anyways so not sure it's breaking. Is there other functionality we're providing via `Link` that is now unsupported?~

We can work around this by targeting `.#{$prefix}--tile--clickable:visited` directly to retain any current functionality 👍🏻

emyarod: another functionality that would be broken is disabled links

tw15egan: Seems like we've got an issue with that as-is. I'll create a separate issue to track that 

disabled-tile

emyarod: yeah the link itself should be disabled already so it looks like it would just be a visual issue. the root element probably does not need to be changed and the original referenced issue is just due to browser settings

tw15egan: Agreed, this will prevent the incorrect color from being rendered if the link has been visited and retain the original element

packages/components/src/components/tile/_tile.scss
```diff
@@ -76,7 +76,9 @@
   }
 
   .#{$prefix}--tile--clickable:hover,
-  .#{$prefix}--tile--clickable:active {
+  .#{$prefix}--tile--clickable:active,
+  .#{$prefix}--tile--clickable:visited,
+  .#{$prefix}--tile--clickable:visited:hover {
```
emyarod, line 81: I think these would then be user land overrides unless we have confirmation that active or visited links in tiles need to remain `text-01`?

tw15egan: I believe we always want the links inside `ClickableTile` to be `text-01` regardless, @laurenmrice can you confirm?

Sidenote, where is the browser setting to change this? I wasn't able to find this and I would imagine end users would also run into this issue

emyarod: visited link styles are controlled by the browser and whether or not the setting is exposed to the user depends on the specific browser being used

tw15egan: Yeah, I was under the assumption we don't really want to show visited styles on our components. `Link` should allow it, but this should always be `text-01` even when the link has been visited.

laurenmrice: Yes! Correct

laurenmrice (approved):
Looks correct to me 👍

emyarod (approved):
looks good to me

Issue #8151, opened 2021-03-22
Update `size` prop in components to use consistent terms for size


joshblack:
Currently size in component code are: 
![image](https://user-images.githubusercontent.com/11670886/107261025-cb6cf780-6a0c-11eb-89ea-d628da6c9693.png)

But should be:
- Small: 32px
- Medium/default: 40px
- Large: 48px

This issue is for updating the API of components that accept a `size` prop to follow this convention. If it's possible, it would be preferred to add these options in as part of the existing API and deprecate existing options before we remove them in v11.

tw15egan:
~Just to clarify, the prop names will be `small`, `medium` (the default prop value), and `large`?~

~This will make it easier to maintain backward compatibility, as some current defaults are "lg" which would incorrectly map to `large` (existing xl) styles~

I don't think will be an issue and t-shirt sizes are more concise

tw15egan:
List of components that use the size prop:

- [ ] Accordion
- [ ] Button 
- [ ] Listbox
   - [ ] Combobox
   - [ ] Dropdown
   - [ ] Multiselect
   - [ ] Filterable Multiselect
- [ ] Content Switcher
- [ ] DatePicker
- [ ] FileUploader
   - [ ] FileUploaderButton
   - [ ] FileUploaderDropContainer
   - [ ] FileUploaderItem
- [ ] Link
- [ ] Modal
   - [ ] ComposedModal
- [ ] Number Input
- [ ] Overflow Menu
- [ ] Search
- [ ] Select
- [ ] Tag
- [ ] Text Input
- [ ] Time Picker
- [ ] Toggle

Questions:
- Do we want to keep the `field` button size? 
- `Data Table` has a few sizes (`compact`, `short`, `normal`, `tall`). Do we want to change any of these? 
- Treeview has a size prop with `default` and `compact`. Do we want to change these?

joshblack:
@tw15egan great questions, I'm down for `sm`, `md`, `lg` or `small`, `medium`, `large`, whatever makes more sense. 

@aagonzales what would you recommend for TJ's questions above for `field` on Button, `DataTable`, and `TreeView`? 👀

tw15egan:
@joshblack looking into it a bit more, it seems like we should be good just sticking with the current t-shirt sizes vs. the full name. This will also give us a bit more flexibility for adding `xs` variants ( `extra-small` is a bit cumbersome)

tw15egan:
[...]

PR #8160, opened 2021-03-22
feat(React): updates to size prop

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/8151

This PR will update the `size` prop across components to use `sm`, `md`, and `lg`. 

For now, I am adding in comments anywhere `xl` is used to mark for deletion when we create our alpha branch. Until then, `lg` and `xl` should function interchangeably. I've also added `md` to the allowed values for `size` for consistency throughout.  

#### Changelog

**New**

- `lg` is now the preferred `size` value to get the "bigger" version of a component. 
- `md` is added to the `size` propType definition. Will be set as a default where needed.

**Changed**

- Changed the `size` prop knob to reflect new values

List of components that use the size prop and that will be updated:

- [x] Accordion
- [ ] Button 
- [x] Listbox
   - [x] Combobox
   - [x] Dropdown
   - [x] Multiselect
   - [x] Filterable Multiselect
- [x] Content Switcher
- [x] DatePicker
- [ ] FileUploader
   - [ ] FileUploaderButton
   - [ ] FileUploaderDropContainer
   - [ ] FileUploaderItem
- [x] Link
- [x] Modal
   - [x] ComposedModal
- [x] Number Input
- [x] Overflow Menu
- [ ] Search
- [x] Select
- [x] Tag
- [x] Text Input
- [x] Time Picker
- [x] Toggle

Questions:
- Do we want to keep the `field` button size? This also cascades into `FileUploader`, which has `small` field` and `default` size buttons
- `Data Table` has a few sizes (`compact`, `short`, `normal`, `tall`). Do we want to change any of these? 
- Treeview has a size prop with `default` and `compact`. Do we want to change these? 

Blockers:
- Search cannot be updated yet as it has `sm` `lg` and `xl` styles

#### Testing / Reviewing

Ensure the components listed above that are checked correctly render at the correct sizes
Ensure no prop warnings are emitted when toggling sizes

### DIFF
--- a/packages/components/src/components/accordion/_accordion.scss
+++ b/packages/components/src/components/accordion/_accordion.scss
@@ -73,7 +73,9 @@
   }
 
   // Size styles
-  .#{$prefix}--accordion--xl .#{$prefix}--accordion__heading {
+  // TODO V11: Remove xl selector
+  .#{$prefix}--accordion--xl .#{$prefix}--accordion__heading,
+  .#{$prefix}--accordion--lg .#{$prefix}--accordion__heading {
     min-height: rem(48px);
   }
 
--- a/packages/components/src/components/content-switcher/_content-switcher.scss
+++ b/packages/components/src/components/content-switcher/_content-switcher.scss
@@ -26,7 +26,9 @@
     height: rem(32px);
   }
 
-  .#{$prefix}--content-switcher--xl {
+  // TODO V11: Remove xl selector
+  .#{$prefix}--content-switcher--xl,
+  .#{$prefix}--content-switcher--lg {
     height: rem(48px);
   }
 
--- a/packages/components/src/components/date-picker/_date-picker.scss
+++ b/packages/components/src/components/date-picker/_date-picker.scss
@@ -117,7 +117,9 @@
   }
 
   // Size variant styles
-  .#{$prefix}--date-picker__input--xl {
+  // TODO V11: Remove xl selector
+  .#{$prefix}--date-picker__input--xl,
+  .#{$prefix}--date-picker__input--lg {
     height: rem(48px);
   }
 
--- a/packages/components/src/components/dropdown/_dropdown.scss
+++ b/packages/components/src/components/dropdown/_dropdown.scss
@@ -75,12 +75,16 @@
     }
   }
 
-  .#{$prefix}--dropdown--xl {
+  // TODO V11: Remove xl selector
+  .#{$prefix}--dropdown--xl,
+  .#{$prefix}--dropdown--lg {
     height: rem(48px);
     max-height: rem(48px);
   }
 
-  .#{$prefix}--dropdown--xl .#{$prefix}--dropdown__arrow {
+  // TODO V11: Remove xl selector
+  .#{$prefix}--dropdown--xl .#{$prefix}--dropdown__arrow,
+  .#{$prefix}--dropdown--lg .#{$prefix}--dropdown__arrow {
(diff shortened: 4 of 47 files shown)


tay1orjones (approved):
Verified components render as expected in different size variants. No prop warnings, all are 👍  except for the ones you have unchecked in your list 

> * [ ]  Button
> * [ ]  FileUploader
>   * [ ]  FileUploaderButton
>   * [ ]  FileUploaderDropContainer
>   * [ ]  FileUploaderItem
> * [ ]  Search

jnm2377 (approved):
Sorry I'm just now getting to this. Tested the size variations and all are working correctly. WRT some of your questions:

> Do we want to keep the field button size? This also cascades into FileUploader, which has small fieldanddefault` size buttons

For the sake of consistency,  I vote for updating these to be `sm, md, lg` as well. The FileUploader sizing (field, small, default) confuses me as is.


> Data Table has a few sizes (compact, short, normal, tall). Do we want to change any of these?

I would be fine with updating these. I think it would still make sense with `xs, s, m, lg` with `md` being default. Tbh I think this would be more intuitive. I always have a hard time remembering whether `compact` or `short` is the smallest. Modal has `xs` size, so it wouldn't be the first component to have 4 sizes. And would be consistent with our other sizing.

> Treeview has a size prop with default and compact. Do we want to change these?

Same as above.

I can work on those sizing updates in another PR, if ya want. :)

tw15egan:
@jnm2377 great points! Yeah, maybe we should move these outstanding questions to a new issue, and address them in a separate PR as you said. These ones were the easy ones and mapped directly to the new values 🎉

aagonzales:
- Ok below is basically what you have with name to size collation. So all `sm` things are always 32px, it shouldn't vary based on component. 
-  Default shouldn't ever be the size name. Default is going to vary based on the component, it won't always be at the `md` level. Default will just be in guidance now and not as part of the name. 
- I'm fine with changing data table rows and file uploader size to be consistent with this naming. Row sizes just came before we had any other size variant which is why its the outlier now. 

| Name | Size |
|-----|-------|
| `xs` | 24 |
| `sm` | 32 |
| `md` | 40 |
| `lg` | 48 |
| `xl` | 64 |
| `2xl` | 80 |

tw15egan:
@aagonzales that is super helpful! I'm going to create a new issue with that table to track those outstanding components that map a little differently than these basic form ones.

aagonzales:
@tw15egan the one thing I'm just noticing though like data table doesn't have a `md`/40px row size so would we just skip that size then? I don't really like the idea of mixing values across a single size name. ie sometimes `md` is 40 somethings its 48. 

| Name | Size |
|-----|-------|
| `xs` | 24 |
| `sm` | 32 |
| `md` | n/a |
| `lg` | 48 |
| `xl` | 64 |

tw15egan:
I think if we apply it to DataTable, I would skip the values we don't have (or is there a use-case for adding in that missing value?). I agree, we shouldn't mix what the value of `md` is across components. A user should know `md === 40px`

aagonzales:
I think most components we would probably just add in the skipped value. I'm not so sure about data table though, just skip the `md` value for now I guess.

tw15egan:
@aagonzales I created https://github.com/carbon-design-system/carbon/issues/8370 to properly track the work, and to keep a running list of the components that have / have not been updated

aagonzales (approved):

PR #8162, opened 2021-03-23
fix(datepicker): fixed the datepicker component bidi layout issue. (#8161)

### LINKED ISSUE #8161: Bidi support: The next & previous month arrow icons are inverted in Carbon DatePicker component with RTLCSS
- [x] `carbon-components`
- [x] `@carbon/icons-react`

## Detailed description

While using the RTLCSS to address bidi layout issues, found that the next & previous month arrow icons in the date picker are inverted, those arrow should pointing outward instead of pointing inward.

DatePicker

> see changed?
Expect the next & previous month arrow icons are pointing outward that same as the LTR layout, but the next & previous month arrow icons are pointing inward with RTLCSS, and would like to see the next & previous month arrow icons pointing outward while the CSS was processed by RTLCSS, next month arrow on the left should point to the left, and the previous month arrow on the right should point to the right.

Firefox 86 & Chrome 89.0.4389.82

10.31.0

1. Create a DatePicker component
2. Process the carbon-components.min.css with RTLCSS

Codesandbox in React:
https://codesandbox.io/s/webpack-react-rtlcss-carbon-datepicker-g32t6

## Additional information
- DatePicker with RTLCSS:
圖片

- Expected DatePicker:
圖片

DavidSCChen:
Closes #8161 

Fixed the next & previous month arrow orientation for datepicker component in bidi(RTL) layout by adding the RTLCSS value directives.

#### Changelog

**New**
Added RTLCSS value directives in to the _flatpickr.scss.

#### Testing / Reviewing

Verified the fix has no impact to the ltr layout by using the local storybook development server.
Also created the sandbox to verify the rtl layout:
https://codesandbox.io/s/webpack-react-rtlcss-carbon-datepicker-fixed-hpe4s?file=/app/carbon-components.scss:239237-239282


packages/components/src/components/date-picker/_flatpickr.scss
```diff
@@ -142,6 +142,7 @@
   width: rem(40px);
   padding: 0;
   fill: $icon-01;
+  transform: scale(1, 1) /* rtl: scale(-1,1) */;
```
tw15egan: Due to a limitation with the way sass compiles RTL directives, we need to change the format to something like this so it is correctly applied 

```suggestion
  transform: scale(1, 1) #{'/*rtl: scale(-1,1)*/'};
```

This way, sass correctly interpolates the comment as a directive 

See usage in Checkbox: https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/components/checkbox/_checkbox.scss#L105

DavidSCChen: Hi @tw15egan ,
Thanks for your review, I had updated the code as suggested.

tw15egan (approved):

tay1orjones (approved):

PR #753, opened 2018-05-04
fix(structured-list): Update header alignment to bottom to match designs

alisonjoseph:
## Overview

Resolves #721

Updates header alignment so when text wraps it will align to the bottom.

Issue #721, opened 2018-04-23
StructuredListWrapper doesn't respond to the size of the screen very well


joshblack:
@srlim commented on [Sun Apr 15 2018](https://github.com/carbon-design-system/carbon-components-react/issues/812)

## Detailed description

> Describe in detail the issue you're having.
When the screen size is small, structured list doesn't respond the size very well.
![image](https://user-images.githubusercontent.com/5316797/38788225-903ff460-4100-11e8-8f87-a06fcf497d53.png)

> Is this a feature request (new component, new icon), a bug, or a general issue? bug

> Is this issue related to a specific component? StructuredListWrapper

> What did you expect to happen? What happened instead? What would you like to see changed?
It should be able to contain its contents in the wrapper.

> What browser are you working in? Chrome, Firefox

> What version of the Carbon Design System are you using?
carbon-components: 8.16.8
carbon-components-react: 5.29.5

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
IBM Cloud Private

## Steps to reproduce the issue

https://codesandbox.io/s/0qjlxnmz30

If you reduce the width of the browser size to smaller than 377, module wrapper content isn't contained within the window

---

@joshblack commented on [Mon Apr 23 2018](https://github.com/carbon-design-system/carbon-components-react/issues/812#issuecomment-383686390)

cc @carbon-design-system/design what's the expected visual design at this width? Noticed we could re-create in our codepen example as well if we change the column names from `ColumnA` to `Column A`. Would assume what's shown in the screenshot is valid, as the word will wrap but the text remains centered, but wanted to double check 👍 

---

@aagonzales commented on [Mon Apr 23 2018](https://github.com/carbon-design-system/carbon-components-react/issues/812#issuecomment-383696866)

Headers can wrap up to two lines and should be aligned along the base line (not vertically centered that's being shown. That's a bug on our side)
<img width="574" alt="image" src="https://user-images.githubusercontent.com/11670886/39149420-9bd5bd78-4704-11e8-9a4b-ca06eac9c50d.png">

Structured Lists should trigger a horizontal scroll on smaller screens. Columns should decrease to their smallest size (chosen by each individual use-case made my product side designers).
<img width="594" alt="image" src="https://user-images.githubusercontent.com/11670886/39149219-090448e8-4704-11e8-8578-2e79eb8c225c.png">

---

[...]

PR #8166, opened 2021-03-23
fix(structured-list): top align header text

### LINKED ISSUE #8157: Support _row_ headings for `StructuredList`
### Summary

I would like to have row headings for a `StructuredList`.

### Justification

I seek formatting akin to a dictionary, where there is a meaningful term on the left, and an elaboration for the term on the right:

unneeded | column headings
-- | --
**Purpose** | elaborate on purpose here
**Result** | elaborate on result... this is a long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long, long
**Associations** | list of associations...

Note that GitHub Markdown displays the row heading centered, above.

When I try to use the `head` attribute for the first `StructuredListCell` of a `StructuredListRow` the text formatting is applied, but the alignment of the field drops to the bottom, which looks awkward:
image

I seek a visibly distinct row heading which is even with the top line of the other cells in the row, or centered, or, ideally, a simple way to select which behavior is desired.

### "Must have" functionality

Row headings, bold by default, centered or top aligned.

Nothing specific. 2021?

### Available extra resources

n/a

emyarod:
Closes #8157
related #721
related #753

This PR top aligns headers so that they stay aligned with non-header text in structured lists

#### Testing / Reviewing

Apply the header class to a list row cell and check its vertical alignment


tay1orjones (approved):
Looks good to me 👍

tw15egan:
Should we check if the header is part of `tbody`, and if so change the line-height to match the rest of the row? It's currently getting the header line-height of `1.29` but the rest of the row has a `1.43` line-height 

Screen Shot 2021-03-23 at 4 05 38 PM

emyarod:
the line heights are set by our type mixins so we will need to see if they need to be updated

laurenmrice (approved):
Looks good 🚀

tw15egan (approved):
LGTM 👍 ✅

PR #7611, opened 2021-01-20
fix(CopyButton): ensure copy text is placed over icons

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7324
Closes https://github.com/carbon-design-system/carbon/issues/7605

Ensures Copy button feedback text is placed over other Copy button icons. Also fixes an issue with multiple CodeSnippets being placed on top of each other.

#### Changelog

**New**

- `z-index: 3` added to `.#{$prefix}--copy-btn__feedback`

#### Testing / Reviewing

Go to https://codesandbox.io/s/wonderful-cori-nnk8g?file=/index.js and find the `.bx--copy-btn .bx--copy-btn__feedback` in the inspector. Then add `z-index: 3;` and ensure the feedback tex is placed over subsequent `CopyButton` instances

PR #8171, opened 2021-03-23
feat(CodeSnippet): add feedbackTimeout prop, z-index fix

### LINKED ISSUE #8167: multi and single CodeSnippets do not support the feedbackTimeout prop
A user is able to specify the length of the timeout for an `inline` `CodeSnippet`, but not the `single` or `multi` variants.

### LINKED ISSUE #8168: Investigate z-index issues with CodeSnippet feedback text
Noticed in slack, can be reproduced here: 

https://codesandbox.io/s/restless-surf-611lw?file=/index.js

CodeSnippet feedback text should use the same z-index values as other tooltips

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8167
Closes https://github.com/carbon-design-system/carbon/issues/8168

This adds in the `feedbackTimeout` prop to all variants of `codeSnippet`. 

Additionally, it sets the `z-index` value of the feedback tooltip to the same value as other tooltips.

#### Changelog

**New**

- `feedbackTimeout` prop is now available to all `CodeSnippet` variants. Knob added to storybook.

**Changed**

- added `z('floating')` for icon tooltips, making them consistent with other tooltips. 

#### Testing / Reviewing

Look at the demo story in `CodeSnippet` and ensure no tooltips are rendered behind a code snippet / copy button. Will remove story before merging 👍🏻


tay1orjones (approved):
Looks good to me. Thanks for adding that demo story for review!

andreancardona (approved):
looks awesome, thank you :)

packages/components/src/components/copy-button/_copy-button.scss
```diff
@@ -107,7 +107,6 @@
     .#{$prefix}--copy-btn__feedback {
       @include tooltip--content('icon');
 
-      z-index: 3;
```
emyarod, line 110: can the rest of #7611 (z-index change in `_copy-button.scss`) be reverted as well or does it only apply to this line?

tw15egan: I'll test it out and revert it if it can be 👍🏻

tw15egan: It seems like we need to keep the `z-index: 10` specific to the multi-line variant. The reason being, if we move it back, the `bx--copy-button` for ALL snippets has a z-index of 10. The assistive text tooltip is a child of the copy button, so it inherits the `z-index: 10` from its parent. Even with the `z-index: 6000` coming from the tooltip z-index, it is still in the global z-index scale of 10, with a sub-index of 6000 inside of the parent context. 
Screen Shot 2021-03-24 at 10 59 02 AM

Hope that makes sense, let me know if you know of any workarounds. My usual way of breaking from the z-index parent stack is to use `transform3d` but for some reason, that doesn't seem to work here.

tw15egan:
recheck

PR #8172, opened 2021-03-23
feat(StructuredList): Accessibility refactor

### LINKED ISSUE #1937: AVT 3 - Carbon X Structured List does not announce checkbox or checkbox status using a screen reader

dakahn:
Closes #1937 

This PR is a big overhaul/refactor of StructuredList to provide an accessible experience for keyboard and screen reader users changes include
- new keyboard interaction pattern moving away from tab based selection and using a radio button style selection
- no longer using `<label>` as row elements to handle clicks -- no using React.context
- changed component role from table to grid to more closely follow WAI-Aria spec

### Deprecations
This PR introduces the following deprecations to existing StructuredList API (breaking changes should be behind a feature-flag:
- `label` prop which allowed table rows to select their nested radio buttons via event propagation
- `defaultChecked` maybe unnecessary, but React was throwing a warning about controlled/uncontrolled components
- `onChange` our context setup utilizes the onChange prop on the inputs now
- `value` is now set according to context

### Testing
Ideally, functionality, styling, and mouse based interaction should be identical with keyboard interaction and screen reader accessibility improving greatly 🏄🏽


dakahn:
@andreancardona

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -127,32 +126,38 @@ StructuredListBody.defaultProps = {
   onKeyDown: () => {},
 };
 
+function useFocusWithin() {
```
joshblack: If we want to make this into its own hook we should totally implement this / move over to `internal`. Otherwise if we just want it in this component for now we could totally just remove it 👍

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -127,32 +126,38 @@ StructuredListBody.defaultProps = {
...
-    onKeyDown,
-    tabIndex,
-    children,
-    className,
-    head,
-    label,
-    ...other
-  } = props;
+  const { onKeyDown, children, className, head, ...other } = props;
+  const [hasFocusWithin, setHasFocusWithin] = useFocusWithin();
```
joshblack:
```suggestion
const [hasFocusWithin, setHasFocusWithin] = useState(false);
```

Based on note above, if we want to drop the custom hook we would just update this to

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -127,32 +126,38 @@ StructuredListBody.defaultProps = {
...
   const classes = classNames(`${prefix}--structured-list-row`, className, {
     [`${prefix}--structured-list-row--header-row`]: head,
+    [`${prefix}--structured-list-row--focused-within`]: hasFocusWithin,
   });
 
-  return label ? (
+  return head ? (
+    <div tabIndex={-1} role="row" {...other} className={classes}>
```
joshblack: Does this need a `tabIndex="-1"`?

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -173,41 +178,39 @@ StructuredListRow.propTypes = {
    */
   head: PropTypes.bool,
 
-  /**
-   * Specify whether a `<label>` should be used
-   */
-  label: PropTypes.bool,
```
joshblack: Removing props is totally fine I think, just need to use the `deprecate` prop helper first so that people who use this prop now won't have errors about these props being passed on to DOM nodes through `other`

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -173,41 +178,39 @@ StructuredListRow.propTypes = {
...
 
 export function StructuredListInput(props) {
-  const { className, value, name, title, id, ...other } = props;
-  const classes = classNames(`${prefix}--structured-list-input`, className);
-  const instanceId = id || getInstanceId();
+  const defaultId = useId('structureListInput');
+  const {
+    className,
+    value,
+    name = `structuredListInput-${defaultId}`,
```
joshblack: Style nit

```suggestion
    name = `structured-list-input-${defaultId}`,
```

joshblack (changes requested):

packages/react/src/components/StructuredList/StructuredList.js
dakahn: I think the custom hook is helpful! I'll keep it and moved it to /internal

packages/react/src/components/StructuredList/StructuredList.js
dakahn: false positive from Accessibility Checker/jsx-a11y. I put it in to test fixes and forgot to take it out 🤦🏽

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -231,7 +252,10 @@ StructuredListInput.propTypes = {
   /**
    * Specify whether the underlying input should be checked by default
    */
-  defaultChecked: PropTypes.bool,
+  defaultChecked: deprecate(
+    PropTypes.bool,
+    `\nThe prop \`defaultChecked\` will be removed in the next major version of Carbon.`
```
joshblack: Maybe something like:

> The prop `defaultChecked` is no longer needed and will be removed

Just so people know this is no longer needed

packages/react/src/components/StructuredList/StructuredList.js
joshblack: I think if it's a wrapper around `useState` then we might want to just embed it directly into the component (it doesn't add much value as just a wrapper).

An example of a custom hook may be something like this: https://codesandbox.io/s/focuswithin-wtzp1 that could be generalized

dakahn:

tw15egan:
Noticed a small little visual bug at the top right of the component

Screen Shot 2021-03-30 at 10 50 11 AM

andreancardona:
@tw15egan just fixed that styling issue!

andreancardona:

tw15egan (approved):
Looks great!  👍 ✅

laurenmrice:
In this example: I am showing at first how nothing is selected or focused by default. Then I `tab` into the first row and use the `down` arrow keys and each row gets focus. After I select a row, and use the `up` and `down` arrow keys again, it then changes the selected row while focusing. Is that the correct behavior ?
Apr-01-2021 14-16-19

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -176,40 +190,52 @@ StructuredListRow.propTypes = {
   /**
    * Specify whether a `<label>` should be used
    */
-  label: PropTypes.bool,
+  label: deprecate(
+    PropTypes.bool,
+    `\nThe prop \`label\` will be removed in the next major version of Carbon.`
```
joshblack:
```suggestion
    `\nThe \`label\` prop is no longer needed and will be removed in the next major version of Carbon.`
```

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -267,8 +300,12 @@ export function StructuredListCell(props) {
     [`${prefix}--structured-list-content--nowrap`]: noWrap,
   });
 
-  return (
-    <div className={classes} role={head ? 'columnheader' : 'cell'} {...other}>
+  return head ? (
```
joshblack: Just a style note but feel free to use:

```jsx
if (head) {
   return <span>...</span>
}

return <div>...</div>;
```

If you think it reads better than the ternary 👍

packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -176,40 +190,52 @@ StructuredListRow.propTypes = {
...
 
   return (
     <input
       {...other}
       type="radio"
-      tabIndex={-1}
-      id={instanceId}
+      tabIndex={0}
+      checked={row && row.id === selectedRow}
+      value={row ? row.id : ''}
```
joshblack: Something I'm a little unsure about here is if they still pass in a `value` what should we do with it?

One tricky thing with adding in context is that if, for whatever reason, someone uses `StructuredListInput` without the wrapper components then this component ends up not working because the context is never defined.

I guess this kind of goes back to your point earlier about how this might be too big of a breaking change to safely migrate to in v10? Curious what your thoughts are

joshblack (changes requested):

dakahn:
@laurenmrice Yep! That's the expected interaction. We've basically got a big fancy set of radio-button's keyboard/screen reader wise now 👍🏽

laurenmrice (approved):
lgtm 👍 !

packages/react/src/components/StructuredList/StructuredList.js
dakahn: its worth talking about for sure -- I guess what are the chances or what is the actual use case of _only using_ StructuredListInput? Should we put a call out on Slack or maybe kick this whole can down the road to v11?

joshblack: @dakahn maybe the safest thing is just to wrap it all behind a flag, I guess? If we want to do it in v10, then I think we'll just need to make sure we understand what props teams could be supplying and how to leverage/use them, if so.

For example, if someone is using a `label` then they might be expecting in a test to be able to find a node using that `label`. If we no longer use it in the component, then their test would be break.

packages/react/src/components/StructuredList/index.js
```diff
@@ -4,6 +4,70 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
+import React from 'react';
+import {
+  StructuredListWrapper as StructuredListWrapperNew,
```
joshblack: One small style nit: prefer `Next` over `New` and `Classic` over `Old`

joshblack (changes requested):

joshblack:
Looking great! Just has that small style nit for naming conventions and then snapshots that are failing in CircleCI 👀

joshblack (approved):

PR #8176, opened 2021-03-24
Fix modal overflow on  IE11

### LINKED ISSUE #6919: Modals with lots of content don't size correctly on IE

- [x] `carbon-components`

## Detailed description

On Chrome etc., if the viewport is small enough and/or the modal's content is large enough, the modal will get a scrollbar.

This does not happen on IE11.  Instead, the modal overflows the screen.

See https://the-carbon-components.netlify.app/component/modal--default.html or https://react.carbondesignsystem.com/?path=/story/modal--default for simple test cases.   In the first link, you need to make your browser window quite small, or add content to the modal, in order to see the problem.  In the second link, check the "hasScrollingContent" knob.

## Additional information

Here's the current behavior:

current

Is there some reason the modal is using grid sizing?   It seems like it should just be a flexbox (with the buttons on the bottom in a nested flexbox).  Note that although the `bx--modal-container` is `display:grid` on chrome, it's `display:block` on IE.   

If I switch it to `display:flex` and put a height on the node, IE basically works:

flex

carlosparreno:
Closes #6919

- Modal content scrolls is not displayed. Instead, content overflows the header and footer containers.
- `display: grid`, `grid-template-columns` and `grid-template-rows` are not supported in IE11. Instead we need to use their correspondent IE11 values: `-ms-grid`, `-ms-grid-row` and `-ms-grid-column`
- Solution based on this article https://elad.medium.com/supporting-css-grid-in-internet-explorer-b38669e75d66

#### Changelog

**Changed**

- _modal.scss file

#### Testing / Reviewing

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -773,6 +773,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "carlosparreno",
+      "name": "Carlos Parreño",
+      "avatar_url": "https://avatars.githubusercontent.com/u/25571131?v=4",
+      "profile": "https://github.com/carlosparreno",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ check out our [Contributing Guide](/.github/CONTRIBUTING.md) and our
   </tr>
   <tr>
     <td align="center"><a href="https://github.com/szymonbrandys"><img src="https://avatars.githubusercontent.com/u/79149899?v=4?s=100" width="100px;" alt=""/><br /><sub><b>szymonbrandys</b></sub></a><br /><a href="https://github.com/carbon-design-system/carbon/commits?author=szymonbrandys" title="Code">💻</a></td>
+    <td align="center"><a href="https://github.com/carlosparreno"><img src="https://avatars.githubusercontent.com/u/25571131?v=4?s=100" width="100px;" alt=""/><br /><sub><b>Carlos Parreño</b></sub></a><br /><a href="https://github.com/carbon-design-system/carbon/commits?author=carlosparreno" title="Code">💻</a></td>
   </tr>
 </table>
 
--- a/packages/components/src/components/modal/_modal.scss
+++ b/packages/components/src/components/modal/_modal.scss
@@ -70,8 +70,12 @@
     position: fixed;
     top: 0;
     display: grid;
+    // stylelint-disable-next-line declaration-block-no-duplicate-properties
+    display: -ms-grid;
     grid-template-rows: auto 1fr auto;
+    -ms-grid-rows: auto 1fr auto;
     grid-template-columns: auto;
+    -ms-grid-columns: auto;
     width: 100%;
     height: 100%;
     max-height: 100%;
@@ -225,7 +229,9 @@
   }
 
   .#{$prefix}--modal-header {
+    -ms-grid-row: 1;
     grid-row: 1/1;
+    -ms-grid-column: 1;
     grid-column: 1/-1;
     margin-bottom: $spacing-03;
     padding-top: $spacing-05;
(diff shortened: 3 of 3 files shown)


carlosparreno:
Hi @tw15egan,

I'm trying to use the Netlify deployed version https://deploy-preview-8176--carbon-components-react.netlify.app/ on IE11 unsuccessfully. Is there any known issue with this? It seems to happen in all recent Carbon versions (not in the old ones, e.g. I tried one in July)

Is this a known issue? If so, can you please advise on how to test this?

Thanks in advance

tw15egan:
@carlosparreno I will pull it down locally and verify. I believe some of the theme storybook additions (which are only in the prod deploy) may not be IE11 compatible

carlosparreno:
Great!! Thanks, @tw15egan !

tw15egan:
@carlosparreno upon taking a look, I'm seeing this is adding IE-specific media queries. In your application, when you compile the styles, these IE media queries should be added automatically when using something like [autoprefixer](https://github.com/postcss/autoprefixer).

Can you confirm that you are using an `autoprefixer` as part of your build process, and still seeing this issue?

tw15egan:
What is weird, though, is that `Modal` does not seem to be picking up grid-specific media queries for IE...

modal

You can see on the left that it _is_ getting them for flex though. @tay1orjones any ideas?

carlosparreno:
Hi @tw15egan, thanks for your reply:

We are not using autoprefixer. As per my understanding, all the css in carbon is ready to consume as you import `carbon-components`. Is there any pre-req I'm missing?

As you say, the Modal does not seem to be picking up grid-specific media queries for IE. This is reproducible here if you extend the content text until it overflows the footer section: https://the-carbon-components.netlify.app/?nav=modal

Could you please advise on how to compile this in my application or otherwise where to look at it in Carbon and fix it myself?

Thanks TJ!

carlosparreno:
Hi @tw15egan,

Just wondering if you had a chance to look at my comment above ⬆️ .

Thanks a lot in advance 🤗

tw15egan:
@carlosparreno I just added https://github.com/carbon-design-system/carbon/pull/8215 that updates our browser matrix to include the ie11 grid options for autoprefixer to our compiled css.

In terms of enabling autoprefixer as part of your build process, you can take a look at their docs here: https://github.com/postcss/autoprefixer

The PR I mentioned updates our `gulpfile` use of autoprefixer, but you can also use it in a webpack setup.

tay1orjones:
@carlosparreno Could you confirm #8215 @tw15egan created fixes the issue for you? It has been released in [v10.32.0](https://github.com/carbon-design-system/carbon/releases/tag/v10.32.0).

I think we can close this one in favor of that fix.

carlosparreno:
Hi @tay1orjones ,

I did a quick test in the vanilla version https://the-carbon-components.netlify.app/?nav=modal and I still see the issue. See image:

image

The modal buttons are hidden when the content is too large and the CSS IE11 specific is not present.

Maybe the link above isn't still updated with the latest Carbon version ?

Also, @tw15egan mentioned to me that I needed to use the autoprefixer in my app. I'm a little bit confused by that. As per my understanding, the components are working in IE11 out of the box. Also, I didn't see any step in the install documentation regarding autoprefixer either https://www.carbondesignsystem.com/developing/react-tutorial/step-1/.  Therefore, my guess is that autoprefixer adds the CSS browser-specific in Carbon as part of the build process and the output is shipped supporting IE11. Am I understanding this right?

Thanks a lot in advance.

tay1orjones:
> As per my understanding, the components are working in IE11 out of the box.

@carlosparreno Yes, if you import and use [the compiled css](https://unpkg.com/carbon-components@10.32.0/css/carbon-components.css). Search/find there for `.bx--modal-container` and you'll see the `display: -ms-grid;`. If you compile/build from `.scss` files, you'll need to include autoprefixer in your build pipeline. 

[This codesandbox](https://codesandbox.io/s/bold-taussig-6q2ei?file=/src/index.js) uses the css directly and so should show it with the prefixes in IE11.

carlosparreno:
Cool, I get it. And it's working now as you could also show in your `codesandbox` working and I also get what you guys meant with the `autoprefixer`. 

All good now, I'm closing this.

Thank you all!

carlosparreno:

tw15egan:
@carlosparreno Awesome! Glad we could help you get that sorted out, and thanks for pointing out that issue 👍🏻

Issue #7115, opened 2020-10-22
Visited link style(s) no longer appear


metonym:
When applying the `visited` prop to the `Link` component, the link remains blue.

The expected behavior is that the link should turn purple.

- [x] `carbon-components`
- [x] `carbon-components-react`

In the [React storybook](https://react.carbondesignsystem.com/?path=/story/link--default), toggle the knob "Allow visited styles" in the Link story.

When inspecting the CSS, the "bx--link--visited" class is applied as expected, but the purple color nows sits behind a `:visited` pseudoselector. Is this a regression or a progression?

<img width="149" alt="Screen Shot 2020-10-22 at 9 07 01 AM" src="https://user-images.githubusercontent.com/10718366/96899904-5322fc00-1446-11eb-8ea4-dc19390a057a.png">

emyarod:
may be related to https://github.com/carbon-design-system/carbon/issues/3581 https://github.com/carbon-design-system/carbon/pull/5239

metonym:
@emyarod Thanks for the clarification

PR #5239, opened 2020-01-31
fix(link): remove visited styles by default

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/3581

Removes `:visited` styles by default. If the link has a `bx--link--visited` class, it will receive visited styles once a link has been visited.

Not sure if `bx--link--visited` is the greatest name, since it will have that class _before_ it is actually visited, so open to suggestions. 

#### Changelog

**New**

- `bx--link--visited`, which will add `:visited` styles once a link has been clicked
- `visited` prop that will add the class that will allow visited styles when using React `Link`

**Removed**

- Links will no longer receive `:visited` styles by default.

#### Testing / Reviewing

Click the first link, and see that it **does not** change to visited styles. 
Click the second link, and see that it **does** receive visited styles.

Issue #3581, opened 2019-06-07
[Link]: Inline option should be same color as text and underlined by default


mjabbink:

vpicone:
@dakahn curious if you have any insights here.

mjabbink:
color alone cannot distinguish the inline link for sure. There must be an underline otherwise many cannot see the link at all making it pointless. Even in many cases that are not inline I recommend using the underline but optional is good. Example of no underline are anchor links and links in footer.

snidersd:
@dakahn IMO this issue can be handled in the documentation.  According to the Carbon Component usage documentation, links should be used primarily as navigational elements, and the style section has examples of horizontal and vertical link groups which are considered navigational links that are visually evident from the page design and context, as described in https://www.w3.org/TR/WCAG20-TECHS/F73.html, the failure is when links are used within the text, such as links in a paragraph or items in a list where some are different than others and are presented in colored text. See https://www.w3.org/TR/WCAG20-TECHS/G182.html. 
The technique states that “Visual cues can take many forms including changes to the font style, the addition of underlines, bold, or italics, or changes to the font size.” Therefore an underline is one of a variety of ways to indicate the text is a link.

vpicone:
Adding the additional requirement that the text-decoration should have the appropriate thickness in browsers that support defining this new property (Firefox, Safari, Safari IOS).

The feature is [not yet supported](https://caniuse.com/#feat=mdn-css_properties_text-decoration-thickness) in Edge or Chrome.

mjabbink:
The Technique we want to use is underline as the very articles use in the above links. We do not want inline links to bold or Italic.

vpicone:
@mjabbink per WCAG 2.0, an inline link is considered accessible if it has a 3:1 contrast ratio and an additional distinction is provided on hover.

<img width="693" alt="Screen Shot 2020-01-07 at 10 57 07 AM" src="https://user-images.githubusercontent.com/4078018/71913377-e5b71080-313c-11ea-8e97-b6bc2ac5ab8e.png">

vpicone:
Design requirements:
- All links: blue by default, current color variant
- Regular links: underline on hover
- Inline links: always underlined
- Remove visited styles

vpicone:
We went ahead and overrode/removed the `:visited` styles in the gatsby-theme per @mjabbink. Just a heads up that they're still very much a thing in Carbon.

mjabbink:
👍

PR #8177, opened 2021-03-24
fix(link): update visited link color

emyarod:
Related #3581
Related #5239
Related #7115
Related #8158

based on the latest spec from design

image

This PR restores visited link styles to the Link component. The removal of visited link styles was meant to be scoped to product and possibly dotcom, but it should remain in the core component library. The Gatsby theme should also be unaffected by this https://github.com/carbon-design-system/gatsby-theme-carbon/pull/697

#### Testing / Reviewing

Confirm that visited Carbon links receive the correct styles by setting the Link story's `href` knob to an address in your browser history

### DIFF
--- a/packages/components/src/components/link/_link.scss
+++ b/packages/components/src/components/link/_link.scss
@@ -47,7 +47,7 @@
     }
 
     &:visited {
-      color: $link-01;
+      color: $visited-link;
     }
 
     &:visited:hover {
--- a/packages/react/src/components/Link/Link.js
+++ b/packages/react/src/components/Link/Link.js
@@ -9,6 +9,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classnames from 'classnames';
 import { settings } from 'carbon-components';
+import deprecate from '../../prop-types/deprecate';
 
 const { prefix } = settings;
 
@@ -76,7 +77,11 @@ Link.propTypes = {
   /**
    * Specify whether you want the link to receive visited styles after the link has been clicked
    */
-  visited: PropTypes.bool,
+  visited: deprecate(
+    PropTypes.bool,
+    `\nThe prop \`visited\` for Link has been deprecated and will
+  be removed in the next major release.`
+  ),
 };
 
 export default Link;
--- a/packages/components/src/components/link/link.hbs
+++ b/packages/components/src/components/link/link.hbs
@@ -5,10 +5,7 @@
   LICENSE file in the root directory of this source tree.
 -->
 
-<a href="#"
-  class="{{@root.prefix}}--link{{#if inline}} {{@root.prefix}}--link--inline{{/if}}{{#if visited}} {{@root.prefix}}--link--inline{{/if}}">Link</a>
-<a href="http://www.google.com"
-  class="{{@root.prefix}}--link {{@root.prefix}}--link--visited {{#if inline}} {{@root.prefix}}--link--inline{{/if}}">Link</a>
+<a href="#" class="{{@root.prefix}}--link{{#if inline}} {{@root.prefix}}--link--inline{{/if}}">Link</a>
 <a class="{{@root.prefix}}--link{{#if inline}} {{@root.prefix}}--link--inline{{/if}}" aria-disabled="true">Placeholder
   link</a>
 <p class="{{@root.prefix}}--link--disabled{{#if inline}} {{@root.prefix}}--link--inline{{/if}}">Disabled Link</p>
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -3207,9 +3207,7 @@ Map {
         ],
         "type": "oneOf",
       },
-      "visited": Object {
-        "type": "bool",
-      },
+      "visited": [Function],
     },
   },
   "ListItem" => Object {
--- a/packages/react/src/components/Link/Link-story.js
+++ b/packages/react/src/components/Link/Link-story.js
@@ -17,7 +17,7 @@ const props = () => ({
   className: 'some-class',
   href: text('The link href (href)', '#'),
   inline: boolean('Use the in-line variant (inline)', false),
-  visited: boolean('Allow visited styles', false),
+  visited: boolean('[Deprecated]: Allow visited styles', false),
   onClick: ((handler) => (evt) => {
     evt.preventDefault(); // Prevent link from being followed for demo purpose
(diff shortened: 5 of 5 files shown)


laurenmrice (approved):
👍

tw15egan:
If we are setting visited styles all the time, do we have a need for the `visited` prop? Or should we flip that to `true` by default? Should we still give users a way to bypass `visited` styles and fallback to the correct link token?

emyarod:
that will require confirmation from design, given the original PR which added the `allowVisited` prop was reviewed under the assumption that design wanted optional visited styles

that being said, I believe changing to visited styles being opt-out would likely make the prop unneeded

laurenmrice:
Yes we do want optional visited styles still. Not all products want the visited style showing up because it can look messy.

packages/components/src/components/link/_link.scss
```diff
@@ -47,7 +47,7 @@
     }
 
     &:visited {
-      color: $link-01;
+      color: $visited-link;
```
tw15egan, line 50: I may be confused with this PR, but this change seems to imply we are no longer allowing teams to opt-out of having visited styles? This seems like it could cause unintended style issues in products with links that are navigated multiple times.

I believe the original intent was to provide `visited` styles but let teams allow them based on the `visited` prop, not enable them by default? 


https://github.com/carbon-design-system/carbon/pull/5239#issuecomment-580910038

https://github.com/carbon-design-system/carbon/pull/5239#issuecomment-581636154

emyarod: > this change seems to imply we are no longer allowing teams to opt-out of having visited styles

this PR allows teams to opt out of visited styles. can you explain where the implication is coming from?

emyarod: originally, visited styles were overridden which meant that teams did not have visited styles by default (referenced issues above) so visited styles were opt-in. but support for opt in visited styles was also just partial support due to the added requirement for a CSS class on top of the `:visited` pseudoclass

tw15egan: > this PR allows teams to opt out of visited styles. can you explain where the implication is coming from?

This PR is setting all `visited` links to purple. When I go to the `Link` story and enter an address I've been to, I am seeing `visited` styles. How would a team, that does not want visited styles, opt-out from receiving visited styles? 

Screen Shot 2021-03-29 at 5 27 36 PM

We set the `:visited` state to the link color so that teams would not get visited styles by default, but could opt into them by passing a `visited` prop. I'm just looking for clarification that we are changing our stance and now want all teams to have visited styles by default, as this will have a large impact across product teams. Is there a linked issue that says that we want to enable this functionality by default?

emyarod: to opt out of visited styles teams can target the `:visited` pseudoclass as it was intended. I think opt-in visited styles were flawed to begin with because there is no way to know whether or not a user has visited a link due to privacy issues. for further clarification there is also the attached design spec at the top of the thread

I'm also not sure there's any large impact to returning to following the HTML spec. what kind of impact are you anticipating?

tw15egan: My main concern is just that when product teams pull in this update, all visited links will be purple until they set their own `:visited` styles

emyarod: to me that change doesn't seem different from routine color contrast fixes or token updates, and the behavior would also fall in line with established usability patterns. I see the case for removing visited styles for use cases like the Gatsby theme or in brand/marketing but for the core Carbon component I think it would be a usability antipattern

tw15egan: > to me that change doesn't seem different from routine color contrast fixes or token updates

Not really, tokens aren't opt-in / out. There is just one value. If Cloud has guidance that visited links are the same color as normal links and Cloud PAL pulled in this update, they would then need to ensure they set the correct visited color. Based on Laurens's response in slack, this PR doesn't seem like it's needed. 

Screen Shot 2021-03-30 at 10 21 54 AM

[link](https://ibm-studios.slack.com/archives/GD8JG6JR3/p1617059068007400?thread_ts=1617058490.004700&cid=GD8JG6JR3)

We would also need to update guidance on the website if we change the default behavior 

Screen Shot 2021-03-30 at 10 23 07 AM

Happy to create a new issue that looks into the UX of this component and so we can get feedback from product teams about link behavior, but none of these linked issues are advocating that we should set visited styles on components by default.

emyarod: to clarify this PR was opened based on contradictory information which I received from design last week together with the spec image. although they have retracted that feedback and are now aligned I think the usability issues have been brought to light again (which is separate from whether or not downstream users need to update based on Carbon changes since that would be an expected consequence from reverting antipatterns)

that being said, the guidance would still remain the same, the Gatsby theme would remain unaffected, and the difference lies in what's exported in a themed package vs what's exported in the core component library. even though this issue has been brought up previously it isn't necessarily a hill I'm willing to die on but the lack of usability input on this subject is surprising esp since it directly led to an initial misdiagnosis of #8158

packages/react/src/components/Link/Link.js
```diff
@@ -76,7 +77,11 @@ Link.propTypes = {
   /**
    * Specify whether you want the link to receive visited styles after the link has been clicked
    */
-  visited: PropTypes.bool,
+  visited: deprecate(
+    PropTypes.bool,
+    `\nThe prop \`visited\` for Link has been deprecated and will
+  be removed in the next major release.`
+  ),
```
jnm2377, line 84: If we want users to opt in to visited link styles (meaning by default, there are no visited link style differences), then maybe we shouldn't deprecate this prop. I can see it being used to conditionally add a className that will then add purple visited link styles. i.e. Thoughts? @tw15egan @emyarod @laurenmrice

emyarod: just FYI the basis of the PR is to return to opt-out rather than opt-in styles based on the confusion it caused even our own issue triaging, as well as established research e.g. [1](https://www.nngroup.com/articles/change-the-color-of-visited-links/) [2](https://www.usability.gov/sites/default/files/documents/guidelines_book.pdf) [3](https://design-system.service.gov.uk/styles/typography/)

that being said, given that @dakahn has removed the request for review it seems this is not an a11y concern and primarily a ux one

dakahn: Yeah sorry, meant to comment. As long as contrast requirements are met and we have a logical consistency across the system with various states for links we're good from an accessibility and usability standpoint.

PR #8181, opened 2021-03-24
feat(CodeSnippet): set min number of rows

guigueb:
feat(CodeSnippet): set min number of rows

Closes #

#8056

set default min to 100%
apply the appropriate minHeight collapsed or expanded as needed
add props to component and story... minCollapsedNumberOfRows and minExpandedNumberOfRows
use minExpanded when clearing expandedCode... minExpandedNumberOfRows could be < maxCollapsedNumberOfRows

#### Testing / Reviewing

Testing via Storybook pages may be troublesome as you can't manipulate the contents of the CodeSnippet
to see the reactions when there are fewer rows of data than the min settings.

Also note: the code does not have any 'adjusting' checks if the client passes odd settings.
ie: it makes no since to set minExpandedNumberOfRows > maxExpandedNumberOfRows so there is no 'adjusting' code.
As these are client set - not settable/changeable by the user - I don't think 'adjusting' checks are needed.


tw15egan (approved):
LGTM 👍 ✅

jnm2377 (approved):

PR #8189, opened 2021-03-25
feat(icon-build-helpers): add module-info extension and next react builder

### LINKED ISSUE #8179: Update icons-react to ship icon components with size prop
This issue tackles the integration of our new "icon with size prop" approach for our icons-react package. For this issue, we will need to:

- [ ] Emit icons as individual files, e.g. `@carbon/icons-react/IconName.js` or `@carbon/icons-react/namespace/IconName.js`
- [ ] Support the `size` prop for each component
- [ ] If an icon has bespoke assets per-size, the `size` prop should use that data to handle rendering.

joshblack:
Closes #8179

This PR updates our `icon-build-helpers` with a new experimental builder for v11. This builder outputs a directory called `next` which contains a flat list of all of our icons. The intent is that teams would require icons from these sub-paths instead of importing from an entrypoint. This bundle is built using CommonJS, although this could change in the future.

<details>
<summary>Each icon has approximately the following shape before being compiled:</summary>

```jsx
import React from 'react';
import Icon from './Icon.js';

const ComponentName = React.forwardRef(
  function ComponentName({ children, size = 32, ... rest}, ref) {
    return (
      <Icon width={size} height={size} ref={ref} {...rest>
        <path d="..." />
        {children}
     </Icon>
  }
);

ComponentName.propTypes = {
  sizes: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
};

export default ComponentName;
```
</details>

<details>
<summary>For icons that have multiple asset sizes, we conditionally render each in the component based on an exact match:</summary>

```jsx
const ComponentName = React.forwardRef(
  function ComponentName({ children, size = 32, ... rest}, ref) {
    if (size === 16) {
      return (
        <Icon width={size} height={size} ref={ref} {...rest>
          <path d="..." />
          {children}
       </Icon>
      );
    }

    return (
      <Icon width={size} height={size} ref={ref} {...rest>
        <path d="..." />
        {children}
     </Icon>
  }
);
```
</details>

To accomplish this new format, we introduce a new extension in our metadata ecosystem: `module-info`. This extension iterates through our icons, extracting asset data, optimizing it, and converting it to an AST to be used by a builder. It also embeds information about the JavaScript module, like name and filepath.

#### Changelog

**New**

- Add new extension, `module-info`, to power the experimental builder for React icon components
- Add new builder, `reactNext`, for building our v11 set of icons

**Changed**

- Update the storybook example to work with Yarn 2 (sorry about the diff size 🤦 )

**Removed**

#### Testing / Reviewing

- Visit the storybook for icons-react
- Make sure the icon renders and no console errors are emitted
- Edit the index.stories.js file
  - Add a size prop, set it to 32 and verify the component renders
  - Change the size prop to 16 and verify the path data changes (menu uses multiple assets)
- Verify again that no prop warnings have been emitted


packages/icon-build-helpers/src/builders/react/next.js
```diff
@@ -0,0 +1,256 @@
...
+  // specific size
+  const maxSize = sizeVariants.reduce((maxSize, { size }) => {
+    if (size > maxSize) {
+      return size;
+    }
+    return maxSize;
+  }, -Infinity);
+
+  // Each asset that is not used for our "default" icon will conditionally
+  // render from an if statement in our compnoent
```
joshblack:
```suggestion
  // render from an if statement in our component
```

packages/icon-build-helpers/src/builders/react/next.js
```diff
@@ -0,0 +1,256 @@
...
+  });
+
+  // The "default" icon that will be rendered, based on the max size
+  const returnStatement = sizeVariants.find(({ size }) => {
+    return size === maxSize;
+  });
+
+  // We build up our component source by adding in any necessary deprecation
+  // blocks along with conditionally rendering all asset sizes along with a
+  // default return for the "default" icon
```
joshblack:
```suggestion
  // We build up our component source by adding in any necessary deprecation
  // blocks along with conditionally rendering all asset sizes. We also use a
  // return statement for the "default" icon
```

packages/icon-build-helpers/src/builders/react/next.js
```diff
@@ -0,0 +1,256 @@
...
+const { babelConfig } = require('./next/babel');
+const { svgToJSX, jsToAST } = require('./next/convert');
+const templates = require('./next/templates');
+
+// This builder outputs a collection of CommonJS modules representing our icon
+// components. It does not generate an `index.js` entrypoint file due to the
+// number of icons we currently support (>1600)
+//
+// Each CommonJS module includes an icon component. For more information on the
+// structure of each module, you can view `createIconEntrypoint`
```
joshblack:
```suggestion
// structure of each module, checkout `createIconEntrypoint`
```

packages/icon-build-helpers/src/builders/react/next.js
```diff
@@ -0,0 +1,256 @@
...
+          path.resolve(__dirname, './components/Icon.js'),
+          'utf8'
+        ),
+        './iconPropTypes.js': `
+          import PropTypes from 'prop-types';
+
+          export const iconPropTypes = {
+            size: PropTypes.oneOfType([
+              PropTypes.number,
+              PropTypes.string,
```
emyarod: `size` accepts strings but `width` and `height` seem to take numbers only

image

joshblack: Thanks @emyarod! Great catch. Updated the `Icon.js` prop type to include number/string which should handle this.

dakahn (approved):
Looks good! Built without errors on my machine 👍🏽

vpicone:
@joshblack did you experiment at all with a [shared entrypoint](https://unpkg.com/browse/react-icons@4.2.0/ai/index.js)? I feel like named exports are an expectation from a developer standpoint.

For inspecting the generated code, would it be fair for us to remove ie11 from the browserlist? You mentioned it only exports cjs now, is there a blocker for not exporting both cjs/esm from the start?

joshblack:
@vpicone we could totally go for an entrypoint again. Happy to add it in (and include one for ESM too) if we want.

Basically the situation we want to avoid is running into this again: https://github.com/carbon-design-system/carbon/issues/6390. 

Basically after a certain size babel will just give up on an entrypoint file, which we then switched to using re-exports which caused long require times in Node.js, and then finally settled on that "bucket" approach which basically split up the entrypoint into smaller files and re-exported from them.

vpicone:
@joshblack I definitely think named exports are worth looking into, with enough icons on a page, the imports might get pretty gnarley. Couple of other things:

If a size isn't specified should it default to '1em'? Looks like it defaults to `32x32` at the moment.

The bespoke asset behavior works when size is a number, but not when passed as a string... <sub>maybe this is a feature, not a bug :sweat_smile:</sub>

Screen Shot 2021-04-01 at 5 49 18 PM

joshblack:
>  I definitely think named exports are worth looking into, with enough icons on a page, the imports might get pretty gnarley

@vpicone I just don't know what to do if we run into the problem again because effectively we have the same data, we're just arranging it differently. In other words, we can't embed all the icons source files in the same `index.js` because of babel breaking and if we do the re-export pattern then we run into that Disk I/O problem from that issue.

We can do the bucketing approach again, I guess?

For the number thing, I guess we could match on `'16'` and `'16px'`? I'm not sure though

vpicone:
> @vpicone I just don't know what to do if we run into the problem again because effectively we have the same data, we're just arranging it differently.

Using the current overrides we'll be cutting the number of exports down from 6336 to 1675. On my disk that brings the source file down from 120kb to 33kb. Not the end of the world but worth exploring maybe 🤷 

> For the number thing, I guess we could match on `'16'` and `'16px'`? I'm not sure though

I think that would be the expected behavior right? Considering "16" is valid svg length and equates to pixels, I'd expect it to serve the same asset as `16`. Having the prop's type influence the asset seems non-deterministic.

joshblack:
@vpicone making the update now for the different conditional checks 👍 

For the `index.js` problem, we run into two situations:

**Embed all icon data in one file**

For this one, we're looking at a file that is ~1.7MB in size (I think for font awesome they sit at 1.3MB ish for their entrypoint in `react-icons`)

**Re-exports files from module paths**

For this option, we run into the fact that we have ~1650 file paths to resolve in Node.js. Thankfully this is down from 6000 that we had but ultimately these many imports were causing Node.js to run into some problems with resolution time.

I'll go ahead and add support for the first strategy and we can move to the second/bucket abstraction as-needed 👍

joshblack:
@vpicone just pushed the updates 👍  should have entrypoints for CommonJS/ESM along with logical expression for different options for `size`

vpicone (approved):

Issue #8209, opened 2021-03-29
feat(ComboBox) ComboBox need more/better support of downshift


guigueb:
### Summary

Carbon uses downshift to deliver ComboBox.
Carbon supports some of the downshift options directly as ComboBox props.
Carbon also supports a ComboBox.downshiftProps prop for additional downshift options.

The Carbon ComboBox use of downshift options is mixed, sometimes ignoring, sometimes using, sometime mis-using, them.

```
The Carbon Client should not have to search/read the Carbon ComboBox **CODE** to figure out what is possible.

Ideally, all downshift options should be supported.

However, if the Carbon ComboBox will not support a downshift option or if it modifies a downshift option... those downshift options should be well documented by Carbon.

This can be in the form of...
- a whitelist; listing specifically what downshift options are supported.
- Or a blacklist; listing specifically what downshift options are not supported.
```

### Justification

The support of the ComboBox.downshiftProps props is essential for client to control ComboBox. 
Clients need to have better control to support their applications, to support functionality, styling, and testing.
And in some cases to fix Carbon issues: see #8070.

### Desired UX and success metrics

This is a hard one.

From a Carbon point of view the internal use of downshift should have no impact at all.
The internal use of downshift by Carbon should deliver a working ComboBox as per the Carbon design.
The success metrics for this is that Carbon ComboBox works as per the Carbon design.

The issue is for the support of downshift within ComboBox.
So the effect on UX is completely based on the downshift options the Client passes to Carbon ComboBox.
The success metrics for this is that Carbon ComboBox accepts and reflects the downshift options supplied by the user.

### "Must have" functionality

The must have's is a hard list to determine.
It is based on what downshift options are deemed essential to support/pass on.
Ideally, every/any downshift option should be supported.

If there are some downshift options that are not supported, or modify the clients use of the downshift options...
...they should be well documents by Carbon.

The Carbon Client should not have to search/read the Carbon ComboBox code to figure out what is possible.

Some examples of some of the issues Carbon has with the downshift options...

1. the downshift getter-props should not be applied to <DownShift> but applied the the getter-props directly
2. some downshift options are supplied as direct props to ComboBox (presumably for convince), in this case they need to be 'looked for' as direct props and as downshiftProps.
3. when downshift options are used internally they need to take into account passed in downshift options (by either applying or calling them)
4. some downshift options are state options in ComboBox, making them completely inaccessible as downshift options
5. there may be others

```
[...]

PR #8211, opened 2021-03-29
feat(ComboBox): the ComboBox need more better support of downshift

guigueb:
This is not a real PR.

This has been created as a way to help show downshift issues with ComboBox.
See https://github.com/carbon-design-system/carbon/issues/8209

This is not THE way to change things.
This is not even A way to change things.

This is simply to help show what the downshift issues with ComboBox are.

Not all of the occurrences of the issues are shown here.
When addressing the issues, we need to keep in mind other similar issues.

ie:
For issue (2) I demoed/reworked handleOnChange().
This may not be the only use of onChange that needs to be looked at.
And I did not point out the use of the other direct/downshiftProps usages that need to be looked at.

**
I showed issues for (1) and (2).
I was going to show issues for the other issues but based on Joshblack's comment below I opted not to.
It seems like the API for ComboBox integration will will changing, as such, the support for downshift may not be needed.

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.js
+++ b/packages/react/src/components/ComboBox/ComboBox.js
@@ -372,11 +372,24 @@ export default class ComboBox extends React.Component {
       [`${prefix}--text-input--empty`]: !this.state.inputValue,
     });
 
+    const {
+      // (1) extract the downshift prop-getters to apply later
+      getInputProps: otherGetInputProps = (v) => v || {},
+      getItemProps: otherGetItemProps = (v) => v || {},
+      getLabelProps: otherGetLabelProps = (v) => v || {},
+      getMenuProps: otherGetMenuProps = (v) => v || {},
+      getRootProps: otherGetRootProps = (v) => v || {},
+      getToggleButtonProps: otherGetToggleButtonProps = (v) => v || {},
+      // (1) otherDownShiftProps include all non-prop-getters (actions, state, and props(basic and advanced) )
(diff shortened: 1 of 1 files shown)


joshblack:
@guigueb thanks for taking the time to make this PR! It's incredibly helpful for demonstrating the issues.

I think that, unfortunately, support for full integration with downshift will be minimal. We ran into a tough situation this last release cycle (v10.x) when we exposed library APIs as part of our component APIs and it made updating dependencies incredibly difficult, especially for faster moving targets like Downshift. (For reference, Downshift has moved from v1 to v6 while we've stayed on v10.x) For the rest of v10.x, we'll likely keep support limited to the Component API for the `Downshift` component or equivalent hooks versus and not expose any of the prop getters.

Our current thinking is that we'll reverse this API design decision and instead expose the listbox primitives we have for teams to use with downshift directly. This is definitely still a part of our v11 release planning though so it could change but just wanted to share our current thoughts here.

guigueb:
@joshblack
Thank you for the heads up on the API changes.
You may wish to update the related feat #8209 on this.
BTW:
 downshift is very impressive - I'm glad that our ComboBox issues exposed me to it.
 I will be sorry to see that we won't be supporting/exposing it to clients.
 I'll be interested to see how the API changes.

joshblack:
Sounds good @guigueb! I'll go update that issue and close out this PR and the issue.

Let me know if this is a mistake though and I'd be happy to re-open them!

PR #8215, opened 2021-03-29
fix(ie11): add ie11 to autoprefixer, add grid: autoplace

### LINKED ISSUE #6919: Modals with lots of content don't size correctly on IE

tw15egan:
Closes #6919
Refs https://github.com/carbon-design-system/carbon/pull/8176

Adds `ie >= 11` to our browser matrix, also enables the `grid: autoplace` feature of autoprefixer. There are some [known caveats](https://github.com/postcss/autoprefixer#does-autoprefixer-polyfill-grid-layout-for-ie). If someone could test these in IE11 that would be great (I do not have access at the moment)

#### Changelog

**New**

- `grid: autoplace` feature of autoprefixer

**Changed**

- `ie >= 11` to our browser matrix

#### Testing / Reviewing

Any changes should just be scoped to IE11. A quick search of the compiled css shows 18 instances of the new `-ms-grid` vendor prefix. The following components should be tested in IE11:

- [ ] `FileUploader` (`.bx--file__selected-file`)
- [ ] Inline `Listbox` / `Dropdown` (`.bx--list-box__wrapper--inline` and `.bx--list-box__wrapper--inline .bx--form-requirement`)
- [ ] `Modal` (`.bx--modal-container`, `.bx--modal-header`)


joshblack (approved):

tay1orjones (approved):

wkeese:
Thanks for the fix.  FWIW, I did try to test this on IE11, from the link you gave, but Storybook itself won't run on IE; I think it needs some polyfills / transpilation.

Issue #7525, opened 2021-01-11
Conduct performance audit of Sass @import versus @use for codebase


joshblack:
We've been heavily considering transitioning over to Sass Modules (`@use`), in part because it seemed like it would improve our compilation time for projects using our Sass files.

This issue is to conduct a rough performance audit of our current Sass codebase with `sass-loader` and `@import` and compare it to a converted codebase with Sass Modules.

joshblack:
**Baseline with `sass-loader` and `sass` implementation with no Carbon styles**

Cold start
56533ms

Recompilation
829ms
619ms
767ms

**With `sass-loader` and `@import`**

Cold start
138502ms

Recompilation
50538ms
54899ms
55997ms

Average: 53808ms

**With `sass-loader` and `sass` implementation**

Cold start
95193ms

Recompilation
42531ms
45316ms
40227ms

Average: 42691.33ms

joshblack:
**With `sass-loader`, `sass` implementation, and `@use`**

Cold start
57257ms

Recompilation
3397ms
2594ms
3236ms

Average
~3075ms

PR #8228, opened 2021-03-30
feat(storybook): update sass loaders and use dart-sass

joshblack:
Update storybook config and loader dependencies to use dart-sass, enables use of sass modules in the storybook.

#### Changelog

**Changed**

- update and reconfigure sass loaders for storybook

**Removed**

- remove `fast-sass-loader`

#### Testing / Reviewing

The deploy preview should show the CSS Grid styles in the Auto Columns story as an example. Once approved I can remove the example story.


joshblack:

config/storybook-preset-carbon/index.js
```diff
@@ -36,32 +36,36 @@ module.exports = {
...
+          implementation: require('sass'),
+          includePaths: [path.resolve(__dirname, '..', '..', 'node_modules')],
         },
         sourceMap: true,
       },
     };
 
     const fastSassLoader = {
-      loader: 'fast-sass-loader',
+      loader: require.resolve('fast-sass-loader'),
```
tay1orjones: I don't think we should be using and recommending `fast-sass-loader` as we begin to embrace sass modules. It [doesn't support `@use`](https://github.com/yibn2008/fast-sass-loader/issues/67), but beyond that, the ways it achieves a faster compile time don't seem to match with how `@use` is [intended to work](https://sass-lang.com/blog/the-module-system-is-launched#use-the-heart-of-the-module-system) with `@import` and vice-versa.

> In addition to namespacing, there are a few important differences between `@use` and `@import`:
> 
> - `@use` only executes a stylesheet and includes its CSS once, no matter how many times that stylesheet is used.
> - `@use` only makes names available in the current stylesheet, as opposed to globally.
> - Members whose names begin with `-` or `_` are private to the current stylesheet with `@use`.
> - If a stylesheet includes `@extend`, that extension is only applied to stylesheets it imports, not stylesheets that import it.

`fast-sass-loader` uses regex to grab all the imports and hoist them all into one file. It then caches the result to avoid recompiling. I tried patching it locally to modify the regex to include `@use`, but when it passes the result off to `sass` it still can't resolve files, even the [built in modules](https://sass-lang.com/blog/the-module-system-is-launched#controlling-namespaces) fail. If the resolution problem was fixed, I don't think we'd see appropriate namespacing control as everything would essentially be global. I suspect it would also impact usages of `@forward`.

To my eye, `@use` sort of turns the value proposition of the library upside down. Especially after @joshblack saw compile times significantly improve with sass modules, https://github.com/carbon-design-system/carbon/issues/7525

tay1orjones (approved):
I pushed a commit to remove `fast-sass-loader` and use `sass-loader` in both dev and prod. I tested a fresh storybook dev (`yarn start`) and see an increase in build time: 

| `fast-sass-loader` | `sass-loader` |
| - | - |
| 42s for manager | 51s for manager |
| 53s for preview | 1min 46s for preview |

Recompile time should be less. I also think we'll see a decrease when using sass modules? Also it might help if we imported component-level modules per `*-story.js` file instead of including all styles globally for the whole storybook? 

I'm didn't add the [`fiber` optimization](https://github.com/webpack-contrib/sass-loader#implementation) as it's been deprecated for some time. It was [updated yesterday](https://github.com/laverdet/node-fibers/commit/8f2809869cc92c28c92880c4a38317ae3dbe654d) noting it's incompatible with Node v16. 

All in, I think this should be ready to review now. The deploy preview should show the CSS Grid styles in the `Auto Columns` story as an example. Once approved I can remove the example story.

joshblack:
bump @andreancardona @emyarod when you get a chance to review! @tay1orjones just pushed up updates and moved this from draft.

emyarod (approved):

andreancardona (approved):
yahooo 🎉

tw15egan:
> I pushed a commit to remove `fast-sass-loader` and use `sass-loader` in both dev and prod. I tested a fresh storybook dev (`yarn start`) and see an increase in build time:
> 
> `fast-sass-loader`	`sass-loader`
> 42s for manager	51s for manager
> 53s for preview	1min 46s for preview
> Recompile time should be less. I also think we'll see a decrease when using sass modules? Also it might help if we imported component-level modules per `*-story.js` file instead of including all styles globally for the whole storybook?
> 
> I'm didn't add the [`fiber` optimization](https://github.com/webpack-contrib/sass-loader#implementation) as it's been deprecated for some time. It was [updated yesterday](https://github.com/laverdet/node-fibers/commit/8f2809869cc92c28c92880c4a38317ae3dbe654d) noting it's incompatible with Node v16.
> 
> All in, I think this should be ready to review now. The deploy preview should show the CSS Grid styles in the `Auto Columns` story as an example. Once approved I can remove the example story.


Working locally I'm seeing quite a slowdown in both initial startup, as well as any style updates. It's taking me ~50s for style changes to appear in the storybook when running locally. Anything we can do to speed up dev speed?

Issue #8149, opened 2021-03-22
Update ~25% of components to use new design tokens


joshblack:
As part of #8148, this issue is responsible for taking the first step in updating our components to use the new design tokens. There are probably several strategies that we could follow to implement this, including:

- Add new tokens to the existing themes and have them fallback to the current tokens
- Add new tokens to the existing themes and then toggle usage based on feature flags in the component scss files

It is up to the implementors to figure out which approach will work best going into this release for our team 👀

tw15egan:
Will need to wait for https://github.com/carbon-design-system/carbon/issues/8238 before starting this

tw15egan:
Component token spec: https://github.com/carbon-design-system/carbon/issues/8187

PR #8241, opened 2021-03-30
feat(theme): add new color tokens

### LINKED ISSUE #8238: Add new tokens to theme package
The first step to closing #8149 is to add the new tokens to the theme packages. 

Until we create an alpha branch, we will add the new tokens to the theme package as references to the current tokens. That way, components can be updated to use the new tokens, but the old tokens will still be available. 

For instance: 

```js
// current token references IDL color
export const interactive01 = blue60;

// new token, references old token until we create an alpha branch
export const backgroundBrand = interactive01;
```

Once we have an alpha branch, we can remove the old tokens, and have them reference the IDL colors directly

```js
export const backgroundBrand = blue60;
```

jnm2377:
Closes #8238 
Rel: #8149 

Adds new color tokens with existing tokens as a fallback. Will need to update again when we release v11 to deprecate existing tokens and use new color tokens without fallback. 

#### Testing / Reviewing

- Double check that new token names are correct, and are using the correct fallback color: [color token list](https://www.notion.so/carbondesignsystem/Final-List-of-new-tokens-fc5985c7331341baa6546064399d36da)


jnm2377:

aagonzales:
@jnm2377 Looks like the button tokens and all the state tokens are missing.

Notion is stupid in that it doesn't load the whole table automatically. You'll need to hit the "load 50 more" button at the bottom. 
image

jnm2377:
lol my bad @aagonzales updated! 👍🏽

tw15egan (approved):
Looks great! 🎉 👍🏻 ✅

tw15egan:
@jnm2377 I think we need to add the tokens [here](https://github.com/carbon-design-system/carbon/blob/main/packages/themes/src/tokens.js#L163) to fix the failing tests

jnm2377:
@tw15egan updated! 🤞🏽

aagonzales (approved):
Awesome, looks like they're all there now!

tay1orjones (approved):

PR #8242, opened 2021-03-30
docs(CodeSnippet): Update snippet docs

### LINKED ISSUE #8088: CodeSnippet: Storybook needs some lovin
### Summary

The CodeSnippet Storybook examples need a little work to be more clear/helpful to users.
This is a storybook documentation issue.

### Justification

Clients trying out CodeSnippet in the Storybook environment would like it to be more helpful.

Each story should have story code showing the exact props used to create it (not deconstructed props `{...props}`).

Each story should have a list of knobs available for it.
- in CodeSnippet each story is a variant of CodeSnippet which works on a subset of props, we should show only the knobs that relate to the variant
- but in other components only the playground has knobs (we could go this route)

All the props in CodeSnippet should be reflected in knobs.
The following props are missing... className, children, onClick.

In knobs the prop being applied should be clear.
The prop `Copy button title _____` doesn't clearly indicate what prop is being used.
Possibly follow the other descriptions `Copy button title (copyButtonDescription) ____`.
But I would suggest all of carbon adopt a standard... 
- `propNameFromTheCode: propDesciptionFromTheCode _______`
- which in this case is... `copyButtonDescription: Specify the description for the Copy Button ____`

### "Must have" functionality

Ideally, to be most helpful to clients, in all stories, not just CodeSnippet, we would show...
- the exact code used to create it, not deconstructed props (the client should be able to reproduce the example via copy/paste)
- all the knobs/props available to that story (a subset of them all it if is a variant) reflecting the default values or the values needed to show the story

anytime

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8088
Refs https://github.com/carbon-design-system/carbon/pull/8181

Updates the docs so that they are more helpful  

#### Changelog

**New**

- All 3 variants are rendered in the `Playground` story, and you can change the content displayed in them via the knobs

**Changed**

- Added the prop names to all knobs

**Removed**

- `Multiline` story no longer spreads props so it can be copy / pasted

#### Testing / Reviewing

Ensure CodeSnippet stories are copy / pastable (exception of playground story) and that the prop names are added to the knobs.
Ensure you can change the content of the code snippets


guigueb:
Nice.
I think you missed the copyLabel prop.

And it may be hard to do - but could we get children(CodeSnippet data) to be a prop too.
With the single and multi line variants, modifying the CodeSnippet data will help test the wrapText prop and scollbar behavior.
But most of all it is needed to test the behavior of the multi line min/max - closed/expanded props.

guigueb:
My bad, looks like you already have the children(CodeSnippet data) as a knob prop.
I got lost looking for a prop named (children).

dakahn (approved):

emyarod (approved):

PR #8243, opened 2021-03-30
fix(NumberInput): pass down event change value

### LINKED ISSUE #7457: NumberInput's `onChange` event does not include the value in `event.target.value` when the arrows are used.

- [ ] `carbon-components`

## Detailed description

When using the arrows on the `NumberInput` component, the `event`
parameter of `onChange` does not contain the actual value in `event.target.value`. Instead, the value is only in `event.imaginaryTarget.value`. However, when typing the number directly in the component, both `event.target.value` and `event.imaginaryTarget.value` contain the correct number I am typing.

Yes, `NumberInput`.

> see changed?

I would like to see `event.target.value` contain the correct number when typing directly and also when using the up and down arrows.

Firefox and Chrome

carbon-components.react 7.17.0

> should be aware of?

Instantiate a `NumberInput`, add the `onChange={(event) => {...}}` property and print the `event.target.value`. Then use the arrows in the component. The printed value will be sometimes undefined, and sometimes empty string. By typing in the component, the values of the printed `event.target.value` will be the expected ones.

>
> - Style and vanilla JS:
> - React:

## Additional information

- Screenshots or code
- Notes

jnm2377:
Closes #7457

When using keyboard to update number input, users can now access the updated value. We are now passing down the value as a second parameter in the onChange handler.

#### Changelog

- now passing down `value` in the keyboard onChange handler.
- also now passing it down in the onClick even if `useControlledStateWithValue` is false.

#### Testing
- view storybook example and check your console. Make sure you can the accessed value is the current value for the number input. Can test for click/keyboard updates.


packages/react/src/components/NumberInput/NumberInput-story.js
```diff
@@ -20,6 +20,7 @@ import NumberInput from '../NumberInput';
 import { NumberInput as OGNumberInput } from './NumberInput';
 import NumberInputSkeleton from '../NumberInput/NumberInput.Skeleton';
 import mdx from './NumberInput.mdx';
+import { console } from 'window-or-global';
```
emyarod:
```suggestion
```

packages/react/src/components/NumberInput/NumberInput-story.js
```diff
@@ -88,6 +89,7 @@ export const Default = () => {
       label="NumberInput label"
       helperText="Optional helper text."
       invalidText="Number is not valid"
+      onChange={(e, something) => console.log(something)} //TO-DO: remove before merging
```
emyarod:
```suggestion
```

FYI storybook already shows the event object in the actions tab so this is probably not needed for testing!

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -289,8 +289,8 @@ class NumberInput extends Component {
             onClick && onClick(evt, { value, direction });
             onChange && onChange(evt, { value, direction });
           } else {
-            onClick && onClick(evt, direction);
-            onChange && onChange(evt, direction);
+            onClick && onClick(evt, { value, direction });
+            onChange && onChange(evt, { value, direction });
```
emyarod: I believe this would be a breaking change since the function signature is changing so you probably want to keep the existing signature and add to it instead of changing its shape

emyarod (commented):
would you mind looking into the failing tests? I believe it's related to the function signature change being a breaking change

packages/react/src/components/NumberInput/NumberInput.js
jnm2377: ah, makes sense. Will update! 👍🏽

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -289,8 +289,8 @@ class NumberInput extends Component {
             onClick && onClick(evt, { value, direction });
             onChange && onChange(evt, { value, direction });
           } else {
-            onClick && onClick(evt, direction);
-            onChange && onChange(evt, direction);
+            onClick && onClick(evt, { value, direction });
```
joshblack, line 292: This might be something that we'll need to wrap behind a feature flag since this will act as a breaking change for the `onClick` prop. In other words, this usage would break with this change:

```js
<NumberInput
  onChange={(event, direction) => {
    if (direction === 'down') {
      // ...
    }
  })}
/>
```

packages/react/src/components/NumberInput/NumberInput-story.js
```diff
@@ -80,16 +80,7 @@ export default {
...
-    <NumberInput
-      min={0}
-      max={100}
-      value={50}
-      label="NumberInput label"
-      helperText="Optional helper text."
-      invalidText="Number is not valid"
-    />
-  );
+  return <NumberInput {...props()} />;
```
tw15egan: Can we keep this as-is so that the default story can be copy / pasted? We can add in additional props to the component if we want to display them, but we're trying to only use the `...props()` syntax in the `Playground` example

packages/react/src/components/NumberInput/NumberInput-story.js
```diff
@@ -88,6 +88,9 @@ export const Default = () => {
       label="NumberInput label"
       helperText="Optional helper text."
       invalidText="Number is not valid"
+      id="number-input-id"
+      onClick={action('onClick')}
+      onChange={action('onChange')}
```
emyarod:
```suggestion
```

I think this is fine to remove if the aim is to keep the default story for easy copy pasting

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -257,7 +257,7 @@ class NumberInput extends Component {
           if (useControlledStateWithValue) {
             onChange(evt, { value });
           } else if (onChange) {
-            onChange(evt);
+            onChange(evt, { value: Number(value) });
```
emyarod: should this return the same params as `handleArrowClick`?

packages/react/src/components/NumberInput/NumberInput-test.js
```diff
@@ -466,13 +470,20 @@ describe('NumberInput', () => {
         downArrow.simulate('click');
         expect(onClick).toBeCalled();
         expect(onChange).toBeCalled();
-        expect(onClick).toHaveBeenCalledWith(expect.anything(), 'down');
+        expect(onClick).toHaveBeenCalledWith(
+          expect.anything(),
+          'down',
+          expect.anything()
```
emyarod: since this is the value should we make this assertion more specific? (number and/or string)

joshblack (changes requested):
This looks great! I think the one note is just what @emyarod had shared already, would be great for `onClick` and `onChange` to have the same signatures. For v10 I guess that could be:

```ts
onClick(evt, direction, value);
onChange(evt, direction, value);
```

But in v11 it would be:

```
onClick(evt, { direction, value });
onChange(evt, { direction, value });
```

?

packages/react/src/components/NumberInput/NumberInput-test.js
jnm2377: I think, ideally, yes @emyarod I'm not entirely sure why `direction` isn't passed down as a param here originally. I can add it if we want them to be the same. thoughts? @joshblack

emyarod: was this meant to be a reply to a different review comment (https://github.com/carbon-design-system/carbon/pull/8243/files/18dba3ae39f5e3e6826c9b48536849a7c9dc47c7#r605683234)?

jnm2377: oops, yes. 😓 @emyarod

packages/react/src/components/NumberInput/NumberInput.js
emyarod: @jnm2377 to answer [your question](https://github.com/carbon-design-system/carbon/pull/8243#discussion_r612565905) they should both return the same params

joshblack (approved):

tw15egan (approved):

PR #8255, opened 2021-03-31
fix(FileUploader): replace <strong> with <p> in exported component

### LINKED ISSUE #8252: Use of strong in file uploader causing accessibility issue
From https://github.com/carbon-design-system/carbon-components-vue/pull/1145

> Hello guys,
> We have an accessbility issue in cv-file-uploader component label, <strong/> tag is not the best option for heading contents. To fix this issue I tried to change to <p/> tag, so the Accessbility Checker stoped warning about this problem. Can you check if the alteration is possible?

emyarod:
Closes #8252

This PR replaces a `<strong>` label in the exported `FileUploader` with `<p>` to remove a11y warnings about header text. The example applications have also been updated to match 

#### Testing / Reviewing

Confirm the `FileUploader` is no longer exported with `<strong>` and that there are no related a11y warnings


joshblack (approved):

aagonzales (approved):
Looks good to me, I see no visual update which I think its correct.

tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

tw15egan:
@emyarod I think you just need to sign the DCO

PR #8258, opened 2021-03-31
fix(Tile): update outline-offset

### LINKED ISSUE #6306: `Radio Tile Group` outline disappear problem
Hi, I'm Zhuoyue from the IBM Watson Studio UI team and am currently working on developing UI for AutoAI.

The `Radio Tile Group`'s **outline disappears** under 70% Zoom or smaller. This seems to be the bug for Chrome alone, in the safari everything is fine.

**Safari**      
image 
**Chrome**
image

**Safari**      
image
**Chrome**
image

A temporary fix we are using is to add the following code inside the `tilesContainer` of the respective `.scss`, but this only partially solves the problem. If we zoom in 67% or less, we need an even smaller `outline-offset` to make sure the outline shows up correctly. But under smaller `outline-offset`, the 110% or more zoom in will looks weird. So I think this component should be updated for this situation. Thanks!

```css
:global(.bx--tile--is-selected){
	outline-offset: -1.5px !important;
}
```

andreancardona:
Closes #6306

- fixes RadioTile outline bug that can be seen [here](https://github.com/carbon-design-system/carbon/issues/6306) &  below:
Screen Shot 2021-03-31 at 2 44 40 PM

#### Testing / Reviewing

- [ ] Checkout: https://deploy-preview-8258--carbon-components-react.netlify.app/?path=/story/components-tile--radio
- [ ] Zoom view to less than 100% - ex: 90%, 80%, etc.
- [ ] Make sure Tile --> Radio & MultiSelect outlines can be seen
  - Select a tile 
  - Then click out of the container to confirm bug is fixed


tw15egan:
The radio looks great!

Still seeing some cutoff on the multi select variant
Screen Shot 2021-03-31 at 5 13 34 PM

andreancardona:
> The radio looks great!
> 
> Still seeing some cutoff on the multi select variant
> Screen Shot 2021-03-31 at 5 13 34 PM

Ooo I see! Going to confirm a visual change. Converting this to a draft in the mean time

andreancardona:
@laurenmrice would like your opinion on the visual change to the black border to 2px for the following reasons:

- It matches the focus state 2px outline
- and fixes all the bugs/issues

Thank you!

laurenmrice:
@andreancardona The border for the selected state needs to still be 1px. We use 1px for borders like this across our system unless it is a focus state.

andreancardona:
@tw15egan should be ready for review!

jnm2377 (approved):
LGTM! :)

laurenmrice (changes requested):
- When there are two selected tiles next to eachother, there should not be an overlap of the tile borders. It should still look like 1px border between the tiles. 
Artboard


- When you select a tile, there is a slight jump of the tile text moving out of place. The text should always remain in the same place not matter what state.
Apr-02-2021 16-00-08

andreancardona:
@laurenmrice I'll make sure to look into the jumping text (I think it's an outline v border issue)

But when there are 2 selected tiles - that is the [current styling](https://react.carbondesignsystem.com/?path=/story/components-tile--multi-select)- is that something we still want to update?
Screen Shot 2021-04-02 at 5 40 59 PM

laurenmrice:
@andreancardona Yes, we want to fix that, that is a bug.

packages/components/src/components/tile/_tile.scss
```diff
@@ -197,6 +198,10 @@
     }
   }
 
+  // .#{$prefix}--tile--selectable {
```
tw15egan: Just need to remove this comment and we should be good to go 👍🏻

tw15egan (approved):
LGTM 👍 ✅ 🎉

laurenmrice (approved):
Looks good ! 🙌

PR #8268, opened 2021-04-01
style(TextInput): add specificity to inline selectors

### LINKED ISSUE #8250: Inline labels sometimes show above form controls

- [x] `carbon-components`

## Detailed description

_form.scss defines the CSS of

```
  .#{$prefix}--form-item {
    ...
    flex-direction: column;
  }
```

_text-input.scss "overrides" that with:

```
  .#{$prefix}--text-input-wrapper--inline {
    flex-flow: row wrap;
  }
```

However, the override has the same selectivity as the original CSS, so it only works if _form.scss is loaded before _text-input.scss.   This works in the React code but isn't working for us, likely because we needed to `@include` _form.scss in one of our SCSS files in order to `@extend` bx--label.

Does it make sense to increase the selectivity of the selector in _text-input.scss?

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8250

There were scenarios wherein the inline variant of `TextInput` could render incorrectly if another component / file was importing `form.scss` _after_ `text-input.scss` due to the selectors having the same specificity. 

This PR just prefixes them with `.#{$prefix}--form-item` so that the form styles do not override the `inline` styles 

#### Changelog

**Changed**

- Increased specificity for 5 selectors to ensure the `inline` variant renders correctly no matter the import order. 

#### Testing / Reviewing

You can create a `styles.scss` inside the `TextInput` folder in the React repo, with the following line of code:

```scss
@import '~carbon-components/src/components/form/form';
```

and then in `TextInput-story.js`

```js
import `./styles.scss;`
```

Then, check the inline knob and ensure the label does not render on top of the input. It should stay on the left.


emyarod (approved):
should we select `.bx--text-input-wrapper` instead of `.bx--form-item`?

tw15egan:
@emyarod yeah that's a good point, will probably future-proof it a bit more in case we move form styles out of the components by default. Will update 👍🏻

jnm2377 (approved):

PR #8269, opened 2021-04-01
feat(tokens): new token update pt. 1

tw15egan:
Refs #8149 

Updates the following components to use the new tokens:
- [x] ContentSwitcher
- [x] ContextMenu
- [x] CopyButton
- [x] DataTable
- [x] DatePicker
- [x] Dropdown

#### Changelog

**Changed**

- New tokens replace old tokens in a few components

**Removed**

- Some extra styles that were causing padding differences in `DataTable`, see [here](https://github.com/carbon-design-system/carbon/pull/8269#discussion_r605711813)

#### Testing / Reviewing

Ensure the correct tokens are being used in the following components, and there are no visual changes in the storybook:

- [ ] ContentSwitcher
- [ ] ContextMenu
- [ ] CopyButton
- [ ] DataTable
- [ ] DatePicker
- [ ] Dropdown

### DIFF
--- a/packages/components/src/components/data-table/_data-table-action.scss
+++ b/packages/components/src/components/data-table/_data-table-action.scss
@@ -23,7 +23,7 @@
     width: 100%;
     height: $spacing-09;
     overflow: hidden;
-    background-color: $ui-01;
+    background-color: $layer;
   }
 
   .#{$prefix}--toolbar-content {
@@ -68,7 +68,7 @@
       background-color $duration--fast-02 motion(entrance, productive);
 
     &:hover {
-      background-color: $hover-field;
+      background-color: $field;
     }
   }
 
@@ -89,12 +89,12 @@
 
   .#{$prefix}--toolbar-search-container-expandable.#{$prefix}--search--disabled
     .#{$prefix}--search-magnifier-icon {
-    background-color: $disabled-01;
+    background-color: $layer-disabled;
     cursor: not-allowed;
     transition: background-color none;
   }
 
-  .#{$prefix}--toolbar-search-container-disabled {
+  .#{$prefix}--toolbar-search-container-disabled .#{$prefix}--search-input {
     cursor: not-allowed;
   }
 
@@ -111,7 +111,7 @@
     &::before {
       top: rem(2px);
       height: calc(100% - #{rem(4px)});
-      background-color: $hover-ui;
+      background-color: $field-hover;
     }
   }
 
@@ -148,7 +148,7 @@
 
   .#{$prefix}--toolbar-search-container-active
     .#{$prefix}--search-input:not(:placeholder-shown) {
-    background-color: $hover-field;
+    background-color: $field-hover;
     border: none;
   }
 
@@ -206,11 +206,11 @@
   }
 
   .#{$prefix}--toolbar-action:hover:not([disabled]) {
-    background-color: $hover-field;
+    background-color: $field-hover;
   }
 
   .#{$prefix}--toolbar-action:hover[aria-expanded='true'] {
-    background-color: $ui-01;
+    background-color: $layer;
   }
 
   .#{$prefix}--toolbar-action[disabled] {
@@ -219,7 +219,7 @@
 
   .#{$prefix}--toolbar-action[disabled] .#{$prefix}--toolbar-action__icon {
     cursor: not-allowed;
-    fill: $disabled;
+    fill: $icon-disabled;
   }
 
   .#{$prefix}--toolbar-action:focus:not([disabled]),
@@ -249,7 +249,7 @@
     width: auto;
     max-width: $spacing-05;
     height: $spacing-05;
-    fill: $icon-01;
+    fill: $icon-primary;
   }
 
   //-------------------------------------------------
@@ -290,14 +290,14 @@
 
   .#{$prefix}--toolbar-search-container-persistent
     .#{$prefix}--search-input:hover:not([disabled]) {
-    background-color: $hover-field;
+    background-color: $field-hover;
   }
 
   .#{$prefix}--toolbar-search-container-persistent
     .#{$prefix}--search-input:active:not([disabled]),
   .#{$prefix}--toolbar-search-container-persistent
     .#{$prefix}--search-input:not(:placeholder-shown) {
-    background-color: $hover-field;
+    background-color: $field-hover;
   }
 
   .#{$prefix}--toolbar-search-container-persistent .#{$prefix}--search-close {
@@ -327,7 +327,7 @@
     padding-right: $spacing-06;
     padding-left: $spacing-06;
     overflow-x: auto;
-    background-color: $interactive-01;
+    background-color: $background-brand;
     transform: translate3d(0, 48px, 0);
     transition: transform $duration--fast-02 motion(standard, productive),
       clip-path $duration--fast-02 motion(standard, productive),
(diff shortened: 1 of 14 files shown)


packages/components/src/components/data-table/_data-table-expandable.scss
```diff
@@ -64,36 +64,12 @@
...
   tr.#{$prefix}--parent-row.#{$prefix}--expandable-row + tr[data-child-row] td {
     padding-left: 4rem;
-    border-bottom: 1px solid $ui-03;
+    border-bottom: 1px solid $layer-alt;
     transition: padding-bottom $duration--fast-02 motion(standard, productive),
       transform $duration--fast-02 motion(standard, productive),
       background-color $duration--fast-02 motion(standard, productive);
   }
 
-  //compact child row padding
```
tw15egan, line 73: @jnm2377 I noticed these were added in https://github.com/carbon-design-system/carbon/pull/8080, but it seemed to cause misalignment issues in the size variants of the data table. When I remove these styles, they all align with the parent column (like the first image), but I was unsure if this was an intended visual update

Screen Shot 2021-04-01 at 10 38 46 AM
Screen Shot 2021-04-01 at 10 38 40 AM
Screen Shot 2021-04-01 at 10 38 34 AM
Screen Shot 2021-04-01 at 10 38 28 AM

packages/themes/src/g10.js
```diff
@@ -196,7 +196,7 @@ export const buttonSecondaryActive = activeSecondary;
 export const buttonTertiaryActive = activeTertiary;
 
 export const focusInset = inverse01;
-export const focuseInverse = inverseFocusUi;
+export const focusInverse = inverseFocusUi;
```
tw15egan: @jnm2377 just a heads up, I noticed we had an extra `e` in here so I've updated it in this PR 👍🏻

packages/components/src/components/data-table/_data-table-action.scss
```diff
@@ -68,7 +68,7 @@
       background-color $duration--fast-02 motion(entrance, productive);
 
     &:hover {
-      background-color: $hover-field;
+      background-color: $field;
```
aagonzales: Is this hover for items in the toolbar? I think it needs to be `field-hover` then

packages/components/src/components/data-table/_data-table-expandable.scss
```diff
@@ -64,36 +64,12 @@
 
   tr.#{$prefix}--parent-row.#{$prefix}--expandable-row + tr[data-child-row] td {
     padding-left: 4rem;
-    border-bottom: 1px solid $ui-03;
+    border-bottom: 1px solid $layer-alt;
```
aagonzales: If this is the visible divider between the row then it should use the border token `$border-subtle` instead.

aagonzales: Wondering if all these `border-bottoms` in the rows should be `$border-subtle` instead of `$layer-alt`

packages/components/src/components/data-table/_data-table-inline-edit-cell.scss
```diff
@@ -299,8 +299,8 @@
     padding: 1rem;
     color: $text-error;
     font-size: rem(12px);
-    background-color: $inverse-01;
-    border: 1px solid $ui-03;
+    background-color: $background-inverse;
```
aagonzales: `$background-inverse` pairs with `$inverse-02` not `-01`. So this should either be `$icon-inverse` or `$text-inverse`

packages/components/src/components/data-table/_data-table-inline-edit-cell.scss
```diff
@@ -315,8 +315,8 @@
     z-index: 10;
     width: rem(10px);
     height: rem(10px);
-    background-color: $inverse-01;
-    border: 1px solid $ui-03;
+    background-color: $background-inverse;
```
aagonzales: Same comment: `$background-inverse` pairs with `$inverse-02` not `-01`. So this should either be `$icon-inverse` or `$text-inverse`

aagonzales (commented):
Data table was the only one with some corrections and questions. 

- [x] ContentSwitcher
- [x]  ContextMenu
- [x]  CopyButton
- [x]  DataTable
- [x]  DatePicker
- [x]  Dropdown

tw15egan:
@aagonzales good catches! I've made the changes 👍🏻

aagonzales (approved):
🎉  Nice! Looks good to me now!

tay1orjones (approved):
Automated visual regression testing would've been nice on this one 😅 

I cross referenced storybooks and things look the same to me 👍

tw15egan:
@tay1orjones completely agree 😬

joshblack:
@tw15egan for adding in the tokens in this way, will it end up working out for how folks are currently theming since the tokens map to their old variants? or is this something that we will need to guard with a flag in case things don't match up?

tw15egan:
@joshblack theming seems to look correct in the deploy previews, did you have a specific use-case in mind? Just want to be sure I follow 👍

joshblack:
@tw15egan great question, I tried to make a SassMeister [here](https://www.sassmeister.com/gist/8fbe7482905c99387512d7899ae683bc) to demonstrate what I was thinking might happen.

Basically, if I have a theme that sets `ui-01` to something, but then the token that is used in a component changes (in the gist that is the change between `my-component-v1` and `my-component-v2`) then the way I apply my theme would stop working since it's not getting the value I set for `ui-01`.

I think the deploy previews would work as expected since the new token would be defined across themes through our default themes that have been updated, whereas here the custom theme would not have that update.

Definitely a rough/quick example, let me know if it'd be helpful to hop on a call instead 👀

packages/components/src/components/data-table/_data-table-expandable.scss
jnm2377: The changes were intentional, but I think I added these styles before I addressed this comment: https://github.com/carbon-design-system/carbon/pull/8080#pullrequestreview-612344005 and then forgot to recalculate the padding after the fix. Lemme mess around a bit and see if we still need them. Or figure out what the correct padding should be.

packages/react/src/components/DataTable/stories/expansion/DataTable-expansion-story.scss
```diff
@@ -1,10 +1,6 @@
 @import '~@carbon/type/scss/type';
 @import '~@carbon/themes/scss/themes';
 
-.demo-expanded-td td {
-  padding-left: 4rem;
-}
-
 .demo-inner-container-header {
```
jnm2377, line 4: Do we want to just add this to the expandable styles?

packages/components/src/components/data-table/_data-table-expandable.scss
jnm2377: @tw15egan you're right! These aren't needed after I addressed that comment lol AH! Thanks for that catch!!! Do we want to just add the same padding for all the sizes here? Rather than in the storybook? Or is the padding optional and we just add it in the storybook for prettiness? cc: @aagonzales

jnm2377 (approved):

packages/components/src/components/data-table/_data-table-expandable.scss
aagonzales: I think add the same padding for all the sizes.

tw15egan: This PR will normalize the left-padding between all size variants

packages/react/src/components/DataTable/stories/expansion/DataTable-expansion-story.scss
tw15egan: Should already be getting it from this selector 👍🏻 

`tr.bx--parent-row.bx--expandable-row+tr[data-child-row] td`

Screen Shot 2021-04-12 at 9 12 50 AM

PR #8276, opened 2021-04-01
feat(react): add FeatureFlags component for contextual feature flags

joshblack:
A use-case came up with some of the grid work where we wanted a story where the feature flags for grid or v11 could be turned on and components used within the story would act as expected.

This PR satisfies this use-case by introducing a `FeatureFlags` component and two hooks: `useFeatureFlag` and `useFeatureFlags`. The `FeatureFlags` component is ultimately a context provider and allows the end-user to specify what flags should be enabled or disabled in a given React tree. The `useFeatureFlag` and `useFeatureFlags` hooks are intended to be used by components to determine if a flag is enabled or disabled.

This work includes a refactor to `@carbon/feature-flags` to encapsulate the existing behavior into a new module: `FeatureFlagScope` that is used to create the default "global scope" of feature flags. Consumers can also create their own scopes with a `createScope` function.

#### Changelog

**New**

- `FeatureFlags` component
- `useFeatureFlag` and `useFeatureFlags` hook
- New `FeatureFlagScope` module in `@carbon/feature-flags`

**Changed**

- Refactor `@carbon/feature-flags` to use new `FeatureFlagScope` module

**Removed**


tay1orjones (approved):
🔥 🎉  I really like that there are the two separate hooks - one for individual flag usage and another for the entire scope

packages/feature-flags/src/FeatureFlagScope.js
```diff
@@ -0,0 +1,93 @@
...
+      this.flags.set(key, value);
+    }
+  }
+
+  /**
+   * Check if a feature flag is enabled
+   * @param {string} name
+   * @returns {boolean}
+   */
+  enabled(name) {
```
emyarod, line 89: should we rename this to `isEnabled` to make it more distinguishable from `enable`?

emyarod (approved):

packages/feature-flags/src/FeatureFlagScope.js
joshblack: I can definitely go either way, whatever seems more intuitive for folks 👍  This is also internal for us so we can try out both to see what works best if we want.

PR #8277, opened 2021-04-01
feat(project): update yarn to 3.0.0-rc.1 and add @carbon/react

joshblack:
This PR for updating yarn to source (meaning that Yarn is built from the latest code on GitHub). This approach came up due to a known bug over in Yarn that occurs with `postinstall` scripts on package that have `peerDependencies`.

Thankfully, this issue is fixed. However, there is no release that currently contains this fix outside of using source from Yarn. As a result, we had several options for adding in the `@carbon/react` package:

1. Update to Yarn source which addresses the bug
2. Remove the `postinstall` step for the `carbon-components-react` package
3. Share files between `carbon-components-react` and `@carbon/react` so that there is no dependency

This PR attempts to use approach 1 to solve these problems/limitiations. So far, it seems like it builds successfully with `@carbon/react` and many of the file changes are due to hash changes. However, I wanted to get feedback from everyone to see how they felt and if there were any other strategies I should try out 👀 

Note, for changes to the react package itself here is a link to view them separate from the yarn changes: [Link](https://github.com/carbon-design-system/carbon/pull/8277/commits/daaf20ebedf54f54669c488ddffc3c309ab8b0eb?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.scss&file-filters%5B%5D=dotfile)

#### Changelog

**New**

- Add `packages/carbon-react` for the `@carbon/react` package

**Changed**

- Update `.yarn/*` to build from source

**Removed**

#### Testing / Reviewing

- Verify CI builds work as expected
- Verify no unintended package changes have occurred


packages/carbon-react/src/index.js
```diff
@@ -0,0 +1,201 @@
...
+  unstable_ContextMenu,
+  unstable_ContextMenuDivider,
+  unstable_ContextMenuGroup,
+  unstable_ContextMenuItem,
+  unstable_ContextMenuRadioGroup,
+  unstable_ContextMenuSelectableItem,
+  unstable_useContextMenu,
+  unstable_Heading,
+  unstable_Section,
+} from 'carbon-components-react';
```
mattrosno: For option 3, is this really just using a relative path in these imports? Possible because we have a monorepo? That seams the cleanest until yarn releases its fix?

joshblack: @mattrosno basically we could get away with this by using a relative path and then Rollup could "inline it" in the resulting bundle, I think. Would have to prototype it to find out

emyarod (commented):
is there a link to see more about the yarn bug?

joshblack:
@emyarod really just a thread in their discord but I have this repo to reproduce: https://github.com/joshblack/yarn-berry-postinstall-bug-repro

Basically when you have a packages with peer dependencies and a post install step it fails (seems like yarn treats it like a cycle so the bin is undefined)

emyarod:
if there is an associated a ticket on the yarn repo I just want to see more about the release strategy but given that it's fixed on yarn latest wouldn't it be coming in a future release?

joshblack:
@emyarod no ticket that I could find when I was looking 😞  I was in the process of making a ticket using their tooling but noticed I couldn't reproduce the issue. That lead me to their discord where some of the folks there were helping me out when they realized that it was already fixed on latest in a PR.

When it comes to release, it seems like `2.4.x` will be the last minor in the 2.x release line 😞  Reference: https://dev.to/arcanis/yarn-2-4-log-filters-audits-better-warnings-49io#what-about-yarn-3

This may be why the release has taken a while as they are moving to v3 now.

emyarod:
ideally they would release a patch to resolve the current issues in their latest version. would this be a temporary migration to building from source or would we transition back to versioned releases?

joshblack:
@emyarod it'd be temporary until the next stable release that includes this fix 👍

tay1orjones (approved):
I'm good with temporarily building yarn from source to unblock the effort of getting an all-in-one `@carbon/react` package going 👍

emyarod (approved):
ok just wasn't sure if it was a permanent change especially with only a little information on the underlying yarn issue, looks good to me

joshblack:
@tay1orjones @emyarod seems like there is a 3.0 RC we can try out as well, might be better than building from source so let me try that out too 👀

PR #8282, opened 2021-04-02
Experiment/add cypress for e2e tests

### LINKED ISSUE #8145: Investigate Cypress for visual testing in our React package
Cypress is a tool that has been recommended to us as one that we could use to do visual testing on our components. This issue is for delivering an MVP integration with cypress with ideally one test suite (probably Accordion?) that demonstrates how we could bring this into the project.

This integration should consider the following:

- How simple is it to author tests?
- How simple is it for tests to run in CI?
  - Does running cypress dramatically increase CI time?
- How simple is it to bring in styles for our components?

tay1orjones:
Closes #8145 

- Add and configure Cypress for automated end to end testing
- Add the Cypress eslint plugin
- Add `@testing-library/cypress` to have a more consistent syntax between jest unit tests and cypress tests.
- Include `cypress-real-events` so we can use the <kbd>Tab</kbd> key for keyboard navigation testing.
- Add and configure Percy snapshots in CI only.
- Add simple e2e test file for Accordion for proof of concept

From the original issue - this integration considers the following:

> How simple is it to author tests?

- I think it's pretty easy, it's very similar syntax to jest tests. The assertion syntax is a bit different though.

> How simple is it for tests to run in CI?

- Fairly simple. `ci-checks` runs a `yarn lerna run` command in the root `package.json`, which runs yarn command for cypress in `packages/react`. We only attach Percy to report snaps to their external service in CI (not locally)

> Does running cypress dramatically increase CI time?

- Locally this single test takes 52s to complete

> How simple is it to bring in styles for our components?

- Pretty simple, just import them at the top of the test file. The webpack config file is used by cypress to bring it all in.

#### Changelog

**New**

- add cypress automated end-to-end testing

#### Testing / Reviewing

- CI should pass, snaps should be reported to the Percy dashboard. 
- Cypress command should succeed locally


packages/react/src/components/Accordion/__tests__/Accordion-test.e2e.js
```diff
@@ -0,0 +1,70 @@
...
+    cy.findByText(/Panel 3/).should('not.be.visible');
+
+    // snapshots should always be taken _after_ an assertion that
+    // a element/component should be visible. This is to ensure
+    // the DOM has settled and the element has fully loaded.
+    cy.percySnapshot();
+  });
+
+  it('should tab between items and open on space', () => {
+    // in order for realPress('Tab') to work, something must be focused first.
```
joshblack, line 41: Does a leading `cy.realPress('Tab')` work here at all or nah?

joshblack:
Getting this right now when running the tests: 
image

Is it missing our `babel-plugin-dev-expression`, I guess?

dakahn (approved):
installed, built and tests ran great 🟢 👍🏽

packages/react/src/components/Accordion/__tests__/Accordion-test.e2e.js
tay1orjones: No, unfortunately

joshblack:
@tay1orjones how does the authoring experience work with the update? I had assumed it was `yarn cypress open` but wanted to check. When I use that command I end up seeing this in the GUI: 

image

> Locally this single test takes 52s to complete

Also ooof, that's rough. I assume most of the time is to compile the dev server from webpack that is used by cypress to test?

tay1orjones:
@joshblack This uses the "[Component Test Runner](https://www.cypress.io/blog/2021/04/06/introducing-the-cypress-component-test-runner/)" which they've separated from the standard runner. It's what allows us to `mount()` a component vs having to spin up a dev server and `cy.visit(url)`. So the commands are different: 
- `yarn cypress open-ct`
- `yarn cypress run-ct`

> I assume most of the time is to compile our code with the dev server from webpack that is used by cypress to test?

I think it's actually been reduced since I updated to the new version of cypress! I don't see a total time output on the console, but the compile time is listed at 22.24s. The tests themselves are super quick:

```
Accordion
    ✓ should render (151ms)
    ✓ should tab between items and open on space (477ms)
```

Also there is video recording processing, 2s of the overall time. We might be able to turn that off in CI via a flag. Same with screenshots.

tay1orjones:
Just pushed an update to turn off video and snaps in CI on the `yarn test:e2e:cypress` command.

I spoke too soon on the time reduction, lerna still reports total time of 44.2s to complete the command.

joshblack:
Thanks @tay1orjones! That makes a ton of sense.

package.json
```diff
@@ -27,7 +27,8 @@
     "lint:styles": "stylelint '**/*.{css,scss}' --report-needless-disables --report-invalid-scope-disables",
     "sync": "carbon-cli sync",
     "test": "cross-env BABEL_ENV=test jest",
-    "test:e2e": "cross-env BABEL_ENV=test jest --testPathPattern=e2e --testPathIgnorePatterns='examples,/packages/components/,/packages/react/'"
+    "test:e2e:jest": "cross-env BABEL_ENV=test jest --testPathPattern='e2e/' --testPathIgnorePatterns='examples,/packages/components/,/packages/react/'",
+    "test:e2e:cypress": "lerna run test:e2e:cypress"
```
joshblack: How would you feel about using the command:

```bash
yarn workspace carbon-components-react test:e2e:cypress
```

Instead of adding a new npm task? Just was curious, sometimes these get crazy with the multiple colon joined tasks that I was curious if there was a way to simplify or not.

packages/react/package.json
```diff
@@ -35,7 +35,8 @@
     "prepublish": "yarn build",
     "start": "yarn storybook",
     "storybook": "rimraf node_modules/.cache/storybook && start-storybook -p 9000 -s ./.storybook/assets",
-    "snapshot": "build-storybook && percy-storybook --widths=320,1280"
+    "snapshot": "build-storybook && percy-storybook --widths=320,1280",
+    "test:e2e:cypress": "cypress run-ct --config video=false,screenshotOnRunFailure=false"
```
joshblack: Feel free to use `test:e2e` here, I think

package.json
tay1orjones: @joshblack I'm not opposed to it. My thought was that we may want to leave the door open to cypress testing in other packages, which the lerna command would automatically pick up any other packages that have the `test:e2e:cypress` command.

tay1orjones:
@joshblack I updated the command naming/syntax based on your suggestions 👍

tw15egan:
This is really great! I changed the test to output the video and everything worked as expected

tay1orjones:
@tw15egan Awesome! You can also run the tests with GUI for debugging/authoring with `cypress open-ct`

tay1orjones:
@tw15egan @joshblack let me know if there's any other concerns here 👀  

Otherwise I think this one might be set 👍  just need another approve

tw15egan (approved):
Ship it! 👍🏻 ✅

PR #8283, opened 2021-04-02
fix(RadioButtonGroup): clones RadioButton

### LINKED ISSUE #8225: `RadioButtonGroup` usability issue (renders it's own children regardless of given children types)
## Detailed description

`carbon-components-react@npm:7.31.0`

I've created a wrapper component for `RadioButton` to support our design addition of description text & then spent half an hour scratching my head as to why my component is not rendered.

It's because [you are rendering your version of `RadioButton` regardless of what is given as children](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/RadioButtonGroup/RadioButtonGroup.js#L90).

I'd consider this to be quite an ugly anti-pattern.
If you need to insist on using the only right component, why aren't you using data to render  (an array of records as a prop for `RadioButtonGroup`) instead of remapping children.

## Additional information

image

andreancardona:
Closes #8225 

- We decided to clone the child (radioButton)  instead of forcing our own RadioButton (what we are currently doing) causing issues for users

#### Testing / Reviewing

- Make sure RadioButton works as expected (there should be no changes): https://deploy-preview-8283--carbon-components-react.netlify.app/?path=/story/components-radiobutton--default


packages/react/src/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -99,16 +98,13 @@ export default class RadioButtonGroup extends React.Component {
...
-          checked={value === this.state.selected}
-        />
-      );
+      return React.cloneElement(radioButton, {
+        name: this.props.name,
+        key: value,
+        value: value,
+        onChange: this.handleChange,
+        checked: value === this.state.selected,
+      });
```
tw15egan, line 107: Since all the children may not be `RadioButton` elements, should we check if they are before adding irrelevant data to these elements?

```jsx
      if (radioButton.type.displayName === 'RadioButton') {
        return React.cloneElement(radioButton, {
          name: this.props.name,
          key: value,
          value: value,
          onChange: this.handleChange,
          checked: value === this.state.selected,
        });
      } else {
        return radioButton;
      }
```

Otherwise, I'm seeing the `name` attribute being placed on a custom `div` I rendered adjacent to the first RadioButton

Screen Shot 2021-04-05 at 4 33 24 PM

andreancardona: These are props we want to override.

Thoughts @vpicone 👀

tw15egan: In my example, I had

```jsx
<RadioButtonGroup
    legendText="Radio button heading"
    name="radio-button-group"
    defaultSelected="radio-1">
      <RadioButton labelText="Option 1" value="radio-1" id="radio-1" />
      <div>A custom sublabel</div>
      <RadioButton labelText="Option 2" value="radio-2" id="radio-2" />
      <RadioButton labelText="Option 3" value="radio-3" id="radio-3" />
</RadioButtonGroup>
```

and the `div` was getting the `name` attribute. I think we would only want to apply these props to `RadioButton` elements? 

I could be misunderstanding the original issue! Do we have a different sample code block to test?

vpicone: @tw15egan the aim here is to let folks customize/wrap `RadioButton` with their own custom component. Doing a check on the `displayName` prevents that in the same way that replacing their component with our own does (current state). I don't think we can't solve the arbitrary `div` wrapper scenario without context, but we can pass their custom component props as if it were a `RadioButton` so that they can be applied in their component.

```jsx
const MySpecialRadioButton = (props) => (
  <div className="my-custom-class">
    <RadioButton {...props} />
    <p>My special description</p>
  </div>
);

const App = () => (
  <RadioButtonGroup name="radio-button-group">
    <MySpecialRadioButton
      id="radio-1"
      value="standard"
    />
    <RadioButton id="radio-2" value="standard" />
  </RadioButtonGroup>
);
```

tw15egan: Makes sense! Thanks for clarifying it for me 😄

tw15egan (approved):
LGTM 👍 ✅

joshblack (approved):

PR #8212, opened 2021-03-29
fix(data-table): remove inline edit short table styles

emyarod:
Closes #8206

This PR removes broken style overrides for form controls within short data tables

#### Testing / Reviewing

Confirm that inline selects within the short data table have the correct spacing and alignment for the chevron icon. The extra table column should be removed from the examples before merging

PR #8295, opened 2021-04-05
fix(data-table): add table scope to form control overrides

### LINKED ISSUE #8286: Background color for TextInput, PasswordInput, MultiSelect, ComboBox etc.. is the same as the light variant

- [x] `carbon-components`

## Detailed description

It seems that components with input elements have the "light" background color as the default. I've checked the following components and assume they apply to all/any input components:

- TextInput
- PasswordInput
- MultiSelect
- ComboBox
- ...maybe more

The issue may be because the [styles defined here](https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/components/data-table/_data-table-core.scss#L145) are not scoped to the data table.

**Repro steps**

- visit the Storybook
- the default background color for components with inputs is the same as their light variants
- try toggling the "light" knob (nothing should happen)

[TextInput example](https://react.carbondesignsystem.com/?path=/story/components-textinput--default):

Screen Shot 2021-04-02 at 1 54 44 PM

emyarod:
Closes #8286
Related #8212

This PR scopes the table form control styles to the data table component to prevent standalone form controls from inheriting the styles

#### Testing / Reviewing

Confirm that inline table form controls receive the correct color token while standalone form controls are unaffected

### DIFF
--- a/packages/components/src/components/data-table/_data-table-core.scss
+++ b/packages/components/src/components/data-table/_data-table-core.scss
@@ -142,15 +142,16 @@
   }
 
   // Form Control Overrides
-  .#{$prefix}--list-box input[role='combobox'],
-  .#{$prefix}--list-box input[type='text'],
-  .#{$prefix}--dropdown,
-  .#{$prefix}--list-box,
-  .#{$prefix}--number input[type='number'],
-  .#{$prefix}--number__control-btn::before,
-  .#{$prefix}--number__control-btn::after,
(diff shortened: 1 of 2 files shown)


tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

joshblack (approved):

laurenmrice (approved):
👍 looks good

PR #6925, opened 2020-09-28
fix(a11y): fixing MultiSelect no focus indicator

rjhenriquez:
Closes #6855

Fixing MultiSelect - accessibility issue - no focus indicator visible on clear button

#### Changelog

**Changed**

MultiSelect component

#### Testing / Reviewing

- Go to http://localhost:9000/?path=/story/multiselect--default
- Select a few elements
- Focus on the button to "Clear Selection" by keyboard, it should show an outline

PR #4792, opened 2019-11-28
fix(Tag): ignore click events when disabled

emyarod:
Closes #4791

~~This PR wraps the close icon in a button~~ to utilize native button features (rather than spoofing a button with the SVG element) and ignores the onClick handler when the filter tag is disabled

#### Changelog

**New**

- extend disabled tag styles for filter tag

**Changed**

- ~~wrap SVG icon in button~~ due to changes introduced in #4863 the entire tag is now a `button` and the styles are rescoped to the svg icon within the `button.bx--tag` element
- click handler logic
- update test to remove console warning

#### Testing / Reviewing

Ensure click handlers are no longer fired when the tag filter is disabled and the component matches the visual spec (particularly the disabled state)

PR #8297, opened 2021-04-05
fix(Tag): pass disabled state to tag button

### LINKED ISSUE #7160: Investigate disabled tags firing onClick
Noticed when reviewing #6925 

https://github.com/carbon-design-system/carbon/pull/6925#issuecomment-700315384

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/7160
Refs https://github.com/carbon-design-system/carbon/pull/4792
Refs #6925

Prevents `onClick` from firing on an `interactive` tag when it is disabled. 

#### Changelog

**New**

- If the `Tag` is an interactive tag (`button`), passes along the disabled state to prevent click actions

#### Testing / Reviewing

Set the `disabled` knob to true for the default `Tag` story. Ensure no actions are logged when clicking on the disabled tags


tw15egan:
We could also modify the `filter` variant to handle this scenario as well, but we would need to make the entire `tag` focusable (need design input)

emyarod (approved):
looks good to me

andreancardona (approved):
looks good!

PR #8304, opened 2021-04-05
fix(date-picker): reduce gap between range inputs

### LINKED ISSUE #8234: Time picker vs Date range picker - spacing
We are exploring a custom field type which is a compound of a Text input + Select components. In Carbon there are two similarly constructed components - "Time picker" and "Date range picker".

I wanted to double check what should be the spacing between the two components - 1px or 2px? I noticed the spacing is not consistent in the following Sketch library assets:

Time picker vs Date range picker - spacing@2x

emyarod:
Closes #8234

This PR reduces the gap between the date picker range inputs to match the spec

#### Changelog

**Changed**

- reduce gap between the date picker range inputs
- enable linting for flatpickr styles
- resolve stylelint violations on flatpickr styles

#### Testing / Reviewing

Confirm that the date picker styles are identical to the existing component except for the gap between the date picker range inputs


joshblack (approved):

tw15egan (approved):

joshblack:
bump @laurenmrice when you have a sec!

laurenmrice (approved):
looks good ! 👍

PR #8307, opened 2021-04-05
docs(style): add initial testing style guide

### LINKED ISSUE #8144: Write initial testing style guide
As we start to refactor our test suite in #8143, we should begin to author a living testing style guide that lives in `/docs/style.md`. This testing style guide should cover the following topics and should be added to over time:

- Test structure
  - Where should user-driven (e2e) tests be written
  - Where should automated accessibility tests be written
  - Where should Component API tests be written
- Test recipes for common practices

tay1orjones:
Closes #8144

This adds some initial component testing style documentation. It's not final by any means, just a stake in the ground to center ideas and conversation. Ideally it'll be a living document, updated as we learn new strategies and more throughly inform what/how/why we find to be an effective testing strategy. 

Some of this is still in flux but good to begin to nail down, like e2e testing explorations in #8282 

#### Changelog

**Changed**

- add testing section to `docs/style.md`


docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+    });
+
+    // id
+    // -----
+    // When a component accepts an id prop, it's important
+    // that the node on which the id is placed is consistent
+    // between minor versions. As a result, tests that you
+    // write for id should make assertions around id being
+    // placed on the same node.
+    id(
```
joshblack:
```suggestion
    it(
```

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+##### `ComponentName-test.a11y.js`
+
+- Use `accessibility-checker` and `axe`
+- Optionally configure common props to ensure component variants do not contain
+  accessibility errors.
+
+```js
+describe('ComponentName - AVT1', () => {
+  it('should have no aXe violations', async () => {
+    render(<ComponentName />);
```
joshblack: Should we use the container from this to then pass to the assertion below?

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+});
+```
+
+##### `ComponentName-test.e2e.js`
+
+- These tests are ran with Cypress, which is a bit different in syntax and
+  mental model than Jest/JSDOM tests. It's recommended to become familiar with
+  the core concepts by reading the
+  [Cypress introduction documentation](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress).
+- Trigger Events with ``via`cy.real\*();`
```
joshblack:
```suggestion
- Trigger Events with via `cy.real\*();`
```

joshblack: Related: I think cypress has some testing-library style helpers we could use, right? https://testing-library.com/docs/cypress-testing-library/intro

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+
+import React from 'react';
+import { mount } from '@cypress/react';
+
+describe('ComponentName - e2e', () => {
+  beforeEach(() => {
+    mount(<ComponentName />);
+  });
+
+  it('should render', () => {
```
joshblack: I wonder if we should avoid the "should render" style tests and instead do usage-based examples at the top level. Should render style tests seem like they should be covered in the usage/behavior tests so should be fine.

I guess the only question from that is if we couple the visual snapshot tests with the behavior or do we separate them?

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+
+```js
+/**
+ * @jest-environment node
+ */
+
+import { ComponentName } from '../ComponentName';
+
+describe('ComponentName - SSR', () => {
+  it('should import ComponentName in node/server env', () => {
```
joshblack: We might need this to be something like:

```js
  expect(() => {
   require('path/to/index.js');
  }).not.toThrow();
```

We could also use Jest's module name mapper option to have `carbon-components-react` resolve to the `src/index.js` or CommonJS entrypoint?

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+    expect(ComponentName).toBeTruthy();
+  });
+});
```
joshblack, line 713: We should also add a test for `renderToStatic` for SSR, I think, to catch for any unchecked references to `window` or `document`

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+```
+
+##### Notes on manual testing
+
+- Due to
```
joshblack, line 721: @dakahn is there a good list or document we could point to for manual testing?

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+
+- These tests are ran with Cypress, which is a bit different in syntax and
+  mental model than Jest/JSDOM tests. It's recommended to become familiar with
+  the core concepts by reading the
+  [Cypress introduction documentation](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress).
+- Trigger Events with ``via`cy.real\*();`
+- Keyboard navigation/interaction testing is best done here, in a simulated
+  browser environment, not JSDOM.
+- Take snapshots with Percy
+  - Percy snapshots should be used very sparingly (1-2 per component)
```
joshblack: We should be fine in terms of snapshots with Percy, right? The limit should be pretty high, would be great to get a lot of visual coverage but I totally get that there is a sweet spot to balance here

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+```
+
+##### `ComponentName-test.a11y.js`
+
+- Use `accessibility-checker` and `axe`
+- Optionally configure common props to ensure component variants do not contain
+  accessibility errors.
+
+```js
+describe('ComponentName - AVT1', () => {
```
joshblack: Style nit:

```suggestion
describe('ComponentName AVT1', () => {
```

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+| `ComponentName-test.ssr.js`  | Server side rendering           |
+
+There are corresponding commands to run all categories, individual categories,
+or a combination. Depending on your shell, modifiers can be used to run two
+commands one after another. Refer to the documentation of your shell.
+
+| Command                           | Corresponding test category                                          |
+| --------------------------------- | -------------------------------------------------------------------- |
+| `yarn test`                       | All categories                                                       |
+| `yarn test:unit`                  | Only component unit tests                                            |
```
joshblack, line 580: What do these command look like in the `package.json`? 👀

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+  JSDOM. Testing this in a simulated browser environment, such as Cypress (e2e),
+  is a better environment for this type of testing.
+
+```js
+import { render, screen, findByLabel } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { ComponentName } from '../ComponentName';
+
+describe('ComponentName', () => {
+  it('should render with default props', () => {
```
joshblack: I'm not sure how useful these kinds of tests are, do you think we need them with some of the other tests that we're writing?

docs/style.md
```diff
@@ -539,6 +539,245 @@ import { unstable_Pagination as Pagination } from 'carbon-components-react';
...
+| `yarn test:ssr`                   | Only server side tests                                               |
+| `yarn test:a11y && yarn test:e2e` | In `bash` via `&&`: Run the a11y tests, and if they succeed, run e2e |
+
+#### Recipes
+
+Below are some common receipes for component testing. Many of the pattern/syntax
+details contained within these recipes are enforced via eslint rules declared in
+`eslint-config-carbon`.
+
+##### `ComponentName-test.js`
```
joshblack, line 591: I wonder if we should add a preamble here around what this file is for. I think if we have Cypress then this file decidedly becomes testing the "API" of the component, right?

joshblack (changes requested):
Looking great! Left some comments/questions while reading through it 👀

docs/style.md
tay1orjones: Something like this? Was thinking we could add in during the beginnings of the refactor

```json
scripts: {
  "yarn test": "yarn test:unit && yarn test:a11y && ... maybe we look into running these in parallel rather than sequentially or something ",
  "yarn test:unit": "jest --testPathPattern='.*\\-test\\.js(x)?'",
  "yarn test:a11y": "jest --testPathPattern='.*\\.a11y\\.js(x)?'",
  "yarn test:e2e": "jest --testPathPattern='.*\\.e2e\\.js(x)?'",
  "yarn test:ssr": "jest --testPathPattern='.*\\.ssr\\.js(x)?'",
}
```

docs/style.md
tay1orjones: 👍  I'll take a stab at it. Based on how slow/expensive cypress is I think it makes sense to encapsulate as much as possible here with only a specific limited set of items in the e2e tests. I'm thinking of the testing pyramid and all - ideally this file would cover the majority of component functionality.

That said, I think it's hard to codify right now without having yet dug into refactoring our existing tests off enzyme.

docs/style.md
tay1orjones: Probably not. Aim is to get some sort of absolute baseline for components, like what is the ground level? I think all components should have a className placed on the root node, maybe that's where all component files should start?

joshblack: @tay1orjones nice, I like it. Especially for this file that seems to be testing the "component API" if we're using Cypress for behavior

docs/style.md
joshblack: @tay1orjones that sucks about Cypress being so slow 😞  Do you think it'd be worth doing a quick hackathon to see how far we could take a custom playwright setup that works with Jest?

docs/style.md
tay1orjones: Sort of. It's dependent on how we config and what we want to cover. IIRC we'll have a quota of 200k snaps a month? My thought was that it would run on every PR.

- ~55 components @ 2 snaps each
  - 110 snaps per run
- Ideally we'd check chrome/safari/firefox
  - 110 snaps * 3 browsers = 330 snaps per run
- If we run this test suite in PR checks it would be ran on every update w/ latest from `main` (average 2 of those? 3?)
  - 330 snaps * 3 updates on every PR = 990 snaps per PR
- Average past two months PR count is ~100 pull requests per month
  - 990 snaps per PR * 100 PRs a month = 99,000

Lots of variables there too. Add one more snap per component and it's 150k, 4 snaps per component and we hit quota.

An alternative option would be to not run VRT on every PR, but instead only once a week through a GH action and have it automatically open defect issue(s) for us to address before we release.

TLDR - We'll probably have to choose between more snaps ran less frequently, or less snaps ran more frequently

docs/style.md
tay1orjones: Yeah, this is a good usecase for container. I'll update to reflect.

Reminds me that `accessibility-checker` wants things wrapped in a main tag. I'm not sure if axe does or not. Testing library defaults to rendering inside a div inside the body,so I think we'll need to either wrap the component in `<main>` or use [the `container` option](https://testing-library.com/docs/react-testing-library/api/#container) inside `render()`

joshblack: @tay1orjones we might be able to pass/configure AC like we did for aXe to ignore landmarks so we don't have to worry about that 👀  Not sure if they changed it between versions but worth checking out if so since we wouldn't have to worry about it then.

docs/style.md
dakahn: I really really really like this one https://www.a11yproject.com/checklist/

dakahn (approved):
Gonna approve pending @joshblack 's comments getting resolved. 🏄🏽 happy to have this stuff codified. Should make testing less interpretive and more rote moving forward

docs/style.md
tay1orjones: I think we have to "ignore failures" with a baseline file, rather "than turn off rules." Ruleset config is in progress though, [internal slack convo](https://ibm-studios.slack.com/archives/C036P1CTN/p1617919996193400?thread_ts=1617915215.191700&cid=C036P1CTN) 

> On the automated side, you can use a baseline file to ignore known issues. Depending on the browser/test environment you’re running, you may be able to run the scan on a node instead of on the page, which would only return issues within that subtree.
> ... 
> ...on the ruleset customization, that PR hadn’t been finished yet: https://github.com/IBMa/equal-access/pull/270

joshblack: > you may be able to run the scan on a node instead of on the page, which would only return issues within that subtree.

ooo this seems promising? 👀

docs/style.md
tw15egan: When we last had Percy enabled, it only tested Firefox and Chrome, so that would reduce it a little bit

joshblack:
@tay1orjones since this is a pretty large set of guidance/docs, would it help if we separated out topics into a couple of categories:

- No brainers: these are stuff we're 100% confident in and should be merged in this PR
- Needs some tweaks or examples: these are topics that we're pretty sure about but need a couple of examples to reinforce the point or make sure it's the right direction
- Aspirational: stuff we want to try but don't have the time/bandwidth to fully verify or offer guidance for

The goal being that no brainers get put in this PR, and then we can have issues for each of the separate topics to help make breaking it down easier and help target what needs to be experimented/worked on in a sprint?

docs/style.md
tay1orjones: After some testing, I found that the most consistent way to avoid having to potentially wrap things in `<main>` is to simply always pass the `container` to both Axe and AC 👍

docs/style.md
tay1orjones: @joshblack Yep, if you'd prefer we can have one file that imports/covers the entire package. Though do you think it would be helpful to stub out these simple SSR test files for each component so we'd have them for future expansion of server side testing? Otherwise it would be a one-off file that not all components would have in their folder.

docs/style.md
tay1orjones: @joshblack not opposed! I think that could be a great idea.

tay1orjones:
I've pushed an update incorporating feedback and also pulled out some items to be defined later:

- e2e testing guidance/docs through [#8145](https://github.com/carbon-design-system/carbon/issues/8145#issuecomment-805945758)
- keyboard navigation strategy, https://github.com/carbon-design-system/carbon/issues/8464

joshblack (approved):

andreancardona (approved):
🎉

PR #8308, opened 2021-04-05
fix(ComboBox): add missing dependencies to useEffect

### LINKED ISSUE #8306: [ComboBox]: onInputChange is called on every render, even if input value did not change, causing infinite render loop

- [ ] `carbon-components`

## Detailed description

My application uses the `ComboBox` component, and updates the `items` array based on the input given by the `onInputChange` callback prop. My application also fully controls rendering via calls to `render` from `react-dom`. When we render based on an input change, we enter an infinite loop.

Yes, ComboBox, and potentially others changed in a similar way.

> see changed?

This use of `setEffect` does not properly follow the Rules of Hooks. Since `useEffect` will get called on every render, calling the `onInputChange` callback sets up an infinite loop if the callback itself triggers another render. This pattern is common for fully-controlled applications that need to re-render to update the items that are available to the combo box after user input.

https://github.com/carbon-design-system/carbon/blob/a067cfcc112b309ae36a47f569d24f1942a1cca4/packages/react/src/components/ComboBox/ComboBox.js#L149-L153

Instead, the input value should be added into the [second argument of `useEffect`](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects) to prevent extraneous calls.

It may be a good idea to run [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) as part of the Carbon build to catch problems like this in the future.

- macOS Big Sur x86
- Chrome Version 89.0.4389.114 (Official Build) (x86_64)
- Firefox Developer Edition 88.0b7 (64-bit)

Problem introduced in 10.24.0, not present in 10.23.0.

Introduced by https://github.com/carbon-design-system/carbon/commit/635e0cbb37542f8ecc574b449e4250bd2442116b as part of https://github.com/carbon-design-system/carbon/pull/7768.

1. Call `render` in an `onInputChange` callback passed to a `ComboBox` React component.
2. Change the input value.
3. (infinite loop)

See https://codesandbox.io/s/interesting-feather-1s5kk?file=/src/index.js for a reproduction.

Note that I added a `setTimeout` to the render call in my example to prevent the infinite loop from blocking the thread, which would break CodeSandbox.

emyarod:
Closes #8306

This PR adds missing dependencies to the Combobox `useEffect` to avoid a render loop when the component is controlled and rerendered on input changes

#### Testing / Reviewing

Substitute the Combobox component in the example code sandbox with a built version of the component from this PR, and observe that removing the setTimeout in the example app no longer causes the component to render infinitely

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.js
+++ b/packages/react/src/components/ComboBox/ComboBox.js
@@ -150,7 +150,7 @@ const ComboBox = (props) => {
     if (onInputChange) {
       onInputChange(inputValue);
     }
-  });
+  }, [onInputChange, inputValue]);
 
   const handleSelectionClear = () => {
     if (textInput?.current) {
--- a/packages/react/src/components/ComboBox/ComboBox.js
+++ b/packages/react/src/components/ComboBox/ComboBox.js
@@ -113,6 +113,7 @@ const ComboBox = (props) => {
   );
   const [prevSelectedItem, setPrevSelectedItem] = useState(null);
   const [doneInitialSelectedItem, setDoneInitialSelectedItem] = useState(null);
+  const savedOnInputChange = useRef(onInputChange);
 
   if (!doneInitialSelectedItem || prevSelectedItem !== selectedItem) {
     setDoneInitialSelectedItem(true);
@@ -147,10 +148,14 @@ const ComboBox = (props) => {
   };
 
   useEffect(() => {
-    if (onInputChange) {
-      onInputChange(inputValue);
+    savedOnInputChange.current = onInputChange;
+  }, [onInputChange]);
+
+  useEffect(() => {
+    if (savedOnInputChange.current) {
+      savedOnInputChange.current(inputValue);
     }
-  }, [onInputChange, inputValue]);
+  }, [inputValue]);
 
   const handleSelectionClear = () => {
     if (textInput?.current) {


packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -150,7 +150,7 @@ const ComboBox = (props) => {
     if (onInputChange) {
       onInputChange(inputValue);
     }
-  });
+  }, [onInputChange, inputValue]);
```
joshblack: Would it be helpful if we added a `useSavedCallback` type hook for these situations? It seems like the effect is parameterized by `inputValue` and we wouldn't want to re-run this effect when `onInputChange` happens (especially if we're passed a function that's updated each render)

`useSavedCallback` could look like:

```jsx
const savedOnInputChange = useSavedCallback(onInputChange);

// ...

useEffect(() => {
  const { current: onInputChange } = savedOnInputChange;
  if (onInputChange) {
    onInputChange(inputValue);
  }
}, [inputValue]);
```

And the implementation:

```js
function useSavedCallback(callback) {
  const ref = useRef(callback);

  useEffect(() => {
    ref.current = callback;
  });

  return ref;
}
```

emyarod: I'm not sure if I see this optimization as strictly necessary in this case but it may be worth keeping in mind

tay1orjones: Would `useCallback` be enough to memoize `onInputChange` and avoid unnecessarily re-running the effect?

joshblack: @emyarod I think it's something that's helpful for us to do. In this case, the effect would synchronize with the input value changing and not the handler changing (although it would need that to correctly dispatch the update).

A totally extreme example of this idea would be if the parent component had a timer:

```js
function Parent() {
  const [count, setCount] = useState(0);

  useEffect(() => {
    const intervalId = window.setInterval(() => {
      setCount((count) => count + 1);
    }, 1000);
    return () => {
      window.clearInterval(intervalId);
    };
  }, []);

  return (
    <>
      <p>The current count is: {count}</p>
      <ComboBox onChange={() => { /* ... */ }} />
    </>
  );
}
```

Because we have the effect synchronizing with the callback, we end up with the effect being called repeatedly with the same value. Instead, if the callback is saved and then invoked by an effect parameterized by `inputValue` it would only run when the value changes.

Here is a quick demo of this idea: https://codesandbox.io/s/vigorous-goldwasser-rp3fe?file=/src/App.js

https://user-images.githubusercontent.com/3901764/114059480-07300c00-985a-11eb-8685-bae773bacf57.mov

emyarod: wouldn't saving the handler in a ref inside the parent function accomplish the same effect?

joshblack: I'm not sure, would need to explore. In this scenario, I'm imagining that the parent would be the person using Carbon so just was illustrating that splitting the effect out in this way would help out in the default case without them having to change how they author their code.

emyarod: something like this I assume https://codesandbox.io/s/pensive-sun-4ijti

given that the event handler is provided by the Carbon consumer, I am not sure about optimizing it for them by default (including `useCallback` usage as @tay1orjones suggested). although I think I am less hesitant about baking `useCallback` into our component if we are certain we want to optimize the handler no matter what

tay1orjones: I'm not opposed to optimizing it now to prevent problems down the line, though if it is an uncommon use case that the majority of consumers won't run into we can point them here (or other docs?) to instruct on how they can fix downstream in their application code.

Having this hook could be useful though, it would apply to any callback placed in a useEffect that has a dependency on one or more variables/props. I'm not sure how common that is in the codebase right now though.

joshblack: Definitely meant the counter example as the extreme case, the big thing is that in a default setup this implementation would:

- Have the effect and handler be called for non-related updates
  - The effect would be called because the identity changes for function declarations or arrow functions
- Have the handler be called twice in intended situations like if the parent is using the value from `onChange` to update state
  - State update triggers a render that then creates a new reference for a function declaration or arrow function, this then calls the effect to be run which is the point above

And it seems like introducing this convention would:

- Make sure the handler is only called when the value is changed and at no other time
- Make sure the handler is only called once per input change

nertzy:
Would it be possible to have a first pass of a fix that only considers the inputValue changing and not the callback function?

As a consumer, I only expect `onInputChange` to be called when the input changes. If I pass a new function on every render, I would still only expect one call, to whatever the function is set to at the time the input change occurs.

In my application, `onInputChange` is getting called unconditionally on every render, even though both inputValue and the callback function are not changing. When I type in an unrelated separate text field, it triggers onInputChange on every combo box on the page. We essentially had to write our own condition in our callback to remember the previous value of the input and only continue if the input value actually changed, which feels like a waste of effort.

Thanks for looking at this so deeply. I hope that we can figure out a good solution for the long term.

joshblack:
@nertzy would this thread address what you're describing or would it be another change? 👀  https://github.com/carbon-design-system/carbon/pull/8308#discussion_r607948337

nertzy:
> @nertzy would this thread address what you're describing or would it be another change? 👀 [#8308 (comment)](https://github.com/carbon-design-system/carbon/pull/8308#discussion_r607948337)

If I understand it correctly, I think it would work for my use case. As long as `[inputValue]` is the second argument to the `useEffect` call, then my application won't receive a call to the callback on every render.

I don't know enough about the `useSavedCallback` concept to have any advice on whether that would be necessary.

emyarod:
@joshblack @tay1orjones circling back on this, so is the expectation here to include a new `useSavedCallback` hook or can this fix go in first?

joshblack:
@emyarod I don't think you need to do the hook explicitly, but I think saving the callback in a ref and calling it would be great for this PR and then we can get it in 👍  (Like in this pattern: https://github.com/carbon-design-system/carbon/blob/main/docs/style.md#hooks-that-use-a-callback)

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -113,6 +113,7 @@ const ComboBox = (props) => {
   );
   const [prevSelectedItem, setPrevSelectedItem] = useState(null);
   const [doneInitialSelectedItem, setDoneInitialSelectedItem] = useState(null);
+  const { current: handleInputChange } = useRef(onInputChange);
```
joshblack:
```suggestion
  const savedOnInputChange = useRef(onInputChange);
```

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -147,10 +148,10 @@ const ComboBox = (props) => {
   };
 
   useEffect(() => {
-    if (onInputChange) {
-      onInputChange(inputValue);
+    if (handleInputChange) {
+      handleInputChange(inputValue);
     }
-  });
+  }, [handleInputChange, inputValue]);
```
joshblack:
```suggestion
  }, [inputValue]);
```

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -147,10 +148,10 @@ const ComboBox = (props) => {
   };
 
   useEffect(() => {
-    if (onInputChange) {
-      onInputChange(inputValue);
+    if (handleInputChange) {
```
joshblack:
```suggestion
    if (savedOnInputChange.current) {
```

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -147,10 +148,10 @@ const ComboBox = (props) => {
...
+      handleInputChange(inputValue);
```
joshblack:
```suggestion
      savedOnInputChange.current(inputValue);
```

joshblack (changes requested):

packages/react/src/components/ComboBox/ComboBox.js
joshblack: I think we'll need to not destructure this so that the effect doesn't keep changing like in the previous implementation when passing in the prop.

packages/react/src/components/ComboBox/ComboBox.js
emyarod: I was still getting the lint error about dependencies here when I removed it earlier but it might be a false positive

packages/react/src/components/ComboBox/ComboBox.js
```diff
@@ -147,10 +148,10 @@ const ComboBox = (props) => {
   };
```
joshblack, line 149: I think the missing bit to wrap up the saved callback bit is adding:

```js
useEffect(() => {
  savedOnInputChange.current = onInputChange;
}, [onInputChange]);
```

To make sure the ref is updated when the prop changes

joshblack (approved):

tw15egan (approved):

PR #8313, opened 2021-04-06
fix(ComboBox): highlight selected item based on item id if present

### LINKED ISSUE #7514: ComboBox highlights all items after selection

- [x] `carbon-components`

## Detailed description

When passing in an array of items to `ComboBox`, if the items don't have `id`s on them, after selecting one, all items are given the `bx--list-box__menu-item--highlighted` class and highlighted.

I believe this is because of this line here https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/ComboBox/ComboBox.js#L451

If neither item has an `id`, `undefined === undefined` and it is considered highlighted.

ComboBox (And possibly other listboxes?)

> see changed?

I want to be able to select objects that use a different unique field for identifying them - the Storybook docs don't mention the `id` field is required on object items, and simply say "We try to stay as generic as possible here to allow individuals to pass in a collection of whatever kind of data structure they prefer", so I'd think it should support items of any structure.

Perhaps an `itemToId` function could be passed in as a prop to allow mapping a different field to id for highlighting, that defaults to `item => item.id`

I currently work around this by modifying data from the api and injecting the missing `id` prop, but ideally I should be able to pass it in unchanged.

Chrome

Present as of
```
"carbon-components": "^10.25.0",
"carbon-components-react": "^7.25.0",
```

> should be aware of?

No currently pressing ship dates, but it would be nice to fix.

1. Create a ComboBox selecting from Object items that do not have `id` props
2. Select an item
3. Open the ComboBox again
4. The other displayed items are also highlighted.

https://codesandbox.io/s/sleepy-rain-69vkt

## Additional information

- Screenshots or code

Screen Shot 2021-01-06 at 7 58 17 PM

```
import React from "react";
import { render } from "react-dom";
import { ComboBox } from "carbon-components-react";
import "carbon-components/css/carbon-components.css";

const items = [
  { label: "item1" },
  { label: "item2" },
  { label: "item3" },
  { label: "item4" },
  { label: "item5" }
];

const App = () => (
  <div>
    <ComboBox id="demo" items={items} />
  </div>
);

render(<App />, document.getElementById("root"));
```

emyarod:
Closes #7514

This PR checks whether or not an item contains an id key before highlighting based on a potentially undefined property

#### Testing / Reviewing

Pass in items without an id key to the ComboBox component or duplicate the provided example and confirm that only selected items are highlighted


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #8315, opened 2021-04-06
feat(DatePicker): add disable, enable flatpickr options

### LINKED ISSUE #8312: Datepicker not passing down `disable` flatpickr option
Brought up in slack, our `DatePicker` is not passing down the `disable` option of `Flatpickr`

https://flatpickr.js.org/examples/#disabling-dates

We should pass this down so users have the flexibility to block out certain days, weeks, weekends, etc...

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8312

This adds in support for the Flatpickr `disable` and `enable` options. These options let a user disable specific dates (like weekends), or only enable specific dates (and disabling all others).

#### Changelog

**New**

- Adds in the `enable` and `disable` props to `DatePicker`

**Changed**

- Updated docs to give guidance for these features

#### Testing / Reviewing

Ensure you can use the `enable` and `disable` prop on `DatePicker` to set specific dates. Both props cannot be used at the same time


emyarod (commented):
since we are exposing these options the component should probably also accept prop value changes

image

tw15egan:
@emyarod should be working as expected now 👍🏻

andreancardona (approved):
looks good!

andreancardona (commented):
@tw15egan hey TJ - sorry just saw a bug but also maybe I'm misunderstanding how the disable dates work. 

If I wanted to block out / disable a whole month ex: all of July 2021 - would I do the following as seen in the screen recording - or something else? bock dates after `07/01/21` and before `07/31/2021` if so, there is a bug where I can't scroll into any other months. Let me know your thoughts!

https://user-images.githubusercontent.com/32720851/113922455-895df900-97a4-11eb-9967-172170f0f6f4.mov

tw15egan:
@andreancardona This PR is adding in the `disable` prop, it seems like this is something to do with `minDate` and `maxDate`? I think based on your values for those props you're effectively disabling all dates, I believe that prop should work if you flipped the values (`maxDate={'07/31/2021'}`)

`disabled` can specify exact dates, a range of dates, or you can pass in a function (for example, to disable weekends). 

You can test by editing the `Single` story directly and using something like


DIsable weekends:
```jsx
<DatePicker 
   datePickerType="single"
   disable={[(date) => (date.getDay() === 0 || date.getDay() === 6)]}
>
    <DatePickerInput
      placeholder="mm/dd/yyyy"
      labelText="Date Picker label"
      id="date-picker-single"
    />
  </DatePicker>
```

Disable all of July
```jsx
<DatePicker
    datePickerType="single"
    disable={[
      {
        from: '07/01/2021',
        to: '07/31/2021',
      },
    ]}>
    <DatePickerInput
      placeholder="mm/dd/yyyy"
      labelText="Date Picker label"
      id="date-picker-single"
    />
  </DatePicker>
```

andreancardona:
@tw15egan ahh sorry I see now - thank you for clarifying! Looks good :)

andreancardona (approved):

emyarod (approved):

PR #8320, opened 2021-04-07
fix(Modal): adjust modal grid columns

### LINKED ISSUE #8314: Wide content in modals overflows eccentrically with no scroll availability

- [x ] `carbon-components`
- [x ] `carbon-components-react`

## Detailed description

If you put wide content into the header, body or footer of a modal dialog, and that content has a preferred width greater than the width of the dialog, the dialog does not constrain the containing `<div>` to the dialog width but allows it to overflow the dialog. As no scrolling is enabled, this makes part of the content unavailable to the end user.

For example, put a tab bar with several tabs into the modal header. Ordinarily, when there is insufficient width for the tab bar, the tab bar uses the available width and enables sideways scroll buttons to ensure all the tabs can be reached. However, when placed into the modal header, the tabs will simply extend off the right-hand edge of the dialog, leaving the last tabs invisible and unreachable. Or, if the scroll behaviour does activate, it doesn't allow scrolling all the way to the end.

Yes, modal. In carbon-react, the ComposedModal component can be used to see this problem.

> see changed?

I expected the header, body and footer of the modal to be constrained to the width of the modal. I would like to see it changed so that this happens.

The problem is caused by the way the css grid is being used. `bx--modal-container` currently has the style rule `grid-template-columns: auto;` (see https://github.com/carbon-design-system/carbon/blob/4acadadb8a34d98a40ab09e95704d08cdb1181ca/packages/components/src/components/modal/_modal.scss#L74). This permits wider items in the grid to extend invisibly off the right-hand edge of the modal container. If this is changed to `grid-template-columns: 100%;`, grid items are constrained to the modal width. In the test case below on codesandbox, this can be seen by changing the style rule on the fly: the tab bars immediately adjust to show scrolling behaviour and scroll to reach all the tabs in the bar.

Google Chrome Version 89.0.4389.114 (Official Build) (x86_64)

10.32.0

> should be aware of?

Cloud & Cognitive PAL components

See https://codesandbox.io/s/carbon-issue-8314-testcase-tm54k

## Additional information

image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8314

Adjusts the `grid-template-columns` style so that children elements do not overflow the modal container

#### Changelog

**Changed**

- `grid-template-columns` changed from `auto` to `100%`

#### Testing / Reviewing

Ensure `Modal` still renders correctly


emyarod (approved):
looks good to me

jnm2377 (approved):
LGTM

PR #8326, opened 2021-04-07
feat(Tokens): update color tokens pt. 2

### LINKED ISSUE #8300: Update 25% of components to use new design tokens (2/4)
Ref #8148 
Ref #8149 

Components to be updated: 

- [x] FileUploader
- [x] Form
- [x] InlineLoading
- [x] Link
- [x] List
- [x] ListBox
- [x] Loading
- [x] Modal
- [x] MultiSelect
- [x] Notification
- [x] NumberInput

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8300

More color token updates. Left a comment for any deprecated `light` tokens, but left them as is for now. List of components updated:

- [x] FileUploader
- [x] Form
- [x] InlineLoading
- [x] Link
- [x] List
- [x] ListBox
- [x] Loading
- [x] Modal
- [x] MultiSelect
- [x] Notification
- [x] NumberInput

#### Changelog

**Changed**

- Updated color tokens

#### Testing / Reviewing

Ensure the components listed above have no regressions

### DIFF
--- a/packages/components/src/components/list-box/_list-box.scss
+++ b/packages/components/src/components/list-box/_list-box.scss
@@ -75,15 +75,15 @@ $list-box-menu-width: rem(300px);
     width: $list-box-width;
     height: rem(40px);
     max-height: rem(40px);
-    color: $text-01;
-    background-color: $field-01;
+    color: $text-primary;
+    background-color: $field;
     border: none;
-    border-bottom: 1px solid $ui-04;
+    border-bottom: 1px solid $border-strong;
     cursor: pointer;
     transition: all $duration--fast-01 motion(standard, productive);
 
     &:hover {
-      background-color: $hover-ui;
+      background-color: $field-hover;
     }
   }
 
@@ -98,13 +98,14 @@ $list-box-menu-width: rem(300px);
   }
 
   .#{$prefix}--list-box--expanded {
-    border-bottom-color: $ui-03;
+    border-bottom-color: $border-subtle;
   }
 
   .#{$prefix}--list-box--expanded:hover {
-    background-color: $field-01;
+    background-color: $field;
   }
 
+  // V11: Possibly deprecate
   .#{$prefix}--list-box--expanded:hover.#{$prefix}--list-box--light:hover {
     background-color: $field-02;
   }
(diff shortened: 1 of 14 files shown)


packages/components/src/components/list-box/_list-box.scss
```diff
@@ -252,7 +257,7 @@ $list-box-menu-width: rem(300px);
   }
 
   .#{$prefix}--list-box.#{$prefix}--list-box--inline.#{$prefix}--list-box--expanded:hover {
-    background-color: $field-02;
+    background-color: transparent;
```
aagonzales, line 260: Is this for a light variant? If not what is this line item for? Just want to make sure we're not changing the color of something.

aagonzales (approved):
- [x]  FileUploader
- [x]  Form
- [x]  InlineLoading
- [x]  Link
- [x]  List
- [x]  ListBox
- [x]  Loading
- [x]  Modal
- [x]  MultiSelect
- [x]  Notification
- [x]  NumberInput


I think it all looks good. Just had one comment about one line to double check. Gonna 👍 its

packages/components/src/components/modal/_modal.scss
```diff
@@ -49,6 +49,7 @@
         visibility 0ms linear;
     }
 
+    // V11: Question for design: do we have an updated tokens for fields that exist on `layer`?
```
emyarod, line 52: @aagonzales does this have an answer or is this an open question?

packages/components/src/components/list-box/_list-box.scss
tw15egan: This selector specifically is for the inline variant of dropdowns, when expanded, we want to keep the input field from getting a hover style when the menu is hovered. 

Screen Shot 2021-04-09 at 7 47 44 AM

I believe previously we were using `field-02` for the inline variants, but realize we should switch them to `transparent` in the case they are on a different background. This specific selector was likely was just not updated then.

In their normal state, we have their background set to `transparent` a few lines above, so this just aligns it with that style. If you scroll up a bit to line 233/238 you should be able to see

packages/components/src/components/modal/_modal.scss
aagonzales: I talked to @tw15egan about this. Fields that exist on a layer (like in modal or on a tile) would be styled using an inline theme now. So it would still be `$field` just from the theme in the next step in the layering modal. @joshblack talked about potentially building some of this inline theming logic into the components. 

Some products teams have argued for keeping a `$field-02` and light variant in each theme so each theme is a little more independent. This is a big question we're hoping to answer with our alpha and beta testing for the major release.

tw15egan: For now, I've left a little `// V11: possibly deprecate` note near all usage of `light` specific tokens so we can easily find them once we determine a path 👍🏻

emyarod (approved):
looks good to me

dakahn (approved):

PR #8344, opened 2021-04-09
style(selectable-tile): use checkbox icon instead of round checkmark

janhassel:
The `RadioTile` and `SelectableTile` components are great to give more details or visual cues to multi- and single-select options. However, since both share the same visual style, it can be hard for users to figure out whether they're dealing with a multiselect or a singleselect until they interact with it and observe if other tiles in the same group change or not.

This PR updates the icon for selectable tiles to (1) always be visible and (2) change to an empty checkbox icon (when unselected) and a filled checkbox (when selected).
The radio tile style stays the same.

image

image

#### Changelog

**Changed**

- Replaced checkmark icon with `CheckboxCheckedFilled16` for `SelectableTile`
- When unselected, the icon in a `SelectableTile` is still shown but changes to `Checkbox16`


tw15egan:
This makes sense to me, but want @carbon-design-system/design to give the go-ahead since this will require a kit update as well 👍🏻

joshblack (approved):

tw15egan (approved):
LGTM pending design signoff 👍 ✅

joshblack:
bump @thyhmdo when you have a sec!

thyhmdo:
@janhassel Hello there, I don't think it would make sense to let users hover to see the checkbox here, I imagine they'd still be confused if it's single or multi-select. The multiselect tile isn't obvious enough but I think just changing the icon shape doesn't really solve the problem here. Can you share some visual examples in this context? 

1. Thinking showing all the checkboxes upfront may help eliminate that confusion
2. Having some support text to let users know what they can single or multi-select here.

janhassel:
@thyhmdo Apologies for the long delay!

Good point, it might still not solve all the confusion.
Due to confidentiality I can't post any example screenshots here but I can send you some via Slack if you're interested (just ping me at jan.hassel).

One key difference between checkboxes and radio buttons I see is that usually a radio button group has one option selected by default (although I think this is not forced). In that case, users could associate a group without any selection with checkboxes and as soon as a group has an option selected, the icon makes it clear.

I'm also open to show a reduced icon on unselected tiles if that's something that you folks woudl support. I just hope it doesn't create too much visual clutter.

thyhmdo:
@janhassel hey Jan, no worries, just Slacked you, let me know

janhassel:
After talking with @thyhmdo, we agreed that showing empty checkbox icons for unselected multi-select tiles will help the user much more to understand whether they're dealing with a multi-select or single-select group. The radio tile style is not affected by this.

I updated the original PR description to reflect this as well.

thyhmdo (approved):
Good discussion with @janhassel on this!

tw15egan:
Thanks, @thyhmdo for talking this over with @janhassel!

Jan, the PR looks great, just seeing an issue with disabled multi-select tiles

Screen Shot 2021-04-27 at 10 09 09 AM

tw15egan (changes requested):
Just need to address disabled states

janhassel:
@tw15egan Thanks for noticing that! Should be fixed now 👍

tw15egan (approved):
LGTM 👍 ✅

PR #8346, opened 2021-04-09
fix(breakpoint): update in between breakpoint

### LINKED ISSUE #8186: off-by-one error for breakpoint mixin `carbon--breakpoint-between`

- [ ] `carbon-components`

## Detailed description

This is an extension for the issue https://github.com/carbon-design-system/carbon/issues/5782. It was fixed for the `carbon--breakpoint-down` but not for `carbon--breakpoint-between` which also generates `max-width` in media query.

andreancardona:
Closes #8186 

- Update  `carbon--breakpoint-between`  breakpoint by adding the following: 

      map-get($min, width + 0.02),
      map-get($max, width - 0.02),

- For reference: The `carbon--breakpoint-down` was also updated and that PR can bee seen here: https://github.com/carbon-design-system/carbon/pull/5791/files


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #8349, opened 2021-04-09
[DRAFT] feat(Disclosure): added initial disclosure component and tests

### LINKED ISSUE #6633: Disclosure Component
We've seen a use-case crop up for a generic component that combines a trigger and a pop-up to enable a variety of use-cases. This primitive can be used by teams to build a variety of accessible functionality in their product.

At its core, this component would have two pieces:

- A trigger that the user interacts with by clicking or using their keyboard to disclose the content of the component
- A pop-up that contains the content 

Depending on the type of content, it may be useful for the pop-up to contain a way to close the pop-up (like a `X` present in a modal). Similarly, focus may need to be placed on the first item for certain cases like in a dialog pattern.

In terms of focus order, the content in the pop-up should appear **after** the trigger. This means that a user can do the following:

- Tab to the trigger
- Press spacebar
- Press tab to go into the content of the pop-up
- Press tab to get out of the content of the pop-up (note: this does not dismiss the pop-up)

2021 Release: 
- Accessibility Primitives:[ Disclosure](https://github.com/carbon-design-system/carbon/issues/8293)

andreancardona:
Closes #6633 

- This is a draft PR for the base of the disclosure component 
- Next steps: 

- [ ] add some basic styles
- [ ] update tests: 
   - testing for onClick 
   - make sure ID stays on node 
   - make sure REF stays on node


joshblack:
Closing since you all have the updated PR out!

PR #8350, opened 2021-04-09
feat(Tokens): restrict new tokens behind feature flag

### LINKED ISSUE #8351: Achieve a way to update all components to new tokens, but allow custom theming
We need a way to handle teams implementing custom themes using the old color tokens, but also allow us to update the components to use the new color tokens. 

I think this may be possible by setting the new tokens to the old token values if the `enable-2021-release` feature flag is not set.

See #8269 for exploration

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8351

Based on a [comment](https://github.com/carbon-design-system/carbon/pull/8269#issuecomment-816839719) in the token update PR #8269 

Unless the feature flag is set, all old tokens will map to the new token variables inside the styles package. That way, someone using a custom theme and referencing the old tokens like `$ui-01`, will not be broken when we merge in the token updates. 

In this example, I've updated the `Link` component to use the new `$link-primary` token.

To test, you can look at `.storybook/styles.scss` in the React package. At the top, you can see the `enable-2021-release` flag is set to false. 

Screen Shot 2021-04-09 at 2 10 01 PM

When this flag is false, we have a block of code that sets `$link-primary: $link-01;` inside `packages/components/src/globals/scss/_theme-tokens.scss` 

Screen Shot 2021-04-09 at 2 10 34 PM

That way, all of the old tokens can still be used to create custom themes, but internally we can reference the new tokens.

Then, if you flip that to true, you can see that the `$link-01` variable is no longer changing the link color to red. Now, we can reference the new color token to change the link color.

Screen Shot 2021-04-09 at 2 12 22 PM

**Testing**
Follow the steps above and try some other tokens out. Since no token update shave been merged yet, you'll have to add some new tokens somewhere to test that they properly map.
Also, ensure all token names are correct (mainly hyphenation)


tw15egan:
I believe the theme switching is not working because I'm explicitly setting the `enable-2021-release` flag I the storybook (and I was omitting `enable-css-custom-properties`). Locally, if I add the custom properties flag, I'm able to change it both ways:

Bottom of `.storybook/styles.scss`
```scss
:root {
    @include carbon--theme($carbon--theme--white, true) {
      @include emit-component-tokens($tag-colors);
      @include emit-component-tokens($notification-colors);
      --cds-link-primary: red;
    }
  }
```

Should we enable custom properties by default if the `enable-2021-release` flag is set? We would then need to add another, non `feature-flag` variable to properly disable this, since I'm not able to tell whether a user is trying to set `enable-css-custom-properties: false` or it's just the default value. 

I'll add an example

tw15egan:
Okay, so I just added a block to `feature-flags.scss` that will enable custom properties by default when the `enable-2021-release` is set. In case someone wants to use things under the `enable-2021-release`, but doesn't want css custom properties, I added a `$css--disable-css-custom-properties` flag that is set to `false` by default, that will skip over enabling custom properties. 

```scss
@if feature-flag-enabled('enable-2021-release') {
  @if variable-exists(css--disable-css-custom-properties) ==
    false or
    $css--disable-css-custom-properties ==
    false
  {
    $feature-flags: map-merge(
      $feature-flags,
      (
        enable-css-custom-properties: true,
      )
    );
  }
}
```

I'm going to update this PR with 4 examples in the storybook `.storybook/styles.scss` file for testing. I think it should capture all scenarios, but definitely let me know if I'm missing a use-case

joshblack (approved):
Awesome 🔥

joshblack:
> Should we enable custom properties by default if the enable-2021-release flag is set?

Ugh, definitely get what you mean about this situation. I think if we see that flag flipped on then it's totally fair game to just flip on the custom property flag and assume it's always on while the 2021 flag is on. Not sure if it helps to fix the problem you're describing but hopefully might make some things simpler

tw15egan:
Yeah setting the enable custom properties to always on by default solved the issues I was describing, so I added an extra flag `$css--disable-css-custom-properties` so that a user could enable the new release, but turn off custom properties if needed 👍🏻

tw15egan:
@dakahn and @andreancardona this is good to go now, I'll remove the commented-out scenarios and the `link.scss` changes before merging. Once this is in, I can rebase the token PR's and verify even more with the updated components

andreancardona (approved):
looks good! @tw15egan 🔥

dakahn (approved):

PR #8361, opened 2021-04-12
feat(link): add support for icon with standalone links

janhassel:
The [usage guidelines for the Link](https://www.carbondesignsystem.com/components/link/usage#standalone-link) component point out that standalone links can be paired with icons. However, there is no direct API for it, resulting in different styling across different teams (spacing between text and icon, vertical alignment). This PR adds a `renderIcon` prop to the link to standardize this variant.

#### Changelog

**New**

- `props.renderIcon` for `Link`

#### Testing / Reviewing

Ensure that...
- the icon is not rendered when `props.inline` is passed
- the icon always changes colors together with the text (hover, active, visited)

Question for design: there was no visual specifiation regarding this variant, so I took the image from the usage guidelines as a template. There, `spacing-03` was used between the text and the icon. Let me know if this should be changed.

For reference, left top is the mockup from the usage guidelines, bottom and right are the implementation of this PR.
image


packages/components/src/components/link/_link.scss
```diff
@@ -95,6 +96,12 @@
   .#{$prefix}--link--lg {
     @include type-style('body-short-02');
   }
+
+  .#{$prefix}--link__icon {
```
joshblack, line 99: Would be great to get the following snippet to "Just work":

```jsx
<Link href="...">Text <SomeIcon /></Link>
```

I'm assuming the big challenge is adding in the correct spacing for the icon (in this case `$spacing-03`) so we need a selector to target it and then need to provide that to the component itself, correct?

Just thinking out loud, this may be what we have to end up going with now, but would be interesting to explore either "gap" or custom property based solutions in the future. Gap would be interesting in the context of grid (and flex when it gets it) since I think it only is applied when there are more than one items. `display: inline-grid` plus `grid-column-gap` and `grid-auto-flow: columns` may be a nice trick in the future for these.

janhassel: Interesting idea. We couldn't directly prevent the misuse of inline links with icons then though, could we?

joshblack: Maybe we add in a class for the "inline" link, I guess? Then scope the CSS rules outside of that.

Definitely thinking for later on, right now this seems like a solid step given what our API and constraints are 👍

johnbister (approved):
Looks consistent with the guidelines. Do you know if we’re planning on doing sm/lg variants of link with icon in the future? If so, we’ll want to adjust the spacing accordingly, but for the default link size this looks good to me!

janhassel:
@johnbister There is actually support for `sm` and `lg` links right now.
If you let me know what spacing each size variant should have, I can update the PR accordingly.

| Variant | Font size | Spacing |
| :- | :- | :- |
| `sm` | 0.75rem / 12px | |
| `md` (default) | 0.875rem / 14px | `$spacing-03` |
| `lg` | 1rem / 16px | |

johnbister:
Oh nice!   After playing around with it $spacing-02 and 04 is too big of a jump for the sm/lg sizes and it doesn't look right.

****Let's keep them as is--all $spacing-03.** Looks good! Thanks @janhassel**

emyarod (approved):
looks good to me

tw15egan:
@janhassel just seeing a slight change when the disabled styles are enabled, perhaps a line-height change somewhere?

janhassel:
@tw15egan Thanks for catching that! When disabled, the `display` property was set to `inline` explicitly and therefore overwrote `inline-flex` from the normal/enabled style.

tw15egan (approved):

joshblack (approved):

PR #8366, opened 2021-04-12
feat(PasswordInput): support `warn` and `inline` props

### LINKED ISSUE #8365: PasswordInput missing support for `inline` and `warn` props
(viewable in the current published storybook)

The PasswordInput component currently does not support the `inline` prop or the `warn` prop correctly unlike the normal TextInput component. Setting values for those props in the password input has no effect at the moment

emyarod:
Closes #8365

This PR adds support for the `warn` and `inline` props in the PasswordInput component to match TextInput

#### Changelog

**New**

- `warn` and `warnText` props to enable warning state
- `inline` prop to support inline password input variant

**Changed**

- password input markup more closely matches text input now
- vertical alignment of inline text input elements

#### Testing / Reviewing

Ensure the newly supported props work properly in the PasswordInput component


tw15egan (approved):
LGTM 👍 ✅

laurenmrice (changes requested):
- The `Label` needs to bottom align with the text in the input. Right now its a little higher.
Artboard

- When the inline knob is selected and then selecting the warning state knob, the label and helper text shift down. They should remain in the same place. This also seems to happen for the invalid state.
warning


- *Note - Eventually I think design needs to redesign the inline variant for form components across the board. I have realized it might make more sense to keep the helper text under the input, since it now gets replaced by the invalid and warning text that appears for these states. But this will be a design spec we can provide for a later PR. 👍

laurenmrice (approved):
Looks good! Thanks 🙌

jnm2377 (approved):

PR #8368, opened 2021-04-12
feat(PasswordInput): sync input type prop and state

emyarod:
This PR deprecates `TextInput.ControlledPasswordInput` and adds support for control over the password input field type in the `TextInput.PasswordInput`

this PR includes changes from #8366 so we should review and merge those changes first

#### Testing / Reviewing

Confirm that the controlled password input story which uses `TextInput.PasswordInput` instead of `TextInput.ControlledPasswordInput` functions as expected


joshblack:
Is this one intended to be fully controlled/uncontrolled or is it more the more hybrid pattern where it has internal state and then if the prop passed in changes it updates?

emyarod:
> the more hybrid pattern where it has internal state and then if the prop passed in changes it updates

the component should now follow this pattern

jnm2377 (approved):

tw15egan:
Noticed a few issues:

Size variants are causing misalignment with the password icon
Screen Shot 2021-05-12 at 12 55 54 PM


Not seeing any focus styles on the password icon when using keyboard
Screen Shot 2021-05-12 at 12 56 03 PM

emyarod:
these are not regressions and should be addressed separately (the portions of code this PR modifies are unrelated)

tw15egan:
Sounds good!

tw15egan (approved):

PR #8376, opened 2021-04-12
feat(tokens): update color tokens pt3

### LINKED ISSUE #8303: Update 25% of components to use new design tokens (4/4)
Ref #8148 

Components to be updated: 

- [x] Tag
- [x] Textarea
- [x] TextInput
- [x] Tile
- [x] TimePicker
- [x] Toggle
- [x] Tooltip
- [x] Treeview
- [x] UI Shell (replace old color tokens, old UI Shell tokens will be updated in a separate PR)

We will also need to update all global scss files to reference new color tokens
- [x] `globals/scss`

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8303

Components to be updated: 

- [x] Tag
- [x] Textarea
- [x] TextInput
- [x] Tile
- [x] TimePicker
- [x] Toggle
- [x] Tooltip
- [x] Treeview
- [x] UI Shell (replace old color tokens, old UI Shell tokens will be updated in a separate PR)

We will also need to update all global scss files to reference new color tokens
- [x] `globals/scss`

#### Changelog

**Changed**

- Old color tokens replaced with new ones

**Removed**

- Some duplicate selectors we refactored to one selector

#### Testing / Reviewing

Go over the list above and ensure all components look correct


packages/components/src/components/data-table/_data-table-core.scss
```diff
@@ -264,6 +264,7 @@
     tbody
     tr:not(.#{$prefix}--parent-row):nth-child(even)
     td {
+    // V11: Question for design: This currently maps to $ui-02
     background-color: $data-table-zebra-color;
```
tw15egan: @aagonzales the zebra styles currently map to `$ui-02`. Do we have any updated tokens we can use for these?

aagonzales: No we don't at the moment. If we keep the light variants we will.

tw15egan: Do you know what token we should use for zebra styling then?

aagonzales: I've got to think about it a little. We may need to add a new one.

tw15egan: Sounds good. I noticed it is also used for the `bx--content` class in the [UI Shell ](https://deploy-preview-8350--carbon-components-react.netlify.app/?path=/story/components-ui-shell--fixed-side-nav) where we have it blend in with the background in the light themes, but higher contrast in dark. Didn't see any current tokens that had the same type of contrast bump between light / dark themes. 

I believe it is also used in [Toggle](https://github.com/carbon-design-system/carbon/pull/8376/files#diff-269c960611bf3ff718a8247dfef83564ca602d1213f00a7c405e9737113a331bR384)

packages/components/src/components/toggle/_toggle.scss
```diff
@@ -377,7 +381,7 @@
   .#{$prefix}--toggle-input:active
     + .#{$prefix}--toggle-input__label
     > .#{$prefix}--toggle__switch::before {
-    box-shadow: 0 0 0 1px $ui-02, 0 0 0 3px $focus;
+    box-shadow: 0 0 0 1px $background, 0 0 0 3px $focus;
```
aagonzales: I think this one would be `$focus-inset` right? Its the line between the focus border and the component?

packages/components/src/components/data-table/_data-table-core.scss
aagonzales: I'm not really sure its used correctly in UI Shell. It seems to be jumping layer and going from the background color (first layer) to the third layer (ui-02). I would assume that's a bug and it should be ui-01 instead of -02 there.

aagonzales: Let's use `$layer-accent` for zebra stripes for now.

aagonzales (changes requested):

tw15egan:
@aagonzales just pushed up some changes 👍🏻

aagonzales (approved):

emyarod (approved):

jnm2377 (approved):

packages/components/src/components/ui-shell/_content.scss
```diff
@@ -15,7 +15,7 @@
 @mixin carbon-content {
   .#{$prefix}--content {
     padding: 2rem;
-    background: $ui-02;
+    background: $layer;
```
nertzy, line 18: This changes the background color for the UI Shell content tag.

Everywhere else, `$ui-01` is being replaced with `$layer`. This is the only place where `$ui-02` is being replaced. Is this an error?

tw15egan: This was an intentional change: https://github.com/carbon-design-system/carbon/pull/8376#discussion_r613421974

cc @aagonzales

Sebastian-Nguyen: Hi @tw15egan , I run the same code but on my local, content background color is ui-02, but on cloud the color is ui-01. I use a custom theme: (ui-background: #ffffff, ui-01: #f3f3f3, ui-02: #ffffff).
Do you know what make that difference? What should I do to make it run consistently on both environment?

tw15egan: Your local may be out of date and not pulling in the most recent changes. Seems like cloud is pulling in the latest that has changed it from `ui-02` to `ui-01` (`layer` in v11)

Sebastian-Nguyen: @tw15egan , you're right! I'll pull the new code changes. Thanks!

PR #8383, opened 2021-04-13
fix(button): remove dependency on tooltip component classes

### LINKED ISSUE #8372: Button component hasIconOnly assistive text styling not working

- [ ] `carbon-components`

## Detailed description

When using the `Button` component and importing the style by using 

`@import 'carbon-components/scss/components/button/_button';` 

the assistive text's styling does not work, and is in the wrong position.

Yes, `Button`

> see changed?

The `Button` scss should include styling for the assistive text

Firefox

x.27, reproduced on latest

> should be aware of?

IBM Cloud Docs

1. Create Button (hasIconOnly) component in js.
2. import Button styling
3. Hover on button

https://codesandbox.io/s/vibrant-sun-i3v25?file=/package.json

emyarod:
Closes #8372

This PR updates icon-only button tooltip selectors so that the dependency on tooltip component styles is removed.

#### Testing / Reviewing

This is not directly reviewable in storybook, but when building and using the icon only button the styles should now be standalone and no longer depend on tooltip specific component styles


joshblack (approved):

tw15egan:
Testing locally while only importing button styles, the `top` and `bottom` positions look great, but both `left` and `right` are rendering like this:

Screen Shot 2021-04-13 at 10 25 37 AM

Alignment prop works as expected with `top` and `bottom` positions

emyarod:
it looked like we currently only supported top and bottom positions in the icon-only button (https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/components/button/_button.scss#L198-L205) but I can add support for left and right as well

tw15egan:
Weird, must have gotten nuked somewhere because I definitely remember testing it in https://github.com/carbon-design-system/carbon/pull/8022 ([preview](https://deploy-preview-8022--carbon-components-react.netlify.app/?path=/story/components-button--playground))

andreancardona (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #8394, opened 2021-04-14
fix(text-input): correct padding when in warning state

janhassel:
Currently, the `TextInput` doesn't have any additional padding on the right when in warning state, resulting in the warning icon overlaying longer text. The error state accounts for that already. This PR adds the same additional padding to the warning state.

Before
image
image
image

After
image

#### Changelog

**New**

- Add class `bx--text-input--warning` to text input when in warning state

#### Testing / Reviewing

- Ensure there is enough padding to the right so that the icon doesn't overlay the text but the text is cut off earler (as seen in the screenshots above)


tay1orjones (approved):

tw15egan (approved):
Thanks for the updates, @janhassel ❤️ 🎉 ✅

PR #8411, opened 2021-04-14
feat(grid): add css-grid support to components behind a feature flag

### LINKED ISSUE #8142: Author CSS Grid Components for React
We currently offer support for `Grid`, `Row`, and `Column` components in `carbon-components-react`. With our new CSS Grid support, we will need to either update these components or add variants to work with our CSS Grid classes.

Most likely we should see if we can make use of existing `Grid` and `Column` components (`Row` is no longer needed), potentially behind a feature flag that is added to our new `@carbon/feature-flags` package. These components should use the class names that were authored for the new CSS Grid support.

tay1orjones:
Closes #8142

Modifies the Grid components to use CSS Grid behind a feature flag.

#### Changelog

**Changed**

- `Grid`, `Column` use css grid implementation behind `enable-css-grid` feature flag

#### Testing / Reviewing

The storybook example will work when https://github.com/carbon-design-system/carbon/pull/8228 is in. I can remove it after approvals before merging.


tay1orjones:
@tw15egan @emyarod this one is ready for review 👍

packages/react/src/components/Grid/Grid.js
```diff
@@ -8,30 +8,55 @@
 import { settings } from 'carbon-components';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useContext } from 'react';
+import { useFeatureFlag } from '../FeatureFlags';
 
 const { prefix } = settings;
 
+const GridContext = React.createContext({ hasGridParent: false });
```
joshblack: Might help to have this be:

```suggestion
const GridContext = React.createContext(false);
```

And then the value can be a primitive type / boolean. This can help since on every render of `Grid` the identity of the `value` will change when it's an object. That will then trigger re-renders for each child that uses `useContext(GridContext)` since the value changes. Another option could be keeping it in state if it could change, or having it be a ref if those work better.

packages/react/src/components/Grid/Grid.js
```diff
@@ -8,30 +8,55 @@
...
+
 function Grid({
   as: BaseComponent = 'div',
   condensed = false,
   narrow = false,
   fullWidth = false,
+  subgrid,
```
joshblack: Is there a good use-case for manually specifying something as a grid? Was trying to think through when someone might want to do this this 🤔  Seemed like it doing it automatically would cover most situations.

packages/react/src/components/Grid/Grid.js
```diff
@@ -51,6 +76,11 @@ Grid.propTypes = {
    */
   className: PropTypes.string,
 
+  /**
+   * Specify how many columns wide the Grid should span
+   */
+  columns: PropTypes.number,
```
joshblack, line 80: What do we need this `columns` prop for? 👀

joshblack (changes requested):

emyarod (commented):
not sure if the feature flag is being applied in this example?

image

https://deploy-preview-8411--carbon-components-react.netlify.app/?path=/story/components-grid--experimental-css-grid

tay1orjones:
@emyarod sorry, looks like the netlify deploy preview hasn't update yet

packages/react/src/components/Grid/Grid.js
tay1orjones: Provides control over how many columns are available in the grid in total. The most common would be overriding the default 16 to a 12 column grid for equal spacing of thirds. 

For a subgrid this controls how many columns it will span.

packages/react/src/components/Grid/Grid.js
tay1orjones: My thought was actually the opposite, ensuring we provide a way to opt-out of a nested `<Grid>` automatically being a subgrid.

```js
<Grid>
  <Grid>Automatically a subgrid</Grid>
  <Grid subgrid={false}>but I want to opt out</Grid>
</Grid>
```

packages/react/src/components/Grid/Grid.js
tay1orjones: Great point, fixed

tw15egan (approved):
LGTM once other comments are addressed  👍 ✅

packages/react/src/components/Grid/Grid.js
joshblack: @tay1orjones do you have a situation in mind where that happens? (when someone would want to opt-out) Just trying to think of a use-case

packages/react/src/components/Grid/Grid.js
joshblack: > The most common would be overriding the default 16 to a 12 column grid for equal spacing of thirds.

This seems like it might be against IDL stuff, right? I thought for thirds they basically say drop the last column and do 5/5/5 but I could be misremembering 🤔 

> For a subgrid this controls how many columns it will span.

If I remember correctly this should be handled by the column classes in CSS, right? Is there somewhere in the React code where we need this?

joshblack (approved):
Happy to move forward since it's experimental, had some questions in the comments though for some of the props that might be better moved over to Slack or a new issue?? I'm down either way!

packages/react/src/components/Grid/Grid.js
tay1orjones: @joshblack After thinking it through some more and playing with it in the storybook I don't think there's a good reason to have it as a configurable prop. I'll remove it 👍

packages/react/src/components/Grid/Grid.js
tay1orjones: I've never heard of the 5/5/5! 😅  That could totally be my lack of knowledge on the topic. Is it on the IDL site or in our docs?

> If I remember correctly this should be handled by the column classes in CSS, right? Is there somewhere in the React code where we need this?

Yeah, this controls which column class is applied. By default it's 16, but some users may want a subgrid to be 8, to enable stuff like another non-subgrid on the same "row".

joshblack: Nice, makes sense!

PR #8425, opened 2021-04-15
feat(Components): update color tokens (1/2)

### LINKED ISSUE #8149: Update ~25% of components to use new design tokens

jnm2377:
Closes #8149 

**Changed**

- updates color tokens for:
    - Accordion
    - Breadcrumb
    - Button
    - Checkbox
    - Codesnippet
    - Combobox

Had some questions about some of the tokens (left in comments) - can update once questions are answered.

#### Testing / Reviewing

Test against visual regression for components.


packages/components/src/components/code-snippet/_code-snippet.scss
```diff
@@ -124,13 +124,14 @@
...
     &:hover {
-      background-color: $field-01;
+      background-color: $field;
       cursor: auto;
     }
   }
 
   .#{$prefix}--snippet--light.#{$prefix}--snippet--inline.#{$prefix}--snippet--no-copy:hover {
     background-color: $field-02;
+    // this???
```
tw15egan: I've just been leaving comments anywhere we reference `$field-02` to come back and update once we figure out the inline-theming approach

jnm2377: ahh, good point. I'll update the comments to say `//TO-DO:`

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -100,11 +100,12 @@
 
   .#{$prefix}--accordion__item--disabled,
   .#{$prefix}--accordion__item--disabled + .#{$prefix}--accordion__item {
-    border-top: 1px solid $disabled-02;
+    border-top: 1px solid $button-disabled;
```
aagonzales: Should be `$border-disabled` not button

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -100,11 +100,12 @@
...
+    //should this be $border-disabled (that would make it disabled01)
   }
 
   li.#{$prefix}--accordion__item--disabled:last-of-type {
-    border-bottom: 1px solid $disabled-02;
+    border-bottom: 1px solid $button-disabled;
```
aagonzales: `button-disabled` is actually darker than the `border-subtle` color (in light themes). Disabled borders need to has less contrast. `$border-disabled` is correct.

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -117,7 +118,7 @@
     margin: $accordion-arrow-margin;
     transform: rotate(90deg) #{'/*rtl:ignore*/'};
     transition: all $duration--fast-02 motion(standard, productive);
-    fill: $ui-05;
+    fill: $border-inverse;
```
aagonzales: Is this for the chevron arrow? Should be using `$icon-primary`

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -199,7 +200,8 @@
 
     .#{$prefix}--accordion__arrow {
       transform: rotate(-90deg) #{'/*rtl:ignore*/'};
-      fill: $ui-05;
+      fill: $border-inverse;
```
aagonzales: Should be `$icon-primary`

aagonzales: This should have been using `$icon-01` not `$ui-05` its the same value but different use cases. So the migration isn't straight forward.

packages/components/src/components/accordion/_accordion.scss
```diff
@@ -219,7 +221,7 @@
   .#{$prefix}--accordion.#{$prefix}--skeleton .#{$prefix}--accordion__arrow {
     cursor: default;
     pointer-events: none;
-    fill: $ui-05;
+    fill: $border-inverse;
```
aagonzales: Change to `$icon-primary`

packages/components/src/components/button/_button.scss
```diff
@@ -444,15 +444,15 @@
   }
 
   .#{$prefix}--btn-set .#{$prefix}--btn.#{$prefix}--btn--disabled {
-    box-shadow: rem(-1px) 0 0 0 $disabled-03;
+    box-shadow: rem(-1px) 0 0 0 $layer-selected-disabled;
```
aagonzales: What is this box shadow too? Not sure `$layer-selected-disabled` is the right token for usage.

aagonzales: Is it the divider between buttons? It think `$icon-on-color-disabled` fits the usage a little better though still no quite right. We could add `$button-separator-disabled` if we wanted to be super clean about it but I dont think we need too.

packages/components/src/components/code-snippet/_code-snippet.scss
```diff
@@ -302,7 +303,7 @@
   }
 
   .#{$prefix}--snippet-button:hover {
-    background: $hover-ui;
+    background: $background-hover;
```
aagonzales: This would technically be `$field-hover` since its on top of `$field`

packages/components/src/components/code-snippet/_code-snippet.scss
```diff
@@ -370,12 +371,12 @@
...
     transition: $duration--moderate-01 motion(standard, productive);
-    fill: $text-01;
+    fill: $icon-primary;
   }
 
   .#{$prefix}--snippet-btn--expand:hover {
-    color: $text-01;
-    background: $hover-ui;
+    color: $text-primary;
+    background: $background-hover;
```
aagonzales: This would technically be `$field-hover` since its on top of `$field`

aagonzales (commented):
Just a few edits mostly based on usage distinction.

packages/components/src/components/accordion/_accordion.scss
jnm2377: Yeah, I was confused bc the existing styles use `disabled-02` and that doesn't match up to `border-disabled`. Wasn't sure if it was a style bug or intentional.

packages/components/src/components/accordion/_accordion.scss
jnm2377: yeah, I think it was just not using the correct token originally. I was quite confused 😓

packages/components/src/components/accordion/_accordion.scss
aagonzales: Yeah it can get confusing. I had to really think about this one for a minute.

packages/components/src/components/accordion/_accordion.scss
aagonzales: Yeah I found a lot of places that were using the `$ui-05` color instead of the `$icon-01`. I think this is just because we made those icon tokens later and a lot of the code didn't get updated with the new token.

aagonzales (approved):
🎉

tw15egan (approved):
Awesome work! 🎉 ✅

andreancardona (approved):

PR #8434, opened 2021-04-16
fix(modal): set pagination background color within modals

### LINKED ISSUE #8433: Pagination component’s select doesn’t have a “bx--pagination_select” style

- [ ] `carbon-components`

## Detailed description

The pagination component’s select doesn’t have a “bx--pagination_select” style on it to prevent the default override in modals (something that is already done for the left/right nav buttons - bx--pagination__button). This means that the pagination component’s select style is not overridden in modal container. As a result, the background of the select component (used in pagination component) doesn’t change to the modal background colour.

I am raising this for an external partner, so feel free to ask any questions and I can follow up them to get more details! 🙂 

## Additional information

Screenshot 2021-04-14 at 17 46 40
Screenshot 2021-04-14 at 17 47 04

emyarod:
Closes #8433

This PR sets the correct pagination background color when used within modals. The pagination components should be removed from the stories before merging

#### Testing / Reviewing

Confirm that full size pagination components nested within modals receive the correct background colors

### DIFF
--- a/packages/components/src/components/modal/_modal.scss
+++ b/packages/components/src/components/modal/_modal.scss
@@ -50,6 +50,8 @@
     }
 
     // V11: Question for design: do we have an updated tokens for fields that exist on `layer`?
+    .#{$prefix}--pagination,
+    .#{$prefix}--pagination__control-buttons,
     .#{$prefix}--text-input,
     .#{$prefix}--text-area,
     .#{$prefix}--search-input,
--- a/packages/react/src/components/Modal/Modal-story.js
+++ b/packages/react/src/components/Modal/Modal-story.js
@@ -13,8 +13,6 @@ import {
   object,
   optionsKnob as options,
   select,
-  number,
-  array,
   text,
   withKnobs,
 } from '@storybook/addon-knobs';
@@ -24,7 +22,6 @@ import Select from '../Select';
 import SelectItem from '../SelectItem';
 import TextInput from '../TextInput';
 import mdx from './Modal.mdx';
-import Pagination from '../Pagination';
 
 const sizes = {
   'Extra small (xs)': 'xs',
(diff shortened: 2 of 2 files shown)


tw15egan (approved):
LGTM, thanks for adding in the demo story!

laurenmrice (changes requested):
- The form components are not being displayed correctly in the modal unless you turn on the `Modal contains a form` knob. But when this knob turns on, the `p` text should not be lengthening, it should wrap at 20% of the modal.

Artboard

- The pagination component example is visually broken if the form knob isn’t turned on. When the form knob is turned on, playing with the pagination makes it break.
Screen Shot 2021-04-17 at 10 50 05 AM

Apr-16-2021 13-54-39

- Can we take the specific example of pagination out before being merging (similarly like we did with data table inline inputs?)

emyarod:
I'm unable to replicate the issue with the broken pagination when the form knob is off

image

for the `p` text lengthening when the form knob is on, that appears to be separate from this change and an existing problem so I can address it in another ticket

laurenmrice:
@emyarod It is happening in Firefox (88.0). I am up to date. I also had another person test this and they it is broken for them as well.

- The pagination also seems to have a visual issue with the top rule not extending all the way to the end of the component. This is happening in Firefox/Chrome/Safari.
Screen Shot 2021-04-19 at 1 57 15 PM

emyarod:
I was testing in Firefox as well and I saw the top rule issue but haven't seen the spacing issue from earlier

dakahn (approved):

tay1orjones (commented):
@emyarod I was able to reproduce the issue. FF v86 and then I updated to v88 and it still happened.

modal sizing

emyarod:
my earlier screenshots were from FF 89/90 so I will have to look into this further. I don't believe these would be caused by the color token changes in this PR so it's likely an existing issue with trying to use pagination components within modals

emyarod:
my guess is this would be related to the forced 20% right padding on the non-form modal content since the pagination component isn't responsive within modals. how should the pagination component appear in these scenario?

tw15egan:
We can track the Modal padding issue here https://github.com/carbon-design-system/carbon/issues/8373

I think this PR is good just focusing on the color updates.

laurenmrice (approved):
Approving since color looks good and we can track spacing issues separately. 👍

tw15egan:
@emyarod good to go once stories are reverted 👍🏻

PR #8442, opened 2021-04-16
fix(CodeSnippet): right-align overflow indicator if no copy button

### LINKED ISSUE #8439: CodeSnippet singleline right overflow indicator is too far left if hideCopyButton === true

- [ ] `carbon-components`

## Detailed description

I believe this issue is in the Carbon Design system itself, but I just reproduced it in the React implementation.

If the singleline code snippet doesn't have a copy button, the right overflow indicator will not change its position and will overlay the text in an ugly way:

image

While the left indicator is at the left edge of the snippet, the right indicator is offset by 2.5rem.

1. Open the [Code Snippet playground](https://react.carbondesignsystem.com/?path=/story/components-codesnippet--playground)
2. Active the `hideCopyButton` checkbox under Knobs
3. See that the overflow indicator is placed too far left in the singleline snippet

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8439

Adjusts the right overflow indicator when the `hideCopyButton` prop is passed down so that the overflow indicator aligns to the right. 

#### Changelog

**New**

- `right: 0` when `no-copy` class is present

#### Testing / Reviewing

In the storybook, go to `CodeSnippet --> Playground`  and see that the overflow indicator for the Single variant is there. Now, check the `hideCopyButton` prop and ensure the indicator is now flush against the right edge of the snippet


packages/components/src/components/code-snippet/_code-snippet.scss
```diff
@@ -502,6 +502,10 @@
     right: $carbon--spacing-08;
   }
 
+  .#{$prefix}--snippet--single .#{$prefix}--snippet__overflow-indicator--flush {
+    right: 0;
+  }
```
emyarod:
```suggestion
  .#{$prefix}--snippet--single.#{$prefix}--snippet--no-copy
    .#{$prefix}--snippet__overflow-indicator--right {
    right: 0;
  }
```

packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -164,6 +164,13 @@ function CodeSnippet({
     [`${prefix}--snippet--wraptext`]: wrapText,
   });
 
+  const codeSnippetOverflowClasses = classNames(
+    `${prefix}--snippet__overflow-indicator--right`,
+    {
+      [`${prefix}--snippet__overflow-indicator--flush`]: hideCopyButton,
+    }
+  );
+
```
emyarod:
```suggestion
```

packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -239,9 +246,7 @@ function CodeSnippet({
       {hasLeftOverflow && (
         <div className={`${prefix}--snippet__overflow-indicator--left`} />
       )}
-      {hasRightOverflow && (
-        <div className={`${prefix}--snippet__overflow-indicator--right`} />
-      )}
+      {hasRightOverflow && <div className={codeSnippetOverflowClasses} />}
```
emyarod:
```suggestion
      {hasRightOverflow && (
        <div className={`${prefix}--snippet__overflow-indicator--right`} />
      )}
```

emyarod (commented):
I believe we can use the existing classes instead of introducing a new one

tw15egan:
@emyarod just pushed up a change that removed the extra class 😄

tay1orjones (approved):

emyarod (approved):

PR #8443, opened 2021-04-16
fix(react): update CodeSnippet html from code > pre to pre > code

### LINKED ISSUE #8436: CodeSnippet singleline and multiline use invalid HTML

- [x] `carbon-components`

## Detailed description

I'm currently working on `@carbon/vue` v3. While porting the code snippet component I noticed that the HTML in the official design system docs and in the React implementation is invalid. Both the singleline and the multiline variant have this structure: `<code><pre>CONTENT</pre></code>`.

According to MDN, a [`<code>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code) may only contain phrasing content, but the [`<pre>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) is flow content. This means that sites using the Carbon code snippet component are not valid according to the W3C markup validator (and this is not just a warning, it's an error). You can try this yourself using this small snippet:

1. Open the [W3C Markup Validator](https://validator.w3.org/)
2. Click on "Validate by Direct Input", and paste the following document:
    ```html
    <!DOCTYPE html>
    <html lang="en">
      <head>
        <title>foobar</title>
      </head>
      <body>
        <code>
          <pre>CONTENT</pre>
        </code>
      </body>
    </html>
    ```
3. See the resulting error:
    > Element `pre` not allowed as child of element `code` in this context.
    > Contexts in which element `pre` may be used: Where flow content is expected.
    > Content model for element `code`: Phrasing content.

## Additional information
I believe swapping the `code` and `pre` elements around should fix this as the `code` element is itself phrasing content, which is allowed inside `pre`.

joshblack:
Closes #8436 

#### Changelog

**New**

**Changed**

- Update code snippet to use pre > code instead of code > pre

**Removed**

#### Testing / Reviewing

- Verify no visual style changes have occured

### DIFF
--- a/packages/react/src/components/CodeSnippet/CodeSnippet.js
+++ b/packages/react/src/components/CodeSnippet/CodeSnippet.js
@@ -224,13 +224,11 @@ function CodeSnippet({
         aria-label={ariaLabel || copyLabel || 'code-snippet'}
         onScroll={(type === 'single' && handleScroll) || null}
         {...containerStyle}>
-        <code>
-          <pre
-            ref={codeContentRef}
-            onScroll={(type === 'multi' && handleScroll) || null}>
-            {children}
-          </pre>
-        </code>
+        <pre
+          ref={codeContentRef}
+          onScroll={(type === 'multi' && handleScroll) || null}>
+          <code>{children}</code>
+        </pre>
       </div>
       {/**
        * left overflow indicator must come after the snippet due to z-index and


tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

andreancardona (approved):
🔥

PR #8444, opened 2021-04-16
chore(sass-loader): only use sass-loader in prod or with flag

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/8228

After this PR was merged in, `sass-loader` was drastically increasing reload time (~4000ms to ~50000ms) when style changes were made and working the storybook. This PR adds `fast-sass-loader` back in, and also adds in the `CARBON_REACT_STORYBOOK_USE_SASS_LOADER` environment variable. 

This way, a user should be able to get the fast reload speeds while developing styles, but can also use `sass-loader` if needed.

**Testing**
Pull down locally and run the storybook. Initial startup times should be about the same. By default, `fast-sass-loader` will be used when developing. Make a change to a style in `packages/components` and the refresh should take about 5000ms.

Now, close the storybook and type `export CARBON_REACT_STORYBOOK_USE_SASS_LOADER=true` in the terminal. Now, reload the storybook, and try editing the same style Reload speed will be much longer, due to using `sass-loader` instead of `fast-sass-loader`


config/storybook-preset-carbon/index.js
```diff
@@ -35,6 +37,36 @@ module.exports = {
   ].filter(Boolean),
 
   webpack(config) {
+    const sassLoader = {
+      loader: 'sass-loader',
+      options: {
+        prependData() {
```
joshblack: I think this method was changed to `additionalData` in v10 of `sass-loader

config/storybook-preset-carbon/index.js
```diff
@@ -35,6 +37,36 @@ module.exports = {
...
+          return `
+            $feature-flags: (
+              ui-shell: true,
+              enable-css-custom-properties: ${CARBON_REACT_STORYBOOK_USE_CUSTOM_PROPERTIES},
+            );
+          `;
+        },
+        sassOptions: {
+          includePaths: [path.resolve(__dirname, '..', 'node_modules')],
```
joshblack, line 54: Might help to include `implementation: require('sass')` here to make sure it will use dart sass

joshblack (changes requested):

config/storybook-preset-carbon/index.js
```diff
@@ -35,6 +37,36 @@ module.exports = {
...
+        },
+        sourceMap: true,
+      },
+    };
+
+    const fastSassLoader = {
+      loader: 'fast-sass-loader',
+      options: {
```
tay1orjones, line 62: > Might help to include `implementation: require('sass')` here to make sure it will use dart sass

Also here too for `fast-sass-loader`

config/storybook-preset-carbon/index.js
tay1orjones: @joshblack The config from your initial work on the previous PR also includes a `content` param

https://github.com/carbon-design-system/carbon/pull/8228/commits/dedd96d64a9334227e66c9892b760683446b389d

joshblack (approved):

emyarod (approved):

PR #8445, opened 2021-04-16
fix(ToggleSkeleton): update styles to match design specs

### LINKED ISSUE #6222: Toggle and ToggleSmall Skeleton broken

- [x] `carbon-components`

## Detailed description

The skeleton for ToggleSmall appears to be broken. It looks identical to a ToggleSmall without any props and doesn't show the typical animated skeleton. Additionally it doesn't provide a smaller skeleton for the label like TextInput

https://react.carbondesignsystem.com/?path=/story/togglesmall--skeleton

jnm2377:
Closes #6222

#### Specs:
md-672px-8 column

**New**

- added className for deprecated `ToggleSmallSkeleton` bc it required different styles, however shared all the same classNames

**Changed**

- updated styles for `ToggleSkeleton`

#### Testing / Reviewing

Review `Toggle` skeleton story to make sure it matches design specs.


packages/react/src/components/Toggle/Toggle-story.js
```diff
@@ -6,28 +6,9 @@
  */
 
 import React from 'react';
-import { action } from '@storybook/addon-actions';
-import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
+import { withKnobs, boolean } from '@storybook/addon-knobs';
 import Toggle from '../Toggle';
-
```
tw15egan, line 12: Can we add a Playground story so users can play around with the knobs? Would that be enough to handle the `small` variant, or do you think we should break that into a separate story?

jnm2377: Good question! We can. I just added separate stories since I know we're trying to make things copy/paste-able off the bat. I thought it would help with that. But can def add a playground story. I don't have a strong preference either way.

tw15egan: Yeah definitely! Then let's keep the current stories so that they are copy/paste-able and just add in one with all the works 👍🏻

packages/react/src/components/ToggleSmall/ToggleSmall-story.js
```diff
@@ -62,3 +63,38 @@ Default.parameters = {
       `,
   },
 };
+
+export const Skeleton = () => {
```
tw15egan, line 67: Since we've already deprecated this component, I don't think we need to add in features. Can we just add a `size` prop to the regular `ToggleSkeleton` variant if users want a smaller Toggle skeleton? I'd prefer to not add anything to `ToggleSmall` since we'll be removing it completely in a few months 👍🏻

jnm2377: I didn't add any new features to the `ToggleSmallSkeleton`, I just added the story for it. I fixed style errors that it had, bc wasn't sure if I should still fix the styles for it, but otherwise no new props or anything for this one.

jnm2377: Should I just delete the skeleton styles for it then? I wasn't sure if I should fix the `ToggleSmallSkeleton` or not? With the style updates I made to `ToggleSkeleton` it completely breaks the `ToggleSmallSkeleton`, that's why I added the new class, so that it could be temporarily fixed until it's removed all together. Wasn't sure how I should go about this.

tw15egan: I just tried rendering the `ToggleSmallSkeleton` on the `main` branch and it seems like it's already broken, so no sweat 😄    I'm down either way. If we do keep it, can you just add a `V11: ` somewhere in the comment so that we can easily remember what we need to remove down the line? 

Also, I think we'll need to remove pointer-events when it is a skeleton

laurenmrice (commented):
- I think the skeleton looks good for the toggle itself.
- I think if the toggle is paired with label text above it, we will need to have a skeleton for the label text as well. Here are some specs for the label text above the toggle.

md-672px-8 column copy

andreancardona:
@jnm2377 Let me know when this is ready for review - with the updates. :)

Issue #693, opened 2018-04-10
Structured List Padding Issue


showeimer:
## Detailed description

Left alignment is not consistent when using `bx--structured-list--condensed` between `bx--structured-list-th` and `bx--structured-list-td`.  See below:

<img width="870" alt="screen shot 2018-04-10 at 10 22 06 am" src="https://user-images.githubusercontent.com/26075187/38563143-26c01bb4-3caa-11e8-8f2e-fb700e1b0dd7.png">

> Is this issue related to a specific component? 

Structured List

> What did you expect to happen? What happened instead? What would you like to see changed?

Identical left alignment for Table Header and Table Cells
(http://carbondesignsystem.com/components/structured-list/code)

> What browser are you working in? 

Chrome (Version 65.0.3325.181 (Official Build) (64-bit)

> What version of the Carbon Design System are you using? 

"carbon-components": "^8.6.0",
"carbon-components-react": "^5.29.5",

## Additional information

<img width="591" alt="screen shot 2018-04-10 at 9 37 25 am" src="https://user-images.githubusercontent.com/26075187/38563151-2b14cf34-3caa-11e8-8b90-14fb861930a9.png">
<img width="619" alt="screen shot 2018-04-10 at 9 28 54 am" src="https://user-images.githubusercontent.com/26075187/38563155-2c60700a-3caa-11e8-97c7-4d36de56ab24.png">
<img width="259" alt="screen shot 2018-04-10 at 10 30 41 am" src="https://user-images.githubusercontent.com/26075187/38563246-5e73a77e-3caa-11e8-925a-7d1723bdd0e5.png">
<img width="339" alt="screen shot 2018-04-10 at 10 30 58 am" src="https://user-images.githubusercontent.com/26075187/38563248-60af5d12-3caa-11e8-9a00-32241997b8c2.png">
<img width="221" alt="screen shot 2018-04-10 at 10 31 18 am" src="https://user-images.githubusercontent.com/26075187/38563253-61e72c96-3caa-11e8-8955-d66fd790e80a.png">

PR #8446, opened 2021-04-16
feat(StructuredList): add prop for condensed list

### LINKED ISSUE #5736: feat(structured-list): Add React component support for the condensed structured list styles
### Summary

The `carbon-components` repo has styles for a condensed structured list using the class `{prefix}--structured-list--condensed`. The React component `StructuredListWrapper` should have a prop that toggles this class, similar to how there are props to change the sizing of `Button` and `DataTable`.

The styles can currently be added by manually including the class under the className prop.

### Justification

Since `carbon-components` have the styles, it should be supported in `carbon-components-react`.

A user can set a prop to toggle the class on the StructuredListWrapper.

### "Must have" functionality

An easy way to add the condensed class to the React component.

As there is a workaround to add this feature, there is no specific timeline.

### Available extra resources

I've added this feature in my forked carbon repo already. I will be able to submit a PR to add this feature.

jnm2377:
Closes #5736

**New**

- adds boolean for condensed list support (styles were already supported in vanilla)


tw15egan (approved):
LGTM 👍 ✅

aagonzales:
@jnm2377 Is there a design spec somewhere that you're referencing? Is the removal of the padding-left intentional?

jnm2377:
@aagonzales I didn't do anything with the styles, these already exist. I simply added a prop to be able to toggle the styles in the react component. As far as the padding, it seems like that was removed several years ago bc of this issue: https://github.com/carbon-design-system/carbon/issues/693

aagonzales:
"already supported in vanilla" I think it's that vanilla was never updated. This condensed style matches the old v9 styling of structure list. What you see in the "simple" story is the intended design. I don't think there was ever supposed to be a condensed version with the v10 update. I'll take this to the greater design team and ask them.

jnm2377:
I guess on the dev side we assumed this issue was ok to work on bc no designers ever declined the enhancement in the original issue when Akira asked for input (this was about a year ago). Lmk what design decides.
Screen Shot 2021-04-19 at 12 48 00 PM

aagonzales:
(in very joking a tone) we were a little distracted in March 2020 so I think this one just fell through the cracks. 

We talked about it in our design sync today, we think a condensed version is ok to have but this style is is flush as well as condensed and those would be two separate props. I'm going to mock-up a quick spec to demonstrate what I mean.

dakahn (approved):

joshblack (approved):

aagonzales:
Ok so adding a condensed list is one issue and there's a second one about adding a prop for flush content instead of hanging content. 

I know this PR is just about enabling the prop but I think its an old one and was never properly updated with v10 styling. I have talked to the design team and we're ok with having these 4 variants. Or to start simply adding in the indented padding to the left of the first column on the condensed list to match the padding of the default list.

image

kingtraceyj (changes requested):
For condensed, padding between columns should be 16px.

Current (8px):
image

Proposed (16px):
image

jnm2377:
@kingtraceyj @aagonzales updated to include both options via props. There's a playground story where you can toggle the diff styles to test, and also added knobs to the selection story to view on there.

kingtraceyj (changes requested):
**Condensed, Flush should have 16px between columns**
Current:
image

Proposed:
image

**Hang alignment, default spacing should have 16px spacing**
current is 32px (there's 16px before and after):
image

Proposed:
image

kingtraceyj (approved):
good to go 💥  thanks josefina! looks great.

PR #8451, opened 2021-04-19
feat(project): add dsm cli

joshblack:
**This is currently a draft. All details can and will change**

This PR looks to introduce a CLI called `dsm` (design system manager). This tool is built around bringing improved performance to our monorepo through several optimizations in our build pipeline, including:

- Support for saving and restoring build artifacts from a cache (including in CI)
- Support for incremental builds through the implemented cache
- Support for targeting specific dependencies and groups to avoid full rebuilds of the monorepo


joshblack:
Closing for this project: https://github.com/carbon-design-system/dsm

PR #8461, opened 2021-04-20
fix(tooltip): tooltip margin bug

### LINKED ISSUE #8275: Space between icon and tooltip in Carbon tooltip component does not match sketch kit
## Space between icon and tooltip in Carbon tooltip component does not match sketch kit 

- Should have 8px between icon and tooltip but looks like has 0 or 1px

### Actual in Carbon React

image (2)

### Expected in Carbon Sketch it

Screen Shot 2021-04-01 at 12 44 02 PM

andreancardona:
Closes #8275 

Adds small 8px margin to tooltip to match current design in Sketch kit


laurenmrice (changes requested):
The alignments are off when you change the `direction` prop to `left` , `right` and `top`. After fixing those I would also check them against the `alignment` prop options as well.

Everything else looks good though!

tooltip

packages/components/src/components/tooltip/_tooltip.scss
```diff
@@ -545,6 +545,7 @@
 
   .#{$prefix}--tooltip--shown {
     display: block;
+    margin-top: $spacing-03;
```
tw15egan: ~I think we'll need to add these changes to `src/globals/scss/_tooltip.scss` to account for all the alignment variants~

Looks like we can fix this by adding the correct margin to the applicable `&[data-floating-menu-direction='DIRECTION']` blocks in `src/components/tooltip/_tooltip.scss`

packages/react/src/internal/FloatingMenu.js
```diff
@@ -400,7 +400,7 @@ class FloatingMenu extends React.Component {
         ...styles,
         ...positioningStyle,
         position: 'absolute',
-        margin: 0,
+        // margin-bottom: 0,
```
tw15egan: Can we remove this comment?

laurenmrice (changes requested):
- The left and top directions with start, center and end alignments still need `8px (spacing-03)` between the icon and the caret tip. The others look good !

packages/components/src/components/tooltip/_tooltip.scss
```diff
@@ -514,6 +514,8 @@
     }
 
     &[data-floating-menu-direction='left'] {
+      margin-left: -$spacing-03;
```
tw15egan: Looks like this is an issue with negative numbers and css custom properties. We can do either:

`margin-left: calc(#{$spacing-03} * -1)`

or

`margin-left: rem(-8px)`

packages/components/src/components/tooltip/_tooltip.scss
```diff
@@ -524,6 +526,9 @@
     }
 
     &[data-floating-menu-direction='top'] {
+      margin-top: -$spacing-03;
```
tw15egan: Same as above

andreancardona:
@tw15egan @laurenmrice updated :)

laurenmrice (changes requested):
- The top direction for start,center and end alignment needs to fixed still. The rest of the directions and alignments look great!

packages/components/src/components/tooltip/_tooltip.scss
```diff
@@ -524,6 +526,8 @@
     }
 
     &[data-floating-menu-direction='top'] {
+      margin-left: calc(#{$spacing-03} * -1);
```
tw15egan:
```suggestion
      margin-top: calc(#{$spacing-03} * -1);
```

laurenmrice (approved):
Looks good to me! 🎉

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #8463, opened 2021-04-20
feat(Search): refactor search component

tw15egan:
While working on the `Search` size prop update, I realized we can't use the `feature-flags` hook since it is a class-based component. This is a WIP to update `Search` to a function component.

I've got most of the functionality working, but not sure how to handle the clear action if a user provides their own value (controlled)...

### DIFF
--- a/packages/react/src/components/Search/Search-story.js
+++ b/packages/react/src/components/Search/Search-story.js
@@ -7,7 +7,7 @@
 
 /* eslint-disable no-console */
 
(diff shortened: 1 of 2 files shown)


tw15egan:
Meant to be a draft, my mistake!

tw15egan:
Saving this for a rainy day, able to work around in https://github.com/carbon-design-system/carbon/pull/8474

PR #8474, opened 2021-04-21
feat(Search): update search size prop

### LINKED ISSUE #8371: Update Search `size` prop
Values determined by the following table:

| Name | Size |
|-----|-------|
| `xs` | 24 |
| `sm` | 32 |
| `md` | 40 |
| `lg` | 48 |
| `xl` | 64 |
| `2xl` | 80 | 

Able to workaround with `FeatureFlagContext`

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8371

Updates `Search` to align to the new `size` prop updates

| Name | Size |
|-----|-------|
| `xs` | 24 |
| `sm` | 32 |
| `md` | 40 |
| `lg` | 48 |
| `xl` | 64 |
| `2xl` | 80 | 

#### Changelog

**New**

- `FeatureFlagContext` is now exported from the internal `FeatureFlags` file
- `xl` now maps to `lg` when the feature flag is disabled
- `lg` now maps to `md`  when the feature flag is disabled
- Updated size table is commented out in the story, can be swapped when we have an alpha branch 

**Changed**

- `xl` scss class changed to `lg`
- `lg` scss class changed to `md`

#### Testing / Reviewing

There is a size story inside `Search`, you can see the old prop names still map to the correct values when the feature flag is now enabled. The new feature flag will enable the new size props. Will remove the story before merging.


packages/components/src/components/search/_search.scss
```diff
@@ -89,20 +89,20 @@
   }
 
   // Large styles
-  .#{$prefix}--search--lg .#{$prefix}--search-input,
-  .#{$prefix}--search--lg.#{$prefix}--search--expandable.#{$prefix}--search--expanded
+  .#{$prefix}--search--md .#{$prefix}--search-input,
```
joshblack: Do we need to worry about any of these changes for component implementations like Vue/Angular? 👀

tw15egan: Good point, should probably give them a heads up. I believe the `Button` update also changes 2 class names (`xl` --> `2xl`, `lg` --> `xl`)

packages/react/src/components/Search/Search-story.js
```diff
@@ -61,6 +69,34 @@ export default {
...
+      <Search placeholder="sm" size="sm" id="search-1" />
+      <br />
+      <Search placeholder="lg" size="lg" id="search-1" />
+      <br />
+      <Search placeholder="xl" size="xl" id="search-1" />
+      <br />
+    </div>
+    <br />
+    <br />
+    <FeatureFlags flags={{ 'enable-2021-release': true }}>
```
joshblack, line 87: Do you think it'd be worth toggling when the feature flag stories get shared? Maybe something like:

```jsx
import { enabled } from '@carbon/feature-flags';

export default {
  title: '...',
  // ...
  excludeStories: [
    {
      name: 'StoryName',
      enabled: enabled('enable-2021-release'),
    },
  ].filter(story => {
    return story.enabled;
  }).map(story => {
    return story.name;
  }),
};
```

joshblack (approved):
Looks great! Just had a couple of questions

packages/react/src/components/Search/Search-story.js
tw15egan: Yeah, that's a good idea, I was just going to remove the feature flag story, but I can add it as a separate story that only shows up when the feature flag is toggled at the top level. 

Do we currently have any way to flip the switch at the storybook level?

tw15egan: Hmm, not able to get this working, keeps causing build failures. Were you able to get this working on your end?

packages/components/src/components/search/_search.scss
tw15egan: Actually, on second thought, to avoid problems downstream, I just reverted it back to the original classes and added notes to update the class down the road. I'll update the Button PR in the same way

laurenmrice (approved):
Looks good! 👏

jnm2377 (approved):

Issue #6613, opened 2020-08-04
[ComboBox] Pressing enter when no options are selected does not fire onChange callback


nertzy:
## What package(s) are you using?

- [ ] `carbon-components`
- [x] `carbon-components-react`

## Detailed description

In a ComboBox, when I press enter after typing text that matches none of the options, the `onChange` callback never fires, which means that the drop-down is closed, and the text does not match the selected item. Therefore, the user has no way of knowing that the input represents a different value than is displayed.

## Steps to reproduce the issue

See my code sandbox: https://codesandbox.io/s/distracted-keldysh-pwhii?file=/src/index.js

1. Select an option
2. Type text that matches none of the options
3. Press enter
4. Notice that `onChange` never fired, so now the input is in an inconsistent state.

tw15egan:
Wondering if using the `data-invalid` attribute could be a workaround to this problem? 

```jsx
data-invalid={textValue !== item ? true : null}
```

https://codesandbox.io/s/misty-butterfly-inyun?file=/src/index.js

nertzy:
That workaround doesn't fit our use case particularly well.

We are already using the invalid state to show invalid values upon saving our form. It would be inconsistent UX for us to also introduce an invalid state dynamically before the user submits. 

I'm also not sure if we could show the user a message explaining what is wrong in this case without the message hiding and showing as the user types.

The best thing for our use case would be that some callback fires, so that we can programmatically return the ComboBox to a consistent state. I would prefer one of these options:

* Hitting "Enter" without selecting a value does nothing (instead of closing the drop-down)
* Or, hitting "Enter" or has the same effect as hitting "Escape", returning the textbox back to the current selected item's text

Alternatively, if there was some callback that fired, we could handle it in our own code.

tay1orjones:
The only way we found around this was to use event delegation. Wrap the ComboBox in a div with a keypress handler and call our onChange handler with the value in the textbox via a ref.

```js
  const comboRef = React.createRef();

  const handleOnKeypress = evt => {
    // Current value of input
    const currentValue = comboRef.current.textInput.current.value.trim();
    if (evt.key === 'Enter' && currentValue) {
      const newItem = {
        id: `sampleNewItemID`,
        text: currentValue,
      };
      handleOnChange({ selectedItem: newItem });
    }
  };

  return (
    <div onKeyDown={evt => handleOnKeypress(evt)}>
      <ComboBox 
        ref={comboRef}
        onChange={handleOnChange}
[...]

PR #3862, opened 2019-08-29
fix(Dropdown, ComboBox): adds listener to open/close menu on enter

abbeyhrt:
Closes #2823 

Components using Downshift, like Dropdown and ComboBox, were not opening on enter, this adds `toggleMenu` from Downshift to toggle the menu when `isOpen` changes. 

*Note: this problem also affects MultiSelect, since it uses Downshift but is more complicated to fix. I've made an issue (#3861) and will work on a solution for that separately. 

#### Changelog

**Changed**

- add toggleMenu and onKeydown listener to `ListBox.Field` to toggleMenu on enter

#### Testing / Reviewing

Go to ComboBox and Dropdown, tab into them and make sure they toggle open/close with Enter press.

Issue #3861, opened 2019-08-29
bug(MultiSelect): does not open menu on enter


abbeyhrt:
**Description**
Tab to the Multiselect box
Hit the Enter

The Enter key does not expand or collapse the listbox.

See ARIA Authoring Practices Guide for additional information. 

It states the following:

Enter key:

If the focus is on the button, expands the listbox and places focus on the currently selected option in the list.
If focus is in the listbox , collapses the listbox and keeps the currently selected option as the button label.

-->

## Environment

> Operating system
macOs Mojave Version 10.14.6
> Browser
Chrome

## Detailed description

> What version of the Carbon Design System are you using?
v10

> What WCAG 2.1 checkpoint does the issue violate?
I don't think this technically violates WCAG 2.1. A user can still access the menu with their keyboard, just with the spacebar and not the enter key.

asudoh:
@abbeyhrt Can this be closed or do you have more work to do? Thanks!

abbeyhrt:
Not yet! Still have some work to do on it

asudoh:
Closing given this is not a hard requirement based on our discussion.

Issue #2823, opened 2019-05-23
AVT 2 - React Dropdown - Enter key does not expand and collapse the list


snidersd:
## Environment
Tested on macOS Mojave 10.14.5, Version 74.0.3729.157 (Official Build) (64-bit), with VoiceOver

## Description
Tab to the Dropdown box
Hit the Enter

The Enter key does not expand or collapse the listbox. See [ARIA Authoring Practices Guide ](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html)for additional information. 

It states the following:

Enter key: 

-  If the focus is on the button, expands the listbox and places focus on the currently selected option in the list.
- If focus is in the listbox , collapses the listbox and keeps the currently selected option as the button label.

PR #8476, opened 2021-04-21
fix(ComboBox): prevent invalid option retention on `Enter`

### LINKED ISSUE #8170: ComboBox retains invalid option after typing and clicking enter

- [ ] `carbon-components`

## Detailed description

A value can be displayed in the ComboBox that doesn't refer to a valid option.

ComboBox

> What did you expect to happen? What happened instead? What would you like to see changed?
Typing then hitting enter... and typing then clicking away.... should have the same result.
The ComboBox selection should revert to the entry before the user started to type if no matched item is found.

Issue #6818 ComboBox doesn't support entry of Text not found in list items may be effected by this.
#6818 specifically mentions enter - but enter and click away should both be taken into account. 

FF/CH(Win + Mac), Safari(Mac), Edge(Win)

With the latest version of Carbon, at https://react.carbondesignsystem.com/?path=/story/components-combobox--combobox 

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
IBM Cognos

1. invoke the ComboBox in CodeSandbox - https://react.carbondesignsystem.com/?path=/story/components-combobox--combobox
2. enter "XXX" in the ComboBox input field
3. click outside the input field (the input field reverts to what was there before you started to type)
4. enter "YYY" in the ComboBox input field
5. enter ENTER (the input field retains the value "YYY")

The ComboBox now looks like you selected a valid option of "YYY", but "YYY" is not a valid option.

Side note:
When the ComboBox is in this state - showing an option that does not match an item - clicking on the Combobox will show the drop down list and have the previous selected item still checked.
Essentially showing 'YYY' but having 'Option 1' still being the real selected item.

You can see this easily enough via the CodeSandbox CodeSnippet above.

## Additional information

https://user-images.githubusercontent.com/5973642/112207534-9f3cbd00-8bed-11eb-91a6-d706a286832a.mov

## Attempted workaround

I created this CodeSandbox showing the attempt I made at addressing this issue.
But is does not work, the re-setting value does not work.
If anyone has any comments/suggestions how to fix this please let me know.
https://codesandbox.io/s/combobox-fix-enter-kpdne

### LINKED ISSUE #8179: Update icons-react to ship icon components with size prop

emyarod:
Closes #8170
Related #2823
Related #3862

This PR prevents an invalid option from remaining in the ComboBox on <kbd>Enter</kbd> keypress while retaining the functionality introduced in #3862 in response to #2823 (<kbd>Enter</kbd> expands the closed ComboBox menu). Alternatively, we may be able to remove the <kbd>Enter</kbd> key menu expansion due to this issue being closed as a wontfix https://github.com/carbon-design-system/carbon/issues/3861), but curious to hear others' thoughts on this

#### Testing / Reviewing

perform the test described in the referenced issue and confirm that pressing <kbd>Enter</kbd> when an invalid value is in the input field does not close the menu while retaining the previous selection (if one was made)


jnm2377:
I can see why #3861 was closed, since that one gets a bit more complicated with multiselect/open/close on enter, but I think having an event listener for enter on combo box is fine since you're only selecting one item. That being said, when I tested the component, I realized enter only opens the menu box if there is no item already selected. If I've selected an item and the menu is closed, then I press enter to try to open again, it doesn't open. Is that expected?

jnm2377 (commented):
Looks like it's mostly working. When I was testing, I noticed that if I selected an item, then tried to select an invalid item (`xxx`) and pressed enter, it didn't close. Great! But then if I deleted the `xxx` and closed the menu, the original selected item text was no longer displaying even though it shows as still selected when the menu is open. See video attached.


https://user-images.githubusercontent.com/32556167/115874806-29a15800-a40a-11eb-993c-10d60b9a3546.mov

andreancardona (commented):
also seeing Josefina's comment / bug :/

emyarod:
@jnm2377 @andreancardona this does not appear to be a regression, unless I am misunderstanding you. can you clarify what the expected behavior should be and what existing behavior is WRT #8170?

emyarod:
> That being said, when I tested the component, I realized enter only opens the menu box if there is no item already selected. If I've selected an item and the menu is closed, then I press enter to try to open again, it doesn't open. Is that expected?

regarding this question, yes this is expected as Enter typically does not expand a combobox (ref https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html)

jnm2377:
@emyarod you're right. After testing the current component, it seems that the bug already exists. I can open an issue for it. 

https://user-images.githubusercontent.com/32556167/116612324-080a0a00-a8fd-11eb-9fc0-d091d73ddf68.mov

jnm2377 (approved):

tw15egan (approved):
I think just ensuring enter **does not close the menu with an invalid option** is enough for this PR. I think if we want to change how the list-box menus are opened/closed via keyboard (Enter vs space vs down arrow) we should track in a separate issue and have a consistent pattern throughout. 

This LGTM in reference to the original ticket 👍🏻 ✅

nertzy:
This also closes #6613.

PR #8427, opened 2021-04-15
feat(Components): add new color tokens (2/2)

jnm2377:
Closes #8302 

**Changed**

- Updated color tokens for:
    - Overflow menu
    - Pagination
    - Pagination nav
    - Popover
    - Progress indicator
    - Radio button
    - Search
    - Select
    - Slider
    - Structured list
    - Tabs

#### Testing / Reviewing

Test against visual regressions.

PR #8478, opened 2021-04-21
feat(Themes): add in G80 theme

### LINKED ISSUE #8471: [element] add gray 80 theme
Add Gray 80 theme for dark inline theme layering modal. See [Notion doc](https://www.notion.so/carbondesignsystem/Final-List-of-new-tokens-fc5985c7331341baa6546064399d36da) for values.

*May have some value tweaks later. I was mostly guessing based on values of the other themes.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8471

Adds in the g80 theme for inline theming efforts

#### Changelog

**New**

- `g80` is now available as a theme.

**Changed**

- Added `g80` to the theme switcher in deploy previews

#### Testing / Reviewing

@aagonzales I didn't see a token for `linkPrimaryHover`, I think that's the only one missing


aagonzales:
@tw15egan Some of the components aren't rendering correctly because its tokens haven't been updated yet. Should we have the v10 tokens values defined for Gray 80? Or will that not matter once we update everything.

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+// No token listed in notion doc
+export const linkPrimaryHover = blue30;
+
+export const buttonDangerHover = adjustLightness(buttonDangerPrimary, -8);
+export const buttonPrimaryHover = adjustLightness(buttonPrimary, -8);
+export const buttonSecondaryHover = adjustLightness(buttonSecondary, -7);
+export const buttonTertiaryHover = white;
+
+export const backgroundSelected = gray70;
+export const backgroundSelectedHover = adjustLightness(layerSelected, +7);
```
aagonzales: In the preview this is coming up as a much darker gray. 
image

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+export const focusInverse = blue60;
+export const highlight = blue70;
+
+export const backgroundHover = adjustLightness(background, +6);
+export const layerHover = adjustLightness(layer, +7);
+export const fieldHover = adjustLightness(field, +7);
+export const backgroundInverseHover = adjustLightness(backgroundInverse, -5);
+
+// No token listed in notion doc
+export const linkPrimaryHover = blue30;
```
aagonzales, line 115: Blue30 is correct

tw15egan:
@aagonzales yeah so I added this one in the opposite way of the other theme updates with new tokens in that I set the new token value to the IDL color and referenced the old tokens to the new ones. 

Will g80 be available as a theme in v10? If not, then we probably don't need to add them if we don't expose the theme on the website or anything

joshblack (approved):

aagonzales:
No its irrelevant in v10.

jnm2377 (approved):
🔥

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+export const icon02 = iconSecondary;
+export const icon03 = iconOnColor;
+
+export const link01 = linkPrimary;
+export const link02 = linkSecondary;
+
+export const inverseLink = linkInverse;
+
+export const field01 = field;
+export const field02 = gray70;
```
aagonzales: field02 would actually be Gray60

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+
+// Interaction states
+export const inverseFocusUi = focusInverse;
+
+export const hoverPrimary = buttonPrimaryHover;
+export const activePrimary = buttonPrimaryActive;
+
+export const hoverPrimaryText = linkPrimaryHover;
+
+export const hoverSecondary = '#606060';
```
aagonzales: hoverSecondary = `buttonSecondaryHover`

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+export const activeSecondary = gray80;
```
aagonzales: `buttonSecondaryActive`

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+
+export const hoverTertiary = gray10;
```
aagonzales: `buttonTertiaryHover`

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+export const activeTertiary = gray30;
```
aagonzales: `buttonTertiaryActive`

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+
+export const hoverUI = backgroundHover;
+export const hoverLightUI = '#656565';
```
aagonzales: `#5E5E5E`

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+export const activeUI = backgroundActive;
+export const activeLightUI = gray50;
+export const selectedUI = backgroundSelected;
+export const selectedLightUI = gray60;
```
aagonzales: `Gray50`

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+export const inverseHoverUI = backgroundInverseHover;
+
+export const hoverSelectedUI = backgroundSelectedHover;
```
aagonzales: Change this to `layerSelectedHover`. In v10 `hoverSelectedUI` works for both background and ui-01, it need ti work for both so use the lighter option.

packages/themes/src/g80.js
```diff
@@ -0,0 +1,316 @@
...
+
+export const hoverDanger = buttonDangerHover;
+export const activeDanger = buttonDangerActive;
+
+export const hoverRow = '#4c4c4c';
```
aagonzales: `layerHover`

packages/themes/src/g80.js
tw15egan: Should it be -7?

aagonzales:
Something still isnt rendering correctly with the selected states. I can't tell if its a value in the theme problem or if its just because the old tokens are being used and it will be fine once the new ones come through. 

image

tw15egan:
`layerSelected` is mapped to `gray60`, so in the new theme, it will be correct. It could be a mismatch between what we had with the old tokens and what we're mapping them to for backwards compatibility for now, I see that`selectedUI` is mapped to `backgroundSelected` (`gray70`). Would mapping `selectedUI` to `layerSelected` possibly fix this?

aagonzales:
Ok lets just leave it alone then. I wont look too hard at the components until they've all been updated with the new tokens.

packages/themes/src/g80.js
aagonzales: oh yes that's what it is. should be -7

tw15egan: Fixed!

tw15egan:
@aagonzales I turned on the feature flag so it should all be using the new tokens in all components except the ones in this PR: https://github.com/carbon-design-system/carbon/pull/8427

I'll turn it off before merging but should be the best way to test 👍🏻

packages/themes/src/g80.js
```diff
@@ -0,0 +1,315 @@
...
+
+export const linkPrimaryHover = blue30;
+
+export const buttonDangerHover = adjustLightness(buttonDangerPrimary, -8);
+export const buttonPrimaryHover = adjustLightness(buttonPrimary, -8);
+export const buttonSecondaryHover = adjustLightness(buttonSecondary, -7);
+export const buttonTertiaryHover = white;
+
+export const backgroundSelected = gray70;
+export const backgroundSelectedHover = adjustLightness(layerSelected, -7);
```
aagonzales: `backgroundSelected` not `layerSelected`

aagonzales (commented):
Thanks TJ! That feature flagged really helped. Just one change I found.

packages/themes/src/g80.js
aagonzales: This is my bad, its typed wrong in the notion doc too.

aagonzales: Oops actually it should be `+7` as well. I may have misspoke earlier i think (can't find that previous comment)

tw15egan:
@aagonzales updated 👍🏻

aagonzales (approved):
🎉  cool! looks good now!

PR #8483, opened 2021-04-22
Fix nesting of accordions

alexander-schranz:
Closes #

Fix nesting of accordions.

#### Changelog

**New**

-

**Changed**

- Fix nesting of accordions.

**Removed**

- 

#### Testing / Reviewing

Nest 2 accordions into each other when open the parent the children are also automatically opened.

### DIFF
--- a/packages/components/src/components/accordion/_accordion.scss
+++ b/packages/components/src/components/accordion/_accordion.scss
@@ -197,7 +197,7 @@
         padding-bottom motion(entrance, productive) $duration--fast-02;
     }
 
-    & > .#{$prefix}--accordion__arrow {
+    & > .#{$prefix}--accordion__heading > .#{$prefix}--accordion__arrow {
       transform: rotate(-90deg) #{'/*rtl:ignore*/'};
       fill: $ui-05;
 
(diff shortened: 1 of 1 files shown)


alexander-schranz:
recheck

jnm2377:
Hi @alexander-schranz 👋🏽 Thanks for contributing to our component library! Unfortunately, at this moment we don't want to support nested accordions. We have an unstable/experimental [tree view](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-treeview--default) component that might suit your needs. Supporting nested components will not likely be on our roadmap anytime soon. There are a lot of accessibility concerns with it, but there's an ongoing discussion in this issue, if you're interested: https://github.com/carbon-design-system/carbon/issues/7328

Closing this PR for now.

alexander-schranz:
@jnm2377 thank you we are building with the accordion some kind of tested blocks so the tree view is not somethign we could be using:

Image 1:
Bildschirmfoto 2021-04-23 um 15 24 26

Image 2:
Bildschirmfoto 2021-04-23 um 15 24 29

Image 3:
Bildschirmfoto 2021-04-23 um 15 24 34

I still would recommend you in any cases that you should modifiers like `--active` and `--open` should modify there children only with a `>` selector else you always run into problems of nested CSS. 

I will currently solve it on our side with some !important overrides so still okay for closing this, just thought it would make sense to contribute this maybe back to you.

PR #8503, opened 2021-04-26
fix(tooltip): prevent propagation of escape when open

### LINKED ISSUE #8502: [Tooltip]: escape on button in tooltip propagates

- [ ] `carbon-components`

## Detailed description

If `Tooltip` is a child of a component that has an `onKeyDown` handler for 'Escape', escaping the Tooltip also calls the parent's `onKeyDown`. For example, when is a `Tooltip` in pal's `SidePanel` component, escaping the tooltip unexpectedly closes the panel. 

You can see an example of this behavior here: 
https://codesandbox.io/s/frosty-fog-xuj18?file=/src/index.js
 where `onKeyDown` is called when you try to close the tooltip with escape.

joshblack:
Closes #8502 

Prevent propagation of the escape keypress when the tooltip is open / being dismissed.

#### Changelog

**New**

**Changed**

- Update `handleEscKeyPress` to stop event propagation when we are dismissing the tooltip

**Removed**


tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

PR #8514, opened 2021-04-27
fix(toggle): update disabled toggle tokens

### LINKED ISSUE #6794: Add light variant on Toggle
Just like `TextArea` and `TextInput`, we would also need a light variant for `Toggle` when it's being used in white background, for example:
when using a disabled Toggle on white background:
image

while the expectation is:
image

development.

### "Must have" functionality

Able to see the background of the toggle, it should not blend in to the white background when it's disabled

Internal issue: https://github.ibm.com/wdp-gov/tracker/issues/31292

emyarod:
Closes #6794

This PR updates the disabled toggle color tokens so that the tokens used for toggles within data tables is now the default

#### Testing / Reviewing

Confirm the new disabled token colors match the updated spec


aagonzales (approved):
LGTM!

emyarod:
@aagonzales just to follow up on @tw15egan's question in the ticket, should we be using new token names here?

aagonzales:
oh uh I guess so? It would be `disabled-02: button-disabled` and `disabled-03: icon-on-color-disabled`

andreancardona (approved):

tay1orjones (approved):

PR #8519, opened 2021-04-27
fix(icon-builder-helpers): use default size of 16 for icons

### LINKED ISSUE #8518: Update icons next builder to use 16 as the default size
Our icons builder for the next set of icons currently default to 32x32 as the asset size when no `size` prop is provided. We should use 16x16 instead as that is the most common size.

joshblack:
Closes #8518 

This PR updates our builder to use 16 as the default size instead of 32 since it is the most common size for our icons.

#### Changelog

**New**

**Changed**

- Update next builder in icon-build-helpers to use 16 instead of 32

**Removed**

#### Testing / Reviewing

- Verify that the icon components in `icons-react/es/*` use 16 as the default value for the size prop


andreancardona (approved):
🎉

tw15egan (approved):
LGTM 👍 ✅

PR #8528, opened 2021-04-28
fix(text-input): increase disabled password toggle specificity

### LINKED ISSUE #8375: Password Text Input icon not receiving disabled colors
Noticed while working on https://github.com/carbon-design-system/carbon/issues/8303

Screen Shot 2021-04-12 at 3 38 41 PM

When `disabled`, the eye icon is not getting disabled styles due to a [more specific selector](https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/components/text-input/_text-input.scss#L105) setting the fill. 

We need to either reduce the specificity of the linked selector (ideal) or increase the specificity for the [disabled styles](https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/components/text-input/_text-input.scss#L170)

emyarod:
Closes #8375

related #8366

This PR updates the disabled password toggle selector so that it receives the proper color. It also removes a redundant selector and unnests some style blocks for better readability

#### Testing / Reviewing

Confirm the disabled password input appears correct and no visual regressions were introduced


tw15egan (approved):
LGTM 👍 ✅

laurenmrice (changes requested):
**Both Password stories:**
- The dots when hiding your password are very dark in g80, g90 and g100 themes in Safari only. Seems like a Safari bug.
Screen Shot 2021-04-30 at 2 28 43 PM

----

**Fully controlled password story:**
- You should not be able to interact with the input icon functionality. Happens in Firefox/Chrome/Safari

Apr-30-2021 14-24-38

emyarod:
the fully controlled story issue will be resolved in https://github.com/carbon-design-system/carbon/pull/8368 and the Safari issue seems to be an existing problem which I can address separately as it also impacts the text input

laurenmrice (approved):
Approving since the two issues I mentioned are being addressed in separate issues. Looks good!

dakahn (approved):

PR #8537, opened 2021-04-29
feat(Search): customize icon

### LINKED ISSUE #8480: Custom icon for search component
Actually the same issue as in #8467, but in the search component.

I would like to pass a custom icon for the search component. But in [Search](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Search/Search.js) the icon is hardcoded with icon `Search16`.

Maybe we can solve it like in https://github.com/carbon-design-system/carbon/pull/8468, or have you other suggestions?

fabpico:
Closes #8480


packages/react/src/components/Search/Search.js
```diff
@@ -176,6 +176,7 @@ export default class Search extends Component {
       disabled,
       onChange,
       onKeyDown,
+      renderSearchIcon,
```
tw15egan: Can you add this to the PropType definitions?

tay1orjones (approved):
Hi @FabianPiconeDev - I added a commit here to include the PropTypes definition as well as rename the prop to `renderIcon` to be more consistent with other components that accept an icon as a prop.

tw15egan (approved):

dakahn (approved):

FabianPiconeDev:
@tw15egan Sorry I was a bit busy, I was just about to do it, then I saw that @tay1orjones did it. Thanks! :-)

PR #8560, opened 2021-05-03
refactor(notification): improve accessibility

### LINKED ISSUE #8406: Refactor Notification
Currently, our Notification components support rendering interactive content inside of the notification itself. This content includes components like links or buttons. Unfortunately, Notifications in their current implementation are unable to correct relay this information to screen readers as all semantics are stripped in the content of a notification.

This update to Notification will include some research in order to figure out the correct way to provide accessible options for both text-only notifications and notifications with action items.

Related: #8405, https://github.com/carbon-design-system/carbon/issues/6433, https://github.com/carbon-design-system/carbon/issues/5055'

Discussion: https://github.com/carbon-design-system/carbon/discussions/8533

Links & Resources

- https://www.youtube.com/watch?v=W5YAaLLBKhQ

Packages impacted:
- carbon-components
- carbon-components-react

tay1orjones:
Closes #8406

This PR clarifies the API and intention of our notifications to guide consumers towards more accessible notifications out of the box. 

This restricts Toast notifications to not allow interactive children (actions). InlineNotification is improved to use the appropriate aria-role and focus interaction when actions are present.

The test suite will fail because In order to view the changes in storybook, I had to flip the 2021 feature flag. So before merging I'll need to flip the 2021 release feature flag back to false.

#### Changelog

**New**

- Adds a new `next/Notification.js` that is conditionally exported behind the 2021 release flag
- Notifications can be dismissed/closed by pressing <kbd>esc</kbd>
- NotificationActionButton accepts new `notificationType` prop

**Changed** (all behind 2021 release flag)

- ToastNotification `role` can be `alert`, `log`, or `status`. Default is `status`.
- ToastNotification can not contain interactive children
- InlineNotification can automatically receive focus when actions are present
- InlineNotification role will flip to `alertdialog` when actions are present
 
**Removed** (all behind 2021 release flag)

- Remove unused NotificationTextDetails
- Remove props no longer needed in ToastNotifcation: `caption`, `notificationType`, `title`, `subtitle`
  - `caption`, `title`, `subtitle` should now use `children` instead
  - `notificationType` is now handled internally and is not configurable
- Remove props no longer needed in InlineNotifcation: `notificationType`, `title`, `subtitle`
  - `title`, `subtitle` should now use `children` instead
  - `notificationType` is now handled internally and is not configurable

#### Testing / Reviewing

- Ensure stories apply the appropriate role, focus, and close via the escape key.


joshblack:
Definitely think this could make sense with removing the interactive pieces, are you all also exploring the whole `alertdialog` stuff where the notification (if it has controls) ends up being a dialog? 👀

tay1orjones:
Thanks for the nudge on `alertdialog` @joshblack 🙂  

After more exploration, the thinking on this shifted to be the following:

Toast
 - Can have `alert`, `log` or `status` role
 - No interactive elements (`actions`) or rich text
 - Component never receives focus
 - Timeout is still available and configurable, defaults to off as a best practice
 - <kbd>esc</kbd> will close
 - close button has `aria-hidden="true"`

Inline
 - Can have `alert`, `log`, `status`, or `alertdialog` role
 - Can have `actions`
 - If it has `actions`, a role of `alertdialog` will be forced and the component will steal focus
 - <kbd>esc</kbd> will close
 - If role is _not_ `alertdialog`, close button has `aria-hidden="true"`

New "Persistent" variant (open to other naming suggestions!)
- Mimics the visual styling of a Toast
- Role is set to `alertdialog` and is not a configurable prop
- `actions` are required
- Component always steals focus
 - <kbd>esc</kbd> will close

So the overall idea here is that passive/non-intrusive error, warning, and completion feedback will typically be created with `role="alert"` (Toast or Inline), while confirmation feedback that requires user interaction would use `role="alertdialog"` (Inline or Persistent).

tay1orjones:
I've applied some changes based on feedback from IBMa

- Toast's default role should be `status` (polite) so it's not so intrusive
- Automatically applying focus to a notification with actions should be optional, on by default
- Closing notifications with <kbd>esc</kbd> should be optional, on by default

tay1orjones:
I've removed `PersistentNotification` for now as it needs a design pass. So this will only include changes to Toast and Inline notification behind the v11 flag. I'll open a new issue for PersistentNotification work.

Status checks aren't going to pass here until I turn off the v11 release feature flag, but wanted to keep it on for now so the deploy preview storybook shows the changes for review. Ready for review.

packages/react/src/components/Notification/next/Notification.js
```diff
@@ -0,0 +1,536 @@
...
+
+  /**
+   * Specify an optional className to be applied to the notification action button
+   */
+  className: PropTypes.string,
+
+  /**
+   * Specify the notification type
+   */
+  notificationType: PropTypes.oneOf(['inline']),
```
jnm2377: Is this prop needed if there is only one type of notification for this component? I'm a little confused about what it does for this if there aren't other options.

jnm2377:
It looks like some of the spacing is off if I do Toast notification with hidden close button 👀 
Screen Shot 2021-05-14 at 3 57 43 PM

emyarod (commented):
from the CI log it looks like the themes tests need to be updated to handle custom properties

packages/react/src/components/Notification/next/Notification.js
tay1orjones: It previously had another option of `persistent` for the PersistentNotification but since we're waiting on that I removed it and forgot to delete the prop altogether. Thanks for the catch!

tay1orjones:
> from the CI log it looks like the themes tests need to be updated to handle custom properties

@emyarod I temporarily set `enable-v11-release` to `true` so that the changes could be reviewed in storybook. I believe those failures are due to that and unrelated to the notification changeset.

kingtraceyj (approved):
looks good to me! 💥

jnm2377 (approved):
LGTM

tay1orjones:
@emyarod I can flip the v11 flag off to get CI passing if you think these changes look good 👍

emyarod (approved):
yep changes look good outside of just the CI issue mentioned previously

tay1orjones:
All of this should be behind the v11 flag now. I set it back to `false` and I removed all the changes to the storybook.

I opened an issue to do a docs pass later after all the various v11 Notification changes land, https://github.com/carbon-design-system/carbon/issues/8761

Issue #8370, opened 2021-04-12
Update Size Prop across components


tw15egan:
We want to align all `size` props so that a user knows exactly what size to expect based on the value they provide. 

Based on the following table:

| Name | Size |
|-----|-------|
| `xs` | 24 |
| `sm` | 32 |
| `md` | 40 |
| `lg` | 48 |
| `xl` | 64 |
| `2xl` | 80 | 

### Components to be updated
- [x] Accordion
- [x] Button 
- [x] Listbox
   - [x] Combobox
   - [x] Dropdown
   - [x] Multiselect
   - [x] Filterable Multiselect
- [x] Content Switcher
- [x] DatePicker
- [x] FileUploader
   - [x] FileUploaderButton
   - [x] FileUploaderDropContainer
   - [x] FileUploaderItem
- [x] Link
- [x] Modal
   - [x] ComposedModal
- [x] Number Input
- [x] Overflow Menu
- [x] Search
- [x] Select
- [x] Tag
- [x] Text Input
- [x] Time Picker
- [x] Toggle

Form elements have been worked on in https://github.com/carbon-design-system/carbon/pull/8160

Outstanding work is tracked in https://github.com/carbon-design-system/carbon/issues/8369

PR #8570, opened 2021-05-03
feat(codemod): add in codemod for size prop updates

### LINKED ISSUE #8563: Create codemod for size changes in v11
We have some `size` prop updates in v11 and it would be great if we could author a codemod for them and eventually integrate it into the `@carbon/upgrade` package.

This issue is just to add the transform in the `codemods` package and run it manually on our codebase, similar to the icons work.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8563
Refs https://github.com/carbon-design-system/carbon/issues/8370

This PR adds in a new codemod to help users convert their old `size` prop names to the new convention

Screen Shot 2021-05-03 at 3 00 20 PM

#### Changelog

**New**

- `size-prop-update.js` contains the codemod that will convert the code to the new syntax
- `code-mod-story.js` contains a bunch of current components all using the _current_ naming convention. 
- After you run the codemod, you can paste the results into the second story in `code-mod-story.js`

**Changed**

- `TableToolbarAction` was updated to use the `sm` class, and marked the old classes for removal in V11

#### Testing / Reviewing

Pull down the PR, then navigate to the `codemods` folder. From there, run `yarn jscodeshift -d -t ./transforms/size-prop-update.js '../packages/react/src/components/Test/code-mod-story.js'`. 

After that has run, you will notice a new file in the react package ( `packages/react/src/components/Test/output.js` that contains the updated code. Paste this code into the associated story in `code-mod-story.js` that contains a feature flag. 

Verify all components are using the new props, and that nothing has changed visually 🎉 

(extra code and stories will be removed after approval / before emerging)


packages/react/src/components/Test/code-mod-story.js
```diff
@@ -0,0 +1,298 @@
+/**
```
joshblack: Would be great to add these tests over into the `__testfixtures__` folder and define a test just to help catch regressions 👀 

e.g. for `small-to-size` we did: https://github.com/carbon-design-system/carbon/blob/main/codemods/transforms/__tests__/small-to-size-test.js

with these fixtures:

- Input: https://github.com/carbon-design-system/carbon/blob/main/codemods/transforms/__testfixtures__/small-to-size-prop.input.js
- Output: https://github.com/carbon-design-system/carbon/blob/main/codemods/transforms/__testfixtures__/small-to-size-prop.output.js

joshblack (changes requested):

packages/react/src/components/Test/code-mod-story.js
tw15egan: Oh yeah, definitely! I will work on adding some tests in today 👍🏻

tw15egan: @joshblack added some tests for all components that will be modified

joshblack (approved):
🔥

dakahn (approved):

PR #8577, opened 2021-05-04
fix(Tooltip): remove pointer from TooltipDefinition, add disabled TooltipIcon

### LINKED ISSUE #8542: Cursor should not be the "hand" (pointer) for non-interactive Tooltip components

- [x] `carbon-components`

## Detailed description

> A user has pointed out that seeing the pointer ("hand") icon when he hovers over a TooltipDefinition component led him to think the text was a link (or was otherwise clickable). Since that's not generally the case (unless the component surrounds or is surrounded by a link or button), it probably should use some other cursor (the user suggested the "help" cursor, but I think just leaving it as the default cursor would be fine). The same issue affects the TooltipIcon component.

TooltipDefinition and TooltipIcon

> see changed?
Change the cursor to the default, or the help cursor when hovering over these components.

Chrome

"carbon-components": "10.29.0",
"carbon-components-react": "7.29.0"

> should be aware of?
[IBM Call Home Connect Cloud](https://www.ibm.com/support/call-home-connect/cloud/). No pressing ship/release dates.

1. Go to https://www.carbondesignsystem.com/components/tooltip/usage/
2. Choose the Tooltip Icon or Tooltip Definition variants.
3. Hover your mouse over the trigger component.
4. Note that the mouse cursor is the pointer ("hand"), indicating that the item is clickable. But clicking the item does nothing. 

(Note that this is not applicable for the "interactive" Tooltip component, which the user must click action to display the tooltip.)

## Additional information

- Screenshots or code
- Notes

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8542

Removes the pointer cursor for `TooltipDefinition`, as the content shows on hover and is not interactive. Retains the pointer cursor for `TooltipIcon` if there is an `onClick`.

This also adds in proper `disabled` support for `TooltipIcon`. 

#### Changelog

**New**
- `TooltipIcon` now ships with a `disabled` prop that will change the cursor and icon fill
- Disabled tooltips now use `not-allowed` cursor

**Changed**

- Only add `cursor: pointer` if it is an `icon` tooltip with an `onClick`, otherwise use default cursor

#### Testing / Reviewing

Ensure no pointer cursor is added for `TooltipDefinition`, but it is still added for `TooltipIcon` (if `onClick` is present) and `Icon-only Button`. Ensure disabled styles render correctly for `TooltipIcon` and now tooltip is shown when it is disabled.

### DIFF
--- a/packages/components/src/components/tooltip/_tooltip.scss
+++ b/packages/components/src/components/tooltip/_tooltip.scss
@@ -151,8 +151,6 @@
       color: $text-primary;
 
       &:hover {
-        cursor: pointer;
-
         + .#{$prefix}--tooltip--definition__top,
         + .#{$prefix}--tooltip--definition__bottom {
           display: block;
@@ -182,7 +180,6 @@
     margin-top: $carbon--spacing-04;
     background: $background-inverse;
     border-radius: rem(2px);
-    cursor: pointer;
     pointer-events: none;
 
     p {
@@ -343,7 +340,6 @@
 
     display: inline-flex;
     align-items: center;
-    cursor: pointer;
     font-size: 1rem;
 
     &:focus {
--- a/packages/components/src/components/tooltip/_tooltip.scss
+++ b/packages/components/src/components/tooltip/_tooltip.scss
@@ -349,6 +349,11 @@
     }
   }
 
+  // Disabled styles
+  .#{$prefix}--tooltip__trigger:not(.#{$prefix}--btn--icon-only)[disabled] svg {
+    fill: $icon-disabled;
+  }
+
   .#{$prefix}--tooltip__label .#{$prefix}--tooltip__trigger {
     // Override `margin: 0` from button-reset mixin
     margin-left: $carbon--spacing-03;
--- a/packages/components/src/globals/scss/_tooltip.scss
+++ b/packages/components/src/globals/scss/_tooltip.scss
@@ -79,15 +79,16 @@
   overflow: visible;
   align-items: center;
 
-  @if $tooltip-type == 'icon' {
-    cursor: pointer;
-  }
-
   &:focus {
     @include focus-outline('border');
   }
 
+  @if $tooltip-type == 'definition' {
+    cursor: default;
+  }
+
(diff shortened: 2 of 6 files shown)


joshblack (approved):

jnm2377:
hmmm.. I'm still seeing the pointer on `TooltipDefinition`

https://user-images.githubusercontent.com/32556167/117069629-5a2e9f00-acf2-11eb-9404-4438ebb4253d.mov

tw15egan:
@jnm2377 oof, thanks for catching that! Just added some more changes that should take care of that and retain pointer styles for other variants.

jnm2377 (approved):
🎉 LGTM

John30013:
As the author of #8542, I would say that retaining the pointer in all cases for TooltipIcon is not the way to go. It should only have the pointer cursor if the `onClick` property is set. Otherwise, it should not specify a cursor setting at all, and allow whatever other elements it contain (or surround it) to dictate the cursor style.

tw15egan:
@John30013 I've updated `TooltipIcon` to only have a cursor change if the `onClick` prop is present

John30013:
@tw15egan Thanks for the fix! I noticed (just today) that if the `TooltipIcon` has its `disabled` prop set, the cursor doesn't change to `not-allowed` (which it does on regular buttons). The underlying button _is_ disabled (the function passed in the `onClick` prop doesn't fire), but there's no visual indication that it's disabled (the icon color doesn't change, and the cursor stays the default cursor).

tw15egan:
@John30013 Looks like we didn't have full support for disabled the `TooltipIcon` component, so I've added that in. You should now be able to see 3 different cursors in that story:
1) Interactive tooltip will show cursor
2) Non-interactive shows default cursor
3) Disabled ones show not-allowed cursor

Let me know what you think

@jnm2377 @joshblack you may want to take another look as there have been some bigger changes

John30013:
@tw15egan Looks great!

thyhmdo (approved):
Looks good to me

jnm2377 (approved):

PR #8585, opened 2021-05-05
fix(button): ignore fill="none" paths on icon svgs

### LINKED ISSUE #8583: PlayOutlineFilled32 icon doesn't render correctly when used in a Button

- [x] `carbon-components`
- [x] `@carbon/icons-react`

## Detailed description

When using the PlayOutlineFilled32 icon within the Carbon Button component (React) the whole circle appears filled with only a slight outline of the play button. When I use the icon outside of the button component it renders with a clear play icon. (Easier to see in [the sandbox](https://codesandbox.io/s/playoutlinefilled32-mpie4?file=/src/App.js:210-222))

We are using a ghost button but the issue remains when using the default button kind.

React `<Button>` used with `renderIcon` 

> see changed?

The play icon within the circle to be clearer. 

Chrome (Version 90.0.4430.93)

Reproduced in 10.34.0

> should be aware of?

IBM Robotic Process Automation 2Q release.
It's likely we can use another icon temporarily.

Use the Button component with the renderIcon prop and add the PlayOutlineFilled32 icon
`<Button renderIcon={PlayOutlineFilled32}>`

https://codesandbox.io/s/playoutlinefilled32-mpie4

## Additional information

image

emyarod:
Closes #8583

This PR ignores SVG icon paths with `fill="none"` inline styles when applying icon button color styles. An alternative solution would be to check for and remove extraneous layers on the affected assets (for example the `PlayOutlineFilled32` icon does not appear to require the `<path d="M11,23a1,1,0,0,1-1-1V10a1,1,0,0,1,1.4473-.8945l12,6a1,1,0,0,1,0,1.789l-12,6A1.001,1.001,0,0,1,11,23Zm1-11.3821v8.7642L20.7642,16Z"></path>` while `PlayOutlineFilled16` is unaffected because it does not contain this layer). This approach would be more involved as it requires an audit of every icon

#### Testing / Reviewing

Confirm that all icon button variants render correctly, the temporary PlayOutlineFilled16 and PlayOutlineFilled32 options can be removed before merging


joshblack (approved):

emyarod:
if we can update the source SVGs then this additional selector won't be needed

tw15egan (approved):
LGTM 👍 ✅

PR #8589, opened 2021-05-05
test(components): add e2e tests for selectors to detect changes

### LINKED ISSUE #8520: Detect and report changes in CSS between releases
Between releases, we often will tweak or modify our Sass styles which in turn will cause downstream effects on libraries that consume and distribute the SCSS/CSS files for different libraries/frameworks/implementations.

Unfortunately, this means that many of the changes that we make to our stylesheet would be considered breaking changes for this audience. The lack of consistency or rate of change for certain selectors can be incredibly frustrating for these maintainers and can cause a number of downstream bugs due to lack of communication on our end for these changes.

As we go into our v11 release cycle, we would like to figure out a way to report on changes in CSS structure that can be reported between releases. Ideally, this would be in a changelog for teams that need this information to easily reference what updates they will need to make in order to adopt the latest version. Some of the changes that we would like to detect include:

- Selectors
  - Selector added
  - Selector removed
  - Selector renamed (most likely this is a combo of added/remove)

joshblack:
Closes #8520 

This PR adds in a simple snapshot-based strategy for detecting changes to the selectors that we ship as part of our Sass/CSS files. It uses our compiled CSS file (`carbon-components/css/carbon-components.css`), parses it into an AST, and then finds all the CSS Rules that exist in the stylesheet.

We grab the selectors from each rule and then snapshot them in a test. When selectors change, this snapshot will fail and the release team will be notified. Changes that are intended will then be added to our changelogs so teams know what selectors to be on the lookout for since they have been removed or replaced.

This is the initial implementation of this approach to see if this strategy will work. As we learn more about it, we can change the strategy to better suite the use cases that we have.

#### Changelog

**New**

- Add an e2e test for selectors from `carbon-components`

**Changed**

**Removed**

#### Testing / Reviewing

- Verify that the snapshot produces an array of all the selectors from our styles
- Edit an existing sass file by removing or changing the name of a selector
- Build `carbon-components`
- Run `yarn test:e2e`, it should report a difference in the snapshot for the selectors of `carbon-components`


joshblack:
FYI @tw15egan this is a follow-up to that issue we were talking about, curious if you think this would be a good strategy while we start updating our stylesheets or if there is something else I should be adding in here 👀

tw15egan:
@joshblack this is perfect! Will definitely help with catching class name changes. Is there any way we can output class changes to a specific file or something so that the release team can reference it? Just trying to figure out how we can keep track of what specifically has changed between releases 🤔

joshblack:
@tw15egan good point, I could definitely update the test so that it iterates over each component's scss file instead of doing one big one. Do you think that would do the trick?

tw15egan:
I'm just trying to think of a way that when doing the release, we can pull up a file (or run a command) and see any style changes that occurred between the last release and the current release. Even if it's just being able to diff the `10.34.0` style snapshot and the `10.35.0` snapshot. Maybe we could add it as part of the release action or something? Just brainstorming.

joshblack:
@tw15egan ah okay, that makes sense. Maybe something like this in our CLI?

```bash
yarn carbon-cli diff --styles
```

And it would print out any changes for each component that you could then include in a changelog? What kind of format do you think would work best for the output for when it gets rendered in md? Happy to give it a try 👀

emyarod (approved):
looks good to me, e2e test is catching selector changes in snapshots

tw15egan:
@joshblack yeah that's exactly what I was thinking! I think just something as simple as 

### Previous selector
```scss
.bx--search-magnifier
```

### Updated selector
```scss
.bx--search-magnifier-icon
```


I don't think there should be too many between releases, so that may be all we need

tw15egan (approved):
Works great! 👍 ✅

joshblack:
Sent out a message in the Slack channel for this and seemed like there wasn't that much interest in supporting this strategy. Let me close this out and see if we can do a CLI command instead so it doesn't impact our CI times too much 🤔

PR #8595, opened 2021-05-06
fix(feature-popver-export): add popover to the export path

### LINKED ISSUE #8592: Can not import Popover from 'carbon-components-react' '
I had to manually add this code to the node_modules/carbon-components-react/es/index.js

Zrzut ekranu 2021-05-6 o 18 43 47

It was due to following error when import:

```javascript
import {Popover, PopoverContent} from 'carbon-components-react'
```
Zrzut ekranu 2021-05-6 o 18 47 14

Can I request for adding {Popover, PopoverContent} to the export path?

RadekKpc:
Resolves #8592. In order to fix issue #8592.  Could you do a code review?

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -225,3 +225,8 @@ export {
   Heading as unstable_Heading,
   Section as unstable_Section,
 } from './components/Heading';
+export {
+  Popover as unstable_Popover,
+  PopoverContent as unstable_PopoverContent,
+} from './components/Popover';
+


joshblack:
Copying over from #8592 

> Hi there @RadekKpc! 👋
> 
> This component is intentionally not exported as it is currently under development. We'll make sure to add it to the public entrypoint when it's ready or when we're looking to get feedback on it!

emyarod:
yep looks like that was opened while I was working on it

PR #8599, opened 2021-05-06
fix(overflow-menu): update ARIA role list items in OverflowMenuItem

### LINKED ISSUE #7445: JAWS doesn't navigate overflow menus correctly
Here's a thread with an explanation of what I'm seeing and a demo https://ibm-analytics.slack.com/archives/C036P1CTN/p1607619522139900
When you open an overflow menu, JAWS announces that you’re at a menu and to use arrow keys to move through menu items, but when you hit the down arrow to move to the next menu item, JAWS says “leaving menu”. And when you keep pressing the down arrow to move through more items, focus doesn't move, and hitting enter on one of those menu items does not do the right thing.

## Environment

> Operating system
Running windows 10 in vmware

> Browser
Firefox

JAWS

This happens in the react storybook example https://react.carbondesignsystem.com/?path=/story/overflowmenu--basic

tombrunet:
Follow the menuitem pattern used here: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

Closes #7445

The JAWS screen reader is confused by a `button` inside of a `menuitem`. The containing `ul` is correctly a `menu`, so the li no longer has real semantic meaning. Those are therefore `role=none`. The `button` or `a[href]` element then becomes the focusable element and should have the `role=menuitem`.

#### Changelog

**Changed**

- Changed OverflowMenuItem `li` role to `none` and added `role="menuitem"` to the `button/a[href]`

#### Testing / Reviewing

Need to open an overflow menu in Firefox with JAWS and arrow up/down. It should simply read the options with 1 of X, 2 of X, etc without stating that it's leaving the menu. Reach out to @thbrunet on Slack if an example is deployed somewhere and you just need me to confirm with JAWS.

### DIFF
--- a/packages/react/src/components/OverflowMenuItem/OverflowMenuItem.js
+++ b/packages/react/src/components/OverflowMenuItem/OverflowMenuItem.js
@@ -186,12 +186,13 @@ export default class OverflowMenuItem extends React.Component {
       );
     })();
     return (
-      <li className={overflowMenuItemClasses} role="menuitem">
+      <li className={overflowMenuItemClasses} role="none">
         <TagToUse
           {...other}
           {...{
             'data-floating-menu-primary-focus': primaryFocus || null,
           }}
+          role="menuitem"
           href={href}
           className={overflowMenuBtnClasses}
           disabled={disabled}


tombrunet:

joshblack (approved):

emyarod (approved):
looks good to me

PR #8612, opened 2021-05-07
feat(TooltipIcon): support `renderIcon`

### LINKED ISSUE #8608: Carbon React `TooltipIcon` component should accept a `renderIcon` prop
### Summary

The current `TooltipIcon` implementation requires including the icon as a child element. Most other Carbon React components that include an icon do so via the `renderIcon` prop. I'm requesting that you update `TooltipIcon` to follow this pattern, and allow specifying the icon via a `renderIcon` prop.

### Justification

This change would provide a programming interface that is consistent with other Carbon React components that display an icon (`Button`, `Link`, `OverflowMenu`, and `Tag`).

Developers would be able to use a consistent programming interface to specify the icon to be rendered by any Carbon React component that can render an icon.

### "Must have" functionality

N/A

N/A

### Available extra resources

I'm sure I could implement this myself, but it would take some time as I would need to study other Carbon React components that implement the `renderIcon` prop to see how it's done. It would probably be faster for someone on the Carbon React development team to do this. But I'll be glad to give it a shot.

emyarod:
Closes #8608

This PR adds support for custom icons via the `renderIcon` prop on `<TooltipIcon />`. The current method of using `props.children` to render the icon may be deprecated

#### Testing / Reviewing

Confirm the `renderIcon` prop functions as a drop in replacement for `props.children` custom icon rendering

### DIFF
--- a/packages/react/src/components/TooltipIcon/TooltipIcon.js
+++ b/packages/react/src/components/TooltipIcon/TooltipIcon.js
@@ -27,6 +27,7 @@ const TooltipIcon = ({
   onFocus,
   onMouseEnter,
   onMouseLeave,
+  renderIcon: IconElement,
   tooltipText,
   ...rest
 }) => {
@@ -132,7 +133,8 @@ const TooltipIcon = ({
         id={tooltipId}>
         {tooltipText}
       </span>
-      {children}
+      {IconElement && <IconElement />}
+      {!IconElement && children}
     </button>
   );
 };
@@ -148,7 +150,7 @@ TooltipIcon.propTypes = {
    * Specify an icon as children that will be used as the tooltip trigger. This
    * can be an icon from our Icon component, or a custom SVG element.
    */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
 
   /**
    * Specify an optional className to be applied to the trigger node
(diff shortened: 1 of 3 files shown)


packages/react/src/components/TooltipIcon/TooltipIcon.js
```diff
@@ -148,7 +150,7 @@ TooltipIcon.propTypes = {
    * Specify an icon as children that will be used as the tooltip trigger. This
    * can be an icon from our Icon component, or a custom SVG element.
    */
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
```
tw15egan, line 153: Do we need to add a deprecation notice? Or do we plan to still allow the old functionality, but just surface the new `renderIcon` prop in addition?

emyarod: I mentioned that it can be deprecated but I didn't make any changes to deprecate it. I'm not sure how things will change with the Popover still under development

tw15egan: Yeah I was just wondering if it would help users know about the new feature/prop, or just be annoying 😄  Fine to just leave it as is 👍🏻

tw15egan (approved):
LGTM 👍 ✅

John30013:
Wow, you guys are fast! I had actually created an implementation yesterday, but I'm having trouble getting `yarn build` to complete without errors. I also realized that my use case is actually handled by an icon-only Button (which already has a tooltip). I still think this is a valid feature request, if for no other reason than it harmonizes the programmatic interface with other Carbon React components that render icons. So I'm gratified that you jumped on it. (FYI, the preview story looks great.)

dakahn (approved):

PR #8641, opened 2021-05-10
fix(Tooltip): make screenreader description less verbose

### LINKED ISSUE #7226: Screen reader is quite verbose in "info tip dialog" (aka Tooltip)
Run [Tooltip example](https://carbon-components-react.netlify.app/iframe.html?id=tooltip--default-bottom)
Run JAWS or NVDA in Chrome or Firefox, and in each case, tab to the Tooltip button (which automatically opens the tip dialog).
Here's the output (just from typing one tab key):

#### JAWS Chrome is too verbose:
Tooltip label Button collapsed 
To activate press Enter.
Tooltip label Tooltip label 
Tooltip label Tooltip label dialog This is some tooltip text. This box shows the maximum amount of text that should appear inside. If more room is needed please use a modal instead.
Learn More Link

#### JAWS FF is ok:
Tooltip label Button 
To activate press Enter.
expanded 
Tooltip label dialog This is some tooltip text. This box shows the maximum amount of text that should appear inside. If more room is needed please use a modal instead.
Learn More Link

----
#### NVDA Chrome is too verbose:
Tooltip label  button  collapsed  
Tooltip label Tooltip label
Tooltip label Tooltip label  dialog  This is some tooltip text. This box shows the maximum amount of text that should appear inside. If more room is needed please use a modal instead.
Learn More  visited  link  
Tooltip label Tooltip label
visited  link    Learn More
button    Create

#### NVDA FF is a bit too verbose:
button  collapsed    Tooltip label
expanded
Tooltip label  dialog  This is some tooltip text. This box shows the maximum amount of text that should appear inside. If more room is needed please use a modal instead.
Learn More  link  
Tooltip label
link    Learn More
button    Create

This suggestion in https://github.com/carbon-design-system/carbon/issues/6703#issuecomment-721868383 to only open the info "tip" on click will reduce the verbosity, but still need to try to reduce it down to only what is essential. I think the aria-labelledby on text outside of a button might be causing the extra verbosity. Usually a button's label is either its text content or an aria-label string - it's a bit strange to have an externally-labelled button, so that may be messing up the screen readers. Will have to try a few things, but first - what is the intended use case for the external label?

jnm2377:
Closes https://github.com/carbon-design-system/carbon/issues/7226

Makes screenreader for `Tooltip` less verbose. Found various examples that show accessible tooltips should be built, and almost all of them had the aria-describedby on the trigger element, and the actual "tooltip" content just served as the description, so this is what I went with here.

The one downside I can think of is that there wouldn't be a description if people don't include a `tooltipBodyId`, which currently isn't required. 

**Changed**
- moved `aria-describedby` from the tooltip content to the tooltip trigger (as seen in most examples)

**Removed**
- removed `role="tooltip"` per Carolyn's recommendation since we're already using `role="dialog"`
- removed `aria-labelledby` since we're already using `aria-describedby` (most documentation I found said to use one or the other to make it accessible for screen readers, but I haven't found any examples using both)

#### Testing / Reviewing
- Go to `Tooltip` story
- turn on voice over for testing
- tab to the tooltip icon, which should automatically open the tooltip dialog
- note that the tooltip dialog should be less verbose
    - compare to old dialog:
    - > Tooltip label button collapsed
Tooltip label Tooltip label
Tooltip label Tooltip label dialog This is some tooltip text. This box shows the maximum amount of text that should appear inside. If more room is needed please use a modal instead.
Learn More visited link
Tooltip label Tooltip label
visited link Learn More
button Create

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.js
+++ b/packages/react/src/components/Tooltip/Tooltip.js
@@ -505,12 +505,13 @@ class Tooltip extends Component {
       <>
         <ClickListener onClickOutside={this.handleClickOutside}>
           {showIcon ? (
-            <div id={triggerId} className={triggerClasses}>
+            <div className={triggerClasses}>
               {triggerText}
               <div
                 className={`${prefix}--tooltip__trigger`}
                 {...properties}
-                ref={refProp}>
+                ref={refProp}
+                aria-describedby={tooltipBodyId}>
                 <IconCustomElement {...iconProperties} />
               </div>
             </div>
@@ -519,7 +520,8 @@ class Tooltip extends Component {
               id={triggerId}
               className={triggerClasses}
               ref={refProp}
-              {...properties}>
+              {...properties}
+              aria-describedby={tooltipBodyId}>
               {triggerText}
             </div>
           )}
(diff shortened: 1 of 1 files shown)


dakahn (approved):
These changes have SERIOUSLY improved how this component speaks in JAWS on Chrome and Firefox. Wonderful work! Much less verbose and afaict more understandable 👏🏽👏🏽👏🏽 

(also ran AC against closed and open and seems fine)

joshblack (approved):

Issue #8555, opened 2021-05-03
Convert initial set of components to Sass Modules in `@carbon/styles`


joshblack:
This issue is for converting an initial set of components to Sass Modules to be authored in `@carbon/styles`. This work should involve any explorations needed for what is the ideal setup for authoring component files with sass modules and, ideally, should incorporate whatever we think is necessary in terms of tests.

These components include (but can change):

- [ ] Accordion #8798 
- [ ] Breadcrumb
- [ ] Link #8798
- [ ] List #8798
- [ ] Loading #8798

joshblack:
Some of the steps I've been following:

- Find the component SCSS file from `packages/components`
- Create an `_index.scss` file for this component over in `packages/styles/scss/components/<component>/_index.scss`
- Copy over on the `*.scss` files, ignore the JS/HBS/md files for the most part, into this folder
- Update the `_<component>.scss` file with the following changes:
  - Remove the `import-once` import
  - Remove the `exports` mixin that is used
  - Update `_index.scss` to `@use '<component>';` and include the mixin afterwards

```scss
@use 'example-component';

@include example-component.example-component;
```

- Update the imports of the file from `@import` to `@use`
  - Common examples
  - Duration tokens coming from `_motion.scss` and have a name change
  - `rem` and `em` helpers come from `../utilities/convert`
  - `carbon--spacing` -> `spacing`
  - `carbon--breakpoint` -> `breakpoint`
  - Update tokens to use CSS Custom Properties

---

Some style guide ideas:

**Ordering of imports**

When bringing in imports, the order should be:

- Any `@forward` rules`
- Any sass native imports (`@use 'sass:map')`
- Alphabetical order

PR #8651, opened 2021-05-11
feat(styles): add global utilities to @carbon/styles

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/8555

WIP of porting all of our global scss files needed to work on the 5 components outlined in #8555 

#### Changelog

**New**

- Each mixin the `helper-mixins` now lives in its own file inside of `utilities`
- `keyframes` file added, moved skeleton keyframes into this one
- `component-reset`, which differs from `reset`, as it is scoped to wherever it is called and not applied globally
- `visually-hidden` which allows us to use/offer a mixin to apply hidden styles, but keep text readable by screenreaders


packages/styles/scss/utilities/_text-truncate.scss
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright IBM Corp. 2021
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+.#{$prefix}--text-truncate--end {
```
joshblack: This is a good example of a file where we have both CSS classes we like to emit along with potentially mixins, kind of like visually hidden too, I think.

Do you think it'd be worth offering both mixins and CSS classes? Trying to figure out when would it be a good time to offer one or the other (or both) 🤔

tw15egan: 👍🏻  I believe I updated `visually-hidden` to be a mixin, so I can update these to be mixins as well

joshblack:
FYI @janhassel this PR is working on bringing in our utilities into a `scss/utilities` folder in the new package and thought you might be interested 👀  Part of the hope is to make these part of the Public API / documented so that they are easy to use when building custom components.

Whenever you get the chance, I'd love to hear what you think about this setup and/or if there are any utilities that you think are essential for building custom components

janhassel:
@joshblack @tw15egan This is great stuff, I'm excited to see this approach!

One thing that comes to mind immediately is that sass modules could be used to enforce the access scope better. As of today, even when a sass variable is annotated as `/// @access private`, you'll see teams using it directly anyways for various reasons. I'd recommend to [prefix all private members of a sass module with a hyphen or underscore as this will prevent it from being exposed when used](https://sass-lang.com/documentation/at-rules/use#private-members).

I'll definitely go through some projects and see if there are any recurring utlities that could be added here. 👍

joshblack:
@janhassel definitely, that will be the plan!

packages/styles/scss/utilities/_tooltip.scss
```diff
@@ -0,0 +1,372 @@
...
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+@use 'z-index';
+@use 'box-shadow';
+@use 'convert';
+
+// Does this need to be relative?
+@use 'type';
```
joshblack: Yeah, I think it will in order to resolve the type entrypoint unfortunately.

```suggestion
@use '../../type';
```

joshblack (approved):
Seems great to me once the import gets updated so the test passes 👀

andreancardona (approved):

dakahn (approved):

PR #8655, opened 2021-05-11
fix(TooltipDefinition): sync hidden state with other tooltip components

### LINKED ISSUE #8654: Definition tooltip text won't show up when focus
## Detailed description
There're a button and a definition tooltip on my screen. When I hover the mouse over the button, it automatically adds 'bx--tooltip--visible' to the button and 'bx--tooltip--hidden' to the tooltip. When I move the mouse out, 'bx--tooltip--visible' is removed from the button but 'bx--tooltip--hidden' is still there on the tooltip.
Same problem occurs when I focus the button by tab and then I tab to the tooltip, its text won't ever show up.

1. Focus or hover the button
2. Focus the tooltip (using tab)

Please take a look at this code sandbox: https://codesandbox.io/s/jolly-shannon-rwrbd?file=/index.js

## Additional information
Mouse over button
mouse over button

Mouse out button
mouse out button

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8654

Adds in similar logic to `TooltipDefinition` that is used in icon only buttons and tooltip icons

#### Changelog

**Changed**

- `focus` and `blur` handlers now function similarly to `TooltipIcon`
- Only toggles tooltips if it is the `hasIconOnly` variant of button

#### Testing / Reviewing

Added a test story under `TooltipDefinition`. Ensure focusing on both elements will not prevent the tooltip from appearing.


packages/react/src/components/Button/Button.js
```diff
@@ -53,7 +53,9 @@ const Button = React.forwardRef(function Button(
   const tooltipTimeout = useRef(null);
 
   const closeTooltips = (evt) => {
-    const tooltipNode = document?.querySelectorAll(`.${prefix}--tooltip--a11y`);
+    const tooltipNode = document?.querySelectorAll(
+      `.${prefix}--tooltip--a11y:not(.${prefix}--tooltip__trigger--definition)`
+    );
```
emyarod: I don't think this `:not` selector here is the ideal solution since you would also need to apply this to `TooltipIcon` in order to deal with the scenario where `TooltipIcon` and `TooltipDefinition` are on the same page

rather than adding this `:not` selector I think `TooltipDefinition` should have the same show/hide logic as `TooltipIcon` instead (unless there was a reason for it being excluded initially?)

tw15egan: Makes sense 👍🏻

tw15egan: @emyarod updated to toggle correctly. I didn't add the query selector as I feel like it's already a dirty pattern, and definition tooltips are not likely to be near other actionable icons where it would be getting in the way. Hovering over a button/tooltip icon will still close this tooltip, but hovering a tooltip definition will not close other tooltips.

joshblack (approved):

emyarod (approved):

PR #8656, opened 2021-05-11
fix(DataTable): expand + select padding

### LINKED ISSUE #8472: Data table: expand + select spacing
Update in this PR  https://github.com/carbon-design-system/carbon/pull/8080 caused the padding with expansion + selection to be inconsistent. There's now too much padding between the chevron and checkbox. 

image

In the PR we updated the click target area to be 48px wide. I think we need to reduce that to be 32px so we can get consistent padding and spacing back. 

New touch area should look like this
image

jnm2377:
Closes #8472 

#### Testing / Reviewing

Make sure the expandable + select example has updated padding:
    - expand button should be 32px wide now (with an additional 8px left padding in the parent cell)
    - selectable cell should also be 32px wide
    - the table cell following the select should have 8px left padding now
    - select and expand button focus should now match sizes


tw15egan (approved):
LGTM 👍 ✅

aagonzales (commented):
The spacing looks perfect now!

Just wondering if can we update the focus box height for the expansion chevron to be 32px tall for size small and up? It feels weird to tab through the tall chevron focus to the small checkbox focus. I know checkbox focus is coming from the component probably but it would be nice for visual consistency in tabbing if those focuses were the same. Also let me know if i should just open a new issue for this. 

image

packages/components/src/components/data-table/_data-table-core.scss
```diff
@@ -299,14 +299,19 @@
...
+    width: rem(32px);
     background: $layer-accent;
     transition: background-color $duration--fast-01 motion(entrance, productive);
   }
 
+  .#{$prefix}--table-column-checkbox
+    .#{$prefix}--checkbox:focus
+    + .#{$prefix}--checkbox-label::before {
+    //make checkbox focus larger to match expansion btn focus
+    outline-offset: rem(6px);
```
tw15egan, line 312: Do we need to make sure this is only scoped to larger variants? I think this will affect the xs / small variants (haven't tested yet)

jnm2377: ahhh good point 👍🏽

tay1orjones (commented):
Does the checkbox need more padding-right, or the text more padding-left? The checkbox doesn't look centered between the expansion chevron and the text like it does in the spec from the [issue](https://github.com/carbon-design-system/carbon/issues/8472#issue-864072839)

Screen Shot 2021-05-12 at 10 05 01 AM
Screen Shot 2021-05-12 at 10 05 18 AM

aagonzales:
Oh yeah good catch @tay1orjones looks like its 4px off. I think just add those 4px to padding-right of checkbox?
Top row code, bottom row design spec
image 

image

jnm2377:
@tw15egan @tay1orjones @aagonzales updated! I added a knob to the Development > example story so you can test the expansion + select spacing/focus in different sizes.

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

aagonzales (approved):
Ship it!

PR #8691, opened 2021-05-14
Node update to v14.17 LTS

sstrubberg:
This updates node to v14.17 LTS.

#### Changelog

**Changed**

- Updated CircleCI and Github Action to reflect v14.17
- Updated `node-sass` & `gulp-sass` dependencies across the project to allow build process to complete successfully.

#### Testing / Reviewing

Make sure this doesn't break our build!

Notes:

For some reason the yarn.lock is still retaining and trying to install the old node-sass: 4.13.1 package, even though no projects are using it.


.github/workflows/ci.yml
```diff
@@ -8,36 +8,36 @@ jobs:
...
       - uses: actions/checkout@v2
-      - name: Use Node.js 14.x
+      - name: Use Node.js 14.17
         uses: actions/setup-node@v2.1.5
         with:
-          node-version: '14.x'
+          node-version: '14.17'
       - name: Install dependencies
-        run: yarn install --immutable --immutable-cache
+        run: yarn install
```
joshblack: We should make sure to keep this as `yarn install --immutable --immutable-cache` for CI builds

.github/workflows/ci.yml
```diff
@@ -8,36 +8,36 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js 14.x
+      - name: Use Node.js 14.17
```
joshblack: I think these workflows should end up resolving to 14.17 so should be good to leave it as `14.x`

joshblack (changes requested):

.github/workflows/ci.yml
sstrubberg: If I'm just leaving everything at 14.x, what exactly am I changing here?

joshblack: I think the big thing is the `.nvrmc` change and then making sure both the repo runs/compiles on 14.17 and that the netlify previews work as expected 👀  (I think they use the .nvmrc file to pick up the version)

Totally get if it feels superfluous though! Don't sweat it if it doesn't feel like something you're interested in 👍

PR #8694, opened 2021-05-16
fix(Modal): Increase Overflow menu options z-index with modal open

### LINKED ISSUE #8693: OverflowMenuItem, used inside a Modal appears behind the Modal

- [ ] `carbon-components`

## Detailed description

When an overflow menu is used inside the modal, the menu options render behind the modal.

Issue is related to Modal, OverflowMenu components

The modal has a z-index of 9000, and the overflow menu item has a z-index of 6000, this is causing the overflow menu option to display behind the modal.

Chrome

Tried with version carbon-components-react 7.33.0, 7.35.0

## Additional information

- Screenshots or code

Screenshot 2021-05-13 at 12 53 54

- Notes

I found this issue, #6753, which is a similar issue, but for tooltip instead of overflow menu options. 

This was fixed in #6757, I can create a similar fix and create a PR.

sj-rai:
Closes #8693 

This PR fixes the issue of overflow menu options being displayed behind the modal, when the overflow menu is inside the modal, by increasing the z-index of the overflow-menu-options when modal is open.

I made the change similar to #6757, which was the same issue as this, but for tooltip instead of overflow menu.


sj-rai:

packages/components/src/components/modal/_modal.scss
```diff
@@ -375,6 +375,10 @@
     z-index: z('modal');
   }
 
+  .#{$prefix}--body--with-modal-open .#{$prefix}--overflow-menu-options {
+    z-index: z('modal');
+  }
```
tw15egan: Could we combine this with the above selector?

```suggestion
// z-index fix for #8693
  .#{$prefix}--body--with-modal-open .#{$prefix}--tooltip, 
  .#{$prefix}--body--with-modal-open .#{$prefix}--overflow-menu-options {
    z-index: z('modal');
  }
```

sj-rai: Yes, good idea. I'll combine it.

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #8188, opened 2021-03-25
fix(sketch): fetch default icon layer styles from IDL Sketch library

emyarod:
Closes #8127

This PR connects the default icon shared layer styles to the IDL V2 Sketch library instead of regenerating them in the document. The icons sync command is now no longer dependent on the colors sync command or `@carbon/colors` 

#### Changelog

**New**

- link icon shared layer styles to IDL V2 Sketch library

**Changed**

- decouple icons sync command from colors sync command
- use param objects for icon sync functions and helpers

#### Testing / Reviewing

Confirm that icon layer shared styles are linked to the IDL V2 Sketch library when syncing icons

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/6207179/carbon-elements.sketchplugin.zip)

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/6238013/carbon-elements.sketchplugin.zip)

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/6267500/carbon-elements.sketchplugin.zip)

PR #8700, opened 2021-05-17
refactor(sharedStyles): use param object instead of list of params

emyarod:
This PR adds in the refactoring from #8188 while it remains blocked. The `syncColorStyle`  function now uses a param object instead of a list of params. This helped with linting by triggering warnings with unused params while refactoring. It will also help in case new params need to be added in the future, since the order of the function params will be unimportant within the config object.

#### Testing / Reviewing

Confirm there are no regressions with the current plugin behavior since this change is 100% organizational

[carbon-elements.sketchplugin.zip](https://github.com/carbon-design-system/carbon/files/6495000/carbon-elements.sketchplugin.zip)

### DIFF
--- a/packages/sketch/src/commands/icons/shared.js
+++ b/packages/sketch/src/commands/icons/shared.js
@@ -57,6 +57,14 @@ function removeDeprecatedSymbolArtboards({ icons, sizes, symbolsPage }) {
   });
 }
 
+/**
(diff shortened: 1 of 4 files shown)


joshblack (approved):

laurenmrice (approved):
Looks good! No regressions

andreancardona (approved):

Issue #8556, opened 2021-05-03
Add docs for CSS Grid Usage


joshblack:
Since we're looking to get feedback on CSS Grid in our beta release, it's important that we have a baseline set of documentation to be available that would be expected when working with the grid. This documentation should include:

- [x] Adding a CSS Grid [story](https://react.carbondesignsystem.com/?path=/story/components-grid--auto-columns) to our storybook after #8554 
- [ ] ~Updating [Grid implementation](https://app.slack.com/www.carbondesignsystem.com/guidelines/2x-grid/implementation) page on our website after https://github.com/carbon-design-system/carbon-website/issues/2252~ this will be contained elsewhere, https://github.com/carbon-design-system/carbon/issues/9243
- [x] Updating [Grid.mdx](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Grid/Grid.mdx)
- [x] Updating the Elements grid example

PR #8702, opened 2021-05-17
docs(carbon-react): add storybook v6

joshblack:
Includes storybook v6 for the `@carbon/react` package. In order to get this working with hoisting in the monorepo, we had to inline the `storybook-preset-carbon` package in the `packages/react` folder 😞 

#### Changelog

**New**

- Add storybook v6 to `packages/carbon-react`

**Changed**

- Inline `storybook-preset-carbon` in `packages/react`

**Removed**

- `storybook-preset-carbon`

#### Testing / Reviewing

- Verify that the netlify previews succeed and deploy successfully
- Make sure `yarn storybook` works for `packages/react`
- Make sure `yarn storybook` works for `packages/carbon-react`


tw15egan (approved):
Both storybooks working well for me 👍🏻 ✅

tay1orjones (approved):
The existing stubbed out grid story in `carbon-react` doesn't use the right base grid class. Should be `bx--css-grid` instead of `bx--grid`, but this is unrelated to the changeset here and if I manually change the class in the inspector I can see the styles are applied 🎉 

image


I'll be updating the grid story soon anyway in #8556 so no hold up here

PR #8705, opened 2021-05-18
feat(datepicker): add onOpen callback

### LINKED ISSUE #8698: Add onOpen event handler to DatePicker
### Summary

Add `onOpen` event handler to React DatePicker to allow hooking into when the date picker opens.

### Justification

Developers may want to trigger some custom cose based on the calendar picker being opened. DatePicker already exposes an `onClose` event handler and `onOpen` is a handler offered by Flatpickr.

For our specific use-case, we have a legacy application that renders modal content inside an iframe. Then content can contain a form with a DatePicker. If the modal content happens to be shorter than the height of the DatePicker, then the DatePicker is not usable. If we can know when the picker is open, we can add additional height to the iframe temporarily, to allow the picker be usable, restoring the original height when it's closed.

Describe the full user experience for this feature. Also define the metrics by
which we can measure success/failure for the user.

### "Must have" functionality

`onOpen` event handler on React DatePicker

Do you want this work within a specific time period? Is it related to an
upcoming release?

_NB: The Carbon team will try to work with your timeline, but it's not
guaranteed. The earlier you make a request in advance of a desired delivery
date, the better!_

### Available extra resources

We have a small squad of developers available to contribute code to this.

tay1orjones:
Closes #8698 

This adds an `onOpen` callback prop that exposes flatpickr's `onOpen` to consumers while still calling the internal `onHook`.

#### Changelog

**New**

- datepicker: add onOpen callback

**Changed**

- datepicker-story: add onClose and onOpen callback props as actions

#### Testing / Reviewing

Datepicker should function the exact same as it did before with no regressions. The playground story should show action functions being called on open and close of the datepicker passing along the expected args.


emyarod (approved):
looks good to me

andreancardona (approved):

andreancardona:
@tay1orjones any clue what's going on with the CI checks?

tay1orjones:
@andreancardona There's an ongoing [incident](https://www.githubstatus.com/incidents/m16jzl31gnqt) with GitHub actions. It sounds like it should be coming back up soon

PR #4268, opened 2019-10-08
fix(tooltip): focus management with non-interactive elements

elizabethsjudd:
Closes #4232 

Tooltips with non-interactive elements as the content need to have focus displayed on the tooltip itself. The user should be able to click within a tooltip without it closing.

#### Changelog

**Changed**
- demo for tooltips to show how a tooltip with non-interactive elements behaves
- updated the styling for the heading which was removed during a bad merge
- updated the focus styles to work on the content element of the tooltip
- moved aria attributes from tooltip to content to prevent duplicate information read by screen readers on non-interactive tooltips and updated others for consistency 

#### Testing / Reviewing
Test all three demos under interactive tooltip for regression behavior. The changed behavior should be as described above on the middle demo.  Tooltips should be passing AVT 1, 2, and 3 after this PR is merged.

Issue #4232, opened 2019-10-04
Regression bug: Vanilla Tooltip focus management


elizabethsjudd:
I submitted these PRs (https://github.com/carbon-design-system/carbon/pull/3476 and https://github.com/carbon-design-system/carbon/pull/3148) which corrected a11y bugs on Tooltips in Carbon vanilla and in the latest version of Carbon this behavior is no longer working

## What package(s) are you using?

- [x] `carbon-components`
- [ ] `carbon-components-react`

## Detailed description

> Describe in detail the issue you're having.

 - Tooltips without interactive elements should focus the tooltip on open so that screen reader users have access to the content
- The user should be able to click inside the tooltip without it closing the tooltip

> Is this issue related to a specific component?

Interactive tooltip

> What did you expect to happen? What happened instead? What would you like to
> see changed?

- The tooltip (when there are not interactive elements within it) should receive focus
- The user could select content that was within the tooltip without it closing

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

10.6.4

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Watson Health Design Pattern & Asset Library. We have a release early Nov

## Steps to reproduce the issue

1. Open the tooltip in the demo below
2. the tooltip should receive focus (there is a white border that displays when it gets focus)

https://codepen.io/elizabethsjudd/pen/oNNNgpM?editors=1000

asudoh:
Hi @elizabethsjudd thank you for reporting - Would you want to send a PR with Karma test cases for those? Thanks!

elizabethsjudd:
@carbon-design-system/developers can you assign this issue to me and I'll make the fix and add tests for this issue.

elizabethsjudd:
@asudoh I'm running in to issues with a similar interaction as this bug I logged: https://github.com/carbon-design-system/carbon/issues/3835

on `track-blur.js` line 45, we need to know if the user clicked within the element and if so, we don't want to close the tooltip.  Right now we have no way of knowing that information from the `event` object.  Do you have any thoughts on how we could test for this use-case as I know Carbon doesn't want me to add a click event to the document (which is how WH handled the same interactions).

asudoh:
My knee-jerk reaction to that is the caller code `handleBlur(event)` can test if a certain element contains `event.target`. BTW having this question made me wonder if your draft PR hasn't fixed the issue yet...?

elizabethsjudd:
[...]

PR #8713, opened 2021-05-18
fix(Tooltip): prevent focus hijacking by other tooltips with focus trap

### LINKED ISSUE #8653: Tooltip has closing issue on Firefox

- [x] `carbon-components`

## Detailed description

Using Firefox and the Tooltip component, I found cases where clicking on other widgets on the page does not close the tooltip and cause confusion.

Tooltip

> see changed?

I expect the tooltip to be closed when clicked elsewhere.  That is working for Chrome, Safari and Edge.

Firefox 88.0.1

10.34

> should be aware of?

IBM Cloud
Some functions already in production.

Use Firefox browser:
1. Go to https://cloud.ibm.com/apidocs/hs-crypto#actiononkey
2. Scroll to "Request" and then "Request body" table with a dropdown and a "View" tooltip.
3. Click on "View".  The tooltip is opened.
4. Click on the left "one of" dropdown.    The dropdown does not open and the tooltip closed and reopened.

Second scenario
1. Go to https://dev.console.test.cloud.ibm.com/apidocs/cloudant#postbulkget.
2. Scroll to the Request body table which shows two Tooltip triggers.
3. Click on BulkGetQuery and the tooltip shows up.
3. Click on BulkGetAttsSinceQuery and notice both tooltips show up. Close outside would still open the tooltip once.

I only see these for firefox.

## Additional information
- Screenshots or code

https://user-images.githubusercontent.com/25369764/117868652-d7fd2800-b267-11eb-95be-2b8b76b73dc7.mov

https://user-images.githubusercontent.com/25369764/117868666-daf81880-b267-11eb-8746-9d77125b4554.mov

emyarod:
Closes #8653
related #7260
related #7319

#7260 reported focus hijacking when clicking from tooltip to tooltip and was resolved in #7319. However, this only addressed the use case where the tooltips contained no focusable children. Focus would be restored to a tooltip's trigger element on loss of focus and multiple tooltips with focus traps would conflict with each other, leading to situations where multiple tooltips could be open at once. The order of the focus events was different in Firefox, but this conflict of focus events occurred in all browsers. This PR updates tooltip focus logic so that pages with multiple tooltips (particularly ones with focusable children/focus traps) will not hijack focus from the already focused elements

#### Changelog

**New**

- (temporary) test story adapted from #7260

**Changed**

- tooltip focus logic on tooltip close

#### Testing / Reviewing

Verify that focus is applied to the expected elements when navigating pages with Carbon tooltips using the mouse or keyboard in all browsers, regardless of whether or not the tooltips contain focusable children


jnm2377:
It's working fine with mouse, but ran into an issue on keyboard, that if I tab into a non interactive tooltip then tab somewhere else, it stays open, even if I open another tooltip. With the interactive tooltip, I couldn't tab out of it bc it had the focus trap, so I had to press esc, which then closed both of the tooltips.

https://user-images.githubusercontent.com/32556167/118712418-616fa580-b7e6-11eb-98c3-1b7a6ee8d863.mov

emyarod:
this is the existing behavior and I believe also the expected behavior. these tooltips are more like toggletips in that they require explicit dismissal, so the tooltip with no interactive children needs an explicit <kbd>Esc</kbd> keypress or clicking on a non-child element to dismiss it (as opposed to the definition and icon tooltips which will dismiss on blur)

jnm2377:
I'm not sure I would call that the expected behavior. I think if it's not interactive, it behaves more like an icon tooltip, whose [expected behavior](https://carbondesignsystem.com/components/tooltip/usage/#icon-tooltip) is to be shown on hover and focus. So I would expect the non interactive one to close when tabbing to another element since it no longer has focus. Similar to

jnm2377:
I think if it's expected to work more like a toggle tooltip, then we need to add some sort of logic so that if the focus is on the "toggle" button, and the tooltip is shown, then you can't tab out of it until you press `esc` to close the tooltip, or if you can tab out of it, then the tooltip should close automatically.

jnm2377:
cc: @dakahn any thoughts? (see comments above)

emyarod:
I was referring to the behavior for the interactive tooltip above ("They persist [sic] until intentionally dismissed by clicking outside of the tooltip") which is meant to be different from the icon tooltip (remember the interactive tooltip's trigger doesn't require an icon)

If the expected behavior should be revised in that regard, I think it can be revisited by the design team. but in any case, given that it is not a regression and is not the topic of the referenced issues, I believe its scope lies outside of the current PR

joshblack (approved):

emyarod:
regarding the point about keeping focus on the trigger button if the tooltip does not contain a focusable element, I believe that is an a11y violation and would also prevent selection of text within the tooltip (e.g. #4268 #4232)

jnm2377:
> I was referring to the behavior for the interactive tooltip above ("They persist [sic] until intentionally dismissed by clicking outside of the tooltip") which is meant to be different from the icon tooltip (remember the interactive tooltip's trigger doesn't require an icon)
>
> If the expected behavior should be revised in that regard, I think it can be revisited by the design team. but in any case, given that it is not a regression and is not the topic of the referenced issues, I believe its scope lies outside of the current PR

Right. But if you're navigating via keyboard, you can't click outside of the tooltip. Moving focus is the equivalent of clicking outside of the tooltip for keyboard users, so shouldn't the non interactive one close? I thought the scope of this PR was to not have tooltips remain open if they lose focus (whether its via mouse or keyboard).

jnm2377:
> regarding the point about keeping focus on the trigger button if the tooltip does not contain a focusable element, I believe that is an a11y violation and would also prevent selection of text within the tooltip (e.g. #4268 #4232)

I'm not sure what you mean. I wasn't suggesting that.

emyarod:
this PR is resolving focus hijacking described in the referenced issues but it seems that you are describing keyboard navigation behavior which doesn't appear to be a regression WRT this PR or the referenced issues. on that topic, the dismissal action for keyboard users would be with <kbd>Esc</kbd>

jnm2377 (approved):

PR #8546, opened 2021-04-30
feat(UI icons): new icons march and april

conradennis:
Submitting new and updated UI icons from March and April 2021
all updates and additions approved by BXD

33 net new icons
4 updated icons
1 removed icon

updated categories.yml with additions
updated deprecated.yml with change
updated ui-icon-master.ai with additions and updates

**NOTE: I included the glyph status icons in the 32 folder but they are only to be sized at 16px and should not be scaled to other sizes. If this is an issue, happy to move them to another folder in the repo**

**EDIT: Moved the glyph icons out of the 32 folder**

#### Changelog

**New—33 total**

– svg/glyph--caution-inverted.svg
– svg/glyph--caution.svg
– svg/glyph--circle-fill.svg
– svg/glyph--circle-stroke.svg
– svg/glyph--critical.svg
– svg/glyph--incomplete.svg
– svg/glyph--square-fill.svg
– svg/glyph--undefined.svg
– 32/bookmark--add.svg
– 32/breaking-change.svg
– 32/connect--recursive.svg
– 32/connect--source.svg
– 32/connect--target.svg
– 32/connection--receive.svg
– 32/connection--send.svg
– 32/connection--two-way.svg
– 32/convert-to-cloud.svg
– 32/delivery--add.svg
– 32/folder--open.svg
– 32/folder--parent.svg
– 32/lifesaver.svg
– 32/location--filled.svg
– 32/parameter.svg
– 32/search--locate.svg
– 32/select--window.svg
– 32/text--font.svg
– 32/thumbs-down--filled.svg
– 32/thumbs-up--filled.svg
– 32/touch--1--filled.svg
– 32/touch--1-down--filled.svg
– 32/touch--1-down.svg
– 32/touch--2--filled.svg
– 32/zoom--fit.svg

**Changed—4 total**

– 32/data--base--alt.svg
– 32/thumbs-down.svg
– 32/thumbs-up.svg
– 32/zoom--area.svg

**Removed—1 total**

– 32/letter--Aa--large.svg (replaced with 32/text--font.svg)

PR #8724, opened 2021-05-19
fix(icon-build-helpers): support glyph-only assets

emyarod:
related #8546

This PR defaults to glyph assets in the icon builders when no default asset size is defined

copying my [comment](https://github.com/carbon-design-system/carbon/pull/8546#issuecomment-844479351) from the PR thread

> the breakage is because no other size is specified for these new glyphs, whereas the existing glyphs have at least one other size defined. should we just include another size for the new glyphs? if not then we need to update the builder to accommodate glyph-only assets, if that's desired


tw15egan:
It seems like in the originating PR, they were against adding sizes and wanted them to remain `glyph-only`, so I'm good with this fix! Thanks for working on this

joshblack (approved):

tw15egan (approved):

PR #8726, opened 2021-05-20
feat(Plex): add Plex package to carbon-react storybook

### LINKED ISSUE #8711: Add locale switcher support to storybook v6 in `@carbon/react`
Storybook v6 makes it easy to specify a locale for the storybook and makes it available in the context for a component.

For our IBM Plex integration, we are looking to add in an option for Arabic to storybook and wire it up so that we can add in styles for including the Arabic font faces from plex.

Most likely this will follow the same model as the theme in [`preview.js`](https://github.com/carbon-design-system/carbon/pull/8702/files#diff-c4c2336ccc3fca94a4eb4aae72cc3135e4b901fbebb964aea89fd78a2e1610c9R84-R94), something like:

```jsx
export const decorators = [
  (Story, context) => {
    const { locale, theme } = context.globals;

    React.useEffect(() => {
      document.body.setAttribute('data-carbon-theme', theme);
    }, [theme]);

    React.useEffect(() => {
      document.documentElement.lang = locale.value;
    }, [locale]);

    return <Story {...context} />;
  },
];
```

Then, you can target the styles with `html[lang="ar"]`. I don't think it's known how to bring in fonts quite yet, whether we load them up all at once or load them as-needed so that will be something the implementor will need to figure out.

### LINKED ISSUE #8712: Bring in IBM Plex font faces into storybook
As part of https://github.com/carbon-design-system/carbon/issues/8710 we would like to bring in multiple IBM Plex font faces into the storybook of `@carbon/react`.

This issue will follow #8711 which will add locale support to the storybook. Using this locale information, we should figure out how to appropriately:

- Load in font face declarations from IBM Plex without generating a large initial CSS bundle
- Correctly configure the `font-family` for the localized font

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8711
Closes https://github.com/carbon-design-system/carbon/issues/8712

Adds in the `@ibm/plex` package, as well as enhancing the `locale` switcher to allow a user to change the locale to `Arabic`. This can be used to test that the `IBM Plex Sans Arabic` is properly loaded in. 

#### Changelog

**New**

- a `Plex` story, that shows the 3 font styles currently supported in the `@ibm/plex` package (`Sans`, `Mono`, and `Arabic`)
- A `fontWeight` and `fontSize` switcher, to test the different weights (and increase `font-size` for easier testing)

#### Testing / Reviewing

In the storybook, use the locale switcher to switch between `en` and `ar` and ensure the correct fonts are loaded when viewing the `Arabic` story

### DIFF
--- a/packages/carbon-react/.storybook/styles.scss
+++ b/packages/carbon-react/.storybook/styles.scss
@@ -5,7 +5,31 @@
 // LICENSE file in the root directory of this source tree.
 //
 
+$feature-flags: (
+  enable-v11-release: true,
+);
+
+@use "~@ibm/plex/default" as sans;
+@use "~@ibm/plex/mono";
+@use "~@ibm/plex/arabic";
+
 @use '../index.scss' as styles;
+
+// For now, including all weights for testing
+@include arabic.all;
+// @include arabic.light;
+// @include arabic.regular;
+// @include arabic.semibold;
+
+@include sans.all;
+// @include sans.light;
+// @include sans.regular;
+// @include sans.semibold;
+
+@include mono.all;
+// @include mono.light;
+// @include mono.regular;
+// @include mono.semibold;
 
 body {
   @include styles.theme();
@@ -21,4 +45,17 @@ body {
 
 [data-carbon-theme='g100'] {
   @include styles.theme(styles.$g100);
+}
+
+html[lang='en'] body {
+  font-family: 'IBM Plex Sans', 'Helvetica Neue', Arial, sans-serif;
+}
+
+html[lang='ar'] body {
+  font-family: 'IBM Plex Sans Arabic', 'Helvetica Neue', Arial, sans-serif;
+}
+
+.text-mono {
+  font-family: 'IBM Plex Mono', Menlo, 'DejaVu Sans Mono',
+    'Bitstream Vera Sans Mono', Courier, monospace;
 }
--- a/packages/carbon-react/.storybook/styles.scss
+++ b/packages/carbon-react/.storybook/styles.scss
@@ -17,19 +17,8 @@ $feature-flags: (
 
 // For now, including all weights for testing
 @include arabic.all;
-// @include arabic.light;
-// @include arabic.regular;
-// @include arabic.semibold;
-
 @include sans.all;
-// @include sans.light;
-// @include sans.regular;
-// @include sans.semibold;
-
 @include mono.all;
-// @include mono.light;
-// @include mono.regular;
-// @include mono.semibold;
 
 body {
   @include styles.theme();
--- a/packages/carbon-react/.storybook/preview.js
+++ b/packages/carbon-react/.storybook/preview.js
@@ -24,6 +24,11 @@ export const globalTypes = {
           title: 'English',
           value: 'en',
         },
+        {
+          right: '🇵🇸',
+          title: 'Arabic',
+          value: 'ar',
+        },
       ],
     },
   },
@@ -83,11 +88,15 @@ configureActions({
 
 export const decorators = [
   (Story, context) => {
-    const { theme } = context.globals;
+    const { locale, theme } = context.globals;
 
     React.useEffect(() => {
       document.body.setAttribute('data-carbon-theme', theme);
     }, [theme]);
+
+    React.useEffect(() => {
+      document.documentElement.lang = locale;
+    }, [locale]);
 
     return <Story {...context} />;
   },
--- a/packages/carbon-react/index.scss
+++ b/packages/carbon-react/index.scss
@@ -5,4 +5,7 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-@forward '@carbon/styles';
+@forward '@carbon/styles' with (
+  $css--font-face: false,
+  $css--default-type: false,
+);
--- a/packages/carbon-react/package.json
+++ b/packages/carbon-react/package.json
@@ -44,7 +44,8 @@
     "@carbon/icons-react": "^10.33.0",
     "@carbon/styles": "^0.7.0",
     "@carbon/telemetry": "0.0.0-alpha.6",
-    "carbon-components-react": "^7.36.0"
+    "carbon-components-react": "^7.36.0",
+    "@ibm/plex": "6.0.0-next.3"
   },
   "devDependencies": {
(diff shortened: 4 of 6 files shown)


packages/carbon-react/.storybook/styles.scss
```diff
@@ -22,3 +46,16 @@ body {
 [data-carbon-theme='g100'] {
   @include styles.theme(styles.$g100);
 }
+
+html[lang='en'] body {
+  font-family: 'IBM Plex Sans', 'Helvetica Neue', Arial, sans-serif;
+}
+
+html[lang='ar'] body {
+  font-family: 'IBM Plex Sans Arabic', 'Helvetica Neue', Arial, sans-serif;
```
joshblack, line 44: @vpicone do we have font stacks that we could use for sans/arabic/etc? I know we define some over in: https://github.com/carbon-design-system/carbon/blob/main/packages/type/scss/_font-family.scss#L13 but was curious if you think this should be in the `@ibm/plex` package instead

packages/carbon-react/.storybook/styles.scss
```diff
@@ -22,3 +46,16 @@ body {
 [data-carbon-theme='g100'] {
   @include styles.theme(styles.$g100);
 }
+
+html[lang='en'] body {
+  font-family: 'IBM Plex Sans', 'Helvetica Neue', Arial, sans-serif;
```
joshblack, line 40: I think with our styles we might be able to do:

```suggestion
  font-family: styles.font-family('sans');
```

Totally get if it doesn't work now though, but we should definitely add it in through the `scss/_type.scss` entrypoint if not

joshblack (changes requested):
image

Hmm running into this for rendered fonts, even though the font-family one is working:

image

Am I missing a step potentially? 🤔

packages/carbon-react/.storybook/styles.scss
vpicone: @joshblack BoldMonday doesn't have any recommended fallback fonts outside of Latin. Might be worth exploring what other folks like Noto are doing.

joshblack: @vpicone I think the `font-family` rules for Mono/Sans/Serif are mentioned in the README of the project: https://github.com/IBM/plex#web-usage

Just was asking about codifying those in the sass itself since it seemed like they are a part of using the package on the web 👍

vpicone: @joshblack Totally, I was thinking about the non-english languages that aren't supported by fallback/web safe stack.

joshblack: @vpicone would we end up doing something like:

```css
system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif
```

? Seems to work for FB across the different languages, falling back to specific local files ultimately (I think)

tw15egan:
@joshblack we were seeing that in testing in Chrome, but Safari seemed to be rendering correctly. We'll need to dig deeper, not sure what the issue is

packages/carbon-react/.storybook/styles.scss
vpicone: Definitely, there's also some Web Safe fonts for each language that might be more appropriate matches, this comes into play when the font-display is swap and you get the flash of System text. But using the systems font is simple and safe.

tw15egan:
@joshblack it seems like it may be caused because the `@ibm/plex` font files are being installed into the _monorepo_ `node_modules` folder, and not the `@carbon-react`, so we may need to configure the storybook config to handle that use-case? I've tried fiddling with `file-loader` in `main.js` but not having any luck

joshblack:
@tw15egan if you add in `resolve-url-loader` into webpack does that help out? https://www.npmjs.com/package/resolve-url-loader

packages/carbon-react/.storybook/styles.scss
tw15egan: Should we include this via `@ibm/plex` or in the `@carbon/type`? Might be good to create a separate issue to track this enhancement as it doesn't currently exist

packages/carbon-react/.storybook/styles.scss
tw15egan: I'm down to just use system fonts as fallbacks! I just copied what we had in the readme for now so the linter would stop yelling at me for not declaring any fallbacks. Whatever we decide, we can then make it reusable as suggested in https://github.com/carbon-design-system/carbon/pull/8726#discussion_r637062234

packages/carbon-react/.storybook/styles.scss
joshblack: @tw15egan ooo good call, carbon type would totally be the way to go for this. Definitely get wanting to get this PR in and then we can tackle getting in carbon type as an enhancement 👍

tw15egan:
@joshblack I've pushed a bunch of fixes and created https://github.com/carbon-design-system/carbon/issues/8764 to track our `font-family` stack strategy. This PR demonstrates the ability of the `@ibm/plex` package to serve as a standalone package / delivery of the Plex font, and should give us a great path forward from moving away from Google Fonts

joshblack (approved):

vpicone (commented):
Hmm, looks like maybe font-face declarations might still be getting emitted? Do we want to fiddle with those error messages/feature flag logic in this PR?

Screen Shot 2021-05-25 at 6 36 43 PM

vpicone:
Also I can't find the locale switcher 🤦

tw15egan:
@vpicone I think we should tackle that deprecation warning in a separate PR on the `next` branch as we get ready to move forward with using the `@ibm/plex` package as a dependency. I believe that warning is coming from the `components` package. I'm only seeing that in the build process, not when running the `carbon-react` storybook locally. Are you seeing it there? 

Also, the locale switcher is the little globe icon at the top: 
Screen Shot 2021-05-26 at 10 28 59 AM

joshblack:
Yeah, I think you were looking at the netlify preview @vpicone which is the current stable storybook. The `@carbon/react` one doesn't have a deploy preview yet so you need to check it out locally 👍

vpicone:
@joshblack AH NUTS that was it. Thank you.

dakahn (approved):

PR #8735, opened 2021-05-21
feat(styles): add accordion styles and tests

joshblack:
Part of #8555

This converts our accordion styles to Sass modules and also updates some of the existing packages to make it work. There were some challenges that came up where it would be great to get feedback on:

- Should the `_theme.scss` module export all theme members, or just colors?
- How do we want to handle namespaces? In other words, when should we use `@use 'module-name'` versus `use 'module-name' as *` to bring members in to scope

#### Changelog

**New**

- Add `_tokens.scss` file to themes to map variables like `$interactive-01` over to their CSS Custom Property variant
- Add accordion component scss folder

**Changed**

- Update utilities to make the namespace import less awkward
  - For example, `@include button-reset.button-reset()` becomes `@include button-reset.reset()`

**Removed**

#### Testing / Reviewing

- Make sure the test for accordion passes in styles (which means the styles can compile and be configured)


joshblack:
cc @tw15egan would love your thoughts here 👀

dakahn (approved):

emyarod (approved):
looks good to me, and I believe a couple of snapshots will need to be updated

PR #8736, opened 2021-05-21
ci: update build to exclude sketch package

joshblack:
This PR consolidates our `node-sass` dependencies to only one version so it only compiles once. It also excludes `@carbon/sketch` from the build step in CI (but not when doing releases)

#### Changelog

**New**

**Changed**

- Pin node-sass to 6.0.0 in the project
- Add `--ignore '@carbon/sketch'` to build command in CircleCI

**Removed**

#### Testing / Reviewing

- Verify CI passes as expected (and hopefully is faster!)


tw15egan (approved):
Looks like it shaved ~5 minutes off the CI time 🎉

tay1orjones (approved):

PR #8752, opened 2021-05-24
fix(MultiSelect.Filterable): onMenuChange is not being called

### LINKED ISSUE #8749: Multiselect.Filterable onMenuChange is not being called when the menu closes by an outer click

- [ ] `carbon-components`

## Detailed description

I need to use the prop `onMenuChange` of the MultiSelect.Filterable component, but the function is not being triggered when  the user closes the menu by clicking outside.

MultiSelect.Filterable

The function be triggered with `false`, the function is not being called.

10.35

Matheus-Rangel:
Closes #8749 

Currently the function was relying on the inputBlur switch case to call the handleOnMenuChange, but that was not enough and most of the times when clicking outside the dropdown the onMenuChange prop was not being called.

Fixing the issue by Calling handleOnMenuChange  function inside the handleOnOuterClick.

#### Changelog
**Changed**
- Calling handleOnMenuChange  function inside the handleOnOuterClick instead of modifying the state directly
 
**Removed**
- The call of the handleOnMenuChange inside the inputBlue switch case to avoid duplicate false calls to the onMenuChange.


Matheus-Rangel:

tw15egan (approved):
LGTM, thanks for contributing a fix! 👍 ✅

emyarod (approved):
looks good to me

tay1orjones (approved):

PR #8760, opened 2021-05-25
feat(react): add disclosure hook

### LINKED ISSUE #8819: Implement initial `useDisclosure` hook
As part of our work on disclosure, we would like to start out by building a `useDisclosure` hook that encapsulates the [disclosure design pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure) from WAI-ARIA.

This hook should:

- Return the props necessary for the button and content parts of a disclosure pattern
- Should be used in an `unstable_Disclosure` story as an example
- Include tests to verify that the behavior matches the keyboard interaction and WAI-ARIA Roles/States/Properties required by the pattern

joshblack:
Closes #8819 

Related to: #6633

New component! It's Disclosure!

This is phase 1 of 5ish. Built out Disclosure with Text. Next up will be Menus!

Sketch file: 
- https://ibm.ent.box.com/folder/138835990231

**Testing/Reviewing:**

- Checkout the `unstable_Disclosure` component in Storybook and confirm the three stories work as expected.
  - Body (default)
  - Header and Body (with and without divider)
  - Body and Footer (with and without divider)
- Run the tests
  - `yarn test --watch packages/react/src/components/Disclosure/__tests__/useDisclosure-test.js`
  -  Confirm that all 5 tests pass.
- Confirm that we didn't break `unstable_Popover` in the process.


aledavila:

aledavila:
recheck

aledavila:
@joshblack Ready for review !

dakahn (changes requested):
Firstly, big fan of `Linky link`

Secondly, i'm able to get the disclosure to show up _on top of_ the tooltip. Which is a little awkward -- is that intentional?

sstrubberg:
> Firstly, big fan of `Linky link`
> 
> Secondly, i'm able to get the disclosure to show up _on top of_ the tooltip. Which is a little awkward -- is that intentional?

Firstly... thx ;)

Secondly, that is so weird. This whole time the alignment was just fine being top-left on popover. Not sure what changed there.

packages/icons-vue/examples/vue-cli/yarn.lock
```diff
@@ -1,7738 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
emyarod, line 1: is this file deletion intentional?

emyarod (commented):
is having the tooltip together with a disclosure the intended UX? also is the disclosure only dismissible by another click on the trigger button?

https://user-images.githubusercontent.com/8265238/122245240-2c10b300-ce8b-11eb-9ea8-37e40e6366be.mov

on a side note maybe we should increase the whitespace around the story or change the default positioning of the disclosure so that more of the component is visible without editing the page

andreancardona:
> Firstly, big fan of `Linky link`
> 
> Secondly, i'm able to get the disclosure to show up _on top of_ the tooltip. Which is a little awkward -- is that intentional?

It's from our changes 

> is having the tooltip together with a disclosure the intended UX? also is the disclosure only dismissible by another click on the trigger button?
> 
>  Screen.Recording.2021-06-16.at.10.08.58.mov 
> on a side note maybe we should increase the whitespace around the story or change the default positioning of the disclosure so that more of the component is visible without editing the page

TBD on the tooltip - we weren't sure also - waiting for a visual review on that. :)

packages/icons-vue/examples/vue-cli/yarn.lock
sstrubberg: Not entirely clear when that file was touched. Thanks noticing it though.

joshblack:
I think for the issue to close you can use: https://github.com/carbon-design-system/carbon/issues/8819 👀

joshblack:
Looking awesome! I think for the issue that we want to get in for this sprint, we can focus on getting in the files for `Disclosure/index.js` that has the hook, the test file for this hook, and some sample story that uses the hook and can bring in style updates / what-not in a future PR. What do you all think?

sstrubberg:
> Looking awesome! I think for the issue that we want to get in for this sprint, we can focus on getting in the files for `Disclosure/index.js` that has the hook, the test file for this hook, and some sample story that uses the hook and can bring in style updates / what-not in a future PR. What do you all think?

Sounds great @joshblack . I think we have all three of those in place currently. Just trying to get the popover menu to return to a hidden state when the user clicks anywhere else on the page. Do we have a hook for that anywhere?

joshblack:
> Just trying to get the popover menu to return to a hidden state when the user clicks anywhere else on the page

Yeah, definitely! You can use the `useOutsideClick` hook: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/internal/useOutsideClick.js

andreancardona:
Closing this for now and separating storybook styles and hook functionality into 2 different PRs

PR #8789, opened 2021-05-27
fix(breadcrumb): prevent wrapping to second line

### LINKED ISSUE #8779: [Breadcrumbs] Wrapping weirdly. Not behaving according to guidance
## Detailed description

Breadcrumbs wrapping to 3 lines when guidance on website says `it should never wrap to second line`. Seeing this bug downstream in Dotcom. 

https://react.carbondesignsystem.com/?path=/story/components-breadcrumb--breadcrumb

May-27-2021 09-26-25

image

tay1orjones:
Closes #8779

Prevent breadcrumbs from wrapping to a second line. I do have some reservations on this change request and would like some feedback:

The initial issue points this out, but the [breadcrumb guidance](https://www.carbondesignsystem.com/components/breadcrumb/usage/#overflow-content) on the webstie states

> Breadcrumbs should never wrap onto a second line.

@carbon-design-system/design How literally should we interpret this? I _think_ the intent is that when breadcrumbs would be too long and begin wrapping, consumers should instead switch to the "overflow menu mode".  I'm not sure if we should force breadcrumbs to stay on a single line like this and use `overflow: hidden` to ensure they don't resize when they hit the browser edge on smaller screen sizes.

#### Changelog

**Changed**

- breadcrumb: prevent items from wrapping to second line

#### Testing / Reviewing

Breadcrumb items should not wrap at small screen sizes. Try adding double/triple the amount of breadcrumb item `li`'s and see how it clips the ones that are beyond the browser's edge.

https://user-images.githubusercontent.com/3360588/119901559-ae046080-bf0b-11eb-8d14-307ba20ab4d0.mov

### DIFF
--- a/packages/components/src/components/breadcrumb/_breadcrumb.scss
+++ b/packages/components/src/components/breadcrumb/_breadcrumb.scss
@@ -21,8 +21,8 @@
     @include type-style('body-short-01');
 
     display: flex;
-    flex-wrap: nowrap;
     overflow: hidden;
+    flex-wrap: nowrap;
   }
 
   .#{$prefix}--breadcrumb-item {
--- a/packages/components/src/components/breadcrumb/_breadcrumb.scss
+++ b/packages/components/src/components/breadcrumb/_breadcrumb.scss
@@ -20,12 +20,9 @@
     @include reset;
     @include type-style('body-short-01');
 
-    display: inline;
-
-    @include carbon--breakpoint(md) {
-      display: flex;
-      flex-wrap: wrap;
-    }
+    display: flex;
+    flex-wrap: nowrap;
+    overflow: hidden;
   }
 
   .#{$prefix}--breadcrumb-item {


jnm2377:
Hmm... I have some initial questions.

- should we have guidelines as to how many breadcrumbs deep we recommend users going to (max limit)? We do this with tabs, which of course isn't the same kind of navigation, but bc of the visual style being similar... might be something that  might help prevent this sort of issue.
- if we do want to go with this hidden/scrolling behavior on smaller screens, would we want to add some sort of fade for when it overflows like the way we do with code snippet? To indicate that there's more content? 
- or...could we add that overflow menu functionality by default? I feel like I prefer this solution since it's already what we recommend.

emyarod:
bump @thyhmdo for a visual review and clarification on breadcrumb guidance!

thyhmdo:
@jnm2377 I thought we have a pretty clear guidance here on using overflow menu: "When space becomes limited, use an overflow menu to truncate the breadcrumbs. The first and last two page links should be shown, but the remaining breadcrumbs in between are condensed into an overflow menu. Breadcrumbs should never wrap onto a second line." Unless this doesn't work on mobile screen size. 

We don't have a thought of what number could be good for breadcrumbs right now, but similar to tabs (no more than 6) would be ideal. I think if it comes to user research and testing, the product teams need to see if it's viable for their users.

tw15egan:
I feel that unless we auto-convert the `Breadcrumbs` to the overflow menu variant on small screen sizes, they should break into multiple lines. Auto converting to the overflow menu variant is a much larger enhancement and is likely out of scope for this sprint.

tay1orjones:
@tw15egan Yeah, I agree. Adding the auto-convert to overflow on small sizes would be the ideal solution to this.

I'm good with keeping the existing behavior until we can take that on, in the meantime close this PR and use the contents as a guide for how teams can override the wrapping behavior. I'll update the original issue. 👍

PR #8797, opened 2021-05-28
fix(css-grid): various bugfixes, first docs pass

tay1orjones:
Part of #8556

First pass at CSS Grid stories and docs. Also a few bugfixes to the css grid implementation; mixed mode gutters, add aspect ratio, among others.

#### Changelog

**New**

- New stories covering various parts of CSS Grid
- CSS Grid mdx storybook docs
- Configured storybook breakpoint switcher to use CDS defined breakpoints

**Changed**

- CSS Grid styling for mixed mode gutters, max width, among other small fixes

#### Testing / Reviewing

Spin up the new storybook and check it out

https://user-images.githubusercontent.com/3360588/120017114-16564f00-bfab-11eb-8fe6-d7f70557ad2b.mov


packages/carbon-react/.storybook/main.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
             postcssOptions: {
               plugins: [
                 require('postcss-custom-properties')(),
-                require('rtlcss')(),
+                // require('rtlcss')(),
```
joshblack:
```suggestion
```

packages/grid/examples/css-grid/src/pages/layouts.js
```diff
@@ -1,5 +1,5 @@
 import Head from 'next/head';
-import { Tile, Button, UnorderedList, ListItem } from 'carbon-components-react';
+// import { Tile, Button, UnorderedList, ListItem } from 'carbon-components-react';
```
joshblack:
```suggestion
```

Might help for this diff to just remove any commented out code so we have a minimal working example without needing to come back and clean anything up in the future 👍

packages/grid/scss/modules/_css-grid.scss
```diff
@@ -35,6 +36,7 @@
   --cds-grid-columns: 4;
   --cds-grid-column-size: 1fr;
   --cds-grid-gutter: 2rem;
+  --cds-grid-gutter-wide: 2rem; /* used to configure appropriate margins for condensed subgrids inside wide grids */
```
joshblack: Style nit

```suggestion
  // Used to configure appropriate margins for condensed subgrids inside wide grids
  --cds-grid-gutter-wide: 2rem; 
```

joshblack (changes requested):
Looking great! Left a couple of QOL fixes/updates

One note for the span classes, feel free to update to have the behavior match CSS Grid instead of our old behavior if it makes sense. I think aligning with CSS Grid terminology instead of having our implementation try and map back to our v10 work makes a ton of sense 👍

tay1orjones:
@joshblack thanks! I revised the span/offset classes to match CSS Grid more appropriately, improved the offset story, and addressed the other feedback items 👍

joshblack (approved):

dakahn (approved):

PR #3054, opened 2019-06-13
feat(text-input): add read-only variant

emyarod:
Closes #3018

Closes #1672

This PR adds read-only support to the text input and textarea components alongside a mechanism to convert the inputs to controlled mode based on the `onChange` handler and `value` prop. This PR also reimplements character counter (which was previously reverted but depends on the controlled/uncontrolled separation). Since controlled/uncontrolled separation itself doesn’t provide any benefit from user’s perspective until character counter is introduced, it is being introduced alongside the character counter feature here

#### Changelog

**New**

- mechanism to turn `TextInput` and `TextArea` to controlled components based on `onChange` handler and `value` prop
- `readOnly` mode styles
- character counter for `TextInput` and `TextArea` (ref: https://github.com/carbon-design-system/carbon/pull/2745 which was previously reverted)

#### Testing / Reviewing

Ensure the read-only text input and character counters behave as expected

Issue #3018, opened 2019-06-11
Read only text input


emyarod:
ref: https://github.com/carbon-design-system/carbon/issues/2177 by @shixiedesign 

A read only variant of Text input is suggested here: https://github.com/carbon-design-system/carbon-contribution/issues/13#issue-421692303

> The “Disabled” option is not viable, as the label and entered data is not accessible. Proposal is to simply remove the line at the base of the "Filled" version of the component, so that it closely matches with the Disabled component, but passes accessibility.

## Overview of component (updated May 13)
- Read-only variant will be keeping underline and dropping the background of the field. We need the underline for low-vision users to recognize this as a text field. Also loosing the underline makes the element a button-style, which would be confusing. 
- Not moving the text position after all. Keeping text 16px away from left of border helps with preserving the form structure. 
- Regular text cursor and allows text selecting
- Add `not-editable` icon, icon in PR https://github.com/carbon-design-system/carbon/pull/2710
- Optional tooltip when hovering `not-editable` icon, to provide reason for the field being read-only 
- Overflow text in the field will show in tooltip on hover
- Should be considered a state, possible `readonly` attribute
- Follow the standard readonly attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly

## Design spec (updated May 13)

![image](https://user-images.githubusercontent.com/15144993/57665202-9970fc00-75c0-11e9-9a3f-b628d7b168a9.png)

## In context explorations: 
![image](https://user-images.githubusercontent.com/15144993/57496411-6fee6280-7298-11e9-82cc-71f2b96e0255.png)
![image](https://user-images.githubusercontent.com/15144993/57496415-75e44380-7298-11e9-8397-3a79fb5e9a75.png)

## Finalized design spec. Ready for dev 🤖 updated May 23

![Read-only 1](https://user-images.githubusercontent.com/15144993/58283670-77b90700-7d6e-11e9-8207-fab575af77fb.png)

### Light variant spec
Only difference from default is changing field background to `$field-01` so it is the same as background color. 

![Read-only 2](https://user-images.githubusercontent.com/15144993/58283648-6a038180-7d6e-11e9-8a04-f77b83f590c4.png)

kaylarakower:
@shixiedesign @emyarod Another proposal to consider (I created this issue prior to seeing @shixiedesign 's, so just some food for thought). The design was created for a slightly use case as none of the fields would be editable.

[...]

Issue #2177, opened 2019-03-21
Read only form controls


shixiedesign:
A read only variant of Text input is suggested here: https://github.com/carbon-design-system/carbon-contribution/issues/13#issue-421692303

> The “Disabled” option is not viable, as the label and entered data is not accessible. Proposal is to simply remove the line at the base of the "Filled" version of the component, so that it closely matches with the Disabled component, but passes accessibility.

## Overview of component (updated May 13)
- Read-only variant will be keeping underline and dropping the background of the field. We need the underline for low-vision users to recognize this as a text field. Also loosing the underline makes the element a button-style, which would be confusing. 
- Not moving the text position after all. Keeping text 16px away from left of border helps with preserving the form structure. 
- Regular text cursor and allows text selecting
- Add `not-editable` icon, icon in PR https://github.com/carbon-design-system/carbon/pull/2710
- Optional tooltip when hovering `not-editable` icon, to provide reason for the field being read-only 
- Overflow text in the field will show in tooltip on hover
- Should be considered a state, possible `readonly` attribute
- Follow the standard readonly attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly

## Design spec (updated May 13)

![image](https://user-images.githubusercontent.com/15144993/57665202-9970fc00-75c0-11e9-9a3f-b628d7b168a9.png)

## In context explorations: 
![image](https://user-images.githubusercontent.com/15144993/57496411-6fee6280-7298-11e9-82cc-71f2b96e0255.png)
![image](https://user-images.githubusercontent.com/15144993/57496415-75e44380-7298-11e9-8397-3a79fb5e9a75.png)

chrisconnors-ibm:
awaiting specification.

shixiedesign:
Another design possibly from Jordan https://github.com/carbon-design-system/carbon-contribution/issues/13#issuecomment-476725248

![image](https://user-images.githubusercontent.com/15144993/55974735-f12ce800-5c4d-11e9-94a7-03d0d59fdd9a.png)

laurenmrice:
not stale

lovemecomputer:
i have a couple of questions for implementation:

- 
- have we resolved/decided behavior

can one of the images above should be considered spec for implementation, or do we still need to define a spec?

lovemecomputer:
i have a couple of questions for implementation:

- **semantic:** if this is a unique state, should we think of it as being labelled `read-only`/`readOnly`?
[...]

PR #8806, opened 2021-05-31
feat(text-input): add readonly variant

janhassel:
Ref #2177, #3018, #3054

Adds a readonly variant to the TextInput component.

Intentional differences from the original design spec:
- No tooltip on the `EditOff16` icon
  - Looking at the planned changes of the tooltip component in v11, adding full support for that at this time seems inefficient. It would require the addition of multiple props that aren't really in the scope of the text input but rather of the tooltip (`tooltipAlignment`, `tooltipPosition`) and might introduce accessiblity issues that are to be addressed with the tooltip changes in v11.
  - In the meantime, teams can still inform their users about the reason why a field is read-only with a couple of options:
    - Use the helper text as explaination
    - Include a tooltip in the form label (quick poc with disabled field: https://jfckr.csb.app/)
- No tooltip on the entire input to show overflow text
  - Same as above
  - Users have the ability to "scroll" inside the field with both their mouse or their keyboard (arrow keys) to reveal overflowing text (native browser behaviour)

#### Changelog

**New**

- `props.readonly` on `TextInput`

#### Testing / Reviewing

- Ensure `readonly` attribute is set to `<input>` accordingly
- Ensure no visual regressions appear with any combination of `invalid`, `warn`, `readonly` on both the `TextInput` and the `PasswordInput` component
- Ensure that no error states are shown when a field is marked as readonly

Side note: this PR also adds a helper `useNormalizedInputProps` which aims to reduce some of the many conditional renderings that were introduced over time (e.g. `{!invalid && !warn && !isFluid && !inline && helper}`) by outsourcing them and providing a single entry for the icons and validation messages.
The intent is to add readonly support to other inputs as well (like `NumberInput`). This utility should help there as well and streamline the inputs. Let me know what you think!


packages/react/src/components/TextInput/TextInput.js
```diff
@@ -241,6 +238,11 @@ TextInput.propTypes = {
    */
   placeholder: PropTypes.string,
 
+  /**
+   * Whether the input should be read-only
+   */
+  readonly: PropTypes.string,
```
sstrubberg:
```suggestion
  readonly: PropTypes.bool,
```

image

☝️ When the PropType is set to string, I get the following in my console. When I turn it into a bool it goes away

sstrubberg (changes requested):
Read-only appears to take precedence over warn and invalid. Was that the intended outcome?

Also, the stuff below.

packages/react/src/components/TextInput/TextInput.js
janhassel: Thanks for catching that!

janhassel:
@sstrubberg Yes, when `readonly` is passed, both `invalid` and `warn` should be ignored. I based this on @shixiedesign's [comment from two years ago](https://github.com/carbon-design-system/carbon/pull/3054#pullrequestreview-251780808):
> Is it possible to disable "error state" on read-only? This situ shouldn't arise. It doesn't give user a way forward since field cannot be edited.

sstrubberg (approved):

packages/react/src/internal/useNormalizedInputProps.js
```diff
@@ -0,0 +1,59 @@
...
+import {
+  WarningFilled16,
+  WarningAltFilled16,
+  EditOff16,
+} from '@carbon/icons-react';
+import { settings } from 'carbon-components';
+
+const { prefix } = settings;
+
+export function useNormalizedInputProps({
```
tay1orjones, line 53: I think this is a good point of abstraction. Could you add a jsdoc-type comment for it so we have some context on what this hook accomplishes, when it should be used, expected param types, etc?

tay1orjones (changes requested):
When `disabled`, I think the icon should probably be modified to match text color?

image

Also how should we treat it with the `light` prop? Selecting between the readonly and light knobs there is no visual difference other than the icon. @carbon-design-system/design 

https://user-images.githubusercontent.com/3360588/120539919-b77b4600-c3ad-11eb-9d48-3cb1f0a041a5.mov

janhassel:
@tay1orjones Actually, now that you bring it up: `readonly` should probably take precedence over `disabled` / block `disabled` from being applied. Or can you think of a use case where a disabled read-only field would make sense?

packages/react/src/internal/useNormalizedInputProps.js
janhassel: Pushed an update adding some guidance aorund it. Let me know what you think of it!

johnbister (commented):
If read only takes precedence over disabled being applied, it looks fine visually. But if you think of a use case where a disabled read-only would be used, I would make the icon the same color as the disabled text like Taylor said. Let me know what you decide and I’ll approve.

johnbister:
'readonly' taking precedence over 'disabled' looks good for textinput! 

@janhassel I don't see the 'readonly' variant for the passwordinput.  Is this PR still adding the 'readonly' variant to both textinput and passwordinput? I apologize if I'm just missing something.

janhassel:
@johnbister My plan would be to add support for `readonly` step by step to more components. I think building out the ground work in this one and then apply it to others can help keep the PRs manageable.

johnbister (approved):
lgtm!

johnbister (approved):
lgtm!

janhassel:
@tay1orjones Regarding the `light` prop question: Afaik the design intent is to not have any background when the input is readonly, so I set it to `transparent`. Therefore there is no visual differenation between `light` and regular.

tay1orjones (approved):
Thanks for the follow up @janhassel , sounds good 👍

PR #8810, opened 2021-06-01
fix(button): fix icon color on disabled icon only ghost buttons

janhassel:
With the recent selector change in #8585 regarding the icon color on buttons, the specificity increased and caused disabled icon-only ghost buttons to use `$icon-primary` instead of `$icon-on-color-disabled`. This PR adds the same selector addition to the disabled icon-only ghost buttons.

Before
image

After
image

#### Changelog

**Changed**

- Increased css specificity for icon color fill on disabled icon-only ghost buttond

#### Testing / Reviewing

- Ensure the icon color on disabled icon-only ghost buttons matches `$icon-on-color-disabled`

### DIFF
--- a/packages/components/src/components/button/_button.scss
+++ b/packages/components/src/components/button/_button.scss
@@ -289,7 +289,7 @@
 
   .#{$prefix}--btn--ghost.#{$prefix}--btn--icon-only[disabled]
     .#{$prefix}--btn__icon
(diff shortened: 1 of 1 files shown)


tw15egan (approved):
LGTM 👍 ✅

joshblack (approved):

PR #8820, opened 2021-06-01
fix(Tooltip): avoid stealing focus when dismissing with mouseclick

### LINKED ISSUE #8180: Tooltip should not force focus after user clicks elsewhere on the page
- [x] `carbon-components`

## Detailed description / Steps:
My page has used the Tooltip component to show a "Breaking change" tooltip notice.
1. I click on the Tooltip trigger text and it shows the tooltip content inside.
2. I leave it opened and scroll down the page a bit and click on somewhere that I normally could click on, e.g., an accordion, a copy button, etc.
3. I noticed instead of the focus staying there, the page jumps back to where the Tooltip was and the focus is changed to the Tooltip trigger text.

Component:
Tooltip component in carbon-components-react

Expected result:
I would expect after user clicks anything else on the page, the focus should left the tooltip or tooltip trigger.

Browser:
I use Chrome primarily but I can reproduce on FF and Safari too.

Version of the Carbon Design System:
10.30 but I think same in 10.31

Product: IBM Cloud APIDocs page

Additional information:

https://user-images.githubusercontent.com/25369764/112367868-d88a3100-8cb0-11eb-9398-82a5ce65f818.mov

emyarod:
Closes #8180

This PR updates the interactive tooltip's user initiated open/close logic to avoid stealing focus after dismissal via mouse click

#### Testing / Reviewing

Confirm that the interactive tooltip trigger does not steal focus only after the tooltip is dismissed by user click

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.js
+++ b/packages/react/src/components/Tooltip/Tooltip.js
@@ -300,7 +300,11 @@ class Tooltip extends Component {
       if (!open && tooltipBody && tooltipBody.id === this._tooltipId) {
         this._tooltipDismissed = true;
         const currentActiveNode = event?.relatedTarget;
-        if (!currentActiveNode && document.activeElement === document.body) {
+        if (
+          !currentActiveNode &&
+          document.activeElement === document.body &&
+          event?.type !== 'click'
+        ) {
           this._triggerRef?.current.focus();
         }
       }


tay1orjones (approved):
Looks good 👍  Focus isn't stolen on close via click

https://user-images.githubusercontent.com/3360588/120567290-0093c000-c3d7-11eb-953e-c2441c84dcd0.mov

emyarod:
bump @dakahn when you have a chance!

dakahn (approved):

PR #8829, opened 2021-06-02
fix(menu): render through portal to avoid z-index and overflow issues

janhassel:
Renders the `Menu` component through a portal to avoid it being misplaced, cut off or invisible in certain setups (such as the DataTable toolbar).

#### Changelog

**Changed**

- render `Menu` through `ReactDOM.createPortal()`

#### Testing / Reviewing

- Ensure keyboard navigation still works as expected
- Ensure the position of the menu is corretl attached to the trigger point
- Ensure that when the menu is closed through `blur`, the focus is returned to the trigger element (if any exists)
  - I added a test story "unstable_menu > OverflowMenu > In Toolbar" to test this.
  - **The story should be removed prior to merging and only serves as an example setup to illustrate the purpose of this PR**


joshblack:
@janhassel in the past we got tripped up with portals (specifically the AC violation that content should be rendered in a landmark) and I was curious if this approach handles the use-case where screen reader uses can exit the menu with their virtual cursor and end up in a part of the document that they don't expect?

For an example, I wrote a bit about this over in: https://github.com/carbon-design-system/carbon/issues/7565#issuecomment-786921600 and just was curious if this approach was using an `aria-owns` trick or something to get around that 👀

janhassel:
@joshblack Good point! I only thought of the focus management so far but not about the virtual screen reader cursor. Just pushed an update with `aria-owns` and it seems to work great (at least with VoiceOver on my end). Only caveat is that `aria-owns` creates a parent/children relationship and not an adjacent one, so I had to wrap the trigger button in a container for it to work:

```jsx
<div aria-owns="menu-id">
  <button>...</button>
</div>
```

I don't think this should cause any issues though as I set `display: inline-block` on the container div to keep the flow predictable.

Could you check if the update works for you?

sstrubberg (dismissed):
Keyboard commands are firing like expected, but when I click on the menu, then click it again to close it, it doesn't close.

Jun-04-2021 10-02-21

janhassel:
@sstrubberg Good catch! Should be fixed now.

tay1orjones (commented):
Should focus return to the trigger element after an item is selected in the menu? Here I select one and then hit tab, which places focus on the first tab stop. 


https://user-images.githubusercontent.com/3360588/121713487-8a601f00-caa2-11eb-9113-b73aacfa235f.mov

EDIT - I'm also I'm seeing the same with the voiceover virtual cursor. Here I start on the default OverflowMenu - I select an option from the list, hit tab, focus is returned to the trigger element. On the Menu "In Toolbar" story it goes back to the first tab stop instead of the trigger element

https://user-images.githubusercontent.com/3360588/121714913-f0997180-caa3-11eb-829c-41a848904ba4.mov

janhassel:
@tay1orjones Apologies for the long delay! I just pushed an update which returns the focus to the trigger (if any) whenever the menu is closed. It should work with both mouse and keyboard operation (including VoiceOver controls).

joshblack:
Just a quick note, before merging we should give @janhassel a heads up so he can toggle off the demo story that he wrote 👀

tay1orjones (approved):

joshblack (approved):

PR #8833, opened 2021-06-02
feat(Storybook): upgrade to v6

tw15egan:
Starts work on upgrading the `react` storybook to v6 (finally)

#### Changelog

**Changed**

- Upgraded Storybook to v6
- Changes to `@storybook/addon-actions` in `Tag.js`

**Removed**

- `@storybook/addon-notes`

#### Testing / Reviewing

Still to do:
- [ ] Issue with `@babel/plugin-proposal-private-property-in-object` and the babel config


tw15egan:
@joshblack take two (or three)! I can actually get the storybook to compile but it is throwing a `babel` warning quite excessively. I've tried a few workarounds in `.storybook/babelrc.js` but I'm not able to get it working. Feel free to push up any fixes you can think of 👍

joshblack:
@tw15egan it seems like it's an issue that they're looking to address in 6.3, there are some workarounds over in: https://github.com/storybookjs/storybook/issues/14805 in the interim, or we could try out the prerelease/wait for 6.3 to be released 👀

PR #8844, opened 2021-06-04
feat(data-table): add data-table to sass modules style package

tay1orjones:
Refs #8555

Begins work on bringing `DataTable` to the `@carbon/styles` package

#### Changelog

**New**

- include refactored core and partial files added to `/styles`, including a test

#### Testing / Reviewing

Test is failing for me locally, but I'm not sure why. 🤔  Any suggestions @joshblack? 

```
Undefined mixin.
       ╷
    11 │ @include data-table.data-table;
```


packages/styles/scss/components/__tests__/data-table-test.js
```diff
@@ -0,0 +1,37 @@
...
+       $_: get('mixin', meta.mixin-exists('data-table-core', 'data-table-core'));
+       $_: get('variables', map.keys(meta.module-variables('data-table-core')));
+
+       $_: get('mixin', meta.mixin-exists('data-table-expandable', 'data-table-expandable'));
+       $_: get('variables', map.keys(meta.module-variables('data-table-expandable')));
+
+       $_: get('mixin', meta.mixin-exists('data-table-sort', 'data-table-sort'));
+       $_: get('variables', map.keys(meta.module-variables('data-table-sort')));
+     `);
+    expect(unwrap('mixin')).toBe(true);
```
joshblack, line 83: I think we might need to have a couple of tests for data table, one for each file we'll want teams to be able to import.

For the public API test, I think we'll only need two calls: one for mixins one for variables. The `meta.module-variables` should take in `data-table` as the argument since that's the namespace being loaded in.

We should also have similar style tests for each file we're expecting people to be able to bring in together 👀

packages/styles/scss/components/data-table/_data-table-action.scss
```diff
@@ -0,0 +1,586 @@
...
+@use '../../theme' as *;
+@use '../../type' as *;
+@use '../../utilities/button-reset';
+@use '../../utilities/convert' as *;
+@use '../../utilities/focus-outline' as *;
+
+/// Data table action styles
+/// @access private
+/// @group data-table
+@mixin data-table-v2-action {
```
joshblack: Just a refactor note for any v2 references we still have leftover 👀 

```suggestion
@mixin data-table-action {
```

packages/styles/scss/components/data-table/_data-table-action.scss
```diff
@@ -0,0 +1,586 @@
...
+@use '../../motion' as *;
+@use '../../spacing' as *;
+@use '../../theme' as *;
+@use '../../type' as *;
+@use '../../utilities/button-reset';
+@use '../../utilities/convert' as *;
+@use '../../utilities/focus-outline' as *;
+
+/// Data table action styles
+/// @access private
```
joshblack: These private annotations will need to go away unless the mixin itself is private

packages/styles/scss/components/data-table/_data-table-action.scss
```diff
@@ -0,0 +1,586 @@
...
+
+  .#{$prefix}--toolbar-search-container-persistent
+    .#{$prefix}--search-close::before {
+    display: none;
+  }
+
+  //-------------------------------------------------
+  //TOOLBAR BUTTONS
+  //-------------------------------------------------
+  .#{$prefix}--overflow-menu.#{$prefix}--toolbar-action {
```
joshblack: Ah, didn't think about this but will we need specific component dependencies in for data table before we are able to land data table?

packages/styles/scss/components/data-table/_data-table-inline-edit.scss
```diff
@@ -0,0 +1,65 @@
...
+//
+
+@use '../../config' as *;
+@use '../../spacing' as *;
+@use '../../theme' as *;
+@use '../../utilities/button-reset';
+@use '../../utilities/convert' as *;
+@use '../../utilities/focus-outline' as *;
+
+.#{$prefix}--inline-edit-label {
```
joshblack: We might be able to ditch this style sheet for now, I'm not sure where this gets used 🤔  Do you happen to know @tw15egan ?

packages/styles/scss/components/data-table/_data-table-skeleton.scss
```diff
@@ -0,0 +1,67 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+
+@use '../../config' as *;
+@use '../../spacing' as *;
+@use '../../theme' as *;
+@use '../../utilities/convert' as *;
+@use '../../utilities/skeleton' as *;
+
+.#{$prefix}--data-table.#{$prefix}--skeleton th {
```
joshblack: We should make sure to wrap this in some kind of mixin, maybe `data-table-skeleton` or `skeleton`?

packages/styles/scss/components/data-table/_data-table-sort.scss
```diff
@@ -0,0 +1,240 @@
...
+@use '../../config' as *;
+@use '../../motion' as *;
+@use '../../spacing' as *;
+@use '../../theme' as *;
+@use '../../utilities/button-reset';
+@use '../../utilities/convert' as *;
+@use '../../utilities/focus-outline' as *;
+
+/// Data table sort styles
+/// @access private
```
joshblack: For mixins that aren't hypen-prefixed we'll have to treat them as public unless we're saying we don't want teams to include this file directly 👀 

```suggestion
/// @access public
```

packages/styles/scss/components/data-table/_data-table.scss
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@use 'data-table-action';
```
joshblack: Along with `@use` I think we'll need to `@forward` any variables for modules that teams could potentially configure.

packages/styles/scss/components/data-table/_mixins.scss
```diff
@@ -0,0 +1,26 @@
+// @access public
```
joshblack, line 9: Might need a copyright banner at the top of this file

joshblack (changes requested):

packages/styles/scss/components/data-table/_data-table-inline-edit.scss
tw15egan: I'm not too sure, I think it got ported over to Cloud PAL. It should be able to be safely removed

packages/styles/scss/components/data-table/_data-table-action.scss
tay1orjones: Fixed 👍

packages/styles/scss/components/data-table/_data-table-action.scss
tay1orjones: Fixed 👍

packages/styles/scss/components/data-table/_data-table-action.scss
tay1orjones: Yeah, we'll need those before we add DataTable to the new `/carbon-react` storybook. I don't think we need them to land this port over to sass modules though?

packages/styles/scss/components/data-table/_data-table-inline-edit.scss
tay1orjones: Awesome! I removed it.

packages/styles/scss/components/data-table/_data-table-sort.scss
tay1orjones: Fixed 👍

packages/styles/scss/components/data-table/_data-table.scss
tay1orjones: Yep, thanks for catching this. I found 5 variables but 4 of them have no usage when I find all - I'll make note of them in a review, I think we can delete them

packages/styles/scss/components/data-table/_mixins.scss
tay1orjones: Fixed 👍

packages/styles/scss/components/data-table/_data-table-skeleton.scss
tay1orjones: Fixed 👍

tay1orjones:
@joshblack Thanks for the review! It prompted me to reorganized this quite a bit. 

I removed `_data-table-core.scss` and instead moved these core styles to be within `_data-table.scss`. The feature modules (action, expandable, skeleton, sort) moved to their own folder, each with an `_index.scss`. I kept the mixin names the same. You can see in the test file, but usage would look like:

```scss
@use '../data-table';
@use '../data-table/skeleton';

.my-datatable {
  @include data-table();
}

.my-datatable-skeleton {
  @include data-table-skeleton();
}
```

The goal being that users can bring in only the core styles without having to bring in the feature files if they don't need them. I did modify `packages/styles/scss/components/_index.scss` to import each of the feature files though so it should all be included in the centralized/main bundle. Tests for all this are updated and should be passing.

packages/styles/scss/components/data-table/_vars.scss
```diff
@@ -0,0 +1,37 @@
...
+// /// @type Number
+// /// @access public
+// /// @group data-table
+// $data-table-row-height: 2rem !default;
+
+// TODO - REMOVE? FIND ALL FINDS NO USAGES
+// /// @type Color
+// /// @access public
+// /// @group data-table
+// $data-table-zebra-color: $layer-accent !default;
```
tay1orjones: I couldn't find usage of these variables in a find all in the entire monorepo. Safe to remove?

joshblack: Do itttt

If you can add it to our v11 migration file for carbon-components that'd be awesome (just flagging them as removed)

joshblack (approved):

packages/styles/scss/components/data-table/_vars.scss
tay1orjones: @joshblack Sweet! Removed and updated the migration doc

dakahn (approved):

PR #8856, opened 2021-06-07
fix(components,themes): move theme fallback to themes package

### LINKED ISSUE #8838: Add compatibility for theme mixin to map v10 to v11 tokens
We currently provide a mapping between v10 and v11 tokens in our [`theme-tokens.scss`](https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/globals/scss/_theme-tokens.scss) file. For teams that may be using `carbon--theme` directly, whether from another Sass import or from `@carbon/themes`, this mapping is unavailable to them and their theming strategy no longer works.

This issue is to add an interop/compat mixin to the `carbon--theme` mixin directly. This should provide mappings from v10 token values to v11 tokens. Since `carbon--theme` is our current recommendation/practice for applying themes, this should allow existing code to work as intended with components in Carbon using the new tokens under the hood.

Another approach we could take is to toggle which tokens are used based on the feature flag being set.

joshblack:
Closes #8838 

This PR moves our logic for adding fallback theme token values from the components package to the themes package.

#### Changelog

**New**

**Changed**

- Update themes generated mixin step to include component token mappings for new tokens being used in components

**Removed**

- Removed token mappings from components package

#### Testing / Reviewing

- Run the storybook and verify that the token values are correct in the component
- View the netlify preview and make sure custom properties work as expected


tw15egan (approved):
Verified locally that all existing functionality still works 👍🏻 ✅

dakahn (approved):

PR #8863, opened 2021-06-08
fix(notification): include displayname wrapper to prevent enzyme selector failure

tay1orjones:
These conditional exports for v11 cause duplicate component instances which can cause downstream consumers' tests to fail if they're using enzyme selectors. 

This will not fix downstream tests that rely on constructor selectors. Recommendation is that constructor selectors not be used in favor of more stable selectors that do not rely on the react tree hierarchy such as displayname, data-testid, or similar.

#### Changelog

**Changed**

- add displayname for notification and structured list conditional exports

#### Testing / Reviewing

In the react dev tools you should see that the `v11Switch` annotation is present on the outermost instance of the associated components
image


joshblack:
@tay1orjones what do you think about adding in a custom component creator for stuff like this? Maybe something like:

```js
function createComponentToggle(spec) {
  const { name, flag = 'enable-v11-release', next, classic } = spec;

  function ComponentToggle(props) {
    const enabled = useFeatureFlag(flag);
    if (enabled) {
      return React.createElement(next, props);
    }
    return React.createElement(classic, props);
  }

  ComponentToggle.displayName = `FeatureToggle(${name})`;

  return ComponentToggle;
}

// Usage, ToastNotification.js
const ToastNotification = createComponentToggle({
  name: 'ToastNotification',
  next: ToastNotificationNext,
  classic: ToastNotificationClassic,
});
```

tay1orjones:
@joshblack Totally! I thought as I made the initial changes "there's an abstraction to be made here" 😄 

I added the `createComponentToggle` amended to support `forwardRef` (double check my method there please, I _think_ I've got it right)

I discovered that Toggle and OverflowMenu were using similar next/classic toggles, so I refactored those to use this too.

tay1orjones:
Also worth noting that it appears the root problem of double wrappers only happens with named exports, OverflowMenu and Toggle are defaults and don't have the double wrap like the named exports of StructuredList and Notification components.

joshblack (approved):

dakahn (approved):

PR #8869, opened 2021-06-09
docs(Popover): update docs with more clear usage guidelines

dakahn:
Closes #

Adds more descriptive and thorough usage docs for Popover 🏄🏽

#### Testing / Reviewing

Writing usage docs for primitives is kinda funky so make sure the way we're talking about stuff makes sense here.

### DIFF
--- a/packages/react/src/components/Popover/Popover.mdx
+++ b/packages/react/src/components/Popover/Popover.mdx
@@ -20,8 +20,7 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
 **Note: this component is currently experimental and does not follow semver.**
 
 A popover is a UI element that _pops over_, or renders on top of, the UI on a
-page. It's a component "primitive" meaning it can be used to help make more
-complex components that utilize this behavior. It is often styled as what we'd
+page. It's a component primitive that is designed to be used alongside other components to build out complex interactions. It is often styled as what we'd
 traditionally consider a tooltip.
 
 ## Usage 
--- a/packages/react/src/components/Popover/Popover.mdx
+++ b/packages/react/src/components/Popover/Popover.mdx
@@ -31,6 +31,7 @@ import { popover } from 'carbon-components-react';
 
 function MyPopover() {
   return (
+    <TriggeringUiElement/>
     <Popover 
       open={open} 
       align={top} 
@@ -45,20 +46,25 @@ function MyPopover() {
 
 <Props />
 
-## Popover Alignment 
+## Caret Alignment 
 
-A popover can be aligned relative to its triggering element in the following directions:
+Users can describe the caret alignment in one of twelve directions. Caret alignement dictates position and direction of the popover.
 
-```
-    'top-left'            'top'        'top-right'
-               ___________________
-              |                   |
-      'left'  |  Trigger element  |  'right'
-              |___________________|
+- 'top',
+- 'top-left',
+- 'top-right',
 
-'left-bottom'        'bottom'      'right-bottom' 
+- 'bottom',
+- 'bottom-left',
+- 'bottom-right',
 
-```
+- 'left',
+- 'left-bottom',
+- 'left-top',
+
+- 'right',
+- 'right-bottom',
+- 'right-top',
 
 ## Feedback
 
--- a/packages/react/src/components/Popover/Popover.mdx
+++ b/packages/react/src/components/Popover/Popover.mdx
@@ -19,12 +19,46 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
 
 **Note: this component is currently experimental and does not follow semver.**
 
-The Popover component is used to build layers that appear on top of the current
-surface.
+A popover is a UI element that _pops over_ or renders on top of other UI on the
+page. It's a component "primitive" meaning it can be used to help make more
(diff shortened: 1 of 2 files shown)


tw15egan (approved):

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
-surface.
+A popover is a UI element that _pops over_ or renders on top of other UI on the
+page. It's a component "primitive" meaning it can be used to help make more
+complex components that utilize this behavior. It is often styled as what we'd
+traditionally consider a tooltip.
+
+## Usage 
+
+```jsx
+import { popover } from 'carbon-components-react';
```
joshblack:
```suggestion
import { Popover, PopoverContent } from 'carbon-components-react';
```

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
 
 **Note: this component is currently experimental and does not follow semver.**
 
-The Popover component is used to build layers that appear on top of the current
-surface.
+A popover is a UI element that _pops over_ or renders on top of other UI on the
```
joshblack:
```suggestion
A popover is a UI element that _pops over_, or renders on top of, the UI on a
```

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
+page. It's a component "primitive" meaning it can be used to help make more
+complex components that utilize this behavior. It is often styled as what we'd
```
joshblack:
```suggestion
page. It's a component primitive that is designed to be used alongside other components to build out complex interactions. It is often styled as what we'd
```

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
+traditionally consider a tooltip.
+
+## Usage 
+
+```jsx
+import { popover } from 'carbon-components-react';
+
+function MyPopover() {
+  return (
+    <TriggeringUiElement/>
```
joshblack:
```suggestion
    <TriggerElement />
```

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
+    <Popover 
+      open={open} 
+      align={top} 
+      relative>
```
joshblack, line 37:
```suggestion
     >
```

We might need to figure out what to do with this `relative` thing since I think we need the container to be relative in this case 🤔

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
+      <PopoverContent className="custom-classname"/>
```
joshblack, line 38: Would it make sense to indicate that the content here is what is rendered in the popover/tooltip container?

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
 
 ## Component API
 
 <Props />
 
+## Caret Alignment 
+
+Users can describe the caret alignment in one of twelve directions. Caret alignement dictates position and direction of the popover.
+
+- 'top',
```
joshblack, line 52: Instead of a list, could we instead give a couple of examples of configuration?

Something like:

> For example, when you use `align="top"` the Popover component will align to the top of the trigger element. Likewise, `align="bottom"` will align the Popover component to the bottom of the trigger element.
>
> You can also configure the caret position by using `top-left` or `top-right`, and `bottom-left` or `bottom-right`. This will place the caret on either the left or right side of the Popover. By default, the caret will render in the middle of the Popover.
>
> These rules also apply to `left` and `right` alignment, but instead you will. use `left-top` and `left-bottom`, or `right-top` and `right-bottom`, to specify the alignment of the caret in the Popover.

packages/react/src/components/Popover/Popover.mdx
```diff
@@ -19,13 +19,53 @@ import { Preview, Props, Story } from '@storybook/addon-docs/blocks';
...
+- 'bottom-right',
+
+- 'left',
+- 'left-bottom',
+- 'left-top',
+
+- 'right',
+- 'right-bottom',
+- 'right-top',
+
```
joshblack, line 67: Would it be helpful to talk about no caret support? 🤔

joshblack (changes requested):
Looks great! Left a couple of content notes, let me know what you think 👀

packages/react/src/components/Popover/Popover.mdx
andreancardona: yes! I'll make a note*

PR #8893, opened 2021-06-11
fix(DatePicker): avoid aggressive range input clearing

### LINKED ISSUE #8892: Range DatePicker clears all inputs when end date field is empty

- [ ] `carbon-components`

## Detailed description

When one of the two input fields in the range DatePicker is cleared, the component will clear both input fields and remove any existing data entered

Date picker

> see changed?

rather than clearing all entered data, the component should allow users to set an invalid state if it detects that the range inputs are not completely filled. attempting to move focus from the end date input field should not trap focus on the end date input for an additional tab stop

1. Enter a start date and end date in a range DatePicker
2. Clear one of the two input fields and move focus away from that input field
3. Observe that both input fields are cleared

This is likely due to the clearing logic being scoped to the single date picker and not accounting for the range variant (https://github.com/carbon-design-system/carbon/blob/de323e34a5dbf867805a1756c9cf816c1febf34c/packages/react/src/components/DatePicker/DatePicker.js#L450)

emyarod:
Closes #8892

This PR updates the DatePicker so that clearing part of a range DatePicker will not clear all input fields and reset the component

#### Testing / Reviewing

Populate the inputs in a range date picker and then clear the end date. After moving focus, the other input should remain unmodified. Focus should not be trapped on the end date input after it has been cleared


sstrubberg (changes requested):
Still getting some weird interactions with datepicker. Also, I noticed that the calendar pops up if you click anywhere in the label.

Jun-14-2021 09-40-14

emyarod:
@sstrubberg can you elaborate on what weird interactions are occurring? updated the description to specify clearing the end date

for clicking on labels to activate inputs, that is the default behavior of form controls

sstrubberg (approved):
I think I wasn't understanding the original intent when I first reviewed. Getting it now :)

andreancardona (approved):

Issue #4828, opened 2019-12-06
WCAG 2.1 AA compliance


dakahn:
This is an evolving periodically updated (_last updated 9/1/2020_) list that represents our current understanding of the overall accessibility of Carbon Components React across four spectrums:

- 🤖 **has automated accessibility testing issues/violations**
  - we work toward compliance with _all_ known automated testing systems, but use DAP/Accessibility Checker (IBM's in house tools) for compliance tracking.
- ⌨️ **has keyboard accessibility issues/violations** 
  - Any and all mouse interactions are expected to be operable using only the keyboard. And all interactable elements should be reachable in a logical way/order. We defer to the WAI-ARIA Authoring Practices Guidelines.
- 👁️‍🗨️ **has visual accessibility issues/violations**
  - All colors/themes (etc) should meet or improve on WCAG 1.2 guidelines for color contrast. Wayfinding -- or focus states -- should be clear and perceivable in all themes and colors as well
- 👂 **has screen reader accessibility issues/violations**
  - All interactions and content should be accessible using JAWS (in Chrome/Edge and Firefox on Windows 10), NVDA (in Chrome/Edge and Firefox on Windows 10), and VoiceOver (on Safari in macOS)

---

1. Accordion
2. Breadcrumb
3. Button👁️‍🗨️
4. Checkbox👁️‍🗨️⌨️
5. Code snippet 👂 
6. Content switcher👁️‍🗨️⌨️
7. Data table 👁️‍🗨️👂 
8. Date picker⌨️👁️‍🗨️
9. Dropdown🤖⌨️ 👂 
10. File uploader
11. Form
12. Inline loading⌨️👂 
13. Link
14. List
15. Loading
16. Modal👁️‍🗨️👂 
17. Notification
18. Number input👂
19. Overflow menu👁️‍🗨️⌨️
20. Pagination👁️‍🗨️
21. Progress indicator
22. Radio button👁️‍🗨️
23. Search
24. Select👁️‍🗨️
25. Slider
26. Structured list⌨️👂
27. Tabs👂
28. Tag
29. Text input
30. Tile
31. Toggle
32. Tooltip👂
33. UI Shell👂

pascal-potvin:
Just wanted to see if it was possible to get an updated timeline for the ones which have  🚨open accessibility issues. Our products rely heavily on accessibility. Thanks

dakahn:
[...]

PR #8898, opened 2021-06-14
fix(StructuredList): add role attribute to StructuredListRow

atikenny:
Relates to #4828

Fix the accessibility problem in the StrcuturedListRow.

#### Changelog

**Changed**

- added the row role to the `StructuredListRow`

#### Testing / Reviewing

Check the accessibility of the component in Storybook using the IBM Able Chrome plugin.

#### Note

Tried to fix the related issue about the missing role for the label but that invalidates another rule that labels cannot have roles specified.


atikenny:

tay1orjones:
> Fix the accessibility problem in the StrcuturedListRow.

@atikenny could you clarify what the specific problem is that this change solves?

StructuredList was updated recently with some accessibility improvements https://github.com/carbon-design-system/carbon/pull/8172, though I believe they're all behind the v11 flag for now.  @dakahn was this addressed as part of it - moving away from the `table` role to the `grid` role?

atikenny:
> > Fix the accessibility problem in the StrcuturedListRow.
> 
> @atikenny could you clarify what the specific problem is that this change solves?
> 
> StructuredList was updated recently with some accessibility improvements #8172, though I believe they're all behind the v11 flag for now. @dakahn was this addressed as part of it - moving away from the `table` role to the `grid` role?

This PR fixes the original problem with the `table` role missing a `row` role in it.

andreancardona (approved):

dakahn (changes requested):
@tay1orjones we have a total accessibility overhaul behind our feature flags right now yeah. But as for fixing existing problems with v10 StructuredList I'm all for that. I still see 16 errors for `role="row"` in my Accessibility Checker though 👀

atikenny:
> @tay1orjones we have a total accessibility overhaul behind our feature flags right now yeah. But as for fixing existing problems with v10 StructuredList I'm all for that. I still see 16 errors for `role="row"` in my Accessibility Checker though 👀

I tried fixing those as well, but unfortunately, I wasn't able to (it used to be 20 problems now it's 16).
The problem is that instead of a table we use a collection of divs and labels. But we would also like to use a label as a row in that table. This is impossible with the table design, but we cannot ditch the label as then we would lose the selection use-case of this component.

dakahn (approved):
i mean i'm cool with this. It fixes four errors 🏄🏽

tay1orjones (approved):
Thanks for the input @dakahn!

PR #8916, opened 2021-06-15
fix: wifi icon name in icon index

### LINKED ISSUE #2406: Wifi icon names are wrong in Carbon Icons documentation
On the [Carbon Design System Icons page](https://www.carbondesignsystem.com/guidelines/icons/library/), search for "wifi". Mouse over the "code" icon for the two Wifi icons (WiFi and WiFi off), and note the capitalization of the icon names: `WiFi32` and `WiFiOff32`.

However, if you try to import the icons in React using those names, you get an import error. Looking at the actual code in the `@carbon/icons-react` package, I see the icon names are spelled with lowercase "f"s: `Wifi32` and `WifiOff32`. 

Please correct the documentation.

johnbister:
Closes [#2406](https://github.com/carbon-design-system/carbon-website/issues/2406)

This PR corrects the wifi icon name when copied from the icon index on the CDS site. Currently, the icon index copies (< WiFi32 />) instead of (< Wifi32 />) which causes an import error.  The friendlyname for these icons retain the capital "F" despite going through pascalcase so I changed it to "f". 

#### Testing / Reviewing
-Make sure the icon index copies Wifi32 for the Wifi icon
-Make sure the icon index copies WifiOff32 for the Wifi off icon

### DIFF
--- a/packages/icons/icons.yml
+++ b/packages/icons/icons.yml
@@ -16056,14 +16056,14 @@
   sizes:
     - 32
 - name: wifi
-  friendly_name: WiFi
+  friendly_name: Wifi
   aliases:
     - wifi
     - internet
   sizes:
     - 32
 - name: wifi--off
-  friendly_name: WiFi off
+  friendly_name: Wifi off
   sizes:
     - 32
   aliases:


sstrubberg (approved):

emyarod (commented):
just to confirm, do we no longer want to use the Pascal case `WiFi` for these assets?

also does the wifi pictogram need to be updated accordingly?

johnbister:
@emyarod I took a closer look and there are other instances where the icon name in the index does not match what it is in the actual icon package. 

Instead of just a one-off fix for wifi, I’ll make a larger issue to audit the icon index and see what’s happening. I’ll close this for now and work on a larger fix after we get through v11

Issue #8899, opened 2021-06-14
Sass Module Refactor


joshblack:
This issue currently tracks our efforts refactoring our codebase to Dart Sass using Sass modules.

**Components**

- [x] Accordion
- [x] Breadcrumb
- [x] Button
- [x] Checkbox
- [x] Code Snippet
- [x] Combobox
- [x] Content Switcher https://github.com/carbon-design-system/carbon/pull/8862
- [x] Copy Button
- [x] Data Table @tay1orjones https://github.com/carbon-design-system/carbon/pull/8844
- [x] Date Picker @tay1orjones  https://github.com/carbon-design-system/carbon/pull/8895
- [x] Dropdown 
- [x] File Uploader @tay1orjones https://github.com/carbon-design-system/carbon/pull/8920
- [x] Floating Menu
- [x] Form
- [x] Grid https://github.com/carbon-design-system/carbon/pull/8797
- [x] #9027
- [x] Link
- [x] Listbox 
- [x] List
- [x] Loading https://github.com/carbon-design-system/carbon/pull/9030
- [x] Menu @tay1orjones https://github.com/carbon-design-system/carbon/pull/8922
- [x] Modal @tay1orjones https://github.com/carbon-design-system/carbon/pull/8923
- [x] MultiSelect 
- [x] Notification https://github.com/carbon-design-system/carbon/pull/9004
- [x] Number Input 
- [x] Overflow Menu
- [x] Pagination Nav https://github.com/carbon-design-system/carbon/issues/9068
- [x] Pagination @jnm2377 https://github.com/carbon-design-system/carbon/pull/8953/
- [x] Popover @dakahn 
- [x] Progress Bar https://github.com/carbon-design-system/carbon/issues/9075 @tw15egan 
- [x] Progress Indicator @jnm2377 https://github.com/carbon-design-system/carbon/pull/9022
- [x] Radio Button @dakahn #8980 
- [x] Search @dakahn #8967 
- [x] Select @jnm2377 https://github.com/carbon-design-system/carbon/pull/8953/
- [ ] Skeleton @jnm2377 https://github.com/carbon-design-system/carbon/issues/9130
- [x] Slider @sstrubberg #8981
- [x] Structured List @sstrubberg #8931
- [x] Tabs @sstrubberg 
- [x] Tag @dakahn 
- [x] Text area #9037 
- [x] Text Input @sstrubberg #8988
[...]

PR #8920, opened 2021-06-15
feat(file-uploader): add FileUploader to @carbon/styles

tay1orjones:
Refs #8899

Brings `FileUploader` to the `@carbon/styles` package

#### Changelog

**New**

- include refactored file added to `/styles`, including a test


packages/styles/scss/components/file-uploader/_file-uploader.scss
```diff
@@ -0,0 +1,351 @@
...
+@use '../../utilities/focus-outline' as *;
+@use '../../utilities/high-contrast-mode' as *;
+@use '../../utilities/visually-hidden' as *;
+
+@use '../button';
+@use '../form';
+@use '../loading';
+
+// TODO - how should we resolve this? Should we keep it? Is the '@use type' above enough?
+// @use '../../globals/scss/vendor/@carbon/elements/scss/type/styles';
```
tay1orjones: Is the `@use '../../type' as *;` above enough to remove this?

tay1orjones: Sounds like the answer to this is yes, the vendor stuff can be removed. I've pushed an update to fix.

joshblack (changes requested):
Could you include the story files too @tay1orjones? Just to double-check that the styles work 👀  (sorry we forgot to include this in the template 🤦 )

tay1orjones:
@joshblack Yes! Thank you for the reminder. I took a stab at not only bringing them over but refactoring them to use the new storybook 6 args, argTypes, and controls.

joshblack:
@tay1orjones seems like a weird storybook issue coming up 🤔

tw15egan (approved):
The controls look ✨ wonderful ✨  💯

tay1orjones:
@joshblack I'm not sure what happened, but it's up and working now since I updated with latest from `main`.

joshblack (approved):

PR #8931, opened 2021-06-16
feat(structured-list): sass module refactor

sstrubberg:
Refs #8899 

Brings `StructuredList` (& `FeatureFlags`) component(s) to the `carbon-react` package. Also brings `StructuredList` into `@carbon/styles`.

#### Changelog

**New**

- Styles added for `StructuredList`
- Stories added for `StructuredList`

#### Testing / Reviewing

Pull down the PR locally, go to `carbon-react` folder and run `yarn storybook`. Ensure `StructuredList` renders properly.

#### Current Bugs

- You'll notice that there is a radio button on the `Selection` story. Having a challenging time getting that to go away.

image


packages/carbon-react/src/components/StructuredList/StructuredList.stories.js
```diff
@@ -0,0 +1,203 @@
...
+
+import {
+  StructuredListWrapper,
+  StructuredListHead,
+  StructuredListBody,
+  StructuredListRow,
+  StructuredListInput,
+  StructuredListCell,
+} from '.';
+import StructuredListSkeleton from './StructuredList.Skeleton';
```
joshblack:
```suggestion
import {
  StructuredListWrapper,
  StructuredListHead,
  StructuredListBody,
  StructuredListRow,
  StructuredListInput,
  StructuredListCell,
 StructuredListSkeleton
} from '../StructuredList';
```

Let me know if this works 🤔 Would be great to get them all from the same export

packages/carbon-react/src/components/StructuredList/StructuredList.Skeleton.js
```diff
@@ -0,0 +1,76 @@
+/**
```
joshblack: I think we can remove this and use the exported one from `carbon-components-react` directly (unless did you maybe run into issues with it not being exported? 👀

joshblack (approved):
Just left a small comment, I think you're good to go after that! The radio input stuff should be addressed when that refactor lands, I think.

packages/styles/scss/components/structured-list/_structured-list.scss
```diff
@@ -0,0 +1,226 @@
...
+//
+
+@use '../../config' as *;
+@use "../../type" as *;
+@use '../../motion';
+@use '../../theme' as *;
+@use "../../utilities/focus-outline" as *;
+@use "../../utilities/skeleton" as *;
+@use "../../utilities/high-contrast-mode" as *;
+
```
tw15egan:
```suggestion
@use '../../utilities/convert' as *;
```

Since there is a reference to the `rem` function on L136, we'll need to bring in the convert functions. Or we can change this line to `2.5rem` since it's the only instance in this file

dakahn (approved):

Issue #8911, opened 2021-06-15
Add Percy visual regression testing (VRT) snapshots for all components


tay1orjones:
### Summary

In #8282 Cypress was set up for e2e testing and also added the Percy integration to facilitate visual regression testing via snapshots taken in the e2e tests. We need to enable a snapshot for each component, this epic/issue is to track the status of adding a single baseline snapshot for each component.

### Process

High level steps for each component:
1. List in [a comment below](https://github.com/carbon-design-system/carbon/issues/8911#issuecomment-882998388)
1. Open the PR, ensure percy status check UI shows the component has an associated snapshot
   - This may require approval in the Percy UI

### Components
- [x] Accordion
- [x] AccordionItem
- [x] #9263
- [x] AspectRatio https://github.com/carbon-design-system/carbon/pull/8958
- [x] Breadcrumb https://github.com/carbon-design-system/carbon/pull/8958
- [x] BreadcrumbItem https://github.com/carbon-design-system/carbon/pull/8958
- [x] #9264
- [x] Button https://github.com/carbon-design-system/carbon/pull/8958
- [x] https://github.com/carbon-design-system/carbon/issues/9265
- [x] ButtonSet https://github.com/carbon-design-system/carbon/pull/8958
- [x] Modal https://github.com/carbon-design-system/carbon/pull/8958
- [x] https://github.com/carbon-design-system/carbon/issues/9266
- [x] https://github.com/carbon-design-system/carbon/issues/9267
- [x] https://github.com/carbon-design-system/carbon/issues/9268
- [ ] Add VRT for ContentSwitcher
- [ ] Add VRT for Copy
- [ ] Add VRT for CopyButton
- [ ] Add VRT for DataTable
- [ ] Add VRT for DatePicker
- [x] https://github.com/carbon-design-system/carbon/issues/9274
- [ ] Add VRT for ErrorBoundary
- [ ] Add VRT for FileUploader
- [ ] Add VRT for Form
- [x] https://github.com/carbon-design-system/carbon/issues/9278
- [x] #9279
- [ ] Add VRT for List components
- [ ] Add VRT for Loading
- [x] https://github.com/carbon-design-system/carbon/issues/9283
- [ ] Add VRT for Notification components
- [ ] Add VRT for NumberInput
- [ ] Add VRT for OverflowMenu
- [ ] Add VRT for Pagination
- [ ] Add VRT for ProgressIndicator
- [ ] Add VRT for RadioButton
[...]

PR #8958, opened 2021-06-18
chore(test): add baseline e2e tests and percy snapshots

tay1orjones:
Refs #8911

Adds baseline end to end tests for 5 components: `AspectRatio`, `Breadcrumb`, `Button`, `ButtonSet`, `Modal`. These baseline tests include a percy snapshot for each component. 

`Modal` got some special treatment as an example for the future of how to test and snapshot components at the different breakpoint widths. Even with this, the modal tests all together only take .75s to execute!

#### Changelog

**New**

- test: e2e and percy snapshots

#### Testing / Reviewing

- Status checks should pass
- Percy status check (and UI) should register new snapshots for the 5 components listed in this PR.

If you're interested to see what the tests are doing: 
  - pull down this branch
  - cd into `packages/react`
  - run `yarn cypress open-ct`
  - this will open the cypress gui in a new instance of chrome where you can select and run tests, and view the UI it's snapshotting and testing against.


emyarod (approved):
looks good to me

image

I don't have permission to approve the new snapshots on the percy dashboard

tay1orjones:
@emyarod could you try going to browserstack.com, signing out, and then sign back in with SSO - then hop over to percy.io and sign in with browserstack? It recently switched over to SSO, so your old/non-ibm creds may not work.

emyarod:
got it, thanks!

jnm2377 (approved):

PR #8967, opened 2021-06-21
feat(search): sass module refactor

### LINKED ISSUE #8940: feat(Search): add Search to @carbon/styles
This issue is for updating the Search component to Sass Modules in the `@carbon/styles` package. In this refactor, we will:
- [x] Copy over styles for for the component to `packages/styles/scss/components/{component}/_{component}.scss`
  - Any additional files will live in the folder `packages/styles/scss/components/{component}/`
- [x] Create an `_index.scss` at `packages/styles/scss/components/search/_index.scss`
- [x] Create a test file at `packages/styles/scss/components/__tests__/{component}-test.js`
- [x] add the storybook stories to @carbon/react
- [x] `@forward` the mixin and any variables from `_{component}.scss` in `_index.scss`
- [x] Include the styles from `_{component}.scss` by calling the mixin in `_index.scss`
- [x] in `{component}-test.js` test the public API/config options of the `/_index.scss` file
- [x] The `_{component}.scss` will then need to be refactored to use the new import paths and conventions. You can find common recipes here: https://github.com/carbon-design-system/carbon/issues/8899#issuecomment-860887959

dakahn:
Closes #8940 

Sass module refactor for Search component

## Testing

Component should behave and appear the same 👍🏽

### DIFF
--- a/packages/styles/scss/components/search/_search.scss
+++ b/packages/styles/scss/components/search/_search.scss
@@ -11,19 +11,20 @@
 
 @use '../../theme' as *;
 @use '../../config' as *;
+@use '../../type';
+@use '../../motion' as *;
+@use '../../spacing' as *;
 @use '../../utilities/placeholder-colors' as *;
 @use '../../utilities/focus-outline' as *;
 @use '../../utilities/visually-hidden' as *;
-@use '../../motion' as *;
-@use '../../spacing' as *;
 @use '../../utilities/component-reset' as *;
 @use '../../utilities/skeleton' as *;
 @use '../../utilities/high-contrast-mode' as *;
-@use '../../type';
+@use '../../utilities/convert' as *;
 @use '../../utilities/button-reset';
 
 /// Search styles
-/// @access private
+/// @access public
 /// @group search
 @mixin search {
   .#{$prefix}--search {
--- a/packages/styles/scss/components/search/_search.scss
+++ b/packages/styles/scss/components/search/_search.scss
@@ -1,0 +1,368 @@
+//
+// Copyright IBM Corp. 2016, 2018
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+//-----------------------------
+// Search
+//-----------------------------
+
+@use '../../theme' as *;
+@use '../../config' as *;
+@use '../../utilities/placeholder-colors' as *;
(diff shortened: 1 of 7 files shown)


packages/styles/scss/components/search/_search.scss
```diff
@@ -0,0 +1,368 @@
...
+@use '../../config' as *;
+@use '../../utilities/placeholder-colors' as *;
+@use '../../utilities/focus-outline' as *;
+@use '../../utilities/visually-hidden' as *;
+@use '../../motion' as *;
+@use '../../spacing' as *;
+@use '../../utilities/component-reset' as *;
+@use '../../utilities/skeleton' as *;
+@use '../../utilities/high-contrast-mode' as *;
+@use '../../type';
```
tay1orjones: If there's no mixin name conflicts you can use the `as *`

```suggestion
@use '../../type' as *;
```

which then simplifies the usages:

```diff
-@include type.type-style('body-short-01');
+@include type-style('body-short-01');
```

Also a small style nit I got from Josh: order of imports should be:

* top-level (colors, config, etc)
* component imports (form, others used)
* utilities
* local

packages/styles/scss/components/search/_search.scss
```diff
@@ -0,0 +1,368 @@
...
+@use '../../utilities/button-reset';
+
+/// Search styles
+/// @access private
```
tay1orjones:
```suggestion
/// @access public
```

dakahn:
Seems like the styles are broken in the Story -- :| ON IT

tay1orjones (changes requested):
Just a couple super minor things

packages/styles/scss/components/search/_search.scss
dakahn: Ah, yeah I was getting more specific with type because both /type and /component-reset forwarded `@include reset` which was throwing an error

sstrubberg (approved):

packages/styles/scss/components/search/_search.scss
dakahn: If there's a better way to take care of those overlapping imports like that let me know tho

tay1orjones: No that's totally cool, I ran into the same thing!

tay1orjones (approved):

tw15egan (approved):

PR #8971, opened 2021-06-21
feat(styles): add component token support for Sass Migration effort

joshblack:
This PR updates the `theme` mixin to include component token maps which can be generated by a `get-tokens` helper now available in `utilities/component-tokens`. This should help unblock any efforts that use component tokens in our sass module refactor.

More usage info for this approach can be found in: https://github.com/carbon-design-system/carbon-v11/discussions/1

#### Changelog

**New**

- New utilities for working with component tokens, specifically a helper that lets you get the tokens from a token map for a specific theme

**Changed**

- Update theme helper to take in a list of maps to allow component tokens to be passed in

**Removed**


andreancardona (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #8988, opened 2021-06-22
feat(text-input): storybook and carbon/styles

sstrubberg:
Refs #8899 

Brings `TextInput` to the `carbon-react` & `@carbon/styles` packages.

#### Changelog

**New**

- Styles added for `TextInput`
- Stories added for `TextInput`

#### Testing / Reviewing

Pull down the PR locally, go to `carbon-react` folder and run `yarn storybook`. Ensure `TextInput` renders properly.

Run `yarn test --watch packages/styles/scss/components/__tests__/text-input-test.js` and make sure you get green checkmarks.

#### Current Bugs

Having trouble rendering `FullyControlledTogglePasswordVisibility` story


packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+  .#{$prefix}--btn.#{$prefix}--btn--icon-only.#{$prefix}--text-input--password__visibility__toggle.#{$prefix}--tooltip__trigger svg,
+    .#{$prefix}--btn.#{$prefix}--btn--icon-only.#{$prefix}--text-input--password__visibility__toggle.#{$prefix}--tooltip__trigger:hover
+    svg {
+    @include high-contrast-mode('icon-fill');
+  }
+}
+
+@include exports('text-input') {
+  @include text-input;
+}
```
tay1orjones: I believe this should be removed

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+
+    width: 100%;
+    height: rem(40px);
+    padding: 0 $carbon--spacing-05;
+    border: none;
+    border-bottom: 1px solid $border-strong;
+    background-color: $field;
+    color: $text-primary;
+    transition: background-color $duration--fast-01 motion(standard, productive),
+      outline $duration--fast-01 motion(standard, productive);
```
tay1orjones: Small change to duration tokens - there's a couple other spots in the file to update in addition to this one
```suggestion
    transition: background-color $duration-fast-01 motion(standard, productive),
      outline $duration-fast-01 motion(standard, productive);
```

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+/// @group text-input
+@mixin text-input {
+  .#{$prefix}--text-input {
+    @include reset;
+    @include type-style('body-short-01');
+    @include focus-outline('reset');
+
+    width: 100%;
+    height: rem(40px);
+    padding: 0 $carbon--spacing-05;
```
tay1orjones: Spacing tokens don't need the `carbon--` anymore. There's a few more usages in the file that can be updated.

```suggestion
    padding: 0 $spacing-05;
```

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  // TODO V11: Remove xl selector
+  .#{$prefix}--text-input--xl,
+  .#{$prefix}--text-input--lg {
+    height: rem(48px);
+  }
+
+  .#{$prefix}--text-input--sm {
+    height: rem(32px);
```
tay1orjones, line 66: I don't see the `convert` mixin being imported above. Maybe tests still pass because it brings in the form styles, which `@use`s `convert`?

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+  .#{$prefix}--text-input--sm.#{$prefix}--password-input {
+    padding-right: $carbon--spacing-07;
+  }
+
+  .#{$prefix}--text-input--lg.#{$prefix}--password-input {
+    padding-right: $carbon--spacing-09;
+  }
+
+  .#{$prefix}--text-input::placeholder {
+    @include placeholder-colors;
```
tay1orjones, line 82: I also don't see the `placeholder-colors` mixin `@use`ed above either, similar to `convert`. I guess a question to @joshblack - should we still `@use` mixins even when they're included in other dependent components?

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+    fill: $support-error;
+    transform: translateY(-50%);
+  }
+
+  .#{$prefix}--text-input__invalid-icon--warning {
+    fill: $support-warning;
+  }
+
+  .#{$prefix}--text-input__invalid-icon--warning path:first-of-type {
+    fill: $carbon__black-100;
```
tay1orjones: I think this should be updated to be the following:
```suggestion
    fill: $black-100;
```

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+    .#{$prefix}--text-input--password__visibility__toggle {
+      right: $carbon--spacing-08;
+    }
+  }
+
+  //-----------------------------
+  // Skeleton
+  //-----------------------------
+  .#{$prefix}--skeleton.#{$prefix}--text-input {
+    @include skeleton;
```
tay1orjones, line 255: Also no corresponding `@use` for this mixin

packages/carbon-react/src/components/TextInput/TextInput.stories.js
```diff
@@ -0,0 +1,221 @@
...
+  info: {
+    text: `
+    Fully controlled text field with password visibility toggle.
+  `,
+  },
+};
+
+export const Skeleton = () => <TextInputSkeleton />;
+
+Skeleton.storyName = 'skeleton';
```
tay1orjones: Storybook v6 automatically infers the name of the story from the export name, so I think we can remove these for most cases.

packages/carbon-react/src/components/TextInput/TextInput.stories.js
```diff
@@ -0,0 +1,221 @@
...
+
+Skeleton.parameters = {
+  info: {
```
tay1orjones: This addon is deprecated and isn't configured in the v6 storybook. I've just been removing them altogether. I think ideally the content would be moved to the `TextInput.mdx`?

packages/carbon-react/src/components/TextInput/TextInput.stories.js
```diff
@@ -0,0 +1,221 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState } from 'react';
+import { action } from '@storybook/addon-actions';
+import { withKnobs, boolean, select, text } from '@storybook/addon-knobs';
```
tay1orjones: Do you get errors in the console from this since `addon-knobs` is deprecated/removed in v6? I was unable to get DatePicker stories working with the knobs functions in the props object/functions. I ended up stripping them all down to the default values from the knob params, like this:

```diff
-defaultValue: text(
-      'Default value (defaultValue)',
-      'This is not a default value'
-    ),
+ defaultValue: 'This is not a default value',
```

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -0,0 +1,384 @@
...
+//-----------------------------
+@use '../../theme' as *;
+@use '../../type' as *;
+@use '../../utilities/component-reset';
+@use '../../utilities/focus-outline' as *;
+@use '../../utilities/button-reset';
+@use '../form';
+
+/// Text input styles
+/// @access private
```
tay1orjones:
```suggestion
/// @access public
```

tay1orjones (changes requested):
I found a couple things to update. I'm unsure on the strategy if we should/shouldn't `@use` mixins that are `@use`d in other dependent component files (`form` in this case). Maybe Josh can add his opinion on it.

Also I believe there should be an `_index.scss` alongside the `text-input.scss` file that does the whole forward/use deal

```
@forward 'textinput';
@use 'textinput';

@include textinput.textinput;
```

Edit: oh and just remembered the component needs to be used in the central entrypoint `packages/styles/scss/components/_index.scss`

packages/carbon-react/src/components/TextInput/TextInput.stories.js
```diff
@@ -0,0 +1,221 @@
...
+      TextInputSkeleton,
+      'TextInput.PasswordInput': TextInput.PasswordInput,
+    },
+  },
+};
+
+export const Default = () => (
+  <TextInput
+    type={select('Form control type (type)', types, 'text')}
+    {...props.SharedInputProps()}
```
emyarod: I think the prop knobs need to be removed from the non-playground stories

emyarod (commented):
some component styles appear to be missing

image

packages/carbon-react/src/components/TextInput/TextInput.stories.js
sstrubberg: Nope. No console errors.

packages/carbon-react/src/components/TextInput/TextInput.stories.js
sstrubberg: If i straight up remove them, then there's nothing being passed into those stories. How do fill the props?

emyarod: they would still take default props or any other hard coded props and they just won't be configurable via knobs. if I'm not mistaken only the playground stories (currently not in `carbon-react` yet from what I can tell) would have knobs

tw15egan: Yeah, we can skip the playground stories for now and just use the simple examples with hardcoded props. We can go through the new storybook at a later time to add in more controls/knobs.

emyarod (approved):

tw15egan:
Noticing an issue with the tooltips on the password text input... but you have the `Tooltip` import, so I'm not too sure what might be causing it. 

Screen Shot 2021-06-24 at 12 07 29 PM

tw15egan (approved):

tay1orjones (approved):
Solid! 🎉

PR #9006, opened 2021-06-23
feat(useDisclosure): adds useDisclosure hook and test

### LINKED ISSUE #8119: Cannot scroll to view absolutely positioned div inside an ExpandableTile - Safari only
- [ ] `carbon-components`

## Detailed description
In Cognos, we have a few cases where we show a calendar control inside an ExpandableTile.  When the user clicks on the control and the calendar displays they cannot scroll to view the entire calendar - this only happens in Safari, in other browsers the user can scroll without any issues.  I was also able to reproduce the issue with a simple absolutely positioned div taking the calendar control out of the equation.

I tried using the latest carbon-components-react and can still reproduce the issue.

Use a ExpandableTile that has a control (button) that shows / hides an absolutely div that is higher than the Tile container.  When the div is shown, try scrolling down to see the entire div.  In Safari, you cannot scroll.

I've created a reproducible test case in codesandbox showing the issue with the Calendar and also a simple div. 
https://codesandbox.io/s/tilebelowthefoldcontent-r578b?file=/src/index.js:700-714

### LINKED ISSUE #8819: Implement initial `useDisclosure` hook

andreancardona:
Closes #8819 

This `useDisclosure` pr creates the hook and the tests (tests for functionality and accessibility)
- [ ] cd packages/react and run `yarn test --watch packages/react/src/components/Disclosure/__tests__/useDisclosure-test.js`


packages/react/src/components/Disclosure/index.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { useRef, useState } from 'react';
+
+function useDisclosure(id) {
+  const [open, setOpen] = useState(false);
+  const ref = useRef(null);
```
joshblack: Seems like this might not be used in the implementation / test so could be removed

joshblack (approved):
Looks great! Just left a comment and should be good to go.

jnm2377 (approved):

PR #9008, opened 2021-06-23
feat(project): add flat exports for namespaced components

### LINKED ISSUE #8901: Flatten nested imports and add deprecation notice
Discussion: https://github.com/carbon-design-system/carbon/discussions/8642

This issue is for tackling the normalization of our export strategy. Currently, we export a mixed of named exports and exports only available as fields on these exports. For example: `TextInput` and `TextInput.ControlledPasswordInput`.

With this change, we would like all exports to be available as named exports. To get there, we will need to:

- [x] Identify all components that are exported as a field on another component
- [x] Determine their export name in the normalized approach
- [x] Add each component as an export using the normalized approach
- [x] Add a deprecation notice to the old import with a message instructing teams on how to migrate to the new import

tay1orjones:
Closes #8901

Adds new flat exports for previously namespaced components, like `PasswordInput`, deprecating (with notices) the old `TextInput.PasswordInput`

#### Changelog

**New**

- added flat exports for `PasswordInput`, `ControlledPasswordInput`, and `FilterableMultiSelect`
- added new `deprecateFieldOnObject()` and deprecation notices for the old exports
- added a migration doc too

#### Testing / Reviewing

* Public API snapshot should now include these new exports at the root level in addition to retaining the old namespaced exports
* I modified the import path of the TextInput in the TextInput story file, so if you go to that story locally and view the PasswordInput stories, you'll see the console warning pop:
image


tay1orjones:
This is ready for review, but I'm unsure how best to fix the tests. The public api snap updates, but the test fails because the components are calling `console.warn()`.

I don't think we want to `jest.spyOn(global.console, "warn").mockImplementation(() => {})` because that would swallow *all* warnings. I'm not sure what the best approach is to fix this 🤔

tw15egan:
@tay1orjones is there any way to check for a specific warning message? Or a snippet from the message like "Please import and use" so that it will catch all of our deprecation warnings, but still emit the rest?

packages/react/docs/migration/11.x-namespaced-exports.md
```diff
@@ -0,0 +1,44 @@
...
+namespace of another component. These have been deprecated, and a new export has
+been made available for each so these can be imported directly. In v11.x the
+namespaced exports will be removed.
+
+## Process
+
+1. Update imports for the components listed in the table below.
+
+```diff
+- import TextInput from 'carbon-components-react';
```
joshblack:
```suggestion
- import { TextInput } from 'carbon-components-react';
```

packages/react/docs/migration/11.x-namespaced-exports.md
```diff
@@ -0,0 +1,44 @@
...
+
+2. Update usages of the components listed in the table below, they no longer
+   need the namespace
+
+```diff
+- <TextInput.PasswordInput/>
++ <PasswordInput/>
+```
+
+## Changes:
```
joshblack, line 38: Wouldn't be too bad as a codemod, I think, right?

packages/react/src/index.js
```diff
@@ -72,7 +74,15 @@ export ListItem from './components/ListItem';
 export Loading from './components/Loading';
 export Modal from './components/Modal';
 export ModalWrapper from './components/ModalWrapper';
-export MultiSelect from './components/MultiSelect';
+import MultiSelect from './components/MultiSelect';
+if (__DEV__) {
```
joshblack: Would it make sense to put these directly on the component export itself instead of in the entrypoint file? This could help out with the direct import use-case, potentially, and catch any usage where we might be using these components from a relative path.

joshblack (changes requested):

packages/react/src/index.js
tay1orjones: @joshblack yes - I just pushed an update moving this to the exports themselves instead of the entrypoint file. Let me know what you think, I had to modify the syntax of the exports a bit.

packages/react/docs/migration/11.x-namespaced-exports.md
tay1orjones: I think it would be pretty chill, yeah. ~I can take a stab at it here if you want. Otherwise I can make a new issue for it~

Opened https://github.com/carbon-design-system/carbon/issues/9190 for the codemod

tay1orjones:
@tw15egan thanks for the idea! I pushed an update to do just that, it matches on the "field has been deprecated" substring.

tw15egan (approved):
LGTM 👍 ✅

sstrubberg (approved):

packages/react/__tests__/PublicAPI-test.js
```diff
@@ -124,6 +124,16 @@ beforeEach(() => {
  * the components that we export and their corresponding API.
  */
 test('Public API should only change with a semver change', () => {
+  // deprecation warnings for fields on objects should be ignored
+  const originalWarn = console.warn;
+  const spy = jest.spyOn(console, 'warn').mockImplementation((message) => {
```
joshblack: I think you can add in a mock for our internal helper here: https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/internal/__mocks__ like we do for warning

Then any test that needs to make sure warning isn't emitted can do `jest.mock('../../internal/deprecateFieldOnObject')`

packages/react/src/components/MultiSelect/index.js
```diff
@@ -4,11 +4,16 @@
...
  */
+import { deprecateFieldOnObject } from '../../internal/deprecateFieldOnObject';
 
 import MultiSelect from './MultiSelect';
 import FilterableMultiSelect from './FilterableMultiSelect';
 
 FilterableMultiSelect.displayName = 'MultiSelect.Filterable';
 MultiSelect.Filterable = FilterableMultiSelect;
 
+if (__DEV__ && !process.env.JEST_WORKER_ID) {
```
joshblack: is the `JEST_WORKED_ID` check to see if we're in a test? If so, I think we can use that `mock` strategy mentioned above for any tests for this method instead

packages/react/src/internal/deprecateFieldOnObject.js
```diff
@@ -0,0 +1,30 @@
...
+function deprecateFieldOnObject(object, field, Component, message) {
+  Object.defineProperty(object, field, {
+    enumerable: true,
+    get() {
+      if (!didWarnAboutDeprecation[field]) {
+        warning(
+          false,
+          message ||
+            `The ${field} field has been deprecated on the ${object.displayName} object. ` +
+              `Please import and use ${Component.displayName} directly.`
```
joshblack: Might need to add in a fallback potentially if there is no `displayName`

```suggestion
              `Please import and use ${Component.displayName || Component.name} directly.`
```

joshblack (changes requested):
Looking great 🔥  Just left a note about the mock strategy that can help out with tests and I think after that should be good to go 🚢

packages/react/__tests__/PublicAPI-test.js
tay1orjones: Ah, perfect. That's much cleaner!

packages/react/src/components/MultiSelect/index.js
tay1orjones: Fixed

packages/react/src/internal/deprecateFieldOnObject.js
tay1orjones: Yes, added

tay1orjones:
Very much appreciate the suggestions @joshblack - all have been addressed. The mock approach feels much cleaner ✨ 👍

joshblack (approved):

PR #9012, opened 2021-06-23
fix(themes): add conditional to fallback guard

joshblack:
This PR addresses a use-case with custom theming that can also impact IBM customers. To re-create the issue today, imagine the following scenario:

```scss
@import '@carbon/themes/scss/themes';

// Custom theme
$carbon--theme: (
  ui-background: #ffffff,
);
@include carbon--theme();

// Here, $ui-background is defined but our new token $background is not
```

`$background` not being defined causes issues in our components since they rely on that token to render correctly.

The `$background` fallback is not applied because our `@if` uses the following checks:

```scss
@if global-variable-exists('feature-flags') and map-get($feature-flags, 'enable-v11-release') != true {
  // ...
}
```

As a result, if `$feature-flags` does not exist as a global variable then we will not hit this branch. The build command now generates:

```scss
@if global-variable-exists('feature-flags') ==  false or
  global-variable-exists('feature-flags') and map-get($feature-flags, 'enable-v11-release') !=  true
{
  // ...
}
```

This will now emit the fallback in two situations:

- `$feature-flags` is not defined
- `$feature-flags` exists but `enable-v11-release` is not true

Both situations will skip emitting the fallbacks if `$feature-flags` exist and `enable-v11-release` is true.

#### Changelog

**New**

**Changed**

- Update builder with new conditional logic
- Update test file for mixins to new sass renderer
- Add v11 test for this bug and make sure builder passes the test

**Removed**

#### Testing / Reviewing

- Verify v11 test passes as expected in `mixins-test.js`
- Double-check my logic in case I'm missing anything 😅

### DIFF
--- a/packages/themes/__tests__/mixins-test.js
+++ b/packages/themes/__tests__/mixins-test.js
@@ -9,22 +9,20 @@
 
 'use strict';
 
-const { convert, createSassRenderer } = require('@carbon/test-utils/scss');
+const { SassRenderer } = require('@carbon/test-utils/scss');
 const { themes } = require('../src');
 
-const render = createSassRenderer(__dirname);
+const { render } = SassRenderer.create(__dirname);
 
 describe('_mixins.scss', () => {
   it('should export a carbon--theme mixin', async () => {
-    const { calls } = await render(`
+    const { unwrap } = await render(`
       @import '../scss/mixins';
 
-      $t: test(mixin-exists(carbon--theme));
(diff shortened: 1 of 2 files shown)


joshblack:
cc @tw15egan would love to get your eyes on this too if you have a sec

tw15egan (approved):
Working as expected 👍🏻

jnm2377 (approved):

PR #9013, opened 2021-06-23
feat(react): add support for layout and text direction

joshblack:
This PR is a preview of potential changes we could make to better support bidirectional layout and text in Carbon.

For those seeing this work for the first time, you may be familiar with the `dir` or `direction` property in HTML and CSS. This direction is commonly seen as the way to enable LTR (left to right) or RTL (right to left) experiences on the web. However, one common misconception is that you will want a page to _always_ be LTR or RTL. In certain situations, you may want the layout direction to be different than the text direction.

The localization section on Material has a great example of this from Google Docs: https://material.io/design/usability/bidirectionality.html#localization In this example, they show that the editor is in a LTR mode but when editing RTL text you will want the text and controls to be mirrored, as appropriate.

This PR looks to add in a distinction between layout and text direction and provides a primitive type, `Text`, that can be used to automatically infer text direction and override it, when appropriate. These primitives include:

- `LayoutDirection` for setting the layout direction in a part of an application
- `TextDirection` for setting the text direction in a part of an application
- `Text` for applying the correct `dir` attribute on a particular text node based on the `TextDirection`, defaulting to `auto`

This draft also includes an update for what this will look like for `RadioButton` and `RadioButtonGroup`. The tl;dr of these changes is to replace any location where text can be rendered with a `Text` component.

The storybook includes several demos:

- A default example with simple Text direction support
- A layout direction example with a default LTR mode switching to RTL for a paragraph written in Arabic
- A text direction override example that allows a consumer to force the text direction of any particular text in the sub-tree


packages/react/src/components/Text/Text.js
```diff
@@ -0,0 +1,95 @@
...
+        value.direction = override;
+      } else if (parentDirection === 'auto') {
+        textProps.dir = override;
+      }
+    } else if (parentDirection !== dir) {
+      textProps.dir = dir;
+      value.direction = dir;
+    } else if (parentDirection === 'auto') {
+      textProps.dir = dir;
+    }
```
joshblack, line 40: This was some tricky conditional logic to work with nested `Text` nodes, curious if you have any suggestions @tay1orjones to simplify 🤔 

Could also just always pass in `dir` no matter what the parent is and call it a day. Would simplify this a ton but would mean extra markup. Wasn't sure which way would be better and was curious what others thought would be best. I've been staring at this too long now 😂

tay1orjones: Yeah that is a tricky block. I don't see a way to simplify. It looks like the complexity just stems from only setting `value.direction` when the `parentDirection` isn't `auto`.

tay1orjones: > Could also just always pass in dir no matter what the parent is and call it a day.

Also I'm not opposed to this. It feels more explicit via the extra markup but I think either is fine. Would be ideal to get feedback from consumers (or as we implement it ourselves) if this auto nesting functionality is preferred.

tay1orjones (approved):

joshblack:
cc @jnm2377 @andreancardona let me know if you have any questions!

jnm2377 (approved):
Phew that was a lot to wrap my head around lol but really awesome!

joshblack:
@kubijo got it! Wasn't sure what background there was and wanted to provide context.

In terms of meaningful changes, I think that in terms of styles the biggest switch in v11 will be the use of logical properties. We'll also be adding previews for RTL directly in storybook 👍

PR #9025, opened 2021-06-24
fix(tile): add link color contrast

### LINKED ISSUE #8701: Links on expandable tile on hover have insufficient color contrast
Hi Carbon love you guys thanks for all you do.

Anyway! I know the below is not the ideal usage of an expandable tile, but I'm reviewing a lot of services lately that use a really, really long expanded state, which means hovering the entire time is unavoidable, and they include learn more links throughout, which I think should be allowed. However, the link color isn't accessible on the tile's hover color:

image
image

Is there a way we can help the services implement this better? Can you optionally disable the background color change hoverstate for this tile ever? Or give the link a different color on hover?

andreancardona:
Closes #8701 

## Testing:

- [ ] Checkout `Tile --> Expandable` and it should be seen as below with `color: $link-secondary`

Screen Shot 2021-06-24 at 10 52 16 AM


packages/components/src/components/tile/_tile.scss
```diff
@@ -52,6 +52,10 @@
       background: $layer-hover;
     }
   }
+  .#{$prefix}--tile-clickable tbody tr:hover td .#{$prefix}--link,
+  .#{$prefix}--tile-clickable tbody tr:hover th .#{$prefix}--link {
```
tw15egan: I think we just need something like

```suggestion
  .#{$prefix}--tile--expandable .#{$prefix}--link {
```

tw15egan (changes requested):

tw15egan (approved):
LGTM 👍 ✅

jnm2377 (approved):

PR #9039, opened 2021-06-25
fix(Toggle): remove aria-label, fix aria-labelledby

### LINKED ISSUE #6567: Toggle and ToggleSmall aria-label is better placed on the checkbox, not the label element
## Environment

> MacOS

> Chrome

> VoiceOver

## Detailed description

VoiceOver on Chrome does not announce the aria-label of the Toggle element because the attribute is on the label element, not the input itself.

I expected the text of the aria-label to be announced by VoiceOver upon focusing on the toggle element.

> What happened instead?

The text was not announced, only the "tick box" and its on/off state. 

Looking at the "accessible name computation" specification, it seems to me that the current setup, having an aria-label on a label, is not in the specification. The aria-label attribute would be better suited on the input element itself.

1. Turn on VoiceOver for Chrome
2. Go to [Storybook](https://react.carbondesignsystem.com/?path=/story/togglesmall--toggled)
2. Navigate to the Toggle element

## Additional information

- It's important to note that the label is announced in Safari, which is probably the most used browser with VoiceOver
- I was unable to test this with NVDA/JAWS at the moment
- But as per specification, I thing the aria attribute should go onto the input element anyways. 
- I tested it by moving the attribute using the Developer Tools, and it worked both with Safari and Chrome

Let me know if you found my reasoning solid, and I'll go ahead and create a PR!

dakahn:
Closes #6567 

Running some changes in aria to try and get this thing speaking better with Voiceover in Chrome on macOS

This PR is a WIP -- need to hit it with a little testing and make sure this isn't a regression on some browsers/readers.

### DIFF
--- a/packages/react/src/components/Toggle/Toggle.js
+++ b/packages/react/src/components/Toggle/Toggle.js
@@ -72,7 +72,6 @@ class Toggle extends React.Component {
 
   static defaultProps = {
     defaultToggled: false,
-    // ['aria-label']: 'Toggle',
     labelA: 'Off',
     labelB: 'On',
     onToggle: () => {},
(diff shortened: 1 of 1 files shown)


emyarod (commented):
converted to draft while this is still WIP

PR #9060, opened 2021-06-29
reenable safeguarding missing enable-v11-release feature flag

### LINKED ISSUE #9055: Button component does not inherit theme token
This is how I am importing the scss
```
$feature-flags: ( // Custom CSS properties must be enabled to dynamically switch themes
  enable-css-custom-properties: true,
  ui-shell: true,
  grid-columns-16: true);

$css--font-face: true;
$css--helpers: true;
$css--use-layer: true;
$css--reset: true;
$css--default-type: true;
$css--plex: true;
//@import '@carbon/colors/scss/colors';

// Use all Carbon themes
@import "carbon-components/scss/globals/scss/vendor/@carbon/themes/scss";
@import "carbon-components/scss/globals/scss/component-tokens";


:root[theme="g100"] {

  @include carbon--theme(map-merge($carbon--theme--g100, (interactive-01: #ff0f0f,
  interactive-02: #ff0f0f,
        interactive-03: #ff0f0f,
        interactive-04: #ff4747,
        link-01: #ff7a7a,
        overlay-01: rgba(22, 22, 22, 0.7),
        hover-primary: #e70413,
        active-primary: #9e0003,
        hover-Primary-Text: #ffa8a8,
        hover-secondary: #606060,
        inverse-support-04: #ff0f0f,
        inverse-focus-ui: #ff0f0f,
        hover-ui: #353535,
        hover-selected-ui: #4c4c4c,
        hover-danger: #ba1b23,
        hover-row: #353535,
        highlight: #9e0028,
        focus: #ff0f0f, //should be the same as interactive 01

        skeleton-01: #353535)), true) {
  }
}

@import "carbon-components/scss/components/button/button";
```
The button background colour will not change as of the interactive-01 token.

Node: v14.17.1
OS: Windows 10
SASS: 1.35.1 & tested on SASS: v2

henryennis:
In the recent carbon-components package update 10.38.0, this safeguard was disabled resulting in theme tokens not being applied to components. This occurs when not using v11 flag.  There is no obvious documentation specifying that you need the feature flag for tokens to work so reenabling this until V11 would help other devs.

Tested in a local repo, tokens now applying.

Resolves #9055

### DIFF
--- a/packages/components/src/globals/scss/_theme-tokens.scss
+++ b/packages/components/src/globals/scss/_theme-tokens.scss
@@ -15,6 +15,104 @@
 // to make sure that they're in sync if one has diverged from the other.
 @include carbon--theme();
 
+
+//If the feature flag enable-v11-release is not enabled tokens will not work correctly.
+@if not feature-flag-enabled('enable-v11-release') {
+  $background: $ui-background;
+  $layer: $ui-01;
+  $layer-accent: $ui-03;
+  $field: $field-01;
+  $background-inverse: $inverse-02;
+  $background-brand: $interactive-01;
+  $interactive: $interactive-04;
+
+  $border-subtle: $ui-03;
+  $border-strong: $ui-04;
+  $border-inverse: $ui-05;
+  $border-interactive: $interactive-04;
+
+  $text-primary: $text-01;
+  $text-secondary: $text-02;
+  $text-placeholder: $text-03;
+  $text-helper: $text-05;
+  $text-on-color: $text-04;
+  $text-inverse: $inverse-01;
+
+  $link-primary: $link-01;
+  $link-secondary: $link-02;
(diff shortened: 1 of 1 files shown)


xSCIven:

xSCIven:
recheck

joshblack:
Hi there @xSCIven! 👋  This should be coming from the `carbon--theme` mixin directly, is that not the case in your project?

joshblack (approved):

joshblack:
Thanks for this @xSCIven! I opened up #9063 as a way we might address this, we may also just go for a full revert on our new token set since this has been problematic for teams over the past couple of releases. 

I'm going to go ahead and close this but we'll still use #9055 to track the fix for this issue. Thanks again for taking the time to create this PR!

PR #8684, opened 2021-05-13
revert(project): revert storybook v6

joshblack:
This PR reverts our storybook v6 changes due to complications in Netlify and other deployments 😞 

#### Changelog

**New**

**Changed**

**Removed**

- Remove changes related to storybook v6

PR #8630, opened 2021-05-10
Add storybook v6 support to `@carbon/react`

joshblack:
Closes #8554

This PR adds in the initial version of Storybook v6 to the `@carbon/react`. However, there are some known issues/limitations that we ran into that will be addressed in follow-up PRs:

- Issues
  - Welcome page styling issues
  - No base styles applied to custom MDX docs
  - Performance problems when importing from the `carbon-components-react` entrypoint
- Enhancements
  - Theme switcher using built-in Storybook functionality 
  - Locale switcher using built-in Storybook functionality
  - Deployments through Netlify
  - Add-ons panel 

#### Changelog

**New**

- Add storybook v6 support to `@carbon/react`

**Changed**

**Removed**

#### Testing / Reviewing

- `cd packages/carbon-react`
- Run `yarn storybook`
- Verify that the project runs and that you can visit the Grid story
- Verify that the reset styles are applied from the sass imports (it should use Plex instead of the default sans font for your OS)

PR #9062, opened 2021-06-29
chore: upgrade storybook from v5 to v6

### LINKED ISSUE #8883: ComboBox storybook props not visible or functional
ComboBox props are not working properly and sometimes not visible in the netlify builds and on local builds. Loading another story with props before returning to the ComboBox playground will display knobs but they will not function as expected. Reverting #8702 seems to resolve the issue but more investigation needs to be done to figure out why this specific component (so far) is not displaying properly in Storybook

emyarod:
Closes #8883 

related #8702

This PR updates the v5 Storybook to v6 for `packages/react` so that there are no clashes with the `carbon-react` storybook and so that component story knobs are functional again

#### Changelog

**Changed**

- upgrade `packages/react` storybook 
- regenerate lockfiles to update dependency versions
- fix outdated story imports

#### Testing / Reviewing

Confirm that component story knobs are functional again as well as the existing Carbon theme and type addon knobs


joshblack:
@emyarod is there a way to upgrade without regenerating the lockfile? Seems like we're getting a lot of big jumps with dependencies, could this be something that we use `yarn upgrade-interactive` to help out with?

emyarod:
@joshblack I tried not to touch the lockfile but that was the only way I could fix storybook build issues after migrating to v6. I suppose we can try and isolate the packages that need to be bumped with upgrade-interactive

joshblack:
@emyarod got it, it would be best to avoid doing a whole lockfile update and instead use `upgrade-interactive` along with `yarn why` to figure out why certain entries in the lockfile are old or duplicated which are causing the build to be a problem

joshblack:
Looking great! I think the only thing left is the deploy step which seems to be failing with:

```
ERR! Failed to load static files, no such directory: packages/react/.storybook/assets
ERR! Make sure this directory exists, or omit the -s (--static-dir) option.
```

emyarod:
looks like it was moved in https://github.com/carbon-design-system/carbon/pull/8630 but not reverted in https://github.com/carbon-design-system/carbon/pull/8684

joshblack:
@emyarod it looks like the color contrast on the selected story text might have changed?

image

Not sure if it's a new theme option or something 🤔

emyarod:
yeah seems to be caused by the new default storybook styles

joshblack (approved):

joshblack:
🎉

tw15egan (approved):
Everything looking/working great, nice work! 🎉

PR #9063, opened 2021-06-29
fix(themes): update themes mixin to use derived variables

joshblack:
This is a draft PR for a potential fix for #9055 

This is option one that we could take, the other option is a full revert of our token changes in `carbon-components`.

The problem comes from us getting the value in custom property scenarios from the theme itself and _not_ the `$button-primary` variable that is re-initialized. This PR updates our theme builder to instead use the variables that have been initialized (or re-initialized) instead of reading from the theme itself.

cc @tw15egan

### DIFF
--- a/packages/themes/__tests__/mixins-test.js
+++ b/packages/themes/__tests__/mixins-test.js
@@ -10,6 +10,7 @@
 'use strict';
 
 const { SassRenderer } = require('@carbon/test-utils/scss');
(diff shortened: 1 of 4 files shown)


joshblack:
Going to move forward with reverting the token updates:

PR one: https://github.com/carbon-design-system/carbon/pull/9064
PR two: coming soon

PR #9077, opened 2021-06-30
feat(eslint): add eslint-plugin-jest

### LINKED ISSUE #9047: Add the `eslint-plugin-jest` package to our eslint config
We would like to add in support for https://github.com/jest-community/eslint-plugin-jest in our eslint config. This issue is to add the rules, come up with a sensible set of defaults, and update the codebase to use them.

There is a question of if we should make these rules in the files that we ship outside of a major change, so it may be that we need to create a separate eslint entrypoint that we use so that users of our config don't start getting a lot of errors.

This issue is complete once the plugin has been added to our config and all rules that are enabled are passing.

joshblack:
Closes #9047 

This PR adds in `eslint-plugin-jest` to our ESLint config. It also sets up a separate `internal.js` file so that we can use new rules without impacting the Public API until we're ready to make the change.

#### Changelog

**New**

- Add jest config and `eslint-plugin-jest` to our eslint config package

**Changed**

- Update tests to new eslint rules

**Removed**

#### Testing / Reviewing

- Verify the ESLint rules are being applied
- Verify ESLint passes as expected


tw15egan (approved):
Tests are passing and rules are bing applied 👍🏻 ✅

jnm2377 (approved):

PR #9078, opened 2021-06-30
fix(@carbon/styles): remove / operator, use math.div

### LINKED ISSUE #8746: Update usage of `/` operator to use `math.div`
The `/` operator will be deprecated in Dart Sass `2.0.0` as pointed out in: https://github.com/carbon-design-system/carbon/issues/8069#issuecomment-846009017

This issue tracks the updates needed in `@carbon/styles` to move away from the `/` operator to `math.div` as has been recommended in: https://sass-lang.com/documentation/breaking-changes/slash-div

Unfortunately, our current Sass targets include `node-sass` and I don't believe this is something we can address in our codebase today. it will, however, be addressed in our v11 release.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8746

WIP, will update as more components are added to the `@carbon/styles` package

#### Changelog

**Changed**

- Replace `/` symbol with the `math.div` function or `*` where possible

#### Testing / Reviewing

Ensure styles remain the same with the changes and no deprecation warnings are thrown


tw15egan:
@joshblack seems like this is causing all the tests to fail in `@carbon/styles` since the test renderer is using `node-sass` under the hood. Any ideas?

joshblack:
@tw15egan hmm, if it's using `SassRenderer` for the tests it should be picking up `sass` 🤔  I wonder if our version is too old or something for the new math function?

As a side-note, would it make sense to start with these changes over in `packages/components` without `math.div` (basically just the multiplication change) and then update `@carbon/styles` to use `math.div` where multiplication doesn't make sense?

This would end up helping out teams who are making the move to dart sass now so they don't have to see warnings in their console all over the place.

tw15egan:
It looks like there are only a few places where we use the `/` sign directly, but I'll update those in this PR. I'm guessing the bulk of the warnings are coming from the `rem` function that will need to be replaced with `math.div`

tw15egan:
Seems like the `@carbon/grid` package is another big one that needs to be updated, as it is using `/` quite a bit in its calculations 
Screen Shot 2021-07-07 at 11 48 00 AM

PR #9081, opened 2021-06-30
fix(icons): update glyph only pipeline process

### LINKED ISSUE #9079: Glyphs in icons are being generated incorrectly in output extension
This issue ultimately derives from glyph icons being added back into the codebase. This caused some downstream issues in our builders, along with errors in our icon demo site. This issue is to figure out how we can correctly ship icon glyphs and make sure that each part of our icons stream works as expected.

joshblack:
Closes #9079 

Updates our icon pipeline to work with glyphs placed in `src/svg`.

#### Changelog

**New**

**Changed**

- Update `icon-build-helpers` to feed through the `glyph` size info throughout the pipeline

**Removed**

#### Testing / Reviewing

- Verify the elements preview works
- Verify glyphs are rendered as only the glyph and no extra sizes, unless those SVG assets are defined


joshblack:
Note to @carbon-design-system/release this will end up being a breaking change, but it should be okay because the glyphs were rendering as sizes that no one should be using / wouldn't be allowed by brand. Let me know if you think we should just do a major version bump for `@carbon/icons-react` or if we can just give people a heads up to change real quick in this release.

joshblack:
There will also be a follow-up PR to rename `glyph--` icons to just the base name 👍

tw15egan (approved):
Preview looks good to me 👍🏻

emyarod (approved):

PR #9090, opened 2021-07-01
fix(jest): update AC matcher to defer initialization

joshblack:
This is a small QOL fix for how we bring in AC. In our e2e tests, often this module would hang due to connection issues. This change will defer initializing AC until the matcher is first used which can help out with tests that don't use this matcher directly.

#### Changelog

**New**

**Changed**

- Defer AC initialization in Jest matcher

**Removed**

#### Testing / Reviewing


tw15egan (approved):
Tests still passing as expected

andreancardona (approved):
looks good!

PR #9104, opened 2021-07-02
fix(react): update radio button group state initialization

### LINKED ISSUE #9041: `RadioButtonGroup` mishandles the input falsy value as an undefined value

- [ ] `carbon-components`

## Detailed description

`RadioButtonGroup` mishandles the input falsy value as an undefined value.

`RadioButtonGroup`

> see changed?

When I write:

```javascript
<RadioButtonGroup {...props} valueSelected={0} defaultSelected={undefined}>
```

The inner state `selected` should be `0` instead of `undefined`.

* https://codesandbox.io/s/loving-currying-pw2wp?file=/src/index.js
* When clicking "select value 0" button, the first radio button should be selected. But it will leave blank with the current implementation.

> Workaround

Set `selectedValue` and `defaultSelected` with the same value.

## Additional information

- One of problematic implementations: https://github.com/carbon-design-system/carbon/blob/62f2e5d012d1ef74e6dc8b7b6087147223c7c586/packages/react/src/components/RadioButtonGroup/RadioButtonGroup.js#L17

joshblack:
Closes #9041 

This updates how we initialize our `selected` state value from a `||` expression to a ternary that checks if the value is undefined. When using `||`, the state would be incorrectly initialized if the value was false-y but not undefined (like the value `0`).

I also took this small change to convert some of the tests for `RadioButtonGroup` to RTL. It also updates the `jest-dom` matchers to be brought in automatically.

#### Changelog

**New**

**Changed**

- Update `RadioButtonGroup.js` to correctly initialize `selected` state
- Update `RadioButtonGroup-test.js` file to start using RTL
- Update `jest-config-carbon` to automatically bring in `jest-dom` matchers

**Removed**

#### Testing / Reviewing

- Review the test case and make sure it mirrors the one brought up in #9041 
- Verify test passes
- Verify component continues to work in Storybook


tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

PR #9107, opened 2021-07-03
fix(components): add missing g80 theme notification layer tokens #9106

### LINKED ISSUE #9106: Gray 80 low contrast notification incorrect background color?

- [x] `carbon-components`

## Detailed description

It appears that the background color for a Gray 80 theme low contrast notification is too light. The g90/g100 themes use a dark background with light text.

Screen Shot 2021-07-03 at 4 52 55 AM

**g80**

```css
.bx--inline-notification--low-contrast.bx--inline-notification--error {
    border-left: 3px solid var(--cds-support-error,#da1e28);
    background: var(--cds-notification-background-error,#fff1f1);
}
```

**g90**

```css
.bx--inline-notification--low-contrast.bx--inline-notification--error {
    border-left: 3px solid var(--cds-support-error,#da1e28);
    background: var(--cds-notification-background-error,#fff1f1);
}
```

## Possible solution

A possible solution may be to add the `g80` theme layer tokens to `components/src/components/notification/_tokens.scss`: https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/components/notification/_tokens.scss

See PR #9107 

1. Visit https://carbon-svelte.vercel.app/components/InlineNotification?theme=g80#low-contrast

metonym:
Closes #9106 

This PR adds the missing `g80` layer tokens to the `carbon-components` notification tokens.

#### Changelog

**Changed**

- fix(components): add missing g80 theme notification layer tokens

#### Testing / Reviewing


joshblack:
Hi @metonym! 👋 

If I remember correctly, the g80 theme was part of our experimental work for the new design token direction we were going. That experimental work ended up with us taking another direction for theming and so the g80 theme is no longer needed and will be removed.

Let me know if the g80 theme is something that you all are depending on at the moment or if there is another use case for this that you would like to bring up 👀

metonym:
@joshblack Thank you for the response Josh.

The Carbon Svelte library is depending on g80 at the moment. Can it be preserved until v11?

joshblack:
Yeah definitely, we can keep it around. Just curious, what are you all using it for? 👀

joshblack (approved):

metonym:
@joshblack Currently, we offer it as another Carbon theme.

It's shipped as a pre-compiled CSS file: https://unpkg.com/browse/carbon-components-svelte@0.39.0/css/

joshblack:
Got it 👍  Thanks!

In the future, what would work best on our end to flag things in Sass as experimental/unstable so that the level of support for functionality/features is more obvious?

metonym:
I think that the existing approach is fine. For us, we've been ambitious in adding as many features from `carbon-components` and `carbon-components-react`, even if they are marked as unstable and experimental within the scope of major version 10.

dakahn (approved):

PR #9116, opened 2021-07-05
fix(data-table): deprecate shouldShowBorder prop

metonym:
This deprecates the `shouldShowBorder` prop in `Table` as it appears that toggling the `.bx--data-table--no-border` has no visual effect. I've checked the compiled `carbon-components` CSS and the selector does not exist.

#### Changelog

**Changed**

- fix(data-table): deprecate Table `shouldShowBorder` prop


tw15egan (approved):

packages/react/src/components/DataTable/Table.js
```diff
@@ -70,7 +71,12 @@ Table.propTypes = {
   /**
    * `false` If true, will remove the table border
    */
-  shouldShowBorder: PropTypes.bool,
+  shouldShowBorder: deprecate(
+    PropTypes.bool,
+    'The `shouldShowBorder` prop has been deprecated.' +
```
joshblack:
```suggestion
    'The `shouldShowBorder` prop has been deprecated and can be safely removed.' +
```

Might be helpful to indicate the "next best action" in the deprecation message. In this case, it seems like we want to share that the user can safely remove this prop as it no longer has an effect.

Suggestion is just one idea for wording this, curious what you think 👀

joshblack (approved):

PR #9121, opened 2021-07-06
fix(NumberInput): allow value to be outside min/max, show validation

### LINKED ISSUE #8965: [NumberInput] When passed value is out of boundary, the value is changed to equal min/max

- [X ] `carbon-components-react`

## Detailed description

Scenario : When a value in passed in, but is out of boundary (greater than max or less than min value), the value is set to the min or max boundry.  This can cause data integrity issues if value is set on page load or comes in from an outside calculation, rather than being set by the user.

> Is this issue related to a specific component

Number Input 
https://react.carbondesignsystem.com/?path=/story/components-numberinput--default

> see changed?

I would expect that if a value is passed to the component that is outside of the min or max value, the exact value passed is displayed, but a field validation error is also shown. This would be consistent with what currently happens if I try to manually type in a value into the component that is outside of the boundary.

Chrome 

Current

1. Go to storybook https://react.carbondesignsystem.com/?path=/story/components-numberinput--default
2. Note that knobs have min=0 , max =100
3. Set "value" prop to 999 
Result: Value is changed to 100. No error is shown.
result

>
https://github.com/carbon-design-system/carbon/blob/fb5d24e5c6b02a9a11d595181b3ec72a3d2fd20d/packages/react/src/components/NumberInput/NumberInput.js#L42

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8965

Allows a `value` prop to be passed in that is larger than the `min` and `max` bounds, instead of capping at the `min` or `max`. If a value is passed in that exceeds these limits, the validation is shown. If the number steppers are used when an invalid number is shown, it will skip the closest valid number.

#### Changelog

**Removed**

- Removed value capping when a `value` is passed directly to `NumberInput` 

#### Testing / Reviewing

In the playground, set the `value` knob to a number outside the min/max value. The validation should be shown, and the stepper should prevent an invalid increment/decrement.


jnm2377 (approved):
LGTM! 🔥

sstrubberg (approved):

PR #9132, opened 2021-07-07
fix(Slider): check validity of number input after onChange

### LINKED ISSUE #8842: Slider Component input box bug

- [x] `carbon-components`

## Detailed description

There is a bug when trying to type in a number into the input box for the slider

Slider Component

> What did you expect to happen? What happened instead? 

If you, for example, have min 30, and your current value is 35, if you want to change it to 34, the box wont allow you to delete the ‘4’ as the value would be ‘3’, even though you are still editing the number.

> What would you like to see changed?

Might be better to delay the validation until the user has left the input box, or mark the box as invalid and not automatically change the value

Chrome

carbon-components:  10.35.0,
 carbon-components-react: 7.35.0

jnm2377:
Closes #8842 

This PR updates `Slider` so that form validation happens after any changes are made. Previously, number input wouldn't allow you to backspace if the number would be invalid, even if you were changing it to something valid.
> If, for example, you have min 30, and your current value is 35, if you want to change it to 34, the box wont allow you to delete the ‘4’ as the value would be ‘3’, even though you are still editing the number.

#### Changelog

**New**

- `isValid` state
- `onBlur` event handler for form validation

**Changed**

- `onChange` handler no longer has form validation 
- fixed padding for slider invalid input (had extra padding from regular number input invalid styles which have an invalid icon, but slider does not have icon).

#### Testing / Reviewing

- For testing, go to the `Slider` story, and use the text input to type an invalid number (test both less than the min and more than the max). After clicking away, the invalid state should appear. Then update the text input to a valid number and click away. The invalid state should disappear now.


packages/react/src/components/Slider/Slider-story.js
```diff
@@ -58,7 +58,7 @@ export const Default = () => (
   <Slider
     labelText="Slider Label"
     value={50}
-    min={0}
+    min={30}
```
emyarod, line 61: is this story knob change intentional?

emyarod (commented):
it seems like even if an `invalid="false"` prop is provided the component will be rendered as invalid no matter what, so I think we need to handle props and state sync

https://user-images.githubusercontent.com/8265238/124783434-ae304c80-df0a-11eb-946c-eb4dcafbdd20.mov

I believe this would fix the interactions with the controlled slider example as well

https://user-images.githubusercontent.com/8265238/124783975-1a12b500-df0b-11eb-9811-bc923a91bbe2.mov

packages/react/src/components/Slider/Slider-story.js
jnm2377: Yes, I updated it to test easier for the min, and also to reproduce the bug. We can revert this before merging.

jnm2377:
> it seems like even if an `invalid="false"` prop is provided the component will be rendered as invalid no matter what, so I think we need to handle props and state sync
> 

@emyarod Do you have any suggestions on how to do that? I guess, I'm thinking maybe kinda how we merge refs in ComboBox? Is that what you meant?
https://github.com/carbon-design-system/carbon/blob/311c4ab776624d175d5a983b166083f37c57ba71/packages/react/src/components/ComboBox/ComboBox.js#L311

emyarod:
most of our class components will perform a props/state sync in `getDerivedStateFromProps` along with any relevant change handlers. maybe the NumberInput would be a good reference to look at due to the similarities to this

sstrubberg (approved):

dakahn (approved):

PR #9138, opened 2021-07-07
fix(select): use non transparent inline select background

### LINKED ISSUE #9112: Inline Select component needs background in Chrome/Windows

- [x] `carbon-components`

## Detailed description

In Chrome on Windows a `<select>` background needs to be set to fit the `<color>`, otherwise the options can become unreadable. This is very OS and browser-specific; e.g. it is not a problem for Firefox on Windows or Chrome on Mac.

Affected is the `<Select>` component with the inline class set without mouse hover in a dark theme. E.g. used in `<Pagination>`.

With hover (CSS sets background):
image

Without hover:
image

1. Use dark theme
2. Add `<Pagination>`
3. Open one of the dropdowns
4. Move mouse away

The [official component page](https://www.carbondesignsystem.com/components/pagination/usage/) demonstrates this behavior.

## Additional information

Relevant CSS that seems to cause this:
```css
.bx--select--inline .bx--select-input {
    ...
    ...
}
```

The select already has an explicit background otherwise.

emyarod:
Closes #9112

This PR sets a ui-background color for inline selects so that themed select options on Windows will be visible (Alternatively we can set explicit values on the select options and optgroups but the menus themselves may still be discolored because of how different OS and browser combinations handle menu rendering and CSS custom properties)

#### Testing / Reviewing

Confirm that the select options and option groups are visible on dark Carbon themes in all browser and OS combinations


joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

dakahn (approved):

PR #9139, opened 2021-07-07
fix(HeaderMenu): spread extra props into root element

### LINKED ISSUE #9134: React HeaderMenu does not render data-* attributes
## Brief description

When integrating Carbon with automated UI testing tools such as Cypress or Selenium, the recommended approach is to use `data-*` attributes such as `data-cy` to allow the UI testing framework to select the correct element for automation. The React `<HeaderMenu>` component does not correctly render `data-*` attributes.

By contrast other react components (such as `<Header>`) do correctly propagate `data-*` attributes.

- [x ] `carbon-components`
- [x ] `carbon-components-react`

## Detailed description

Using the following markup

```
<Header aria-label="navbar" data-cy="navbar">
        <HeaderName href="/">My App</HeaderName>
        <HeaderGlobalBar>
          <HeaderNavigation aria-label="navigation">
            <HeaderMenu className="entity" aria-label="Entities" menuLinkName="Entities" data-cy="entity">
            ...
            </HeaderMenu>
         </HeaderNavigation>
        </HeaderGlobalBar>
</Header>
```

The expectation is that the rendered HTML for the `<HeaderMenu>` will contain the attribute `data-cy="entity"`; instead the markup is missing.

Yes - HeaderMenu. I can see the feature working for other components (such as Header), but I have not made an explicit survey of the full toolkit.

> see changed?

See above.

N/A - issue is visible by inspecting the raw HTML.

```
    "carbon-components": "^10.38.0",
    "carbon-components-react": "^7.38.0",
```

> should be aware of?

GHHS SPM AI Sept 2021. 

See above.

Issue is visible by inspecting the markup generated in this code sandbox:
https://codesandbox.io/s/vibrant-gareth-m5q2g?file=/src/index.js

emyarod:
Closes #9134

This PR allows for additional props to be spread onto the UI Shell HeaderMenu which should improve support for testing libraries like Cypress which inject data attributes

#### Testing / Reviewing

Confirm that any extra props placed on the UI Shell HeaderMenu are ultimately rendered in the DOM


tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):

PR #9149, opened 2021-07-08
feat(styles): add support for layer sets

joshblack:
This PR adds in support for layer sets to `packages/styles`. Layer sets contain several tokens that map to a shorthand version of that token, for example: `border-strong` maps to `border-strong-01`, `border-strong-02`, or `border-strong-03` depending on which `layer` it is rendered in.

The code in this PR adds in support to generate these mappings and will automatically redefine them using the `.#{$prefix}--layer` selector. By default, the tokens will map to the first level and are defined on `:root` selector, the `.#{$prefix}--layer` will map to the second level, and `.#{$prefix}--layer .#{$prefix}--layer` will map to the third level in the layer set.

#### Changelog

**New**

- Add support for layer sets in scss/utilities/_layer-set.scss
- Add support for CSS Custom Property helpers in scss/utilities/_custom-property.scss

**Changed**

**Removed**

#### Testing / Reviewing

- Review docs/comments to make sure things make sense to maintainers


tw15egan:
Would it be possible to add a story to the `@carbon/react` storybook to showcase the functionality / verify it's working as intended?

joshblack:
@tw15egan I think the test case can help out with seeing what gets generated. Kind of next steps on my end for this was to use this module after the theme update PR got merged in and then we need to go through and update the components to use it.

In the interim, to see what it does the test basically looks like:

```scss
@use '../layer-set' with (
  $layer-sets: (
    field: (
      rgba(0, 0, 0, 0.3),
      rgba(0, 0, 0, 0.2),
      rgba(0, 0, 0, 0.1),
    ),
    background: (
      rgba(0, 0, 0, 0.8),
      rgba(0, 0, 0, 0.7),
      rgba(0, 0, 0, 0.6),
    ),
  )
);
```

Which then outputs:

```css
:root {
  --cds-field: rgba(0, 0, 0, 0.3);
  --cds-background: rgba(0, 0, 0, 0.8);
}

.cds--layer {
  --cds-field: rgba(0, 0, 0, 0.2);
  --cds-background: rgba(0, 0, 0, 0.7);
}

.cds--layer .cds--layer {
  --cds-field: rgba(0, 0, 0, 0.1);
  --cds-background: rgba(0, 0, 0, 0.6);
}
```

So now you can use `var(--cds-field)`, or simply `$field`, and it will map to the correct value depending on if at the top-level or nested in a `cds--layer` class.

tw15egan:
Makes sense! Was just thinking of examples to show off its functionality, definitely doesn't need to be in this PR. Everything else looks great!

tw15egan (approved):
LGTM 👍 ✅

sstrubberg (approved):
🔥

PR #9157, opened 2021-07-09
feat(FilterableMultiselect): add optional autoComplete prop

### LINKED ISSUE #8993: [FilterableMultiSelect]: `autocomplete="off"` ignored by Chrome

- [ ] `carbon-components`

## Detailed description

It seems like [Chrome has made the choice](https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164) to ignore `autocomplete="off"` on inputs, meaning that _sometimes_ FilterableMultisect will try to autocomplete inputs that happen to be labelled by something that would be in an address like "Name" or "State". To be honest, I wasn't able to reproduce this but one of Cloud PAL's users encountered this with "State"

image (4)
  
Like I said, I wasn't able to reproduce this, but given that Chrome explicitly said that they ignore `autocomplete="off"` it seems likely that more people could run into this problem 

## Additional information

A solution, according to Chrome, is if you put anything as the value for `autocomplete` that they _don't_ recognize, they autofill won't be suggested. 

"In cases where you really want to disable autofill, our suggestion at this point is to utilize the autocomplete attribute to give valid, semantic meaning to your fields. If we encounter an autocomplete attribute that we don't recognize, we won't try and fill it.

As an example, if you have an address input field in your CRM tool that you don't want Chrome to Autofill, you can give it semantic meaning that makes sense relative to what you're asking for: e.g. autocomplete="new-user-street-address".  If Chrome encounters that, it won't try and autofill the field."

dakahn:
Closes #8993 

[Chrome no longer supports the HTML5 attribute autocomplete's `off` configuration](https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164).  

As a workaround this PR adds an optional prop to FilterableMultiselect called `autoComplete` allowing Chrome users to utilize the _unique_ browser feature of passing a string describing the expected input (eg `autoComplete="super-secret-password"`) in order to disable autocomplete. A method described in the bug report closure above as
> ...if you have an address input field in your CRM tool that you don't want Chrome to Autofill, you can give it semantic meaning that makes sense relative to what you're asking for: e.g. autocomplete="new-user-street-address".  If Chrome encounters that, it won't try and autofill the field.

#### Testing / Reviewing

1. pull down and checkout the PR
2. start the react Storybook and observe `autocomplete="false"` on the Filterable Multiselect's input
3. in react/.../Multiselect/MultiSelect-story.js on line 195 add the prop `autoComplete="user-input"
4. in the React Storybook observe `autocomplete="user-input"` on the Filterable Multiselect's input

@abbeyhrt _maybe have your user give this spin and see if they can still recreate the issue_

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -3578,6 +3578,9 @@ Map {
         "ariaLabel": Object {
           "type": "string",
         },
+        "autoComplete": Object {
+          "type": "string",
+        },
         "compareItems": Object {
           "isRequired": true,
           "type": "func",
--- a/packages/react/src/components/MultiSelect/FilterableMultiSelect.js
+++ b/packages/react/src/components/MultiSelect/FilterableMultiSelect.js
@@ -36,6 +36,11 @@ export default class FilterableMultiSelect extends React.Component {
     ariaLabel: PropTypes.string,
 
     /**
+     * Specify autocomplete attribute value. Defaults to 'off'
+     */
+    autoComplete: PropTypes.string,
+
+    /**
      * Specify the direction of the multiselect dropdown. Can be either top or bottom.
      */
     direction: PropTypes.oneOf(['top', 'bottom']),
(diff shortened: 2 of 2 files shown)


joshblack (changes requested):
Would it make sense for us to specify a "custom value" for the autocomplete instead of making it a prop?

PR #9226, opened 2021-07-19
fix(BreadcrumbItem): remove href check for children of type string

### LINKED ISSUE #9224: React Breadcrumb uncaught error if string without href

- [ ] `carbon-components`

## Detailed description

The breadcrumb has the option of last item being the current page. In this scenario the breadcrumb may well be a string with no href supplied. This however produces an unhandled error.

e.g.
```
    <BreadcrumbItem isCurrentPage={true}>
      The current page
    </BreadcrumbItem>
```

produces

image

It is possible to work around this by wrapping the text in a an element e.g.

```
    <BreadcrumbItem isCurrentPage={true}>
      <span>The current page</span>
    </BreadcrumbItem>
```

BreadcrumbItem

> see changed?

No unhandled error should occur

Chrome

10.38

> should be aware of?

IBM Cloud Cognitive components package

See above.

emyarod:
Closes #9224

This PR resolves an uncaught type error where breadcrumbs without `href` s and with children of type string were incorrectly being processed as React elements

#### Changelog

**Changed**

- breadcrumb item typechecks
- add additional breadcrumb item to storybook examples

#### Testing / Reviewing

Confirm the breadcrumb stories function as expected with the added breadcrumb items


andreancardona (approved):
🎉

tw15egan:
If a breadcrumb does not have a link, should it still be styled as a link?

emyarod:
bump @jnm2377 when you have a chance

emyarod:
I believe so assuming it is still interactive, but that would be left up to the application to make sure

jnm2377 (approved):

PR #9236, opened 2021-07-20
feat(carbon-react): add Layer component

### LINKED ISSUE #9185: Build out a `Layer` component as a replacement for the `light` prop
This work builds on top of the contextual tokens work that's been implemented for `@carbon/styles`. `Layer` will be a simple component that will be a replacement for the `light` prop. the way it works is that it will be a simple container that will render the `cds--layer` class name. This class name remaps any contextual tokens that are in play to the next step in their layer set.

```jsx
// Before
<Dropdown light />

// After
<Layer>
  <Dropdown />
</Layer>
```

The `Layer` component can be configured to be any element type that the user wants.

This work should include usage documentation.

joshblack:
Closes #9185 

This PR adds in the `Layer` component and associated layer set tokens in `@carbon/styles`.

#### Changelog

**New**

- Add `Layer` component to `@carbon/react`

**Changed**

- Update `scss/_theme.scss` to include layer sets

**Removed**

#### Testing / Reviewing

- View the storybook story and make sure that the `var(--bx-layer)` token is changing as intended across each theme we support


sstrubberg (approved):

emyarod (approved):
looks good to me

Issue #9180, opened 2021-07-12
Create migration doc that specifies how to move from tokens in v10 to v11


tw15egan:
We'll need to document how a team that is currently utilizing the v10 tokens can update their usage to the new v11 tokens. The notion doc from @aagonzales should help with this 

Notion doc: https://www.notion.so/carbondesignsystem/Final-List-of-new-tokens-fc5985c7331341baa6546064399d36da
Component specs: https://github.com/carbon-design-system/carbon/issues/8187

PR #9245, opened 2021-07-21
docs: add v11 color token tables

johnbister:
#### Changelog

**New**

-Adds new theming and color token tables for v11 migration


tw15egan:
Looks like this will close https://github.com/carbon-design-system/carbon/issues/9180 🎉 💯

docs/migration/11.x-color.md
```diff
@@ -0,0 +1,178 @@
...
+
+| **Contextual tokens**     |                    |              |                                                                                            |                                          |                                          |                                          |                                          |                                          |
+|---------------------------|--------------------|--------------|--------------------------------------------------------------------------------------------|------------------------------------------|------------------------------------------|------------------------------------------|------------------------------------------|------------------------------------------|
+| **New Token Name**        | **Old Token Name** | **Status**   | **Usage**                                                                                  | **White**                                | **Gray 10**                              | **Gray 80**                              | **Gray 90**                              | **Gray 100**                             |
```
jnm2377: This looks great!! Just have a few suggestions. I think it would be good to refer to these as v10 and v11 rather than new and old. I also think maybe it would be better to have v10 column first, and then v11. I feel like that's the typical order we do for other migration docs so that people can find what they are "currently" on and then see what needs to change. Also, not sure if the status column is really needed. Is that useful for designers? I think on the dev side it's not important, but if design needs it then makes sense to leave it. 
```suggestion
| **v11**        | **v10** | **Status**   | **Usage**                                                                                  | **White**                                | **Gray 10**                              | **Gray 80**                              | **Gray 90**                              | **Gray 100**                             |
```

docs/migration/11.x-color.md
```diff
@@ -0,0 +1,178 @@
+# Color Tokens
+
+Note: everything in this file is WIP.
```
andreancardona:
```suggestion
<InlineNotification>

**Note:** everything in this file is WIP.

</InlineNotification>
```

andreancardona:
🎉

aagonzales:
I'm not seeing a website preview link?

jnm2377:
@aagonzales these migrations docs live in the monorepo, not the website. We have a different draft PR for the website docs.

jnm2377 (approved):

laurenmrice (approved):

PR #9257, opened 2021-07-22
feat(Popover): add to @carbon/styles

dakahn:
Adds Popover to @carbon/styles

#### Testing / Reviewing

Check out how i'm bringing in popover in `Popover/index.js`. Almost certain this isn't how we want to do this. Also @joshblack as for _showing usage_ in a Story -- I used the same style as we had in experimental, but that might be more than we want for stable lemme know.


packages/carbon-react/src/components/Popover/index.js
```diff
@@ -0,0 +1,144 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { settings } from 'carbon-components';
```
joshblack: I think this is fine if you want to just remove `packages/react/src/components/Popover/index.js` so the source of truth lives in `@carbon/react`

joshblack (approved):

andreancardona (approved):

andreancardona:
@dakahn I think testing needs to be updated

PR #9325, opened 2021-07-26
fix(font-family): added system fonts

### LINKED ISSUE #8764: @carbon/type: determine ideal font family stack, and add as mixin to type package
Based on https://github.com/carbon-design-system/carbon/pull/8726#pullrequestreview-665660046 and https://github.com/carbon-design-system/carbon/pull/8726#pullrequestreview-665661296, we should determine an ideal `font-family` font stack, and utilize this as part of our `@carbon/type` package. This could then be consumed by the `@ibm/plex` and `@carbon/react` packages

sstrubberg:
Closes #8764 

Updates font-families in the `@carbon/type` package with the following:

- Arabic
- Devanagari
- JP
- KR
- Thai

Replaced all the `Helvetica Neue` stuff with `system-ui, -apple-system, BlinkMacSystemFont, '.SFNSText-Regular', sans-serif`


joshblack:
@sstrubberg I think we can keep the existing font stacks the same but use the system UI ones for the new fonts that we're going to be shipping through the Plex package. I think these include:

- Arabic
- Devanagari
- JP
- KR
- Thai

What do you think? 👀

joshblack (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #9340, opened 2021-07-27
feat(ProgressIndicator): add animation

### LINKED ISSUE #9156: Progress Indicator Animations
### Summary

Requesting to enhance the progress indicator component with animations showing the blue line progressing as steps are completed and the icons change to help the user better understand their progress as they complete forms. Also, requesting a slight change to the "optional text" under each text to instead a show a summary of what the user selected in that step.

This would require motion design and development work.

### Justification

This would reduce the cognitive load of the user and help them better make connections between the steps. The addition of the "summary text" could also provide a "summary" step for our users - allowing them to review progress made without going back and taking additional steps. 

When a user completes a step, the progress line would animate to show their progress. At the same time, the summary text would fade in as well as the blue step indicator would transform to the check mark. Here's a mock up of how this could work:
Horizontal Progress Tracker
Progress Tracker 01

Success/failure could be measured through user testing the current implementation of this component with the proposed changes OR how current forms work without vs with and user's preferences. 

### "Must have" functionality

- Blue line/icons animating

Some additional functionality to consider:

- Creating anchor links in the indicator so users can easily revisit steps
- Showing error states in the progress indicator steps when users try to complete the form or miss a section

no specific time line

### Available extra resources

I (@lisa-j-kaiser) am happy to help with any design/animation and @kristastarr (krista.starr@ibm.com) is working on development of this component to contribute back.

kristastarr:
Closes #9156 

Progress Indicator Animations: 
- The blue line at the top of a completed step animates from left to right. 
- The "current" and "completed" status icons transition in opacity when they appear and disappear. 
- There is a subtle transition on the optional label text on initial page load. 

#### Changelog

**New**
Added keyframes and transitions to _progress-indicator.scss

**Changed**
Changed the ProgressIndicator component in the React package - this change would require a `StepData` object to be passed in as a prop to the `ProgressIndicator` component, in order `ProgressIndicator` to render a `ProgressStep`. (Currently the `ProgressStep`s are being rendered as `React.Children` in `ProgressIndicator`).

Rationale: Properties of `this.props.children` cannot be modified, so currently are being re-rendered using `React.cloneElement` with the updated "current" or "completed" property each time the CurrentIndex changed.  When I added the CSS for the animations, this caused all the icons to animate every single time there is a change to the status of any single `ProgressStep`.  Rendering `ProgressStep` components from inside the `ProgressIndicator` enabled me to modify the "current" and "completed" properties of each `ProgressStep` to trigger a re-render of a step/icon only when there is a change to that specific `ProgressStep`.

#### Testing / Reviewing
You can see the changes by running the Carbon React Storybook. 

recheck


kristastarr:

johnbister:
The motion looks great to me @kristastarr !  Do you think we should include what errors and disabled states would look like in this pattern as well. For instance, if the user cannot click ahead a few steps because prior steps determine what is displayed in subsequent steps.  There was a great discussion with @lisajkaiser about this last week and I think it would be great to include these scenarios as well.

joshblack (changes requested):
The animation looks slick! I think the only big note from the implementation is that it seems like the PR makes a change to `ProgressIndicator` where it reverts it to a class component versus the function component that we refactored it to recently.

From a motion perspective, will we need to add in support for disabling this animation when the user has a no motion preference? 👀

kristastarr:
@joshblack Absolutely, I added media queries to disable the animations for users who prefer reduced motion.

packages/react/src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -26,208 +26,219 @@ const defaultTranslations = {
...
-  });
-
-  const handleKeyDown = (e) => {
-    if (matches(e, [keys.Enter, keys.Space])) {
-      onClick();
-    }
+export class ProgressStep extends React.PureComponent {
+  constructor(props) {
+    super(props);
+  }
```
jnm2377, line 36: Still trying to figure out what the status of this PR is, but we're actually working on making our components functional components for v11. Devs agreed that it makes more sense esp since most devs these days work with the latest React. We don't want this to be a class component.

jnm2377 (changes requested):

packages/react/src/components/ProgressIndicator/ProgressIndicator.js
jnm2377: ref: https://github.com/carbon-design-system/carbon/issues/9712

kristastarr: Hey, sorry, it looked like I needed to re-write some of the tests to get them to pass and unfortunately I haven't done it yet.  I understand what you're saying about functional components, but I could not find a way to apply the animations without making it into a class component. Do you want to close this PR for the moment?  I still plan to fix the failing tests, and I will review and seek advice on the feasibility of accomplishing the animation on a functional component.

jnm2377: I'm curious what issues you were running into with the functional component, and also if there are other ways we might be able to implement this without breaking changes. Referring to the `stepData` prop. Why is that needed instead of handling the props through `ProgressStep`? It's not our typical API pattern for data of parent/child components like `Tabs`, `Accordion`, `ContentSwitcher`. I think there's a bigger discussion to be had about the implementation.

kristastarr: Please bear with my long explanation here :sweat_smile:   I'll try to explain how/why I ended up doing it the way I did, and maybe we can come to a better solution - any thoughts or suggestions are greatly appreciated!  I definitely agree it would be preferable to not introduce any breaking changes.  

The issue I was having was how to prevent a re-render of all the `ProgressStep` components each time there was a change to the `ProgressIndicator`'s props.  Once I added in the CSS keyframes animations, the animations were re-triggered every time each `ProgressStep` was re-rendered, which caused a weird flashing effect (see below although please note that that is an earlier version that did not yet include the line animations, only the icon animations).  The goal was for each `ProgressStep` & its icon to re-render only when there was a change to its status.  


https://user-images.githubusercontent.com/21047571/137981027-2c5dcc5f-6ddc-4937-b7a8-3e4c3c746a04.mp4


In the current implementation of ProgressIndicator.js, `renderSteps()` clones each child element each time there is a change to the `ProgressIndicator`'s `currentIndex`, and so the `cloneElement()` is what was triggering the re-render.  Instead of cloning the element, I wanted the element to persist but just change it's props if necessary.  I first thought to use `React.memo()` but from my understanding, that is not something that can be used with React.Children.  So that's how I ended up passing the StepData in to render the `ProgressStep`s with the newly added `isCurrent` and `isComplete` props, rather than rendering them as cloned React.Children.  The solution I found was to use `React.PureComponent` to make use of its comparison of those props to only re-render a `ProgressStep` if it changes, hence the Class component. 

As a next step, I addressed the lint and format errors, and I will take a look at the typical pattern for for handling parent/child components in the components you mentioned. (The PublicAPI-test is still failing but since we may eliminate the changes causing that fail, I will address it later if need be.) @matthewgallo will take a look at this as well and we will hopefully find a solution that will work as a functional component and also one without any breaking changes.

jnm2377: Hey @kristastarr sorry for the late reply. Thank you for that detailed explanation, it's very helpful. I'd really like to help get this into production since it's something our designers are interested in. I think going forward, we can do a few things:

1. Make this an RFC or create a separate RFC to get more input from the rest of our dev team on the implementation. 
2. I am also planning on bringing this up in our next dev team meeting (this week), and hopefully get some thoughts on how we feel about the API changes. Maybe even think about alternate solutions.
3. Based on feedback and input from the rest of the team, we can decide what to do, how to help, etc.

How does that sound?

jnm2377:
Hey @kristastarr After talking to our dev team, we would prefer for the component to keep the existing API, where the steps are rendered as children and not a prop data array. We've just found that our components that follow the data prop pattern can be a pain both for users and maintainers. It can just limit people sometimes. So our requests would be to keep the functional component and API as is. 

It seems that this PR hasn't had some activity for quite some time, so I'm going to close this for now. Please feel free to open another PR with these changes once you're ready. 🙂

kristastarr:
I am still planning to work on implementing the animation with a functional component.  If yes, I will check in with y’all to see if I should do a new PR to Carbon, and if not we will plan to incorporate it into the Carbon for IBM products.

PR #9346, opened 2021-07-28
hot-fix(pagination): fix style bug

andreancardona:
This is a hot fix reverts this PR: https://github.com/carbon-design-system/carbon/pull/9138

That _did not_ fix the original issue: https://github.com/carbon-design-system/carbon/issues/9112

image (5)


joshblack (approved):

tw15egan:
This was recently added in via https://github.com/carbon-design-system/carbon/pull/9138. Is there an alternative fix for #9138 that doesn't break the component in other use-cases?

andreancardona:
@tw15egan good question. 

The first time I see the bug is in this PR actually: https://github.com/carbon-design-system/carbon/pull/9149

tw15egan:
It's fine if there isn't, we should just re-open the other issue so we don't forget about it 👍🏻

tw15egan (approved):

andreancardona:
@tw15egan Ooo one other think. I just asked D.A. to look at this bug again, and it seems it's still not fixed: https://github.com/carbon-design-system/carbon/pull/9138

@tw15egan I don't have Windows support. Can you confirm this bug still exists? 

Also it seems to be a bug in `svelte` and `vanilla`

Am I understanding that correctly? 👀

tw15egan:
I don't currently have access to a Windows machine, but if that fix didn't work then lets definitely revert it so that the pagination bug is fixed

tw15egan (approved):
LGTM 👍 ✅

andreancardona:
@tw15egan re-requested your review. Reverting the original PR: https://github.com/carbon-design-system/carbon/pull/9138
Sorry for the confusion!

PR #9355, opened 2021-07-28
feat(experimental-pagination): add onchange handler

### LINKED ISSUE #9313: unstable_Pagination does not fire onChange for next and previous buttons
Pagination component has a performance issue when data is huge.

#162

As an alternative we are trying to use unstable_Pagination, but it does not trigger onChange on click of next and previous buttons.

Any other alternatives to handle the performance issue with Pagination is also welcome.

We are on "@carbon/ibm-security": "1.41.0",

ibm-security issue reference -> https://github.com/carbon-design-system/ibm-security/issues/1051

tay1orjones:
Closes #9313 

Adds an `onChange` prop/handler to `unstable_Pagination`. The signature mirrors the callback available in the stable `Pagination`.

#### Changelog

**New**

- unstable_Pagination: added onChange handler

#### Testing / Reviewing

In the storybook, the unstable_Pagination/Playground story should fire an onChange action when next/prev buttons are clicked.


andreancardona (approved):

jnm2377 (approved):
🔥

PR #9364, opened 2021-07-29
fix(UIShell): update story to show proper right panel transition

### LINKED ISSUE #9095: Transition in right sidebar not working
Transitions applied via css class to headerPanel are ignored.

Using the HeaderContainer along with its render method from the react carbon components story causes header panel transitions to be ignored

Credit:
https://stackoverflow.com/questions/62703153/react-carbon-design-system-transition-in-right-sidebar-not-working

Additional information:
Got transitions working by removing the headerContainer and making use of react useState to manage the state of sideNav.

Furthermore the headerContainer component should not exist at all as state management should be done through your framework.

unsure if this is a bug or the examples in carbon react story needs to be updated.

References:
https://react.carbondesignsystem.com/?path=/story/components-ui-shell--header-base-w-navigation-actions-and-side-nav

Using chrome browser: Version 91.0.4472.124 (Official Build) (64-bit)

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9095

Updates the UI Shell with Actions and Right Panel to properly use state to manage the expansion so that it shows the proper transitions. 

#### Changelog

**Changed**

- Updates the story to show proper animations

#### Testing / Reviewing

Go to the `Header Base w/ Actions and Right Panel with Animation` story and click the notification icon. Ensure the transition works as expected


sstrubberg (approved):
Jul-30-2021 11-01-42

🚀

laurenmrice (commented):
- Just wondering if the open panel should also be connected to the header icon it was triggered from.

Ex:
Screen Shot 2021-08-02 at 1 24 32 PM

-----
- This tooltip gets cutoff and should probably have the caret tip positioned on the right side, so the tooltip opens to the left.

Screen Shot 2021-08-02 at 1 28 11 PM

tw15egan:
@laurenmrice fixed, nice catch 👍🏻

jnm2377 (approved):

laurenmrice (changes requested):
- [x] The tooltip position and icon to panel connection look good!

- One last thing I noticed is that when you click the icon in the header to open the panel, it shifts down a little bit in the header. So then it looks like its jumping around when you close or open the panel. This happens in Firefox/Chrome/Safari. My giph is taken from Safari as an example:
Aug-03-2021 10-41-22

tw15egan:
@laurenmrice should be fixed now, thanks for catching it!

laurenmrice (approved):
Looks great! 🎉

PR #9399, opened 2021-08-03
fix(ListBox): add title attribute to selected item

### LINKED ISSUE #9397: Hovering over the selected item from a dropdown doesn't show the complete text

## Description

`Dropdown`, `Combobox`

> see changed?

It is expected to see the entire text when we hover over a selected option from the dropdown. Like it shows the full text when we hover over any option when the dropdown is open.

Chrome 

For Dropdown:
1. Go to [Carbon Components React](https://react.carbondesignsystem.com/?path=/story/components-dropdown--default) portal, `Dropdown` page
2. Select the second option from the dropdown list - `Lorem, ipsum dolor sit amet consectetur adipisicing elit.`
3. Because of its length the string is truncated and it is expected to show the entire string when we hover over it. But it is not the case in the current implementation. 

For Combobox:
1. Go to [Carbon Components React](https://react.carbondesignsystem.com/?path=/story/components-combobox--combobox) portal, `Combobox` page
2. Select the second option from the dropdown list - `An example option that is really long to show what should be done to handle long text`
3. Because of its length the string is truncated and it is expected to show the entire string when we hover over it. But it is not the case in the current implementation.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9397

Adds title attribute so that long text can be revealed when it is the currently selected item

Screen Shot 2021-08-03 at 2 06 14 PM

#### Changelog

**New**

- Adds in `title` attribute to the selected item `span` / `input`

#### Testing / Reviewing

- [ ] ComboBox
- [ ] Dropdown 

Pick the text the longest text/text that is truncated. When the Dropdown / Combobox is closed, ensure the text is shown on hover

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.js
+++ b/packages/react/src/components/ComboBox/ComboBox.js
@@ -306,6 +306,7 @@ const ComboBox = React.forwardRef((props, ref) => {
                   aria-expanded={rootProps['aria-expanded']}
                   aria-haspopup="listbox"
                   aria-controls={inputProps['aria-controls']}
+                  title={textInput?.current?.value}
                   {...inputProps}
                   {...rest}
                   ref={mergeRefs(textInput, ref)}
--- a/packages/react/src/components/Dropdown/Dropdown.js
+++ b/packages/react/src/components/Dropdown/Dropdown.js
@@ -132,7 +132,6 @@ const Dropdown = React.forwardRef(function Dropdown(
 
   const menuItemOptionRefs = useRef(items.map((_) => React.createRef()));
 
-  console.log();
   return (
     <div className={wrapperClasses} {...other}>
       {titleText && (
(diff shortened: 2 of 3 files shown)


dakahn:
Is it possible for non-mouse users to see the full list item label somehow or is it also truncated in the listbox?

tw15egan:
@dakahn doesn't seem like it, this just matches the list menu experience (which also doesn't have a way for a non-mouse user to view the truncated text). A longer-term solution should be created in a separate issue and would need to be implemented consistently throughout the components

andreancardona (approved):
🎉

dakahn:
@tw15egan i'll bang out a ticket

dakahn (approved):

andreancardona:
@tw15egan is this ready to be merged? :)

PR #9418, opened 2021-08-04
fix(DataTable): add position relative to prevent jumping

### LINKED ISSUE #9403: Clicking table checkbox scrolls everything out of viewport

- [x] `carbon-components`

## Detailed description

If a carbon table is contained in a scrollable `<div>`, then clicking the checkbox on chrome may scroll everything out of view.
Clicking a checkbox should not cause any scrolling.

Tested on Carbon 10.40.

I suspect there's a similar problem with vanilla checkboxes, which could be solved with a `position:relative` on the `bx--checkbox-wrapper`.

1. Open https://codesandbox.io/s/adoring-sound-cvryv?file=/index.html
2. Scroll down to the table.
3. Click the checkbox.
4. Everything scrolls out of the viewport.

## Additional information

I originally filed this as https://bugs.chromium.org/p/chromium/issues/detail?id=1235323 but it seems to be a carbon issue, rather than a chrome issue per see.  The root issue is that the checkbox's `<input>` is outside of the viewport, causing a scroll when clicking the `<label>`.

It only happens when the scroll is on a `<div>` rather than on the document itself.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9403

Adds `position: relative` to prevent jumping when a checkbox is checked in a scrollable div. 

#### Changelog

**New**

- `position: relative` to data table

#### Testing / Reviewing

Check out https://codesandbox.io/s/adoring-sound-cvryv?file=/index.html and add `position: relative` to `.bx--data-table-container`. Scroll down and check a checkbox, the page shouldn't jump.

Ensure all data table variants remain unchanged.


dakahn (approved):
Works for me! 🏄🏾

joshblack (approved):

PR #9425, opened 2021-08-05
fix(button): mouse out

### LINKED ISSUE #8961: Tooltip doesn't disappear on mouseout when icon-only Button was previously disabled

- [x] `carbon-components`

## Detailed description

I have several icon-only `Button`s (rendered by a functional component) that, when clicked, execute a synchronous action (e.g., making a network request—simulated in the minimal test case below with a `setTimeout()`). While the action is in progress the button is disabled via a local boolean state variable. 

Before clicking the button, hovering over it displays its tooltip, which disappears as expected when the mouse pointer moves off the button. However, after clicking the button, waiting for the "synchronous event" to complete and the button to become enabled, the tooltip stops disappearing when moving the mouse pointer off the button. Hovering over another `hasIconOnly` button causes the "stuck" tooltip to disappear, but hovering over the originally clicked button causes its tooltip to become "stuck" again.

`<Button>` with `hasIconOnly` property and `disabled` property that is controlled by a boolean state variable.

> see changed?
The tooltip should disappear when the button is no longer hovered.

Microsoft Edge, Version 91.0.864.48 (Official build) (64-bit) (Windows 10)
I have confirmed the issue also occurs in Chrome, but **does not** occur in Firefox (so this might be a Blink rendering engine bug/issue).

10.37.0 / 7.37.0

> should be aware of?
[IBM Call Home Connect Cloud](www.ibm.com/support/call-home-connect/cloud). No pressing dates.

1. Go to my minimal test case at https://j0h0z.csb.app/.
2. Mouse over the buttons. Confirm that the tooltip appears when the button is hovered, and disappears when the mouse moves off the button.
3. Click one of the buttons. The clicked button becomes disabled for some amount of time.
4. After the button becomes enabled again, mouse over the button to display the tooltip.
5. Now move the mouse off the button. The tooltip does not disappear. 
    - Note: in step 3, if you leave the mouse over the button until it becomes enabled and then move the mouse off the button, the tooltip _will_ disappear. But if you then hover that button again and then move off, the tooltip becomes "stuck".
6. Now mouse over one of the other buttons on the page. Note that when its tooltip appears the "stuck" tooltip disappears.
7. If you click another one of the buttons, it will _also_ gain the "stuck" tooltip behavior (that behavior will affect any button that was disabled when it was clicked, and then re-enabled).

https://j0h0z.csb.app/

## Additional information

- Screenshots or code
- Notes

sstrubberg:
Closes #8961 
Refs https://github.com/carbon-design-system/carbon/issues/7040
Refs https://github.com/carbon-design-system/carbon/issues/4560
Refs https://github.com/carbon-design-system/carbon/pull/8022

*** DO NOT MERGE** until the `MouseOut` logic is removed from the Button story.

Refactored `Button` as it pertains to how it handles tooltips.

**Changed**

- In order to isolate the issue, we ended up stripping Button down of most of it's logic to get a sense of what conditionals were happening when the functions `handleFocus`, `handleBlur`, `handleMouseEnter`, and `handleMouseLeave` ran. In doing so we fairly quickly realized we didn't need the `alllowTooltipVisibilty`, `isHovered`, and `isFocused` states; and replaced them with simply `isTooltipVisible`. This took care of most of the work, but then we found that there were instances where multiple tooltips would occupy the screen simultaneously in some conditions. We ended up gutting the old `closeTooltips` function and replaced it with the global store for tracking and closing tooltips.

#### Testing / Reviewing

Open storybook and examine Button, specifically `iconOnly` button and the `MouseOut` story (the story that needs to be removed before merging the PR). Ensure the tooltips on the button fire on/off correctly with mouse and keyboard control.


tw15egan:
It seems like tooltips on focus are no longer appearing



I'd take a look at https://github.com/carbon-design-system/carbon/pull/8022 and ensure all use-cases are still being handled, and we'll also need to port these same changes over to `TooltipIcon`

dakahn:
Added the `WIP` label since it seems like this one is still cookin 👍🏾

PR #9434, opened 2021-08-06
chore(project): add quietDeps for sass tests

joshblack:
Our test output is pretty noisy with warnings for `/` in sass tests. This adds the `quietDeps` flag to the sass test renderer so these no longer are emitted as part of the test. We will still see these warnings during development, however.

#### Changelog

**New**

**Changed**

- Add `quietDeps` to sass renderer

**Removed**

#### Testing / Reviewing

- Verify that in the tests that we don't see any sass deprecation notices in the output

### DIFF
--- a/packages/test-utils/src/renderer.js
+++ b/packages/test-utils/src/renderer.js
@@ -34,6 +34,7 @@ const SassRenderer = {
           },
         },
         includePaths: [cwd, ...nodeModules],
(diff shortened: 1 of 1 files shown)


tw15egan (approved):
Tests are much quieter 🙏

dakahn (approved):

PR #9435, opened 2021-08-06
chore(project): update storybook to 6.3.6

joshblack:
Update storybook to 6.3.6

#### Changelog

**New**

**Changed**

- Update storybook to 6.3.6

**Removed**

#### Testing / Reviewing

- Make sure netlify previews for storybook work as expected


tw15egan (approved):
Both storybooks rendering as expected 👍✅

joshblack:
Nice catch by percy, cc @tay1orjones any idea why it might not be showing `show more` anymore? 🤔

tw15egan:
Weird, they are showing in the previews and when running locally

dakahn (approved):

tay1orjones:
@joshblack wow, sorry this fell through my notifications. I believe @jnm2377 pushed up a fix for the `show more` not rendering - we just needed to wait/assert that the show more button was settled in the dom before taking the snapshot.

I'll update with latest from main and I think it should resolve

Issue #9465, opened 2021-08-11
Update components to place `className`  on outer node


joshblack:
We currently inconsistently place `className` for our components. In v11, we will have each component place the `className` prop on the **outermost** node. This issue is for updating components in a non-breaking way using the `useFeatureFlag` hook to place the `className` prop on the outermost node.

```jsx
function Demo() {
  const enabled = useFeatureFlag('enable-v11-release');

  // Figure out common pattern for placing `className` on outermost node
}
```

This issue is for filling out the list below of components that currently place `className` on the outermost node and for addressing this using the pattern described above.

A list of components that place `className` on the outermost node:

- [ ] TODO

tw15egan:
Components that do not place `className` on the outermost node:

- [ ]

PR #9491, opened 2021-08-17
feat(classNames): move classNames to outer wrapper

### LINKED ISSUE #5853: [carbon-react] Input className prop inconsistencies
- [x] `carbon-components`

## Summary
The elements which receive extra classnames one can provide to carbon components vary for the different inputs. This makes it really hard to apply some custom styling to individual input components consistently.

- For text inputs and textareas, the className gets added to the actual `input` (or `textarea`) element
- In number inputs and the select component, it is added to the outer-most element (`.bx--number` / `.bx--select`), where `.bx--form-requirement` is a child.
- For Dropdown, ComboBox and MultiSelect the className is added to a wrapper element adjacent to `.bx--form-requirement`.

Is it planned to make this behaviour consistent and therefore more predictable in v11?

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/9465
Refs https://github.com/carbon-design-system/carbon/pull/9502
Closes https://github.com/carbon-design-system/carbon/issues/5853

Uses the `feature-flags` package to move the `className` prop to the outer wrapper.

#### Changelog

**New**

- Moved the `className` prop to the outermost wrapper so that all elements inside are targetable via css

**Changed**

- If the `className` prop was being added inside an element, it now returns null when the V11 feature flag is enabled. Instead, it is placed on the outer wrapper

#### Testing / Reviewing

Test the following components, which each have a test story that will be removed before merging:

- [ ] Checkbox
- [ ] Combobox / Dropdown / Multiselect
- [ ] DatePicker
- [ ] NumberInput
- [ ] RadioButtonGroup 
- [ ] Slider 
- [ ] TimePicker

Ensure `className` is placed on the outer wrapper if wrapped in the `FeatureFlag` element. `className` should remain in the same spot otherwise

### DIFF
--- a/packages/react/src/components/Checkbox/Checkbox.js
+++ b/packages/react/src/components/Checkbox/Checkbox.js
@@ -133,7 +133,7 @@ Checkbox.propTypes = {
    */
   wrapperClassName: deprecate(
     PropTypes.string,
-    `\nThe prop \`wrapperClassName\` for Checkbox has been deprecated in favor of \`className\`. \`className\` is now placed on the outer wrapper.`
+    `\nThe prop \`wrapperClassName\` for Checkbox will be deprecated in V11 in favor of \`className\`. \`className\` will then be placed on the outer wrapper.`
   ),
 };
 
--- a/packages/react/src/components/Checkbox/Checkbox.js
+++ b/packages/react/src/components/Checkbox/Checkbox.js
@@ -30,24 +30,18 @@ const Checkbox = React.forwardRef(function Checkbox(
 ) {
   const enabled = useFeatureFlag('enable-v11-release');
 
-  const labelClasses = enabled
-    ? `${prefix}--checkbox-label`
-    : classNames(`${prefix}--checkbox-label`, className);
+  const labelClasses = classNames(`${prefix}--checkbox-label`, [
+    enabled ? null : className,
+  ]);
   const innerLabelClasses = classNames(`${prefix}--checkbox-label-text`, {
     [`${prefix}--visually-hidden`]: hideLabel,
   });
 
-  const wrapperClasses = enabled
-    ? classNames(
-        `${prefix}--form-item`,
-        `${prefix}--checkbox-wrapper`,
-        className
-      )
-    : classNames(
-        `${prefix}--form-item`,
-        `${prefix}--checkbox-wrapper`,
-        wrapperClassName
-      );
+  const wrapperClasses = classNames(
+    `${prefix}--form-item`,
+    `${prefix}--checkbox-wrapper`,
+    [enabled ? className : wrapperClassName]
+  );
 
   return (
     <div className={wrapperClasses}>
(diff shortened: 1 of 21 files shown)


tw15egan:
cc @andreancardona

jnm2377 (approved):
Tested the component stories and they're working as expected! 🎉 🎉 🔥

packages/react/src/components/Checkbox/Checkbox.js
```diff
@@ -127,7 +131,10 @@ Checkbox.propTypes = {
   /**
    * The CSS class name to be placed on the wrapping element
    */
-  wrapperClassName: PropTypes.string,
+  wrapperClassName: deprecate(
```
joshblack, line 134: If people followed the deprecation notice and used `className` would the logic above pick up that `wrapperClassName` is not being used and it uses `className` instead? Wasn't 100% sure so figured I'd ask 👀

tw15egan: Good catch! I can update the wording to be a bit more clear. How's this sound?

"The prop `wrapperClassName` for Checkbox will be deprecated in V11 in favor of `className`. `className` will then be placed on the outer wrapper."

joshblack: Ah so sorry, totally misread what was happening. For cases where we have both a `className` prop and a class name to target an inner node (like `wrapperClassName` in this case) do you think we should give people a way to target the inner node? Or is it better to constrain them to the `className` on the outermost node? I think that's what tripped me up, I thought we were adding in `className` as a replacement to `wrapperClassName` but I didn't realize that the component already had a `className` prop

tw15egan: I figured it would be best that if a user passes in `className`, it is always placed on the outer node across the repo. That would still give them access to the inner nodes via the class they pass in. It would also reduce these `wrapperClassName` props that were added later on because we were placing `className` on the inner nodes and people had no way to target the outer nodes.

tw15egan: That way the experience is consistent across the repo. You pass in `className`, it always goes on the outer node and you have access to all internal nodes

joshblack: Sounds good! So in this case, would the CSS they would write be: `.my-custom-class .bx--form-item` to target what `wrapperClasses` sets or are you thinking it'd be something else?

tw15egan: I think it'd be `.my-custom-class.bx--form-item` since they're both placed on the same elements but yeah that's what I was thinking 👍🏻

tw15egan:
bump @joshblack when you have a chance 👍🏻

joshblack (approved):
So sorry about the delay! Looks great 👀

PR #9493, opened 2021-08-17
feat(Text): add text to Accordion, Button, Checkbox, Combobox, Contentswitcher

### LINKED ISSUE #9462: Update 5 components to use `Text`
Over in: https://github.com/carbon-design-system/carbon/pull/9013 we are adding in support for an experimental `Text` component. This component allows for automatic inferring of text direction through `dir="auto"`.

This issue is for implementing this `Text` in five components:

- [ ] Accordion
- [ ] Button
- [ ] Checkbox
- [ ] Combobox
- [ ] ContentSwitcher

dakahn:
Closes #9462


tay1orjones (approved):
🎉

packages/react/src/components/Accordion/AccordionItem.js
```diff
@@ -84,11 +85,13 @@ function AccordionItem({
         onKeyDown={onKeyDown}
         type="button">
         <ChevronRight16 className={`${prefix}--accordion__arrow`} />
-        <div className={`${prefix}--accordion__title`}>{title}</div>
+        <Text as="div" className={`${prefix}--accordion__title`}>
+          {title}
+        </Text>
       </Expand>
-      <div id={id} className={`${prefix}--accordion__content`}>
+      <Text as="div" id={id} className={`${prefix}--accordion__content`}>
```
joshblack: I think we'll have to keep this as a `div` since the content could contain things other than text nodes, potentially

packages/react/src/components/Button/Button.js
```diff
@@ -214,6 +215,9 @@ const Button = React.forwardRef(function Button(
     component = 'a';
     otherProps = anchorProps;
   }
+
+  const childrenWithText = <Text as="span">{children}</Text>;
```
joshblack: Ah, does Button potentially accept more than text children? This may be something where we would recommend that the consumer who wants the text translated either sets the direction themself or uses the `Text` component

joshblack (changes requested):
Looking great! Just a couple of usage notes/questions

joshblack:
I think a big thing for the conversion will be any text that we accept as props that the user cannot control rendering of will need to be wrapped in a `Text` component. If they can supply children or configure the nod already, then we don't need to convert that part of a component. Curious what you all think after going through these five components 🤔

dakahn:
@joshblack that makes sense to me!

packages/react/src/components/Accordion/AccordionItem.js
joshblack: I think specifically that we should keep this as before instead of changing it to `Text`. Sorry that was unclear!

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -115,7 +116,7 @@ export default class ContentSwitcher extends React.Component {
               ...data,
               index: this.state.selectedIndex,
               name: child.props.name,
-              text: child.props.text,
+              text: <Text>child.props.text</Text>,
```
joshblack: I think this would be a breaking change since we're changing the type of `this.props.onChange`. I don't know if we need to wrap this in `Text` either since this isn't a user-facing snippet of text

joshblack (changes requested):

joshblack (changes requested):
Not sure if you're still looking for a review from me, but I think the change left is in `AccordionItem` to wrap the title in a `Text` and to not wrap the content in a `Text`

joshblack:
@dakahn saw the re-review request, I think my last review comment is here: https://github.com/carbon-design-system/carbon/pull/9493#pullrequestreview-746369685

tay1orjones:
@joshblack fixed the accordion item 👍

joshblack (approved):

wajnberg:
Ok I checked all the 5 components:
This is what I found:
- Button is not working (see above)
- Accordion is OK
- Checkbox is OK
- Combobox is partially OK. It is working for title and helper text. But what about the options themselves ?
image
- Contentsswitcher is not working
image

I further looked at the code you committed and I can actually see that no code has been changed for Button (apart from a cosmetic change ) and ContentSwitcher.

Thank you for your feedback

Issue #8826, opened 2021-06-01
Dialog Component


joshblack:
## About

`<Dialog>` is a primitive used to create modal and non-modal dialogs.

## Accessibility

- When dialog is activated focus is moved to the dialog.
- Default initial focus is on the first interactive element in the dialog.
- The dialog should have an accessible name and announce itself as a "dialog".
- The dialog should be dismissable by 
  - activating a "close" button inside the dialog 
  - pressing the <kbd>escape</kbd> key
  - blurring the dialog by clicking or tapping outside of it
- If the dialog is modal obscured contents should be inert and inaccessible via keyboard and virtual cursor by constraining focus to the dialog in a "ring".
- If the dialog is non-modal keyboard and virtual cursor focus is not constrained to the dialog and can leave the dialog causing it to close.
- When the dialog is closed focus should return to the triggering element unless the dialog is non-modal and was closed by focus or virtual cursor leaving the dialog.

**Links & Resources**

- https://w3c.github.io/aria-practices/#dialog_modal
- https://www.tpgi.com/the-current-state-of-modal-dialog-accessibility/
- https://www.scottohara.me/blog/2019/03/05/open-dialog.html
- https://www.youtube.com/watch?v=JS68faEUduk
- https://svinkle.github.io/launchy/
- https://github.com/WICG/inert
- https://github.com/scottaohara/accessible_modal_window

## API Design

```jsx
function Demo() {
  const [open, setOpen] = useState(false);
  return (
    <Button onClick={() => setOpen(true)}>Open Dialog</Button>
    <Dialog open={open} onDismiss={() => setOpen(false)}>
      {children}
    </Dialog>
  );
}
```

**Props**

| Name | Type | Description |
| --- | --- | --- |
| `children` | Node | |
| `modal` | boolean | Specify if the dialog is a modal or non-modal dialog |
| `onDismiss` | Function | |
| `open` | boolean | |
| `initialFocusRef` | Object | A ref for the node that is focused when the dialog is opened |

**Links & Resources**

- https://reach.tech/dialog
- https://react-spectrum.adobe.com/react-spectrum/Dialog.html
- https://chakra-ui.com/docs/overlay/modal
- https://www.radix-ui.com/docs/primitives/components/dialog

## Requirements

- Both modal and nonmodal dialogs are supported
- TODO

## Open issues

- https://github.com/carbon-design-system/carbon/issues/2739

## Explorations

[...]

PR #9496, opened 2021-08-17
feat(react): add dialog component

joshblack:
This is the first part of #8826 where we would like to add in an official `<Dialog>` component to help with building both modal and nonmodal dialog.

This exploration has a storybook story that you can view locally but is currently not enabled in the deployed storybook.

#### Changelog

**New**

- Add unstable Dialog component exploration

**Changed**

**Removed**

#### Testing / Reviewing

- Pull down the PR locally
- Run the storybook in `packages/react`
- comment out the line in `Dialog-story` that says `includeStories: []`
- Try out the "Dialog Example" story for keyboard interaction and focus management


sstrubberg (approved):
Looks great! I had a question about the intended outcome on the attached mp4. When I open multiple dialogs, but close the original, then re-open it, I would expect it to open to where I was last instead of starting from the beginning. Thoughts?

https://user-images.githubusercontent.com/15822070/131915180-86e7d29f-adc3-4698-bb64-6ba04126e709.mp4

tw15egan (approved):
Focus trapping working as expected, focus returning to open button after closing and able to continue tabbing, and not seeing any accessibility violations. Nice work! 👍🏻 ✅

joshblack:
@sstrubberg I think it's expected since it's closing all the nested dialogs, effectively. The state isn't being kept and would most likely need to be something we add in on top of this implementation 👀

PR #9499, opened 2021-08-18
chore: adopt issue form templates

janhassel:
This is a quick exploration of how switching to GitHub's new issue form templates could look like.
So far, I only converted the bug report template as a discussion starter. Interested in hearing your thoughts!

#### Changelog

**Changed**

- Converted bug repot issue template to isse form


janhassel:
Closing in favor of #9579

PR #9502, opened 2021-08-18
feat(textInput, textArea, Select): move classNames to outer wrapper

### LINKED ISSUE #9465: Update components to place `className`  on outer node

### LINKED ISSUE #9469: [Tooltip] remains open when focus moves to other elements on the page

- [ ] `carbon-components`

## Detailed description

If you have multiple tooltips on a page and you tab through them all tooltips will stay open after they've lost focus. Often, they overlap blocking one another.

Tooltip

> see changed?

I would expect the tooltip to close when it no longer has focus.

Chrome 92.0.4515.131 

10.41.0/7.41.0

> should be aware of?

[Minimal repo sandbox](https://codesandbox.io/s/agitated-northcutt-ypjhj)

## Additional information

123816423-4d2ec600-d8c5-11eb-969c-82ef563a2b19

- Notes

Also tracking in the iot repo here: https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2437

andreancardona:
Closes #9465 

Uses the `feature-flags` package to move the `className` prop to the outer wrapper.

#### Changelog

**New**

- Moved the `className` prop to the outermost wrapper so that all elements inside are targetable via css

**Changed**

- If the `className` prop was being added inside an element, it now returns null when the V11 feature flag is enabled. Instead, it is placed on the outer wrapper

#### Testing / Reviewing

Test the following components, which each have a test story: `Class Name Change Test` that will be removed before merging:

- [ ] Select
- [ ] TextArea
- [ ] TextInput

Ensure `className` is placed on the outer wrapper if wrapped in the `FeatureFlag` element. `className` should remain in the same spot otherwise


packages/react/src/components/Select/Select.js
```diff
@@ -110,8 +115,14 @@ const Select = React.forwardRef(function Select(
...
   return (
-    <div className={`${prefix}--form-item`}>
+    <div
+      className={
+        enabled
+          ? classNames(`${prefix}--form-item`, className)
+          : `${prefix}--form-item`
+      }>
       <div className={selectClasses}>
+        {console.log('hi', className)}
```
tw15egan:
```suggestion
```

packages/react/src/components/TextArea/TextArea-story.js
```diff
@@ -12,9 +12,10 @@ import { withKnobs, boolean, number, text } from '@storybook/addon-knobs';
 import TextArea from '../TextArea';
 import TextAreaSkeleton from '../TextArea/TextArea.Skeleton';
 import mdx from './TextArea.mdx';
+import { FeatureFlags } from '../FeatureFlags';
 
 const TextAreaProps = () => ({
-  className: 'some-class',
+  // className: 'some-class',
```
tw15egan:
```suggestion
  className: 'some-class',
```

tw15egan:
Looks like `TextInput` is still adding the class to the inner input. I think we need to remove it from [here](https://github.com/carbon-design-system/carbon/pull/9502/files#diff-7820afc8fe28f023d88aad65b058606d670d1451929594acfce5505010358fb2R56)

Screen Shot 2021-08-18 at 9 43 55 PM

tw15egan (approved):
LGTM once test stories are removed

dakahn (approved):

PR #9508, opened 2021-08-19
Use math.div instead of slash for sass

### LINKED ISSUE #8746: Update usage of `/` operator to use `math.div`

egriff38:
Closes #8746

This PR attempts to replace all instances of sass division by slash with `math.div`. First time contributor 😄 

#### Changelog

**Changed**

- changes `*.scss` file slash division `$var1 / $var2` over to `math.div($var1, $var2)`
- Added `@use 'sass:math'` where necessary, below imports.
- Changes one instance from dividing by 4 to multiplying by 0.25

#### Testing / Reviewing
- ✔️  `yarn build` passed
- ✔️  `yarn lint` passed
- 👎  `yarn test` fails, but only for the one line packages/layout/scss/_convert.scss#24. I suspect it may be an issue with node-sass?


egriff38:

joshblack:
Hi there @egriff38! 👋  Thanks so much for the first-time contribution! 🙏  We're so excited that you were able to take the time to make this PR.

I think currently @tay1orjones was taking a look at in in our sprint. I think our specific challenge is how to support both `node-sass` and `sass` users with these updates since, unfortunately, `math.div` is not available in `node-sass`.

One option for this could be that anytime we want to update a file to use `math.div` we could use the `filename.import.scss` trick, I'm curious if you think that could work with the changes that you made or if you think this already satisfies our `node-sass` requirements 👀

egriff38:
Hi @joshblack, thanks for the welcome! I'm not familiar with the `filename.import.scss` trick. I do want to ask though why carbon continue to support node-sass when it was [announced deprecated](https://sass-lang.com/blog/libsass-is-deprecated) about 10 months back, and users opt in to their own sass compilers if they import the `.scss` files directly? Is there a significant portion of the user base still using `node-sass` over `sass` for compilation?

Besides the fix you mentioned, I'm unaware of any ways to make my change backwards compatible besides just leaving the files as they are and running the [division migrator](https://sass-lang.com/documentation/cli/migrator#division) to maintain 2 separate versions (Though that would mean double the stylesheets to test...). I used the migrator for part of my PR and it worked much better than my find and replace to generate the changes.

joshblack:
@egriff38 I think that's very fair to ask and will try my best to answer, let me know if there is anything else that I can help to clarify!

> Is there a significant portion of the user base still using node-sass over sass for compilation?

Unfortunately yes, the majority of teams using Carbon are still using `node-sass` at the moment. We're trying to start transitioning people over to Dart Sass before v11 where we will make the switch officially over to Dart Sass.

This is where that `*.import.scss` idea came to mind to support `@import` through dart sass with the new `math.div` while still supporting `node-sass` which would resolve to `*.scss`. Reference: https://sass-lang.com/documentation/at-rules/import#import-only-files

egriff38:
I think I understand what you mean. Do you think there could be some kind of polyfill module/import specifically for division that resolves to `a / b` for node-sass and `math.div` for dart-sass? Having a single utility polyfill file that gets imported everywhere could allow easier transition down the line to just `@use sass:math`. I need to better understand how the import-only files work though to know how to do that. We need to be sure the implementation is mutually exclusive across compilers so node-sass doesn't get a syntax error and dart-sass doesn't give deprecation warnings.

joshblack (approved):

andreancardona (approved):
@egriff38 @tay1orjones 🎉

metonym:
For anyone running into the `math.div is undefined` error after upgrading, you must use `sass` version 1.33.0 or greater:

> (node:73511) UnhandledPromiseRejectionWarning: Error: Undefined function.
> return math.div($px, $carbon--base-font-size) * 1rem;

[sass version 1.33.0 changelog](https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1330)

PR #9510, opened 2021-08-20
refactor(react): remove default placeholder from Combobox, TextArea, TextInput, MultiSelect, FilterableMultiselect

### LINKED ISSUE #5022: Removing placeholder text from Carbon Design System
### Summary

According to our guidance placeholder text should be avoided and definitely not used to convey critical information (input format etc) to the user. That said, there has been some discussion recently about what it might take (from a dev and design standpoint) to take placeholders out (?).

### Justification
Lots of research around placeholders and user expectations. Here's some I had on hand. Feel free to edit and add other links:

https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Placeholder_Research

### Questions

There are many, but one that came up in conversation was -- what would the blast radius of removing this feature be? Potentially any of our text inputs can have hidden labeling (and be using Placeholder as an ad hoc label 🤷‍♂

dakahn:
partially closes #5022

### Changes

- Removes default placeholder for 
1. Combobox
2. TextArea
3. TextInput
4. MultiSelect
5. FilterableMultiselect

### Notes
Three inputs with a  default placeholder were excluded from this PR (this PR now only partially closes 5022) because the removing default placeholder text without further design consideration would have resulted in worse usability potentially. These components are:
- Search
- Datepicker
- Timepicker

Any other instances of Placeholder in the codebase were not required or used in example storybook stories as far as i could tell 👍🏾


sstrubberg (approved):
LGTM!

jnm2377 (approved):

tay1orjones (approved):

packages/react/src/components/TextInput/TextInput.js
```diff
@@ -72,7 +72,7 @@ const TextInput = React.forwardRef(function TextInput(
     type,
     ref,
     className: textInputClasses,
-    title: placeholder,
```
joshblack: I _think_ this is considered a breaking change since existing code is now incorrect so might be worth flagging this

joshblack (changes requested):
Looking great! I think in this PR we saw: ComboBox, TextArea, and TextInput covered. Was curious if we were also going to cover:

- TimePicker
- ControlledPasswordInput
- PasswordInput
- Search
- [x] FilterableMultiSelect
- DatePickerInput
- TableToolbarSearch

Which I think also have a `placeholder` prop: https://carbon-react-explorer.vercel.app/props/placeholder

I think there are several with `placeholderText` too that I wasn't sure if they applied either but wanted to confirm: https://carbon-react-explorer.vercel.app/props/placeHolderText

dakahn:
@joshblack 

It doesn't look like we have helper text functionality for:
- tableToolbarSearch
- search
- datepicker

So without some design we'd probably need to punt removal of placeholder on those components since we don't really have a codified way of showing -- date format in DatePIcker apart from placeholder text for example.

That said it looks like:
- timepicker
- filterableMultiselect
- passwordInput
- controlledPasswordInput
Have placeholder functionality and so i'll circle back through and change these 👍🏾

tay1orjones (approved):

joshblack (approved):

aagonzales (approved):
LGTM!

PR #9526, opened 2021-08-24
fix(FormGroup): only place className on outer wrapper

### LINKED ISSUE #4572: [FormGroup]: Custom className prop gets applied to both the `label` and the `fieldset` elements
Screen Shot 2019-11-05 at 1 13 14 PM
---
Screen Shot 2019-11-05 at 1 15 30 PM
---
This makes it impossible to style these elements independently:
Screen Shot 2019-11-05 at 1 15 41 PM

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/4572

This ensures the `className` prop is only placed on the outer wrapper, and not duplicated. I've placed this behind a feature flag, but we may just want to go ahead and make this change as it would be difficult to style them independently as-is

#### Changelog

**New**

- Test story in `FormGroup`

**Changed**

- `className` prop is only placed on the outer wrapping `fieldset` element

#### Testing / Reviewing

Go to the test story and ensure the custom class is only placed on the outer wrapper

### DIFF
--- a/packages/react/src/components/FormGroup/FormGroup-story.js
+++ b/packages/react/src/components/FormGroup/FormGroup-story.js
@@ -13,6 +13,7 @@ import RadioButtonGroup from '../RadioButtonGroup';
 import RadioButton from '../RadioButton';
 import Button from '../Button';
 import mdx from './FormGroup.mdx';
+import { FeatureFlags } from '../FeatureFlags';
 
 const props = () => ({
   disabled: boolean('Disabled (disabled)', false),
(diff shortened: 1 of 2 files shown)


jnm2377 (approved):

tw15egan:
bump @joshblack when you have a chance 👍🏻

joshblack (approved):
Seems great! Quick question on the test stories, do we want those showing up in the production storybook or are you removing/disabling them before merge? 👀

joshblack:
So sorry about the delay @tw15egan! Totally missed this 🤦

tw15egan:
@joshblack will get rid of them before merging 🙂

PR #9528, opened 2021-08-24
fix(menu): remove unnecessary outside click handler

janhassel:
With React v17, the OverflowMenuNext (v11) component wasn't opening anymore supposedly due to smarter execution of the hooks which led to `canBeClosed` being `true` immediately when opened, thus allowing the `ClickListener` to fire its `onOutsideClick` event and closing the menu since the registered click was still on the trigger element.

A somewhat recent update to the Menu component added an `onBlur` event in order to correctly return focus for keyboard users. After some testing I noticed that it renders the `ClickListener` unnecessary as it more accurately detects when the menu should close in this context. Removing the `ClickListener` and shifting the `rootRef` on the `<ul>` element promises to reduce complexity and fixes the original issue of the menu not opening (more accurately: closing immediately) on React v17.

#### Changelog

**Changed**

- Moved `rootRef` to `<ul>` element

**Removed**

- Removed `ClickListener` wrapper on `Menu`
- Removed unused `autoclose` prop (not a breaking change since it's marked as experimental/unstable)

#### Testing / Reviewing

- Ensure both the Context menu as well as the unstable_Menu->OverflowMenu stories still work as expected (open correctly and close on outside click) with React v16
- Test the same with React v17
- Try to open the ContextMenu in **Safari** by using right-click and (more importantly) *ctrl+click**. The menu should stay open in both cases until an outside click is registered (or a menu item is pressed).


tay1orjones (approved):
Things look good with `react@16.x `

I did a `yarn upgrade-interactive` and pulled `packages/react` react/react-dom deps up to `v17.0.2` and re-ran storybook. Things look good there too in both safari and chrome. 👍 

One thing I did notice in the overflow menu story -- when it opens there is a brief flash of hover/red on the danger item at the bottom. Here's a video, it might help to download it and play at half speed to see it. It's very fast.

https://user-images.githubusercontent.com/3360588/130846432-e758752f-70c2-4dba-ba64-54d39519e29c.mov

andreancardona (approved):
looks good! :)

PR #9533, opened 2021-08-24
chore(carbon-react): added dependencies for better install experience

### LINKED ISSUE #9529: Add all peer dependencies to `@carbon/react` for easier install
We currently get some warnings of uninstalled `peerDependencies` when installing `@carbon/react`. This issue is to add those peer dependencies automatically to `@carbon/react` so users no longer see this notice.

sstrubberg:
Closes #9529 

Added  `carbon-components` and `carbon-icons` to the Dependencies for a better `@carbon/react` installation experience. Users should no longer receive the following warnings on install.

image


jnm2377:
Please correct me if I'm wrong, but I thought peer dependencies still [require you to install the dependencies](https://samuelsson.dev/difference-between-npm-dependencies/)? But actual dependencies won't need to be installed? I think the better option would be to go with dependencies in that case so people don't actually have to install these. However, if peer dependencies don't require installation on their end, then it's cool 👍🏽

sstrubberg:
> Please correct me if I'm wrong, but I thought peer dependencies still [require you to install the dependencies](https://samuelsson.dev/difference-between-npm-dependencies/)? But actual dependencies won't need to be installed? I think the better option would be to go with dependencies in that case so people don't actually have to install these. However, if peer dependencies don't require installation on their end, then it's cool 👍🏽

thanks for the heads up!

jnm2377 (approved):

tw15egan (approved):

PR #9556, opened 2021-08-28
feat(tooltip): Add optional prop to auto orientate within parent bounds

### LINKED ISSUE #7321: Tooltip is escaping the screen when it is close right left bottom f the screen

- [x] @console/pal v1.302.1
- [x] carbon-components v10.23.2
- [x] carbon-components-react v7.23.2

## Detailed description

- Tooltip is escaping from the screen when label is close to top bottom left or right borders

 - Tooltip stays within the screen borders 
> What happened instead? 
 - Tooltip escapes the screen

>What would you like to see changed?
 - Tooltip should stay in the screen while the sharkfin adapts and stays on the icon to which it is attached

 - Google Chrome

> What offering/product do you work on?
 - Ibm cloud Power-iaas. This issue is causing many accessibility and responsivity problems and it is very blocking for our product 

 - create a label with Tooltip icon which is close to screen borders the tooltip will get out of at least one of the sides.

##  Reduced test case in CodeSandbox
 - no need to create test case its visible on https://www.carbondesignsystem.com/components/tooltip/usage/ 
 - direct sandbox link :  https://codesandbox.io/s/sikg6

## Additional information
image

TuxedoFish:
Closes #7321 

- Tooltip is escaping from the screen when label is close to top bottom left or right borders
- Adds `autoOrientation` prop which allows the `align` and `direction` props to be updated

#### Changelog

**New**

- Adds in an `AutoOrientation` story to the storybook for the Tooltip
- Adds in `getBestOrientation` function to FloatingMenu
- Adds in `autoOrientation` prop to the Tooltip which uses this function

**Changed**

- direction and align are initially set via props but the rendered menu uses state

**Removed**

- N/A

#### Testing / Reviewing

- Turning off autoOrientation should have the existing behaviour of tooltips falling outside of the container bounds 
- Turning on autoOrientation should stop the tooltip from being able to fall outside of the bounds of its container


TuxedoFish:

Harry-Liversedge:

tw15egan:
This is really great! It's looking good when aligned to the left of the screen, but seeing some issues when I set `justifyContent: flex-end` and align to the right of the screen 

Screen Shot 2021-08-30 at 1 11 33 PM


Screen Shot 2021-08-30 at 1 11 10 PM

TuxedoFish:
> This is really great! It's looking good when aligned to the left of the screen, but seeing some issues when I set `justifyContent: flex-end` and align to the right of the screen
> 
> Screen Shot 2021-08-30 at 1 11 33 PM
> 
> Screen Shot 2021-08-30 at 1 11 10 PM

Thanks so much 😄 !

Ahh yes so I saw this but then I found I could recreate this on the public storybook for 7.42.0 (https://react.carbondesignsystem.com/?path=/story/components-tooltip--default-bottom)

Screenshot (5)
Screenshot (6)

All this PR should do in theory is if tooltip direction is bottom and that would move it outside of the container, then it switches it to top. But then in this case it looks like the edge case of positioning it up against the right edge seems to be not quite behaving as expected? 

I can have a look into why that is happenning but to me, it seems to be an existing issue not something added by this change, do you want me to work on fixing it in this PR @tw15egan?

tw15egan:
Right you are @TuxedoFish, I'm also seeing the behavior in the existing storybook. I'll create a separate issue, and if you'd like, you can submit a separate PR from this one. Since this is an existing bug, and your PR introduces new functionality, I'd like to keep them separate. Thanks for digging into the issue! 😄

packages/react/src/components/Tooltip/Tooltip.js
```diff
@@ -263,6 +272,159 @@ class Tooltip extends Component {
...
+    const scrollY = container.position !== 'static' ? 0 : pageYOffset;
+    const relativeDiff = {
+      top: container.position !== 'static' ? container.rect.top : 0,
+      left: container.position !== 'static' ? container.rect.left : 0,
+    };
+    const { width, height } = menuSize;
+    const { top = 0, left = 0 } = offset;
+    const refCenterHorizontal = (refLeft + refRight) / 2;
+    const refCenterVertical = (refTop + refBottom) / 2;
+    console.table(container);
```
tw15egan:
```suggestion
```

TuxedoFish: Oh damn, yes sorry about that, I'll go through and remove any console statements

tw15egan: No worries!

TuxedoFish:
> Right you are @TuxedoFish, I'm also seeing the behavior in the existing storybook. I'll create a separate issue, and if you'd like, you can submit a separate PR from this one. Since this is an existing bug, and your PR introduces new functionality, I'd like to keep them separate. Thanks for digging into the issue! 😄

Awesome, yeh I would be happy to work on that!

tw15egan (approved):
LGTM, thanks for contributing!

TuxedoFish:
> LGTM, thanks for contributing!

Thanks!

tay1orjones (commented):
@TuxedoFish Thanks for taking a look at this! 

Adding to the `flex-end` situation a bit - I wonder if the initial positioning logic is the cause? I can replicate the placement being incorrect, but as soon as the browser window is resized, it snaps into place:


https://user-images.githubusercontent.com/3360588/131559007-6021e4f8-758d-4e1e-892b-1f1e7a64997f.mov

I also see similar incorrect behavior with the default flex/justify properties on small viewport sizes with resize:

https://user-images.githubusercontent.com/3360588/131559080-8707aaed-02a5-4820-a60c-c0667c614a89.mov

tay1orjones:
I also think it could be helpful for the auto orientation story to render four tooltips/triggers instead of just one. If one was placed in each corner of the screen, it would be very easy to see how the auto orientation prop augments the positioning logic to avoid placing tooltips outside of the container.

TuxedoFish:
> @TuxedoFish Thanks for taking a look at this!
> 
> Adding to the `flex-end` situation a bit - I wonder if the initial positioning logic is the cause? I can replicate the placement being incorrect, but as soon as the browser window is resized, it snaps into place:
> 
>  flex-end.with.resize.mov 
> I also see similar incorrect behavior with the default flex/justify properties on small viewport sizes with resize:
> 
>  small.sizes.and.resize.mov

Yes I see that too, I also noticed that if I output the bounding rectangle that is used in Floating Menu before and after resizing the window, it is getting different widths and heights. 


https://user-images.githubusercontent.com/3999881/131625562-7ace6ffe-9583-4a2b-bc66-810df67e1b69.mp4


I saw that the bx--tooltip has a min-width and max-width so the width can be a certain range. It seems like when the tooltip is positioned for the first time on those edges it then seems to assume a larger size which then means the position looks off potentially? Not sure how to go about fixing this?

I have also pushed that change, sounds like a good idea, I have kept them all using the same props from storybook and then positioned them in the four corners!

tay1orjones:
@TuxedoFish if it's a problem of having outdated/invalid parameters of the positioning values, I wonder if moving the logic to `getSnapshotBeforeUpdate` instead of `componentDidUpdate` would be useful 🤔

TuxedoFish:
> @TuxedoFish if it's a problem of having outdated/invalid parameters of the positioning values, I wonder if moving the logic to `getSnapshotBeforeUpdate` instead of `componentDidUpdate` would be useful 🤔

I gave it a try and it didn't solve it straight away but I'll have to investigate further! @tay1orjones are you happy for me to work on this in a separate PR for #9561.

If so is there any changes you want me to make to this PR or are you happy to approve? Thanks,

TuxedoFish:
I have opened #9606 as a potential fix for the alignment of tooltips on the right hand side (#9561)

tay1orjones (approved):
Sorry for the delay @TuxedoFish - this looks great, thanks for the contribution!

PR #9559, opened 2021-08-30
feat(Multiselect): add itemToElement to Multiselect

### LINKED ISSUE #9557: Support rendering items as node for MultiSelect
### Summary

Currently, carbon only supports `itemToString` for displaying items in `MultiSelect` component. We would like to have an additional prop `itemToElement` like DropDown so we can show icons, or other components in the items.

### Justification

It's common to not only show string in an item. Users may have needs to show more than text in the items, for example, icons, tooltips.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9557

Adds in `itemToElement` support to `Multiselect` so that it functions similarly to `Dropdown` and `Combobox`

#### Changelog

**New**

- `itemToElement` prop added to `Multiselect` and `FilterableMultiselect`

**Changed**

- Added a story for easier testing

#### Testing / Reviewing

Go to the `Item To Element` story under `MultiSelect` and ensure the component renders as expected and is still searchable.


packages/react/src/components/MultiSelect/MultiSelect.js
```diff
@@ -447,6 +461,7 @@ MultiSelect.defaultProps = {
   compareItems: defaultCompareItems,
   disabled: false,
   locale: 'en',
+  itemToElement: null,
```
joshblack:
```suggestion
```

I think we could leave it undefined with the checks above before calling, what do you think?

joshblack (approved):

packages/react/src/components/MultiSelect/MultiSelect.js
tw15egan: Definitely 👍🏻

andreancardona (approved):
pending Josh's question, looks good :)

PR #9571, opened 2021-08-30
fix(tile): added deprecation warnings for v11

sstrubberg:
Refactored most of Tile to be a Functional component and added deprecation warnings for [handleClick](https://carbon-react-explorer.vercel.app/props/handleClick) and [handleKeyDown](https://carbon-react-explorer.vercel.app/props/handleKeyDown). Added deprecation warnings to Button as well.

**Halp**

Tests are failing all over the place. Could use another set of eyes.


packages/react/src/components/Tile/Tile.js
```diff
@@ -21,190 +21,176 @@ import { composeEventHandlers } from '../../tools/events';
 
 const { prefix } = settings;
 
-export class Tile extends Component {
-  static propTypes = {
-    /**
-     * The child nodes.
-     */
-    children: PropTypes.node,
+export const Tile = ({ children, className, light = false, ...other }) => {
```
joshblack: Small style nits (almost all for codebase consistency, not correctness):

- Our style guide uses function declarations for components: https://github.com/carbon-design-system/carbon/blob/main/docs/style.md#writing-a-component
- `other` -> `rest`
- `classNames` -> `cx`

joshblack (commented):
Looking great! Just left a style nit comment, the only other thing that came to mind was if we needed to use `React.forwardRef` here.

Sometimes, our class components are used in a way where people are passing in a `ref` to them and they consider it a breaking change if `ref` no longer works for a component. I'm not sure what the best way forward is with this and would love to hear your thoughts on if we should go for the functional components or if `React.forwardRef` may be an okay compromise 👀

PR #9591, opened 2021-09-02
feat(react): add display names to forward ref components

### LINKED ISSUE #9459: Some components do not have a 'displayName' set

aledavila:
Closes #9459 

Added `displayName` to components that are using forwardRef

#### Changelog

**New**

Added to the following components:
- Popover
- ComboBox
- ListBoxMenuItem
- PaginationNav
- Toolbar
- TextInput (PasswordInput)
- HeaderMenuItem
- SideNav
- SideNavLink
- SideNavMenuItem
- SideNavSwitcher
- Switcher

#### Testing / Reviewing

Make sure components work as intended. Make sure display names are correct


tay1orjones (approved):

joshblack (approved):

joshblack:
@aledavila I think updating snapshots is the last thing and then we're good to go!

joshblack:
@aledavila seems like there are some test failures coming up:

image

PR #9593, opened 2021-09-03
fix(Tile): recalculate tile rect height on component did mount

### LINKED ISSUE #9187: Issue with using ExpandableTile within a Tab

- [ ] `carbon-components`

## Detailed description

>Describe in detail the issue you're having.

The ExpandableTile component's default "unexpanded" state does not have the correct height when the page first renders. The below screenshot shows how they're way too short.
image

And the tile boxed in red shows what it's actually supposed to look like, with some elements AboveTheFold. You can "fix" this bug by expanding and collapsing the tile, but the user shouldn't have to do that to achieve the desired appearance.
image

-----------

After some poking around, it may have something to do with the Tabs and how the selected tab function works. By default, the first Tab ("Active") is selected, and the second "Upcoming" tab is not by default selected (and is where the ExpandableTiles are located). 

This is how the page loads right now, with the "Active" Tab being the default selected Tab. You can see how the ExpandableTiles under the "Upcoming" Tab have the incorrect collapsed height, which can be fixed by expanding/collapsing it.
9d599280-db28-11eb-9e60-88e2a798e003

However, when the default selected Tab is "Upcoming," then the bug does not occur.
6d11f400-db28-11eb-88e1-fb20ffab73d4

It might have something to do with [this](https://github.com/carbon-design-system/carbon/blob/e7ef150ba1d74e2e864a9d6d67e612e824dcbf51/packages/react/src/components/Tile/Tile.js#L516) line that calls `getBoundingClientRect()` because when I tried calling that function from the Inspector it always returned zeros for every dimension.

ExpandableTile and Tabs

Firefox

> should be aware of?
This is for the Commitments and Subscriptions page for the billing microservice.

1. Go to https://test.cloud.ibm.com/billing/subscriptions
2. Login with username `orlando.mendoza1+sld_08@mail.test.ibm.com` and password `ServiceLD01!`
3. Click on the "Upcoming" Tab

## Additional information

[Link to another issue within IBM that documents this bug](https://github.ibm.com/ibmcloud/TOB/issues/124)

dakahn:
Closes #9187 

Adds a resize observer to Tile that recalculates the rect height to account for tiles initially rendered as hidden (like in a set of tabs)


sstrubberg (changes requested):
Getting a CORS error in deploy preview.

image

This reminds me *exactly* of a [PR](https://github.com/carbon-design-system/carbon/pull/9504) i just recently spent way too much time debugging. I think it's Storybook acting up. I ended up removing the Storybook Actions Add-on for Button so it would work.

Also, looks like tests are failing.

packages/react/src/components/Tile/Tile.js
```diff
@@ -654,7 +672,9 @@ export class TileAboveTheFoldContent extends Component {
     const { children } = this.props;
 
     return (
-      <span className={`${prefix}--tile-content__above-the-fold`}>
+      <span
+        id="abovethefold"
```
joshblack:
```suggestion
```

joshblack (changes requested):
Just a small note with the `id` change.

A question that I don't know if we considered, why do we need to calculate the height of the "above the fold" content in the first place? Can't we just _only_ render the above the fold content until the tile is expanded and then render the rest of the content? 🤔

packages/react/src/components/Tile/Tile.js
```diff
@@ -507,11 +507,23 @@ export class ExpandableTile extends Component {
         };
   }
 
+  resizeObserver = null;
+  resizeElement = React.createRef();
```
jnm2377: Is `resizeElement` needed? I don't see it being used anywhere after this, so I'm a bit confused about it.

dakahn: Nope! i'll take this out. artifact from some trouble shooting 👍🏾

dakahn:
> Just a small note with the `id` change.
> 
> A question that I don't know if we considered, why do we need to calculate the height of the "above the fold" content in the first place? Can't we just _only_ render the above the fold content until the tile is expanded and then render the rest of the content?

That makes sense to me @joshblack -- should I open a ticket for a minor tile refactor?

joshblack:
@dakahn sounds great to me!

packages/react/src/components/Tile/Tile-test.js
```diff
@@ -19,6 +19,14 @@ import { settings } from 'carbon-components';
...
 const { prefix } = settings;
 
+const observe = jest.fn();
+const unobserve = jest.fn();
+
+window.ResizeObserver = jest.fn(() => ({
+  observe,
+  unobserve,
+}));
+
```
joshblack: Make sure to do this in a `beforeEach` block and then restore it in `afterEach`, otherwise it may interfere with other tests in the future that expect this to be the native `window.ResizeObserver`.

For example:

```js
let nativeResizeObserver = null;

beforeEach(() => {
  nativeResizeObserver = window.ResizeObserver;
  window.ResizeObserver = jest.fn(() => {
    return {
      observe: jest.fn(),
      unobserve: jest.fn(),
    };
  });
});

afterEach(() => {
  window.ResizeObserver = nativeResiveObserver;
});
```

Another option could be using `jest.spyOn`

```js
let mockObserve;
let mockUnobserve;

beforeEach(() => {
  mockObserve = jest.spyOn(window.ResizeObserver, 'observe').mockImplementation(() => {});
  mockUnobserve = jest.spyOn(window.ResizeObserver, 'unobserve').mockImplementation(() => {});
});

afterEach(() => {
  mockObserve.mockRestore();
  mockUnobserve.mockRestore();
});
```

joshblack (changes requested):

dakahn:
@joshblack wrapped the resizeObserver work in before/after each blocks and scoped it to the expandable tile tests, but it doesn't seem to be firing off properly

joshblack:
@dakahn I think it'll need to be `global` instead of `window`, potentially, since it's saying `ResizeObserver` is undefined and so it's not resolving it from the global scope

dakahn:
Hmmm -- @joshblack something funnier is going on I don't understand. If I lift up `window.ResizeOBserver = ....` globally it works (like I would expect), but if I wrap it in `beforeEach` it breaks, but maybe `beforeEach` has to be _inside_ a `describe` block?

joshblack:
@dakahn yeah, I'd make sure that it's inside of the `describe` block for sure 👍  (let me know if I'm misunderstanding!)

packages/react/src/components/Tile/Tile-test.js
```diff
@@ -236,6 +237,19 @@ describe('Tile', () => {
   });
 
   describe('Renders expandable tile as expected', () => {
+    window.ResizeObserver = jest.fn(() => {
+      return {
+        observe: jest.fn(),
+        unobserve: jest.fn(),
+      };
+    });
+
```
dakahn: @joshblack for some reason (something about the async nature of window.ResizeObserver?) wrapping this assignment in a `beforeEach` block causes ResizeObserver to be undefined and error. If I'm not mistaken though the `afterAll` reset below should be correctly resetting the observer back to window default, right?

joshblack:
@dakahn just pushed up a fix that should cover it

joshblack (approved):

jnm2377 (approved):

sstrubberg (changes requested):
Still getting CORS errors on deploy preview.

sstrubberg (approved):

sstrubberg:
Looks like the CORS errors are only present on Netlify's deploy preview. They work perfectly locally and in prod.

PR #9598, opened 2021-09-03
fix(progress-indicator): fix vertical small screen bug

### LINKED ISSUE #9261: Vertical Progress Indicator: missing spacing when flex wraps on narrow widths

- [ ] `carbon-components`

## Detailed description

when using the `ProgressIndicator` component with the `vertical` prop set to `true`, e.g. `<ProgressIndicator vertical />`, narrow container width causes an the SVG and text in each step to wrap. this is due to the `flex-wrap` CSS property. since the text of the step has no margin/padding, it sits directly against the vertical line. presumably, there should be some spacing. this definitely occurs in the newest Chrome Dev version but I haven't tested others yet. other browsers likely have the same issue.

while the below reproduction steps are somewhat contrived, this can happen whenever the container is too small. so when `ProgressIndicator` is in a row, this can occur at much larger screen sizes. for example, in [@carbon/ibm-security](https://ibm-security.carbondesignsystem.com/?path=/story/patterns-wizard--default), our `Wizard` will cause overflow overlapping. in a specific example of ISV, we're seeing the flow-wrap at 707px.

1. open [Carbon React storybook to `ProgressIndicator`](https://react.carbondesignsystem.com/?path=/story/components-progressindicator--default)
2. open bottom drawer, select `Knobs` tab, and check the `vertical` prop
3. resize the visible area to less than 314px (using dev tools and making them really big or the device screen size simulator)
4. observe the SVGs and text splitting to separate lines and the lack of spacing on the left seen in the screenshot

## Additional information

- Carbon Storybook vertical progress indicator narrow
- IBM Security `Wizard` 
IBM Security `Wizard`
- ISV `Tearsheet` 
ISV `Tearsheet`

andreancardona:
Closes #9261 

## Testing

- [ ] Checkout the progress indicator / change the knob to vertical / add a long title to one of the options
- [ ] In the inspector, change the screen to be less than 307px
- [ ] The lines should no longer push into the left border and should look like below

Screen Shot 2021-09-03 at 10 18 18 AM


jnm2377:
Hmm.. wondering if we should be finding a way to implement an ellipsis or somehow be more specific about text wrapping. The issue still occurs below 307 breakpoint, but also I think it could happen at any breakpoint if the vertical progress indicator is within a small container:
Screen Shot 2021-09-04 at 6 46 33 PM
Screen Shot 2021-09-04 at 6 43 35 PM

joshblack:
@jnm2377 that makes sense, if we could have the text wrap then it would really help us out here since we need to use tooltips as soon as ellipsis get introduced 😞

andreancardona:
@joshblack @jnm2377 Just pushed up a fix. let me know what you think!

jnm2377 (approved):

joshblack (approved):

PR #9606, opened 2021-09-05
fix(tooltip): Fix tooltip alignment aligned on right hand side of container

### LINKED ISSUE #9561: Tooltip issue when aligned to right of screen
Noticed while reviewing https://github.com/carbon-design-system/carbon/pull/9556

Screen Shot 2021-08-30 at 1 11 10 PM Screen Shot 2021-08-30 at 1 11 33 PM

Additional info provided by @TuxedoFish confirming the issue was an existing bug

https://github.com/carbon-design-system/carbon/pull/9556#issuecomment-908284591

Possibly due to the storybook being inside of a flex container causing positioning issues when right-aligned

TuxedoFish:
Closes #9561

- When tooltip is aligned up against RHS the tooltip appears out of position

#### Changelog

**New**

- N/A

**Changed**

- Add a callback to setState(floatingPosition:...) inside _updateMenuSize
- Will refire _updateMenuSize if the new bounding rectangle is different 
- This can only be fired once 

**Removed**

- N/A

#### Testing / Reviewing

- Align the tooltip up against the RHS and see that the tooltip is correctly positioned


jnm2377 (approved):
Sorry for the delay in reviewing. I compared to live tooltip story, and fix is working. 👍🏽 Thanks for the contribution! 💯

tay1orjones (approved):

PR #9626, opened 2021-09-09
docs(file-uploader): show onAddFiles action in storybook

janhassel:
Right now, the [FileUploaderDropContainer story](https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader-drop-container) is showing `onChange` as the only event in the actions panel.

`onChange` isn't in the prop types though but gives the impression to work because it is emitted on the `<label>` element and therefore is called when the underlying `<input type="file">` fires an `onChange`.

The problem is: this means that `onChange` is only called when the user clicks on the input and selects a file. If the user drags and drops a file on the container, `onChange` is not called.

This PR updates the action in the storybook to show the preferred (and listed in the prop types) `onAddFiles`. which reacts to both: manual selection by clicking and dropping a file.

#### Changelog

**Changed**

- FileUploaderDropContainr story: use `onAddFiles` instead of `onChange` as action

#### Testing / Reviewing

- Visit current react storybook, try to drop a file -> there shouldn't be any action shown in the story
- Run the react storybook with this PR and drop a file -> there should be an action shown: `onAddFiles`


sstrubberg:
Hey @janhassel, thanks for the PR. It works like a champ locally but in the deploy preview I'm getting a CORS error in the console when I try to drag n' drop the file. This is something I'm noticing with Storybook Actions throughout our entire Storybook. REF #9593

janhassel:
@sstrubberg I wonder if this just due to the netlify preview. If I drag and drop a file in the story on react.carbondesignsystem.com I don't see that error.

sstrubberg:
@janhassel That would be awesome if that were the case. I'm down to merge and then see what happens in prod. We can always submit a PR to remove the actions if it busts.

sstrubberg (approved):

tay1orjones (approved):

Issue #9535, opened 2021-08-24
v11 Prop Changes 


joshblack:
**Goal**

We want the props for the components in our React library to be consistent and predictable. Props should follow common or familiar conventions and should be consistent across components in the design system.

**Links & Resources**

- https://carbon-react-explorer.vercel.app/

**Related issues**

- https://github.com/carbon-design-system/carbon/issues/5853
- https://github.com/carbon-design-system/carbon/issues/5972
- https://github.com/carbon-design-system/carbon/issues/5367
- https://github.com/carbon-design-system/carbon/issues/5022
- https://github.com/carbon-design-system/carbon/issues/2343
- https://github.com/carbon-design-system/carbon/issues/2529
- https://github.com/carbon-design-system/carbon/issues/4356

**Checklist**

- [x] Change instances of `handle*` to `on*`
  - [x] [`handleClick`](https://carbon-react-explorer.vercel.app/props/handleClick)
  - [x] [`handleKeyDown`](https://carbon-react-explorer.vercel.app/props/handleKeyDown)
  - [x] [~`handleOpen`~](https://carbon-react-explorer.vercel.app/props/handleOpen) - Only exists in `ModalWrapper` - component no longer maintained.
  - [x] [~`handleOverflowMenuItemFocus`~](https://carbon-react-explorer.vercel.app/props/handleOverflowMenuItemFocus) - Only exists in `OverflowMenu` and is not consumable by the user.
  - [x] [~`handleSubmit`~](https://carbon-react-explorer.vercel.app/props/handleSubmit) - Only exists in `ModalWrapper` - component no longer maintained.
  - [x] [~`handleTabClick`~](https://carbon-react-explorer.vercel.app/props/handleTabClick) - Is a child of `Tabs` and is not consumable by the user.
  - [x] [~`handleTabKeyDown`~](https://carbon-react-explorer.vercel.app/props/handleTabKeyDown) - Is a child of `Tabs` and is not consumable by the user.
- [x] #10317
- [x] #9622
- [x] #9623
- [x] #9624
- [x] #9641
  - [ ] Find a list of components that share state in a event handler and change them
- [x] #9763
- [ ] Update any handlers that use `evt.persist()`
- [ ] titleText vs labelText, different for dropdown and number input
- [ ] placeholder vs placeholderText
  - [ ] Check-in with DA/Taylor because they may be removing this
[...]

PR #9628, opened 2021-09-09
chore(Tile): updated props for v11

sstrubberg:
REF #9535

#### Changelog

- Added deprecation warnings for handleClick and handleKeyDown to `Tile`.
- Updated the PublicAPI test snapshot.
- Touched up `Tile` by changing `{...other}` -> `{...rest}` and `classNames` -> `cx`.
- Updated the `Tile` story with new props.
- Added comments in the appropriate areas to remove deprecations for the next major release.

#### Testing / Reviewing

- Check the Tile stories and make sure everything is firing off as it should.
- Make sure tests are passing.


packages/react/src/components/Button/Button.js
```diff
@@ -262,6 +262,14 @@ Button.propTypes = {
    */
   disabled: PropTypes.bool,
 
+  /**
+   * Deprecated in v11. Use 'onClick' instead.
+   */
+  handleClick: deprecate(
```
joshblack: Is `handleClick` an existing prop for `Button` I couldn't find it in the original source file and it seems like we're adding it in here.

joshblack (changes requested):

packages/react/src/components/Button/Button.js
sstrubberg: You're right. I think I mistakenly added it because I saw those event handlers. I'll remove it.

packages/react/src/components/Button/Button.js
```diff
@@ -219,7 +219,7 @@ const Button = React.forwardRef(function Button(
       onMouseLeave: composeEventHandlers([onMouseLeave, handleMouseLeave]),
       onFocus: composeEventHandlers([onFocus, handleFocus]),
       onBlur: composeEventHandlers([onBlur, handleBlur]),
-      onClick: composeEventHandlers([handleClick, onClick]),
+      onClick: composeEventHandlers([onClick, handleClick]),
```
vpicone, line 223: Just out of curiosity what's the significance of the order change here? I haven't messed with the composeEventHandlers helper yet.

packages/react/src/components/Tile/Tile.js
```diff
@@ -295,6 +318,9 @@ SelectableTile.defaultProps = {
   selected: false,
   tabIndex: 0,
   light: false,
+  onClick: () => {},
```
vpicone, line 319: `defaultProps` actually might be deprecated for functional components in the future. [The official position of the react team](https://github.com/yannickcr/eslint-plugin-react/issues/2396) is to use ES6 default values for function components.

Regardless, I think it should be one or the other. Splitting the info up between es6 default values and `defaultProps` is super confusing so this is fine but maybe worth bringing up in dev meeting.

packages/react/src/components/Tile/Tile.js
```diff
@@ -549,7 +579,8 @@ export class ExpandableTile extends Component {
       },
       () => {
         this.setMaxHeight();
-        this.props.handleClick(evt);
+        // TODO: Remove handleClick prop when handleClick is deprecated
+        this.props.handleClick?.(evt) || this.props.onClick(evt);
```
vpicone: Should the same optional chaining exist for `onClick` so we only call it if they pass it in? Would prevent the need for the no-op defaultProp functions.

packages/react/src/components/Tile/Tile.js
```diff
@@ -574,20 +605,20 @@ export class ExpandableTile extends Component {
       tileMaxHeight, // eslint-disable-line
       tilePadding, // eslint-disable-line
       handleClick, // eslint-disable-line
-      onClick,
-      onKeyUp,
+      onClick, // eslint-disable-line
```
vpicone: Probably worth adding a note why its disabled so we can change the rule or fix it in the future

vpicone (commented):
Couple quick questions

packages/react/src/components/Tile/Tile.js
```diff
@@ -174,9 +195,11 @@ export class ClickableTile extends Component {
...
         href={href}
         className={classes}
-        {...other}
-        onClick={this.handleClick}
-        onKeyDown={this.handleKeyDown}>
+        {...rest}
+        // TODO: replace with onClick when handleClick prop is deprecated
+        onClick={composeEventHandlers([this.onClick, this.handleClick])}
+        // TODO: replace with onKeyDown when handleKeyDown prop is deprecated
+        onKeyDown={composeEventHandlers([this.onKeyDown, this.handleKeyDown])}>
```
joshblack: I think these handlers should stay the same since `this.handleClick` is now conditionally calling `props.handleClick` or `props.onClick`. I couldn't find `this.onClick` defined on the class so let me know if I'm missing something 🤔

packages/react/src/components/Tile/Tile.js
```diff
@@ -616,9 +647,11 @@ export class ExpandableTile extends Component {
...
         aria-expanded={isExpanded}
         title={isExpanded ? tileExpandedIconText : tileCollapsedIconText}
-        {...other}
-        onKeyUp={composeEventHandlers([onKeyUp, this.handleKeyUp])}
-        onClick={composeEventHandlers([onClick, this.handleClick])}
+        {...rest}
+        // TODO: replace with onKeyUp when handleKeyup prop is deprecated
+        onKeyUp={composeEventHandlers([this.onKeyUp, this.handleKeyUp])}
+        // TODO: replace with onClick when handleClick prop is deprecated
+        onClick={composeEventHandlers([this.onClick, this.handleClick])}
```
joshblack: Where is `this.onClick` and `this.onKeyUp` coming from in this snippet? 👀

joshblack (changes requested):

packages/react/src/components/Button/Button.js
sstrubberg: nothing. It's just an array of helpers and I saw an opportunity to make it consistent with the others on*'s

packages/react/src/components/Tile/Tile.js
sstrubberg: removed the lonely es6 default value for now. Totally hear you that this should be a workstream to drop `defaultProps` in a not too distant future.

joshblack (approved):

packages/react/src/components/Tile/Tile.js
```diff
@@ -196,10 +196,8 @@ export class ClickableTile extends Component {
         href={href}
         className={classes}
         {...rest}
-        // TODO: replace with onClick when handleClick prop is deprecated
```
vpicone: I actually didn't mind these comments, @joshblack do you have a preference for tracking things we'll need to rip out post-v11.

vpicone (approved):

PR #9636, opened 2021-09-10
fix(Select): fix background issues with inline variant optgroups

### LINKED ISSUE #9112: Inline Select component needs background in Chrome/Windows

### LINKED ISSUE #9625: [a11y]: Pagination dropdown white-on-white for g90 and g100 themes on Windows
Refs https://github.com/carbon-design-system/carbon/issues/9112

### Package

carbon-components

### Browser

Chrome

### Operating System

Windows

### Package version

10.43

None, I don't think EAAC can check select dropdowns?

### Assistive technology

no

### Description

When you open one of Pagination's dropdowns (either items per page, or page number) and then **move the mouse away**, the dropdown becomes white-on-white:

Screen Shot 2021-09-09 at 16 21 28

NB: This only happens with a dark carbon theme (g90 or g100), **and** on Windows.

When hovering a `<select>` , Carbon sets the background color to `rgb(53, 53, 53)` and things work fine.

But when not hovering, Carbon sets each `<select>`'s `background-color` to `transparent`, which (non-intuitively) makes the dropdown gray on the Mac, but white on Windows.

This seems to fix it:

```
.#{$prefix}--select-option {
	background: $background-hover;
}
```

### WCAG 2.1 Violation

Minimum contrast

### CodeSandbox example

https://carbon-components-react.netlify.app/?path=/story/components-pagination--pagination

### Steps to reproduce

Access https://carbon-components-react.netlify.app/?path=/story/components-pagination--pagination and switch the carbon theme to g100.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9625
Closes #9112 

Add in a bit more specificity to override default Firefox styles on `optgroup` as well as sets the correct background-color

#### Changelog

**Changed**

- `background: $background-hover;` as per the issue. Needed to add `optgroup` specificity to correct it in Firefox as well

#### Testing / Reviewing

I have not been able to test in Windows, but this fixes the Firefox issue. I'm aware the `optgroup` does not match the `option`, but it is at least readable now. The background is set via the `select` background color (known Firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1394491))

For testing:
Need to check each theme in these three scenarios, in Windows and Mac (Firefox)
- [ ] Normal `Select`
- [ ] Inline `Select`
- [ ] Pagination (uses inline `Select`)


tw15egan:
It looks like a lot of this is related to https://bugzilla.mozilla.org/show_bug.cgi?id=1394491

Basically, the `select` `optgroup` background styles are set by the `select` background and can't be overridden in Firefox. I've removed the `transparent` styles, and set explicit background in `Pagination` and `Select`. 

For testing:
Need to check each theme in these three scenarios, in Windows and Mac (Firefox)
- [ ] Normal `Select`
- [ ] Inline `Select`
- [ ] Pagination (uses inline `Select`)

tay1orjones (approved):
> I'm aware the optgroup does not match the option, but it is at least readable now.

I get what you're saying after looking at inline:

image

It all looks good for me on Mac - I don't have a windows environment set up yet. @dakahn would you be able to test firefox on windows?

dakahn (approved):

PR #9644, opened 2021-09-10
chore(Tabs): added prop deprecation warnings for v11

sstrubberg:
REF #9535

#### Changelog

- Added deprecation warnings for handleTabClick and handleTabKeyDown to `Tab`.
- Updated the PublicAPI test snapshot.
- Added comments in the appropriate areas to remove deprecations for the next major release.

#### Testing / Reviewing

- Check the `Tab` stories and make sure everything is firing off as it should.
- Make sure tests are passing.

### DIFF
--- a/packages/react/src/components/Tab/Tab.js
+++ b/packages/react/src/components/Tab/Tab.js
@@ -25,16 +25,20 @@ export default class Tab extends React.Component {
     disabled: PropTypes.bool,
 
     /**
-     * A handler that is invoked when a user clicks on the control.
-     * Reserved for usage in Tabs
+     * Deprecated in v11. Use 'handleTabClick' instead.
      */
-    handleTabClick: PropTypes.func,
+    handleTabClick: deprecate(
+      PropTypes.func,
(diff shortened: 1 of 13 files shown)


dakahn (approved):

packages/react/src/components/Tab/Tab.js
```diff
@@ -66,6 +70,18 @@ export default class Tab extends React.Component {
...
+     * Reserved for usage in Tabs
+     */
+    onTabClick: PropTypes.func,
+
+    /**
+     * A handler that is invoked on the key down event for the control.
+     * Reserved for usage in Tabs
+     */
+    onTabKeyDown: PropTypes.func,
+
```
joshblack, line 84: For this component, would we want `handleTabClick` to go to `onClick` versus `onTabClick`?

joshblack (changes requested):

vpicone (approved):
I agree with @joshblack, the `Tabs` container/parent component has an `onClick` prop. We'd want to mirror that hear. Other than that looks good to me!

PR #9652, opened 2021-09-13
9648 add data test id to accordion

### LINKED ISSUE #9648: [Feature Request]: Add data-testid to Accordion
### Summary

We need unique test ids for all components and major areas within the components. This is a first small step in that direction. 

### Justification

To enable support of testing automation in applications that consume the PAL library, we are working to provide a dedicated `data-testid` data attribute for each component. This is also needed in order to start creating a testability contract and we would like the carbon library to offer the same support. See discussion here
https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2446

1. The Accordion will get a default property and value for the `data-testid` property which is added to the outmost html element, in this case the `ul` element.
2. The AccordionItem will also get a data-testid to the list element, the button and the content area. 
3. If the `data-testid` is supplied to the component this id will be used, if not a default value will be used. Internal components will use available testId values from parent to attempt to provide a unique value.
4. The components and subcomponents can be targeted via
[`querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)
or
[`querySelectorAll`](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll).

### Required functionality

_No response_

_No response_

### Available extra resources

_No response_

### Code of Conduct

bjornalm:
Closes #9648
Adds test ids to the accordion component and child components.

#### Changelog

**New**
- Added testId prop to Accordion, Accordion.Skeleton and AccordionItem
- Added unit tests

**Changed**
- Updated snaps

**Removed**
-

#### Testing / Reviewing

Make sure code looks ok and that all relevant components of Accordion have gotten a data-testid attribut that can be a default value or a custom one.


bjornalm:

bjornalm:
recheck

PR #9659, opened 2021-09-14
feat(props): update table expand header props

aledavila:
#### Changelog

This updates table expand header. There are still questions if accordion item: if we truly need the renderExpand prop and what its purpose is. 

**Changed**

- changed prop names for TableHeaderExpand
-  `enableExpando` > `enableToggle`


packages/react/src/components/DataTable/TableExpandHeader.js
```diff
@@ -56,15 +58,28 @@ TableExpandHeader.propTypes = {
...
   className: PropTypes.string,
 
   /**
    * Specify whether an expand all button should be displayed
    */
-  enableExpando: PropTypes.bool,
+  // enableExpando:PropTypes.bool,
+  enableExpando: deprecate(
+    PropTypes.bool,
+    'The `renableExpando` prop has been deprecated in favor of `enableToggle`. This prop will be removed in the next major release.'
```
abbeyhrt:
```suggestion
    'The `enableExpando` prop has been deprecated in favor of `enableToggle`. This prop will be removed in the next major release.'
```

abbeyhrt (approved):
Looks good to me other than a very small typo 🎉

packages/react/src/components/DataTable/TableExpandHeader.js
aledavila: good catch thanks

andreancardona (approved):

PR #9673, opened 2021-09-15
chore: change comment type for stylint-ignore comments

### LINKED ISSUE #9672: New comment polluting compiled CSS output
### Package

@carbon/type

### Browser

_No response_

### Package version

10.35.0

### Description

As part of PR #9480, comments like the following have been added to some SCSS files:
image

The problem is that this style of comment, `/* ... */`, in sass means that the comment is preserved into the CSS output, meaning that the output CSS from our builds now contains many hundreds of copies of this comment and have become substantially larger as a result. These comments only apply at build/lint time and do not need to be preserved into the CSS output. They should be changed to `// ...` style comments, which are discarded during sass compile.

@tw15egan @joshblack FYI

### CodeSandbox example

no

### Steps to reproduce

n/a

### Code of Conduct

dcwarwick:
Closes #9672

#### Changelog

`packages/type/scss/_styles.scss`
`packages/type/scss/modules/_styles.scss`
In both files, replace a CSS-preserved comment (`/* ... */`) with a non-preserved comment (`// ...`) since the comment is only a linter instruction required at build/lint time and not in the final CSS.

#### Testing / Reviewing

Built file `carbon-components/css/carbon-components.css` currently contains 472 copies of the line `/* stylelint-disable number-max-precision */`. After this PR is applied and the files built, it should contain none (and thus be a little over 21KB smaller!).


jnm2377:
Looks like some of the snapshots need to be updated.

joshblack (approved):

sstrubberg (approved):

sstrubberg (changes requested):
Just fix up those testing errors and you're good to go :)

joshblack (approved):

sstrubberg (approved):

dcwarwick:
Thanks @jnm2377 -- sorry, I've been away and just catching up with things now I'm back :-) many thanks all.

PR #9677, opened 2021-09-15
feat(styles): merge v10 and v11 for compat themes

### LINKED ISSUE #9617: Fix bug with setting theme in compatability mode resulting in a noop
This issue was originally reported by @janhassel where the compat theme was not able to have its value changed.

Code to reproduce:

```scss
@use '@carbon/react/scss/reset';
@use '@carbon/react/scss/compat/themes';
@use '@carbon/react/scss/compat/theme' with (
  $theme: themes.$g90
);

body {
  background: theme.$background;
  color: theme.$text-primary;
}
```

Here, we expected the tokens to use values from the g90 theme but they use the value from the white theme instead. It seems like this is related to the issue Jan reported in that the theme would not allow you to correctly retrieve `interactive-01` as a token because the themes were not being merged

joshblack:
Closes #9617 

This updates our compat themes entrypoint to merge v10 and v11 themes to make the use case of setting the theme from the compat entrypoint simpler.

#### Changelog

**New**

**Changed**

- Update Sass to v1.41.0
- Update themes compat entrypoint to merge v10 and v11 themes
- Update test for setting themes in compat mode to new usage

**Removed**

#### Testing / Reviewing

- verify tests pass as expected
- Verify that you can pass in a compat theme and both v10 and v11 tokens are defined, and v11 tokens match the theme used from compat


sstrubberg (commented):
The timing of this one is quite perfect because @jnm2377 and I are working on the compat example right now. Is there anything that should be updated in [the docs](https://github.com/carbon-design-system/carbon/blob/main/packages/styles/docs/sass.md#compatibility) to reflect these changes? Specifically the changes you've made to the test.

joshblack:
@sstrubberg good point! I'll update now.

sstrubberg (approved):

dakahn (approved):

PR #9685, opened 2021-09-16
feat(@carbon/upgrade): add check for git status before upgrading

jnm2377:
This PR adds a feature to our `@carbon/upgrade` cli that checks a user's git status before moving forward with updates. Similar to [react-codemon](https://github.com/reactjs/react-codemod/blob/master/bin/cli.js#L22).

This implementation doesn't allow for any forceful continuation of updating if git status is not clean. It simply exits right from the start. Curious if people think this is ok, or if there should be an option to keep going. 😬 

#### Changelog

- adds `is-git-clean` package
- runs check for git status on init
- exits process if git status is not clean
- continues with upgrade if git status is clean

#### Testing / Reviewing

For testing, you have to checkout locally.

1. make a small change in any `carbon` monorepo file and save it, but don't commit it.
2. in your terminal, run `./packages/upgrade/bin/carbon-upgrade.js` from your root
3. you should see the cli console.log exit bc git status is not clean
4. undo your changes, so git status is clean
5. run cli command again
6. you should see it continue with the update process now

https://user-images.githubusercontent.com/32556167/133674484-27045b53-0fb8-48e8-90f3-58c335ff46dd.mov


joshblack (approved):
Looks great!

sstrubberg (approved):

PR #9687, opened 2021-09-16
feat(react): add usePrefix hook and examples

dakahn:
Adds a `useRef()` hook as well as examples of how to use it for both class and functional React components 🏄🏾 

#### Testing / Reviewing

in the carbon-react preview click into Accordion, Accordion-Skeleton, and ContentSwitcher and inspect those components using your browsers dev tools. Make sure you see the classname prefix applied correctly


packages/react/src/internal/usePrefix.js
```diff
@@ -0,0 +1,8 @@
+import { settings } from 'carbon-components';
```
joshblack, line 8: Make sure to include the copyright banner for the file 👍

joshblack (changes requested):

joshblack:
@dakahn I think the CI failure comes from a snapshot that you need to update 👍

packages/components/src/globals/grid/__tests__/__snapshots__/grid-test.js.snap
```diff
@@ -17,7 +16,7 @@ h2 {
 h3 {
   font-size: 1.75rem;
   font-weight: 400;
-  line-height: 1.28572;
+  line-height: 1.29;
```
joshblack: I think you may need to rebuild and then update snapshots, some of these changes seem to be unrelated

joshblack (changes requested):

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -68,6 +66,8 @@ export default class ContentSwitcher extends React.Component {
     size: PropTypes.oneOf(['sm', 'md', 'lg', 'xl']),
   };
 
+  static contextType = PrefixContext;
```
jnm2377, line 69: Nothing wrong here, just curious why some components need to use `PrefixContext` directly instead of `usePrefix`? Is it bc they're class components?

jnm2377:
hmmm looks like one of the ContentSwitcher tests is failing bc it's expecting a className... might be the prefix specifically that's not showing up the same?

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -130,6 +130,7 @@ export default class ContentSwitcher extends React.Component {
   };
 
   render() {
+    const prefix = this.context;
```
jnm2377, line 133: Wondering if you also need to update the classNames in this file to be `this.prefix` like you did for `ComposedModal` over in this PR: https://github.com/carbon-design-system/carbon/pull/9691/files. I think that may be why tests are failing?

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
dakahn: Yeah, exactly! Class components can't use hooks so they have to grab context via the more old school this.context method

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
jnm2377: see: https://github.com/carbon-design-system/carbon/pull/9687#issuecomment-922062739

dakahn: setting `contextType` to `PrefixContext` should mean that `this.context` is our prefix for everything in that scope. In a few other files I then took the extra step of either assigning `prefix = this.context` for extra readability, but I'll probably revert that change in the other PRs

dakahn: looking into the failed tests now

joshblack (approved):

dakahn:
reviewers, i added some unit tests for `usePrefix` 👍🏾

joshblack (approved):

jnm2377 (approved):

PR #9689, opened 2021-09-17
feat(Carbon-React): add prefix hook refactor Loading -> UnorderedList

dakahn:
ref #9687 

Refactors prefix to use context for components Loading thru UnorderedList 🏄🏾 

#### Testing / Reviewing

Test that the components function and have classname prefixes as expected in the corresponding storybooks


jnm2377:
Ah, looks like tests are failing bc `One of the commands failed: Error: Cannot find module '../../internal/usePrefix'`, which is being added in this PR: https://github.com/carbon-design-system/carbon/pull/9687

So I think this is blocked until that one is merged.

joshblack (approved):

jnm2377 (approved):

PR #9690, opened 2021-09-17
feat(UIShell): add prefix context refactor

dakahn:
ref #9687 

Refactors prefix to use context for UIShell and it's parts 🏄🏾 

#### Testing / Reviewing

Test that the components function and have classname prefixes as expected in the corresponding storybooks


packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -53,8 +47,13 @@ class HeaderMenu extends React.Component {
     tabIndex: PropTypes.number,
   };
 
+  static contextType = PrefixContext;
+  prefix = this.context;
```
joshblack: This is a new pattern to me for setting a class instance property based on context, is this a common pattern? I've always seen it used as `this.context` or if it needs a rename something like:

```js
const { context: prefix } = this;
```

My worry is that this would not change the value of `this.prefix` if the context value changes since this is only called when the class is initialized

packages/react/src/components/UIShell/HeaderNavigation.js
```diff
@@ -40,6 +38,8 @@ export default class HeaderNavigation extends React.Component {
     };
   }
 
+  static contextType = PrefixContext;
+  prefix = this.context;
```
joshblack: Same question/comment as above

joshblack (changes requested):

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -53,8 +47,14 @@ class HeaderMenu extends React.Component {
     tabIndex: PropTypes.number,
   };
 
+  static contextType = PrefixContext;
+
+  static defaultRenderMenuContent = () => (
+    <ChevronDown16 className={`${this.context}--header__menu-arrow`} />
```
joshblack: When it's a static method, I don't think `this.context` will be defined (because it's static).

Down below, what you could do instead is `MenuContent ? <MenuContent /> ? <ChevronDown16 className={`${this.context}--header__menu-arrow`} />` and I think that would work

packages/react/src/components/UIShell/__tests__/__snapshots__/HeaderMenu-test.js.snap
```diff
@@ -25,7 +25,7 @@ exports[`HeaderMenu should render 1`] = `
               test
               <svg
                 aria-hidden="true"
-                class="bx--header__menu-arrow"
+                class="undefined--header__menu-arrow"
```
joshblack: I think this is related to the static changes where `this.context` is undefined

joshblack (changes requested):

joshblack (approved):

sstrubberg (approved):

PR #9702, opened 2021-09-20
fix(datepicker): pass through flatpickr inline prop

### LINKED ISSUE #5713: Pass down options to Flatpickr on the DatePicker component

tay1orjones:
Closes #5713 

Passes the `inline` prop through to flatpickr. As stated in the linked issue, `inline` isn't something formally baked from a design standpoint but we shouldn't block teams from passing this through to meet their own unique use cases. 

To this end, I didn't add `inline` as part of the component API, but instead passed it through by spreading `...rest` props and checking for `inline` in componentDid/WillUpdate.

#### Changelog

**Changed**

- datepicker: pass through inline prop

#### Testing / Reviewing

- Tests should pass, `inline` should be able to be passed through to flatpickr from DatePicker


joshblack (approved):

abbeyhrt (approved):
Looks good to me! 🎉

PR #9706, opened 2021-09-20
fix(tile): remove position relative

### LINKED ISSUE #9676: [Bug]: Tile styling interfering with IBM Cloud Cognitive PageHeader component
### Package

carbon-components-react

### Browser

Chrome

### Package version

7.43.0

### Description

I am using the @carbon/ibm-cloud-cognitive `PageHeader` component within my application along with a `Grid` that contains a couple of `Tile` components to be used within rows and columns. 

I was having an issue where my grid content was overlapping my headere. The header has a `position: sticky` and I figured out that the tiles were the culprit. They have a `position: relative` at the top-most level that's making them overlap. I was able to fix this issue by unsetting the position field. However, I suppose the tiles should ideally work fine with the header.

Screen Shot 2021-09-15 at 5 03 14 PM

### CodeSandbox example

https://codesandbox.io/s/elegant-montalcini-5sfqd?file=/src/components/PageDetails.scss

### Steps to reproduce

1. Import `PageHeader` from @carbon/ibm-cloud-cognitive
2. Create a class that renders the PageHeader and a Grid with tiles within it.
3. Try to scroll

### Code of Conduct

abbeyhrt:
Closes #9676 

This PR removes `position: relative` from Tile's container, `bx--tile` and adds it to `bx--tile-selectable` since it is needed to position checkmark boxes, radio buttons, etc in the selectable tiles. 

**Changed**

- removed `position: relative` from `bx--tile` 
- change `$carbon--spacing-05` to `$spacing-05` because I can

#### Testing / Reviewing

Check to make sure that everything looks the same with this positioning removed.


joshblack:
@abbeyhrt I think we use `position: relative` for things like selectable or expandable tile where we use `position: absolute` to position the checkmark or chevron (just looking at the tile styles that would be my guess, at least)

abbeyhrt:
@joshblack gotcha, that makes sense. I added `position: relative` to `bx--tile-selectable` and that fixed the positioning of the checkmarks and whatnot. Let me know if that works!

joshblack (approved):

dakahn (approved):

PR #9710, opened 2021-09-21
Add color for filename text

makeupsomething:
#### Changelog

**Changed**

In a dark theme like Gray 100 the text for the filename is dark on a dark background and is not readable.
If this is not the right place for this change then feel free to close the PR 🙏


tay1orjones:
I can repro the bug in the new carbon-react-next storybook that uses the `@carbon/styles` package.

In the old storybook it's inhereited from the body
Screen Shot 2021-09-23 at 3 09 53 PM

In the new storybook, a different `color` declaration is defined with a dark color. You can see your override here too flipping it back to a light color
image

I'm not sure why the root issue is happening though. The override fixes it, but would be nice to avoid this element from ever receiving the dark text style. Perhaps it's coming from the type style mixin?

Issue #9512, opened 2021-08-20
[Datatable][isSortable] console error occurs when a column has been sorted and the user then deletes the header


yonghuay:
## What package(s) are you using?

- [x] `carbon-components`
- [x] `carbon-components-react`

## Detailed description

I have got an issue with removing a previously sorted header. Basically, when a column has been sorted and the user then deletes the header, a console error occurs.
I have created a sample code in codesandbox. 

https://codesandbox.io/s/rough-bird-x6of7?file=/index.js

## Steps to reproduce the issue

1. Create a data table.
2. Enable the sort function.
3. Add a new column function and remove the column function.
4. Sort the newly added column and then remove it.
Results: console error occurs.

Please create a reduced test case in CodeSandbox
https://codesandbox.io/s/rough-bird-x6of7?file=/index.js

## Additional information

- Screenshots or code

![Untitled 2](https://user-images.githubusercontent.com/32554015/130226023-164cc4d9-ee2f-4bab-84dc-cfd559ddf7a1.gif)

- Notes

tay1orjones:
Hi @yonghuay, yeah this is because it's trying to sort on a column that is no longer there. A few options:

* Place a `key` on the table and change it's value when you delete a header. This will create a new instance of the component with the sorting by default as it was on initial render.
* The `sortBy` action is made available in your render prop function so that you can [programmatically sort the table](https://react.carbondesignsystem.com/?path=/docs/components-datatable-sorting--usage#programmatic-sorting) by a different header. What I would recommend is on deletion of a header that was sorted on, use `sortBy` to sort the table via a different available header.

Let me know what you think. I'll close this for now but we can reopen at any time if there's something actionable we can do besides the options I outlined above.

yonghuay:
Hi Taylor, @tay1orjones 

Thanks for the suggestions, I passed them on to my colleague. He told me that he used option 1 to solved this issue. 

Thank you very much.

PR #9714, opened 2021-09-21
fix(DataTable): guard npe when cells being sorted do not exist

### LINKED ISSUE #9713: [Bug]: DataTable throws exception if the sorted column is not rendered
### Package

carbon-components-react

### Browser

Chrome

### Package version

v7.7.0

### Description

DataTable throws an NPE when the cells being sorted do not exist.

### CodeSandbox example

https://ijsul.csb.app/

### Steps to reproduce

In the code sandbox, click "Port" column header to sort by Port, then click the blue button at the bottom to see the exception.

### Code of Conduct

holmansze:
Closes https://github.com/carbon-design-system/carbon/issues/9713

DataTable throws an NPE when the cells being sorted do not exist.

We have a table settings feature for DataTable where users can show or hide table columns. When the column is hidden, it is removed from `headers` prop to DataTable. If this column happens to be used for sorting, an NPE is thrown.

The fix is simply an NPE check in the sorting code. When the cells being sorted do not exist, the sort compare function returns 0 (equal), and the table renders in the original (unsorted) order.

#### Changelog

**New**

- n/a

**Changed**

- added NPE check in the sorting code

**Removed**

- n/a

#### Testing / Reviewing

New code tested by removing the column being sorted from the `headers` prop. Worked as expected - table renders properly without sorting. When the column is added back, the table becomes sorted.


holmansze:

hoiwingh:

hoiwingh:
recheck

tay1orjones (approved):
Thanks for taking a look at this! Previously I've recommended the use of a `key` or the `sortBy` function to correct the sort state when a column is removed from the table.
https://github.com/carbon-design-system/carbon/issues/9512#issuecomment-914656452

This PR definitely makes sense - the use of a `key` is no longer necessary, and there is still the option to use the `sortBy` function if reverting to unsorted is not desired.

It would be nice if a new test case was added to the sorting tests to prevent any regression of this functionality, but won't hold up a merge https://github.com/carbon-design-system/carbon/blob/eff767ddfc4bc41b8ebffff72017591ca86bb9e1/packages/react/src/components/DataTable/state/__tests__/sorting-test.js#L8

holmansze:
@tay1orjones - thanks for your review! Unit tests added.

jnm2377 (approved):

jnm2377:
Thanks for the contribution!

holmansze:
Thanks @tay1orjones and @jnm2377 !

PR #9719, opened 2021-09-22
feat(styles,carbon-react): switch to cds prefix

joshblack:
Update our beta packages to use `cds` as the default prefix instead of `bx`. This PR includes two significant changes:

- Update the default `$prefix` variable in `@carbon/styles` to be `cds`. It is still configurable under `scss/_config.scss`
- Add a temporary `src/prefix.js` file in `@carbon/react` that changes the prefix in `carbon-components`. This will be removed in v11 and replaced by `PrefixContext`

#### Changelog

**New**

**Changed**

- Switch default prefix from `bx` to `cds` for styles and React

**Removed**

#### Testing / Reviewing

- View the deployed v11 storybook and verify that:
  - Styles are using the cds prefix
  - CSS Custom Properties are using the cds prefix
  - React components are using the cds prefix

### DIFF
--- a/packages/carbon-react/package.json
+++ b/packages/carbon-react/package.json
@@ -92,6 +92,8 @@
   },
   "sideEffects": [
     "es/feature-flags.js",
-    "lib/feature-flags.js"
+    "lib/feature-flags.js",
+    "es/prefix.js",
+    "lib/prefix.js"
   ]
 }
--- a/packages/carbon-react/.storybook/preview.js
+++ b/packages/carbon-react/.storybook/preview.js
@@ -11,6 +11,9 @@ import { configureActions } from '@storybook/addon-actions';
 import { white, g10, g90, g100 } from '@carbon/themes';
 import React from 'react';
 import { breakpoints } from '@carbon/layout';
+import { settings } from 'carbon-components';
+
+settings.prefix = 'cds';
 
 export const globalTypes = {
   locale: {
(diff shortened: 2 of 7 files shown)


packages/carbon-react/package.json
```diff
@@ -92,6 +92,8 @@
   },
   "sideEffects": [
     "es/feature-flags.js",
-    "lib/feature-flags.js"
+    "lib/feature-flags.js",
+    "es/prefix.js",
```
abbeyhrt, line 96: Out of curiosity, why do these files (really all of them) have to be listed as sideEffects?

tay1orjones: I'm also curious. My guess is since it's pulled from `carbon-components` these should never be pruned when tree shaking?

tay1orjones (approved):

jnm2377 (approved):
YAY!!! It's working! 🔥

packages/carbon-react/package.json
joshblack: @abbeyhrt @tay1orjones basically these modules are not "pure" because they cause side effects in other modules (feature flags, components).

Bundlers like webpack/rollup will assume modules are pure by default if `sideEffects: false` is listed. This means that if exports from these entrypoints are used that the bundler will actually drop the file from the final output. This is not what we want because we always want these files to be included and their side effects run.

Annotating these files as having side effects ensures that they are included in the final bundle even though we aren't using exports from these files.

Hope this helps!

Issue #9712, opened 2021-09-21
Upgrade class components to functional components


joshblack:
**Why are we doing this?**

According to the React team, switching a component from a class component to a functional component is technically a breaking change. This is because class components accept a `ref` that points to the class instance by default. There is no way to opt-out of this behavior. As a result, if we switched a component to a functional component it would no longer have a `ref`, and existing code that uses a `ref` would break. Similarly, using `React.forwardRef` would produce a component where the `ref` would no longer point to a class instance and code could similarly break.

We have moved some components over to the new format despite these situations. The rationale at the time was that we weren't intending people to use these parts of a component (specifically class instance methods) and if teams were trying to use the `ref` to find a node in the document their code would work if we used `React.forwardRef`.

With v11, it seemed like a great opportunity to convert our remaining class components to the new functional component style with hooks.

**Effort size legend**
❇️ - sm
✴️ - md
🆘 - lg

**Tasks**

We would like to convert the following components to functional components that use hooks where appropriate:

- [x] ✴️ src/components/UIShell/SideNavMenu.js @joshblack 
  - [x] ✴️ SideNavMenu 
- [x] ❇️ src/components/UIShell/HeaderNavigation.js @joshblack 
  - [x] ❇️ HeaderNavigation 
- [ ] ❇️ src/components/UIShell/HeaderMenu.js @joshblack 
  - [ ] ❇️ HeaderMenu
- [x] ~~src/components/ToggleSmall/ToggleSmall.Skeleton.js~~ (this component is deprecated)
  - [x] ~~ToggleSmallSkeleton~~
- [x] ~~src/components/Toggle/Toggle.Skeleton.js~~ (this component is deprecated)
  - [x] ~~ToggleSkeleton~~
- [x] ~~src/components/Tooltip/Tooltip.js~~
  - [x] ~~Tooltip~~ (component is being replaced in another update)
- [ ] ✴️ src/internal/ClickListener.js
  - [ ] ✴️ ClickListener
- [x] ~~src/internal/FloatingMenu.js~~
  - [x] ~~FloatingMenu~~ (component will be replaced by popover)
- [x] ~~src/components/ToolbarSearch/ToolbarSearch.js~~
  - [x] ~~ToolbarSearch~~
- [x] ~~src/components/Toggle/Toggle.js~~
  - [x] ~~Toggle~~ (already covered in Toggle/next)
[...]

PR #9722, opened 2021-09-22
feat(Tab): make tab functional component

jnm2377:
Ref #9712 

#### Changelog

- refactors `Tab` to be a functional component
- adds new tests

**Removed**

- `role` prop bc it was no longer being used (changed in this PR: https://github.com/carbon-design-system/carbon/pull/6989/files) + added deprecation notice for it
-  ~`scrollable` classNames since there's a TODO: note to remove in major release~
- `href` prop bc it was deprecated in v10, but left deprecation note for it
- `renderAnchor` bc it was deprecated in v10, but left deprecation note for it (PR: https://github.com/carbon-design-system/carbon/pull/7418/files)

#### Testing / Reviewing

1. Pull PR down locally
2. in `packages/react` run `CARBON_ENABLE_V11_RELEASE=true yarn storybook` to use the v11/next component
3. test out the Tabs story:
- [ ] ensure styles are working correctly
- [ ] ensure click selection is working
- [ ] ensure keyboard navigation (arrow keys to nav between tabs) is working
- [ ] ensure tabs are responsive (scrollable)
- [ ] ensure scrollable tabs update L/R overflow buttons accordingly
     - [ ] test scrolling via mouse/trackpad
     - [ ] test scrolling via overflow button click
 - [ ] ensure tests are passing

#### To-do:
- [ ] ~update styles to work for non `--scrollable` classNames in v11~
- [x] add new RTL tests for Tab
- [x] add new RTL tests for Tabs


dakahn (approved):
Nice job!

packages/react/src/components/Tab/next/Tab-test.js
```diff
@@ -0,0 +1,206 @@
+import React from 'react';
+import { default as Tab } from './Tab';
+import { cleanup, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+describe('Tab', () => {
+  afterEach(cleanup);
```
tay1orjones: This is [called automatically](https://testing-library.com/docs/react-testing-library/api/#cleanup) now so we shouldn't need this anymore

packages/react/src/components/Tab/next/Tab.js
```diff
@@ -0,0 +1,187 @@
...
+  onKeyDown: PropTypes.func,
+
+  /*
+   * An optional parameter to allow overriding the anchor rendering.
+   * Useful for using Tab along with react-router or other client
+   * side router libraries.
+   **/
+  renderAnchor: deprecate(
+    PropTypes.func,
+    'The renderAnchor prop has been deprecated in favor of renderButton. It has been removed in v11.'
```
tay1orjones: Since this is `next`, effectively v11 - are we saying `renderAnchor` is deprecated in v10 and will be removed in v11? Or is it to be deprecated in v11 and removed in v12?

If it's going to be deprecated in v10 and removed in v11, could we just delete the prop here in the `next` version of the component?

packages/react/src/components/Tab/next/Tab.js
```diff
@@ -0,0 +1,187 @@
...
+   **/
+  renderContent: PropTypes.func,
+
+  /**
+   * Provide an accessibility role for your Tab
+   */
+  role: deprecate(
+    PropTypes.string,
+    'The role prop has been deprecated and removed in v11. Role is now built into the element.'
+  ),
```
tay1orjones: Same thing here, I think we can just delete this since it's in the `next` version of the component?

packages/react/src/components/Tab/next/Tab.js
```diff
@@ -0,0 +1,187 @@
...
+   */
+  handleTabKeyDown: PropTypes.func,
+
+  /**
+   * Provide a string that represents the `href` of the Tab
+   */
+  href: deprecate(
+    PropTypes.string,
+    'The href prop has been deprecated and removed in v11.'
+  ),
```
tay1orjones: Same thing here, I think we can just delete this since it's in the next version of the component?

packages/react/src/components/Tabs/index.js
```diff
@@ -5,5 +5,15 @@
...
+import { default as TabsNext } from './next/Tabs';
+import { default as TabsClassic } from './Tabs';
+
+const Tabs = FeatureFlags.enabled('enable-v11-release')
+  ? TabsNext
+  : TabsClassic;
+
 export * from './Tabs.Skeleton';
-export default from './Tabs';
+// export default from './Tabs';
```
tay1orjones:
```suggestion
```

packages/react/src/components/Tabs/next/Tabs-test.js
```diff
@@ -0,0 +1,264 @@
+import React from 'react';
+import { default as Tabs } from './Tabs';
+import { default as Tab } from '../../Tab/next/Tab';
+import { render, screen } from '@testing-library/react';
+import { fireEvent } from '@testing-library/dom';
+
+describe('Tabs', () => {
```
tay1orjones, line 7: Love these new tests! 🎉  Looks awesome, thanks for taking it on

packages/react/src/components/Tabs/next/Tabs.js
```diff
@@ -0,0 +1,615 @@
...
+    }
+    return 0;
+  };
+
+  const getTabAt = useCallback(
+    (index, useFresh) =>
+      (!useFresh && tabs.current[index]) ||
+      React.Children.toArray(children)[index],
+    [tabs, children]
+  );
```
tay1orjones: Did you discover what/if `useFresh` was used for?

tay1orjones (commented):
Hefty refactor! Looks awesome 🔥  I pulled up the existing storybook and started up the local storybook with the flag and went through the same interactions between them both with keyboard and mouse, small screens w/ scrolling, etc and it appears to be functionally the same without regressions.

Just a couple minor comments and questions below. Otherwise looks good to merge!

packages/react/src/components/Tabs/next/Tabs-test.js
jnm2377: thank you @tay1orjones! It was actually pretty fun to work on and learn it 😅

packages/react/src/components/Tabs/next/Tabs.js
jnm2377: @tay1orjones I did not... 👀

packages/react/src/components/Tab/Tab.js
```diff
@@ -77,12 +77,12 @@ export default class Tab extends React.Component {
     /*
      * An optional parameter to allow overriding the content rendering.
      **/
-    renderContent: PropTypes.func,
+    renderContent: deprecate(PropTypes.func),
```
jnm2377: @tay1orjones I deprecated these in v10 Tabs since they're not being used at all... I'm not sure if they were removed and never deprecated. That seems to be a pattern for some of these props. i.e. role is hard coded into each element. renderContent prop is specified but never used (and since it's specified it's not being passed as `...other`). Lmk your thoughts on this.

packages/react/src/components/Tab/Tab.js
```diff
@@ -77,12 +77,12 @@ export default class Tab extends React.Component {
...
 
     /**
      * Provide an accessibility role for your Tab
      */
-    role: PropTypes.string.isRequired,
+    role: deprecate(PropTypes.string),
```
jnm2377, line 85: same as above

jnm2377: this used to set the role way way wayyyyy back before we had a monorepo: https://github.com/carbon-design-system/carbon-components-react/pull/152/files

but at some point we decided to hard code it... here: https://github.com/carbon-design-system/carbon-components-react/pull/368 and the prop was never deprecated.

packages/react/src/components/Tab/Tab.js
jnm2377: jk, this prop isn't being used in this file but it's used in `Tabs.js` when we clone the children tabs.

packages/react/src/components/Tabs/next/Tabs.js
tay1orjones: @jnm2377 I don't see it being used anywhere, I think we can remove it. Crawling back in the git history a bit, it was [used a long time ago](https://github.com/carbon-design-system/carbon/blame/74740a755b1411119bcfbe53dda3c353c0644710/packages/react/src/components/Tabs/Tabs.js#L294).

tay1orjones (approved):

packages/react/src/components/Tabs/next/Tabs.js
jnm2377: Oof @tay1orjones way to find that! That was ages ago. 😅 Sounds good! I'll remove 👍🏽

PR #9726, opened 2021-09-23
fix(layer): use correct prefix

janhassel:
Removes the hard-coded `bx` prefix from the `<Layer>` component, ensuring that is uses the correct prefix defined in the settings.

#### Changelog

**Changed**

- Updated hard-coded prefix with prefix from `carbon-components`

#### Testing / Reviewing

Ensure correct prefix is emitted based on the `carbon-components` settings

### DIFF
--- a/packages/carbon-react/src/components/Layer/index.js
+++ b/packages/carbon-react/src/components/Layer/index.js
@@ -8,6 +8,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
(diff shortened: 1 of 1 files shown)


joshblack:
@janhassel I made #9728 which should address this for Theme and Layer, let me know if it's missing any!

janhassel:
@joshblack Looks great, I wasn't aware of the `usePrefix` hook yet. Gonna close this one then.

PR #9731, opened 2021-09-23
feat(carbon-react): remove label from default props

### LINKED ISSUE #9466: Do not provide accessible labels by default, for example: "Provide icon description"
There are several situations where we use `defaultProps` to provide a placeholder accessible label. In v11, these props should be treated as required with no default props.

```jsx
function TestComponent() {
  // ...
}

// From
TestComponent.propTypes = {
  iconDescription: PropTypes.string,
};

TestComponent.defaultProps = {
  iconDescription: 'Provide a description',
};

// To
TestComponent.propTypes = {
  iconDescription: PropTypes.string.isRequired,
};

TestComponent.defaultProps = {
};
```

This may be tricky to get right with feature flags and most likely will require using the top-level feature flag package instead of being able to use a hook 😞 

Below is a list of components that currently do this:

- [ ] FileUploader
- [ ] SelectItemGroup
- [ ] NumberInput
- [ ] Switch
- [ ] OverflowMenuItem

Note: there may have been some that have been missed, this was a quick search!

dakahn:
partially closes #9466 

Removes various labelling from `defaultProps` for v11 🏄🏾 

#### Changelog

**Changed**

- src/components/OverflowMenu/OverflowMenu.js
- src/components/Switch/Switch.js

#### Testing / Reviewing

In a v11 feature flagged environment test and see (via the Storys i'd suppose) if omitting proper labeling causes a console error from React 🦖 

attn @abbeyhrt


abbeyhrt (changes requested):
Looks good to me! I think there was another one in OverflowMenuItem that needs to be updated too https://github.com/carbon-design-system/carbon/blob/b510ad99b6723aeaee95ed4fea672426a768c56c/packages/react/src/components/OverflowMenuItem/OverflowMenuItem.js#L103

packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -97,7 +98,9 @@ class OverflowMenu extends Component {
     /**
      * The ARIA label.
      */
-    ariaLabel: PropTypes.string,
+    ariaLabel: FeatureFlags.enabled('enable-v11-release')
+      ? PropTypes.string.isRequired
+      : PropTypes.string,
```
joshblack, line 103: In this component, are we looking for `ariaLabel` or `iconDescription` to be the accessibility label for the component?

packages/react/src/components/Switch/Switch.js
```diff
@@ -112,7 +113,7 @@ Switch.propTypes = {
 
 Switch.defaultProps = {
   selected: false,
-  text: 'Provide text',
+  text: FeatureFlags.enabled('enable-v11-release') ? null : 'Provide text',
```
joshblack: This change might not be relevant with the Content Switcher refactor, potentially, so you might be able to leave it alone since this should be a `children` prop anyways

joshblack (changes requested):

packages/react/src/components/OverflowMenu/OverflowMenu.js
dakahn: I figured iconDescription was not necessarily the accessibility label

dakahn: it could be specified but we would still allow a default?

joshblack (approved):
Sorry I blanked on this! Looks good

abbeyhrt (approved):
Ah sorry! Looks good to me, sorry for ignoring it for so long!

Issue #9641, opened 2021-09-10
Create a consistent API for event handlers  in v11


joshblack:
We would like any `on*` or event-handler-related props to have consistent API across the codebase. During an initial review, we noticed that several patterns are floating around the codebase that has been developed throughout the years:

- `(value) => void`
- `(value, event) => void`
- `(event) => void`
- `(event, value) => void`

**Checklist**

- [ ] Conduct an initial audit of existing conventions for event handlers
- [ ] Document our conventions that guide our project refactor
- [ ] Refactor the codebase

joshblack:
## Audit

- [x] Accordion
- [x] AccordionItem
  - Current: `onHeadingClick: ({ isOpen: boolean, event: Event }) => void`
  - Proposed: should `isOpen` just be value (type `ValueState`)
  - Proposed: should we use `onChange` for all controlled?
  - `rest.handleAnimationEnd` -> `onAnimationEnd`
- [x] AccordionSkeleton
- [x] AspectRatio
- [x] Breadcrumb
- [x] BreadcrumbItem
- [x] BreadcrumbSkeleton
- [x] Button
- [x] ButtonSet
- [x] ButtonSkeleton
- [ ] Checkbox
  -  Proposed `onChange: (event, { checked: boolean, id: string })`
  - Current `onChange: (state: boolean, id: string, event: Event)`
  - Should we do `onChange` for native event and `onValueChange` for controlled?
- [ ] CheckboxSkeleton
- [ ] ClickableTile
- [ ] CodeSnippet
- [ ] CodeSnippetSkeleton
- [ ] Column
- [ ] ComboBox
- [ ] ComposedModal
- [ ] Content
- [ ] ContentSwitcher
- [ ] ControlledPasswordInput
- [ ] Copy
- [ ] CopyButton
- [ ] DataTable
- [ ] DataTableSkeleton
- [ ] DatePicker
- [ ] DatePickerInput
- [ ] DatePickerSkeleton
- [ ] DirectionButton
- [ ] Dropdown
- [ ] DropdownSkeleton
- [ ] ErrorBoundary
- [ ] ExpandableTile
- [ ] FeatureFlags
- [ ] Filename
- [ ] FileUploader
- [ ] FileUploaderButton
- [ ] FileUploaderDropContainer
- [ ] FileUploaderItem
- [ ] FileUploaderSkeleton
- [ ] FilterableMultiSelect
- [ ] FluidForm
- [ ] Form
[...]

PR #9737, opened 2021-09-24
feat(search): add onClear callback prop to Search/DataTableSearch

### LINKED ISSUE #9738: [Search]: add onClear callback when input is cleared.
### Summary

We need a way in the carbon-addons-iot-react repo to only trigger search when a user hits Enter or blurs the input. The problem is with all changes pushed through onChange there's no way to tell if they backspaced to delete everything or if the "Clear search input" button was clicked.

### Justification

To improve performance in complex tables by only trigger server calls when the user specifically hits enter or blurs the input. The onClear callback allows us to know the difference between multiple backspaces and a clear. For example, they may want to backspace and delete everything, type a new search and then hit 'Enter' to trigger it.

_No response_

### Required functionality

_No response_

_No response_

### Available extra resources

I can write the PR.

### Code of Conduct

kevinsperrine:
Closes #9738

We needed a way in the `carbon-addons-iot-react` repo to only trigger search when a user hits `Enter` or blurs the input. The problem was with all changes pushed through `onChange` there's no way to tell if they backspaced to delete everything or if the "Clear search input" button was clicked. This PR adds an `onClear` callback prop to distinguish between the two.

#### Changelog

**New**

- added onClear to the Search/DataTableSearch components

#### Testing / Reviewing

Added new tests to Search component to confirm callback is called.


tay1orjones (commented):
Thanks for the PR @kevinsperrine ! 🎉  

This addition makes sense to me. Another option would be to modify the signature of the `onChange` to include a new `isClear` param or something, but based on https://github.com/carbon-design-system/carbon/issues/9641 I don't think we'd want to do that.

I see the `onClear` fire in the DataTable storybook, but I don't see it firing in the Search story?

https://user-images.githubusercontent.com/3360588/134743664-993ddcd3-90ec-4570-aad6-c91d8f0bb3fd.mov

Otherwise this looks solid 👍

kevinsperrine:
> I see the `onClear` fire in the DataTable storybook, but I don't see it firing in the Search story?

@tay1orjones Thanks for the heads up. I have no idea what's going on in the deploy preview. Locally, there are onChange, onKeyDown, and the onClear handlers being fired, but only onChange in the preview. 🤔 

search-actions

tay1orjones (approved):
@kevinsperrine I'm not sure what's going on there. Locally I see the same though; `onKeyDown` and `onClear` logging as expected to the actions pane. It might be something with the deploy preview environment and the fact that you're a first time contributor so some status checks had to be approved to be ran. The production storybook was/is logging things fine. 

I'm good to merge this, if the storybook issue persists in production I can open a new issue to hunt it down.

abbeyhrt (approved):
LGTM! Thanks for the contribtion!

packages/react/src/components/Search/Search-test.js
```diff
@@ -208,6 +210,12 @@ describe('Search', () => {
         input.simulate('change', eventObject);
         expect(onChange).toHaveBeenCalledWith(eventObject);
       });
+
+      it('should invoke onClear when input value is cleared', () => {
+        wrapper.setProps({ defaultValue: 'test' });
+        wrapper.find('button').simulate('click');
+        expect(onClear).toHaveBeenCalled();
+      });
```
jnm2377: it looks like this test is failing

Screen Shot 2021-10-06 at 1 39 14 AM

jnm2377: @kevinsperrine we can merge once this is updated and tests pass :)

kevinsperrine: @jnm2377 Thanks. It's fixed now. I adjusted the test to account for shallow rendering and all is well.

Issue #9666, opened 2021-09-15
[a11y]: Tabbed content does not have an associated name


danielannan17:
### Package

carbon-components-react

### Browser

Firefox

### Operating System

MacOS

### Package version

v7.43.0

### Automated testing tool and ruleset

IBM Equal Access Accessbility Checker

### Description

Components with the ARIA role are expected to have an associated name

### CodeSandbox example

https://8u88r.csb.app/

### Steps to reproduce

1. Download IBM Equal Access Accessibility Checker firefox extension
2. Visit https://8u88r.csb.app/
3. Open chrome dev tools
4. Navigate to 'Accessibility Assessment' tab
5. Click 'Scan' - you should now see the violation
<img width="548" alt="Screenshot 2021-09-15 at 08 55 12" src="https://user-images.githubusercontent.com/11006124/133396372-d16d8df5-e749-4226-b0b0-641e05e3f56e.png">

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

abbeyhrt:
Hi @danielannan17 👋  Thanks for making this issue!  Tabbed content is aria-labelled by `Tab` so `Tab` needs to have an `id` for it to correctly label the tab content. You can see this in the code sandbox you sent, if you add an `id` to `Tab` the accessibility violation goes away. [Example with id added](https://codesandbox.io/embed/eager-wildflower-4meuw?fontsize=14&hidenavigation=1&theme=dark)

abbeyhrt:
Closing since the issue has been addressed.

PR #9741, opened 2021-09-24
fix(carbon-react): removes default props for accessible placeholders

### LINKED ISSUE #9466: Do not provide accessible labels by default, for example: "Provide icon description"

abbeyhrt:
Closes #9466
(Once #9731 is merged) 

Removes default props in v11 for accessible placeholder's and ensures that the props are required. 

Also makes `Tab`'s `id` prop required since the component has an accessibilty violation if it's not provided (reference #9666) 

#### Changelog

**Changed**

- iconDescription for `FileUploader` is now required and default is empty
- iconDescription's and invalidText's default prop for `NumberInput` are now empty
- default prop for label is now empty and `label` and `id` are required

#### Testing / Reviewing

Verify that everything works as expected


packages/react/src/components/FileUploader/FileUploader.js
```diff
@@ -97,7 +100,9 @@ export default class FileUploader extends React.Component {
   static contextType = PrefixContext;
 
   static defaultProps = {
-    iconDescription: 'Provide icon description',
+    iconDescription: FeatureFlags.enabled('enable-v11-release')
+      ? ''
```
joshblack:
```suggestion
      ? undefined
```

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -190,10 +191,14 @@ class NumberInput extends Component {
   static defaultProps = {
     disabled: false,
     hideLabel: false,
-    iconDescription: 'choose a number',
+    iconDescription: FeatureFlags.enabled('enable-v11-release')
+      ? ''
```
joshblack:
```suggestion
      ? undefined
```

joshblack (changes requested):
I think for the default props we'll want the value to be `undefined` versus an empty string.

For dropping isRequired, is it the case that all of these are no longer required? It seemed like we would still want them but want the user to provide them versus the default props being used. Let me know if I'm misunderstanding 👀

abbeyhrt:
> For dropping isRequired, is it the case that all of these are no longer required? It seemed like we would still want them but want the user to provide them versus the default props being used. Let me know if I'm misunderstanding 👀

I think you might have it backwards! If v11 is enabled, they are required, if not, they aren't

aledavila (approved):

packages/react/src/components/Tab/Tab.js
```diff
@@ -98,7 +103,7 @@ export default class Tab extends React.Component {
 
   static defaultProps = {
     role: 'presentation',
-    label: 'provide a label',
+    label: FeatureFlags.enabled('enable-v11-release') ? '' : 'provide a label',
```
joshblack:
```suggestion
    label: FeatureFlags.enabled('enable-v11-release') ? undefined : 'provide a label',
```

packages/react/src/components/Tab/Tab.js
```diff
@@ -44,7 +45,9 @@ export default class Tab extends React.Component {
     /**
      * The element ID for the top-level element.
      */
-    id: PropTypes.string,
+    id: FeatureFlags.enabled('enable-v11-release')
+      ? PropTypes.string.isRequired
+      : PropTypes.string,
```
joshblack: Do we auto-generate the id for the component if it doesn't exist already?

packages/react/src/components/SelectItemGroup/SelectItemGroup.js
```diff
@@ -55,7 +56,7 @@ SelectItemGroup.propTypes = {
 
 SelectItemGroup.defaultProps = {
   disabled: false,
-  label: 'Provide label',
+  label: FeatureFlags.enabled('enable-v11-release') ? '' : 'Provide label',
```
joshblack:
```suggestion
  label: FeatureFlags.enabled('enable-v11-release') ? undefined : 'Provide label',
```

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -190,10 +191,14 @@ class NumberInput extends Component {
   static defaultProps = {
     disabled: false,
     hideLabel: false,
-    iconDescription: 'choose a number',
+    iconDescription: FeatureFlags.enabled('enable-v11-release')
+      ? undefined
+      : 'choose a number',
```
joshblack, line 196: Should we just update `iconDescription` to `label` in a general sense for components that require labels?

packages/react/src/components/NumberInput/NumberInput.js
```diff
@@ -190,10 +191,14 @@ class NumberInput extends Component {
...
     step: 1,
     invalid: false,
-    invalidText: 'Provide invalidText',
+    invalidText: FeatureFlags.enabled('enable-v11-release')
+      ? ''
```
joshblack:
```suggestion
      ? undefined
```

joshblack (changes requested):

packages/react/src/components/NumberInput/NumberInput.js
abbeyhrt: I think a lot of the time, the iconDescription is a distinct thing from the label and depends component to component, so it's something we could talk about but I'm not sure if it should prevent this PR from being merged in

joshblack (approved):

PR #9749, opened 2021-09-27
fix(styles): update grid-gap to column-gap for inline dropdown

joshblack:
This came from a report in Slack. Currently, our inline dropdown uses `grid-gap` which adds both row and column spacing. This PR switches it to use `column-gap` so that only column spacing is used.

image

#### Changelog

**New**

**Changed**

- Update styles for dropdown so that only column spacing is used

**Removed**

#### Testing / Reviewing

- View the deploy preview and verify that the row spacing as shown in the screenshot above no longer exists


tay1orjones (approved):

jnm2377 (approved):

jnm2377 (changes requested):
ughh... wait... after inspecting, I'm realizing it doesn't actually change the spacing. If I add `column-gap: 0` specifically, it does.

Screen Shot 2021-09-29 at 11 17 10 AM
Screen Shot 2021-09-29 at 11 17 20 AM

packages/components/src/components/dropdown/_dropdown.scss
```diff
@@ -24,7 +24,7 @@
   .#{$prefix}--dropdown__wrapper--inline {
     display: inline-grid;
     align-items: center;
-    grid-gap: rem(24px);
+    grid-gap: 0 rem(24px);
```
jnm2377, line 27: I think this is backwards. Should it be?
```suggestion
    grid-gap: rem(24px) 0;
```

packages/styles/scss/components/dropdown/_dropdown.scss
```diff
@@ -25,7 +25,7 @@
   .#{$prefix}--dropdown__wrapper--inline {
     display: inline-grid;
     align-items: center;
-    grid-gap: rem(24px);
+    grid-gap: 0 rem(24px);
```
jnm2377, line 28: same as above

packages/components/src/components/dropdown/_dropdown.scss
joshblack: @jnm2377 I think the syntax for gap is `<row-gap> <column-gap>` and we are setting 0 for the row gap to remove the vertical spacing caused by the listbox styles. Let me know if I'm misunderstanding 👀

jnm2377:
wait... is the column spacing the vertical spacing or the horizontal spacing? 😭 Now I'm confused lol

joshblack:
@jnm2377 haha, totally get that. I believe that column is horizontal and row is vertical spacing.

packages/components/src/components/dropdown/_dropdown.scss
jnm2377: ah ok ok. I think I was misunderstanding which one we were removing! I thought we were removing column gap. Gotcha gotcha.

jnm2377 (approved):

PR #9755, opened 2021-09-27
chore(react): update storybook to accept process.env

joshblack:
This PR enables us to turn on feature flags in our default storybook by running the following command:

```bash
CARBON_ENABLE_V11_RELEASE=true yarn storybook
```

This is useful when working on components that have prop changes or are switching from class to functional components

#### Changelog

**New**

**Changed**

**Removed**

#### Testing / Reviewing

- Run storybook using `CARBON_ENABLE_V11_RELEASE=true yarn storybook`
- Pick a random component story and change it's entrypoint (`src/components/<component>/index.js) to use the following lines at the top:

```js
import * as FeatureFlag from '@carbon/feature-flags';

console.log(FeatureFlag.enabled('enable-v11-release'));
```

When looking at the storybook story, this should log out `true`


sstrubberg (approved):
Worked like a champ. Could we also throw a yarn command in there like `yarn storybook:v11` or something like that so we don't have to remember the `CARBON_ENABLE_V11_RELEASE=true yarn storybook` every time?

dakahn (approved):

PR #9765, opened 2021-09-29
fix(MultiSelect): fix phantom selection on MenuKeyDownEnter 👻

### LINKED ISSUE #9367: Undefined selection in `MultiSelect`
When interacting with the `MultiSelect` component, you can trigger an undefined selection by opening the control with enter or space, and then pressing enter or space again while focus is on the listbox.

The first item in the list should be treated as the active descendant. Currently, it appears that this item is not being rendered but is somehow being treated as the first item by downshift.

https://user-images.githubusercontent.com/3901764/127558300-69a85afd-0930-4992-9b65-0d74c5e93879.mov

dakahn:
Closes #9367 

Adds a conditional to the switch case on l193 that checks that an item was actually highlighted and selected. Otherwise it just calls `break`

#### Testing / Reviewing

In the MultiSelect storybook hi <kbd>tab</kbd> to focus the form element. Then hit <kbd>enter</kbd> to open the element. Hit <kbd>enter</kbd> again and *no selection should be made*


andreancardona:
@dakahn looks good to me after testing :)

Just one thing: https://react.carbondesignsystem.com/?path=/story/components-select--default

Is the expectation for default `Select` to open on enter or _just_ key down?

aledavila (approved):
lgtm. doesnt close or choose anything on enter

tay1orjones (approved):

PR #9777, opened 2021-09-30
fix(accordion): update arrow spacing for start variant

janhassel:
Currently, the chevrons in the "start"-variant of the Accordion (with the chevrons on the left)  have 8px of left spacing whereas the chevrons in the "end"-variant (with the chevrons on the right) have a right spacing of 16px.

After checking with @aagonzales, the left spacing on the "start"-variant should also be 16px / spacing-05.

Before
image

After
image

#### Changelog

**Changed**

- Updated left spacing of chevron in "start"-variant to use `$spacing-05`/16px.
- Added a `$spacing-05`/16px spacing to the right of the accordion title based on spec from @aagonzales:
  - image

#### Testing / Reviewing

1. Run react storybook / open netlify preview
2. Ensure that the chevrons always have 16px to their left and right regardless of variant
3. Ensure the title has a right spacing of 16px in the "start"-variant


tay1orjones (approved):
Looks good, thanks for adding it to the `styles` package as well!

jnm2377 (approved):
Looks great! Thanks for the fix!

jnm2377:
@aagonzales could you verify this is all up to spec? It's a visual bug fix for start chevron accordions.

 https://deploy-preview-9777--carbon-components-react.netlify.app/?path=/story/components-accordion--playgroundhttps://deploy-preview-9777--carbon-components-react.netlify.app/?path=/story/components-accordion--playground

aagonzales (approved):
Yeah it looks correct, if the padding area includes the 16px icon then it would be 32px/spacing-07 there instead.

image

Issue #9624, opened 2021-09-08
Deprecate usage of `iconDescription` for v11


joshblack:
We currently use the `iconDescription` prop in the following components: [Link](https://carbon-react-explorer.vercel.app/props/iconDescription)

We want `iconDescription` to no longer be needed. The accessible label for a component should come from either content or a direct `aria-label`, `aria-labelledby`, or `aria-describedby` prop.

- [x] Select
- [x] Button (should be deprecated with `IconButton`)
- [x] TimePickerSelect (I think this can be deprecated / removed)
- [x] RadioTile
- [x] SelectableTile
- [x] ~~SearchFilterButton~~ deprecated
- [ ] OverflowMenu (used for `aria-label` and `<title>`, should change to `label`)
- [ ] NumberInput (used for both increment and decrement buttons)
- [ ] NotificationButton (used for label on `button`, no tooltip)
- [ ] NotificationIcon (used for `<title>` in SVG)
- [ ] ToastNotification (passed to `NotificationButton` for close)
- [ ] InlineNotification (passed to `NotificationButton` for close)
- [ ] Modal (uses `iconDescription` for close label)
- [ ] InlineLoading (could use `label`, mismatch with `Loading` which uses `description`)
- [ ] FileUploaderItem (this gets passed to `Filename`)
- [ ] Filename (should this just be `label`?)
- [ ] FileUploader (this gets passed to `Filename`)
- [ ] DatePickerInput (is this needed for this input?)
- [ ] TableToolbarMenu (this gets passed to `OverflowMenu`)
- [ ] TableBatchAction (this gets passed to `Button`)
- [x] CopyButton (updated to use `label`)
- [ ] ModalHeader (used for close icon label)
- [x] AccordionItem

joshblack:
Deprecation warning should read something like:

> The `iconDescription` prop is no longer needed and can be safely removed. This prop will be removed in the next major release of Carbon.

OR

> The `iconDescription` prop has been deprecated, use the `soAndSo` prop instead. This prop will be removed in the next major release of Carbon.

dlavrenuek:
@joshblack Now that the `iconDescription` is gone from `ToastNotification` it is no longer possible to set a localized title for  the close button (`NotificationButton`). This does not seem to be the desired behavior, is there already a plan to have this title being set with other means?

PR #9787, opened 2021-10-01
chore(modal): added deprecation warning for iconDescription

sstrubberg:
REF #9624 

Added deprecation warnings for `iconDescription` for v11.

- `Title` and `aria-label` attributes now use the `ariaLabel` prop instead of `iconDescription`.
- The close button now gets an `aria-hidden="true"`

#### Testing

Review Modal in Storybook and ensure tests are firing off correctly.

#### Feedback

Unclear on why the tests are failing because of the deprecation warning. Feedback welcome.

### DIFF
--- a/packages/react/src/components/Modal/Modal.js
+++ b/packages/react/src/components/Modal/Modal.js
@@ -75,9 +75,12 @@ export default class Modal extends Component {
     hasScrollingContent: PropTypes.bool,
 
     /**
-     * Provide a description for "close" icon that can be read by screen readers
+     * The iconDescription prop is deprecated. Use aria-label, aria-labelledby, or aria-describedby to label your component.
      */
-    iconDescription: PropTypes.string,
+    iconDescription: deprecate(
+      PropTypes.string,
+      'The iconDescription prop is no longer needed and can be safely removed. This prop will be removed in the next major release of Carbon.'
+    ),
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Modal/Modal.js
```diff
@@ -217,7 +220,8 @@ export default class Modal extends Component {
     primaryButtonDisabled: false,
     onKeyDown: () => {},
     passiveModal: false,
-    iconDescription: 'Close',
+    // iconDescription is deprecated in v11 use aria-label instead
```
abbeyhrt, line 223: Out of curiosity, do we usually remove the default prop value even if the prop hasn't been deprecated yet?

packages/react/src/components/Modal/Modal.js
```diff
@@ -409,16 +413,20 @@ export default class Modal extends Component {
         Array.isArray(secondaryButtons) && secondaryButtons.length === 2,
     });
 
+    // ['aria-label']: ariaLabel,
```
abbeyhrt, line 416: Should this be commented out?

abbeyhrt:
Should `aria-label` for `Modal` now be required in V11?  Just thinking that it might be since these changes seem In the same vein as this issue from last sprint #9466

PR #9797, opened 2021-10-04
fix(Modal): fix Formik bug in Modal

### LINKED ISSUE #9523: [Modal]: Use of Modal with Formik makes secondary button require a double click
- [x] `carbon-components`

## Detailed description

We have recently been updating to the latest Carbon package versions, and it broke some of our integration tests. Our updates are as follows:
`carbon-components`: `10.23.2` => `10.40.2`
`carbon-components-react`: `7.23.2` => `7.40.2`

Upon investigation, we found that some of our Modals now require double clicking the secondary button before its event is triggered. Upon further testing, it seems to be related to the use of Formik in our application.

We have been able to determine that this behavior only exists in Carbon starting in versions `10.29.0` and `7.29.0` of `carbon-components` and `carbon-components-react`, respectively. [This PR](https://github.com/carbon-design-system/carbon/pull/7775/files#diff-22f235c4784785309ea4abec2fa146f38d2be0d56b42466af68a247a0fe898b3) from that release seems to be the most likely culprit. (I believe the only other change to Modal for this release occurred [here](https://github.com/carbon-design-system/carbon/pull/7690/files#diff-22f235c4784785309ea4abec2fa146f38d2be0d56b42466af68a247a0fe898b3), which was then overridden by https://github.com/carbon-design-system/carbon/pull/7775.)

I was able to create a simple CodeSandbox with barebones use of Formik which results in the behavior we're seeing: https://codesandbox.io/s/carbon-react-modal-formik-bug-2bho4

- Note that single clicking the Cancel secondary button does not result in the `onSecondarySubmit` function triggering. This function only triggers on the second or subsequent clicks. The first click also results in a console error which looks similar to the following: `Uncaught Error: Cannot remove node 11 because no matching node was found in the Store.`
- Note that the CodeSandbox uses versions `10.29.0` for `carbon-components` and `7.29.0` for `carbon-components-react`, as these are the first versions in which this bug is present. Note that the bug no longer occurs if these versions are adjusted to `10.28.0` and `7.28.0`, respectively. Also note that the bug still exists when using the latest versions of `carbon-components` and `carbon-components-react`.

JohnWColby:
Closes #9523 

Breaking the secondary buttons out into their own sub-component [(here)](https://github.com/carbon-design-system/carbon/commit/647ec1e75#diff-22f235c4784785309ea4abec2fa146f38d2be0d56b42466af68a247a0fe898b3R437) created the bug described in the above issue. I understand that this was probably done for readability / maintainability, but it also is causing additional re-renders when used with Formik that are resulting in incorrect behavior.

#### Changelog

**Changed**

- Directly build out the secondary button set rather than using a sub-component

**Removed**

- Remove SecondaryButtonSet sub-component

#### Testing / Reviewing
I tested this by linking my local version of Carbon to my local version of our application, but a more generic way to test would be to run [this CodeSandbox](https://codesandbox.io/s/carbon-react-modal-formik-bug-2bho4) locally, linked to a local version of Carbon on this branch.

### DIFF
--- a/packages/react/src/components/Modal/Modal.js
+++ b/packages/react/src/components/Modal/Modal.js
@@ -419,7 +419,8 @@ export default class Modal extends Component {
         onClick={onRequestClose}
         title={iconDescription}
         aria-label={iconDescription}
-        ref={this.button}>
+        ref={this.button}
+      >
         <Close20
           aria-label={iconDescription}
           className={`${this.modalCloseButtonClass}__icon`}
@@ -451,32 +452,6 @@ export default class Modal extends Component {
       alertDialogProps['aria-describedby'] = this.modalBodyId;
     }
 
-    const SecondaryButtonSet = () => {
-      if (Array.isArray(secondaryButtons) && secondaryButtons.length <= 2) {
-        return secondaryButtons.map(
-          ({ buttonText, onClick: onButtonClick }, i) => (
-            <Button
-              key={`${buttonText}-${i}`}
-              kind="secondary"
-              onClick={onButtonClick}>
-              {buttonText}
-            </Button>
-          )
-        );
-      }
-      if (secondaryButtonText) {
-        return (
-          <Button
-            kind="secondary"
-            onClick={onSecondaryButtonClick}
-            ref={this.secondaryButton}>
-            {secondaryButtonText}
-          </Button>
-        );
-      }
-      return null;
-    };
-
     const modalBody = (
       <div
         ref={this.innerModal}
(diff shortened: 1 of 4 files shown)


ColbyJohnIBM:

packages/react/src/components/Modal/Modal.js
```diff
@@ -416,7 +416,8 @@ export default class Modal extends Component {
         onClick={onRequestClose}
         title={iconDescription}
         aria-label={iconDescription}
-        ref={this.button}>
+        ref={this.button}
+      >
```
JohnWColby: These changes throughout this file seemed to be added via `husky` during the commit process-- hopefully they fit with this repo's conventions.

aledavila (changes requested):
Looks like this completely breaks modal as I'm not seeing in the preview at all.

ColbyJohnIBM:
Interesting, the preview excludes the following components that appear on https://react.carbondesignsystem.com/:
AspectRatio
ComposedModal
CopyButton
DataTable
ErrorBoundary
FormGroup
FormLabel
Grid
Link
Loading
Modal
OrderedList
TooltipDefinition
TooltipIcon
UnorderedList

It also includes a few components that don't exist there:
Layer
Popover
Theme

I'm not sure why this might be...

ColbyJohnIBM:
Is there a way to get netlify to run again with the latest? Or does it do that automatically?

tay1orjones:
Sorry for the confusion @ColbyJohnIBM, there are three different deploy previews built for each pull request:

1. `carbon-react-next` 
   - New storybook from a new v11 package `@carbon/react`, `./packages/carbon-react`, currently in beta.
2. `carbon-components-react` 
   - This is the one you'll want to validate against, it's the preview for the existing storybook for 
   `carbon-components-react`, `./packages/react` 
3. `carbon-elements` 
   - Preview specifically for the elements packages; `grid`, `layout`, `colors`, etc.

The Netlify comment body will indicate which link is for which preview.

These deploys will all automatically update/re-deploy to the same URL when new commits are pushed to the pull request.

jnm2377 (approved):

aledavila (approved):
works as expected

PR #9803, opened 2021-10-05
fix(grid): make subgrid work with grid offset

janhassel:
Ref https://github.com/carbon-design-system/carbon/discussions/9723#discussioncomment-1430432

Ensures that subgrid is supported corectly when the parent grid uses offset.

#### Changelog

**Removed**

- Removed explicit `col-end-n` rules since the existing `col-span-n` and `col-start-n` classes together are already achieving the expected styling

#### Testing / Reviewing

Add the test case from the discussion as a story and verify the appearance / behaviour:
```jsx
export const Test = () => {
  return (
    <Grid>
      <Column sm={4} md={{ span: 6, offset: 1 }} lg={{ span: 8, offset: 1 }}>
        <div style={{ background: 'gray' }}>Full</div>
        <Grid>
          <Column sm={2} md={3} lg={4}>
            <div style={{ background: 'gray' }}>Half</div>
          </Column>
          <Column sm={2} md={3} lg={4}>
            <div style={{ background: 'gray' }}>Half</div>
          </Column>
          <Column sm={2} md={3} lg={4}>
            <div style={{ background: 'gray' }}>Half</div>
          </Column>
          <Column sm={true} md={true} lg={true}>
            Auto col
          </Column>
        </Grid>
      </Column>
    </Grid>
  );
};
```


packages/grid/scss/modules/_css-grid.scss
```diff
@@ -254,10 +254,6 @@
   .#{$prefix}--col-start-#{$i} {
     grid-column-start: $i;
   }
-
-  .#{$prefix}--col-end-#{$i} {
-    grid-column-end: $i;
-  }
```
tay1orjones, line 260: Do you think there's some value in having this as an option available to devs when they're composing columns with these bare classes? 

I know there's some functional duplication if we leave these in, but I wonder if some may prefer the mental model of _"`start` on column x and `end` at column x"_, while others may prefer _"`start` on column x, then `span` x columns"_

tay1orjones (approved):
I have one question on my review below, but both your example story and the existing offset story look great. 

https://user-images.githubusercontent.com/3360588/136612968-4def9805-a440-40f2-8c46-ee6539452753.mov

I pushed one small update to this - I noticed the existing stories' styles weren't being applied due to the prefix change.

packages/grid/scss/modules/_css-grid.scss
janhassel: @tay1orjones That's a very good point which I actually discussed with @ConradSchmidt a bit. In the end, I decided against it mainly for the following two reasons:

**1. CSS size**
In order for both approaches to work we'd need to emit the correct value for `--cds-grid-columns` on all possible combinations of `col-start` with `col-end`.

For example to accommodate a scenario where on `lg` breakpoint a container should start at the 5th column and end at the 9th, we'd need the following css:
```css
.cds--lg:col-start-5.cds--lg:col-end-9 {
  --cds-grid-columns: 4;
}
```

Even though this is not hard to accomplish thanks to SCSS loops, it would result in a lot of additionally generated css with all its implications.

**2. API Consistency**
CSS allows us to write it either way. We can use `grid-column-start` and `grid-column-end`, a combination (`grid-column-start` and `grid-span`) or even a short version `grid-span: n / n`. In the `Column` helper components (at least in React and Vue, haven't checked the others), we are bound to the API defined as `{ span: n, offset: n }`. Reusing this structure could help developers who use the helper components and css classes interchangeably by providing a more consistent developer experience.

---

I'm open to change this though and implement the solution outlined in the first point if you feel strongly about this. Let me know what you think! 🙂

tay1orjones: > in React and Vue ... we are bound to the API defined as { span: n, offset: n }. Reusing this structure could help developers who use the helper components and css classes interchangeably by providing a more consistent developer experience.

This really resonates! Reduced CSS size is definitely a benefit, too. Removing `col-end` sounds good 👍

dakahn (approved):
lgtm

PR #9817, opened 2021-10-07
docs(text): add component examples to storybook, update switch text prop

tay1orjones:
Minor changes stemming from https://github.com/carbon-design-system/carbon/pull/9493#issuecomment-937835496

#### Changelog

**New**

- Text: Add more `Text` usage examples to the storybook

**Changed**

- Switch: update `text` prop to accept `node`, so `Text` can be used for bidi text

#### Testing / Reviewing

- carbon-components-react storybook should show a new `Component Examples` story with a few component examples stemming from the conversation
- The ContentSwitcher/Switch should not throw a prop-types warning


dakahn (approved):

aledavila (approved):
lgtm

packages/react/src/components/Switch/Switch.js
```diff
@@ -107,7 +107,7 @@ Switch.propTypes = {
   /**
    * Provide the contents of your Switch
    */
-  text: PropTypes.string.isRequired,
+  text: PropTypes.node.isRequired,
```
joshblack: For consistency, would it help if this component to accept a `children` prop for the generalized `PropTypes.node` use-case? Having `text` not be `PropTypes.string` can be confusing sometimes.

tay1orjones: @joshblack Yeah I dig that idea. ~I'll deprecate `text`, too.~ Actually it seems like there could be a benefit to keeping text around - it has styles associated with it to enable truncation.

packages/react/src/components/Switch/Switch.js
```diff
@@ -58,6 +59,7 @@ const Switch = React.forwardRef(function Switch(props, tabRef) {
       {...commonProps}>
       <span className={`${prefix}--content-switcher__label`} title={text}>
         {text}
+        {children}
```
joshblack: Just thinking out loud, does it make sense to do `text !== undefined ? text : children` so that we swap between them instead of rendering both?

tay1orjones: @joshblack Yeah for sure. I just updated it

joshblack (approved):

PR #9827, opened 2021-10-08
feat(checkbox): add onClick prop

### LINKED ISSUE #9577: Not able to prevent click event propagation on Checkbox/TableSelectRow

- [ ] `carbon-components`

## Detailed description

Discover while we implement the 'onRowClick' on the table rows with `TableSelectRow`. I don't see a way to completely prevent the click from propagating from the Checkbox or the TableSelectRow because the event seems to be from the `label`
https://codesandbox.io/s/sparkling-sky-2d3uj?file=/src/index.js

Checkbox
TableSelectRow

> see changed?

`Checkbox.onClick` or `TableSelectRow.onSelect` should prevent the click from bubbleing up.

Chrome

Latest

> should be aware of?

CP4D - WKC

1. Step one
2. Step two
3. Step three
4. etc.

https://codesandbox.io/s/sparkling-sky-2d3uj?file=/src/index.js

tay1orjones:
Closes #9577 

Adds an `onClick` handler to the `div` wrapping the checkbox `input` and its `span` label. 

Make note of the code comments I added. I'm curious if there's any additional perspectives on how to solve this issue.

#### Changelog

**New**

- checkbox: add `onClick` handler

#### Testing / Reviewing

- Open the carbon-components-react storybook
- View the Checkbox Playground story
- Click a checkbox, 3 events should fire: 
   1. `onChange`
   2. `onClick` with `event.target` indicating the `span`
   3. `onClick` with `event.target` indicating the `input`
- Click a checkbox _label_, the same set of 3 events should fire


packages/react/src/components/Checkbox/Checkbox.js
```diff
@@ -44,7 +45,15 @@ const Checkbox = React.forwardRef(function Checkbox(
...
-    <div className={wrapperClasses}>
+    /* 
+      The a11y rules below are disabled because <div> element has checkbox 
+      input and label elements as children that allows keyboard interaction. 
+      The onClick handler facilitates consumers' ability to stop click events 
+      from bubbling beyond the checkbox without having to implement their own 
+      wrapper element around <Checkbox>.
+    */
+    /* eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
+    <div className={wrapperClasses} onClick={(evt) => onClick(evt)}>
```
tay1orjones, line 56: @dakahn Is this okay from your perspective? It seems so based on [this note on the first rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md#case-the-event-handler-is-only-being-used-to-capture-bubbled-events):
>If your element is catching bubbled click or key events from descendant elements, there are no appropriate roles for your element: you will have to deactivate the rule.

Do we need a `keyPress` also as indicated from [this note on the second rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/click-events-have-key-events.md)?

abbeyhrt (approved):
@tay1orjones Looks good to me! I'm only seeing `onChange` being called in the Playground but that seems like the same problem as the other PR adding an `onClick` recently, where the new event showed up locally but not in the deploy preview.

dakahn (approved):

packages/react/src/components/Checkbox/Checkbox.js
tay1orjones: We looked into this together and we're good as is for now 👍

PR #9828, opened 2021-10-08
feat(modal): deprecate iconDescription

sstrubberg:
REF #9624 

Added deprecation warnings for `iconDescription` for v11.

#### Changelog

- Removed iconDescription from Storybook
- `title` and `aria-label` attributes now use the `ariaLabel` prop instead of `iconDescription`.
- The close button now gets an `aria-hidden="true"`
- Removed tests involving `iconDescription`. Added a new test to ensure the `aria-hidden` is applied to the `Close20` icon.

#### Testing / Reviewing

Review Modal in Storybook and ensure tests are firing off correctly.

#### Questions

- Should the `aria-label` be required now? REF [9466](https://github.com/carbon-design-system/carbon/issues/9466) - thx @abbeyhrt


packages/react/src/components/Modal/Modal-test.js
```diff
@@ -47,19 +47,9 @@ describe('Modal', () => {
...
-    it('adds new iconDescription when passed via props', () => {
-      mounted.setProps({ iconDescription: 'new description' });
-      expect(mounted.props().iconDescription).toEqual('new description');
-    });
-
-    it('should have iconDescription match Icon component description prop', () => {
-      const description = mounted.find(Close20).props()['aria-label'];
-      const matches = mounted.props().iconDescription === description;
-      expect(matches).toEqual(true);
+    it('should have aria-hidden on the Icon component', () => {
```
dakahn:
```suggestion
    it('should not place the svg icon in the accessibility tree', () => {
```

dakahn: just rewriting this one a bit to keep it in the perspective of a user using the component. A little background on what I mean by accessibility tree
https://developers.google.com/web/fundamentals/accessibility/semantics-builtin/the-accessibility-tree

packages/react/src/components/Modal/Modal-test.js
```diff
@@ -47,19 +47,9 @@ describe('Modal', () => {
...
+      const ariaHidden = mounted.find(Close20).props()['aria-hidden'];
+      expect(ariaHidden).toEqual('true');
```
dakahn, line 52: it could also be helpful to test that the SVG is not a tab stop, but we may do this elsewhere

dakahn (changes requested):

dakahn:
glorious work 
image

dakahn (approved):
looks good to me 🏄🏾

aledavila (approved):
lgtm just looks like you need to update snapshots

Issue #9805, opened 2021-10-05
Audit and consolidate styles between `carbon-components` and `@carbon/styles`


tay1orjones:
Leading up to v11 we have migrated styles from `carbon-components` over into `@carbon/styles` and in the process converted the styles to be configured as sass modules.

Since then, there have been bug fixes and new features made to `carbon-components` changing the styles. These changes need to be found and ensure they have been ported over to `@carbon/styles`. In addition there have been purposeful changes made in `@carbon/styles` for v11 that are breaking changes and should not be ported back to `carbon-components`.

Moving forward:
 - Changes to `carbon-components` should also be made to `@carbon/styles`, or tracked here to be migrated to `@carbon/styles` at a later date during a final audit to take place closer to v11 stable release.
 - Purposeful (breaking) changes made only to `@carbon/styles` for v11 can be noted on this issue in attempt to ensure these changes are retained during the final audit.

### Steps to take for each component are outlined in the [**comment below ↓**](https://github.com/carbon-design-system/carbon/issues/9805#issuecomment-1026166086)

 **Checklist**
 
 - [x] #10481
- [x] #10482
- [x] #10483
- [x] #10484
- [x] #10567
- [x] #10568
- [x] #10569
- [x] #10570
- [x] #10571
- [x] #10572
- [x] #10573
- [x] #10574
- [x] #10575
- [x] #10576
- [x] #10577
- [x] #10578
- [x] #10579
- [x] #10580
- [x] #10581
- [x] #10582
- [x] #10583
- [x] #10585
- [x] #10586
- [x] #10584
- [x] #10587
- [x] #10630
- [x] #10631
- [x] #10632
- [x] #10633
- [x] #10660
- [x] #10661
- [x] #10675
- [x] #10718
- [x] #10719
- [x] #10720
- [x] #10721
- [x] #10722
- [x] #10723
- [x] #10724
- [x] #10725
- [x] #10726
- [x] #10727
- [x] #10728
- [x] #10729
- [x] #10730
- [x] #10731
- [x] #10732

tay1orjones:
It appears the progress Indicator styles are incorrect in v11

tay1orjones:
**Steps for each component**
1. Find differences between v10/v11 components
   - Diff the two style files, identify any meaningful differences other than tokens/sass modules syntax
   - Visually review the component in existing storybook and new storybook
2. Identify if the change should be present in v10, v11, or both.
   - Look at git history of those changes
[...]

PR #9830, opened 2021-10-08
Remove external/outside margin from components

### LINKED ISSUE #5367: Clarify default component spacings

tay1orjones:
Closes #5367
Refs #9805 

Removes external/outside "[harmful margin](https://mxstbr.com/thoughts/margin/)" from components listed in the original issue. The changes are only within `@carbon/styles` because these are considered breaking changes and should only be used with v11. 

#### Changelog

**Removed**

- form: remove fieldset, fileuploader, label margin
- notifications: remove external margin from inline, toast, and actionable notification
- structured-list: remove external margin

#### Testing / Reviewing

- Existing styles for these components should remain unchanged in the `carbon-components-react` storybook
- Styles for these components should be updated within the `carbon-react-next` storybook


tay1orjones:
A few components listed in the original issue have already had margin removed or were fixed to no longer include margin:
- InlineLoading
- Loading
- Toggle
- SmallToggle

cc @janhassel

vpicone:
Do we need to document these somewhere? Could see it being a surprise under some circumstances. re: feature flag, I think that's a design decision we have to make (or maybe have already made? @joshblack). Is the goal for folks to be able to switch from `carbon-components` to `@carbon/styles` as a drop-in change without needing to modify their code?

I could see the benefit of that as far as the adoption of the new package is concerned.

dakahn (approved):
Looks good to me 🏄🏾

tay1orjones:
Update on this - the storybooks should both remain unchanged as they both have the v11 feature flag turned off. To validate the margin removal, you have to pull down locally and add turn on the v11 feature flag.

sstrubberg (approved):

PR #9833, opened 2021-10-11
chore(react): use babel runtime on es builds

salvoravida:
chore(react): use babel runtime on es builds

Hi guys!

This PR adds @babel/runtime on es builds. -> less **60/100k** on your bundle.js depended on how many components are used.


salvoravida:
recheck

salvoravida:
> @salvoravida Thanks for doing this! I think we'll also need to include `@babel/runtime` as a dependency, potentially? 👀
It is already included by babel. not needed.

joshblack (approved):

tay1orjones (approved):

tay1orjones (changes requested):
Sorry - actually needs a `yarn` to update the lockfile. I'll pull this down and do it.

tay1orjones (approved):
Done 👍  LGTM if status checks pass

Issue #9763, opened 2021-09-28
Update `light` prop usage with `Layer` component


joshblack:
We currently use the `light `prop in situations where a component may live on different background colors (namely if it sits on `ui-background` and `field-01`.

However, as the number of layers grows (up to 4) we end up alternating using `light` or not in light themes. In dark themes, however, this striping is not possible as the colors should lighten.

In v11, we will use the `Layer` component to accomplish this. This issue is to setup the following components which have a `light` prop to work with the `Layer` component and deprecate usage of the `light` prop for v11. Specifically:

- In v11, the `light` prop will be deprecated (but not in v10)
- There should be stories for these components around how to use it with the `Layer` component

Below is the list of components impacted:

- [x] Select	
- [x] ListBox
- [x] TimePicker
- [x] RadioTile
- [x] Tile
- [x] ClickableTile
- [x] SelectableTile
- [x] ExpandableTile
- [x] TextInput
- [x] ControlledPasswordInput
- [x] PasswordInput
- [x] TextArea
- [x] Tabs
- [x] Slider
- [x] Search
- [x] OverflowMenu
- [x] NumberInput
- [x] FilterableMultiSelect
- [x] MultiSelect
- [x] Dropdown
- [x] DatePicker
- [x] ContentSwitcher
- [x] ComboBox
- [x] CodeSnippet

tay1orjones:
@aledavila @abbeyhrt is there anything left for this one?

PR #9838, opened 2021-10-11
feat(react,carbon-react): add layer stories

abbeyhrt:
Partially addresses #9763

This PR adds examples of components using the new `Layer` component instead of `light` prop. It also deprecates the `light` prop in the relevant components. 

#### Changelog

**New**
- `with Layer` story for various components 

**Changed**

- {{change thing}}

**Removed**

- deprecates `light` prop

#### Testing / Reviewing

Check the new stories out and make sure everything looks okay.


packages/react/src/components/ListBox/ListBox.js
```diff
@@ -112,9 +113,12 @@ ListBox.propTypes = {
...
 
   /**
-   * Specify if the control should use the light variant
+   * The light prop functionality is being replaced with the Layer component that will be available in Carbon V11.ß
    */
-  light: PropTypes.bool,
+  light: deprecate(
+    PropTypes.bool,
+    'The `light` prop has been deprecated and will be removed in the next major release in favor of the Layer component'
+  ),
```
tay1orjones: Should these be deprecated in v10? I interpreted the [initial issue](https://github.com/carbon-design-system/carbon/issues/9763#issue-1010063714) to mean that v10 `light` would not be deprecated, v11 `light` would be deprecated, v12 `light` would be removed.

If so, we'll have to wrap these deprecation warnings in a feature flag.

tay1orjones: I'm not sure if it would make more sense to write a custom validator that uses `deprecate` or extend `deprecate` to take a new feature flag argument. Maybe a new `deprecateBehindFlag(flag, proptype, deprecation)` that uses `deprecate`?

abbeyhrt: Ah I think you're right. The language around deprecation is so confusing to me. For me, the issue meant that we warn in v10 that `light` will be deprecated and then in the v11, it's actually removed(deprecated). I'll update the PR to remove the warnings. 

Separately, I think extending deprecate to take in a feature flag argument is a great idea, it's something we'll probably do more in the coming days and it would be great to have it baked into `deprecate`

dakahn (approved):
Pulled these down and ran em. These seem good to me barring and changes requested above

aledavila (approved):
Lgtm

PR #9850, opened 2021-10-12
fix(repo): add check around math.div usage

### LINKED ISSUE #9851: [Repo]: usage of `math.div` breaks some build tools that do not have the ability to upgrade sass.
### Package

### Browser

Chrome

### Package version

v10.41.0

### Description

The addition of math.div added in 10.41.0 do not check for the existence of the function, and causes some build tools to break. This breaks backwards compatibility of some libraries, because the sass version cannot be changed by the user.

For example, in `ai-apps/angular` and `carbon-addons-iot-react` the angular library is trying to maintain parity with react by using the react styles; however, angular v11 only support dart-sass v1.32.x and `math.div` was added in 1.33. These math.div updates prevent the `ai-apps/angular` package from re-using styles from carbon and still maintain backwards compatibility with angular v11.

### CodeSandbox example

n/a

### Steps to reproduce

You can see an example of this in the `carbon-addons-iot-react` build logs from https://app.netlify.com/sites/ai-apps-pal-angular/deploys/615df5061dc8800008ef3170.

```
3:19:49 PM: ERR! => Failed to build the preview
3:19:49 PM: ERR! ./src/side-panel/side-panel.scss
3:19:49 PM: ERR! Module build failed (from /opt/build/repo/node_modules/sass-loader/dist/cjs.js):
3:19:49 PM: ERR! SassError: Undefined function.
3:19:49 PM: ERR!    ╷
3:19:49 PM: ERR! 41 │   @return math.div($px, $carbon--base-font-size) * 1rem;
3:19:49 PM: ERR!    │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
3:19:49 PM: ERR!    ╵
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/layout/_convert.import.scss 41:11  carbon--rem()
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/layout/_breakpoint.scss 16:23      @import
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/type/_styles.import.scss 23:9      @import
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/type/_reset.scss 10:9              @import
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/_css--reset.scss 9:9                                            @import
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/_helper-mixins.scss 23:9                                        @import
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/vendor/carbon-components/scss/globals/scss/_mixins.scss 8:9                                                @import
3:19:49 PM: ERR!   src/vendor/@ai-apps/styles/scss/globals/_vars.scss 2:9                                                                                     @import
3:19:49 PM: ERR!   src/side-panel/side-panel.scss 1:9                                                                                                         root stylesheet
```

### Code of Conduct

kevinsperrine:
Closes #9851

Some build tools have older versions of sass that does not include math.div. This adds a wrapper around math.div to maintain backwards compatibility.

#### Changelog

**New**

- adds checks around math.div usage to confirm the function exists

**Changed**

- adds checks around math.div usage to confirm the function exists and uses it if so, or falls back to `/`, if not. 
- sometimes uses `* 0.5` when previously was dividing by two

**Removed**

- n/a

#### Testing / Reviewing

confirmed existing tests pass. There shouldn't be any usages of `math.div` that are not wrapped in `@if meta.function-exists('div', 'math') {`


tay1orjones (commented):
I double checked to make sure all usages are wrapped in the conditional.

A huge caveat here is that anytime `math.div` is used moving forward, we'll have to actively remember to wrap it in this conditional. It's likely to get missed and break the build for consumers using sass <1.33.0. 

My first thought is to tackle this with linting. The `stylelint-scss` plugin we use for stylelint unfortunately rejected the [proposal](https://github.com/kristerkari/stylelint-scss/issues/528) of a rule for no-division. 

This conditional could be abstracted out into its own function to effectively replace `math.div`, but direct usages of `math.div` would still need to be disallowed in favor of using the abstraction. I think this could be done utilizing Stylelint's [`function-disallowed-list`](https://stylelint.io/user-guide/rules/list/function-disallowed-list) rule.

tay1orjones (approved):
@kevinsperrine I think since we're using math.div across various different packages we'll need to make a new internal package in the monorepo to share it between all the other packages. No need to do that here though - I think let's merge this as-is and we can tackle a the abstraction through a new issue.

abbeyhrt (approved):

Issue #6477, opened 2020-07-15
ListBox: need a way to disable a single listbox item in a dropdown, combobox, or multiselect


Donisius:
Use this template if you want to request a new feature, or a change to an
existing feature.

If you are reporting a bug or problem, please use the bug template instead.

### Summary

Please describe your request in one or two sentences.

We need a way to disable a single listbox item from a dropdown or combobox.

### Justification

On `carbon-components-angular` it is possible to single disable dropdown items but there are no styles to visually indicate that the item is disabled.

### Desired UX and success metrics

A dropdown list item should be able to be disabled, and have a visual indication that it is disabled.

_Carbon is a collaborative system. We encourage teams to build components and
submit them for integration as either add-ons or core components._

PR #9859, opened 2021-10-13
feat(multiselect): allow disabled listbox items

### LINKED ISSUE #9610: [Feature Request]: To disable the options in MultiSelect
### Summary

We would like to disable the option(s) in MultiSelect because in the beginning, the options is valid, but afterward, it becomes invalid. But we don't want to remove the invalid options directly because of user experience concerns. 

### Justification

As above.

We want something like below.
image

### Required functionality

_No response_

As soon as possible. Thanks.

### Available extra resources

_No response_

### Code of Conduct

tay1orjones:
Closes #9610 
Refs #6477 

Modifies MultiSelect to pass through the `disabled` option on items to downshift. Adds new styles to properly display disabled items.

#### Changelog

**New**

- multiselect: added listbox disabled item styles

**Changed**

- multiselect: enable `disabled` option for listbox items

#### Testing / Reviewing

- View the `carbon-components-react` storybook default MultiSelect story. The third option in the listbox should have the `disabled` attribute and appear disabled.
- View the `carbon-react-next` storybook default MultiSelect story. The third option in the listbox should have the `disabled` attribute and appear disabled.
- Would be awesome to get a visual review to make sure the disabled styles I added are good


tay1orjones:
A few notes regarding the accessibility of this: 

- The `disabled: true` is passed through to Downshift. Downshift applies a `disabled` attribute to the element with `role="option"`
- The `option` role [inherits `aria-disabled`](https://www.w3.org/TR/2010/WD-wai-aria-20100916/roles#option), which to my understanding can be used interchangeably with the `disabled` attribute.
- Disabled options do not receive focus and are skipped when using keyboard navigation. From the [`aria-disabled` spec](https://www.w3.org/TR/2010/WD-wai-aria-20100916/states_and_properties#aria-disabled):
  - > (aria-disabled) Indicates that the element is perceivable but disabled, so it is not editable or otherwise operable ... Disabled elements might not receive focus from the tab order. For some disabled elements, applications might choose not to support navigation to descendants.

Overall I think we're within bounds accessibility-wise. @dakahn @mbgower definitely would appreciate any critique you might have of my thinking here!

dakahn:
Yep downshift does its job when it comes to disabled. Nice! I'd actually recommend `aria-disabled` over using `disabled` in general because the latter can cause a confusing experience potentially.

dakahn (approved):
Totally agree with keeping the disabled item focusable -- good call there. Looks good 👍🏾

laurenmrice (changes requested):
The disabled text and icon should be color token $disabled-02, not -03.

laurenmrice (approved):
lgtm!

PR #9869, opened 2021-10-14
feat(ContentSwitcher): add icon only ContentSwitcher variant

### LINKED ISSUE #9843: ContentSwitcher: icon only variant
For certain use cases and context an icon only variant of the ContentSwitcher is required
image

dakahn:
Closes #9843 

adds an icon-only variant of ContentSwitcher

#### Changelog

**Changed**

- adds `hasIconOnly` prop and `aria-label` requirement along with it with ContentSwitcher.js
- adds icon-only story to ContentSwitcher storybook 
- adds styling for `#{$prefix}--content-switcher--icon-only` in ContentSwitcher.scss

#### Testing / Reviewing

- [ ] WIP ensure added and existing tests are passing
- [ ] ensure icon-only content switcher meets design specification
- [ ] ensure changes don't cause styling regressions (or others) in existing Contentswitcher

### DIFF
--- a/packages/react/src/components/ContentSwitcher/ContentSwitcher.js
+++ b/packages/react/src/components/ContentSwitcher/ContentSwitcher.js
@@ -35,6 +35,11 @@ export default class ContentSwitcher extends React.Component {
     className: PropTypes.string,
 
     /**
+     * Specify if the button is an icon-only button
+     */
+    hasIconOnly: PropTypes.bool,
+
+    /**
      * `true` to use the light variant.
      */
     light: deprecate(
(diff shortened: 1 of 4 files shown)


kingtraceyj (changes requested):
i should have sent these specs sooner. container size should be equal width and height. in this instance should be 40X40px

**as is:**
image

**should be:**
image

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -34,6 +34,11 @@ export default class ContentSwitcher extends React.Component {
      */
     className: PropTypes.string,
 
+    /**
+     * Specify if the button is an icon-only button
+     */
+    hasIconOnly: PropTypes.bool,
```
joshblack, line 40: Is there a way to do `hasIconOnly` by just letting them render icons as children of `Switch`? Wasn't sure how different the CSS is but would nice to not have to have this as a prop

joshblack:
@dakahn does the icon switcher require different a11y considerations than the text-based one?

PR #9875, opened 2021-10-15
fix(tooltip): prevent escape keydown from bubbling

### LINKED ISSUE #8502: [Tooltip]: escape on button in tooltip propagates

tay1orjones:
Closes #8502 

Prevents the `escape` from bubbling through `onKeyDown` on `Tooltip`. Includes a new test to cover this functionality. I had to more thoroughly mock the debounce to include the `cancel()` Tooltip uses.

#### Changelog

**Changed**

- tooltip: prevent escape keydown from bubbling

#### Testing / Reviewing

- The default bottom tooltip story has been modified to include a handler mimicing the bug reproduction from the original issue. It should _not_ get fired. Kind of silly, but we're explicitly trying to prevent something from happening.
- If you pull down locally and revert my changes to Tooltip, the handler in storybook will incorrectly fire reproducing the initial problem.
- ~⚠️  The story needs reverted before this is merged⚠️~ Edit: removed ✅

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.js
+++ b/packages/react/src/components/Tooltip/Tooltip.js
@@ -703,8 +703,16 @@ class Tooltip extends Component {
               this._tooltipEl = node;
             }}
             updateOrientation={this.updateOrientation}>
+            {/* This rule is disabled becuase the onKeyDown event handler is only
+             * being used to capture and prevent the event from bubbling: */}
+            {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
             <div
               className={tooltipClasses}
+              onKeyDown={(event) => {
+                if (keyDownMatch(event, [keys.Escape])) {
+                  event.stopPropagation();
+                }
+              }}
               {...other}
               id={this._tooltipId}
               data-floating-menu-direction={storedDirection}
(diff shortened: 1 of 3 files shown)


abbeyhrt (approved):
Woo! Looks good to me 🎉

packages/react/src/components/Tooltip/Tooltip.js
```diff
@@ -703,8 +703,16 @@ class Tooltip extends Component {
               this._tooltipEl = node;
             }}
             updateOrientation={this.updateOrientation}>
+            {/* This rule is disabled becuase the onKeyDown event handler is only
```
sstrubberg:
```suggestion
            {/* This rule is disabled because the onKeyDown event handler is only
```

sstrubberg (approved):

PR #9876, opened 2021-10-15
feat(react): add unstable_Stack component

joshblack:
Closes #9821

This PR adds in a `Stack` component to be used for vertical spacing along with other components in the design system. This is a useful utility in a component-based model when margin is no longer used by components.

By default, the `Stack` uses a `gap` prop where you can provide a number or a string. If it's a number, it will map to our spacing scale. If it's a string, it will be a custom value and requires a unit.

#### Changelog

**New**

- Add Stack component, related stories, and tests

**Changed**

- Update storybook to include Stack in `carbon-react`

**Removed**

#### Testing / Reviewing

- Verify CSS is included as expected
- Verify story renders as expected
- Look over the API and make sure it meets our API requirements / is understandable


packages/react/src/components/Stack/index.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export { default as Stack } from './Stack';
```
tay1orjones: Does this need to be re-exported from `carbon-react`?

packages/react/src/components/Stack/Stack.js
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * Specify the orientation of them items in the Stack
+   */
+  orientation: PropTypes.oneOf(['horizontal', 'vertical']),
+
+  /**
+   * Provide the step from the spacing scale to be used as the gap for the
+   * items rendered by the Stack
+   */
+  step: PropTypes.oneOf(
```
tay1orjones: My first thought was that I'd prefer this to be named `spacing` instead of `step` , but after thinking on it more I think I prefer `step` as it's a _step_ in the spacing _scale_.

I also would be open to removing it in favor or merging this in with `gap`. So then `gap` could be a string, number, or one of these values.

tay1orjones (commented):
I'm super into this implementation! I think the usage of grid gap is a fantastic idea.

I'm happy to knock out some tests for it once any other feedback is incorporated

packages/react/src/components/Stack/Stack.js
joshblack: @tay1orjones that makes a lot of sense to me, I was trying to figure out how to do this without having conflicts. In other words, did you mean 12px or the 12th step in the spacing scale 🤔

tay1orjones: I meant 12th step in the spacing scale, but in looking at it again I guess combining it all into `gap` would be sort of hefty... and we'd actually need to use 12px, aka `step.value`?

```js
gap: PropTypes.oneOfType([
  PropTypes.string, 
  PropTypes.number, 
  PropTypes.oneOf(Array.from({ length: SPACING_STEPS }).map((_, step) => {
      return step.value;
    })
  )
]),
```

Alternatively we could leave `gap` as `string || number` proptype and then in the _usage_ docs in storybook show how to import a step from `@carbon/layout` and use the value for `gap`. Side benefit here would be that `Stack` wouldn't have any direct dependence on `@carbon/layout`. Would reduce the scss a bit too since we wouldn't need the `${prefix}--stack-scale-${step}` classes

```js
import { spacing02 } from '@carbon/layout';
...
<Stack gap={spacing02} /> 
```

tay1orjones (approved):
Newest changes feel great! 🔥  Consolidating down to just `gap` and also very much in support of the `VStack` and `HStack` exports

packages/react/src/components/Stack/index.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import React from 'react';
+import { Stack } from './Stack';
+
+const HStack = React.forwardRef(function HStack(props, ref) {
+  return <Stack {...props} ref={ref} orientation="horizontal" />;
+});
+
+const VStack = React.forwardRef(function HStack(props, ref) {
```
joshblack:
```suggestion
const VStack = React.forwardRef(function VStack(props, ref) {
```

dakahn (approved):
Nice work!

PR #9885, opened 2021-10-18
feat(react): upgrade ExpandableTile from class to function

abbeyhrt:
Partially addresses #9712 

Updates ExpandableTile to functional component and updates the tests since they relied on the component being a class. 

#### Changelog

**Changed**

- tests
- ExpandableTile class -> function

#### Testing / Reviewing

Check that ExpandableTile works the same as it currently does


packages/react/src/components/Tile/next/Tile.js
```diff
@@ -398,307 +392,267 @@ SelectableTile.propTypes = {
...
+  /**
+   * An ID that can be provided to aria-labelledby
+   */
+  id: PropTypes.string,
+
+  /**
+   * `true` to use the light version. For use on $ui-01 backgrounds only.
+   * Don't use this to make tile background color same as container background color.
+   */
+  light: PropTypes.bool,
```
sstrubberg: the light prop needs to be deprecated

packages/react/src/components/Tile/next/Tile.js
```diff
@@ -398,307 +392,267 @@ SelectableTile.propTypes = {
...
+TileAboveTheFoldContent.displayName = 'TileAboveTheFoldContent';
 
-    return (
-      <span ref={ref} className={`${prefix}--tile-content__below-the-fold`}>
-        {children}
-      </span>
-    );
-  }
-);
+export function TileBelowTheFoldContent({ children }) {
```
sstrubberg: Why rip out the `React.forwardRef()`?

abbeyhrt: Not intentional! I accidentally was working on a branch without your changes merged in and didn't catch this when I merged in main, good catch! Thanks!

sstrubberg (changes requested):
Looks great! How did you know to dig up that useIsomorphicEffect hook?!

I think all you need is to replace your function with the `React.fowardRef()`.

tay1orjones (approved):

sstrubberg (approved):

Issue #7772, opened 2021-02-10
React - How to dynamically disable a single item inside of Carbon Dropdown component


dhebert-dev:
Hello. I am trying to dynamically disable a single item inside of the Carbon Dropdown component in React. What would be the best approach? I have noticed in the carbon storybook that there is a disable property but it seems to only disable the entire dropdown. I have also tried putting the items inside of an array, giving the item I want disabled the disable=true property and then passing it to the Dropdown component but that did not work. Thanks for your help!

tw15egan:
Hey there @dhebert-dev, it doesn't seem like disabling an individual item is currently supported (Rel https://github.com/carbon-design-system/carbon/issues/6477). As a workaround, you can use the `Select` element, which allows disabling a specific `SelectItem`

dakahn:
I'm going to convert this to an issue since this might be helpful in the future.

PR #9913, opened 2021-10-20
feat(listbox): support disabled items in combobox, dropdown, filterablemultiselect

### LINKED ISSUE #6477: ListBox: need a way to disable a single listbox item in a dropdown, combobox, or multiselect

tay1orjones:
Closes #6477

Refs #9859, #9610, #7772, #7780

Adds support for disabling individual items inside `ComboBox`, `Dropdown`, and `FilterableMultiSelect`.

#### Changelog

**Changed**

- combobox, dropdown, filterablemultiselect: enable disabled items prop through downshift, update stories/docs/styles

#### Testing / Reviewing

View each of the stories for these components. They have been updated to have a single disabled item for testing purposes.


packages/react/src/components/Dropdown/Dropdown.js
```diff
@@ -181,7 +181,11 @@ const Dropdown = React.forwardRef(function Dropdown(
         <ListBox.Menu {...getMenuProps()}>
           {isOpen &&
             items.map((item, index) => {
-              const itemProps = getItemProps({ item, index });
+              const itemProps = getItemProps({
+                item,
+                index,
+                disabled: item?.disabled,
```
joshblack:
```suggestion
                disabled: item.disabled,
```

I think `item` should always be defined / an object, or is that not the case?

joshblack (changes requested):

packages/react/src/components/Dropdown/Dropdown.js
tay1orjones: @joshblack `item` is but `disabled` is optional and may not always be available

joshblack: @tay1orjones nice, if `item` will be always be defined then what is the purpose of using `?.`? I had thought that the intent of optional chaining was that it would short-circuit if the value on the left hand was null-ish. In this case, since `item` is always an object, it seems like it'd be the same as `item.disabled` since they'd both return `undefined`.

tay1orjones: @joshblack Yeah! Sorry I overlooked this, I could've sworn I ran into a problem where `item` wasn't available. The optional chain fixed it, but I can't figure out how to repro it. Could be from my initial explorations and unrelated to how I ended up implementing this. 

I just pushed up a commit removing them 👍

joshblack (approved):

dakahn (approved):

PR #9914, opened 2021-10-20
feat(TimePickerSelect): updated props for v11

sstrubberg:
REF #9535 #9624 

Refactoring `TimePickerSelect` with prop changes in preparation for migration to a functional component.

#### Storybook

- Removed `iconDescription` from "Select" in the `TimePicker` story.
- Added an `aria-label` knob to "Select" in the `TimePicker` story.

#### TimePickerSelect

- Added a deprecation warning to `iconDescription`.
- Remove the unused `inline` prop.
- Put the props in alphabetical order.
- Renamed `...other` -> `...rest`
- Added an `ariaLabel` prop.
- Added proper aria attributes to the Chevron.

#### Testing
- Make sure `TimePicker` tests are passing.
- Ensure `TimePicker` does it's thing in Storybook, especially as it pertains to `TimePickerSelect`.


packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -37,7 +42,7 @@ export default class TimePickerSelect extends Component {
     disabled: PropTypes.bool,
 
     /**
-     * Specify whether the label should be hidden, or not
+     * hideLable has been deprecated in v10 and will be removed in v11.
```
abbeyhrt:
```suggestion
     * hideLabel has been deprecated in v10 and will be removed in v11.
```

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -46,20 +51,19 @@ export default class TimePickerSelect extends Component {
...
 
     /**
      * Specify a custom `id` for the `<select>`
      */
     id: PropTypes.string.isRequired,
 
-    /**
-     * Specify whether you want the inline version of this control
-     */
-    inline: PropTypes.bool,
```
abbeyhrt, line 61: Should `inline` be deprecated too or is it cool for it to just be removed?

tay1orjones: It doesn't look like it was being used, and it had an `// eslint-disable-line` next to it in the props destructuring. Looks like it's been around for a long time and hasn't been used. I think it's safe to delete. I'm curious if this is what you understood as well @sstrubberg?

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -46,20 +51,19 @@ export default class TimePickerSelect extends Component {
...
 
     /**
-     * Provide a description for the twistie icon that can be read by screen readers
+     * iconDescription is deprecated in v10 and will be removed in v11.
      */
-    iconDescription: PropTypes.string.isRequired,
+    iconDescription: deprecate(
+      PropTypes.string,
+      'The `iconDescription` prop for `TimePickerSelect` is no longer needed and has ' +
+        'been deprecated. It will be removed in the next major release.'
```
tay1orjones: Should we recommend `ariaLabel` usage in its place?

```suggestion
      'The `iconDescription` prop for `TimePickerSelect` is no longer needed and has ' +
        'been deprecated. It will be removed in the next major release. Use `ariaLabel` instead.'
```

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
sstrubberg: Took the words right out of my mouth :)

tay1orjones (commented):
Overall looks good!

Do we need to change the `labelClasses` block w/ `visually-hidden` to always be true since the `hideLabel` default has been removed? https://github.com/carbon-design-system/carbon/pull/9914/files#diff-a67cbb553a741d3befb1f3b64288b2c0acf8e9f033976bf1f2bd5a7553138ad8R97-R100

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -69,21 +73,19 @@ export default class TimePickerSelect extends Component {
...
 
   render() {
     const {
-      id,
-      disabled,
+      ['aria-label']: ariaLabel = 'open list of options',
       children,
-      iconDescription,
       className,
-      hideLabel = true,
```
tay1orjones, line 83: @sstrubberg Cool! I think we should leave in this `true` default until then

tay1orjones (approved):

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -16,6 +16,11 @@ const { prefix } = settings;
 
 export default class TimePickerSelect extends Component {
   static propTypes = {
+    /**
+     * Provide aria-label to the <select> element
+     */
+    ariaLabel: PropTypes.string,
```
joshblack: Does this need to be `aria-label` according to how it's used in `render`?

```suggestion
   ['aria-label']: PropTypes.string,
```

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -46,20 +51,19 @@ export default class TimePickerSelect extends Component {
...
 
     /**
-     * Provide a description for the twistie icon that can be read by screen readers
+     * iconDescription is deprecated in v10 and will be removed in v11.
      */
-    iconDescription: PropTypes.string.isRequired,
+    iconDescription: deprecate(
+      PropTypes.string,
+      'The `iconDescription` prop for `TimePickerSelect` is no longer needed and has ' +
+        'been deprecated. It will be removed in the next major release. Use `ariaLabel` instead.'
```
joshblack: I think the message would need to be that `iconDescription` has been replaced with `aria-label` and will be removed in the next major release. Use `aria-label` instead.

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
```diff
@@ -107,17 +109,18 @@ export default class TimePickerSelect extends Component {
...
+          {...rest}>
           {children}
         </select>
         <ChevronDown16
           className={`${prefix}--select__arrow`}
-          aria-label={iconDescription}>
-          {iconDescription && <title>{iconDescription}</title>}
-        </ChevronDown16>
+          aria-label={ariaLabel ? ariaLabel : iconDescription}
+          aria-hidden="true"
```
joshblack:
```suggestion
```

I think we might want to keep the `aria-hidden` off of the SVG, right? So the label could be read. Let me know if I'm misunderstanding

joshblack (changes requested):

packages/react/src/components/TimePickerSelect/TimePickerSelect.js
sstrubberg: This came up in a [previous PR](https://github.com/carbon-design-system/carbon/commit/a9c13fffe27b61c5662b785af0010489caa18ae1#diff-22f235c4784785309ea4abec2fa146f38d2be0d56b42466af68a247a0fe898b3) i did for Modal. @dakahn had suggested this for accessibility. Let me know if I misinterpreted that.

abbeyhrt (approved):
🎉

joshblack (approved):

PR #9923, opened 2021-10-21
feat(TimePickerSelect): class -> functional

sstrubberg:
REF #9712 

Refactored `TimePickerSelect` into a functional components.

**Changed**

Removed `hideLabel` prop and `label` logic.
Removed deprecated iconDescription.
Added aria-label to the `select` element
Added aria-hidden to the icon.

**Testing / Reviewing**

1. Pull down locally
2. Go to `packages/react`
3. Run `yarn start:v11` to run old storybook with v11 components.

**Questions**

- Should I have just removed the `hideLabel` prop and **kept** the `label` logic? In other words, do we need `<label>` for anything?


joshblack (approved):

packages/react/src/components/TimePickerSelect/next/TimePickerSelect.js
```diff
@@ -0,0 +1,89 @@
...
+  const prefix = usePrefix();
+
+  const selectClasses = cx({
+    [`${prefix}--select`]: true,
+    [`${prefix}--time-picker__select`]: true,
+    [className]: className,
+  });
+
+  return (
+    <div className={selectClasses} ref={ref}>
```
jnm2377: wondering if we would prefer the ref be on the select element instead? Since that's the input here? cc: @joshblack

joshblack: That makes sense to have the `ref` be on the `<select>` to me 👍

abbeyhrt (approved):

jnm2377 (approved):

PR #9924, opened 2021-10-21
fix(MultiSelect): remove aria hidden on ListBoxSelection button

dakahn:
Multiselect was throwing errors in Accessibility Checker for having button with interactive descendants being aria-hidden. This removes the aria-hidden attribute from that button and fixes the error

note: we might want to put this back on and build out some kinda of escape to clear functionality (as seen elsewhere) if we don't actually want this button to be in the accessibility tree

#### Testing / Reviewing

- make sure that Multiselect is working as intended
- make sure that Multiselect with selections isn't throwing errors in Accessibility Checker

[Internal slack discussion](https://ibm-studios.slack.com/archives/C046Y0YUD/p1634776657291700)


tay1orjones (approved):
👍 No errors in accessibility checker chrome plugin

aledavila (approved):

Issue #9919, opened 2021-10-21
[Feature Request]: 


TannerS:
### Summary

The Carbon react component `fileUploaderDropContainer` has a prop to accept file types. However the regex for this, does not consider file extensions for `_` or `-`. We wish to request this be added.

### Justification

I cant speak for other usages, but for our microservice, we allow multiple file types, and when adopting the `fileUploaderDropContainer`, we learned the following file types do not work, blocking us from using it 

```
gz_aa
gz_ab
gz_ac
pub_key
sjc03-raid-2
sjc03-raid-log-1
tgz-aa
tgz-ab
tgz-ac
tgz-ad
tgz-ae
tgz-af
tgz-ag
tgz-ah
tgz-ai
tgz-aj
tgz-ak
tgz-al
tgz-am
```

### Desired UX and success metrics

Can allow anyone using this component a wider range of accepted file typed 

### Required functionality

We wish to add this functionality to add the `fileUploaderDropContainer` to the microservice but limited on the file types we support

### Specific timeline issues / requests

We wish to get this sooner than later but we can work details out, but overall, I think this change would be good for everyone 

### Available extra resources

I Dont mind trying to make the PR for this, but I need to learn the carbon system and how to run this library locally 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

PR #9929, opened 2021-10-22
feat(fileuploaderdropcontainer): added pattern for accepted file ext

TannerS:
Closes #

https://github.com/carbon-design-system/carbon/issues/9919

#### Changelog

**Changed**

I made a change based off a feature request I proposed and if the idea is worth taking into account, I thought I could take a gamble at it for my first carbon PR and get some experience with contributing to carbon

Basically, we found ourselves allowing file extensions that contain `-` an `_` for our project. Even though we allow it in the `accept`  prop, the built in regex does not allow it. So I thought if we could give he user access to changing/modifying that regex directly to work with what they pass into the `accept` prop it could help cases like ours. This best making changes to the regex like hard coding things for our need or every person's need that may or may not come a need.

#### Testing / Reviewing

My attempt at testing was a few things.

Regression test was my main thing. I added the default regex as a default prop to keep default behavior and tested this in a local storybook copy (after modifying the story book) and added unit test. 

#### Thoughts

This idea is something I believe can be useful but i understand if this idea gets thrown out. If it does or does not, if there is any advice for the code or unit test (coding design, style, framework, etc) please let me know so I can learn and do better next time, thanks!


joshblack (approved):
🎉

jnm2377 (approved):
Thanks for contributing this! It looks great! 🎉

Issue #6519, opened 2020-07-22
[code snippet] add line number modifier


laurenmrice:
Add a line number functionality to the multi-line code snippet. Possibly should be a prop to turn on and off since this is a modifier for the existing multi-line code snippet.

-----------------------------------------------------------------------------------------------------------------------------

Spec:

![lg-1056px-16 column](https://user-images.githubusercontent.com/43969356/88200283-91244000-cc0b-11ea-8ccf-492af36bb070.png)

PR #9932, opened 2021-10-24
refactor(CodeSnippet) children proptype to node

### LINKED ISSUE #9846: [Feature Request]: Allow React nodes as children in CodeSnippet
### Summary

Allowing for react nodes in CodeSnippet allows developers to utilize the component when rendering more complex code snippet visualisations. 

### Justification

In our case, I wanted to add line-numbers to the component, so we did `codeString.split("\n").join((line) => <span>{line}</span>)` to achieve that effect. Typescript was bothered by this though, as the component expects `children` to be a string. Doing `<CodeSnippet>{parsedCodeString as any}</CodeSnippet>` worked perfectly fine, but it makes me uneasy considering the component could potentially change and need that prop to be a string specifically. 

Allow developers to pass `ReactNode`, `ReactElement` or `ReactNode[]` (or similar) as `children` to the `CodeSnippet` component, and update the types to reflect this. 

### Required functionality

The ability to pass react nodes as children to `CodeSnippet` without throwing a Typescript error or having to do `as any` to buypass said error.

Whenever you're able to get to it, no worries. You may consider this a question as well - do you **need** the `children`-prop on the `CodeSnippet` component to be a string, or is it just fed directly into the resulting `<code>`-element's children? Since "Show more"/"Show less" still works fine when I passed ReactNode children as `any`, I'm assuming you are calculating that based on a ref-height, and not the string's length?

### Available extra resources

None as of now, but very interested in contributing down the line

### Code of Conduct

imp-dance:
Closes #9846 

Allows for passing nodes to the CodeSnippet components, and alternatively passing a `copyText` string if you want to overwrite copy value from the node's ~`textContent`~  `innerText`.

#### Changelog

**New**

- Ability to pass `copyText` as alternative to the code-element's ~`textContent`~  `innerText` value.

**Changed**

- CodeSnippet's `children`-prop now accepts nodes as well, according to propTypes.
- The value to be copied in the CodeSnippet component depends on whether you pass a copyValue. If you don't, then the copy-text is determined by the code-element's ~`textContent`~  `innerText` value. 

#### Testing / Reviewing

This is my first time trying to contribute, so sorry for any silly issues. I ran `yarn test` and one of the tests failed, the following: 

`* Public API should only change with a semver change`

I was afraid that I wasn't updating the type in all the appropriate places. Maybe someone could point me in the right direction to fix this?


packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -146,7 +148,9 @@ function CodeSnippet({
   }, [handleScroll]);
 
   const handleCopyClick = (evt) => {
-    copy(children);
+    if (copyText || innerCodeRef?.current) {
+      copy(copyText ?? innerCodeRef?.current?.textContent);
```
joshblack: Just curious, when would we want to use `textContent` vs `innerText` for the default copy behavior? 👀

imp-dance: I was curious about the same thing! It seems to pretty much boil down to this quote from StackOverflow:

> innerText returns the visible text contained in a node, while textContent returns the full text. For example, on the following HTML `<span>Hello <span style="display: none;">World</span></span>`, innerText will return 'Hello', while textContent will return 'Hello World'.

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext

imp-dance: @joshblack I realise now that I've been a bit quick to pick `textContent`. I guess to answer your question specifically - when we know that the children of the element may possibly contain hidden elements that should not be included in the copy value - it's important to pick `innerText`. 

So actually, I think innerText might be a better option here. We don't know if it'll contain hidden text, but it could - since we're allowing nodes. `innerText` seems to be a better representation of what value would be copied if the user had manually highlighted all the text of the element and copied it. 

If it was a question of performance and you were calling `innerText` many times over and over then possibly `textContent` would be more performant but in this case that doesn't really matter much. 

What do you say, should we go for `innerText` instead?

joshblack: Nice, thanks for the explainer @imp-dance! I really appreciate it.

`innerText` totally makes sense to me based on what you wrote above 👍

imp-dance: Alright! Just pushed a commit to the PR which should resolve this 👍

joshblack:
So sorry about the delay on my end for this @imp-dance! 

Implementation wise, I think we're good to go but just wanted to ask to see if you felt like this was backwards-compatible or if we should reserve these kinds of changes for a breaking change 👀

imp-dance:
No worries, @joshblack ! 

I don't think this is a breaking change, since it's just additive, technically. `ReactNode` in TypeScript, or `PropTypes.node` both accept `string` as valid input. The new prop, `copyText` is also additive, so I suppose that won't break current implementations either.

joshblack:
Thanks, @imp-dance!

cc @carbon-design-system/developers-system for any other thoughts on the API addition 👀

dakahn:
i think the use case of advanced syntax highlighting is a good one here! 🏄🏾

dakahn (commented):
Would we want to provide an example use case in our Storybook for passing a node as a child here?

tay1orjones (commented):
Makes sense to me 👍  

I too went down the rabbit hole of looking at `innerText` vs `textContent` before seeing you had a whole conversation about it. I agree `innerText` looks like the better option.

I also think we should have an example of this in storybook for this as @dakahn suggested. Bonus points if a simple test case could be added covering the new proptypes.

imp-dance:
I was doing something akin to this to achieve line-numbers, could be a potential example: 

```jsx
export const withLineNumbers = () => (
  <CodeSnippet {...props()} type="multi">
    {snippetText()
      .multi.split('\n')
      .map((line, index) => (
        <div
          key={line + index}
          style={{
            whiteSpace: 'pre',
          }}>
          <span
            style={{
              display: 'inline-block',
              userSelect: 'none',
              width: '2rem',
              opacity: 0.5,
            }}>
            {index + 1}
          </span>
          {line} {'\n'}
        </div>
      ))}
  </CodeSnippet>
);
``` 

I was actually using a CSS counter and the ::before psuedo element but I think this implementation would be more clear in the example.

tay1orjones:
@imp-dance Very cool idea to use a css counter and pseudo elements. An additional concern I have though is that pseudo elements should only be used for decorative purposes, otherwise we'll fail [success criterion 1.3.1](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html). I think line numbers could be considered meaningful content and not just decorative in nature.

I think the example you posted could be good. Particularly because it would also showcase the `copyText` prop. 

For styling in stories, we typically create a `.scss` file in the same directory as the story file. This `.scss` file is then imported in the story file. Here's an example from Tabs:
https://github.com/carbon-design-system/carbon/blob/fd8ec5eb6522530ffc29d89f1dbdbfd9bd8feb45/packages/react/src/components/Tabs/Tabs-story.scss#L12-L15
https://github.com/carbon-design-system/carbon/blob/fd8ec5eb6522530ffc29d89f1dbdbfd9bd8feb45/packages/react/src/components/Tabs/Tabs-story.js#L19

tay1orjones:
> if you're considering adding features like for example line numbers, keep in mind that implementing them might become harder if this goes through, as we'll have to consider the possibility of nodes instead of just a simple string.

@imp-dance This is also a good point. We have an open issue to implement line numbers over in #6519.

imp-dance:
@tay1orjones Just a new commit with the story, refactored with proper scss instead of inline-styles. 👍

tay1orjones (approved):
Sorry for the delay here, this looks good! I've re-requested reviews from my fellow reviewers as well 👍

joshblack (approved):

dakahn (approved):

dakahn:
@imp-dance looks like we just need to update our snapshots tests, go ahead an run `yarn test -u` and commit those changes, friend 🏄🏾

Issue #9888, opened 2021-10-18
Update components to use latest tokens in `@carbon/styles`


joshblack:
The following components are still using the compatibility theme in our `@carbon/styles` component. This issue is to figure out the correct tokens for these components and make the appropriate updates.

List of components

- [x] Code Snippet @sstrubberg 
- [x] Data Table (expandable) @sstrubberg 
- [x] Date Picker @sstrubberg 
- [x] Dropdown @sstrubberg 
- [x] Flatpickr @sstrubberg 
- [x] List Box @sstrubberg 
- [x] Number Input @sstrubberg 
- [x] Overflow Menu @sstrubberg 
- [x] Popover @joshblack 
- [x] Radio Button @sstrubberg 
- [x] Search @sstrubberg 
- [x] Structured List @motou 
- [x] Tile @sstrubberg 
- [x] TreeView @motou

PR #9937, opened 2021-10-25
style(CodeSnippet): replaced tokens

sstrubberg:
REF #9888 

Updated tokens to v11.


joshblack (changes requested):
Make sure to update `@use '../../compat/theme'` to `@use '../../theme'` 👍 

Unfortunately, our internal usage of the compat theme is causing some issues downstream with beta testers so we want to make sure we're using the theme import instead 👀

jnm2377 (approved):

joshblack (approved):

PR #9939, opened 2021-10-25
feat(styles): add default feature flag config

### LINKED ISSUE #9927: [Bug]: The v11 feature flag is not enabled in the styles package
In our styles package, we do not include the feature flags package nor do we have the v11 feature flag turned on. As a result, styles that have been brought over that use `feature-flag-enabled` do not work as intended.

This issue is to update the usage of `feature-flag-enabled` to use the feature flags package directly and correctly any issues that are caused by not having the flag defined.

joshblack:
Closes #9927 

Update our usage of feature flags in the project. This also updates our storybook to correctly initialize feature flags.

#### Changelog

**New**

**Changed**

- Add feature flag side effect import to storybook
- Add default feature flag config for styles
- Update `feature-flag-enabled` usage in notification to use feature-flags
- Update `toggle` to use feature-flags namespace
- Update Sass to latest version

**Removed**

#### Testing / Reviewing

- Verify that the toggle story renders as expected in the carbon-react-next storybook


tay1orjones (approved):
Notifications and Toggle look good in the next storybook 👍

sstrubberg (approved):
🚀

PR #9951, opened 2021-10-26
style(DatePicker): updated tokens

sstrubberg:
REF #9888 

Update tokens to v11 based on [the following specs](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/11.x-color.md).

- Brought in button tokens.

Updated the following:
- `ui-01` -> `layer-01`
- `text-01` -> `text-primary`
- `hover-ui` -> `background-hover`
- `ui-05` -> `border-inverse`
- `disabled-02` -> `text-disabled`
- `icon-01` -> `icon-primary`
- `interactive-01` -> `button-primary`
- `interactive-04` -> `border-interactive`
- `text-05` -> `text-helper`
- `link-01` -> `link-primary`
- `text-04` -> `text-on-color`
- `ui-02` -> `layer-02`


packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -338,22 +338,22 @@
     }
 
     &:hover::after {
-      border-top-color: $interactive-01;
-      border-bottom-color: $interactive-01;
+      border-top-color: $background-brand;
+      border-bottom-color: $background-brand;
```
tay1orjones: I think it might make more sense to have these map to the `button-primary` token since these styles appear to be related to the increment up/decrement down buttons

```suggestion
      border-top-color: $button-primary;
      border-bottom-color: $button-primary;
```

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -423,29 +423,29 @@
...
     &:hover {
-      background: $hover-ui;
+      background: $background-hover;
     }
 
     &:focus {
       @include focus-outline('outline');
 
-      outline-color: $interactive-01;
+      outline-color: $background-brand;
```
tay1orjones: Same thing here since I believe this is a button?

```suggestion
      outline-color: $button-primary;
```

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -475,18 +475,18 @@
...
   .flatpickr-day.inRange {
     background-color: $highlight;
-    color: $text-01;
+    color: $text-primary;
   }
 
   .flatpickr-day.selected {
-    background-color: $interactive-01;
-    color: $text-04;
+    background-color: $background-brand;
```
tay1orjones:
```suggestion
    background-color: $button-primary;
```

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -501,23 +501,23 @@
...
-    background: $ui-01;
-    color: $text-01;
+    background: $layer-01;
+    color: $text-primary;
   }
 
   .flatpickr-day.endRange.inRange.selected {
-    background: $interactive-01;
-    color: $text-04;
+    background: $background-brand;
```
tay1orjones:
```suggestion
    background: $button-primary;
```

tay1orjones (changes requested):
Let me know what you think about these suggestions. I'm not sure when a `brand` token wins out against `button` if the element is effectively a button.

packages/styles/scss/components/date-picker/_flatpickr.scss
sstrubberg: Good call. It was breaking when i used `$button-primary`. I probably need to find where the import for that comes.

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -222,19 +223,19 @@
...
-      background-color: $hover-ui;
+      background-color: $background-hover;
     }
   }
 
   .flatpickr-next-month.disabled svg,
   .flatpickr-prev-month.disabled svg {
     cursor: not-allowed;
-    fill: $ui-05;
+    fill: $border-inverse;
```
jnm2377: this is an svg, so should it use an icon token? I remember when updating other component tokens, some of the old tokens weren't correct to begin with.

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -222,19 +223,19 @@
...
   }
 
   .flatpickr-next-month.disabled:hover svg,
   .flatpickr-prev-month.disabled:hover svg {
-    fill: $ui-05;
+    fill: $border-inverse;
```
jnm2377: same as above

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -304,15 +305,15 @@
     border-bottom: 0;
 
     &::after {
-      border-bottom: rem(4px) solid $icon-01;
+      border-bottom: rem(4px) solid $icon-primary;
```
jnm2377: and I think this would need a border token, not icon token

packages/styles/scss/components/date-picker/_flatpickr.scss
```diff
@@ -304,15 +305,15 @@
...
     }
   }
 
   .numInputWrapper .arrowDown {
     top: rem(11px);
 
     &::after {
-      border-top: rem(4px) solid $icon-01;
+      border-top: rem(4px) solid $icon-primary;
```
jnm2377: same as above, border token?

jnm2377 (commented):
I just have some questions around some of the tokens. Maybe we need design review on this for the tokens?

packages/styles/scss/components/date-picker/_flatpickr.scss
tay1orjones: `icon-disabled` or `icon-on-color-disabled ` maybe since this is the disabled state of the next/prev button icon?

tay1orjones (approved):
Looks good once @jnm2377's feedback above is addressed 👍

jnm2377 (approved):

PR #9956, opened 2021-10-27
fix(DataTable): fixes overflowing sort button and select column spacing

### LINKED ISSUE #9837: [Bug]: Batch checkbox in DataTable header is incorrectly sized and is clipped by the next table header when sorted
### Package

carbon-components

### Browser

Chrome, Firefox (probably all browsers)

### Package version

v10.45.0

### Description

The checkbox for the batch actions on the DataTable header is clipped by the next column. This isn't obvious until you add sorting which changes the background color of the next column.

image

It's very obvious when using the derived @carbon/ibm-security library as the header background is opaque making the checkbox be clearly clipped (without needing sorting.)

image

The provided codesandbox demonstrates the issue when you sort the column next to the checkbox column (it clips the checkbox) but again its not the sorting that's the issue, the checkbox column is incorrectly sized but its only noticeable using carbon-components when sorting is used.

### CodeSandbox example

https://codesandbox.io/s/2lx3v

### Steps to reproduce

Add batch actions and isSortable to a DataTable then sort by the column next to the batch action checkbox column - the sorted column header will then partially cover the batch action checkbox.

Or without sorting using the devtools view the width of the checkbox column and the column next to it, you'll see they overlap.

### Code of Conduct

jnm2377:
Closes #9837 

#### Changelog
- adds specific width to checkbox column, so that it's not clipped by sortable headers.
- adds selectable + sortable story under the Selection stories

#### Testing / Reviewing

- Review the Select with Sortable story, review that the checkbox is no longer being clipped by the sortable header
- there should still be 20px of spacing between the checkbox and the sortable header text (4px right padding on checkbox + 16px left padding on sortable btn)
Screen Shot 2021-10-26 at 8 50 55 PM
Screen Shot 2021-10-26 at 8 50 39 PM


sstrubberg (approved):
Nicely done updating the `@carbon/styles` package too! 🚀

laurenmrice (approved):
Looks good to me!

joshblack (approved):

tay1orjones (approved):

PR #9959, opened 2021-10-27
fix(Tooltip): remove click stopPropagation

### LINKED ISSUE #9958: [Bug]: [React] Tooltip handleClickOutside is buggy with React 17 and two Tooltips
### Package

carbon-components-react

### Browser

Chrome

### Package version

10.46.0

### Steps to repro:

- Install latest carbon-components-react
- Install React 17
- Implement a page with at least two Tooltip components on it
- Click on one of the tooltips to open it
- Open the other tooltip

### Expected result:

- The first Tooltip closes immediately when I click the other one

### Actual result:

- The first Tooltip stays open
- Clicking on the first tooltip again causes it to close and reopen immediately, creating a "blinking" effect

**The bug is not present with React 16**

### Suggested fix:

Remove the e.stopPropagation() call from the handleMouse method in Tooltip.js

### Notes:

With React and React-DOM 17 installed, the handleClickOutside is buggy. React 17 rewrote the event handling which could introduce breaking changes. The event listeners are not on the HTML element anymore but on the React root DOM element. This might be the root cause of this issue. 

Anyway: the Tooltip.js `handleMouse` method calls an `e.stopPropagation()` on click events. At the same time, there's a `handleClickOutside` method that captures every click that's outside the Tooltip itself, and closes the tooltip on such an event. 

This leads me to believe that the `handleClickOutside` is prevented when you click on another Tooltip. It works fine if you click anywhere else, of course. But the other tooltip calls the stopPropagation method on the SyntheticEvent object and it prevents the ClickOutside from executing. 

It's highly interesting that the bug is not present with React 16, although just looking at the code it would look like it should not be affected by the react version.

Please note that I couldn't figure out why the stopPropagation is called in the handleMouse method. There might be a good reason for it to exist.

### CodeSandbox example

https://codesandbox.io/s/carbon-components-react-forked-ujvjk?file=/src/index.js

### Steps to repro using the Codesandbox example:

- Click on one of the tooltips to open it
- Open the other tooltip

### Expected result:

- The first Tooltip closes immediately when I click the other one

### Actual result:

- The first Tooltip stays open
- Clicking on the first tooltip again causes it to close and reopen immediately, creating a "blinking" effect

### To see it working with React 16:

On the left hand side panel, change the react and react-dom dependencies to version 16.

Go back to the steps to repro. 

The actual result in this case is the Tooltip closes as expected.

### Code of Conduct

FrivalszkyP:
Closes #9958

#### Changelog

**Removed**

- Remove e.stopPropagation() from the click event handler. In React 17, the stopPropagation prevents other Tooltip's `clickoutside` handlers from executing. It probably affects other clickoutside event handlers from executing. 

**Please note: removing it might cause other side effects but I couldn't come up with any scenarios like that.**

#### Testing / Reviewing

See steps to repro in this Issue: #9958


tay1orjones:
Hey @FrivalszkyP! Thanks for submitting this change. Good catch here - the changes to event pooling in React v17 are probably the culprit. 

The `stopPropagation ` was added quite a while ago. The example used in the commit message https://github.com/carbon-design-system/carbon/commit/d32424ca421f920f7ee471dcef15e8f9d522fc8d is regarding tooltips inside buttons.

I'll do some quick tests on this to see if I can find any side effects with tooltips in modals or tooltips.

tay1orjones (approved):
I couldn't find any side effects 👍

FrivalszkyP:
@tay1orjones, thanks for being so thorough, I really appreciate it!

FrivalszkyP:
@tay1orjones I tested it and the Tooltip inside a Button issue could definitely be a problem if we merge this PR. On the other hand, a Tooltip inside a Button is just a button inside another button, in terms of the AOM, so it should be avoided anyway. So the original reason the stopPropagation() was added would have been better handled by refactoring the design and _not_ adding the Tooltip inside a Button.

tay1orjones:
@FrivalszkyP Yes I agree, tooltips shouldn't be placed inside buttons. We do use a tooltip for the `hasIconOnly` variant, but when clicking/dismissing the tooltip the same behavior is seen between the deploy preview storybook here and the production storybook.

https://user-images.githubusercontent.com/3360588/139895134-370b6ce0-7696-40f3-8d61-8c38ee161098.mov

FrivalszkyP:
@tay1orjones interestingly, it works differently for me. Chrome 95 on MacOS Big Sur 11.6. Using the deploy preview, clicking on the tooltip bubble does not close the bubble for me. I'll take a closer look on this, also I'll add storybook `action`s for the button onclick event and see whether the event fires on my branch. Thanks for bringing this to my attention!

dakahn (approved):

ZachStowellIBM:
@FrivalszkyP @tay1orjones - I'm actually running into an issue due to these changes.

We have a DataTable, that has sortable headers.  One of these headers has a tooltip in it.
Screen Shot 2021-12-14 at 4 49 14 PM
Screen Shot 2021-12-14 at 4 49 29 PM

The issue we run in to is the exact scenario you were discussing above, about tooltips inside of buttons (since the ability to toggle the sort order is effectively a button). 

What's the best course of action here from a Carbon perspective?  This seems like a use case where we **would** want to `stopPropagation`, however there's no mechanism in the Tooltip component to do so.

tay1orjones:
@ZachStowellIBM Unfortunately, interactive controls can't have focusable descendants like this. Nested interactive controls [are not announced by screen readers](https://dequeuniversity.com/rules/axe/4.2/nested-interactive). Because of this, placing a Tooltip inside a parent element that is interactive will fail [WCAG Success Criterion 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value). 

Until this PR, the Tooltip masked this problem for quite some time by calling `stopPropagation` internally. It's been there all along though. The event propagation can be stopped in an `onClick` on an element wrapping the Tooltip, but I'd caution that as a temporary workaround.

Rel: https://github.com/carbon-design-system/carbon/issues/10284

ZachStowellIBM:
Thanks for the info @tay1orjones!  We'll be sure to make changes on our end.

PR #9963, opened 2021-10-27
fix(OverflowMenu): remove handlers from icon props

### LINKED ISSUE #9962: [Bug]: [React] OverflowMenu disabled state can be circumvented by clicking on the icon
### Package

carbon-components-react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

10.46.0

### Description

There's a bug inside the OverflowMenu component. It attaches the keyboard and the click handlers to **both** the button element and to the icon element inside the button. Thus, these events can be triggered by the button **and** the icon both. 

This means that if you also pass the `disabled` prop to the OverflowMenu component, even though it will disable the button, the icon inside it will still propagate the click event. That's because the button itself might be disabled but that setting does not apply to the non-interactive elements inside the button.

Proposed solution: the handleClick and the onKeyDown handlers should not be passed down to the iconProps.

### CodeSandbox example

https://codesandbox.io/s/carbon-components-react-forked-xeijh?file=/src/index.js

### Steps to reproduce

- Implement OverflowMenu with any icon (default is good as well)
- Add `disabled` attribute to the OverflowMenu
- Confirm that the button element receives the disabled property
- Click on the button
- Click on the icon inside the button

Expected result:

- Clicking the button and clicking the icon inside the button should make no difference. The button should be disabled

Actual result:

- When you click anywhere inside the button that's not the icon, the button does not react.
- When you click on the icon inside the button, the overflowmenu opens.

### Code of Conduct

FrivalszkyP:
Closes #9962

OverflowMenu disabled state can be circumvented by clicking on the icon. That's because the icon props also receive the click handler. 

#### Changelog

**Removed**

- Remove the two event handlers from the OverflowMenu iconProps

#### Testing / Reviewing

- Open the OverflowMenu Playground in the Storybook
- Click the `disabled` knob
- Click on the icon inside the overflowmenu trigger button.


joshblack (approved):

dakahn (approved):

FrivalszkyP:
I'm seeing that the snapshot test is failing. For some reason, if I execute the test updates as described in the CONTRIBING.md, none of my test files get updates. Even though they should be updated because I removed two handler props. Can someone help me out?

Issue #9941, opened 2021-10-25
Ship initial version of popover based on design guidance 


joshblack:
This issue is for bringing Popover over the initial hurdle and stabilizing it for usage in other components.

**Checklist**

- [ ] Terminology
- [ ] Docs
  - [ ] Storybook docs
  - [ ] Edits on usage docs
- [ ] Component tokens
  - [x] `offset`
  - [x] `border-radius`
  - [x] `drop-shadow`
  - [x] `caret-offset`
  - [ ] `inline` and `block` padding
- [ ] `aria-owns` for menus
- [ ] Placement
  - [ ] Auto placement

**Questions**

- [ ] How can popover width be tied to grid columns?
- [ ] How would we accomplish a 'tab tip' variant? Would this be an icon button + popover?
- [ ] Should we make it clear that the "UI Trigger" is from a different component in usage?
- [ ] How do we make placement work in RTL? (block/inline terminology)
- [ ] Tab tip -> how would we do the shadow?
- [ ] Should we use placement and alignment as the pop names?
- [ ] Should styles be RTL out of the box?

PR #9965, opened 2021-10-27
refactor(popover): update styles and add new story

joshblack:
Reference #9941

#### Changelog

**New**

- Add tests for popover

**Changed**

- Update popover to use new tokens and update styles to use new CSS features now available

**Removed**

#### Testing / Reviewing

- Look at the story in storybook and verify that it works in every direction


packages/styles/scss/components/popover/_popover.scss
```diff
@@ -18,28 +18,47 @@
 /// @access private
 /// @group components
 @mixin popover {
-  $popover-text-color: $text-01;
-  $popover-caret-offset: 1rem;
-  $popover-offset: 8px;
+  // The background color for the popover container
+  $popover-background-color: custom-property.get-var(
```
abbeyhrt, line 41: Woah! I've never seen this `get-var`, what does it do? Is it getting that theme layer token and then assigning `popover-backround-color` to it?

joshblack: @abbeyhrt good question! Here's a breakdown from the module: https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/utilities/_custom-property.scss#L10-L29

Basically `get-var` gives you back the `var(--cds-token-name)` format of the token.

abbeyhrt (approved):

laurenmrice (changes requested):
- The dropshadow should be `0 2px 6px 0 rgba(0, 0, 0, 0.2)`

- When a radio button under the "Caret Position" section is selected, the caret tip is not showing up in the storybook demo on the popover.  Is that related to the changes in this PR?

aledavila (approved):
lgtm

laurenmrice (changes requested):
**Storybook demo spec:**
- [x] Instead of using a 48px container and a 32px icon for the trigger, lets use a `32px container` and a `16px icon`.
- [x] When the caret tip is set to true: there should be a `4px space` between the caret and the container bounding box.
storybook demo spec


-----

**Direction alignments:**
When the caret tip is set to true, most of the alignments looks good, but some popovers just need a 2px nudge)
- [x] Top-left (move popover 2px to right)
- [x] Top-right (move popover 2px to left)
- [x] Bottom-left (move popover 2px to right)
- [x] Bottom-right  (move popover 2px to left)
- [x] Left-bottom (move popover 2px up)
- [x] Left-top (move popover 2px down)
- [x] Right-bottom (move popover 2px up)
- [x] Right-top (move popover 2px down)

joshblack:
@laurenmrice should be good to go for a re-review!

dakahn (approved):
Cool!

laurenmrice (changes requested):
- Directions look great in Safari/Chrome/Firefox! 🎉

- The only thing we need to adjust now is that the 16px checkbox icon looks like it was enlarged to fill the 16px container around it. The true asset of the 16px icon has some padding around the artwork inside of the 16px container. https://carbon-elements.netlify.app/icons/examples/preview/#Checkbox16

- Example of current code versus design spec:
Artboard

joshblack:
@laurenmrice updated!

laurenmrice (approved):
Looks great ! 🎉

PR #9969, opened 2021-10-27
style(tile): update color tokens

motou:
REF #9888

- Updated the following:
  - ui-02 -> layer-02


motou:
It seems that the PR https://github.com/carbon-design-system/carbon/pull/10008 has been merged, so this PR is deprecated

PR #9970, opened 2021-10-27
style(structured-list): update theme import

motou:
REF #9888


joshblack (changes requested):
It seems like there are some tokens still being used from `compat/theme` potentially like `$hover-row` that need to be replaced 👀

sstrubberg (approved):

joshblack (approved):

motou:
@aagonzales sure, the token doc is updated.

tay1orjones (approved):

sstrubberg:
@motou Can you enable this?

image

PR #9971, opened 2021-10-27
style(search): update color tokens

motou:
REF #9888

- Updated the following:
  - disabled-01 -> text-on-color-disabled
  - disabled-01 -> field-disabled


packages/styles/scss/components/search/_search.scss
```diff
@@ -71,11 +71,11 @@
   .#{$prefix}--search-input[disabled] {
     border-bottom: 1px solid transparent;
     background-color: $field-disabled;
-    color: $disabled-01;
+    color: $text-on-color-disabled;
```
motou, line 74: This could be a regression bug. Since `$field-disabled` is equal to `$disabled-01`, it means that the text color and the background color of a disabled search input are the same.

Hence I changed the text color to `$text-on-color-disabled` (equal to `$disabled-03`)

tay1orjones:
Thanks so much for the contribution here @motou and for adding the feedback suggestion on Scott's PR! I'm going to close this as a duplicate since we've already got some eyes from the design team over on the other PR.

tay1orjones:
Duplicate of #9993

PR #9974, opened 2021-10-27
feat(react): add ContainedTab and ContainedTabs

### LINKED ISSUE #9952: [Tabs]: Separate and rename container tabs to ContainedTabs, ContainedTab
Part of #8487

Container tabs need to be renamed to be "Contained". Additionally, the spec and implementation between line and contained tabs is beginning to diverge. It appears separating these out into their own component makes the most sense from a usage perspective. `ContainedTabs` should have `ContainedTab` as children, not `Tab` or `LineTab`. `ContainedTabs` and `Tabs` (or `LineTabs` if we end up making it it's own component) may share a common primitive of `Tabs` to reuse scrolling/focus/etc functionality. 

- [ ] Add new ContainedTab/ContainedTabs under `/next`
   - [ ] ContainedTab should accept a new `secondaryLabel` prop
- [ ] Remove `type` from Tab/Tabs since contained tabs are going to be their own component

abbeyhrt:
Closes #9952 

Breaks out `ContainedTabs` and `ContainedTab` into their own components and removes `type` prop from `Tabs`. 

#### Changelog

**New**

- ContainedTabs
- ContainedTab

**Removed**

- `type` prop from `Tabs`

#### Testing / Reviewing

Check that ContainedTabs story works as expected and matches Container tabs in current storybook


abbeyhrt:
@joshblack let me know what you think!

joshblack (approved):

tay1orjones (approved):

PR #9993, opened 2021-10-28
style(Search): updated tokens

sstrubberg:
REF #9888


aagonzales (approved):

aledavila (approved):

joshblack (approved):

packages/styles/scss/components/search/_search.scss
```diff
@@ -70,12 +70,12 @@
 
   .#{$prefix}--search-input[disabled] {
     border-bottom: 1px solid transparent;
-    background-color: $field-disabled;
-    color: $disabled-01;
+    background-color: $field;
+    color: $field;
```
motou: Looks like the same issue that I discovered here: https://github.com/carbon-design-system/carbon/pull/9971#discussion_r737782924

The color and the background color are the same? ;)

tay1orjones (changes requested):
Thanks for the catch @motou ! Those shouldn't be the same value.

tay1orjones (approved):

PR #10003, opened 2021-10-31
fix(ComposedModal): Address several problems with selectorsFloatingMenus prop on ComposedModal

### LINKED ISSUE #10000: [Bug]: Several problems with `selectorsFloatingMenus` prop on ComposedModal
### Package

carbon-components-react

### Package version

v7.46.0

### Description

It is difficult to use the `selectorsFloatingMenus` prop on ComposedModal, for several reasons:

1. There is a mistake in the proptypes, declaring this prop as type 'string' when it should be array of string. The logic in the `wrapFocus` method assumes that an array will be provided, and causes errors if a string is supplied. If an array is supplied, it works correctly, but issues warnings (in non-production environments/builds) because of the proptypes mismatch. The proptypes should be fixed.
2. The value for this prop is not restructured out in the ComposedModal render method, meaning that it remains in the `...rest` object and thus gets passed through to the `<div>`, and this is both unnecessary and causes React error messages about adding camel-case attributes onto HTML nodes. The value should be remove from `...rest` by restructuring, so that the value is not passed through to HTML.
3. If no value is provided, an array of standard Carbon component CSS selectors is used as allowed focus escapes for the ComposedModal. However, if a value is provided then it replaces that array. This means that if a user needs to add a selector, but still wants other Carbon components to work within modals, they will need to include the same standard CSS selectors in the array they supply along with the value(s) they want to add. This is awkward, error-prone, and a maintenance risk (further selectors may need to be added in the future). It would be better if any value supplied for this prop ADDED TO, rather than replaced, the standard set of selectors.

### Code of Conduct

dcwarwick:
Closes #10000

#### Changelog

**Changed**

- packages/react/src/components/ComposedModal/ComposedModal.js -- fixed proptype, and remove prop from `...others` that are passed through to the DOM.
- packages/react/src/internal/wrapFocus.js -- default Carbon CSS selectors for floating menus/popus are added to prop value rather than replaced by it.

#### Testing / Reviewing

Build, tests, and storybook can help verify that this PR does not break existing usage. In particular, a story or test case with a menu/tooltip/date-picker inside a composed modal would be useful to verify that focus handling continues to work as expected.

I will also apply the changes locally to a build of https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1360 and verify that the issues that were problematic with that fix are resolved by these changes.


packages/react/src/components/ComposedModal/ComposedModal.js
```diff
@@ -91,7 +91,7 @@ export default class ComposedModal extends Component {
     /**
      * Specify the CSS selectors that match the floating menus
      */
-    selectorsFloatingMenus: PropTypes.string,
+    selectorsFloatingMenus: PropTypes.arrayOf(PropTypes.string),
```
joshblack, line 94: I think this is technically a breaking change, would we need to do `PropTypes.oneOfType()` and then support either option until the next major release?

dcwarwick: I would suggest not necessary, on the following basis: the prop is currently declared as a string, BUT actually passing a string causes a JavaScript error that aborts rendering. This is because `wrapFocus.js` line 32 tries to apply the array `some()` method to the passed value, which fails for strings. So there is no utility at all in actually passing a string, so supporting it as an alternative option isn't useful in this case. The only way to actually use this prop currently is to pass the array of strings and ignore the properties warnings.

joshblack: Nice, makes sense to me! Thanks so much for clarifying.

tay1orjones (approved):
These make sense to me - thanks for including so much detail in the original issue!

How fun that you got to open _and close_ the 10,000th issue on the repository! 🏅 🎉

joshblack (approved):

PR #10028, opened 2021-11-02
fix(form): restore label margin, remove margin from FormLabel only

### LINKED ISSUE #9947: Restore `margin-bottom` to labels inside input components
The `margin-bottom` on `.#{$prefix}--label` was recently removed in https://github.com/carbon-design-system/carbon/pull/9830. This removed margin from `FormLabel`, but also removed all margin from labels inside `TextInput`, `RadioButton`, etc.

This issue is to make sure that the form labels internal to components have margin built in, while the `FormLabel` component does not have any margin built in.

tay1orjones:
Closes #9947 

Ensures that `FormLabel` does not have margin-bottom, while other form labels built into components (TextInput, RadioButton, etc) have margin-bottom.

#### Changelog

**New**

- styles: form label "no margin" modifier class

**Changed**

- styles: restore margin to form label default class 

#### Testing / Reviewing

- With the v11 flag turned off 
  - The various form input component stories (`TextInput`, `RadioButton`, etc), the labels should retain their margin-bottom. No layout-breaking change should be being introduced here.
- With the v11 flag turned on
  - The `FormLabel` stories should not have margin-bottom
  - The various form input component stories (`TextInput`, `RadioButton`, etc), the labels should retain their margin-bottom.


joshblack (approved):

sstrubberg (approved):
lgtm!

PR #10029, opened 2021-11-02
chore(project): update internal react deps to v17

tay1orjones:
This change allows us to get more visibility into how our code works in storybook, builds, etc with React v17.

We've been getting some bug reports for issues only present when using the components with React v17 and in v11 we'll be bumping the peer deps to no longer include React v16. This change will help surface issues now, before we make React v17 the default.

#### Changelog

**Changed**

- project: update internal react deps to v17

#### Testing / Reviewing

- Ensure both storybooks build and all stories are viewable
- Status checks need to pass
- Generally, any defects discovered in components should be contained in new issues for triage and fixing


package.json
```diff
@@ -30,10 +30,10 @@
...
     "ast-types": "^0.14.0",
     "node-sass": "^6.0.0",
-    "react": "~16.9.0",
-    "react-dom": "~16.9.0",
-    "react-is": "~16.9.0",
-    "react-test-renderer": "~16.9.0"
+    "react": "~17.0.2",
+    "react-dom": "~17.0.2",
+    "react-is": "~17.0.2",
+    "react-test-renderer": "~17.0.2"
```
tay1orjones, line 36: @joshblack is there any problem you know of if we update the `resolutions` here? I don't think it should have any impact on downstream consumers, but let me know if I should revert this.

joshblack: Should be good! This will force one version for us in the monorepo so that we don't have to worry about multiple versions of React in the monorepo and the weird hook shenanigans that come from that.

packages/react/package.json
```diff
@@ -61,7 +61,7 @@
     "lodash.isequal": "^4.5.0",
     "lodash.omit": "^4.5.0",
     "lodash.throttle": "^4.1.1",
-    "react-is": "^16.8.6",
```
joshblack, line 64: Will this have any problems with people using React 16? 🤔

abbeyhrt: Out of curiosity, What does `react-is` do?

tay1orjones: @abbeyhrt It allows us to check if things are a valid React element type. I think the only place we use it is with `isValidElementType()` in the Public API snapshot to assert that components are valid before we add them to the snapshot "queue".

@joshblack I think we're good. There's no changelog that I could find, so I crawled through [the diff between versions](https://github.com/facebook/react/compare/v16.4.0...v17.0.2#) and it looks like there are no breaking changes. I think they bumped the versions of all core packages so they'd be the same values, even if they didn't include breaking changes.

They deprecated two aliases, `ReactIs.isAsyncMode()` and `ReactIs.isConcurrentMode()`. Will be removed in v18. `ReactIs.typeOf()` now supports some new types, `REACT_LAZY_TYPE`, `REACT_MEMO_TYPE`, `REACT_SUSPENSE_TYPE`.

tay1orjones:
Yikes. This causes almost 300 test failures due to Enzyme not yet having an adapter available for React v17. https://github.com/enzymejs/enzyme/issues/2429

tay1orjones:
@abbeyhrt I added the unofficial enzyme adapter for react v17 you recommended and it cut the failing tests down to just a handful! I've updated the failing ones here - proptype warnings using string substitution needed updated assertions with full argument lists.

@joshblack Let me know what you think about this. The unofficial adapter should just be a stop-gap until the official one is released or we eventually manage to remove enzyme completely in favor of react testing library.

abbeyhrt (approved):
Looks good to me!

joshblack (approved):

PR #10035, opened 2021-11-02
feat(tabs): make auto-width the default behind v11 flag

### LINKED ISSUE #10031: [Tabs]: make auto-width the default for both line and contained tabs

tay1orjones:
Closes #10031 

Makes tabs have automatic width by default in v11 instead of the static `10rem` they have in v10.

#### Changelog

**Changed**

- tabs: add auto-width behind v11 flag

#### Testing / Reviewing

- The deploy previews should still show the v10 behavior of having a static `10rem` width
- Pull down and run storybook with `yarn start:v11` - tabs should have `width: auto`

### DIFF
--- a/packages/components/src/components/tabs/_tabs.scss
+++ b/packages/components/src/components/tabs/_tabs.scss
@@ -710,9 +710,7 @@
       @include focus-outline('reset');
       @include type-style('body-short-01');
 
-      @if feature-flag-enabled('enable-v11-release') {
-        width: auto;
-      } @else {
+      @if not feature-flag-enabled('enable-v11-release') {
         width: rem(160px);
       }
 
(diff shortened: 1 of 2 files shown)


abbeyhrt (approved):

packages/components/src/components/tabs/_tabs.scss
```diff
@@ -710,8 +710,13 @@
       @include focus-outline('reset');
       @include type-style('body-short-01');
 
+      @if feature-flag-enabled('enable-v11-release') {
+        width: auto;
```
joshblack: Do we need to explicitly set it to `auto` because of another `width` declaration? I had assumed this was the default value so just wanted to check 👀

dakahn (approved):
Looks good to me

packages/components/src/components/tabs/_tabs.scss
tay1orjones: You're right, `auto` isn't needed. I'll remove 👍

PR #10041, opened 2021-11-03
fix(OverflowMenu): changed filename for story

### LINKED ISSUE #9804: [Bug]: Menu detaches from trigger on keyboard navigation
### Package

carbon-components-react

### Browser

Chrome

### Package version

10.45.0

### Description

For `OverflowMenu` implementation using `unstable_Menu`, navigating to a menu item via keyboard causes the container to scroll and the menu to detach from the trigger. I'm not sure if this behavior is intended, I think it's a bug.

### CodeSandbox example

n/a - present in storybook

### Steps to reproduce

1. Check out the [storybook story](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-menu-overflowmenu--custom-icon)
2. Click/enter to open the menu
3. Press the down arrow key to move down in the menu items
4. The bug - the container will scroll and the menu will "detach" from the trigger.
5. Go back up via keyboard arrow and the container will scroll back and "reattach" the menu to the trigger

https://user-images.githubusercontent.com/3360588/136066130-d6b61aae-12ee-48ec-922c-96437736b2cf.mov

### Code of Conduct

sstrubberg:
Closes #9804 

Cribbed a function from the old OverflowMenu to `preventDefault` on all the arrow keys when the state is set to open.

#### Testing

1) Open the next storybook
2) Navigate to the [unstable OverflowMenu](https://deploy-preview-10041--carbon-react-next.netlify.app/?path=/story/experimental-unstable-menu-overflowmenu--overflow-menu)
3) Inspect the element with your devtools and toss in `height: 42rem` or some other ridiculous amount of rem.
4) Open the overflow menu with either your mouse or keyboard, then use your keyboard to navigate the menu items.
5) It should be as expected instead of the menu items detaching from icon when you navigate

#### Questions

I feel like this code should be working, but for the life of me, it isn't. Would love an extra set of eyes on this :)


abbeyhrt (approved):

sstrubberg:
> Since this is adding in the story versus fixing the bug (let me know if I'm misunderstanding) should we remove the `Closes #9XYZ` text so the issue won't be closed before a fix lands? 👀

I guess the reason I put the `Closes #9XYZ` is because I couldn't recreate the issue. Totally cool to remove the closing text and then closing the issue separately. Whatever's clever!

sstrubberg:
@tay1orjones you drummed up the original issue. Any thoughts on this?

tay1orjones:
@sstrubberg I can't replicate the bug in the `carbon-react-next` storybook, but the original repro was made in the `carbon-components-react` storybook. I don't see the story being included in the `carbon-components-react` storybook. I think it should be in both.

It totally may be that the bug was fixed via a different change to how Menu is positioned (it seems to automatically pop "up" when there's not enough room to pop "down"), but I'm not sure.

sstrubberg:
> @sstrubberg I can't replicate the bug in the `carbon-react-next` storybook, but the original repro was made in the `carbon-components-react` storybook. I don't see the story being included in the `carbon-components-react` storybook. I think it should be in both.
> 
> It totally may be that the bug was fixed via a different change to how Menu is positioned (it seems to automatically pop "up" when there's not enough room to pop "down"), but I'm not sure.

Thanks for the clarification. I'll update this accordingly and see if I get anything new.

tay1orjones:
Just took a look at this with @sstrubberg and we discovered two things:
1. The story doesn't show up in the old storybook because files in `next` folders are now ignored via #9920
https://github.com/carbon-design-system/carbon/blob/e7b27cf0b047a7b991bb01a5fec3160facd9299a/packages/react/.storybook/main.js#L48
2. The bug can be reproduced in the new storybook - it only happens when the container has scroll.


https://user-images.githubusercontent.com/3360588/140816297-c40df8e5-5efb-4e07-9db1-8cfdd73654c7.mov



I think we probably just need to add an `event.preventDefault();` to the `handleKeyDown` for arrow keys.

joshblack (approved):

abbeyhrt (approved):

tay1orjones (approved):

PR #10062, opened 2021-11-05
feat(bidi): add Text to `BreadcrumbItem`, `DataTable`, `Dropdown`, `FormLabel`, `Link`, and `MultiSelect`

### LINKED ISSUE #9879: Update `BreadcrumbItem`, `DataTable`, `Dropdown`, `FormLabel`, `Link`, and `MultiSelect` to use `Text`
Refs https://github.com/carbon-design-system/carbon/issues/7264#issuecomment-839870634

This issue is for updating another batch of components (or just their documentation) to use the new `<Text>`, `<TextDirection>`, and `<LayoutDirection>` as was done to Accordion, Button, Checkbox, Combobox and Contentswitcher in #9493.

- [x] Breadcrumb
- [x] DataTable
- [x] Dropdown
- [x] FormLabel
- [x] Link
- [x] MultiSelect

tay1orjones:
Closes #9879

#### Changelog

**Changed**

- fix(tablebatchactions): use `Text` internally where appropriate to support bidi text
- docs(text): add storybook .mdx docs, clarify internal `Text` usage

#### Testing / Reviewing

- View stories for the associated components, there should be no regressions to style or functionality of the components


packages/react/src/components/DataTable/TableBatchActions.js
```diff
@@ -58,13 +59,13 @@ const TableBatchActions = ({
       className={batchActionsClasses}
       {...rest}>
       <div className={batchSummaryClasses}>
-        <p className={`${prefix}--batch-summary__para`}>
+        <Text as="p" className={`${prefix}--batch-summary__para`}>
```
joshblack: Should this be on the inner `<span>` instead of the outer `<p>`?

packages/react/src/components/DataTable/TableHeader.js
```diff
@@ -115,7 +116,9 @@ const TableHeader = React.forwardRef(function TableHeader(
         onClick={onClick}
         {...rest}>
         <span className={`${prefix}--table-sort__flex`}>
-          <div className={`${prefix}--table-header-label`}>{children}</div>
+          <Text as="div" className={`${prefix}--table-header-label`}>
```
joshblack: In the case of `children` being `PropTypes.node` should we worry about `Text` since the caller can localize their text?

In other words, when they give us a `string` we have to handle it correctly but if it's a `PropTypes.node` type then they should be able to handle it.

packages/react/src/components/Dropdown/Dropdown.js
```diff
@@ -142,9 +145,9 @@ const Dropdown = React.forwardRef(function Dropdown(
   return (
     <div className={wrapperClasses} {...other}>
       {titleText && (
-        <label className={titleClasses} {...getLabelProps()}>
+        <Text as="label" className={titleClasses} {...getLabelProps()}>
```
joshblack: Same question here since it can be `PropTypes.node`

packages/react/src/components/FormLabel/FormLabel.js
```diff
@@ -24,9 +25,9 @@ const FormLabel = ({ className, children, id, ...other }) => {
   );
 
   return (
-    <label htmlFor={id} className={classNames} {...other}>
+    <Text as="label" htmlFor={id} className={classNames} {...other}>
```
joshblack: Same question here since it can be `PropTypes.node`

joshblack (changes requested):
Looking great! Just had a question on `PropTypes.node` usage with `<Text>` and what you were thinking about that 👀

packages/react/src/components/DataTable/TableBatchActions.js
tay1orjones: I initially thought it would be best to place it on the element with the classname but having it on the innermost node is probably preferable. I'll change it

packages/react/src/components/DataTable/TableHeader.js
tay1orjones: Ah, yes thanks for catching this. I thought children was a string here. I searched DataTable subcomponents for common text node elements like `span`, `p`, and `div` and forgot the check the proptypes on this one. I'll remove it

packages/react/src/components/Dropdown/Dropdown.js
tay1orjones: Yep, will remove

packages/react/src/components/FormLabel/FormLabel.js
tay1orjones: Yep, will remove

dakahn (approved):
Loooks great to me 🏄🏾

packages/react/src/components/Dropdown/Dropdown.js
```diff
@@ -128,7 +129,9 @@ const Dropdown = React.forwardRef(function Dropdown(
   const ItemToElement = itemToElement;
   const toggleButtonProps = getToggleButtonProps();
   const helper = helperText ? (
-    <div className={helperClasses}>{helperText}</div>
+    <Text as="div" className={helperClasses}>
+      {helperText}
+    </Text>
```
joshblack: If I remember correctly, `helperText` could be a `PropTypes.node`, right?

packages/react/src/components/MultiSelect/MultiSelect.js
```diff
@@ -224,15 +225,15 @@ const MultiSelect = React.forwardRef(function MultiSelect(
 
   return (
     <div className={wrapperClasses}>
-      <label className={titleClasses} {...getLabelProps()}>
+      <Text as="label" className={titleClasses} {...getLabelProps()}>
         {titleText && titleText}
```
joshblack: It may be that we want the outer `<label>` and then wrap `titleText` in the `Text` component, if `titleText` is not `PropTypes.node`

joshblack (changes requested):

packages/react/src/components/Dropdown/Dropdown.js
tay1orjones: @joshblack Yeah, good point. It can also be a string. Do you think we should conditionally use a `Text` based on if it's a string or not?

tay1orjones:
@joshblack I've pivoted here to ensure that we do not wrap props of type `node` with `Text`. I've also added docs to clearly state that carbon component internals will only be using `Text` to surround props of type `string`. When props are of type `node` it's the consumers responsibility to compose `Text` appropriately within the value of the prop.

joshblack (approved):

PR #10067, opened 2021-11-08
fix(styles): update popover styles to exclude pointer events

joshblack:
This is a quickfix for popover styles. Currently, when building out things like `DefinitionTooltip` the interactive trigger does not receive mouse events. This is because popover is built in a way that covers the trigger completely in order to position the popover correctly. This results in the popover receiving all click events and not the trigger.

In this PR, we add `pointer-events: none` to the popover container so that it does not receive mouse events. We also add in `pointer-events: auto` to the popover content so that it can still receive mouse events.

#### Changelog

**New**

**Changed**

- Add `pointer-events: none` to popover
- Add `pointer-events: auto` to popover-content

**Removed**

#### Testing / Reviewing

I used the following story in `Popover.stories.js` to test:

```jsx
export const Default = () => {
  return (
    <>
      <Popover open>
        <button
          type="button"
          onClick={() => {
            console.log('click');
          }}>
          Hello
        </button>
        <PopoverContent className="p-3">Test</PopoverContent>
      </Popover>
      <p className="p-3">
        Sit tempore numquam repellendus illum soluta. Accusamus illum eveniet
        quisquam nostrum quod. Voluptatem necessitatibus deleniti quia sapiente
        quia distinctio Assumenda quisquam maxime impedit blanditiis quae
        Debitis recusandae ab aspernatur suscipit.
      </p>
    </>
  );
};
```

With this story, you can click on the button and see if the `console.log` is emitted. If it is, then this PR is working as intended. Without these changes, the `onClick` handler would never be called.


dakahn (approved):
image

Nice! Ya got it figured out. 🏄🏾

jnm2377 (approved):

PR #10087, opened 2021-11-11
fix(SideNav): add screen size check for aria-hidden value

### LINKED ISSUE #9895: [a11y]: Aria-hidden incorrectly set on sidebar
### Package

carbon-components-react

### Browser

Chrome

### Operating System

MacOS

### Package version

7.45.0

Axe Devtools

### Assistive technology

_No response_

### Description

The sidebar on desktop will have `aria-hidden="true"` even though it's open and so should be accessible

### WCAG 2.1 Violation

_No response_

### CodeSandbox example

https://codesandbox.io/s/ibm-sidenav-a11y-jo16k?file=/src/index.js

### Steps to reproduce

1. Open the codesandbox preview in a new window so that it is wide enough to show the sidenav expanded
2. Inspect the `nav` component of the sidebar and see that it has `aria-hidden="true"` but is visible

### Code of Conduct

dakahn:
Closes #9895 

The sidenav on screens greater than 1056px default to open (with the menu button hidden). The default expanded state is bugged (potentially in HeaderContainer which seems to expect isSideNavExpanded as a prop?). This adds a check that basically account for a screen width of larger than 1056 where the side nav is opened and not aria-hidden. 

I consider this a stopgap while functional component refactors are happening to the UIShell and it's components. 🏄🏾 

🎆 🎆 🎆 NOW FEATURING A USEMATCHMEDIA HOOK 🎆🎆 🎆 

#### Testing / Reviewing

check that the aria hidden value on the <nav> element inside header is properly set to false when sidenav is visible and true when it isn't 👍🏾 
image


packages/react/src/components/UIShell/SideNav.js
```diff
@@ -121,14 +121,21 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
     eventHandlers.onMouseLeave = () => handleToggle(false, false);
   }
 
+  function getAriaHiddenValue() {
+    if (expanded === false && window.innerWidth < 1056) {
+      return true;
+    }
+    return false;
+  }
+
```
joshblack: I think we'll need to do two things for this:

- Make sure we don't call `window` if the component is being SSR'd
- Make sure that this value doesn't go "stale", in other words this component should update if the width goes below 1056 or greater than 1056 so we don't get a mismatch when the browser resizes

For the first part, we can use `src/internal/environment` to get `canUseDOM` to guard against calling `window`, we can also put calls to window in a `useEffect` which don't run on the server.

This comes up for the second case where we most likely want to use `window.matchMedia`: https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

This is a browser API that lets us specify a media query and listen to changes for if the media query applies, or not. Here's a codesandbox to try and show this at play: https://codesandbox.io/s/amazing-kapitsa-6lyd5 👀

joshblack (changes requested):

packages/react/src/components/UIShell/SideNav.js
jnm2377: ah yeah, good call. Vince did something [similar](https://github.com/carbon-design-system/gatsby-theme-carbon/blob/c9a8e074a459c65ede37aea578d48c266cce3bc4/packages/gatsby-theme-carbon/src/components/ArtDirection.js#L7-L8) for `ArtDirection` component of Gatsby theme, bc it changes w/ diff breakpoints.

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -121,12 +121,9 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
 
-  function getAriaHiddenValue() {
-    if (expanded === false && window.innerWidth < 1056) {
-      return true;
-    }
-    return false;
-  }
+  const isCollapsedWidth = useMatchMedia(`(max-width: 1055px)`);
+  const ariaHidden = expanded === false && isCollapsedWidth;
+  console.log(ariaHidden);
```
joshblack:
```suggestion
```

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -121,12 +121,9 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
     eventHandlers.onMouseLeave = () => handleToggle(false, false);
   }
 
-  function getAriaHiddenValue() {
-    if (expanded === false && window.innerWidth < 1056) {
-      return true;
-    }
-    return false;
-  }
+  const isCollapsedWidth = useMatchMedia(`(max-width: 1055px)`);
```
joshblack: Would `isSideNavCollapsed` work better for this?

packages/react/src/internal/useMatchMedia.js
```diff
@@ -0,0 +1,37 @@
...
+
+    const mediaQueryList = window.matchMedia(mediaQueryString);
+    mediaQueryList.addEventListener('change', listener);
+
+    // make sure media query list and matches state are synced
+    setMatches(mediaQueryList.matches);
+
+    return () => {
+      mediaQueryList.removeEventListener('change', listener);
+    };
```
joshblack: For consistency, might help to make either `mediaQueryList` here `mql` or change `mql` on line 14 to `mediaQueryList`

packages/react/src/internal/useMatchMedia.js
```diff
@@ -0,0 +1,37 @@
...
+    if (canUseDOM) {
+      const mql = window.matchMedia(mediaQueryString);
+      return mql.matches;
+    }
+    return false;
+  });
+
+  useEffect(() => {
+    function listener() {
+      setMatches(mediaQueryList.matches);
```
joshblack:
```suggestion
    function listener(event) {
      setMatches(event.matches);
```

packages/react/src/internal/useMatchMedia.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const mediaQueryList = window.matchMedia(mediaQueryString);
+    mediaQueryList.addEventListener('change', listener);
+
+    // make sure media query list and matches state are synced
```
joshblack:
```suggestion
    // Make sure the media query list is in sync with the matches state
```

packages/react/src/internal/useMatchMedia.js
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { useState, useEffect } from 'react';
+import { canUseDOM } from '../../internal/environment';
+
+export function useMatchMedia(mediaQueryString) {
```
joshblack, line 11: If you can add tests for this, that'd be great. There's a good example from: https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom that talks about it

We basically would want to test that:

- The initial state of the hook matches if the `mediaQueryList` matches
- When `change` is fired, the hook returns a value that matches what the change event matches
- When the `mediaQueryString` is changed, the state is kept in sync so that the returned value matches the new `mediaQueryList` matches value

joshblack (changes requested):

jnm2377:
@dakahn lmk when this is ready for review :)

dakahn:
@joshblack pushed up some tests for that hook. I'm at a total loss for your second testing comment. Wouldn't testing the state is in sync be testing the onChange was calling properly for the hook? 

also it seems like eslint jest have renamed some rules in a recent update. I'm getting this error when I commit
image


Which corresponds [with this rule](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/valid-describe-callback.md) -- that it doesn't seem like i'm breaking, but I could just be missing something.

joshblack (approved):

joshblack:
@dakahn let me know if you need with the tests! Totally free this afternoon

packages/react/src/internal/useMatchMedia.js
```diff
@@ -0,0 +1,38 @@
...
+    return false;
+  });
+
+  useEffect(() => {
+    function listener(event) {
+      setMatches(event.matches);
+    }
+
+    const mediaQueryList = window.matchMedia(mediaQueryString);
+    // console.log(mediaQueryList);
```
abbeyhrt:
```suggestion
```

abbeyhrt (approved):
Looks good to me! Just a small commented out console left :)

jnm2377 (approved):

PR #10095, opened 2021-11-11
feat(react): add unstable_IconButton

### LINKED ISSUE #9967: Ship initial version of icon button based on design guidance

joshblack:
Closes #9967 

This PR adds in the `IconButton` component under `unstable` for v11

image

#### Changelog

**New**

- Add `src/components/IconButton` plus related tests and playground story

**Changed**

- Update `tooltip` to include `icon-tooltip`
- Update `tooltip` to have two new custom properties: `tooltip-padding-block` and `tooltip-padding-inline`

**Removed**

#### Testing / Reviewing

- Visit the `unstable_IconButton` story in our next storybook and verify that the component works and looks as expected
- Look at the Component API and double-check that it makes sense and matches our API guidelines
- Make sure tests pass as expected


joshblack:
@laurenmrice this should show up under the next storybook under the `unstable_IconButton` story! Let me know if you have any questions 👀

tay1orjones (approved):
Looks really great! I'm excited to see one less dependency in the button styles🏅

joshblack:
bump @jnm2377 @laurenmrice when you get a sec 👀

jnm2377 (approved):
Sorry for the delayed review. This looks 🔥

laurenmrice (changes requested):
**Tooltip sizing**
- [x] The caret for the icon button tooltip should be 8px x 4px in size.
- [x] There should be 2px padding on the top and the bottom of the text inside of the tooltip
Artboard

-----
**For `align`**
Take out alignments for:
- [x] left-bottom
- [x] left-top
- [x] right-bottom
- [x] right-top

**For `kind`**
The following buttons do not come as icon only buttons and should be removed from the dropdown:
- [x] danger
- [x] danger--primary
- [x] danger--ghost
- [x] danger--tertiary

aagonzales (changes requested):
Looks like the height of the tooltip container is 23px not 22.
image

Also the tooltip hangs around for a while after mousing away, is that intentional? I could see problems with that is there are multiple icons and a row.
Nov-19-2021 12-34-32

joshblack:
Thanks @aagonzales!

For the delay, there is a delay for entrance and exit that is a part of the spec. These are also configurable values as props so teams can have set the exit delay to 0 for that classic icon group approach 👀  Another option on our end could be to implement such that only one tooltip is visible at a time.

For the height, I just updated it to use `body-short-01` which should address the issue 👍

joshblack:
@aagonzales let me know if you have any more questions!

aagonzales (approved):
I think its fine from the comments I posted. Not sure if @laurenmrice had any other.

laurenmrice (approved):
🚀

Issue #2343, opened 2019-04-23
Form spacing inconsistencies 


elizabethsjudd:
## Detailed description

> Describe in detail the issue you're having.

Fieldsets and form items should have consistent spacing within a form.  Currently Fieldsets have a margin-bottom of `$carbon--spacing-07` (https://github.com/IBM/carbon-components/blob/master/src/components/form/_form.scss#L20) but form-item doesn't have any margin-bottom meaning we have to manually add it.  The demos on the Carbon site have a manual 2rem `margin-bottom` added to form-item.

Secondly according to the guidelines, the spacing should be `24px/1.5rem` which would be `$carbon--spacing-06`.

> Is this issue related to a specific component?

Form

> What did you expect to happen? What happened instead? What would you like to see changed?

Form items and fieldsets have consistent spacing (either they both have it or neither have it) and follow the recommended guidelines outlined on the Carbon site. https://www.carbondesignsystem.com/components/form/style#recommended

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

v10.1.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Watson Health Design Pattern & Asset Library (PAL)

## Steps to reproduce the issue

1. Inspect the form demo on the carbon site.
1. note the margin bottom is not part of the component styles
1. note the margin bottom value does not align with the recommended values

asudoh:
@emyarod Would you be able to look into this? Thanks!

emyarod:
hi @elizabethsjudd thanks for reporting this! IIRC the v9 `.bx--form-item` class had a `2rem` bottom margin applied to it, but [this rule was never brought to v10](https://github.com/IBM/carbon-components/commit/b8bf8b85639c57fe7336fe6c11c939aa6e22371c#diff-4e8ca57486af317304770962b8ed1fc9).

[...]

PR #10100, opened 2021-11-12
docs(form): use stack for form spacing examples, fix label spacing

### LINKED ISSUE #9948: Create Stack usage examples/documentation for components with removed margin
For the components that had external margin removed in https://github.com/carbon-design-system/carbon/pull/9830 we should 

- [x]  Update their stories to use the Stack component 
   - https://github.com/carbon-design-system/carbon/pull/9876
- [x] Write migration guidance for these components (to use Stack)
- [x] The website needs to be updated as well to use the Stack, particularly the Form examples where we have a `2rem` margin hardcoded

Related https://github.com/carbon-design-system/carbon/issues/9947

Should resolve https://github.com/carbon-design-system/carbon/issues/2343

tay1orjones:
Part of #9948 
Refs #2343
Also fixes some incorrect styles for #9947 I made in #10028. I also forgot to mirror the changes into `@carbon/styles`, so I've done that here.

Use a `Stack` in the `Form` story/example in the `carbon-react-next` storybook.

#### Changelog

**New**

- style(form): include margin on labels and new `.#{$prefix}--label--no-margin`

**Changed**

- style(form): include label spacing by default, add `--no-margin` class suffix that was missed
- docs(form): use `Stack` for form spacing example

#### Testing / Reviewing

- View the `carbon-react-next` storybook, the `Form/Default` story.
- Form items should be properly spaced. @carbon-design-system/design could you double check me on this?


aagonzales:
Not sure if this is because of the stack or not but I noticed a few issues inside of the field sets.

Like with the search field being too small 
image

And for some reason I feel like there's too much padding after the labels in toggle and file uploader
image

tay1orjones:
Thanks @aagonzales! I've updated the PR to fix those. Search now has the new `size` prop, the labels have unified spacing and I improved the spacing between toggle buttons too

aagonzales (approved):
Looks good!

dakahn (approved):

PR #10108, opened 2021-11-12
fix(TextInput): only add aria-describedby when there's helper to link to

### LINKED ISSUE #10107: [Bug]: TextInput with no helper text uses invalid aria-describedby
### Package

carbon-components-react

### Package version

7.47.2

### Description

With #9873, created by @dakahn to address #9630, an `aria-describedby` has been added to the TextInput generated DOM, set to the id of a helper text DIV. However, the helper text DIV is only generated *if* the `helperText` prop has been set. This means that when the `helperText` prop is NOT set the generated DOM includes an `aria-describedby` which contains an id that doesn't exist.

When `aria-describedby` contains an id that doesn't exist, this is an error, and causes accessibility testing tools to report violations whenever TextInput components without `helperText` set are used. For example, `accessibility-checker` v3.1.11 reports violations as follows:
image
The above example is from the ExportModal test suite of @carbon/ibm-cloud-cognitive. This test has been temporarily skipped in order to allow us to continue work until this issue is fixed.

### Code of Conduct

dcwarwick:
Closes #10107 

The `aria-describedby` should only be set when there is a helper text, since otherwise the id it is being set to cannot be found. This change ensures it is only set when `helperText` is set, which is also when the helper text is generated.

#### Changelog

**Changed**

- TextInput.js

#### Testing / Reviewing

I made this exact change manually to the instance of carbon-components-react used by @carbon/ibm-cloud-cognitive, and ran the test suite for TextInput, and the accessibility failure that was previously being reported was no longer reported.


joshblack (approved):

dakahn (approved):
Thank you so much!

PR #10120, opened 2021-11-16
fix(tabs): debounce scroll handler

### LINKED ISSUE #9698: [Bug]: Tabs do not have left scroll button when scrolled to the right side
### Package

### Browser

Safari

### Package version

10.43.0, 7.43.0

### Description

When a user scrolls tabs the left scroll arrow does not appear unless a resize event happens to the page (when the scroll position for the tabs is not 0 on the left).  

I found that using older React (like Sandbox defaulted to) worked but would throw a ton of errors. Using latest react (17.0.2) did not have errors but did not show the arrow unless I resized.

See recording here:
example3

Project reference: IBM Aspera

### CodeSandbox example

https://codesandbox.io/s/sleepy-ardinghelli-y8pry

### Steps to reproduce

Scroll list with either the right arrow or a trackpad. The left arrow does not appear when content goes off screen.  However, a resize event of window will trigger it to appear.  

Expected: The arrow appears as soon as Tabs container is not left scrolled at 0.

### Code of Conduct

tay1orjones:
Closes #9698 

Updates Tabs to debounce the scroll handler. 

@carbon-design-system/release this will need to be cut into a new patch release for v10.48. I'm more than happy to assist!

#### Changelog

**Changed**

- tabs: debounce scroll handler

#### Testing / Reviewing

- The original issue was only present in Safari. 
- In Safari, Check the `tabs` stories in the `carbon-components-react` storybook. 
- Scroll back and forth multiple times (10-20 times) and ensure that React does not throw an error; `Error: Maximum update depth exceeded.`

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.js
+++ b/packages/react/src/components/Tabs/Tabs.js
@@ -153,6 +153,15 @@ export default class Tabs extends React.Component {
 
   _handleWindowResize = this.handleScroll;
 
+  /**
+   * The debounced version of the `scroll` event handler.
+   * @type {Function}
+   * @private
+   */
+  _debouncedHandleScroll = null;
+
+  _handleScroll = this.handleScroll;
+
   componentDidMount() {
     if (!this._debouncedHandleWindowResize) {
       this._debouncedHandleWindowResize = debounce(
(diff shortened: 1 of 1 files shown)


tay1orjones:
@dabrad26 thanks for the steps to repro the original issue in Safari! Let me know what you think on this one, I believe it should solve the issue. 👍

sstrubberg (approved):
🚀

joshblack (approved):

jnm2377 (approved):

PR #10134, opened 2021-11-17
chore(project): update prettier to 2.4.1

joshblack:
This PR updates our prettier dependency to 2.4.1

#### Changelog

**New**

**Changed**

- Update `package.json` to use latest prettier
- Update `prettier-config-carbon` to use latest prettier
- Use `bracketSameLine` instead of `jsxBracketSameLine`
- Run format on files

**Removed**

#### Testing / Reviewing

- Verify formatting changes are helpful, sometimes prettier misformats sass files and it can make things confusing


joshblack:
Actually, it looks like 2.4.1 breaks with `@forward '...' with ()` in Sass Modules so we'll need to wait for 2.5.0 which has the fix 👀

PR #10140, opened 2021-11-18
refactor(popover): update <div> to <span>

joshblack:
This PR refactors our popover to use `<span>` instead of `<div>` so that popovers may be used in inline tags like `<p>` for components like `DefinitionTooltip`. It also adds in an e2e test for VRT.

#### Changelog

**New**

- Add e2e test for Popover

**Changed**

- Change all tags to `<span>` for Popover

**Removed**

#### Testing / Reviewing

- Verify Percy has picked up the new VRT for Popover
- Verify Popover still renders as expected across browsers


jnm2377 (approved):
LGTM. Popover is working as expected. 👍🏽

jnm2377:
oh, just noticed Percy shows a change in the snapshot, but I can't actually tell what has changed... 🤔 It looks right to me..

aledavila (approved):
lgtm. Percy just says new snapshot not even a changed one

dakahn (approved):
Nice! Good ol reliable spans to the rescue 👍🏾

PR #10147, opened 2021-11-19
chore(OverflowMenuItem): refactored to functional component

sstrubberg:
REF #10022


abbeyhrt:
@sstrubberg this isn't from your changes, it's in https://carbon-react-next.netlify.app/ and this PR's deploy preview but it looks like OverflowMenu is a bit broken right now (I can't access the item with a keyboard either)
Screen Shot 2021-11-30 at 10 53 26 AM
.

packages/react/src/components/OverflowMenuItem/next/OverflowMenuItem.js
```diff
@@ -0,0 +1,190 @@
...
+    closeMenu,
+    disabled = false,
+    handleOverflowMenuItemFocus,
+    hasDivider = false,
+    href,
+    isDelete = false,
+    index,
+    itemText = FeatureFlags.enabled('enable-v11-release')
+      ? null
+      : 'Provide itemText',
```
jnm2377: I think we could get rid of this conditional here, since `/next` stuff is always for v11

jnm2377:
Looks good! just had one comment :)

jnm2377 (approved):

jnm2377:
@abbeyhrt I can take care of the `OverflowMenu` component in a different PR.

abbeyhrt:
It looks like there is some funny business with the mixed class `OverflowMenu` and the function `OverflowMenuItem`. Locally running `yarn start:v11` , I can't use arrow keys to get to the items, seemingly because `handleOverflowMenuItemFocus` is undefined. It's supposed to be defined coming from `OverFlowMenu` cloning the children and adding that prop but I don't think that it works the same now that the children aren't classes.

abbeyhrt (approved):
`OverflowMenuItem` looks good to me! Just some weirdness with props being defined from the parent that should be resolved when OverflowMenu is refactored

PR #10151, opened 2021-11-19
fix(inlinecheckbox): do not propagate click event from label

### LINKED ISSUE #9577: Not able to prevent click event propagation on Checkbox/TableSelectRow

tay1orjones:
Closes #9577 

Ensures that the InlineCheckbox only allows click events to propagate from the `input` (via `onSelect`) and not the `label`.

#### Changelog

**Changed**

- inlinecheckbox: do not propagate click event from label

#### Testing / Reviewing

- View the `carbon-components-react` storybook, the DataTable/Sorting/Usage story. 
- Click a checkbox to select a table row. In the actions pane only one `onClick` event should fire w/ the `input` as the target.
- If you pull down this PR and remove my `stopPropagation` call, you can view the original bug - `onClick` fires twice. Once from the `label` and once from the `input`.


tay1orjones:
cc @nelsonchen90

joshblack (approved):

dakahn (approved):

PR #10187, opened 2021-11-29
feat(table-toolbar-search): add onBlur/onFocus handling

### LINKED ISSUE #10077: [TableToolbarSearch]: handle passed onFocus/onBlur events without breaking expand behavior
### Summary

We would like to trigger a search callback to update the table data when the search field is blurred, however, this breaks the automatic expand/contract behavior of the search input when either of these events are passed. It would be nice if the internal onFocus/onBlur would perform their actions and then call the external onFocus/onBlur callbacks afterward.

Example reproduction:
https://codesandbox.io/s/weathered-feather-mh35x?file=/src/index.js

1. Click on the Search icon
2. click outside of the search box

Expected: The search box should collapse

### Justification

_No response_

_No response_

### Required functionality

_No response_

_No response_

### Available extra resources

Yes, I will happily write the PR if this behavior is acceptable.

### Code of Conduct

kevinsperrine:
Closes #10077

add handling to onBlur/onFocus props passed to the TableToolbarSearch. If they are provided, they are called with the event and the handleExpand function as parameters. It's then up to the consumer to call handleExpand to expand/contract the input as desired. Existing automatic expand/contract is maintained if no onBlur/onFocus events are passed.

#### Changelog

**New**

- onBlur prop supported for TableToolbarSearch
- onFocus prop supported for TableToolbarSearch
- added tests to confirm

#### Testing / Reviewing

- added onFocus/onBlur actions to the data-table with toolbar story.
- unit tests should pass


packages/react/src/components/DataTable/TableToolbarSearch.js
```diff
@@ -96,6 +98,9 @@ const TableToolbarSearch = ({
     }
   };
 
+  const internalOnFocus = (event) => handleExpand(event, true);
```
abbeyhrt:
```suggestion
  const handleOnFocus = (event) => handleExpand(event, true);
```

We are working towards a standard of internal functions using `handleOn...()` with the prop for that functionality being `on...()`, like you've already done here :)

packages/react/src/components/DataTable/TableToolbarSearch.js
```diff
@@ -96,6 +98,9 @@ const TableToolbarSearch = ({
...
+  const internalOnBlur = (event) => !value && handleExpand(event, false);
```
abbeyhrt:
```suggestion
  const handleOnBlur = (event) => !value && handleExpand(event, false);
```

Same as above

packages/react/src/components/DataTable/TableToolbarSearch.js
```diff
@@ -110,8 +115,10 @@ const TableToolbarSearch = ({
       }
       onChange={onChange}
       onClear={onClear}
-      onFocus={(event) => handleExpand(event, true)}
-      onBlur={(event) => !value && handleExpand(event, false)}
+      onFocus={
+        onFocus ? (event) => onFocus(event, handleExpand) : internalOnFocus
```
abbeyhrt:
```suggestion
        onFocus ? (event) => onFocus(event, handleExpand) : handleOnFocus
```

packages/react/src/components/DataTable/TableToolbarSearch.js
```diff
@@ -110,8 +115,10 @@ const TableToolbarSearch = ({
...
+      }
+      onBlur={onBlur ? (event) => onBlur(event, handleExpand) : internalOnBlur}
```
abbeyhrt:
```suggestion
      onBlur={onBlur ? (event) => onBlur(event, handleExpand) : handleOnBlur}
```

abbeyhrt (changes requested):
Looks good! Just a couple small changes to fit with our code practices

packages/react/src/components/DataTable/TableToolbarSearch.js
kevinsperrine: Thanks, @abbeyhrt. Updated.

abbeyhrt (approved):
Looks good to me! Thanks for contributing!

jnm2377 (changes requested):
When I look at the `with Toolbar` story, the search is no longer expanding 😓

kevinsperrine:
> When I look at the `with Toolbar` story, the search is no longer expanding 😓

Oh, weird! When I added the bit to include the arguments in the story action, it breaks and causes a cross-origin error in the deploy preview. I'll remove that bit and push it up.

jnm2377 (approved):

jnm2377:
Thanks for the contribution! 🎉

PR #10189, opened 2021-11-29
refactor(styles): remove unused tabs styles

abbeyhrt:
Trying to break up my Tabs PR. This PR removes the styles that are not being used in next/Tabs. Future PR will switch class names from `cds--tabs--scrollable` to `cds--tabs` . 

 `cds--tabs--trigger` is only relevant to vanilla so they can be safely removed for v11 in `@carbon/styles`

#### Changelog

**Removed**

- lots of styles. 
 
#### Testing / Reviewing

Check next story and make sure that Tabs looks the same as before.


packages/styles/scss/components/tabs/_tabs.scss
```diff
@@ -831,13 +442,6 @@
       color: $text-on-color-disabled;
     }
 
-    //-----------------------------
-    //  Tab Content Container
-    //-----------------------------
-    .#{$prefix}--tab-content {
-      padding: $spacing-05;
-    }
-
```
abbeyhrt, line 840: Just as an FYI to reviewers, `tab-content` is still a relevant class name but these styles can be removed because they are a duplicate.

tay1orjones (approved):

aledavila (approved):
Lgtm !

Issue #8487, opened 2021-04-22
Tab Updates and enhancements


kingtraceyj:
We are updating the usage docs to account for more use cases.
### Improvement on existing tabs

1. Name change (Default tabs we are now calling line tabs and container tabs we are now calling contained tabs)
2. Updating default functionality (We are recommending line tabs use auto-spacing and contained tabs can use either auto-spacing or grid)

### Adding modifiers
**Line tabs**

- Icon-only
- Label with icon

**Contained tabs**
- Icon-only
- Label with icon
- Label with secondary label
- Label with secondary label and icon

**Adding in motion**
- TBD - new issue will come for this

_See below more more details._

Pieces delivered in v11:
- [x] #10086
- [x] https://github.com/carbon-design-system/carbon/issues/9952
- [x] https://github.com/carbon-design-system/carbon/issues/9988
- [x] https://github.com/carbon-design-system/carbon/issues/9991
- [x] #10031
- [x] #10249
- ~Can/should LineTab(s) be separate components as well so `Tabs` can be a common primitive for both contained and line implementations?~

```[tasklist]
### Tasks
- [x] https://github.com/carbon-design-system/carbon/issues/11766
- [ ] https://github.com/carbon-design-system/carbon/issues/12844
- [ ] https://github.com/carbon-design-system/carbon/issues/12843
- [ ] https://github.com/carbon-design-system/carbon/issues/12813
- [ ] https://github.com/carbon-design-system/carbon/issues/12845
- [x] https://github.com/carbon-design-system/carbon/issues/12846
- [x] https://github.com/carbon-design-system/carbon/issues/12848
- [ ] https://github.com/carbon-design-system/carbon/issues/13351
- [ ] https://github.com/carbon-design-system/carbon/issues/13353
- [ ] https://github.com/carbon-design-system/carbon/issues/13389
- [ ] https://github.com/carbon-design-system/carbon-website/issues/2660
- [ ] https://github.com/carbon-design-system/carbon/issues/13758
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/682
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3556
```

kingtraceyj:
### Name update
In docs we will now refer to tabs as line tabs and container tabs as contained tabs. We're getting rid of the work default since we use it differently through our components.
 

### Default functionality
We now recommend using auto-spacing for line tabs. Contained tabs can use both auto-spacing or the grid.

**Line:**
[...]

PR #10194, opened 2021-11-30
fix(styles): replace cds--tab--scrollable class with cds--tabs

abbeyhrt:
Ref: #10086, #8487

Switches `cds-tabs--scrollable--**` to `cds-tabs--**` and creates a next version of TabsSkeleton. I know that class names aren't technically a breaking change but i went ahead and separated it since we are already exporting `/next` versions of tabs.  

#### Changelog
**New** 

- TabsSkeleton in `/next`

**Changed**

- Minor story changes to import to ensure that the right component was being used. 
- reordered Tabs styles 
- replaced `cds--tabs--scrollable` with `cds--tabs`

**Removed**

- outdated comments

#### Testing / Reviewing

Check to make sure that Tabs looks the same as before and that `cds--tabs--scrollable` is replaced everywhere but only for `@carbon/react`


aledavila (approved):
lgtm tabs look the same

tay1orjones (approved):

PR #10197, opened 2021-11-30
feat(DefinitionTooltip): add DefinitionTooltip

dakahn:
This PR adds the DefinitionTooltip component (utilizing our Popover primitive 🍖) and DefinitionTooltip styles ⚡ 
image

#### Testing / Reviewing

In the `carbon-react` package run the storybook and check out the unstable_tooltip > Definition story. Make sure it matches intended spec and behavior. 🏄🏾


dakahn:
@kingtraceyj @ing you for design review. Feel free to unassign yourself and tag in whoever if that's not right

packages/react/src/components/Tooltip/next/DefinitionTooltip.js
```diff
@@ -0,0 +1,62 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import { Popover, PopoverContent } from '../../Popover';
+import { match, keys } from '../../../internal/keyboard';
+import { usePrefix } from '../../../internal/usePrefix';
+
+const DefinitionTooltip = ({ children, definition }) => {
```
joshblack: Make sure to update this to a function declaration per our style guide

packages/react/src/components/Tooltip/next/DefinitionTooltip.js
```diff
@@ -0,0 +1,62 @@
...
+        onMouseOver={() => setOpen(!isOpen)}
+        onKeyDown={handleKeyDown}
+        onClick={() => {
+          setOpen(!isOpen);
+        }}>
+        {children}
+      </button>
+      <PopoverContent
+        className={`${prefix}--definition-tooltip`}
+        id="definition-id">
```
joshblack: We should use the `useId()` hook so this is dynamic, otherwise we'll have conflicting ids

packages/react/src/components/Tooltip/next/DefinitionTooltip.js
```diff
@@ -0,0 +1,62 @@
...
+import { match, keys } from '../../../internal/keyboard';
+import { usePrefix } from '../../../internal/usePrefix';
+
+const DefinitionTooltip = ({ children, definition }) => {
+  const [isOpen, setOpen] = useState(false);
+  const prefix = usePrefix();
+
+  function handleKeyDown(event) {
+    if (match(event, keys.Escape) && isOpen) {
+      setOpen(false);
```
joshblack, line 23: We should `event.stopPropagation()`, I think, so that this doesn't close an outer container like a Modal

packages/react/src/components/Tooltip/next/DefinitionTooltip.js
joshblack: I think we'll also want to add support for `...rest` on the inner button `<button>`

packages/react/src/components/Tooltip/next/Tooltip.stories.js
```diff
@@ -68,6 +69,22 @@ export const Duration = () => {
...
+      <p>
+        The first duty of every{' '}
+        <DefinitionTooltip definition={definition}>Starfleet</DefinitionTooltip>{' '}
+        officer is to the truth, whether it’s scientific truth, or historical
+        truth, or personal truth! It is the guiding principle on which Starfleet
+        is based.
+      </p>
+    </div>
+  );
+};
```
joshblack: We should try and use example copy that's consistent across docs 👀

packages/styles/scss/components/tooltip/_tooltip.scss
```diff
@@ -37,6 +39,27 @@ $tooltip-padding-inline: custom-property.get-var(
...
+
+    &:focus {
+      @include focus-outline.focus-outline;
+
+      border-bottom-color: theme.$border-interactive;
+      border-radius: 0;
+    }
+
+    border-bottom: 1px dotted theme.$border-strong;
+  }
```
joshblack:
```suggestion
  .#{$prefix}--definition-term {
    @include button-reset.reset;

    border-bottom: 1px dotted theme.$border-strong;
    border-radius: 0;
  }
  
  .#{$prefix}--definition-term:focus {
    @include focus-outline.focus-outline;

    border-bottom-color: theme.$border-interactive;
  }
```

joshblack (changes requested):
Left a couple of comments! Could you also add a test file for this using RTL?

tay1orjones:
> Make sure it matches intended spec and behavior.

I'm not sure exactly where to find the spec (has it changed from the way it is in v10?), but in this video I went through three things and the last stood out to me as odd.

1. Focus via keyboard <kbd>tab</kbd>, works as expected ✅ 
2. Hover via mouse, works as expected ✅ 
3. Click - it focuses the trigger, but the tooltip goes away. Then I click focus off the trigger, and the tooltip reappears. I frantically click around and can't get the tooltip to dismiss. I hover over the trigger again and notice that I've discovered some type of reverse-hover situation where hovering on the trigger dismisses the tooltip. I click the trigger again, it's focused, the tooltip appears, and then it's all good again.


https://user-images.githubusercontent.com/3360588/145135228-a31ad801-2056-4e89-bdcc-fea34efd3cb9.mov

packages/react/src/components/Tooltip/next/Tooltip.stories.js
dakahn: Star Wars guy, huh?

dakahn: Grabbed the copy from Modal -- lemme know if there's something more apropos

dakahn:
@tay1orjones thanks for the QA pass, so if i'm reading right we have three bugs:
1. tooltip should dismiss on blur
2. tooltip when shown, shouldn't hide when trigger is hovered with the mouse
3. if the tooltip is shown, clicking the trigger should close the tip

packages/react/src/components/Tooltip/next/DefinitionTooltip.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  function handleOnClick() {
+    if (isOpen) {
+      setOpen(false);
+      return;
+    }
+    setOpen(true);
```
joshblack:
```suggestion
    setOpen(!isOpen);
```

packages/styles/scss/components/tooltip/_tooltip.scss
```diff
@@ -11,6 +11,8 @@
 @use '../../type';
 @use '../../utilities/custom-property';
 @use '../../utilities/convert';
+@use '../../utilities/_button-reset';
+@use '../../utilities/_focus-outline';
```
joshblack:
```suggestion
@use '../../utilities/button-reset';
@use '../../utilities/focus-outline';
```

packages/styles/scss/components/tooltip/_tooltip.scss
```diff
@@ -37,6 +39,27 @@ $tooltip-padding-inline: custom-property.get-var(
...
+    border-bottom: 1px dotted theme.$border-strong;
+    border-radius: 0;
+  }
+  
+  .#{$prefix}--definition-term:focus {
+    @include focus-outline.focus-outline;
+
+    border-bottom-color: theme.$border-interactive;
+  }
+  .#{$prefix}--definition-tooltip {
```
joshblack:
```suggestion
  }

  .#{$prefix}--definition-tooltip {
```

joshblack (changes requested):

joshblack (approved):

tay1orjones (approved):

dakahn:
merged without a design review -- so reverted that merge. i'll reopen 😮‍💨

PR #10198, opened 2021-11-30
feat(react): refactor filterable multi select to function

### LINKED ISSUE #10023: [FilterableMultiSelect]: upgrade to functional component

abbeyhrt:
Closes #10023 

Refactors FilterableMultiSelect to a functional component. 

#### Changelog

**Changed**

- updated FilterableMultiselect/index.js so that it respects the v11 flag
- fixes minor token bug where hovered and active item wasn't a different color from hovered items

#### Testing / Reviewing

Check that FilterableMultiSelect works the same as before.


packages/react/src/components/MultiSelect/next/FilterableMultiSelect.js
```diff
@@ -0,0 +1,599 @@
...
+  const menuId = `${id}__menu`;
+  const inputId = `${id}-input`;
+
+  function handleOnChange(changes) {
+    if (onChange) {
+      onChange(changes);
+    }
+  }
+
+  /// look into this weirdness
```
tay1orjones: Assuming there's no more weirdness 😄 

```suggestion
```

dakahn (approved):
Looks great!!! 🏄🏾

tay1orjones (approved):

packages/react/src/components/MultiSelect/next/__tests__/FilterableMultiSelect-test.js
```diff
@@ -0,0 +1,168 @@
...
+const openMenu = (wrapper) => {
+  wrapper.find(`[role="combobox"]`).simulate('click');
+};
+
+describe('FilterableMultiSelect', () => {
+  let mockProps;
+
+  beforeEach(() => {
+    // jest.mock('../../../internal/deprecateFieldOnObject');
+
```
joshblack:
```suggestion
```

Can remove if no longer needed

packages/react/src/components/MultiSelect/next/FilterableMultiSelect.js
```diff
@@ -0,0 +1,598 @@
...
+  const [highlightedIndex, setHighlightedIndex] = useState(null);
+  const textInput = useRef();
+  const filterableMultiSelectInstanceId = useId();
+
+  const context = FeatureFlagContext;
+  let enabled;
+
+  if (context.enabled) {
+    enabled = context.enabled('enable-v11-release');
+  }
```
joshblack: We can update this to:

```suggestion
  const enabled = useFeatureFlag('enable-v11-release');
```

packages/react/src/components/MultiSelect/next/FilterableMultiSelect.js
```diff
@@ -0,0 +1,598 @@
...
+  if (prevOpen !== open) {
+    setIsOpen(open);
+    setPrevOpen(open);
+  }
+
+  const inline = type === 'inline';
+  const showWarning = !invalid && warn;
+
+  // needs to be capitalized for react to render it correctly
+  const ItemToElement = itemToElement;
```
joshblack: We should do this where we are listing props above like:

```js
  itemToElement: ItemToElement
```

packages/react/src/components/MultiSelect/next/FilterableMultiSelect.js
```diff
@@ -0,0 +1,598 @@
...
+
+  function clearInputValue() {
+    setInputValue('');
+  }
+
+  useEffect(() => {
+    if (textInput.current && inputValue == '') {
+      textInput.current.focus();
+    }
+  }, [inputValue]);
```
joshblack: I think we might get in trouble with this effect based on input value. Is there a place where we are "resetting" the input value that we are trying to hook into so that we can focus the text input? I think we should make sure this happens there instead of in an effect.

This way, the focus will be the result of a user action instead of performed by React when `inputValue` changes.

joshblack (changes requested):

packages/react/src/components/MultiSelect/next/FilterableMultiSelect.js
abbeyhrt: I updated `clearInputValue` to also reset focus to the textInput and removed the effect, let me know if that works!

joshblack (approved):

PR #10205, opened 2021-12-01
feat(styles): add support for IBM Plex from the `@ibm/plex` package

joshblack:
This PR adds in support for bringing in IBM Plex to Carbon through `scss/fonts`. This update allows teams to bundle the latest version of `@ibm/plex` into their project, in particular for bundlers like webpack. It also allows teams that self-host to use our system to correctly determine `url()` values for `@font-face` blocks.

Overall, this PR by default allows IBM Plex Mono, Sans, Serif to be included with the light, regular, and semibold font weights. It also allows teams to include other font weights and other font families such as:

- IBM Plex Sans Arabic
- IBM Plex Sans Devanagari
- IBM Plex Sans Hebrew
- IBM Plex Sans Thai
- IBM Plex Sans Thai Looped

This PR **does not** include the following:

- IBM Plex Sans JP
- IBM Plex Sans KR

#### Changelog

**New**

- Add `scss/fonts` entrypoint to `@carbon/styles`

**Changed**

- Update how fonts are brought in with `@carbon/styles`
- Update how storybook brings in fonts in `carbon-react`

**Removed**

- Removed `scss/_font-face.scss`

#### Testing / Reviewing

- `packages/styles/scss/fonts`
  - Review the implementation and see if it makes sense to you
  - Make sure implementation is tested and the tests pass
  - Review the `scss/fonts/README.md` file and make content suggestions where appropriate if something is unclear or missing from the docs
- `packages/carbon-react`
  - Verify that all files from `packages/styles/scss/fonts` are correctly re-exported in `packages/carbon-react/scss/fonts`
  - In the storybook, visit the Elements / IBM Plex story and do the following:
    - Verify each font loads as a network resource
    - Verify fonts with unicode ranges only load the relevant unicode range
    - There's a video below of what this would look like

https://user-images.githubusercontent.com/3901764/144306731-2327256c-2f97-40ca-a34d-8096b84a9032.mov

At the end, you'll see that I noticed that IBM Plex Sans Thai was loaded incorrectly and didn't not have the correct font resource loaded


joshblack:
FYI @vpicone

packages/styles/scss/fonts/unicode/_index.scss
```diff
@@ -0,0 +1,124 @@
...
+
+/// All available unicode ranges where the keys are unicode range names and the
+/// value is the list of unicode ranges that are applicable
+/// @type {List}
+$ranges: (
+  Cyrillic: (
+    'U+0400-045F',
+    'U+0472-0473',
+    'U+0490-049D',
+    'U+04A0-04A5',
```
tay1orjones, line 18: Using the unicode ranges to chunk these is super cool! A few questions though:
- Where do these ranges come from?
- How do we validate that they're correct and we haven't missed any and/or that the font file actually has a glyph for all these? 

I assume in some way we're probably just going off of what's in the font files from Bold Monday and trusting what they've defined as Cyrillic, but I'm curious if there's more going on there?

packages/styles/scss/fonts/README.md
```diff
@@ -0,0 +1,134 @@
...
+Carbon:
+
+| Font                      | Available | Entrypoint                    |
+| ------------------------- | --------- | ----------------------------- |
+| IBM Plex Mono             | ✅        | `scss/fonts/mono`             |
+| IBM Plex Sans Arabic      | ✅        | `scss/fonts/sans-arabic`      |
+| IBM Plex Sans Devanagari  | ✅        | `scss/fonts/sans-devanagari`  |
+| IBM Plex Sans Hebrew      | ✅        | `scss/fonts/sans-hebrew`      |
+| IBM Plex Sans KR          |           |                               |
+| IBM Plex Sans JP          |           |                               |
```
tay1orjones, line 14: Are these left out for now just based on how large the file size is?

packages/styles/scss/fonts/README.md
```diff
@@ -0,0 +1,134 @@
...
+You can also completely override the default behavior and provide your own URL
+resolver through `scss/fonts/src`:
+
+```scss
+@use 'sass:meta';
+@use './my-custom-resolver' as resolver;
+@use '@carbon/styles/scss/fonts/src' with (
+  // Here, "resolve" is the name of the function in the "resolver" module
+  $resolver: meta.get-function('resolve', 'resolver')
+);
```
tay1orjones, line 105: It's awesome that you were able to figure this out. I see below that you use `meta.call()` to call it, very cool! TIL higher order functions are possible in sass 🎉

tay1orjones (approved):
This is really great work, I'm super impressed! Huge value in this one. 

Just a couple questions below but ultimately LGTM 👍

packages/styles/scss/fonts/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```scss
+@use '@carbon/styles/scss/fonts' with (
+  $fonts: (
+    IBM-Plex-Sans-Arabic: true,
+  ),
+);
+```
+
+You can also configure it to disable specific fonts:
```
aledavila, line 32: If you disable a font here do you need to enable another one by default or will the other 2 default ones just take precedent ?

packages/styles/scss/fonts/README.md
```diff
@@ -0,0 +1,134 @@
...
+@use '@carbon/styles/scss/config' with (
+  $css-font-face: false,
+);
+```
+
+## Features
+
+Each font is available as an entrypoint in the `fonts` folder. You can use these
+entrypoints to include specific font weights, styles, and more for IBM Plex. For
+example, if you only want to include the regular font weight for IBM Plex Sans
```
aledavila, line 55: Should we surface the available font weights here?

aledavila (approved):
looks great josh! all tests pass and the implemention is really clean. Thanks for going over this earlier

packages/styles/scss/fonts/unicode/_index.scss
joshblack: @tay1orjones good question, these splits come from Bold Monday if I understand correctly and are ones determined by Google for Google Fonts

packages/styles/scss/fonts/README.md
joshblack: @tay1orjones we ran into two problems when implementing:

- JP was not included in the current Plex beta
- These fonts support hinted and unhinted variants which complicated an already pretty complicated resolver logic lol, felt better to add later instead of all in one
- Like you said, the font files and resulting package can be HUGE

packages/styles/scss/fonts/README.md
joshblack: @aledavila great question, basically there are the three fonts that are enabled by default. If you disable one of them, the others will still be enabled.

The trick for this is in the `fonts/_index.scss` file that merges our "default" fonts with the given `$fonts` map:

```scss
$-default-fonts: (
  IBM-Plex-Sans: true,
  IBM-Plex-Serif: true,
);

// This is the variable that callers would change with `@use`
$fonts: () !default;

// Merge the two maps, prefer values in what the user gives us over our default values
$fonts: map.merge($-default-fonts, $fonts);
```

packages/styles/scss/fonts/README.md
joshblack: Good call, definitely 👍

abbeyhrt (approved):
Looks great! Very cool to see, I've never thought about fonts before and it's awesome to officially have this in carbon now!

packages/styles/scss/fonts/README.md
aledavila: Oh cool. Thank !

PR #10207, opened 2021-12-01
chore(release): v10.49.0-rc.0

dakahn:
Release PR for vX.Y.Z-rc.0


joshblack (changes requested):
It seems like we're adding some extra packages to our yarn cache that aren't supposed to be there. Could it be that the new `www/package.json` hasn't been updated to the rc version of `@carbon/react`?

PR #10223, opened 2021-12-03
refactor(react): move all next stories to react package

joshblack:
This PR updates our project conventions to use `next` folders in the `react` package instead of stories in the `carbon-react` folder. It also moves all new components into `react`, as well, so that they may rely on internals freely instead of having to cross package boundaries.

This change also removes the `start:v11` task as it is no longer needed 👍 

#### Changelog

**New**

**Changed**

- Move component stories from `carbon-react` to `react` and update any imports
- Move Layer component to react
- Move Theme component, useTheme hook to react
- Update re-exports to match new locations

**Removed**

- Remove extra index.js files from `carbon-react`
- Remove `start:v11` task

#### Testing / Reviewing

- Verify moved stories show up in `carbon-react`
- Verify moved stories do not show up in `react` storybook
- Verify packages export components and hooks as-expected


tay1orjones (approved):

jnm2377:
Looks great! Storybooks look right to me, I just had one question around some of the experimental components in our v10 storybook. I noticed some of them are also being included in the v11 experimental section, but some aren't (i.e. heading, text, etc). Should these be included in v11 too?
#### v10
Screen Shot 2021-12-06 at 2 17 13 PM

#### v11
Screen Shot 2021-12-06 at 2 17 19 PM

joshblack:
@jnm2377 good question, I think they're okay in v10 for now since they were added before and aren't necessarily dependent on v11 stuff. I think for ones that are, we can definitely move them over 👍

jnm2377 (approved):

aledavila (approved):

PR #10224, opened 2021-12-03
feat(DatePicker, DatePickerInput): refactors to functional component

### LINKED ISSUE #10010: [DatePickerInput]: upgrade to functional component

jnm2377:
Closes #10010 

Copied over from this draft: https://github.com/carbon-design-system/carbon/pull/10156

#### Changelog

- Refactors DatePickerInput to functional component
- Adds separate component for calendar icon (in same file)


packages/react/src/components/DatePickerInput/next/DatePickerInput.js
```diff
@@ -0,0 +1,321 @@
...
+  const helper = helperText ? (
+    <div className={helperTextClasses}>{helperText}</div>
+  ) : null;
+
+  let error = null;
+  if (invalid) {
+    error = <div className={`${prefix}--form-requirement`}>{invalidText}</div>;
+  } else if (warn) {
+    error = <div className={`${prefix}--form-requirement`}>{warnText}</div>;
+  }
```
tay1orjones: I think it might be easier to read and maintain these ternarys/conditionals if they were down in `render()`. What do you think?

packages/react/src/components/DatePickerInput/next/DatePickerInput.js
```diff
@@ -0,0 +1,321 @@
...
+
+  const containerClasses = cx(`${prefix}--date-picker-container`, {
+    [`${prefix}--date-picker--nolabel`]: !label,
+  });
```
tay1orjones: This could move up with the rest of the `cx` statements and rely directly on `labelText` instead of `label`.

packages/react/src/components/DatePickerInput/next/DatePickerInput.js
```diff
@@ -0,0 +1,321 @@
...
+  warn: PropTypes.bool,
+};
+
+DatePickerInput.propTypes = {
+  datePickerType: PropTypes.oneOf(['simple', 'single', 'range']),
+  iconDescription: PropTypes.string,
+  invalid: PropTypes.bool,
+  openCalendar: PropTypes.func,
+  warn: PropTypes.bool,
+};
```
tay1orjones:
```suggestion
```

tay1orjones (changes requested):
Just a couple minor things

joshblack (changes requested):
I think we'll need to make an update in `DatePicker` to make sure the `openCalendar` functionality works.

If I remember correctly from our work @jnm2377, the difference is in the `assign*` helpers in the class that assign the refs for `inputField` and `inputToField`. Let me know if I'm misremembering 🤔

jnm2377:
> I think we'll need to make an update in `DatePicker` to make sure the `openCalendar` functionality works.
> 
> If I remember correctly from our work @jnm2377, the difference is in the `assign*` helpers in the class that assign the refs for `inputField` and `inputToField`. Let me know if I'm misremembering 🤔

@joshblack I opened the PR based on your suggestion to try to get `DatePickerInput` in while continuing work on `DatePicker`, but should I just close this then? I know there’s still work to be done on `DatePicker` but seems like you're saying separating into different PRs won’t work?

abbeyhrt (approved):
Looks good to me!! 🎉

packages/react/src/components/DatePicker/next/DatePicker-story.js
```diff
@@ -0,0 +1,167 @@
+/**
```
tay1orjones: This file should be `DatePicker-stories.js`, right? I didn't think we wanted duplicate story files in `next` folders

tay1orjones (changes requested):

packages/react/src/components/DatePicker/next/DatePicker-story.js
jnm2377: You're probably right. Tbh I've been really confused with next/current stories. So should I just delete this file? @tay1orjones

tay1orjones: @jnm2377 Yeah no sweat! `*-stories.js` are what's pulled into the new storybook and should use the `next` version of components.

So I think just make sure every example in this `next/DatePicker-story.js` is already represented/included in the `next/DatePicker-stories.js` and then you can delete this `next/DatePicker-story.js`.

I'd also double check the imports in `next/DatePicker-stories.js` to be sure it's importing the `next` version of DatePicker, DatePickerInput, etc.

packages/react/src/components/DatePicker/next/DatePicker.js
```diff
@@ -0,0 +1,636 @@
...
+          startInputField.current.value = change.value;
+        }
+        return;
+      }
+
+      if (calendarRef.current) {
+        calendarRef.current.set(change.name, change.value);
+      }
+    });
+  }, [changes]); //eslint-disable-line react-hooks/exhaustive-deps
```
joshblack: This block seems to use `useTrackedChanges` but that hook is currently only returning an array. Does this need to be updated for the cDU scenario from the class component?

joshblack (changes requested):

packages/react/src/components/DatePicker/next/DatePicker.js
jnm2377: @joshblack I updated the PR. Tbh syncing prev & current "states" is always a little confusing for me. I was trying to wrap my head around how to best do this, and just went with something similar to how the class component did it. Lmk if this works. 😅

packages/react/src/components/DatePicker/next/DatePicker.js
```diff
@@ -0,0 +1,656 @@
...
+
+      if (end) {
+        end.removeEventListener('keydown', handleArrowDown);
+        end.removeEventListener('change', handleOnChange);
+      }
+    };
+  }, [savedOnChange, savedOnClose, savedOnOpen]); //eslint-disable-line react-hooks/exhaustive-deps
+
+  // component did update equivalent
+  useEffect(() => {
```
joshblack, line 422: Totally get what you mean by what you were saying the other day 😭  This is such a tough one to refactor.

I feel like we have only a handful of options that we can realistically do:

- One effect per calendar setting

```jsx
useEffect(() => {
  if (calendarRef.current) {
    calendarRef.current.set({ dateFormat });
  }
}, [dateFormat]);

useEffect(() => {
  if (calendarRef.current) {
    calendarRef.current.set({ minDate });
  }
}, [minDate]);

// ...
```

- Pros
  - Simple approach, each setting gets an effect
  - Simple to extend as we just can add on using the existing pattern
- Cons
  - Cumbersome, we have so many fields 😭 
  - Awkward "first render" problem where we only want to run this code _after_ the first render


- We could create a custom way to track these changes

```jsx
function useTrackChanges(dependencies) {
  const [changes, setChanges] = useState([]);
  const [snapshot, updateSnapshot] = useState(dependencies);

  useEffect(() => {
    const changes = Object.keys(dependencies)
      .filter((key) => {
        return dependencies[key] !== snapshot[key];
      })
      .map((key) => {
        return {
          key,
          value: dependencies[key],
        };
      });

    if (changes.length > 0) {
      setChanges(changes);
    }

    updateSnapshot(dependencies);
  }, [dependencies]);

  return changes;
}

function DatePicker({ dateFormat, minDate }) {
  const changes = useTrackChanges({
    dateFormat,
    minDate,
    // ...
  });

  useEffect(() => {
    changes.forEach(change => {
      const { key, value } = change;
      calendarRef.current.set({ [key]: value });
    });
  }, [changes]);
}
```

- Pros
  - Simpler `useEffect` for adding changes to the calendar
- Cons
  - We're working outside of React's system (a little bit)
  - Since the object identity may not always be stable, we may be doing equality checks that we may not need (fixable with useMemo, however)

---

When you're looking at some of these options and what you have so far, are there any that stand out to you or others that you think we should consider? 👀

jnm2377:
@joshblack this is ready for re-review. I updated as best I could some of the tests we went over yesterday, but ended up simplifying a bit just for the sake of getting this in. 😅

packages/react/src/components/DatePicker/next/DatePicker-test.js
```diff
@@ -0,0 +1,233 @@
...
+const prefix = 'bx';
+
+function getFlatpickrCalendar() {
+  return document.querySelector(`.${prefix}--date-picker__calendar`);
+}
+
+// function isCalendarVisible() {
+//   const calendar = document.querySelector(`.${prefix}--date-picker__calendar`);
+//   return calendar.classList.contains('open');
+// }
```
joshblack:
```suggestion
```

packages/react/src/components/DatePicker/next/DatePicker.js
```diff
@@ -0,0 +1,648 @@
...
+
+  return {
+    onMonthChange: updateCurrentMonth,
+    onValueUpdate: updateCurrentMonth,
+    onOpen: updateCurrentMonth,
+    onReady: [setupElements, updateCurrentMonth, register],
+  };
+};
+
+//returns true if children don't have labels
```
joshblack:
```suggestion
/**
 * Determine if every child in a list of children has no label specified
 * @param {Array<ReactElement>} children
 * @returns {boolean}
 */
```

packages/react/src/components/DatePicker/next/DatePicker.js
```diff
@@ -0,0 +1,648 @@
...
+          item.classList.contains('today') &&
+          calendar.selectedDates.length === 0
+        ) {
+          item.classList.remove('no-border');
+        }
+      });
+    }
+  }
+
+  // component did mount equivalent
```
joshblack:
```suggestion
```

packages/react/src/components/DatePicker/next/DatePicker.js
```diff
@@ -0,0 +1,648 @@
...
+
+      if (end) {
+        end.removeEventListener('keydown', handleArrowDown);
+        end.removeEventListener('change', handleOnChange);
+      }
+    };
+  }, [savedOnChange, savedOnClose, savedOnOpen]); //eslint-disable-line react-hooks/exhaustive-deps
+
+  // update calendar config if props change
+  // component did update equivalent
```
joshblack:
```suggestion
```

packages/react/src/components/DatePicker/next/DatePicker.js
```diff
@@ -0,0 +1,648 @@
...
+  useEffect(() => {
+    if (calendarRef.current && inline) {
+      calendarRef.current.set('inline', inline);
+    }
+  }, [inline]);
+
+  useEffect(() => {
+    if (calendarRef.current) {
+      calendarRef.current.set({ value });
+      updateClassNames(calendarRef.current);
```
joshblack: Is `updateClassNames` something that can be pulled out of the component or does it change based on data?

joshblack (changes requested):
Looking great! Left a couple comments/questions

packages/react/src/components/DatePicker/next/DatePicker.js
jnm2377: @joshblack If I understand correctly, it changes based on data. It's used when the component mounts and also here.

joshblack: @jnm2377 to avoid the `react-hooks/exhaustive-deps` warnings (and prevent future bugs) we should pull out `updateClassNames` of the component and add any additional arguments that it needs / depends on in the scope.

This should remove the need to get rid of the `react-hooks/exhaustive-deps` warnings (at least in cases where it's because of `updateClassNames`

tay1orjones (approved):

packages/react/src/components/DatePicker/next/DatePicker.js
jnm2377: @joshblack Is this a blocker for this PR? I looked into it yesterday afternoon, but was running into some issues bc it's used in the [useEffect](https://github.com/carbon-design-system/carbon/blob/1139c0e15820d3e023c93a764ef65fa777f094bb/packages/react/src/components/DatePicker/next/DatePicker.js#L269-L271) that creates the flatpickr calendar, and then it's used to actually create the flatpickr [calendar instance](https://github.com/carbon-design-system/carbon/blob/1139c0e15820d3e023c93a764ef65fa777f094bb/packages/react/src/components/DatePicker/next/DatePicker.js#L333-L340), so I'm not sure how that would work w/ pulling out the function. If it's working as expected, was wondering if this is ok as is?

joshblack: @jnm2377 the bug that we can run into is that if `updateClassNames` in the future is updated to use some kind of state then that state value would be stale (since the `useEffect` is only seeing the first "version" of this function).

What error were you seeing when you moved `updateClassNames` into the outermost scope of the page?

jnm2377: @joshblack when I pull it out, it no longer creates the flatpickr calendar instance (from the initial useEffect), so then there are a bunch of additional errors from other parts of the code that reference the instance. I think I understand the potential issue you're describing. I guess, I feel like that's more of an enhancement to support a state.

joshblack: @jnm2377 I think what might be happening is that `updateClassNames` requires `prefix` (it's using it from the parent scope) and so when it's moved to the top-level scope things are erroring out, included the calendar creation that you're talking about.

If it is updated to include `prefix` as an argument I think you won't see that issue anymore, let me know!

---

For background on why this is something I noticed, in general, we do not want to disable the eslint rule for the rules of hooks. There are specific moments where that is okay (if we can prove the dependencies will always be up-to-date or missing deps are stable) but they should be isolated since ignoring the rules of hooks will lead to hard-to-find bugs later down the line (in my opinion)

jnm2377: @joshblack I had included prefix differently (had also pulled it out to be used in the function), but hadn't tried including it as an argument. 👍🏽

jnm2377: @joshblack pushed the changes, ready for review

joshblack (approved):

PR #10226, opened 2021-12-03
refactor(styles,react): update layer styles to work in inline themes

### LINKED ISSUE #10049: [Bug]: Theme components/zones does not affect layer tokens
### Package

carbon-components

### Browser

Chrome, Firefox

### Package version

v11 / Modules

### Description

The theme components fails to affect a portion of color tokens that uses the new layer convention. The layer tokens are defined and evaluated at :root, and since CSS custom properties points to the *value* at the scope of the evaluation and not the *reference*, they remain unaffected by inner theming zones.

### CodeSandbox example

https://codesandbox.io/s/vigilant-rosalind-vfp20

### Steps to reproduce

Upload something.

The `cds--file__selected-file` box's background color is white despite the theme being g100. It refers to `--cds-field` which is defined at `:root` and at `:root` the `--cds-field` value defaults to the white one.

### Code of Conduct

joshblack:
Closes #10049 

This component updates our layering system to work in inline themes (aka zones). Currently, our system is flawed because it relies on the nesting of selectors. In other words, we expect the nesting of the `.cds--layer` class to determine its level. This does not work in inline themes because these areas of the page "reset" their layer level.

More specifically, we would expect the following to work for the `<Layer>` component:

```jsx
<div>By default, use an 01 layer token for a `layer` token</div>
<Layer>
  <div>Use an 02 layer token for a `layer` token</div>
  <Layer>
    <div>Use an 03 layer token for a `layer` token</div>
  </Layer>
</Layer>
```

This works in our current implementation, and each `Layer` component is essentially a `<div>` with the `cds--layer` class. Each level of the `cds--layer` class maps the `layer` token to the appropriate level:

```css
// We start at two since `--cds-layer` maps to `--cds-layer-01` by default
.cds--layer {
  --cds-layer: var(--cds-layer-02);
}

.cds--layer .cds--layer {
  --cds-layer: var(--cds-layer-03);
}
```

However, this does _not_ work in our inline themes as the nesting rules no longer apply.

```jsx
<div>By default, use an 01 layer token for a `layer` token</div>
<Layer>
  <div>Use an 02 layer token for a `layer` token</div>
  <Theme theme="g100">
    <div>Use an 01 layer token for a `layer` token</div>
  </Theme>
</Layer>
```

Here, the inline theme _does not_ use an 01 token and instead uses an 02 token value.

This PR addresses this problem by creating dedicated classes for layer sets:

- `.cds--layer-one`
- `.cds--layer-two`
- `.cds--layer-three`

We now use context to communicate the "layer" inside of React. This will correctly apply the layer class and allows us to "reset" the layer when using an inline theme.

#### Changelog

**New**

- Add a new `_layer.scss` entrypoint

**Changed**

- Update our styles entrypoint and `_zone.scss` entrypoint
- Update our Theme component to correct set the initial layer class
- Update our storybook to correctly pick-up side effects from our files
- Update our storybook to correctly set the default theme

**Removed**

- Remove `_layer-set.scss`

#### Testing / Reviewing

- View the layer storybook stories and verify they work as intended
- View the theme "with layer" story and verify it works as intended


joshblack:
Let me know if you have any questions @abbeyhrt @tay1orjones @dakahn, happy to dive into any details if it's helpful!

abbeyhrt (approved):
Very cool! It looks good as far as I can tell!

packages/carbon-react/.storybook/preview.js
```diff
@@ -8,19 +8,10 @@
 import './styles.scss';
 
 import { configureActions } from '@storybook/addon-actions';
-import * as FeatureFlags from '@carbon/feature-flags';
 import { white, g10, g90, g100 } from '@carbon/themes';
 import React from 'react';
 import { breakpoints } from '@carbon/layout';
-import { settings } from 'carbon-components';
-
-FeatureFlags.merge({
```
abbeyhrt, line 17: Out of curiosity, why are the feature flags not needed anymore?

packages/react/src/components/Layer/index.js
```diff
@@ -9,19 +9,28 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { usePrefix } from '../../internal/usePrefix';
+import { LayerContext } from './LayerContext';
```
tay1orjones, line 12: @joshblack I think this is why the CI is failing. Does `LayerContext` path need to be qualified further being inside the `next` folder?

packages/react/src/components/Theme/index.js
```diff
@@ -9,8 +9,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { usePrefix } from '../../internal/usePrefix';
+import { LayerContext } from '../Layer/LayerContext';
```
tay1orjones, line 12: Same thing here with the path needing updated to include `next`. Or maybe `LayerContext` should be pulled out of the `next` folder?

packages/react/src/components/Layer/index.js
joshblack: Good spot! Updated

packages/react/src/components/Theme/index.js
joshblack: Good call, updated!

packages/carbon-react/.storybook/preview.js
joshblack: Great question! It seems like since our files in `src` were not included in the `sideEffects` field in `package.json` that webpack was incorrectly removing the files that had side effects. Once they were added, any imports we made to components from the project correctly brought in side-effects and now it works as intended 🎉

joshblack:
@dakahn @tay1orjones should be good to go with all the story fixes!

tay1orjones (approved):

PR #10236, opened 2021-12-06
Multi select controlled

### LINKED ISSUE #4812: MultiSelect - ability to have more control over and clear selected items
### Summary

I would like to request that the MultiSelect component be updated so that it's possible to have more control over the selection of items and clearing of already selected items, including clearing all selected items at once. Something similar was previously discussed in this issue that I found with a solution that I ended up using for now for my use case (https://github.com/carbon-design-system/carbon/issues/2623). Also in that issue, the implementation for this requested enhancement that was floated as possibly being the way to go was surfacing Downshift's action props to help with clearing selection/items/etc.

### Justification

This came up in our project when we needed to be able to clear all selections in the MultiSelect component for our Watson Discovery React components from one button that also clears all selections across other components, which use Carbon's Checkbox. You can see the related pull request here, which has more detail, if it helps: https://github.com/watson-developer-cloud/discovery-components/pull/11

The desired UX is that a developer who uses the MultiSelect React component is able to control the component's selections, both selecting and deselecting, from a button or other component instead of one at a time by the user. We can measure the success for the developer if it's possible to do this with props, and there is no longer any need to reference the DOM, as in the solution included in the previous issue above ^.

### "Must have" functionality

Ability to clear all selected items at once is the "must have" functionality for this request.

No, but I wanted to surface this since it's a question that was asked in the past for a similar use case to get it back on the team's radar.

### Available extra resources

I am a front-end developer who would be able/willing to help with implementing these changes. I wanted to first open this feature request and get feedback from the Carbon team though before proceeding. Please let me know any additional details/info I can provide, thank you!

dezkareid:
Closes #4812 

Unable to control react/MultiSelect items selected

#### Changelog

**New**

- Add a feature to control selected items on Multilselect

#### Testing / Reviewing

You can see Multiselect/controlled story and see how items are controlled


dezkareid:

packages/react/src/components/MultiSelect/MultiSelect.js
```diff
@@ -465,6 +467,11 @@ MultiSelect.propTypes = {
    */
   useTitleInItem: PropTypes.bool,
 
+  /**
+   * Specify value to control items selected
+   */
+  value: PropTypes.array,
```
tay1orjones: I think this is a good opportunity to unify prop naming with similar components Dropdown and Combobox and refer to this as `selectedItems`

```suggestion
  /**
   * For full control of the selected items
   */
  selectedItems: PropTypes.array,
```

packages/react/src/components/MultiSelect/MultiSelect.js
```diff
@@ -85,6 +86,7 @@ const MultiSelect = React.forwardRef(function MultiSelect(
     disabled,
     initialSelectedItems,
     onChange,
+    controlledItems: value,
```
tay1orjones: selected
```suggestion
    selectedItems
```

packages/react/src/components/MultiSelect/MultiSelect-story.js
```diff
@@ -143,6 +144,44 @@ export const Default = withReadme(readme, () => {
...
+          setSelectedItems([{ id: 'downshift-1-item-0', text: 'Option 1' }]);
+        }}>
+        click to select only Option 1
+      </Button>
+      <Button
+        onClick={() => {
+          setSelectedItems([]);
+        }}>
+        click to clear selections
+      </Button>
```
tay1orjones, line 180: I just added these to this PR to make testing a bit easier in the storybook 👍

tay1orjones (changes requested):
Thanks so much for this contribution @dezkareid! Below are a couple of things I noticed at first glance. 

I'd like to dig in a bit further on the changes to `useSelection` and understand more fully what's going on in that logic.

packages/react/src/components/MultiSelect/MultiSelect.js
dezkareid: Sure, working on it

packages/react/src/components/MultiSelect/MultiSelect-story.js
dezkareid: thanks

packages/react/src/components/MultiSelect/MultiSelect.js
dezkareid: > 

I have a comment here, I can use the name selectedItems on props but there is one property on the same scope, I don't know if just rename it on destructucting `selectedItems: selected` or rename the other selectedItems variable, is on line 97.

Also on useSelection can be a little confusing because there is a property called selectedItems

tay1orjones (approved):
Thanks for making those changes, this looks great 👍

dakahn (approved):

fabpico:
For those who need a temporary workaround (eg. dont want yet to upgrade carbon), you can force rerender MultiSelect and use `initialSelectedItems`.

PR #10239, opened 2021-12-07
feat(pagination): add forwardedRef for pagination container div

### LINKED ISSUE #10240: [Feature Request]: add forward ref to Pagination container
### Summary

We have a few situations where it would be nice to check the size of the Pagination container and apply styles differently. We can achieve this by wrapping Pagination in an extra div, but it would be nice to avoid the extra markup and check the Pagination container directly.

### Justification

_No response_

_No response_

### Required functionality

_No response_

_No response_

### Available extra resources

I've written the PR. 

### Code of Conduct

kevinsperrine:
Closes #10240

This PR adds a `forwardedRef` prop to the Pagination component. This allows us to check the size of the component and adjust styles as necessary in different circumstances.

#### Changelog

**New**

- add `forwardedRef` prop to Pagination component. This allows us to check the `getBoundingClientRect` of the pagination container and change classes or styles as necessary. We _can_ achieve this via wrapping the Pagination with an extra div, but it would be nicer to avoid the extra markup.

#### Testing / Reviewing

Pagination should be unaffected in stories and the new test should pass confirming the presence of the ref.


packages/react/src/components/Pagination/Pagination.js
```diff
@@ -312,7 +321,7 @@ export default class Pagination extends Component {
     const pageSizes = mapPageSizesToObject(_pageSizes);
 
     return (
-      <div className={classNames} {...other}>
+      <div className={classNames} ref={forwardedRef} {...other}>
```
jnm2377, line 316: Hmm.. thoughts on doing something like `rest.ref` rather than adding a whole new prop for it in v10, and having to deprecate it in v11 when it's no longer needed?
DatePicker class component did this with `inline` to not have to introduce it as a prop: https://github.com/carbon-design-system/carbon/blob/ddaadaad1e95dd44dc34b5c16268437568123793/packages/react/src/components/DatePicker/DatePicker.js#L340

kevinsperrine: Good call. That is definitely cleaner. I will adjust.

jnm2377:
hi @kevinsperrine just wanted to check in on this and see if it's something you're still working on?

kevinsperrine:
> hi @kevinsperrine just wanted to check in on this and see if it's something you're still working on?

Hi @jnm2377, Sorry, I forgot about this after the holidays. Yes', I'll adjust and push up the changes today.

jnm2377 (approved):

aledavila (approved):

PR #10241, opened 2021-12-07
fix(checkbox): do not propagate click event from label

### LINKED ISSUE #10122: [Bug]: Checkbox onClick event firing twice not allowing checkboxes to be selected
### Package

carbon-components-react

### Browser

Chrome

### Package version

v10.48.0

### Description

The onClick event is being fired on both the label and the input causing the checkbox to check and immediately uncheck.

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-checkbox--playground
https://codesandbox.io/s/cool-lederberg-qwoc1?file=/src/index.js

### Steps to reproduce

Try to select a checkbox in the playground. https://react.carbondesignsystem.com/?path=/story/components-checkbox--playground

### Code of Conduct

tay1orjones:
Closes #10122 

Mirror Checkbox to have parity with recent changes to InlineCheckbox https://github.com/carbon-design-system/carbon/pull/10151 and only allow click events to propagate from the `input` and not the `label`. 

#### Changelog

**Changed**

- fix(checkbox): do not propagate click event from label

#### Testing / Reviewing

- Open the carbon-components-react storybook
- View the Checkbox default story - it should function as normal
- View the Checkbox Playground story
- Click a checkbox input, 2 events should fire: 
   1. `onChange`
   1. `onClick` with `event.target` indicating the `input`. 
       * Sometimes storybook actions addon won't show `target` appropriately. You may have to place a console.log or debugger in the dev tools to see the event.target for `onClick`
- Click a checkbox _label_, the same set of 2 events should fire

### DIFF
--- a/packages/react/src/components/Checkbox/Checkbox-test.js
+++ b/packages/react/src/components/Checkbox/Checkbox-test.js
@@ -109,6 +109,17 @@ describe('Checkbox', () => {
       expect(call[2].target).toBe(inputElement);
     });
   });
+  it('should only propagate click events from the input', () => {
+    const onClick = jest.fn();
+    render(
+      /* eslint-disable jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */
+      <div onClick={onClick}>
+        <Checkbox id="checkbox" labelText="checkbox" />
+      </div>
+    );
+    userEvent.click(screen.getByRole('checkbox'));
+    expect(onClick).toHaveBeenCalledTimes(1);
+  });
 });
 
 describe('refs', () => {
--- a/packages/react/src/components/Checkbox/Checkbox.js
+++ b/packages/react/src/components/Checkbox/Checkbox.js
@@ -77,7 +77,14 @@ const Checkbox = React.forwardRef(function Checkbox(
           }
         }}
       />
-      <label htmlFor={id} className={labelClasses} title={title || null}>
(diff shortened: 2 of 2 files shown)


abbeyhrt (approved):
Looks good to me!

joshblack:
@tay1orjones is the double click coming from something odd that we're doing? Was trying to see if this happened with default checkboxes and it seemed to work as expected

tay1orjones:
@joshblack I have no clue. I can't figure it out. If only one click is used, the tooltip body never shows up. I tried using the async helpers, `waitFor`, etc. and tried all methods for testing appearance/disappearance from [RTL's docs](https://testing-library.com/docs/guide-disappearance/) but no dice.

joshblack:
@tay1orjones I'd love to pair up and also dig in to see what's happening. Would be a bummer if we had to add `stopPropagation` calls throughout labels but I totally get that it may be an option if we can't figure out why this is happening.

tay1orjones:
@joshblack Ah I totally thought this was a different PR! Disregard my previous comment. 

In this case the browser checkbox naturally propagates clicks from associated labels through the input. Here's an ultra reduced testcase https://codepen.io/tay1orjones/pen/zYdbRvd

You can see two events when clicking on the label, one event when clicking on the input.

tay1orjones:
Closing in favor of https://github.com/carbon-design-system/carbon/pull/10264

PR #10257, opened 2021-12-08
refactor(number-input): refactor to functional component

### LINKED ISSUE #10017: [NumberInput]: upgrade to functional component

joshblack:
Closes #10017 

Update our `NumberInput` component to a functional component.

#### Changelog

**New**

**Changed**

- Update `NumberInput` to a functional component

**Removed**

#### Testing / Reviewing

- Verify component works as expected in storybook
- Review test changes to make sure they match previous tests
- Review refactor to make sure functionality not covered in tests is not list


packages/react/src/components/NumberInput/next/NumberInput.js
```diff
@@ -0,0 +1,456 @@
...
+  'decrement.number': 'decrement.number',
+};
+
+const defaultTranslations = {
+  [translationIds['increment.number']]: 'Increment number',
+  [translationIds['decrement.number']]: 'Decrement number',
+};
+
+const NumberInput = React.forwardRef(function NumberInput(props, forwardRef) {
+  const enabled = useFeatureFlag('enable-v11-release');
```
tay1orjones, line 29: Do we have a convention that props destructure should always be first?

tay1orjones (approved):
One minor question, but looks solid to me. 

I was surprised to find this didn't require any useEffect's. Also breaking out Label and HelperText into their own components feels super good 👍

packages/react/src/components/NumberInput/next/NumberInput.js
joshblack: @tay1orjones I think this is the _only_ exception since `enabled` is used for the default values on `props` lol.

It was tricky to see what made more sense, it used to use `FeatureFlags.enabled` statically and so this seemed like an appropriate trade-off. Let me know what you think

dakahn (approved):

PR #10282, opened 2021-12-13
feat(www): refactor insights pages and add packages

joshblack:
This PR cleans up some of our www pages and adds support for packages. This is to make it easier to get details for a specific package (like component props) in a future PR. Here's a quick video of the changes

https://user-images.githubusercontent.com/3901764/145865950-691d49f3-1329-4a22-8c26-739ca62b5d75.mov

#### Changelog

**New**

- Add new project support and test components like `Flex`, `Box`, and `Text` to help with building out pages

**Changed**

- Update build task for `@carbon/react` to include `_layer` in output

**Removed**

#### Testing / Reviewing

- Review as much or as little as you want! `www` is just a playground for project stats
- If you want to run it locally, make sure you `export GH_TOKEN=...` with your GitHub Personal Access Token and run `node www/tasks/collect.js` to get all the GH data before running `yarn develop` in `www`


www/src/crypto/murmur.js
```diff
@@ -0,0 +1,87 @@
+/**
```
abbeyhrt, line 1: Can you explain this whole file? This looks mainly like nonsense to me haha

abbeyhrt (approved):
Looks good as far as I can tell! I'm definitely curious about how `murmer.js` works!

www/src/crypto/murmur.js
joshblack: @abbeyhrt thankfully this is a file that is just copy-pasted around the internet lol. You can find the implementation for murmur hash on GitHub, but effectively it's a specific type of hash that is quick and generally useful for looking things up based on id.

This can be nice if you need something that generally needs a "stable" id, whether that's for a URL or for looking things up from different parts of a system. In terms of why murmur hash, it could 100% be replaced with any other hash function and we'd be fine 😂  it's just the one I've used most recently.

tay1orjones (approved):

PR #10288, opened 2021-12-14
fix(DataTable): expand all header uses correct aria label

### LINKED ISSUE #10254: [Bug]: Expand all rows table header button shows the wrong aria-label
### Package

carbon-components-react

### Browser

Chrome

### Package version

v7.49.1

### React version

_No response_

### Description

When all the rows are collapsed the header expand all button's aria-label shows "Collapse all rows" instead of "Expand all rows"
image

And then the opposite happens here
image

Fix should be to just change !isExpanded to isExpanded here
https://github.com/carbon-design-system/carbon/blob/83f414f2c9186796bd31959786369c903eb959af/packages/react/src/components/DataTable/DataTable.js#L259

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-datatable-expansion--batch-expansion

### Steps to reproduce

https://react.carbondesignsystem.com/?path=/story/components-datatable-expansion--batch-expansion

### Code of Conduct

jnm2377:
Closes #10254 

#### Changelog

- applies "collapse all" text when `isExpanded` is true

#### Testing / Reviewing

- view Expand All story > inspect the Expand All header `button` in the dom > when it's collapsed, the aria label should now say "expand all", and vice versa. Compare to wrong label in current storybook.


joshblack (approved):

abbeyhrt (approved):
🚀

PR #10296, opened 2021-12-15
fix(DataTable): fix small toolbar sizing and color bug

### LINKED ISSUE #10058: [Bug]: TableToolbarMenu size and color incorrect
### Package

carbon-components

### Browser

Chrome

### Package version

@carbon/react@0.8.1

### Description

The element does not respect the size constraints when `<TableToolbar size="sm">`, causing the icon to be misaligned:
image
The parent to the svg element does not have a set `color` and thus the svg `fill` cannot correspond to the current theme:
image
This should probably be `$icon-01`.

### CodeSandbox example

https://codesandbox.io/s/youthful-driscoll-vbjk5?file=/src/index.js

### Steps to reproduce

1. Using `@carbon/react@0.8.1`, create a TableTooblar with a TableToolbarMenu.
2. Inspect the SVG and note the `fill` property is set to `currentcolor`
3. Inspect the OverflowContainer and note that it does not conform to the height of the containing TableToolbarMenu

The icon fill can be fixed by adding `color: var(--cds-icon-primary)` to the `TableToolbarMenu`, although IMO it would be nice if every icon had the `fill` property defaulting to `$icon-01` instead of needing to inherit that from its parents color.

### Code of Conduct

jnm2377:
Closes #10058 

#### Changelog

- cleans up data table stories in v10:
    - deletes `DataTable-story.js` since storybook wasn't using those stories
    - adds basic table examples (w/ sizes + zebra)
    - adds toolbar examples (w/ different sizes)
    - adds small batch actions example
    - renamed some examples to be more semantic
- adds v11 data table stories which were missing
- fixes small toolbar sizing issues
- fixes batch action sizing issues
- fixes "color" issues in overflow menu: `OverflowMenu` v11 refactor wasn't adding the icon class name to the trigger icon, so it wasn't getting icon fill theming, which was also present in DataTable
#### Testing / Reviewing

- View v10 and v11 toolbar examples to check styles are correct. We don't currently have a story that shows the bug, so can checkout locally and comment out the new styles to see what the existing bug looks like, if you need a reference on what to look for.
- Test out different theme colors for v11 DataTable, and check to see that overflow menu icon fill is now changing color according to theme


abbeyhrt:
@jnm2377 oof, the data table stories were so chaotic, thank you for taking that on and re-organizing so it worked with `@carbon/react`! I like the new arrangement overall but could we have some stories that aren't under folders? Just maybe like the `Usage` story or "Playground" story

abbeyhrt:
Couple of spacing issues with the toolbar  in `@carbon/react` but the v10 one looks all good! 

Screen Shot 2021-12-16 at 2 15 43 PM
Screen Shot 2021-12-16 at 2 15 25 PM
Screen Shot 2021-12-16 at 2 15 17 PM

jnm2377:
@abbeyhrt updated the bugs you found 👍🏽

abbeyhrt:
Nice! I think it looks mostly good but, at least comparing the v11 data table to v10, it looks like there is a little too much space to the left of the icon here:
v11:
Screen Shot 2022-01-13 at 2 58 36 PM

v10: 
Screen Shot 2022-01-13 at 3 00 28 PM

jnm2377:
@abbeyhrt and @tay1orjones updated the styles, so v11 toolbar search icon should be aligned now. Sorry the PR files might be a little confusing with all the story changes. The fixes pertaining to this bug were really in 3 files:
- `packages/components/src/components/data-table/_data-table-action.scss`
-  `packages/styles/scss/components/data-table/action/_data-table-action.scss`
- `packages/react/src/components/OverflowMenu/next/OverflowMenu.js` 

The rest are story updates.

tay1orjones:
Checks were failing due to updates pulled in from https://github.com/carbon-design-system/carbon/pull/10535. I pushed an update converting `$layer-disabled` to use `$layer` instead 👍

abbeyhrt (approved):
Looks great!!

tay1orjones (approved):

PR #10302, opened 2021-12-15
feat(jest-config): add support for loading scss files

joshblack:
Add support for importing sass files from Jest. This is helpful for visibility checks in JSDOM that are based on CSS.

#### Changelog

**New**

**Changed**

- Update our CSS transformer to compile and insert the require'd CSS into the document head
- Update our transform options to apply to scss or css

**Removed**

#### Testing / Reviewing

- Verify stylesheet is loaded in the accordion test


jnm2377:
Hmm. Looks like some theme tests are failing.

joshblack:
@jnm2377 just pushed up a fix 👀

packages/react/src/components/Accordion/__tests__/Accordion-test.js
```diff
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import 'carbon-components/src/components/accordion/_accordion.scss';
```
tay1orjones, line 8: Would it make sense to also add an example test/assertion to fully validate the styles are included as we expect?

The `.#{$prefix}--accordion__content` initially has `display: none;` and could be asserted as such in the first "should render" test with `expect(getByText('Panel A')).not.toBeVisible();`

tay1orjones: I guess this would moreso apply to AccordionItem tests

joshblack: @tay1orjones I'm down either way, just let me know. When porting it over I was only looking at if the styles were being brought in or not

tay1orjones: That's fair, we'll validate this stuff as the refactor to RTL progresses

tay1orjones (approved):

packages/react/src/components/Accordion/__tests__/Accordion-test.js
jnm2377: How do we check if the styles are being brought in or not?

joshblack: @jnm2377 one quick way could be to change the `Accordion-test` to `console.log(document.styleSheets)` and see if one is loaded and explore the rules inside of it 👀  Another way could be to print out `document.head.innerHTML` which would print out the `<style>` tag that contains the styles

jnm2377 (approved):

PR #10314, opened 2021-12-17
feat: allow externally rendered tabs

### LINKED ISSUE #10313: Separate tab and tab content rendering
### Summary

The current Tabs component in React renders the tab content immediately after the list of Tab buttons. It would be useful to be able to separate these two pieces of content without losing the automatic management of the tab content.

### Justification

In @carbon/ibm-products we place the tabs inside a header component which has a background and a bottom border separating it from the content. We wish to use the tabs to control the content displayed below the header and as a result, have to wire these up using onClick.the 

While the intention is to separate the content from the tabs, no UX changes are expected. This does allow for tab and content to be rendered separately, so we may with to document that is not desirable.

### Required functionality

In @carbon/ibm-products we can remove the need for our users to wire up the Tabs to the content when placing the content in a separate section.

_No response_

### Available extra resources

Can apply solution used in PR above to Carbon.

### Code of Conduct

lee-chase:
Closes #10313

Allows the user to supply a `targetRef` in which to place the content of the tabs.

#### Changelog

**Changed**

packages/react/src/components/Tabs/Tabs-story.js
packages/react/src/components/Tabs/Tabs-story.scss
packages/react/src/components/Tabs/Tabs.js

#### Testing / Reviewing

Added an alt background to the tabs in storybook to show they are in a distinct section. Should probably remove before merging.


lee-chase:
It occurred to me while submitting this PR that it might make more sense to renter the Tab control buttons in a alternate location. In the case of @carbon/ibm-products this would mean Tab content can be defined where we wish to display rather than pass length content into the PageHeader section.

abbeyhrt:
Hey @lee-chase in v11, we're making some updates to the way that `Tabs` are composed so that now, for example, `TabContent`(which will be named `TabPanel`) is a separate component from `Tab`. The new composition looks like this: 
 ```
<Tabs>
    <TabList>
      <Tab>Tab Label 1</Tab>
      <Tab>Tab Label 2</Tab>
      <Tab>Tab Label 3</Tab>
    </TabList>
    <TabPanels>
      <TabPanel>Tab Panel 1</TabPanel>
      <TabPanel>
        Tab Panel 2 <Button>Example button</Button>
      </TabPanel>
      <TabPanel>Tab Panel 3</TabPanel>
    </TabPanels>
  </Tabs>
```
with those changes incoming, do you think that the ability to separate the rendering of `Tab` and `TabContent/TabPanel` would be possible?

**Comment by lee-chase** [comment]
> Hey @lee-chase! In v11, we're making some updates to the way that `Tabs` are composed so that now, for example, `TabContent`(which will be named `TabPanel`) is a separate component from `Tab`. The new composition looks like this: 
> 
>  ```
> 
> <Tabs>
> 
>     <TabList>
> 
>       <Tab>Tab Label 1</Tab>
> 
>       <Tab>Tab Label 2</Tab>
> 
>       <Tab>Tab Label 3</Tab>
> 
>     </TabList>
> 
>     <TabPanels>
> 
>       <TabPanel>Tab Panel 1</TabPanel>
> 
>       <TabPanel>
> 
>         Tab Panel 2 <Button>Example button</Button>
> 
>       </TabPanel>
> 
>       <TabPanel>Tab Panel 3</TabPanel>
> 
>     </TabPanels>
> 
>   </Tabs>
> 
> ```
> 
> with those changes incoming, do you think that the ability to separate the rendering of `Tab` and `TabContent/TabPanel` like you're wanting would be possible? 

It would need to be possible to render the TabPanels outside of the Tabs. As per my PR the best way to facilitate this is in React using a portal.

**Comment by joshblack** [comment]
@lee-chase how are you thinking focus management will work in the portal use-case for tabs?

Specifically, was thinking about the flow of:

- Tab to the active tab in the list of tabs
- Use arrow keys to navigate to a specific tab
- Press tab to navigate to the content

In the portal use-case, I believe that it would remove the content from the document flow and focus would need to be programmatically shifted to the content in the portal. Is that you're understanding as well or is there a part of this flow that I'm missing from the PR? 👀

**Comment by lee-chase** [comment]
@joshblack documentation.  The goal is to allow tabs to be rendered and reserve space inside one section and the tab content in the following section. I do not think we should try to manage the tabbing between content and buttons.

e.g.
```html
<header>
  <!-- Tab buttons are here for presentation purposes at the bottom of the header -->
</header>
<!-- 
<main>
  <!-- Tab content is the first thing interactive content that comes after the Tab buttons-->
</main>
```

`@carbon/ibm-products` achieves the same thing using `onSelectionChange`, this was based on internal components that used this pattern.

joshblack:
@lee-chase got it, thanks for clarifying! I really appreciate it.

Based on what you're saying, it seems like the intent is to be able to have the tab list and tab panels be rendered in separate sub-trees, ideally without having a common ancestor. You brought up `onSelectionChange` which I imagine is something like:

```jsx
function DemoComponent() {
  const [index, setIndex] = React.useState(0);

  return (
    <>
      <header>
        <TabList onSelectionChange={(index) => setIndex(index)}>
          <Tab>One</Tab>
          <Tab>Two</Tab>
          <Tab>Three</Tab>
        </TabList>
      </header>
      <main>
        <TabPanel hidden={index === 0} />
        <TabPanel hidden={index === 1} />
        <TabPanel hidden={index === 2} />
      </main>
    </>
  );
}
```

Is that close or am I missing something critical?

lee-chase:
Great summary and example component, although I'd tweak slightly for our use case where the header component is defined in a separate package. The portal suggestion removes the need for the user to handle the tab change.

```js
// Component library component (npm package)
function MyHeader({tabsConfig, onTabChange, ...rest}) {
  return (
    <header>
      <!-- other stuff based on ...rest -->
      <TabList onSelectionChange={onTabChange}}>
        { tabsConfig.map(tab => <Tab>{{tab}}</Tab>) }
      </TabList>
    </header>
  )
};

// Use of library component
function DemoComponent() {
  const [index, setIndex] = React.useState(0);
  const tabsConfig = ['One', 'Two', 'Three'];
  const restOfHeaderConfig = {};
  
  return (
    <>
      <MyHeader tabsConfig={tabsConfig} onTabChange={(index) => setIndex(index)} {...restOfHeaderConfig} />
      <main>
        <TabPanel hidden={index === 0} />
        <TabPanel hidden={index === 1} />
        <TabPanel hidden={index === 2} />
      </main>
    </>
  );
}
```

joshblack:
@lee-chase awesome, love the example. Thanks for adding it.

When it comes to the approach for splitting the rendering of a component into distinct sub-trees, I think my preference would be one of:

- Have a common parent who provides context
- Provide behavior through hooks and allow the owner to wire things up between components

I think the portal approach clearly can work for this use-case but does not align with our typical usage of portals.

Are either of these approaches feasible in your opinion or is the portal approach the clear cut winner for you?

PR #10315, opened 2021-12-18
feat: allow separately rendered tab controls

### LINKED ISSUE #10313: Separate tab and tab content rendering

lee-chase:
Closes #10313 
Relates to #10314

Allows the user to supply a `controlsTargetRef` in which to place the tab controls.

As noted in #10314 it may be more logical to declare the content where it should be located int eh DOM and allow the tab controls to be rendered elsewhere. This PR demonstrates this options.

#### Changelog

**Changed**

packages/react/src/components/Tabs/Tabs-story.js
packages/react/src/components/Tabs/Tabs-story.scss
packages/react/src/components/Tabs/Tabs.js

#### Testing / Reviewing

Added an alt background to the tabs in storybook to show they are in a distinct section. Should probably remove before merging.


joshblack:
Left a comment over on: #10314, curious what you think over there 👀

jnm2377:
Since this has gone stale, I'm going to close this for now. We can re-open it if further discussions warrant it :)

PR #10335, opened 2022-01-04
feat(eslint-config-carbon): add storybook config

joshblack:
This PR introduces an ESLint configuration for `eslint-plugin-storybook`. This plugin enables a variety of rules for the style and/or convention of our storybook. These rules will make sure our stories are structured consistently for Component Story Format v3.

#### Changelog

**New**

- Add `eslint-plugin-storybook` to our ESLint configuration and corresponding config

**Changed**

- Updated files based on new rules, more often than not this fell into two categories:
  - Moving `parameters.component` to `component`
  - Updating stories to be PascalCase

**Removed**

#### Testing / Reviewing

- Verify the plugin is running on story files (`*.stories.js`, `*-story.js`)


packages/carbon-react/src/components/Icons/Icons.stories.js
```diff
@@ -9,6 +9,7 @@ import './Icons.stories.scss';
 import React from 'react';
 import { Bee, Bicycle, ChevronUp } from '../../../icons';
 
+// eslint-disable-next-line storybook/csf-component
```
abbeyhrt, line 12: Why do these lines need to be disabled?

abbeyhrt (approved):
Looks good to me!

packages/carbon-react/src/components/Icons/Icons.stories.js
joshblack: @abbeyhrt basically the rule is that every story _must_ have a `component:` key, which is great except we sometimes have stories that aren't for a specific component. This rule therefore doesn't really apply and so we ignore it 👀

dakahn:
looks like we've got a storybook error about missing stories

joshblack:
@dakahn should be good to go now!

dakahn (approved):

PR #10357, opened 2022-01-06
feat(react): deprecate `renderExpando` and add `renderToggle`

### LINKED ISSUE #10317: Rename "expando" to "expand"
We want to use `expand` nomenclature instead of `expando`. This update will fall into one of a few categories:

1. Prop name usage
   - In these cases, props should be renamed. Existing props deprecated.
2. Internal component usages (functions, variables, etc)
   - If it's not in the public API, we can update it now without worry.
3. Story, documentation, comment usages
   - These by and large should be updated now to use the new `expand` term. In cases of documentation relating to prop name usage, we can update to use the new `expand*` prop instead of the deprecated prop.

The following is a list of usages of `expando` in the codebase that need to be looked at (found via a find all search):
- [ ] `components/demo/js/data-table-demo-expand-all-manager.js`
- [ ] `react/src/components/Accordion/AccordionItem.js`
- [ ] `react/src/components/DataTable/TableExpandHeader.js`
- [ ] `react/src/components/DataTable/stories/expansion/DataTable-expansion-story.js`
- [ ] `react/src/components/TreeView/TreeNode.js`
- [ ] There do not appear to be any usages of `expando` on the website, but double check

sstrubberg:
Closes #10317 

#### Changed
- Deprecated `renderExpando` in favor of `renderToggle` in AccordionItem so it wouldn't create a breaking change.
- Renamed internal `Expando` component to `Toggle` in AccordionItem for consistency.
- Replaced `enableExpando` with `enableToggle` in the DataTableExpansion story to finalize the deprecation started in the TableExpandHeader component.

#### Testing
- Make sure there are no breaking changes.
- Ensure Accordion and DataTableExpansion stories still work as expected.


packages/react/src/components/Accordion/AccordionItem.js
```diff
@@ -134,15 +134,15 @@ AccordionItem.propTypes = {
...
    */
   open: PropTypes.bool,
 
   /**
-   * The callback function to render the expando button.
+   * The callback function to render the expand button.
    * Can be a React component class.
    */
-  renderExpando: PropTypes.func,
+  renderExpand: PropTypes.func,
```
abbeyhrt: Would this be a breaking change?

sstrubberg: > Would this be a breaking change?

I was curious about that myself. This doesn't appear to be publicly exposed, so I thought no. Totally could be mis-reading that though.

packages/react/src/components/DataTable/stories/expansion/DataTable-expansion-story.js
```diff
@@ -334,7 +334,7 @@ export const BatchExpansion = () => (
           <TableHead>
             <TableRow>
               <TableExpandHeader
-                enableExpando={true}
+                enableToggle={true}
```
joshblack:
```suggestion
                enableToggle
```

packages/react/src/components/Accordion/AccordionItem.js
```diff
@@ -16,15 +16,15 @@ import { useId } from '../../internal/useId';
...
+const defaultRenderExpand = (props) => <button type="button" {...props} />;
 
 function AccordionItem({
   children,
   className: customClassName,
   iconDescription, // eslint-disable-line
   open = false,
   onHeadingClick,
-  renderExpando: Expando = defaultRenderExpando,
+  renderExpand: Expand = defaultRenderExpand,
```
joshblack: One thing that I don't know if we decided on from the expando -> expand prop change was when we want to use expand versus toggle.

Does it make sense to refer to this as a toggle in the form as `renderToggle`? Expand implies on where-as toggle implies on/off, if I remember the terminology that we were using.

joshblack (changes requested):
Thanks for taking this on! Left a couple of notes. I believe `renderExpando` is a public prop and so we should also make sure to handle the migration case for it (let me know if I'm misunderstanding)

abbeyhrt (approved):

packages/react/src/components/Accordion/AccordionItem.js
```diff
@@ -41,6 +42,8 @@ function AccordionItem({
     [customClassName]: !!customClassName,
   });
 
+  const Toggle = renderToggle || renderExpando; // remove renderExpando in next major release
```
joshblack, line 45: I think this logic may result in a breaking change if the `renderToggle` prop above has a default value.

Specifically, since `renderToggle` is _always_ defined it will always be used instead of `renderExpando`. As a result, `renderExpando` would not be used.

I think to fix it we would need to remove one of the optional values, probably:

```
  renderExpando = defaultRenderToggle, // remove renderExpando in next major release
  renderToggle,
```

So that `renderToggle` is used if it's defined, otherwise it will fallback to `renderExpando` and support existing usage.

Let me know if this makes sense!

joshblack (changes requested):

packages/react/src/components/Accordion/AccordionItem.js
sstrubberg: Ok, i removed the default prop to `renderToggle`.

sstrubberg: I'm worried that `renderToggle` won't receive the goodies from `defaultRenderToggle` now. Should I be worried?

joshblack: @sstrubberg Since `renderExpando` has the `defaultRenderToggle` fallback it will still use that value if neither `renderToggle` nor `renderExpando` is provided as a prop. Let me know if that makes sense!

joshblack (approved):

PR #10365, opened 2022-01-07
feat(react): add toggletip component

### LINKED ISSUE #10280: ToggleTip
**About**

This component replaces the interactive use-case of the Tooltip component. Specifically, tooltips may no longer have interactive descendants. As a result, this component allows users to have interactive descendants in popovers using a disclosure pattern. This component toggles the visibility of the popover by clicking on the trigger.

**Behavior**

- The trigger for the toggletip receives focus
- The trigger toggles the visibility of the toggletip through:
  - Mouse click
  - Keyboard enter, space
- The toggletip may be dismissed on:
  - Keyboard Escape
  - Blur
  - Click away from toggletip or trigger
- Customizable icon?

**Public API**

We will deliver several components for Toggletip:

- Toggletip
- ToggletipLabel
- ToggletipButton
- ToggletipContent

Existing API

```jsx
<Tooltip triggerText="..." renderIcon={...}>
  <p>Lorem ipsum</p>
  <div className="bx--tooltip__footer">
    <Link />
    <Button />
  </div>
</Tooltip>
```

Proposed API

```jsx
<ToggletipLabel>Label</ToggletipLabel>
<Toggletip>
  <ToggletipButton>ICON</ToggletipButton>
  <ToggletipContent>
    ...
    <ToggletipActions>
    </ToggletipActions>
  </ToggletipContent>
</Toggletip>
```

**Open questions**

- How do we allow this component to be flexible for different targets and layouts?
- What specific layouts would we like to provide?
- What layering rules should this component adhere to? Is it always on top, z-index wise?

**Links & Resources**

joshblack:
Closes #10280

Adds the toggletip set of components to this project. This is the initial API Design for `Toggletip` and its corresponding components. It includes test stubs but no tests yet for the implementation while we go through API review 👀 

#### Changelog

**New**

- Add Toggletip components
- Add `useWindowEvent` to `useEvent`

**Changed**

- Update `useEvent` to supports refs

**Removed**

#### Testing / Reviewing

- Go through the API design for this component
  - Does it make sense?
  - Does it follow our standards?
  - Do you see any problems with it?
- Go through the test stubs, do you see any behavior that should be covered that isn't?


packages/react/src/components/Toggletip/index.js
```diff
@@ -0,0 +1,252 @@
...
+  children: PropTypes.node,
+
+  /**
+   * Provide a custom class name to be added to the outermost node in the
+   * component
+   */
+  className: PropTypes.string,
+
+  /**
+   *
```
tay1orjones:
```suggestion
   * Specify if the toggletip should be open by default
```

packages/react/src/internal/useEvent.js
```diff
@@ -21,10 +21,33 @@ export function useEvent(element, eventName, callback) {
...
     element.addEventListener(eventName, handler);
 
     return () => {
       element.removeEventListener(eventName, handler);
     };
-  }, [element, eventName]);
+  }, [elementOrRef, eventName]);
+}
+
+export function useWindowEvent(eventName, callback) {
```
tay1orjones, line 33: This will have differing behavior between React 17 and 18. State updates originating from native event handlers in 17 are not automatically batched, but they will be in 18. https://github.com/reactwg/react-18/discussions/21

It shouldn't be a problem, but wanted to mention it.

tay1orjones (approved):
Looks good to me! Composing `ToggletipLabel` outside of `Toggletip` looks a bit strange but I like the preference towards flexibility.

packages/react/src/internal/useEvent.js
joshblack: Thanks for the heads up!

> This shouldn't be necessary because we handle discrete events originating outside of the React event system the same way, by reading window.event. There may be some use cases that don't automatically work though, and for those you can use flushSync. It should be used sparingly though, because if it's done too much you'll just effectively force React back into sync mode.

Found this comment in there, seems like we're okay to move forward but need to be aware of this / add it in for specific situations?

tay1orjones: @joshblack Yeah exactly 👍

laurenmrice (changes requested):
- The hover state for the icon looks like its `$icon-primary`, so that is correct. 👍
-----
- [x] There should be 8px between the top of the caret tip and info icon.
toggletip spacing

- [x] The focus on the link and button inside of the popover container need to be using` $focus-inverse`.
toggletip focus

------

- Also just a question, are direction for top, bottom, left, and right going to be provided eventually? Or should they be a part of this PR.

abbeyhrt (approved):
Looks good to me!!

joshblack:
@laurenmrice thanks for the review! Just updated with the changes. I also included a "playground" story with the directions (although the "top" option will run into the top of the page, will figure out a good way to center stories in a future PR).

Separately, it looked like the button was using the correct token but let me know if you still are seeing that problem in the latest update!

laurenmrice:
@joshblack The $focus-inverse looks good in the light themes. But for the dark themes, the button still gets incorrect focus styling. The focus around the button should be blue and not white. 

In storybook:
Screen Shot 2022-01-25 at 11 08 53 AM

joshblack:
@laurenmrice great point, thanks! I'll make an update today

joshblack:
@laurenmrice just pushed up the change so that the button uses `$focus-inverse`. Let me know if that worked!

laurenmrice (approved):
It looks good to me now! 👍

Side note: Eventually design is going to add another focus color token for $focus-inset that is inverse for these situations when a button appears on an inverse background. It will help improve some of the color contrast.

PR #10385, opened 2022-01-11
feat(tabs): make debounce wait configurable prop

tay1orjones:
Refs #9698, #10120

In some situations the the call stack exceeded bug identified in #9698 still happens. This PR adds a new prop so the scroll handler's debounce wait ms is configurable for consumers who identify the need for a more aggressive debounce.

#### Changelog

**New**

- tabs: provide prop to configure scroll handler debounce wait ms

**Changed**

- tabs: increase default debounce wait by 25ms to 150ms

#### Testing / Reviewing

- The original issue was only present in Safari. Particularly with a trackpad as an input device. 
- In Safari, Check the `tabs` stories in the `carbon-components-react` storybook. 
- Scroll back and forth multiple times (10-20 times) and ensure that React does not throw an error; `Error: Maximum update depth exceeded.`


dabrad26 (approved):
Looks good. Thanks!

dakahn (approved):

aledavila (approved):

PR #10393, opened 2022-01-12
feat(fonts): add google font support via flag

### LINKED ISSUE #10353: Add support for hosted option through Google Fonts
As part of our IBM Plex for v11, we added support for using fonts directly from the `@ibm/plex` package. However, we would like to keep our hosted option available that is available in v10. This will help teams migrate from v10 to v11 and also provides an alternate setup out-of-the-box when using Carbon.

The default behavior in v11 will be to use fonts directly from `@ibm/plex`, to enable hosting we should support a flag in the config package that enables hosted fonts and disables custom `@font-face` blocks from the package.

The difference between hosted and non-hosted is simply the value of the `url()` in the `@font-face` block. Most likely we could use our resolver to implement the logic for getting the correct `url` path when hosted is set to true in config.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/10353

Adds in out-of-the-box support for Google fonts by enabling the `$use-google-fonts` flag. It has been turned on by default for testing purposes. 

If turned on, it will create the `src` field from google data, rather than resolve to the installed plex package. 

#### Changelog

**New**

- `$use-google-fonts` flag, which if enabled will provide font files directly from the Google static server rather than the `@ibm/plex` package itself. 
- Map of all the google data in `@styles/scss/fonts/_src.scss` 

#### Testing / Reviewing
First, navigate to the IBM Plex story in the `carbon-react` storybook. Then, in dev tools, go to the Network tab, hit record, and do a hard refresh. The `woff` files should resolve to Google servers. If testing locally, flip the flag to `false` and see that the path is now set to the locally installed font files. 

Before merging the flag should be set to `false` by default.

### DIFF
--- a/packages/styles/scss/__tests__/__snapshots__/config-test.js.snap
+++ b/packages/styles/scss/__tests__/__snapshots__/config-test.js.snap
@@ -8,5 +8,6 @@ Object {
   "font-display": "swap",
   "font-path": "~@ibm/plex",
   "prefix": "cds",
+  "use-google-fonts": false,
 }
 `;
--- a/packages/styles/scss/_config.scss
+++ b/packages/styles/scss/_config.scss
@@ -36,7 +36,7 @@ $font-display: 'swap' !default;
 /// @group config
 $font-path: '~@ibm/plex' !default;
 
-/// Specify if the fonts should be provided by Google Fonts
+/// Specify if IBM Plex should be provided by Google Fonts
 /// @access public
 /// @type String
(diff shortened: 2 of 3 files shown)


abbeyhrt (approved):
Looks good to me! Pulls in fonts from google as expected 👍

jnm2377 (approved):

jnm2377:
🔥 great work on this!

PR #10400, opened 2022-01-12
fix(slider): ensure slider and input are both labelled by the label

tay1orjones:
Refs https://github.com/carbon-design-system/carbon/pull/10252 #10172

1. The `label` is associated with the slider via `htmlFor`
1. The slider no longer has the redundant `aria-labelledby`
1. The input is `aria-labelledby` the `label` when a more explicit `ariaLabelInput` is not passed
1. If a `ariaLabelInput` is passed, the `input` does not receive `aria-labelledby` and instead receives `aria-label={ariaLabelInput}`

#### Changelog

**Changed**

- slider: improve aria labels

#### Testing / Reviewing

- View the storybook to see that the above bullet list statements are true. It'll be slightly different with the v11 flag since the `ariaLabelInput` doesn't receive a default.

### DIFF
--- a/packages/react/src/components/Slider/Slider.js
+++ b/packages/react/src/components/Slider/Slider.js
@@ -603,10 +603,7 @@ export default class Slider extends PureComponent {
             ? classNames(`${prefix}--form-item`, className)
             : `${prefix}--form-item`
         }>
-        <label
-          htmlFor={`${id}-input-for-slider`}
-          className={labelClasses}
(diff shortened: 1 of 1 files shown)


tay1orjones:
@mbgower I think this hits the mark you were aiming for in your original issue?

abbeyhrt (approved):
Looks good to me!

dakahn (approved):

Issue #5702, opened 2020-03-25
Add optional character count to text input


kaylarakower:
### Summary

Text input with optional character count

![label + character count](https://user-images.githubusercontent.com/48567372/77562853-edae6300-6e96-11ea-94f7-38e039f8783c.jpg)

[Image and sketch file](https://ibm.box.com/s/z6h89nz5nrx5h6j9ajq7ubns75qvzlpe)

### Justification

To create and edit pages on IBM.com, content authors need to know the character limit for all pieces of text, whether it is a single line of text or a paragraph. Currently, character counts are only supported for a text area, Issue #1672  (However, it is missing in [storybook](https://react.carbondesignsystem.com/?path=/story/textarea--default) ). 

When a user is typing into a form field which has a maximum value, that value is communicated to the user, and if the limit is reached, they are informed whilst typing.

### "Must have" functionality
(Copied from #1672 )
Tell the user ahead of time what the value is
Tell the user clearly when they have hit the limit (and stop them entering text)

### Contributors
Team: Marketing Platforms - CMS - EASE
Designer: Kayla Rakower, kayla.rakower@ibm.com

emyarod:
this feature is all but finished. the required component changes to correctly implement this were not merged in and this effort was deprioritized, but when this becomes a focus in a project cycle the work is most of the way there with a few updates (related https://github.com/carbon-design-system/carbon/pull/2745 https://github.com/carbon-design-system/carbon/pull/3054 https://github.com/carbon-design-system/carbon/pull/3106 https://github.com/carbon-design-system/carbon/pull/3097)

Mikadv:
Ah year back now. Any news on this?

TannerS:
Bump, is this actively being worked on? @joshblack  or @tay1orjones ? like is this in the works or road map for carbon or cloud pal?

TannerS:
https://github.com/carbon-design-system/carbon/issues/5702

joshblack:
Hi there @TannerS and @Mikadv! 👋  This is not currently on our roadmap but it is a candidate for Q2 2022 after we ship our v11 release 👍

TannerS:
Issue has been merged

tay1orjones:
@TannerS Great! In the future if you use the specific [autolinking syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) GitHub will auto close the related issue(s) when the PR is merged.

[...]

Issue #1672, opened 2019-01-21
[Text input] Add character limit/counter


tomlroach:
### Summary

Text fields and text areas may have a maximum number of characters. This character limit might need to be communciated: 
1. to users ahead of time, so they can format their input,
2. Whilst a user is entering their input so they know how many characters remain
3. In the form of a validation error when the user reaches the limit

Design usage guidelines and/or enhancements to the components to support that design would be desirable. 

### Justification

It's a common pattern which you can see on a number of IBM products and tools and would be good to standardise. 

### Desired UX and success metrics

When a user is typing into a form field which has a maximum value, that value is communicated to the user, and if the limit is reached, they are informed whilst typing. 

### "Must have" functionality
Tell the user ahead of time what the value is
Tell the user clearly when they have hit the limit (and stop them entering text) 

Might also include a "x characters remaining" indicator so users know how many more they can enter.

emyarod:
cc @IBM/carbon-designers for a spec or mockup on text input and textarea character limit UX

jeanservaas:
@tomlroach this is great added functionality and we'll need to explore and possibly text a pattern before settling on a final spec for this:

- settling on a max
- how to deal with characters remaining
- whether to actually limit the characters or colorize the overflow as a warning

After the v10 release, we will start prioritizing opportunities to expand on existing functionality and get stuff like this into the pipeline. 

good discussion of this here and I really like the text colorization example at the bottom:
https://ux.stackexchange.com/questions/13055/character-limits-on-fields-pros-and-cons-and-best-practices

aagonzales:
Adding general text area for devs. I'd be open to doing the character count in a different way. I know some times people prefer it counting down instead up. I know Andy Nelson had some [ideas/opinions](https://ibm-cloudplatform.slack.com/archives/C0M053VPT/p1519839354000302) on it 

![text area - white theme](https://user-images.githubusercontent.com/11670886/51865526-25119600-230c-11e9-9a7c-4ed7840217c5.png)

aagonzales:
Could put it all in the helper text area? 
![image](https://user-images.githubusercontent.com/11670886/51865847-2099ad00-230d-11e9-8843-e48705df0765.png)

connor-leech:
https://github.com/carbon-design-system/carbon-website/issues/1477 I made an issue about the same thing, added to website repo for site guidance

emyarod:
has the spec for this been finalized?

asudoh:
[...]

PR #10442, opened 2022-01-13
feat(textarea): added char counter

TannerS:
Closes #

https://github.com/carbon-design-system/carbon/issues/1672
https://github.com/carbon-design-system/carbon/issues/5702

#### Changelog

**New**

Adds two new props to enable a textarea counter, and an actually max counter limit. Added some styles too.

#### Testing / Reviewing

Added functionality to test live changes in storybook and added unit test. I am all for people helping me look over my unit test for feedback

Added
1) Styles
2) Coded changes (2 props)

Behavior
1) Ability to enable and disable the counter and re-render
2) Follows the text areas maxlength attribute so the user can only type up until it hits the max then does not allow anymore (no error validation option) 
3) both props have to be enabled for this to work. I originally thought of just adding a counter prop which will turn on the counter but I then thought to give the user the ability to turn it off and on without having to erase the counter data BUT this can be discussed in this PR
4) The counter should use the same styles as the label to look the same

Test
1) Added unit test. I did have issues finding a way to write a test to grab the value after adding text. Even after looking at many ways online so my test may be limited.


TannerS:
@abbeyhrt fyi

packages/components/src/components/text-area/_text-area.scss
```diff
@@ -99,6 +99,14 @@
       color: transparent;
     }
   }
+
+  .#{$prefix}--text-area__counter-wrapper {
+    width: 100%;
+  }
+
+  .#{$prefix}--text-area__counter {
+    float: right;
```
abbeyhrt: Would it be possible to use flex box to position the text area counter instead of floats?

abbeyhrt (changes requested):

packages/components/src/components/text-area/_text-area.scss
TannerS: you wish has been granted

packages/react/src/components/TextArea/TextArea.js
```diff
@@ -102,7 +119,10 @@ const TextArea = React.forwardRef(function TextArea(
           ? classNames(`${prefix}--form-item`, className)
           : `${prefix}--form-item`
       }>
-      {label}
+      <div className={`${prefix}--text-area__counter-wrapper`}>
+        {label}
+        {counter}
+      </div>
```
tw15egan: I think we may want to do something like 

```suggestion
{label}
{enableCounter && (
    <div className={`${prefix}--text-area__counter-wrapper`}>{counter}</div>
)}
```

so that the `text-area__counter-wrapper` class is only applied when the `enableCounter` prop is provided.

packages/styles/scss/components/text-area/_text-area.scss
```diff
@@ -102,4 +102,12 @@
       color: transparent;
     }
   }
+
+  .#{$prefix}--text-area__counter-wrapper {
+    width: 100%;
+  }
+
+  .#{$prefix}--text-area__counter {
+    float: right;
```
tw15egan: These styles don't seem to be applied to any element

packages/styles/scss/components/text-area/_text-area.scss
```diff
@@ -102,4 +102,12 @@
       color: transparent;
     }
   }
+
+  .#{$prefix}--text-area__counter-wrapper {
+    width: 100%;
```
tw15egan, line 108: Should these styles be the same as the other style block in `packages/components`?

packages/react/src/components/TextArea/TextArea.js
```diff
@@ -139,6 +159,11 @@ TextArea.propTypes = {
    */
   disabled: PropTypes.bool,
 
+  /**
+   * Specify whether to display the character counter
+   */
+  enableCounter: PropTypes.bool,
```
tw15egan, line 168: Since a `maxCount` is needed for the counter to display, is this prop necessary? Can we toggle the counter simply with the existence of a `maxCount` prop? Is there a use case where you would want a character count, bit not have a max count?

packages/react/src/components/TextArea/TextArea.js
TannerS: make sense!

TannerS: Well with the latest update, i wanted to wrap the label and the counter in a div to use flex box to position them relative to each other, i think doing this would break that but i would need to test, what are your thoughts on that? wrapping both and using flexbox to position it or wrapping the element individually and finding some other styling to making it huig the right side? @tw15egan

packages/react/src/components/TextArea/TextArea.js
TannerS: @tw15egan I did think about that originally, but I wanted to give the user the option of turning it off and on at any time without having to remove or change the number in the counter. I can't think of any use case at the moment, but my thoughts are if they wanted a way to enable and disable it dynamically, they would have to pass in null or remove the prop, and I think i did see issues where it would show up as undefined for max count instead of as value, but if you have other thoughts I will do what the team thinks is best!

packages/styles/scss/components/text-area/_text-area.scss
TannerS: you are right, sorry, forgot to update the second file to match the flexbox update! will fix

packages/react/src/components/TextArea/TextArea.js
tw15egan: Ah okay, I thought they were going to stack. My mistake. Yes, we will need to have a wrapper around it then. Since it will always be there, perhaps we should just name it `text-area__label-wrapper` so it makes sense even if the counter is not present.

Then you can just do a `display: flex` and `justify-content: space-between` to the wrapper to space them.

packages/styles/scss/components/text-area/_text-area.scss
TannerS: correct this was removed in last update, forgot to update this package also

packages/react/src/components/TextArea/TextArea.js
tw15egan: That makes sense, I'm fine either way I was just curious 👍🏻

packages/react/src/components/TextArea/TextArea.js
TannerS: cool, will change!

TannerS:
@tw15egan updated

TannerS:
@tw15egan @abbeyhrt  my PR failed a unit test unrelated to my code, some timeout error, how do re-run the test check?

tw15egan:
@TannerS can you try running `yarn test -u` at the root, add any changes, and push it up?

TannerS:
@tw15egan  thanks! ya that updated a file. 

I am a big confused, what does that command do? it seemed to update by adding the prop `autoOrientation` to this file `packages/react/src/components/Tooltip/Tooltip.js`

which is part of an active PR `https://github.com/carbon-design-system/carbon/pull/10611/files` 

so I am a bit confused on what that commands does and why did it add that prop that was added to a PR not yet merged?

tw15egan:
`yarn test -u` just updates the snapshot tests, not sure why it would pull in code from a different PR

TannerS:
> `yarn test -u` just updates the snapshot tests, not sure why it would pull in code from a different PR

@tw15egan  I thought maybe I messed up a merge or something but it added that prop without the eslint comment like in the PR above, so I am not sure... 

either way, it may been what I needed to fix it, so thank you!

TannerS:
> `yarn test -u` just updates the snapshot tests, not sure why it would pull in code from a different PR

ok, maybe i just messed up and the merge i did left out a file, either way, it should be fixed, so thanks!

abbeyhrt:
@TannerS This looks great so far! One thing I noticed, messing around with the story, is that you can set a maxCount when enableCounter is false and it does prevent the user from typing passed that maxCount. Is that intended? It seems a bit weird to have a limit that isn't communicated to the user but there definitely could be a use case you have in mind that I don't :)

TannerS:
Hey guys that might be a bug let me test it out in just a bit here thank
you for pointing that out

TannerS:
@abbeyhrt @laurenmrice hey guys, ya it was a bug, I pushed a fix, let me know if the way I fixed it is acceptable / carbon standards !

packages/react/src/components/TextArea/TextArea.js
```diff
@@ -45,11 +52,12 @@ const TextArea = React.forwardRef(function TextArea(
       }
     },
     ref,
+    ...(enableCounter && { maxLength: maxCount }),
```
TannerS: @abbeyhrt @laurenmrice  enabled counter being false but max count still enabled fix here

abbeyhrt: Awesome, thanks for fixing that! This definitely works but conditional spreads like this can be a bit hard to debug later on, would you mind instead, somewhere after `textAreaProps` is defined doing something like: 
```js
if (enableCounter) {
  textareaProps.maxLength = maxCount;
}
```
?

TannerS: sure! and pushed! @abbeyhrt !

packages/react/src/components/TextArea/TextArea.js
```diff
@@ -47,9 +54,13 @@ const TextArea = React.forwardRef(function TextArea(
     ref,
   };
 
+  if (enableCounter) {
+    textareaProps.maxLength = maxCount;
+  }
+
```
TannerS, line 60: @abbeyhrt

abbeyhrt (approved):
🎉 🎉  Looks great to me! Thank you for the contribution 😃

laurenmrice (approved):
Looks great! Thank you for contributing this! ✅ ⭐️

packages/components/src/components/text-area/_text-area.scss
```diff
@@ -99,6 +99,12 @@
       color: transparent;
     }
   }
+
+  .#{$prefix}--text-area__label-wrapper {
+    display: flex;
+    width: 100%;
+    justify-content: space-between;
+  }
```
tw15egan, line 107: One last thing! Can you just duplicate these styles over to the `packages/styles`? That will be the V11 style package, and I want to ensure this feature is available in V11.

TannerS: @tw15egan correct me if I am wrong but I thought I did?

https://github.com/carbon-design-system/carbon/pull/10442/files#r791899919

TannerS: is the the correct path/file?

tw15egan: You're right, my mistake 😅  Looks great!

tw15egan (approved):

Issue #10443, opened 2022-01-13
[a11y]: FileUploaderDropContainer's input needs to be moved outside label


TannerS:
### Package

carbon-components-react

### Package version

v10.50.0

### Automated testing tool and ruleset

IBM Equal Access Accessibility Checker

### Description

the FileUploaderDropContainer has it's input within it's label which sparks a a11y issue 

`3.3.2 Labels or Instructions: Text input is nested in label such that input precedes the label text`

I think I fix for this is moving it outside the label 

<img width="1677" alt="image" src="https://user-images.githubusercontent.com/8866319/149418455-a9c13e5d-5bbe-4618-8274-333dd16d3c23.png">

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/8866319/149418480-72adf362-d38e-4e5d-832a-2b60432f6709.png">

### WCAG 2.1 Violation

3.3.2 Labels or Instructions

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader-drop-container

### Steps to reproduce

You can view it by ruinning the IBM Equal Access Accessibility Checker in the storybook for the FileUploaderDropContainer

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader-drop-container

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

PR #10444, opened 2022-01-13
fix(fileuploaderdropcontainer): moved input out of label to fix a11y bug

TannerS:
Closes #

https://github.com/carbon-design-system/carbon/issues/10443

#### Changelog

**Changed**

Moved the input attribute outside the label

#### Testing / Reviewing

I tested it the same in the story  book against the working copy and ran unit test. Far as I know, it behaves and works the same.

Before:
image

After:
image


TannerS:
@joshblack ya sure once I get back on a laptop, but if it helps, it's the only expanded one. I just forgot to click on more details. From what I gathered, it was something about the element appearing before the label and that was due to the element being inside the label element that had the assigned...I think it was the html forto? Or what ever it is that tells what element belongs to which label

joshblack:
@TannerS ah I see it now in the issue, sorry I didn't see it earlier!

TannerS:
@joshblack all good, let me know if u need a screen shot still. From my testing, it seems to behave the same even moving it outside the label but I wanted more eyes to be sure I'm not missing something

joshblack (approved):

dakahn (approved):

Issue #10345, opened 2022-01-05
MultiSelect: Show selected items in the MultiSelect itself


adamalston:
### Summary

Show selected items in a MultiSelect field removing the need to open the dropdown to see what is selected.

### Justification

Users and engineers have requested this feature in alpha and beta testing.

If a MultiSelect is disabled and there are items within that are selected, how is a user supposed to know what those items are?

### Desired UX and success metrics

Template:
![Template](https://user-images.githubusercontent.com/18297826/148283354-990ca8e8-1900-48d3-8eef-1eb9ec1ccd42.png)

Concept (expanded):
![concept-expanded](https://user-images.githubusercontent.com/18297826/148283561-14fd95f4-50aa-4b61-8b45-a4c257b241ff.png)

Concept (collapsed):
![concept-collapsed](https://user-images.githubusercontent.com/18297826/148285340-99c18ca0-9283-48eb-8340-ca638b29f193.png)

Obviously the design in this concept may need to be changed but the functionality is there - the selected items are shown within the field itself without the need to expand the dropdown to see what is selected.

To generate this concept I put in a workaround to concatenate the display names of the selected items into a label:
```tsx
const label = initialSelectedItems
  .map((item: MultiSelectItem) => item.display)
  .join(', ');

return <MultiSelect label={label} initialSelectedItems={initialSelectedItems} ... />;
```

### Specific timeline issues / requests

Yes, this request is related to an upcoming release which I probably cannot detail here.

### Available extra resources

The images above.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/13435
- [ ] MultiSelect (selected items in field): implementation
- [ ] MultiSelect (selected items in field): website docs
- [ ] MultiSelect (selected items in field): figma kit
```

abbeyhrt:
@adamalston Thank you for making this proposal! Would you be able to provide some user scenarios of when a Multiselect would be disabled but would have items selected?

adamalston:
Here is a screenshot of a field that is read only - all of the options are disabled and some of them are selected by default:

![user interface](https://user-images.githubusercontent.com/18297826/148825954-8bc4699a-4de9-4d6f-876d-44332f37ba23.png)

[...]

PR #10446, opened 2022-01-13
feat(multiselect): added selections in label

TannerS:
Closes #

This is a work in progress PR (really just needs approval and test if allowed), this is a possible solution for this proposal, if this proposal is even chosen to be done or accepted. I thought that, it may be a decent proposal so I would try it out.From my testing, the code works but I still add unit test depending on if this is even wanted: #10345
Change log

New / Changed

Multiselect can now display selected items in the multi select itself if enabled by a new prop
Testing / Reviewing

Tested in story book but missing unit test unit proposal is accepted

Behavior:

    label will display as normal
    If new prop enableSelectionLabel is selected, it will display the selected item in the multiselect going of the itemToString function or the default label property
    if no items are selected, and enableSelectionLabel is checked, it will default to showing the label


packages/react/src/components/MultiSelect/MultiSelect.js
```diff
@@ -272,7 +273,9 @@ const MultiSelect = React.forwardRef(function MultiSelect(
             />
           )}
           <span id={fieldLabelId} className={`${prefix}--list-box__label`}>
-            {label}
+            {itemToString && enableSelectionLabel && selectedItems.length > 0
```
abbeyhrt: Would you mind pulling this out into a helper function? We try to not inline conditionals like when they are checking for multiple conditions, thank you!

TannerS: Done

jnm2377:
Hey @TannerS thanks for opening this PR. I think we still need some design research and input in the original issue before we move forward with implementing in code. I'm gonna close this for now until our designers have had a chance to take a look at the proposed designed changes and given feedback.

PR #10486, opened 2022-01-19
docs(carbon-react): add stories for v11 site

### LINKED ISSUE #10485: Verify all stories appear in v11 storybook
Currently there are a few components missing from the v11 storybook. They need to be present in the new v11 storybook so they can be audited. For example, the v11 storybook's new theme switcher will be very handy to audit and ensure design token consistency across the different themes. 

Here's a tentative list:

- [ ] Accordion
- [ ] Breadcrumb
- [ ] Aspect Ratio
- [ ] CopyButton
- [ ] ErrorBoundary
- [ ] FormGroup
- [ ] FormLabel
- [ ] Link
- [ ] Loading
- [ ] OrderedList
- [ ] OverflowMenu
- [ ] UnorderedList

abbeyhrt:
Closes #10485 

Adds missing stories from the v11 storybook. Right now, these are just the v10 stories copied/pasted over with updated import paths. 

#### Testing / Reviewing

Make sure all of the added stories are showing up in the `carbon-react-next` storybook.

### DIFF
--- a/packages/react/src/components/Accordion/next/Accordion.stories.js
+++ b/packages/react/src/components/Accordion/next/Accordion.stories.js
@@ -1,0 +1,161 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/* eslint-disable no-console */
+
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import {
+  withKnobs,
+  boolean,
+  number,
+  select,
+  text,
+} from '@storybook/addon-knobs';
+import {
+  default as Accordion,
+  AccordionItem,
+  AccordionSkeleton,
+} from '../../Accordion';
+import Button from '../../Button';
+import mdx from '../Accordion.mdx';
+
+export default {
+  title: 'Components/Accordion',
+  component: Accordion,
+  subcomponents: {
+    AccordionItem,
+    AccordionSkeleton,
+  },
+  decorators: [withKnobs],
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+export const accordion = () => (
+  <Accordion>
+    <AccordionItem title="Section 1 title">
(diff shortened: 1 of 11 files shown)


packages/react/src/components/Accordion/next/Accordion.stories.js
```diff
@@ -0,0 +1,161 @@
...
+
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import {
+  withKnobs,
+  boolean,
+  number,
+  select,
+  text,
+} from '@storybook/addon-knobs';
```
tay1orjones, line 18: Knobs should be removed as they don't work in the new storybook. We have to use controls instead. 

We can totally defer the addition of controls to get this PR in, but let's open an issue for it and add it to this/next sprint.

tay1orjones (approved):
Knobs aside, all of the added stories are showing up in the carbon-react-next storybook 👍

packages/react/src/components/Accordion/next/Accordion.stories.js
abbeyhrt: Sounds good! Do we have a good working example using controls? I'm down to switch from them in this PR but I'm just not sure how controls work vs knobs

tay1orjones: @abbeyhrt Cool! Yep, there are a few. Any `*.stories.js` files with usages of "args" or "argTypes".

- `Theme.stories.js` is a pretty solid baseline example, `Tooltip.stories.js` as well.
- `Notification.stories.js` uses some additional parts of the API to apply global controls/args to all stories and additionally applies some extra ones to specific stories.

Also here's a [reference](https://storybook.js.org/docs/react/essentials/controls) from the storybook docs

joshblack:
@abbeyhrt could we have the storybook config point to the files in the `packages/react` directory? That way we wouldn't have to worry about keeping things in sync and any changes we make (like for controls) would benefit v10 and v11

abbeyhrt:
Going to close this to try out a different approach

PR #10494, opened 2022-01-21
feat(upgrade): run migrations from upgrade cli

### LINKED ISSUE #10466: Add ability to run codemods to `@carbon/upgrade`

tay1orjones:
Closes #10466

This PR updates the `@carbon/upgrade` CLI to be able to run one or more `migrations` as part of an upgrade flow. These `migrations` contain a `migrate` function that can run jscodeshift codemods.

The following is now possible:
- `@carbon/upgrade`: upgrade and run all available migrations for a given upgrade
- ~`@carbon/upgrade migrate`: prompt a list of all available migrations, select one to run~ removed based on feedback
- `@carbon/upgrade migrate list`: list all available migrations
- `@carbon/upgrade migrate <migration-name>`: run a specific migration

#### Changelog

**New**

- Provide new `migrate` command to run and list individual migrations
- ~Add a new `imports-to-unified-package.js` codemod to rewrite imports from `carbon-components-react` to `@carbon/react`~ moved to https://github.com/carbon-design-system/carbon/pull/10523

**Changed**

- Modify the upgrade flow to run all available migrations for a given upgrade flow by default
- Move `transforms` from `codemods` package to live within the `upgrade` package

**Removed**

- Remove `codemods` package

#### Testing / Reviewing

- Try out the CLI locally by:
  - Pulling down this PR
  - Run `cd packages/upgrade`
  - Run `yarn build` to build the CLI
  - Run `cd fixtures/sample-project` to use our sample project fixture
  - Run `../../bin/carbon-upgrade.js --help` to verify the CLI is up-and-running
  - Take a look at the options / capabilities of what's there to see if it makes sense
  - Run `../../bin/carbon-upgrade.js` and choose an upgrade path. Verify the CLI runs all the available migrations for that upgrade path. In the case of `v11: carbon-components`, two stub migrations should be called and log out to the console.
  - Run `../../bin/carbon-upgrade.js migrate list` to see a list of migrations should be shown
  - ~Run `../../bin/carbon-upgrade.js migrate` to see a prompt of migrations should be shown allowing a single selection~ removed based on feedback


joshblack:
Was thinking about this the other day, one thing that could simplify our upgrade paths is that we blanket run migrations associated with the upgrade (in this case we would not prompt users to select migrations).

For those who would like to select a specific migration, we could have a `migrate` command that is like:

- `@carbon/upgrade migrate list`, list all available migrations
- `@carbon/upgrade migrate <name>`, run the migration with the given name

tay1orjones:
@joshblack Great point! Having to navigate through the upgrade flow to run a migration is cumbersome.

I just pushed an update to run all migrations with an upgrade, and provide separate `migrate` commands to run migrations individually.

packages/upgrade/src/commands/migrate.js
```diff
@@ -0,0 +1,172 @@
...
+
+  const { cwd } = options;
+  const migrations = upgrades
+    .filter((upgrade) => {
+      return upgrade.migrations && upgrade.migrations.length > 0;
+    })
+    .map((upgrade) => {
+      return upgrade.migrations;
+    })
+    .flat();
```
joshblack:
```suggestion
    .flatMap((upgrade) => {
      return upgrade.migrations;
    });
```

I think this will work as an alternative (just looking at it real quick)

packages/upgrade/src/commands/upgrade.js
```diff
@@ -90,6 +90,16 @@ export async function upgrade(options, availableUpgrades = []) {
...
     }
   }
+
+  const migrations = upgrade.migrations;
+  if (migrations && migrations.length > 0) {
+    for (const migration of migrations) {
+      logger.verbose('running migration: %s', migration.name);
+
+      // Run each migrate function sequentially/non-async
+      migration.migrate(workspace.directory);
```
joshblack:
```suggestion
      await migration.migrate(workspace.directory);
```

Does it make sense to run these in series as `async` chunks? In the case that something in the `migrate` function is async (like reading/writing files)

tay1orjones: @joshblack Yes, I'll also update the migrate command to do this as well

packages/upgrade/src/commands/migrate.js
tay1orjones: @joshblack Good eye, thanks!

packages/upgrade/package.json
```diff
@@ -14,8 +14,12 @@
   "bugs": "https://github.com/carbon-design-system/carbon/issues",
   "files": [
     "bin",
-    "cli.js"
+    "cli.js",
+    "transforms"
   ],
+  "installConfig": {
+    "hoistingLimits": "workspaces"
+  },
```
tay1orjones: This was added to ensure that `jscodeshift` is available in the local workspace's `node_modules`. By default yarn was hoisting it up to the root and `npm-which` couldn't find it for some reason. This only impacts this workspace and shouldn't impact other workspaces/packages.

packages/upgrade/src/cli.js
```diff
@@ -57,6 +62,34 @@ export async function main({ argv, cwd }) {
     })
   );
 
+  cli.usage('Usage: migrate').command(
```
joshblack: Not sure if this was intentional, but `usage()` is used to set the usage info for the cli as a whole versus a specific command.

As a result, this will overwrite the `Usage; $0 [options]` above.

joshblack:
```suggestion
  cli.command(
```

packages/upgrade/src/jscodeshift.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import which from 'npm-which';
+import spawn from './spawn';
+import path from 'path';
+
+let jscodeshift;
+
+function run(transformFileName, options) {
+  console.log(options);
```
joshblack: We should probably either remove this or add it at the verbose level

packages/upgrade/src/jscodeshift.js
```diff
@@ -0,0 +1,54 @@
...
+  if (!jscodeshift) {
+    jscodeshift = which.sync('jscodeshift', {
+      cwd: __dirname,
+    });
+  }
```
joshblack: My vote for finding the bin path would be to not change the workspace config for hoisting limits and find the bin path directly. Maybe something like:

```js
// tools/jscodeshift.js
import path from 'path';
import spawn from './spawn';

let _jscodeshift;

function getBinPath() {
  if (!_jscodeshift) {
    const directory = path.dirname(require.resolve('jscodeshift'));
    _jscodeshift = path.join(directory, 'bin', 'jscodeshift.js');
  }
  return _jscodeshift;
}

export async function run(options) {
  const { cwd, stdio = 'inherit', parser = 'babel', paths, transform } = options;
  const args = [
    paths,
    `-t=${transform}`,
    `--parser=${parser}`, 
    `--ignore-pattern=**/build/**`,
    `--ignore-pattern=**/dist/**`,
    `--ignore-pattern=**/es/**`,
    `--ignore-pattern=**/lib/**`,
    `--ignore-pattern=**/node_modules/**`,
    `--ignore-pattern=**/storybook-static/**`,
    `--ignore-pattern=**/umd/**`,
  ];

  if (options.print) {
    args.push('--print');
  }

  // Not sure what log level this would map to
  if (options.verbose) {
    args.push('-v');
  }

  if (options.dry) {
    args.push('--dry');
  }

  return await spawn(getBinPath(), args, {
    cwd,
    stdio,
  });
}

// Usage
import jscodeshift from './tools/jscodeshift';

async function command() {
  await jscodeshift.run({
    // options...
  });
}
```

packages/upgrade/src/spawn.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright IBM Corp. 2019, 2019
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+import spawn from 'cross-spawn';
```
joshblack: We may be able to get away with `execa` now for promise support: https://www.npmjs.com/package/execa

packages/upgrade/.gitignore
```diff
@@ -1,2 +1,3 @@
 # Generated by `esbuild` using `yarn build`
 /cli.js
+/fixtures/sample-project/example-imports-to-unified-package.js
```
joshblack: What's going on with this file btw? I wasn't sure if it should be tracked or not

packages/upgrade/src/commands/migrate.js
```diff
@@ -0,0 +1,170 @@
...
+  if (selectedMigrations.length === 0) {
+    logger.info('No migrations selected');
+    return;
+  }
+
+  if (selectedMigrations.length > 0) {
+    for (const migration of selectedMigrations) {
+      await runMigration(migration, workspaces, options);
+    }
+  }
```
joshblack: How would you feel about dropping support for this branch? In other words, only support running `migrate list` or `migrate <name>`

packages/upgrade/src/commands/migrate.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * @param {Array<Migrations>} migrations
+ * @returns {<void>}
+ */
+function listMigrations(migrations) {
+  logger.info('Available migrations:');
+  for (const migration of migrations) {
+    logger.info('%s (%s)', migration.name, migration.description);
```
joshblack: UX nit:

```suggestion
    logger.log('\t- %s (%s)', migration.name, migration.description);
```

Could help to have a plain list without the `[info]` preamble

joshblack (changes requested):

packages/upgrade/.gitignore
tay1orjones: @joshblack The goal was that the file would be added, but then we wouldn't track any further changes to the file. So you could run codemods on it but never accidentally push an updated file containing the output of the codemod.

This doesn't work though. We could use git's `--skip-worktree` instead, but if the file is changed, `git pull` may fail and `git status` will show nothing has changed. Might be best to just comment in the file "don't commit changes to this file"?

packages/upgrade/src/commands/migrate.js
tay1orjones: Totally fine with me!

tay1orjones:
@joshblack Amazing suggestions, thank you! All have been addressed and feedback incorporated. `execa` works great, and getting the bin directly like you proposed was flawless ✨

tay1orjones:
I've removed the `imports-to-unified-package` codemod and put it in it's own PR based off of this PR, https://github.com/carbon-design-system/carbon/pull/10523

packages/upgrade/src/commands/migrate.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * @param {Array<Migrations>} migrations
+ * @returns {<void>}
+ */
+function listMigrations(migrations) {
+  logger.log('Available migrations:');
+  for (const migration of migrations) {
+    logger.log('\t- %s (%s)', migration.name, migration.description);
```
joshblack: Super small nit, I shouldn't have set `\t` earlier lol. It looks so funky

image


```suggestion
    logger.log('  - %s (%s)', migration.name, migration.description);
```

packages/upgrade/package.json
```diff
@@ -30,22 +31,20 @@
...
+    "build": "esbuild src/cli.js --bundle --platform=node --outfile=cli.js --target=node14 --external:jscodeshift",
     "clean": "rimraf cli.js",
     "watch": "yarn build --watch"
   },
   "devDependencies": {
     "chalk": "^4.1.1",
     "change-case": "^4.1.2",
-    "cross-spawn": "^7.0.3",
     "esbuild": "^0.14.10",
     "execa": "^5.1.1",
```
joshblack, line 42: @tay1orjones running into this issue when running a specific codemod:

image

I think this is because `execa` is specified as `^5.1.1` where it is treated as a default export versus a named export in v6+

joshblack (changes requested):

packages/upgrade/package.json
tay1orjones: Agh yeah - sorry this was from when I was looking at upping the dep to v6. Decided to keep it at v5 and forgot to change the usage. I'll push an update.

tay1orjones: @joshblack updated 👍

packages/upgrade/src/cli.js
```diff
@@ -42,8 +46,9 @@ export async function main({ argv, cwd }) {
       type: 'boolean',
     });
 
+  // $0: the default command
   cli.usage('Usage: $0 [options]').command(
-    '$0',
+    ['upgrade', '$0'],
     'upgrade your project',
     {},
     run(async (args) => {
```
joshblack, line 51: I think we'll need to update the options to include `TRANFORM_DIR`, running into a regression with the default command:

image

That points to:

image


I believe this is because `options.TRANSFORM_DIR` is not being piped through when going through the default path.

joshblack (changes requested):

packages/upgrade/src/cli.js
tay1orjones: @joshblack I just pushed an update to fix this. I think `TRANSFORM_DIR` really shouldn't be configurable as it's an internal static refection of where we locate the folder of transforms. 

I moved it to be defined directly where we need to use it, `upgrades.js`

joshblack (approved):
🎉 🚀

dakahn (approved):

PR #10499, opened 2022-01-21
feat(react,carbon-react): add unstable_FlexGrid

### LINKED ISSUE #10474: Add  component for flex grid backwards-compat

joshblack:
Closes #10474 

Add in `FlexGrid` component for a stable flexbox-based Grid implementation. The intention is that teams can keep using `FlexGrid`, `Row`, and `Column` components in v11 before moving to the CSS Grid-based implementation.

#### Changelog

**New**

- Add FlexGrid
- Add CSSGrid
- Add GridSettingsContext and associated setter, getter for the context

**Changed**

- Update Grid to toggle on FlexGrid and CSSGrid based on feature flag
- Update Column to toggle classes based on the current "grid mode"
- Move Grid stories to packages/react/src/components/Grid/next/*

**Removed**

- Remove grid folder in packages/carbon-react

#### Testing / Reviewing

- Verify stable grid implementation works as expected in storybook/locally
- Verify CSS grid implementation works as expected in v11 storybook/locally
- Consider the approach to toggling between behaviors and if there is a simpler option available that we could use instead of authoring new components


packages/react/src/components/Grid/Grid.js
```diff
@@ -5,58 +5,18 @@
...
-  );
+import { CSSGrid } from './CSSGrid';
+import { FlexGrid } from './FlexGrid';
+
+function Grid(props) {
+  const enableCSSGrid = useFeatureFlag('enable-css-grid');
+  if (enableCSSGrid) {
+    return <CSSGrid {...props} />;
+  }
+  return <FlexGrid {...props} />;
```
tay1orjones, line 19: I thought this wouldn't be behind a feature flag anymore and the two could coexist at the same time since they no longer share a name(space)?

I assume the goal is for FlexGrid to eventually be deprecated and removed, but from a usage standpoint this means that for all of v11 users will have to wrap every usage in a `<FeatureFlag>` toggle.

packages/react/src/index.js
```diff
@@ -64,7 +64,12 @@ export FluidForm from './components/FluidForm';
 export FormGroup from './components/FormGroup';
 export FormItem from './components/FormItem';
 export FormLabel from './components/FormLabel';
-export { Grid, Row, Column } from './components/Grid';
+export {
+  Grid,
+  Row,
+  Column,
+  FlexGrid as unstable_FlexGrid,
+} from './components/Grid';
```
tay1orjones, line 72: Should we also introduce a `CSSGrid` export? Or are you thinking we'll use `Grid` as the default implementation - FlexGrid for now in v10 and then in v11 switch it over to be CSSGrid by default?

tay1orjones (approved):
A couple questions, but I think this looks great 🎉 🏅

packages/react/src/components/Grid/Grid.js
joshblack: @tay1orjones I think you're dead-on, I'm thinking the timeline is:

- In v10, we default to `FlexGrid` but will enable `CSSGrid` under the `Grid` namespace
- In v11, we can safely remove `FlexGrid` from `Grid` and rename `CSSGrid` to `Grid`. `FlexGrid` is also an export from the package

packages/react/src/index.js
joshblack: @tay1orjones exactly, `Grid` will be the default in v11 and `FlexGrid` would also be available for the classic behavior.

packages/react/src/components/Grid/Grid.js
tay1orjones: @joshblack Ah, yes that makes more sense now. Sounds solid!

jnm2377 (approved):
This looks great!

Issue #10295, opened 2021-12-15
Verify that `size` changes work as expected across components


joshblack:
In v11, we standardized how our `size` prop worked across components. This issue is for verifying that these work as intended across our components. The standard for the `size` prop is that it is defined as follows:

```
ComponentName.propTypes = {
  size: PropTypes.oneOf([/* sizes */]),
};
```

Where sizes contains a set of the following values:

- `sm`
- `md`
- `lg`
- `xl`

Specifically, we need to go through the following components:

- [x] [Accordion](https://github.com/carbon-design-system/carbon/pull/10509) @sstrubberg 
- [x] [Button](https://github.com/carbon-design-system/carbon/pull/10532) @sstrubberg 
- [x] [ButtonSkeleton](https://github.com/carbon-design-system/carbon/pull/10559) @sstrubberg 
- [x] ComboBox - https://github.com/carbon-design-system/carbon/pull/10640
- [x] ComposedModal
- [x] ContentSwitcher
- [x] ControlledPasswordInput `DEPRECATED`
- [x] [DatePickerInput](https://github.com/carbon-design-system/carbon/pull/10912) @sstrubberg 
- [x] DataTable @sstrubberg 
- [x] Dropdown https://github.com/carbon-design-system/carbon/pull/10640
- [x] DropdownSkeleton https://github.com/carbon-design-system/carbon/pull/10640
- [x] FileUploaderButton @sstrubberg 
- [x] FileUploaderDropContainer @sstrubberg 
- [x] FileUploaderItem @sstrubberg 
- [x] FileUploader @sstrubberg 
- [x] FilterableMultiSelect https://github.com/carbon-design-system/carbon/pull/10640
- [x] Link
- [x] ListBox https://github.com/carbon-design-system/carbon/pull/10640
- [x] OverflowMenu @sstrubberg 
- [x] Menu `Experimental`
- [x] Modal
- [x] MultiSelect https://github.com/carbon-design-system/carbon/pull/10640
- [x] [NumberInput](https://github.com/carbon-design-system/carbon/pull/10930)
- [x] [Pagination](https://github.com/carbon-design-system/carbon/pull/10971) @sstrubberg 
- [x] [PasswordInput](https://github.com/carbon-design-system/carbon/pull/10612) @sstrubberg 
- [x] [Search](https://github.com/carbon-design-system/carbon/discussions/10247#discussioncomment-1788152) @sstrubberg 
[...]

PR #10500, opened 2022-01-21
docs(carbon-react): update glob pattern to include missing stories 

abbeyhrt:
Closes #10485 

Updates `carbon-react` storybook to include all stories from `packages/react` but prefer the `next/*.stories.js` if it exists. 

#### Changelog

**New**

- `*-story.js` stories in v11 storybook 

**Changed**

- Alphabetizes stories

#### Testing / Reviewing

Verify that all components have stories in v11 and that for components that have both a `/next/*.stories.js` file and a `*-story,js` file, the `next` story file takes is what appears (for example: Tabs, TextArea,  Toggle, Button, Combobox, to name a few).

PR #10509, opened 2022-01-24
feat(accordion): update size prop types for v11

sstrubberg:
REF #10295 

- Added a v11 flag for the `size` prop to only accept SM, MD, LG. [REF](https://v11.carbondesignsystem.com/components/accordion/style#sizes). The v10 version will accept SM, MD, LG *and* XL.
- Added 'md' as a default for the `size` prop.
- Added 'false' as a default for the `disabled` prop for better documentation in Storybook.
- Removed defaultProps for `align` prop and made it a default param. Updated the PubicAPI snapshot accordingly.
- Added a story for the v11 Storybook so I could see the changes.

#### Testing
- Make sure everything is snappy in the v11 Storybook and that everything I'm saying above is true.

Something kind of annoying is that when I use the v11 flag in-line with the PropTypes, you'll notice the "custom" values. Not sure if anything can be done about that.
image


joshblack (approved):
Looks great! 🎉 

One quick note, you may not need to include an `Accordion.stories.js` file after: https://github.com/carbon-design-system/carbon/pull/10500 gets merged in as it should pick up `Accordion-story.js` in our next storybook 🎉

abbeyhrt (approved):
Looks good to me!

PR #10523, opened 2022-01-25
feat(upgrade): add codemod, rewrite imports to `@carbon/react`

### LINKED ISSUE #10467: Author codemod to change `import ... from 'carbon-components-react'` to `@carbon/react`

tay1orjones:
Closes #10467 

~Dependent on https://github.com/carbon-design-system/carbon/pull/10494~ _Merged 👍_ 

- [x] add the new codemod
- [x] add tests
- [x] add test fixtures
- [x] run the codemod in the appropriate upgrade paths


packages/upgrade/package.json
```diff
@@ -39,7 +39,6 @@
     "chalk": "^4.1.1",
     "change-case": "^4.1.2",
     "esbuild": "^0.14.10",
-    "execa": "^5.1.1",
```
tay1orjones: I think this came from a garbled up merge bringing in the latest from `main`. I'll push a fix

packages/upgrade/transforms/imports-to-unified-package.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright IBM Corp. 2016, 2020
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
```
joshblack: For the transform name, would it be possible to use a name like: `update-carbon-react-import-to-scoped` or something more specific?

packages/upgrade/transforms/imports-to-unified-package.js
```diff
@@ -0,0 +1,37 @@
...
+ * import { Button } from 'carbon-components-react';
+ * import { Tile, Tooltip } from 'carbon-components-react';
+ * import CodeSnippet from 'carbon-components-react';
+ *
+ * Into:
+ *
+ * import { Button } from "@carbon/react";
+ * import { Tile, Tooltip } from "@carbon/react";
+ * import CodeSnippet from "@carbon/react";
```
joshblack: Is an important part of this codemod that it will transform multiple imports, or are we assuming they will only have one?

packages/upgrade/transforms/imports-to-unified-package.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+'use strict';
```
joshblack:
```suggestion
```

packages/upgrade/transforms/imports-to-unified-package.js
```diff
@@ -0,0 +1,37 @@
...
+
+export const parser = 'babel';
+
+export default function transformer(file, api) {
+  const j = api.jscodeshift;
+
+  return j(file.source)
+    .find(j.ImportDeclaration)
+    .filter((imp) => imp.node.source.value === 'carbon-components-react')
```
joshblack:
```suggestion
    .find(j.ImportDeclaration, {
      source: {
        value: 'carbon-components-react',
      },
    })
```

packages/upgrade/transforms/imports-to-unified-package.js
```diff
@@ -0,0 +1,37 @@
...
+    .forEach((path) => {
+      console.log(path);
+      path.value.source.value = '@carbon/react';
```
joshblack:
```suggestion
     path.get('source').replaceWith(j.stringLiteral('@carbon/react'));
```

Another option for this that's using some of those JSCodeshift APIs we talked about

joshblack (changes requested):
Left some notes, I don't think any of them are deal-breakers though outside of removing the `console.log` from the codemod 👍

packages/upgrade/transforms/imports-to-unified-package.js
tay1orjones: @joshblack My assumption is that they'll have only one, but the codemod will transform multiples if they have them. 

These comments were more to show that different types of imports are supported. Importing default, importing a single named, and importing multiple named are all handled appropriately. I can cut this comment down to be more clear.

packages/upgrade/transforms/imports-to-unified-package.js
tay1orjones: @joshblack Yes! I like that one. I thought mine wasn't that great but ran with it anyway. Thanks for the suggestion 👍

tay1orjones: I might go for `update-carbon-components-react-import-to-scoped` actually to pinpoint `carbon-components-react`

joshblack (approved):

packages/upgrade/transforms/imports-to-unified-package.js
tay1orjones: Updated

packages/upgrade/transforms/imports-to-unified-package.js
tay1orjones: Updated

abbeyhrt (approved):

PR #10525, opened 2022-01-25
fix(theme): update tokens to correct values

### LINKED ISSUE #10292: Update design tokens with latest values
Before our RC0 release, we need to go through the tokens defined in our themes package and verify that the values match the latest in: https://carbondesignsystem.notion.site/final-List-of-new-tokens-c507394e43fe4daa9f0cd0eedb25d1e4

**Checklist**

- [ ] Correct the values for the following tokens
  - [ ] `highlight` should be `blue-80` in `g100`
  - [ ] `toggle-off` should be `gray-60` in `g100`
- [ ] Compare the netlify preview with the final token list in Notion
- [ ] For any changes, update the values for the token in the corresponding theme file under: https://github.com/carbon-design-system/carbon/tree/main/packages/themes/src/next
- [ ] Create a PR with updates

tay1orjones:
Closes #10292 

Updates various tokens for `next` themes, as specified in the root issue: https://github.com/carbon-design-system/carbon/issues/10292#issuecomment-1016631208

#### Changelog

**Changed**

- theme: updated token values

#### Testing / Reviewing

- Check out the elements theme v11 preview - the token changes should be reflected and they should match what's in the [notion doc here](https://carbondesignsystem.notion.site/final-List-of-new-tokens-c507394e43fe4daa9f0cd0eedb25d1e4)


aagonzales:
@tay1orjones Weird, in the code I see the correct token for `layer-hover-01` but in the preview its showing a different token name.

image

aagonzales (approved):
The code itself looks good, its just the preview that's showing a few wrong tokens for some reason 🤷‍♀️ . Going to approve though since the code is correct.

tay1orjones:
@aagonzales Great catch! It looks like `white-hover` and `gray-10-hover` have the same value, so we have to choose which one to use when there is a name collision. The preview site was just grabbing the wrong name for that value. I pushed an update that should fix 👍

dakahn (approved):

abbeyhrt (approved):

PR #10532, opened 2022-01-26
chore(button): size audit

sstrubberg:

- Added FeatureFlag ternary for Button size prop based on guidance provided by design. [REF](https://v11.carbondesignsystem.com/components/button/usage/#button-sizes)
- Removed the `defaultProps` object and used default params [REF](https://github.com/reactjs/rfcs/pull/107)
- Renamed `...other` to `...rest` for consistency with the rest of the codebase.

#### Testing

- Make sure Button is behaving correctly in the v11 & v10 storybooks and that size props are accurately reflected according to the design guidance.
- Ensure I haven't ruined someone's day from the tests I updated.


packages/react/src/components/Button/Button.js
```diff
@@ -124,7 +124,7 @@ const Button = React.forwardRef(function Button(
     return () => document.removeEventListener('keydown', handleEscKeyDown);
   }, []);
 
-  const enabled = useFeatureFlag('enable-v11-release');
+  const enabled = FeatureFlags.enabled('enable-v11-release');
```
joshblack: Something that may be different with the hook versus the top-level import is that the hook taps into a "feature flag scope". 

In other words, there is a global set of feature flags that are either turned on/off. When in React and using hooks, we create a specific group of feature flags that could be different than this global set. This is called a "feature flag scope" in our package: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/FeatureFlags/index.js#L35-L39

This allows different parts of an app to render with different flags (for example, a page could use the v11 flag while all the others do not).

This whole thing gets a little complicated because our props are flagged, but in general we've been using the hook to try and allow for this use-case.

I hope that was helpful, wanted to share more about why the hook is used versus the top-level important and what is different between them 👀

packages/react/src/components/Button/Button.js
```diff
@@ -16,33 +16,34 @@ import { usePrefix } from '../../internal/usePrefix';
...
+    isSelected,
+    kind = 'primary',
     onBlur,
+    onClick,
     onFocus,
     onMouseEnter,
     onMouseLeave,
-    ...other
+    renderIcon: ButtonImageElement,
+    size = 'lg',
```
tay1orjones: Should this be `md`? This might be why VRT is failing in Percy


https://user-images.githubusercontent.com/3360588/151428614-95424108-0fc1-4d26-af0f-319c0e30f771.mov

sstrubberg: I went with the design guidance on the [v11 site](https://v11.carbondesignsystem.com/components/button/usage/#button-sizes). Did i misinterpret that?

image

tay1orjones: Ah, I think in v10 the default needs to be `md`, for v11 the default can be `lg`. I believe the value for v10 `md` is the same as v11's `lg`. As the PR stands now, the storybook for v10 looks wrong, but the v11 storybook looks correct. Percy only runs against v10 code right now.

jnm2377:
Is the v11 icon button story supposed to be broken? It's missing the icon.

packages/react/src/components/Button/Button.js
sstrubberg: That did it! Thx!

sstrubberg:
> Is the v11 icon button story supposed to be broken? It's missing the icon.

Yeah, i noticed that before I even started working on this. Up to ya'll if we want to get this in and then deal with the icon button or get it bundled in this PR.

sstrubberg:
> > Is the v11 icon button story supposed to be broken? It's missing the icon.
> 
> Yeah, i noticed that before I even started working on this. Up to ya'll if we want to get this in and then deal with the icon button or get it bundled in this PR.

Cancel that. I think it's from when i merged something from main

jnm2377 (approved):

tay1orjones (approved):

PR #10537, opened 2022-01-26
fix(data-table): add position relative to InlineCheckbox to prevent scroll

### LINKED ISSUE #6521: DataTable with selection adds invisible floating checkboxes that increase the height of the page

- [x] `carbon-components`

## Detailed description

DataTable with selection adds invisible floating checkboxes that increase the height of the page (for certain layouts). I can fix the problem by forcing bx-checkbox to `display: none` or removing the `position: absolute`

DataTable

That the table only takes up it's own space no matter the layout. 

> see changed?

bx-checkbox doesn't rely on `overflow: hidden` (which seems to be the culprit in my layout), `display: none` is added to the bx-checkbox in a table row

Chrome

latest

> should be aware of?

1. Add DataTable into fixed sized scrollable element (mine was in a grid > tabs> grid) where tab-content was made scrollable.

>
> - Style and vanilla JS:
> - React:

## Additional information

- Screenshots or code
- Notes

Contact me on Slack and I can show a demo of the issue. it's hard for me to reproduce in codesandbox

### LINKED ISSUE #10307: [Bug]: DataTable with stickyHeader inside a ComposedModal scrolls back to the top when selecting rows at the bottom
### Package

carbon-components-react

### Browser

Chrome, Firefox

### Package version

7.50.0

### React version

16.13

### Description

I am using a DataTable inside a ComposedModal, where my DataTable uses TableSelectRow and stickyHeader.
Whenever I try to select a checkbox that is located at the bottom of the table, it scrolls back to the top. 

### CodeSandbox example

https://codesandbox.io/s/frosty-hertz-r4rgn

### Steps to reproduce

1 - Scroll down to see rows that are hidden by the scroll.
2 - Select a row.

Table will jump back to the top. 

### Code of Conduct

joshblack:
Closes #6521
Closes #10307

This issue was reported over on Slack: https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1643032172186100

It details a problem that occurs in the `DataTable` where clicking on a checkbox will cause the page to scroll to the top. This problem occurs because of a change made to `InlineCheckbox` where we no longer have the checkbox within a parent that is `position: relative`.

As a result, the `position: absolute` that comes from our `hidden` mixin (and applied to `bx--checkbox`) is being used in conjunction with `top` in `bx--checkbox` to place the native checkbox at the top of the page. This results in the page jumping because the browser is attempting to put this control in the visible part of the page.

This PR attempts to address this by adding back in a `position: relative` container so that the native checkbox is placed inline instead of the top of the nearest `position: relative` parent.

#### Changelog

**New**

**Changed**

- Add `position: relative` helper for inline checkbox

**Removed**

#### Testing / Reviewing

- View the Data Table story for selection that includes checkboxes
- Decrease the height of the page until you have a vertical scrollbar
- Click on a checkbox
- Verify the page does not jump to the top

### DIFF
--- a/packages/components/src/components/checkbox/_checkbox.scss
+++ b/packages/components/src/components/checkbox/_checkbox.scss
@@ -42,6 +42,7 @@
   // to match the specs.
   .#{$prefix}--checkbox {
     @include hidden;
+
     //even though its hidden, positioning is for native validation to be aligned to checkbox
     top: 1.25rem;
     left: 0.7rem;
@@ -197,6 +198,13 @@
 
     // Add extra spacing when label is present
     margin: rem(1px) 0 0 rem(6px);
+  }
+
+  //-----------------------------------------------
+  // InlineCheckbox
+  //-----------------------------------------------
+  .#{$prefix}--checkbox--inline {
+    position: relative;
   }
 }
 
--- a/packages/react/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
+++ b/packages/react/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
@@ -975,21 +975,25 @@ exports[`DataTable selection should have select-all default to un-checked if no 
                           name="select-all"
                           onClick={[Function]}
                         >
-                          <input
-                            checked={false}
-                            className="bx--checkbox"
-                            id="data-table-7__select-all"
-                            name="select-all"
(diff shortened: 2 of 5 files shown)


tay1orjones (approved):
I'm no longer seeing a jump 👍

tw15egan (approved):
No jumping for me 👍🏻

PR #10539, opened 2022-01-27
fix(Popover): adjust popover content placement

### LINKED ISSUE #10286: [popover]: caret should be 16px away from edge of popover
### Package

carbon-components

### Browser

Chrome, Safari, Firefox

### Package version

v10.49.0

### Description

When the popover component has a caret tip, the caret tip should always be placed `16px` away from the edge of the container. The caret should be aligned to the type that is also following the 16px padding and the caret tip should still be vertically center with the trigger button.

### Screenshots

Example alignment:
Artboard 1

### Steps to reproduce

View the experimental popover story in react:
https://deploy-preview-9965--carbon-react-next.netlify.app/?path=/story/experimental-unstable-popover--playground

### Code of Conduct

dakahn:
Closes #10286 

Minor adjustment to the placement of the popover content which should line up the caret (pointy thing) with the end of the text padding 
145869848-c89e4798-d3df-4036-9701-8021dbfa6f85

#### Testing / Reviewing

make sure whats compiled and on netlify matches the above screen shot 👍🏾


tay1orjones (approved):

joshblack:
@laurenmrice @dakahn this might be coming up due to some of the changes we made to the placement logic and I wanted to see what you all felt like was the right set of trade-offs.  

When looking at the popover, we have two pieces: the caret and the overall box

image

When we are placing these two pieces, everything is relative to the trigger. As a result, we can only place these items in the center of the trigger, or left/right-aligned

image

Screen Shot 2022-01-27 at 11 00 31

Screen Shot 2022-01-27 at 11 00 40

Unfortunately, we can't really control how much distance there is between the edge of the content and the position of the caret.

If we want to control the distance between the edge of the content and the position of the caret, then unfortunately we lose the ability to center the caret relative to the trigger 😞  This is a trade-off and I'm curious what you all think / which one we want to pick.

I'll also share this with the team to see if they have any other thoughts on how to tackle this!

joshblack:
Notes from our meeting today:

image

It seems like we need to:

- Revert the changes to the default placement so that the edge of the popover aligns to the edge of the trigger
- Add in a caret only translation that follows the formula:
  - Center popover
  - Translate half the height (or width) of the caret
  - Add the popover padding value

abbeyhrt:
@dakahn Is this ready for review? I wasn't sure based on the comments above 🙂

dakahn:
@abbeyhrt should be good for a review now 👍🏾

laurenmrice (changes requested):
The alignments `left-bottom`, `left-top`, `right-bottom`, and `right-top` are having some misalignment issues when the caret prop is set to false. 
popover

-------------- 

It should look like this spec:
Artboard

--------------

All the other alignments look great!

dakahn:
@laurenmrice ACK! Great catch. Totally missed that. On it 👍🏾

laurenmrice (approved):
Looks good to me now D.A.! Thank you! ✅

tay1orjones (approved):

abbeyhrt (approved):

joshblack (approved):

PR #10547, opened 2022-01-27
docs(react): updated storybook to use new prop to fix headers a11y issue

### LINKED ISSUE #10167: [a11y]: The 'headers' attribute value "expand" does not reference a valid 'id' in this document
### Package

carbon-components-react

### Browser

Chrome

### Operating System

MacOS

### Package version

v7.48.1

### Assistive technology

_No response_

### Description

The 'headers' attribute should refer to a valid cell in the same table. The 'headers' attribute specifies one or more header cells related to a table cell in a complex table. It has no visual effect on the table, but it provides a programmatic way for assistive technology such as screen readers to describe visual relationships between header and data cells. When a table has multi-level headers, it can become difficult to convey which header text applies to any one cell using only elements and 'scope' attributes. The 'headers' attribute allows for the concatenation of multiple text into an accurate descriptor for an individual data cell.

### WCAG 2.1 Violation

_No response_

### CodeSandbox example

https://cloud.ibm.com/unifiedsupport/supportcenter

### Steps to reproduce

1. Open https://cloud.ibm.com/unifiedsupport/supportcenter
2. Login with IBM ID & password
3. Click View all
4. Check accessibility

image

### Code of Conduct

TannerS:
Closes #10167

https://github.com/carbon-design-system/carbon/issues/10167#issuecomment-1023460882

I recently did this a11y fix here https://github.com/carbon-design-system/carbon/pull/10214 but forgot to add the changes into the storybook (or maybe I added them to the wrong part of the storybook) so this is the storybook fix for this a11y issue 

#### Changelog

**Changed**

Added the id prop to the `TableExpandHeader` component for the expanding data tables to match up with the `TableExpandRow` defaults `expandHeader` props default value 

#### Testing / Reviewing

Ran the IBM a11y tool on the storybook locally for each expanding table and did not see the issue. I also checked the html to be sure the id's got added and matched the headers attribute


dakahn (approved):

aledavila (approved):
lgtm

PR #10552, opened 2022-01-28
feat(css-grid): add percent span prop option

### LINKED ISSUE #10475: Add %-span support (100%, 50%, 25%) to Grid

tay1orjones:
Closes #10475 

Enables CSSGrid columns to span a percentage width of the parent grid

#### Changelog

**New**

- grid: `spanPropTypes` support new string percent options
- grid: responsive per-breakpoint percent span width styles

#### Testing / Reviewing

- View the Grid/Responsive story in the carbon-react-next storybook, there's a new column showing percent span widths.


packages/react/src/components/Grid/Column.js
```diff
@@ -51,6 +51,7 @@ const spanPropType = FeatureFlags.enabled('enable-css-grid')
         start: PropTypes.number,
         end: PropTypes.number,
       }),
+      PropTypes.arrayOf(PropTypes.oneOf(['25%', '50%', '75%', '100%'])),
```
joshblack: Would something like this work:

```js
const percentSpanType = PropTypes.oneOf(['25%', '50%', '75%', '100%']);

//
? PropTypes.oneOfType([
    PropTypes.bool,
    PropTypes.number,
    PropTypes.shape({
      span: PropTypes.oneOfType([PropTypes.number, percentSpanType]),
      offset: PropTypes.number,
      start: PropTypes.number,
      end: PropTypes.number,
    }),
    percentSpanType
  ]);
```

Only reason why I'm curious is if we could support offsets or not with percent span 🤔

abbeyhrt (approved):
Looks good to me! Grid is such a cool component, playing around with the story and the percentages work as expected!

packages/react/src/components/Grid/Column.js
tay1orjones: @joshblack Great idea - I got it working! I added a new column to the `offset` story that uses percent-based span with offsets. Check it out and let me know what you think.

aledavila (approved):

joshblack (approved):

PR #10559, opened 2022-01-31
chore(button): skeleton size prop audit

sstrubberg:

- Added FeatureFlag ternary for ButtonSkeleton size prop based on guidance provided by design. [REF](https://v11.carbondesignsystem.com/components/button/usage/#button-sizes)
- Removed the `defaultProps` object and moved them into default params. [REF](https://github.com/reactjs/rfcs/pull/107)

#### Testing

- Make sure Button is behaving correctly in the v11 & v10 storybooks and that size props are accurately reflected according to the design guidance.
- Ensure I haven't ruined someone's day from the tests I updated.


abbeyhrt (approved):
Looks good to me!

aledavila (approved):

PR #10589, opened 2022-01-31
chore(TextInput): size audit

sstrubberg:

- Added FeatureFlag ternary for TextInput size prop based on guidance provided by design. [REF](https://v11.carbondesignsystem.com/components/text-input/style#sizes)
- Removed the `defaultProps` object moved them to default params. [REF](https://github.com/reactjs/rfcs/pull/107)

#### Testing

- Make sure TextInput is behaving correctly in the v11 & v10 storybooks and that size props are accurately reflected according to the design guidance.


dakahn (approved):

tay1orjones (approved):
Looks the same 👍

PR #10612, opened 2022-02-01
chore(PasswordInput): updated size props

sstrubberg:

- Added FeatureFlag ternary for `PasswordInput` size prop based on guidance provided by design. [REF](https://v11.carbondesignsystem.com/components/text-input/style#sizes)
- Removed the `defaultProps` object and moved them to default params. [REF](https://github.com/reactjs/rfcs/pull/107)

#### Testing

- Make sure `Toggle Password Visibility` story is behaving correctly in the v11 & v10 storybooks and that size props are accurately reflected according to the design guidance.


packages/react/src/components/TextInput/PasswordInput.js
```diff
@@ -6,34 +6,35 @@ import { View16, ViewOff16 } from '@carbon/icons-react';
...
-    light,
+    invalid = false,
+    invalidText = '',
+    labelText,
+    light = false,
+    onChange = () => {},
+    onClick = () => {},
+    onTogglePasswordVisibility,
+    placeholder,
+    size = '',
```
jnm2377: should we have a default size?

packages/react/src/components/TextInput/PasswordInput.js
```diff
@@ -6,34 +6,35 @@ import { View16, ViewOff16 } from '@carbon/icons-react';
...
   {
-    labelText,
     className,
-    disabled,
-    id,
-    placeholder,
-    onChange,
-    onClick,
+    disabled = false,
+    helperText = '',
```
jnm2377: if there's no default helperText, should we just remove the `= ''`?

packages/react/src/components/TextInput/PasswordInput.js
```diff
@@ -6,34 +6,35 @@ import { View16, ViewOff16 } from '@carbon/icons-react';
...
     hideLabel,
+    hidePasswordLabel = 'Hide password',
+    id,
     inline,
-    invalid,
-    invalidText,
-    helperText,
-    light,
+    invalid = false,
+    invalidText = '',
```
jnm2377: same as the helperText comment

packages/react/src/components/TextInput/PasswordInput.js
```diff
@@ -6,34 +6,35 @@ import { View16, ViewOff16 } from '@carbon/icons-react';
...
     invalidText,
-    helperText,
-    light,
+    labelText,
+    light = false,
+    onChange = () => {},
+    onClick = () => {},
+    onTogglePasswordVisibility,
+    placeholder,
+    size = 'md',
```
jnm2377, line 30: sorry, last thing. I thought `md` was just for v11? Would we need a ternary here for v10 sizes: small & large?

sstrubberg: derp, good call. Should I go empty string since there was no default in v10?

sstrubberg: I guess we can also wait to see if Percy finishes out. Md might already be default since before it was just asking for a string.

abbeyhrt: I don't think there are any styles for `md` in v10/v11 since there are no styles for `${prefix}--text-input--md` in `/styles` or `/components` so I guess in a way the default is md size wise, it's just never called out specifically?

sstrubberg: @abbeyhrt Totally. That's what I was assuming as well. The fact the Percy didn't yell at me for setting `md` is pretty telling, i think. @joshblack can we get some confirmation on these assumptions?

jnm2377 (approved):
Both of the stories seem to be working as expected, so approving :)

packages/react/src/components/TextInput/PasswordInput.js
abbeyhrt: Personally, I would go for an empty for the size prop since `md` doesn't have any meaning but it's also something we can change later on since it wouldn't break anything.

abbeyhrt (approved):
Stories look good!

PR #10636, opened 2022-02-03
docs(aspect-ratio): update aspect ratio docs for v11

joshblack:
This PR introduces two updates to our storybook for v11:

- Grid storybook styles are now scoped to a storybook-specific class name
- The Aspect Ratio story has been updated and includes controls

#### Changelog

**New**

- The AspectRatio story has been updated and now uses controls

**Changed**

- The Grid storybook styles are now scoped

**Removed**

#### Testing / Reviewing

- Verify no regression exists between the current grid storybook styles and the ones in this PR
- Verify the AspectRatio story renders as intended (correct aspect ratio and background color per item)
  - Also, verify the controls work when switching the "ratio" prop


jnm2377 (commented):
Looks great! Only change I think it needs is the aspect ratio story css only works for v10, we could just add an additional selector with `cds` or use `prefix`:
https://github.com/carbon-design-system/carbon/blob/3068b4224d82d65e19753b9b597771a0b27974d6/packages/react/src/components/AspectRatio/AspectRatio-story.scss#L8

tay1orjones (changes requested):
Yeah, we're missing the purple boxes

image

joshblack:
Thanks for the catch @jnm2377 @tay1orjones! Updated

jnm2377 (approved):

tay1orjones (approved):

PR #10637, opened 2022-02-03
chore(stylelint): update stylelint to 14.3.0

tw15egan:
Noticed there was a message about `stylelint` v13 being deprecated in VSCode, so I've gone and updated us to v14.3.0.

Screen Shot 2022-02-03 at 2 43 45 PM

**Reference**: 

- https://github.com/stylelint/vscode-stylelint#%EF%B8%8F-stylelint-13x-and-prior-is-no-longer-supported

**Migration**: 

- https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-14.md

#### Changelog

**New**

- `stylelint` now extends `stylelint-config-standard-scss`

**Changed**

- Upgraded `stylelint` from `13.13` to `14.3`

#### Testing / Reviewing

Ensure no warning pops up in the bottom right corner of VS Code when opening the Carbon repo.
Ensure `stylelint` validations still appear as usual


tw15egan:
Note for reviewers: Do we want to disable the rules that are now throwing lint errors?

joshblack (approved):

joshblack:
@tw15egan great point, I think some of them may need to be configured per-directory, like:
`scss/no-global-function-names` should be disabled in `components` but enabled in `styles`

`scss/comment-no-empty` seems like one we can definitely disable

For ones that are unknown, definitely feel free to remove them (like `function-calc-no-invalid`) if there is no replacement in the new version

tw15egan:
@joshblack do you have any examples on how to disable in one package but not the others? Would that be made in `config/stylelint`, or in the `components` repo itself?

joshblack:
@tw15egan I _think_ we would use overrides for this: https://stylelint.io/user-guide/configure#overrides

Let me know if that works!

dakahn (approved):

tw15egan:
@joshblack Got it worked out, good to go now if you want to take another look

packages/styles/scss/components/button/_tokens.scss
```diff
@@ -330,11 +330,11 @@ $button-disabled: (
     ),
     (
       theme: themes.$g90,
-      value: rgba(141, 141, 141, 0.3),
+      value: rgb(141 141 141 / 30%),
```
joshblack, line 333: Is the `rgb` approach preferred moving forward?? Pretty cool format 👀

tw15egan: I agree, I had originally modified it to that format, but I was getting compilation errors when running the style tests. Any ideas on how to get around that?

config/stylelint-config-carbon/plugins/scss.js
```diff
@@ -116,4 +120,19 @@ module.exports = {
...
   },
+  overrides: [
+    {
+      files: [
+        './packages/components/**/*.scss',
+        './packages/type/**/*.scss',
+        './packages/motion/**/*.scss',
+        './packages/layout/**/*.scss',
+        './packages/grid/**/*.scss',
+        './packages/import-once/**/*.scss',
```
joshblack: Is there any way we could include the `scss/modules` directory inside of elements packages? These should be using the right sass functions, I think

tw15egan: i.e. remove `type`, `motion`, `layout` and `grid`? They threw a few errors, but I will see if I can just resolve them

tw15egan: Alright, I've updated the globs to only override the older files, but will still throw validation errors for the `module` folders. I will push up another commit once I've fixed these new lint errors.

Issue #10199, opened 2021-11-30
[Bug]: After high contrast with white is activated, the close icon is invisible


hnjoshi:
### Package

carbon-components

### Browser

Chrome

### Package version

Latest version

## Components / Docs Affected

 
- ToastNotification

## Detailed Description
> Describe in detail the issue you're having.

Problem: After high contrast with white is activated, the close icon is invisible

<img width="897" alt="image" src="https://media.github.ibm.com/user/42629/files/e936d900-21f6-11ec-9e9d-d0a3ade1bf8f">

> What did you expect to happen? What happened instead? What would you like to see changed?

Close icon should look properly as below:
<img width="428" alt="image" src="https://media.github.ibm.com/user/42629/files/08ce0180-21f7-11ec-9f82-a7a5b13f4138">

> What offering/team/product do you work on? Any pressing ship or release dates we should be aware of?
dns-svcs

### CodeSandbox example

.

### Steps to reproduce

.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

hnjoshi:
Linked to https://github.ibm.com/ibmcloud/pal/issues/5280

joshblack:
cc @dakahn when you get a sec, could you confirm that this is an issue in HCM?

dakahn:
@hnjoshi can you talk a little more about what "high contrast with white" means? 

three big things we need to start triaging contrast issues are browser version, operating system, and tool used for high contrast (plugin vs native). 

Thanks

hnjoshi:
![image](https://user-images.githubusercontent.com/8787607/148452319-434a449b-0609-4027-9da4-127d505935e4.png)
In Windows operating system if we choose "High contrast white" theme we can see the problem.
Chrome Version 96.0.4664.110 (Official Build) (x86_64)

dakahn:
Confirmed this bug on Windows 11 using the white high contrast theme
![image](https://user-images.githubusercontent.com/40970507/148464094-04c2d715-5c6a-407a-9a24-bf406f6326c8.png)

Bug not present with black high contrast theme
[...]

PR #10639, opened 2022-02-03
fix(components): add HCM support to notificiation close icon

### LINKED ISSUE #10585: Audit inline-notification styles
High level steps to take for each component are outlined in the [parent issue](https://github.com/carbon-design-system/carbon/issues/9805#issuecomment-1026166086).

If there's no actionable items to PR, just close this issue without a linked PR

### LINKED ISSUE #10586: Audit toast-notification styles
High level steps to take for each component are outlined in the [parent issue](https://github.com/carbon-design-system/carbon/issues/9805#issuecomment-1026166086).

If there's no actionable items to PR, just close this issue without a linked PR

abbeyhrt:
Closes #10586 
Closes #10585 
REF #10199 

Adds HCM for close icon in notification to `/components` as a part of the styles audit. Was previously added in `/styles`. 

#### Testing / Reviewing

If you have a Windows computer, using HCM, check that notifications have an accessible close X icon.

### DIFF
--- a/packages/components/src/components/notification/_inline-notification.scss
+++ b/packages/components/src/components/notification/_inline-notification.scss
@@ -314,6 +314,10 @@
   .#{$prefix}--inline-notification__icon {
     @include high-contrast-mode('icon-fill');
   }
+
+  .#{$prefix}--inline-notification__close-icon {
+    @include high-contrast-mode('icon-fill');
+  }
   /* stylelint-enable */
 }
 
--- a/packages/components/src/components/notification/_toast-notification.scss
+++ b/packages/components/src/components/notification/_toast-notification.scss
@@ -237,10 +237,16 @@
   .#{$prefix}--toast-notification {
     @include high-contrast-mode('outline');
   }
+
   .#{$prefix}--toast-notification__close-button:focus {
     @include high-contrast-mode('focus');
   }
+
   .#{$prefix}--toast-notification__icon {
+    @include high-contrast-mode('icon-fill');
+  }
+
+  .#{$prefix}--inline-notification__close-icon {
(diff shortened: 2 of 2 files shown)


tay1orjones (approved):
Having trouble getting a VM up and going, but this looks correct to me

jnm2377:
Hmmm... I'm reviewing the HCM in white and black since the original issue seemed like it was in white HCM. It looks like the original issue wasn't actually fixed in the white HCM. When I look at the v11 live storybook, the issue is still present. And same for the v10 deploy preview for this PR.

#### v11 storybook currently 
Screen Shot 2022-02-07 at 3 45 25 PM
Screen Shot 2022-02-07 at 3 44 58 PM

#### PR v10 deploy preview
Screen Shot 2022-02-07 at 3 45 32 PM
Screen Shot 2022-02-07 at 3 44 15 PM

That being said... I'm unsure how to proceed here with since the scope of this PR was to audit the styles and make sure they match, not necessarily fix the bug. Should this be a separate issue? What do you think? @abbeyhrt

abbeyhrt:
@jnm2377 Good question and great catch! Since these styles don't actually fix the problem, I'll re-open that issue and close this PR. I would rather actually resolve the issue and then just add whatever the solution is to both places.

tw15egan:
@abbeyhrt we may need to edit the `high-contrast-mode` mixin to target each of the HCM themes. I would have thought `ButtonText`, a system color, would properly change on the white HCM theme though...


```css
@media screen and (-ms-high-contrast: white-on-black)
```

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast

PR #10643, opened 2022-02-04
chore(Search): size audit

sstrubberg:

- Added default for `Search` size prop based on guidance provided by design. [REF](https://v11.carbondesignsystem.com/components/search/style)
- Removed the `defaultProps` object and used default params [REF](https://github.com/reactjs/rfcs/pull/107).
- Add size control for `Search` storybook stories.
- Removed deprecated `small` prop.
- Fixed styles to reflect updated design guidance.

#### Testing

- Make sure `Search` story is behaving correctly in the v11 & v10 storybooks and that size props are accurately reflected according to the design guidance.

#### Known Issue

- `ExpandableSearch` is currently not working and needs to be fixed.


packages/react/src/components/Search/next/Search.js
```diff
@@ -32,8 +32,7 @@ function Search({
   placeholder = '',
   renderIcon,
   role = 'searchbox',
-  size = !small ? 'lg' : 'sm',
-  small,
+  size = 'lg',
```
tay1orjones: Should this be `md`?

tay1orjones (approved):

joshblack (approved):

PR #10655, opened 2022-02-07
docs(react): change import to populate props table for notification

abbeyhrt:
Updates where the notification components are imported from so that the props table populates as expected. 

### Testing/Reviewing ### 

Go to notification story docs and check that the props shows notification's props for `Toast` and `Inline`


joshblack (approved):

tw15egan (approved):

tw15egan:
Nice and easy fix! Found a few more that are also solved by changing the import (https://github.com/carbon-design-system/carbon/pull/10656), nice find 😄

tay1orjones (approved):
Awesome! I'll keep an eye out for things like this

PR #10674, opened 2022-02-09
fix(OverflowMenu): use v10 menu and export v11 menu as separate unstable component

### LINKED ISSUE #10673: Decide on direction for OverflowMenuItem vs MenuItem
The original intent of the new MenuItem was that it would replace OverflowMenuItem. Since then, we've discovered MenuItem does not have feature parity to OverflowMenuItem.

- What is MenuItem missing?
- Can we patch it up by RC0?
- Do we still want to deprecate or remove OverflowMenuItem in favor of MenuItem in v11?
- If we deprecate in v11, can OverflowMenuItem live alongside MenuItem providing consumers a choice in the interim?

---

jnm2377:
Closes #10673 

#### Changelog

- Moved `OverflowMenu/next/OverflowMenu.js` refactor that Jan contributed into a new folder `OverflowMenuV2` and exported as `unstable_OverflowMenuV2`
- Updated `OverflowMenu/index.js` to have class wrapper since it's a class component
- Updated `OverflowMenuItem/index.js` to not use the functional `/next` component for now since OverflowMenu hasn't been refactored to a functional component. I didn't add the class wrapper bc it stops working with OverflowMenu (being keyboard navigable) if it's wrapped.
- Added some v11 styles that were breaking v11 breadcrumb overflow menu
- Added v11 `OverflowMenu` stories which match the v10 component/stories now.
- Updated the inline notification warning on the v11 unstable menu stories
- Updated `unstable_OverflowMenuV2` tests with basic RTL tests

#### Testing / Reviewing

- make sure tests are passing
- review v11 storybook deploy preview
    - look at `Components/OverflowMenu` stories to make sure they look and work as v10 does
    - look at `Experimental/unstable_Menu/OverflowMenuV2` to test the v11 (v2) version of OverflowMenu that was contributed by Jan. Everything should work as expected w/ just the component name changing. Can compare to existing v11 storybook example.


jnm2377:
@joshblack question about the component exports: since I'm updating the `react` package `index.js` to export Jan's menu as a separate unstable component, I didn't include that in the `carbon-react` package `index.js` exports bc it's exporting from `carbon-components-react` and I thought we would need to release first to have the new unstable export available. Is that a correct assumption?

tay1orjones (commented):
Is it possible to avoid the "V2" name?

Curious if we could keep the Menu-based OverflowMenu as `unstable_OverflowMenu` and refer to it as such in both the new and old storybook under `Menu`? The `OverflowMenu` stories for the existing implementation in both old and new storybook would need to explicitly import the "non-next" implementation, I think?


Just trying to think of a way to keep the same naming scheme we have today in v10:

- `OverflowMenu` refers to the existing implementation
- `unstable_OverflowMenu` refers to the new Menu-based implementation

jnm2377:
> Is it possible to avoid the "V2" name?
> 
> Curious if we could keep the Menu-based OverflowMenu as `unstable_OverflowMenu` and refer to it as such in both the new and old storybook under `Menu`? The `OverflowMenu` stories for the existing implementation in both old and new storybook would need to explicitly import the "non-next" implementation, I think?
> 
> Just trying to think of a way to keep the same naming scheme we have today in v10:
> 
> * `OverflowMenu` refers to the existing implementation
> * `unstable_OverflowMenu` refers to the new Menu-based implementation

yes @tay1orjones I'm not married to the idea of this name. I wasn't sure what to call it, honestly. I went with v2 because I know in the past we've done that with DataTable. Once DataTableV2 replaced DataTable (OG), we changed the name to just DataTable. We can definitely raise it up to the team and see what people think is best.

tay1orjones (approved):

aledavila (approved):

Issue #10401, opened 2022-01-13
[Bug]: Getting Accessibility error(AVT) for Radio button group without legendText.


priyanka1304:
### Package

carbon-components, carbon-components-react

### Browser

Chrome

### Package version

v7.50.0

### React version

16.13.1

### Description

1. Getting Accesibility error when radio button group is created with out legendText.
<img width="369" alt="Screenshot 2022-01-13 at 5 11 07 PM" src="https://user-images.githubusercontent.com/35679136/149324282-249f3c14-898e-4b1a-9a03-79d77c619737.png">
<img width="787" alt="Screenshot 2022-01-05 at 6 05 01 PM" src="https://user-images.githubusercontent.com/35679136/149324327-60111291-3a8f-4570-9900-d1083edfcf4b.png">

### CodeSandbox example

https://codesandbox.io/s/sweet-moore-pydjy?file=/src/index.js

### Steps to reproduce

To reproduce this error accessibility checker tool is required.
Screen shots are attached. Those can be refered.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

jnm2377:
Hi @priyanka1304 thanks for opening this. I can confirm I'm able to reproduce this AVT error even in our latest version. I think we have a few options: 
1. make the prop `legendText` required 
2. add a default value for `legendText` so that there's always a legend

@dakahn do you have any thoughts on this?

priyanka1304:
> Hi @priyanka1304 thanks for opening this. I can confirm I'm able to reproduce this AVT error even in our latest version. I think we have a few options:
> 
> 1. make the prop `legendText` required
> 2. add a default value for `legendText` so that there's always a legend
> 
> @dakahn do you have any thoughts on this?

Thanks for reply. If I will give any value to legendText , then a "Radio button heading" will be shown in the UI. But we want to use the radio button group with out radio button heading. can you confirm it's possible or not?

tw15egan:
We have a `hideLabel` prop for `RadioButton` that visually hides the label, but keeps it in the DOM for a11y purposes. We could probably add a similar `hideLegend` prop to `RadioButtonGroup` to resolve this issue.

[...]

PR #10682, opened 2022-02-10
feat(react): added hideLegend prop to RadioButtonGroup

TannerS:
Closes #

- I wanted to take a jab at the proposed solution in this issue: https://github.com/carbon-design-system/carbon/issues/10401. It is not agreed that this solution is the way to go, but I wanted to throw something together to get an idea that it can work. From there, it can be possible to decide if its a solution or not

#### Changelog

**New**

- Added `hideLegend` prop to hide the `RadioButtonGroup` legend to fix an a11y issue.

#### Testing / Reviewing

- Tested in local storybook using IBM a11y tool


tw15egan (approved):
Nice work! Able to hide the legend and not get any a11y violations 👍🏻 ✅ 🎉

tay1orjones (approved):

Issue #10670, opened 2022-02-08
Updating the progress bar in code 


thyhmdo:
## Updating the progress bar in code 
Related issue: https://github.com/carbon-design-system/carbon-website/pull/2688

**Tasks** 
- [x] Adding statuses (error, success, loading)
- [x] Adding alignment variants 
- [x] Adding sizes (big and small)
- [x] RTL support

jnm2377:
@thyhmdo is there a design spec for all the things that need to be updated?

janhassel:
@thyhmdo I guess this issue can be closed now that the four PRs are merged

thyhmdo:
@janhassel yes

PR #10708, opened 2022-02-11
feat(progress-bar): add "small" size variant

janhassel:
Ref #10670

Adds a new prop (`size`) to the `ProgressBar` component and adds the styling necessary to render at `small` size (4px height).

#### Changelog

**New**

- Add `props.size` to `ProgressBar`
- Add respective styles

#### Testing / Reviewing

Verify in storybook that "big" variant is 8px high and "small" 4px.

---

Question regarding the prop values: As I understand components that don't fit the classic t-shirt size names like Tag, Toggle and ProgressBar should be referred to as "big" and "small". So far, the two other components still use `sm` and `md` as their prop values. With v11, will this be changed or should I change the prop values for ProgressBar also to `sm` and `md` for now?


tay1orjones:
@aagonzales what do you think about `big` and `small` vs `sm` and `md` here?

aagonzales:
It says the sizes but there no button to actually change the size. How do you turn the small size on? (i noticed this happening in a lot of stories) 
image

tay1orjones:
@aagonzales Ah yeah, sorry about that. In the new storybook we need to add in controls. Storybook deprecated the knobs addon we use in the v10 storybook, so many stories in the v11 storybook don't have controls. It's going to be addressed this sprint in https://github.com/carbon-design-system/carbon/issues/10751

The v10 storybook has a knob though to view these changes https://deploy-preview-10708--carbon-components-react.netlify.app/?path=/story/experimental-unstable-progressbar--playground

tay1orjones (approved):

aagonzales (approved):

abbeyhrt (approved):

Issue #9623, opened 2021-09-08
Decide on convention of `type`, `kind`, and `variant` for v11


joshblack:
**Background**

We would like to be consistent in our terminology with how a user can configure a component to render differently in terms of appearance. In our codebase currently, we use several style of props to indicate that a component can have different visual appearances, including:

- [`type`](https://carbon-react-explorer.vercel.app/props/type)
- [`kind`](https://carbon-react-explorer.vercel.app/props/kind)

In this proposal, we would like to define what the exact term used must be for components and when to use these other prop names, if at all.

**Links & Resources**

- Design Tools
  - [Framer](https://www.framer.com/support/using-framer/variants/)
    - A Variant is any visual state specified for a component. With Variants, components can be given unique states to which they can animate when an event is happening or has occurred.
- Design Systems
  - Components using `variant`
    - https://spectrum.adobe.com/page/button/
    - https://primer.style/react/Buttons

joshblack:
- Button
  - Uses `kind` because `type` conflicts with `<button type="...">`

joshblack:
**Figma Research**

Figma uses a term called a **Component Set** to describe a collection of components, called variants, that all share their properties. However, the variants will have different combinations of values for these properties.

A **Component Set** could be one-dimensional. For example, a `button` could have a `state` property with the values: `standard`, `hover`, `pressed`, or `disabled`.

Our **Button Component Set**, in this case, would contain the following variants:

![image](https://user-images.githubusercontent.com/3901764/132733643-23c5dd8a-9ff6-4bbc-be5f-cdb5634363b7.png)

In this case, the variants are specifically:

- Button with state: standard
- Button with state: hover
- Button with state: pressed
- Button with state: disabled

A **Component Set** could also be multi-dimensional. For example, a `button` could have several `type`s such as `primary` and `secondary`. In this case, our **Button Component Set** would look like:

![image](https://user-images.githubusercontent.com/3901764/132733964-12307bdf-3d63-4cdd-8efd-b7670d850648.png)

In this case, the variants are:

- Button, type: primary, state: standard
- Button, type: primary, state: hover
- Button, type: primary, state: pressed
- Button, type: primary, state: disabled
- Button, type: secondary, state: standard
- Button, type: secondary, state: hover
- Button, type: secondary, state: pressed
- Button, type: secondary, state: disabled

**Definitions**

**Design Systems**

**Links & Resources**

- https://www.youtube.com/watch?v=y29Xwt9dET0
[...]

PR #10769, opened 2022-02-16
feat(progress-bar): add alignment variants

janhassel:
Ref #10670

Adds a new prop (`type`) to the `ProgressBar` component and adds styling necessary to render the different [alignment variants](https://carbondesignsystem-hz7eutozq-carbon-design-system.vercel.app/components/progress-bar/usage#alignment).

#### Changelog

**New**

- Add `props.type` to `ProgressBar`
  - Options: `default`, `inline` and `indented`
- Add respective styles

#### Testing / Reviewing

Verify visuals with storybook.

---

Just want to make sure that `type` is the preferred prop name right now / for v11. The discussion in #9623 doesn't seem to have been fully concluded so I compared the number of components using `type` vs. `kind` with https://carbon-react-explorer.vercel.app/props and decided on `type` based on that.
Let me know if there has been a decision I on this topic I'm not aware of.

@thyhmdo Can you please confirm that the "inline" alignment variant should never have a visible helper text? I couldn't find any explicit guidance around this but also didn't see any mockups with it applied so I assumed a helper text is not supported in that case.


abbeyhrt:
@janhassel thank you for doing this! I would need a designer to confirm but comparing `indented` to the alignment varients pictured here



it seems like the label is indented a bit too much in the storybook example.  

Screen Shot 2022-03-01 at 3 13 18 PM. 

It almost looks like the padding-left for the label is 8px and padding left for the optional text is 16, but there is no example of the two used together for the indented variant so i'm not sure .

thyhmdo:
@janhassel Sorry coming to this late. 

- Not sure where we landed with the variants either. Since we called the indeterminate and determinate progress bars as variants, I'm not sure if variants or variations would make sense here. But I think we can always fix that later. 
- Yes, the inline variant won't have the optional helper text. 
- Here  are the specs and I will update this on the Style tab. The thing is it looks like either the Optional helper text looks quite close to the progress bar or the label is far away from the progress bar. It should apply 8px for spacing between the progress bar and the text in both Big and Small progress bar cases. 

@abbeyhrt I don't see the indented version on react. How can I view that?

image


progress-bar-style-4

thyhmdo:
thanks Jan!

janhassel:
Let me know if you have any questions around this PR @abbeyhrt @aledavila @thyhmdo

thyhmdo:
thanks @janhassel for updating. Just a question. The https://github.com/carbon-design-system/carbon/pull/10843 looks good to me. The indent alignment does not appear on that doc but on this issue instead. I may have to wait for something to be deployed. Other than that, it looks good!

abbeyhrt (approved):
Looks good to me code-wise! Thank you for the contribution!

aledavila (approved):
indentations look good to me. 

@thyhmdo you should be able to find them visually here. and change them under `knobs` and then under `type`. https://deploy-preview-10769--carbon-components-react.netlify.app/?path=/story/experimental-unstable-progressbar--playground

Let me know if you can't see them and I can help out more :)

janhassel:
@thyhmdo Basically the two pull requests handle two different topics: The alignment variants in this one and the statuses in the other one. As long as both are open and have not yet been merged their features will not be available in the respective other one.

So as soon as this pull request is merged I can update the second one and then the alignment variants will be available as a preview in combination with the statuses.

Let me know if I misunderstood your question!

thyhmdo (approved):
looks good to me!

PR #10780, opened 2022-02-17
test(stylelint): add a11y rules to stylelint

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/10779

POC for adding in an a11y rule to stylelint that ensures that there is a media query to remove animation if a user has `prefers-reduced-motion` enabled

#### Changelog

**New**

- `stylelint-a11y` added to our stylelint config
- `a11y/media-prefers-reduced-motion` set to `true` to offer a better experience to users who enable the `reduce motion` option at the OS level. 

**Changed**

- Added the appropriate media query where needed to fix this issue with existing components

#### Testing / Reviewing

To enable:
Mac OS - `system preferences --> accessibility --> display --> reduce motion` 

Go to a component that has had this media query added. If you open up dev tools, and go to the animation tab, you'll see a little pane for the current animations (when triggered). When `reduce motion` is checked, an animation / transition should no longer render


jnm2377 (approved):
Tested with reduced motion and it's working 👍🏽

joshblack (approved):

PR #10799, opened 2022-02-18
feat(react): add horizontal overflow to Tabs in v11

### LINKED ISSUE #10249: [Tabs]: reimplement horizontal scrolling in `/next` version
Continuing the work from https://github.com/carbon-design-system/carbon/pull/10225, the new Tabs API implemented in `/next` needs to have the horizontal scrolling functionality implemented to match the standard Tabs/Tab in production v10

abbeyhrt:
Closes #10249 

This was a doozy. 
 - Adds horizontal overflow to Tabs. 
 - Adds a hook `usePressable`.  By default, the events are called in this order: onMouseDown -> onMouseUp -> onClick. Because of this, usePressable  coordinates when `onClick`, `onMouseDown`, and `onMouseUp` are called so that onMouseDown isn't called when a user just clicks and that, if a user has been pressing and holding onMouseDown, `onClick` isn't called after it's released. 
- handles window resize to add horizontal overflow buttons when necessary. 
- adjust styles for overflow buttons

#### Testing / Reviewing

Go to the Tabs stories in `next` storybook and try out the horizontal overflow, holding press and just clicking. The first story has been changed to be in a small container so you can mess around with it (I'll remove before merging).  
With the other stories, test out that the overflow updates appropriately on resize.


packages/react/src/components/Tabs/next/Tabs.js
```diff
@@ -241,19 +459,39 @@ TabList.propTypes = {
...
+
   /**
    * If using `IconTab`, specify the size of the icon being used.
    */
   iconSize: PropTypes.oneOf(['default', 'lg']),
+
+  /**
+   * Provide the props that describe the left overflow button
+   */
+  leftOverflowButtonProps: PropTypes.object,
```
joshblack, line 397: Random question, are these props that we're porting over from the existing implementation? 👀

packages/react/src/components/Tabs/next/Tabs.js
```diff
@@ -194,26 +314,124 @@ function TabList({
     }
   });
 
+  useLayoutEffect(() => {
```
joshblack: I think you'll need to use our `useIsomorphicEffect` utility for this to be safe for SSR

packages/react/src/components/Tabs/next/Tabs.js
```diff
@@ -194,26 +314,124 @@ function TabList({
...
-            {React.cloneElement(child, {
-              ref,
-            })}
-          </TabContext.Provider>
-        );
-      })}
+    <div className={className}>
+      <button
+        aria-hidden="true"
+        aria-label="Scroll left"
```
joshblack, line 319: Is the idea that `aria-label` would be configurable through `leftOverflowButtonProps` or is there another way we want people to customize this label?

abbeyhrt: I suppose it could be, `leftOverflowButtonProps` would override the label we have defined here. This was just a carryover from the current implementation though, i'm down to make it more configurable.

packages/react/src/components/Tabs/next/Tabs.js
abbeyhrt: Yep! Exactly

joshblack (approved):

tay1orjones (commented):
Could we call `scrollIntoView` on a tab when it's selected? Selecting "Tab 5" here leaves it out of view.


https://user-images.githubusercontent.com/3360588/155593448-71894515-8ac9-4077-a83c-f39b7fffe4a1.mov

Otherwise looks solid to me!

aledavila:
I think I broke it. When I press and hold to the end I can't go back to the left no matter what I click 

https://user-images.githubusercontent.com/12533409/156319675-3afeaca6-811c-4e6c-9d60-30d28775bddf.mov

abbeyhrt:
@aledavila Yeah it was :/ that should be fixed now if you want to check it out!

tay1orjones (approved):
Yes! Looks great 🚀

Issue #10828, opened 2022-02-23
[Bug]: Allow for setting `focusTrap` to false for <FloatingMenu />


annawen1:
### Package

carbon-components-react

### Browser

Chrome

### Package version

7.51.0

### React version

16.9.0

### Description

We are using both the `HeaderMenu` and `OverflowMenu` under the hood for our Masthead for Carbon for IBM.com. The focus trapping of the `OverflowMenu`  (our Profile Menu) is causing an issue when user interacts with the `HeaderMenus` (Masthead Menu Items). When user first opens the` OverflowMenu`, the focus trapping maintains focus on the `OverflowMenuItem`s. Then when expanding the `HeaderMenu`s, the focus goes to the `HeaderMenu` then back to the `OverflowMenuItem`. Because of the loss of focus after the `HeaderMenu` expanded, the `HeaderMenu`s remain open instead of closing on blur. 

![Feb-23-2022 09-46-37](https://user-images.githubusercontent.com/54281166/155342742-fd70af3a-4fb6-4baf-ac5a-cba836d26fc9.gif)

### CodeSandbox example

https://codesandbox.io/s/header-with-actions-and-nav-forked-9p6zd2?file=/src/index.js

### Steps to reproduce

Open codeSandbox link

1. click on the OverflowMenu (it's all the way on the right of the header, tried to add theming classes but didn't work - the icon is visible on hover 🙏 )
2. click on "Link 5" HeaderMenu - notice that focus is not going to "Link 5"
4. click outside of "Link 5" - notice the "Link 5" drop down does not close

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tw15egan:
@annawen1 since `focusTrap` is hardcoded to `true`, would a viable solution be to add the `focusTrap` prop to `OverflowMenu`, so that you could set `focusTrap={false}` ?

tay1orjones:
It looks like this is resolved via #10829, let us know if not and we can reopen

PR #10829, opened 2022-02-23
fix(OverflowMenu): allow users to set focusTrap to false

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/10828

Allows teams to disable `focusTrap`, but it will still be enabled by default. 

#### Changelog

**New**

- `focusTrap` can now be set via `OverflowMenu` directly

#### Testing / Reviewing

Ensure `focusTrap` still works by default and no a11y violations are thrown.

cc @annawen1


tw15egan:
FYI, setting `focusTrap` to `false` breaks keyboard navigation, open to suggestions on how to get the best of both worlds

tay1orjones (approved):
Would be nice to have something sorted for the lack of keyboard navigation, but until then we might consider making this a silent option. Available but not advertised - not included in storybook prop tables for instance

tw15egan:
@tay1orjones that's what I was thinking. It should allow @annawen1 and the team to remedy the issue on their end since this seems to be a one-off case. If more instances like this are occurring, we can revisit 👍🏻

dakahn (approved):

PR #10835, opened 2022-02-24
chore(IconSkeleton): move story to components

jnm2377:
No related issue, brought up in [slack thread](https://ibm-studios.slack.com/archives/GD8JG6JR3/p1645721046871839)

#### Changelog

- Moves `IconSkeleton` from the deprecated section to the components section.

#### Testing / Reviewing

Review both the v10 and v11 deploy previews:
- `Icon` should still be in the deprecated section, but `IconSkeleton` should no longer be included in those stories
- `IconSkeleton` should have it's own story now in the component section
- make sure styles are matching and tests pass

### DIFF
--- a/packages/react/src/components/Icon/Icon-story.js
+++ b/packages/react/src/components/Icon/Icon-story.js
@@ -9,7 +9,6 @@ import React from 'react';
 import { iconAdd, iconAddSolid, iconAddOutline } from 'carbon-icons';
 import { withKnobs, select, text } from '@storybook/addon-knobs';
 import Icon from '../Icon';
-import IconSkeleton from '../Icon/Icon.Skeleton';
 
 const icons = {
   'Add (iconAdd from `carbon-icons`)': 'iconAdd',
@@ -49,20 +48,6 @@ const props = () => {
   };
 };
 
-const propsSkeleton = {
-  style: {
-    margin: '50px',
-  },
-};
-
-const propsSkeleton2 = {
-  style: {
-    margin: '50px',
-    width: '24px',
-    height: '24px',
-  },
-};
-
 export default {
   title: 'Deprecated/Icon',
   component: Icon,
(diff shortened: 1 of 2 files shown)


joshblack (approved):

tw15egan:
Would it be possible to have it be under the `Skeleton` section, and be named `SkeletonIcon` to match the other `Skeleton` components? I don't want it to get lost being separated from its other skeleton siblings 😄

jnm2377:
> Would it be possible to have it be under the `Skeleton` section, and be named `SkeletonIcon` to match the other `Skeleton` components? I don't want it to get lost being separated from its other skeleton siblings 😄
> 
> (The component itself looks great!)

@tw15egan do you mean rename the component in v11 to `SkeletonIcon`? Or do you mean just changing the story name? I can def move the story to the skeleton section with the others! 👍🏽

tw15egan:
I think aligning them naming-wise would be nice, we could just keep `IconSkeleton` deprecated, and copy the code to a new file called `SkeletonIcon` so that it closely aligns with the other skeleton components. I guess this would be more of a `feat` since it would technically be introducing a new component, and I think it would improve the dev experience.

jnm2377:
> I think aligning them naming-wise would be nice, we could just keep `IconSkeleton` deprecated, and copy the code to a new file called `SkeletonIcon` so that it closely aligns with the other skeleton components. I guess this would be more of a `feat` since it would technically be introducing a new component, and I think it would improve the dev experience.

@tw15egan yeah, I can see how having the name switch would be more consistent. This PR was just intended as a quick doc fix so that people wouldn't be confused about what's being deprecated. I wasn't intending to introduce any changes. That said, we can ask the team what their thoughts are. Seems like the options are:

1. keep the component as is and just move the story
2. deprecate the component and introduce a new one `SkeletonIcon` so that it matches other skeleton components

PR #10838, opened 2022-02-24
feat(SkeletonIcon): add SkeletonIcon component

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/10835

Introduces a new `SkeletonIcon` component to align with the existing skeleton components, like `SkeletonPlaceholder` and `SkeletonText` 

#### Changelog

**New**

- `SkeletonIcon` introduced as a replacement for `IconSkeleton` 
- Introduced a new story, export, and tests for `SkeletonIcon` component

#### Testing / Reviewing

Ensure `SkeletonIcon` renders as expected, and all tests pass


abbeyhrt (changes requested):
Looks good to me! I think the only thing I see is that we also need to export it in [this file](https://github.com/carbon-design-system/carbon/blob/main/packages/carbon-react/src/index.js) as well so it's released in `@carbon/react`

tw15egan:
@abbeyhrt good catch! Updated 😄

abbeyhrt (approved):
🎉

joshblack:

tay1orjones (approved):
So sorry for the delay!

PR #10843, opened 2022-02-25
feat(progress-bar): add support for statuses

janhassel:
Ref #10670

Adds support for ProgressBar statuses.

#### Changelog

**New**

- Added prop `status` to `ProgressBar` component
  - Possible values: "active", "finished", "error"

**Changed**

- Added logic and styling necessary to render different statuses
- Included tests

#### Testing / Reviewing

- Run storybook
- Compare different statuses with the design guidance in https://github.com/carbon-design-system/carbon-website/pull/2688
- Status "finished" should set its value to `max`
- Status "error" should set its value to `0` and emit `aria-invalid`
- In the inline variant, the bar should be visually hidden (though accessible) with only the icon and label being shown
  - This is because the inline variant is supposed to be used when space is tight, e.g. in a data table cell and @thyhmdo and I agreed that priorizing the label and icon over the 100%-full bar would be the best option

#### Notes

- I opted for the term "finished" instead of "success" (as used in the design guidance) to be consistent with `InlineLoading`
- I had to use `eslint-disable` as it claims `aria-invalid` is not supported on `role="progressbar"`. Referencing [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#progressbar) though, `aria-invalid` should be supported. Please correct me if I'm wrong.


joshblack (approved):

joshblack:
Thanks for putting this together @janhassel! Sorry about the delay on my end, let me know if you need anything else from me here!

abbeyhrt (approved):

janhassel:
@joshblack No worries! I just resolved the last merge conflict that came up from the `usePrefix` change so I think this PR would be ready to go.

janhassel:
@joshblack I changed the base branch to v10 as discussed yesterday. Could you check if everything looks up to code now?

joshblack (approved):

PR #10852, opened 2022-02-25
feat(tile): create consistent API between tile components

### LINKED ISSUE #5972: React component API inconsistency between SelectableTile and RadioTile (misc properties)
- [ ] `carbon-components`

## Summary

Currently the APIs of the components SelectableTile and RadioTile are behaving inconsistent in terms of where to forward the spread misc properties (`...other`/`...rest`) to. This results in e.g. the style property being passed to the label element in SelectableTile and to the (invisible) input element in RadioTile. While https://github.com/carbon-design-system/carbon/pull/5946 suggested an extraction of the style property @asudoh and me had the opinion that it would be a better to resolve the inconsistency of the components as a whole. This can be done via forwarding the misc properties to the equivalent elements, but this would cause a breaking change.

This issues purpose is to discuss the solution for resolving the inconsistency.

## Relevant information

Currently a functional workaround is to use DOM refs to pass HTML attributes manually to the underlaying elements.

joshblack:
Closes #5972

This PR addresses #5972 by placing the extra props on tile components on the `<label>`. This should only occur when the `enable-v11-release` flag is flipped

#### Changelog

**New**

**Changed**

- Spread extra props on the `<label>` component in v11 for `RadioTile`

**Removed**

#### Testing / Reviewing

- Verify this change will not cause regressions in impacted components in v10


abbeyhrt (approved):

joshblack:

tay1orjones (approved):

PR #10900, opened 2022-03-02
fix(uishell): provide tabindex as a named prop

tay1orjones:
Surfaced in [slack](https://ibm-studios.slack.com/archives/C046Y0YUD/p1646256236144499):

> We have a situation where this component is usually hidden, and it’s an accessibility issue for tab navigation to traverse this element while it’s hidden. Therefore, we’d like to set its tabindex to -1 when it’s hidden.

~This just moves the tabindex to be set before `...rest` so it can be overridden~ This makes `tabIndex` a named and configurable prop.
#### Changelog

**Changed**

- provide `tabIndex` as a prop for `SwitcherItem` and `HeaderMenuItem`

#### Testing / Reviewing

- The component should still be navigable as it was before
- Adding tabindex=-1 should prevent it from being in the tab order


joshblack (commented):
For situations like this, would it be helpful if our convention was to make this an explicit/named prop?

This would allow us to safely override things that we need to control and allow a way to add in support for things over time. There are definitely things that should always be configurable (like `tabindex`) that we should straighten out but figured this could be a good heuristic going forward. Curious what you think

ColbyJohnIBM:
Is it possible to make a similar change here? We're running into the same issue trying to override the tabIndex for this component: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/UIShell/HeaderMenuItem.js#L37

tay1orjones:
@joshblack Great point, I made them named props.

@ColbyJohnIBM I also made the update in `HeaderMenuItem`

joshblack (approved):

aledavila (approved):

PR #10903, opened 2022-03-03
fix(Select): make changes based on design audit

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/10430

Makes style changes based on design audit

#### Changelog

**Changed**

- Changes color tokens based on design feedback
- Changed `carbon-mini-units` (was not calculating) to `spacing` token 

#### Testing / Reviewing

Ensure `Select` `InlineSelect` and `Pagination` are rendering with the correct tokens, especially in FIrefox.


joshblack (approved):

aledavila (approved):

packages/styles/scss/components/select/_select.scss
```diff
@@ -115,23 +115,23 @@
...
   .#{$prefix}--select-input:disabled ~ .#{$prefix}--select__arrow {
     fill: $icon-disabled;
   }
 
   .#{$prefix}--select--light .#{$prefix}--select-input {
-    background-color: $field-02;
+    background-color: $field;
 
     &:hover {
       background-color: $background-hover;
```
laurenmrice:
```suggestion
      background-color: $field-hover;
```

laurenmrice (changes requested):
Looks great, just two things:

>   optgroup.#{$prefix}--select-optgroup,
>   .#{$prefix}--select-option {
>     // For the options to show in IE11
>     background-color: $**background-hover**;
>     color: $text-primary;

- The background-color should be `$layer-hover` for the options in the menu.

laurenmrice (approved):
Looks good to me 👍

PR #10915, opened 2022-03-04
chore(FileUploader): updated size props

sstrubberg:
REF #10295 

Effects `FileUploaderItem`, `FileUploaderDropContainer`, `FileUploaderButton`, `FileUploader`.

- Modified the `size` prop to only accept SM, MD, LG for v11 by adding a FeatureFlag. 
- Added `md` as a default for the `size` prop.
- Added Controls functionality to Storybook story.
- Removed deprecated `field` styles.
- Removed `size` PropTypes from `FileUploaderDropContainer` because there is no size prop in that file.

#### Testing
- Make sure everything is snappy in the v11 Storybook. Size controls should allow for SM, MD, and LG


abbeyhrt (approved):
Looks good to me! 🎉

tay1orjones (commented):
@sstrubberg Any idea why the text size is larger in the v11 storybook?

image

sstrubberg:
> @sstrubberg Any idea why the text size is larger in the v11 storybook?
> 
> image

Super weird. Looks like this is [already an issue](https://carbon-react-next.netlify.app/?path=/story/components-fileuploader--file-uploader-drop-container).

tw15egan:
It looks like it's because the font styles in the first storybook file uploader example are wrapped in `cds--form-item`. The standalone one isn't. I believe all we need to do is add

```scss
.#{$prefix}--file__drop-container {
    @include type-style('body-compact-01');
    ....
}
```

to L326 of `file-uploader.scss`

tay1orjones (approved):

PR #10917, opened 2022-03-04
chore(project): install and use turborepo

joshblack:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


joshblack:
FYI @tay1orjones in case you were curious 👀  Was trying to figure out how hard it'd be to integrate turborepo here

.github/workflows/ci.yml
```diff
@@ -64,6 +64,11 @@ jobs:
             node_modules
             */**/node_modules
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+      - uses: actions/cache@v2.1.7
+        id: turbo-cache
+        with:
+          path: node_modules/.cache/turbo
+          key: ${{ runner.os }}-turbo-${{ hashFiles('turbo.json') }}
```
tay1orjones: If I understand correctly, this means 

* We'll get shared build artifacts in CI 
* We'll get shared build artifacts in local dev environments just by virtue of using turborepo 
* This will _not_ provide a shared cache that can be shared/used by _both_ CI and local. For that we'd need a [custom remote cache](https://turborepo.org/docs/features/remote-caching#custom-remote-caches) somewhere, right?

tay1orjones (commented):
Love it!

.github/workflows/ci.yml
tay1orjones: Related https://github.com/vercel/turborepo/discussions/381

PR #10918, opened 2022-03-04
fix(react): removes v11 enabled check on ${prefix}--label and adjusts styles

abbeyhrt:
Noticed that `FormLabel` was un-styled in v11. This PR removes the `enable-v11-release` flag check so that `${prefix}--label` isn't conditionally applied and updates `@carbon/styles` to make margin-bottom: 0 the default for `FormLabel` in v11. 

#### Changelog

**Changed**

- Minor code style updates

**Removed** 
 
- Unnecessary v11 flag checks in `_form.scss` 

#### Testing / Reviewing

Check that v10 FormLabel looks the same, with margin still applied. 

Verify that v11 FormLabel is now styled and has no margin-bottom.

### DIFF
--- a/packages/styles/scss/components/form/_form.scss
+++ b/packages/styles/scss/components/form/_form.scss
@@ -21,16 +21,6 @@ $input-label-weight: 400 !default;
 @mixin form {
   .#{$prefix}--fieldset {
     @include reset;
-
-    @if not enabled('enable-v11-release') {
-      margin-bottom: $spacing-07;
-    }
-  }
-
-  @if not enabled('enable-v11-release') {
-    .#{$prefix}--fieldset--no-margin {
-      margin-bottom: 0;
-    }
   }
 
   .#{$prefix}--form-item {
--- a/packages/styles/scss/components/form/_form.scss
+++ b/packages/styles/scss/components/form/_form.scss
@@ -50,17 +50,10 @@ $input-label-weight: 400 !default;
     @include type-style('label-01');
 
     display: inline-block;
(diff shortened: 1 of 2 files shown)


tw15egan (approved):
LGTM 👍 ✅

packages/styles/scss/components/form/_form.scss
```diff
@@ -50,19 +40,12 @@ $input-label-weight: 400 !default;
     @include type-style('label-01');
 
     display: inline-block;
-    margin-bottom: $spacing-03;
```
tay1orjones: I think this needs to stay in. Labels are incorrectly sitting above the inputs without any spacing between label and input:

image

packages/styles/scss/components/form/_form.scss
```diff
@@ -50,19 +40,12 @@ $input-label-weight: 400 !default;
...
     color: $text-secondary;
     font-weight: $input-label-weight;
     line-height: 1rem;
     vertical-align: baseline;
   }
 
-  @if enabled('enable-v11-release') {
-    .#{$prefix}--label--no-margin {
-      margin-bottom: 0;
-    }
```
tay1orjones, line 63: We might want to keep this around until we formally remove all usages of the --no-margin class in the components

tay1orjones (changes requested):
Yeah, those v11 checks are silly since /styles/ is effectively v11. I think a couple styles outlined below we'll want to leave in though.

tay1orjones (approved):

PR #10928, opened 2022-03-07
chore(OverflowMenu): size prop for v11

sstrubberg:
REF #10295 

- Modified the `size` prop to only accept SM, MD, LG for v11 by adding a FeatureFlag. 
- Added `md` as a default for the `size` prop.
- Added Controls functionality to Storybook story.
- Removed deprecated `xl` styles.

#### Testing
- Make sure everything is snappy in the v11 Storybook. Size controls should allow for SM, MD, and LG


joshblack (approved):
Looks great! Confirmed in old and new storybook ✅ 

One quick note for commit conventions, these types of changes could be considered a `feat` vs a `chore` since it's changing the Public API of this component. This totally gets tricky with feature flags but one way to think about it is since the commit is changing the API this would be a feat whereas a chore commit would not change the Public API

dakahn (approved):

PR #10934, opened 2022-03-07
feat(grid): add narrow grid support to CSS Grid

### LINKED ISSUE #10845: Provide experimental narrow grid mode support in CSS Grid
While implementing the narrow grid mode using CSS Grid, we ran into the following cases where narrow grid mode did not work as intended:

- Incorrect grid margin/hang in narrow mode at small screens (mobile breakpoint)
- Narrow mode would break when configured in a subgrid (gutter is incorrect)
- Narrow grid mode columns do not behave correctly and hang too far into the left gutter

With this in mind we removed narrow grid mode support in #10620. Since then, it appears it may be beneficial to still include the narrow grid mode but mark it as experimental due to the problems outlined above. 

**Checklist**

- [ ] Provide narrow grid mode styles in the grid package scss
  - Implementation-wise we have two options:
    1. Re-include the broken implementation that uses `grid-gap` and negative margin OR
    2. Roll a new implementation that does not use `grid-gap` instead adding gutters via padding. Additionally provide helper classes to add/drop the "hang". If we go this route, should we provide a prop on `Column` to mirror this helper class?
- [ ] Provide the narrow grid mode as a configurable prop within the `CSSGrid` component in `@carbon/react`
   - Open question, how do we flag this as experimental? Ideally `CSSGrid` would remain stable without the `unstable` prefix. Could we mark the prop itself as experimental? `unstableNarrow`?
- [ ] Re-include documentation for narrow grid mode in storybook, including how narrow mode is experimental and `FlexGrid` is a stable way to utilize narrow mode.

Most of this can be easily achieved by reverting/reinstating changes made in #10621.

joshblack:
Closes #10845 

This PR reintroduces the narrow grid support to the CSS Grid and includes support for deriving the grid from the `@carbon/grid` package configuration.

It also documents some of the architectural decisions of the package, including testing scenarios, and updates the migration guide for `@carbon/grid` to include any changes.

This PR also updates `@carbon/styles` so that teams can easily use the flexbox grid instead of the CSS Grid with a `$use-flexbox-grid` flag. By default, the grid entry point will emit CSS Grid styles.

#### Changelog

**New**

- Add usage docs for `@carbon/grid`
- Add support for hanging on the grid with `ColumnHang`

**Changed**

- Update v11 migration docs for `@carbon/grid`
- Update usage docs for `@carbon/styles`
- Update styles for CSS Grid Preview
- Update `@carbon/grid` to not emit styles by default
- Update mixin for CSS Grid to support building from config and supporting mixed subgrids

**Removed**

#### Testing / Reviewing

- Grid Preview
- Grid Storybook
  - v11
    - CSS Grid implementation works as expected
    - FlexboxGrid implementation works as expected
  - v10, flexbox implementation is unchanged
- Verify subgrid matches the intended alignment when mixing modes

image


joshblack:
@tay1orjones @abbeyhrt let me know if you want to go over this together at all, the PR size just kept growing 🤦

tay1orjones (approved):
I'm all in man, great work! 🥇 

This will help unblock some issues I'm seeing updating the website to v11. Notably the "not emitting styles by default" change and exposing the scss utility helpers. Thank you!

I think it's worth cutting a patch for this so we can begin to iterate on it with the website asap.

abbeyhrt (approved):
This is all so cool! It looks good to me as far as I can tell, thanks for going over it with me Josh!

PR #10935, opened 2022-03-07
fix(utilities): broaden HCM mixin to include macos query

### LINKED ISSUE #10844: [a11y]:  dropdown, overflow menu and ghost button text contrast issues with "increase contrast" setting on
### Package

carbon-components-react

### Browser

Chrome

### Operating System

MacOS

### Package version

10.42

### React version

7.43.0

Accessibility Checker, WCAG 1.4.3: Text Contrast

### Assistive technology

Increase contrast MacOS accessibility setting

### Description

The focus stage for **overflow menu** options, **ghost button** and inline **dropdown menu** options is against the 1.4.3 Contrast (Minimum) when the accessibility setting "increase contrast" on MacOS is on.
CleanShot 2022-02-23 at 15 59 06@2x
CleanShot 2022-02-23 at 15 50 55@2x
79393000-94c2-11ec-8715-13305a1f4277
 

### WCAG 2.1 Violation

1.4.3 Contrast

### CodeSandbox example

-

### Steps to reproduce

- Enable "increase contrast" on MacOS accessibility settings
- Using the Tab key to move until you see the focus state of the ghost button or option in an overflow menu or inline dropdown

### Code of Conduct

dakahn:
partially closes #10844

Extends our high contrast mixin to include a media query that covers macOS 🏄🏾 

#### Reference 
https://adrianroselli.com/2017/11/os-high-contrast-versus-inverted-colors.html#MacOSICMQs

#### Testing / Reviewing

make sure Dropdown is accessible with macos HCM cranked up 👍🏾


tay1orjones (commented):
@dakahn should this be mirrored over into v10 `packages/components`?

tay1orjones:
I'm not seeing much (any?) difference. Let me know if there's a different setting I should be testing maybe? I just followed the stuff from the original issue.


https://user-images.githubusercontent.com/3360588/157274398-2e057406-c20f-410c-8229-d96ec38dcc2f.mov

tw15egan:
I'm also not seeing anything... Seems like it is currently only supported in Safari, but even there it seems like the `Highlight` CSS color is causing an issue when the `increase contrast` setting is enabled...

dakahn:
Yep! Botched it lol. Good news is it's hitting the query -- bad news is it looks like our HCM styling is the problem. Since macOS HCM is just a css filter applied top the whole page (rather than a more holistic style sheet rewrite like in Windows) i'm going to see how it works to just exclude a truthy `inverted-colors: inverted` query 👍🏾

dakahn:
Looks like we can remove `prefers-contrast` and then browsers in macos wont pick up our HCM styling (which is what we want since they're filters and not stylesheet rewrites). `prefers-contrast` was how we fixed Firefox in Windows since they didn't want to play nice with the proprietary --ms query.

So it's been a few years since we looked at Firefox in Windows HCM and I thought there was a chance we didn't need that flag anymore. It seems like that might be the case 🎉 
image
image
image

I'll give this a more thorough once over in the morning, but anyone else that wants to also take a look on a Windows machine it'd be a big help

tw15egan:
@dakahn I just tested in Windows, and don't see any regressions by changing the selector. I also verified that the changes work with Safari and increased contrast enabled. Nice work 🎉 

Should we backport this to v10 as well?

dakahn:
@tw15egan Radio button was the only one i found, but i updated the component to use the HCM mixin 👍🏾

packages/styles/scss/components/radio-button/_radio-button.scss
```diff
@@ -109,16 +110,7 @@ $radio-border-width: 1px !default;
...
-        // Utilize a system color variable to accommodate any user HCM theme
-        background-color: WindowText;
-      }
-
-      // Firefox only HCM solution
-      @media screen and (prefers-contrast) {
-        // Utilize a system color variable to accommodate any user HCM theme
-        border: 2px solid WindowText;
-      }
+      @include high-contrast-mode('icon-fill');
```
tw15egan, line 113: Can we add this and the mixin updates to the V10 package as well?

dakahn: @tw15egan should i do this in the new fangled v10 branch instead now that @tay1orjones has made that?

tay1orjones: @dakahn Yeah good call, I vote for opening that in a separate PR into the `v10` branch

tay1orjones (approved):

abbeyhrt (approved):

packages/styles/scss/components/radio-button/_radio-button.scss
fbarroso24: @dakahn @tay1orjones @tw15egan has this been delivered to the v10 branch?  If so, can you include a link to that PR?

We're failing a particular WCAG guideline wrt Text Contrast as a result of this issue and our team does not plan to upgrade to v11 in the near term so having this fix in v10 would go a long way.  Thanks.

tay1orjones: @fbarroso24 It has not. Thanks for letting us know, we can get this into the v10 branch and cut a release.

tay1orjones: @fbarroso24 This has been ported to the v10 branch via https://github.com/carbon-design-system/carbon/pull/11330, and is available in [v10.57.1](https://github.com/carbon-design-system/carbon/releases/tag/v10.57.1)

PR #10937, opened 2022-03-08
fix(DataTable): fix disabled inputs when row is selected

### LINKED ISSUE #10913: [Bug]: Disabled radio button becomes invisible with hover+select a row in data table.
### Package

carbon-components, carbon-components-react, @carbon/icons-react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v7.50.0

### Description

In a selection Data table , if any disabled radio button  is present , then the following happens:
1. When hover or select a row,  the visibility of the  disabled radio button becomes very low.
2.  When hover+select the row, the disabled radio button becomes almost invisible.

This issue is already fixed in v11 of Cabon, but in our project, we cann't upgrade Carbon to v11 in this release. So Please deliver the changes in v10 also.

### Screenshots

Screenshot 2022-03-04 at 10 52 08 PM
Screenshot 2022-03-04 at 10 52 44 PM

### Steps to reproduce

1. Create a selection data table.
2. Add radio buttons under any column and mark one disabled.
3. Hover on the row, the disabled one will be very less visible.
4. Hover+select the row, the disabled one will become almost invisible.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/10913

Fixes issues with disabled items while hovering over selected rows

#### Changelog

**New**

- Specific v10 selectors to change the disabled colors when a selected row is hovered, and contains a disabled input element

**Changed**

- Updated tokens for the v11 data-table to ensure the disabled inputs use the disabled tokens that utilize opacity

#### Testing / Reviewing

I've added example stories to both the v10 and v11 storybook for testing, will remove before merging

-- Go to the With Selection story, select the row with disabled items, and then hover over the row. Ensure the icon and texts are still visible.

### DIFF
--- a/packages/components/src/components/data-table/_data-table-core.scss
+++ b/packages/components/src/components/data-table/_data-table-core.scss
@@ -407,6 +407,27 @@
     padding-top: $spacing-05;
   }
 
+  // Disabled radio button, checkbox fix #10913
+  tr.#{$prefix}--data-table--selected:hover
+    .#{$prefix}--radio-button[disabled]
+    + .#{$prefix}--radio-button__label,
+  tr.#{$prefix}--data-table--selected:hover
+    .#{$prefix}--checkbox[disabled]
+    + .#{$prefix}--checkbox-label {
+    color: $disabled-03;
+  }
+
+  // Disabled radio button, checkbox fix #10913
+  tr.#{$prefix}--data-table--selected:hover
+    .#{$prefix}--radio-button[disabled]
+    + .#{$prefix}--radio-button__label
+    .#{$prefix}--radio-button__appearance,
+  tr.#{$prefix}--data-table--selected:hover
+    .#{$prefix}--checkbox[disabled]
+    + .#{$prefix}--checkbox-label:before {
+    border-color: $disabled-03;
+  }
+
   //----------------------------------------------------------------------------
   // Radio
   //----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


tw15egan:
@aagonzales just pushed this up with some sample code (under the `selection` story) to test out a selected, hovered row with disabled items. I ported the stories to both the `carbon-next` storybook, as well as the `carbon-react` storybook to verify this works for v10 and v11. Let me know what you think

aagonzales (commented):
The styles look good! But I wonder if we can change the story example a little. The second column of checkmarks show the disabled feature but its not really a true use-case (that I can think of). That second checkbox row is also showing up in all the examples in the next storybook and not just the selection story. 

I think just having the last column be the radio selection would be fine. Or can we do just one row as disabled (not sure if it would be selectable then though 🤔 )?

tw15egan:
@aagonzales I'm just going to revert the story changes before merging to what they were before, I just wanted to make sure we could test it easily 😄

aagonzales (approved):
oh oops now I see the "will remove before merging" comment, my bad.

abbeyhrt (approved):
Looks good to me!

joshblack (approved):

tw15egan:
@joshblack @abbeyhrt @aagonzales I noticed some existing issues with the disabled link behaving the same way, so I've updated this PR to also take care of that scenario

dakahn (approved):

Issue #10422, opened 2022-01-13
Notification: v11 audit 


aagonzales:
## Audit
A designer and developer will pair on this audit, some tasks may be discipline focused. Follow the checklist below for the features and items to review. If there are no issues with the item then check the item as complete. If any problems or bugs come up when auditing add a comment to this issue with the problem and how to correct it (if you know how). Bugs do not need to be fixed while auditing. 

## Resources
- [Design spec in Sketch](https://ibm.box.com/s/v4hk50eol2wkurvtxz295kw6pz094lq5)
- [V11 Website](https://v11.carbondesignsystem.com/)
- [SCSS](https://github.com/carbon-design-system/carbon/tree/next/packages/components/src/components)
- [Storybook](https://carbon-react-next.netlify.app/?path=/story/getting-started-welcome--welcome)

## Variants
- Low contrast
- High contrast
- Inline
- Toast

## Checklist

**Visual** (in React code) 
_Design and dev check_
- [ ] Component is using the correct design tokens (color and type), see design spec and scss code.
- [ ] Component is rendering correctly across themes and layers, see design spec and storybook.
    - [ ] White theme
    - [ ] Gray 10 theme
    - [ ] Gray 90 theme
    - [ ] Gray 100 theme
- [ ] Component is rendering correctly across browsers (check themes across browsers as well), see storybook.
    - [ ] Firefox
    - [ ] Safari
    - [ ] Chrome
- [ ] Component sizes are rendering and named correctly ([if applicable](https://v11.carbondesignsystem.com/migrating/guide/develop#size-prop-updates-breaking))

 **Website** (v11)
 _Design checks_
- [ ] Style tab has correct design tokens listed 
- [ ] Style and usage tabs are using the correct [size props](https://v11.carbondesignsystem.com/migrating/guide/design#sizing) names (if applicable)
- [ ] Usage tab is up-to-date for with any [v11 behavioral](https://v11.carbondesignsystem.com/migrating/guide/develop#changelog) changes (if any)
- [ ] Live Demo has applied v11 changes
   - [ ] Themes are rendering correctly
   - [ ] Size props are named correctly
   - [ ] No `light` props included

**Design Kits** (Sketch only)
 _Design checks_
- [ ] Correct design tokens (type and color) are used
- [ ] Component using the correct size prop names
- [ ] Any additional v11 behaviors have been added

**Storybook**
 _Dev checks_
- [ ] Confirm that prop table is populating 
[...]

PR #10939, opened 2022-03-08
fix(react): adds tabIndex -1 to notification dismiss buttons 

abbeyhrt:
REF: #10422 

From looking at the [PR](https://github.com/carbon-design-system/carbon/pull/8560) that added in `aria-hidden=true` to the Notification close button, it seemed like there was a specific intent behind it so removing this line might not be the right move! Let me know!

 The button being aria-hidden causes an accessibility violation and removes a way for screenreader users to dismiss notifications, having the button not be hidden removes the violation and allows the notification to be dismissed through screenreader(VO at least)

Also updates comment to state "status" as the default and removes the information for "alertdialog". 

#### Testing / Reviewing

Verify that this change makes sense given the context around InlineNotification, that the component has no accessibility violations and that the Notifications can be dismissed through screenreader.


tay1orjones:
Double check me here @dakahn but I think it was originally added because anything with a `role` of `status || log || alert` can't have interactive children. The close button is effectively an affordance for non-disabled users.

> The button being aria-hidden causes an accessibility violation

Could you share more details on what it specifically fails?

> ...(it) removes a way for screenreader users to dismiss notifications

I believe the intent was that this would be supplanted via the "<kbd>escape</kbd> to close" functionality. Again, reaffirming that the close button is only an affordance for sighted/mouse users.

abbeyhrt:
@tay1orjones Sorry! I put it in the audit and forgot to put it here:


With the "escape to close" functionality, how are screenreader users supposed to be able to dismiss the notification when `closeOnEscape` is false? Separately, it might just be me but I wasn't able to dismiss the notification on escape when using VO when `closeOnEscape` was true, so there was no way to dismiss the notification since I couldn't reach the button either.

abbeyhrt:
It could be the intent that they can't dismiss it visually though as far as I know! I'm focusing on visually dismissing it which  might not be necessary to think about when it comes to screenreaders, as long as the content is read and they can move on, then this could be okay! Like you were saying, the close button is a visual affordance that isn't needed for screenreaders. 

The only thing I think we would need to address for sure is the violation that I posted above.

abbeyhrt:
It could be an invalid violation as well. Looking at it, it seems like there would never be a way to aria-hide a button that receives focus even though there could be valid reasons to do so.

tay1orjones (approved):
Sorry for late getting back to this - I'm totally cool to roll with this to get rid of the error. It's always something we can circle back on without a breaking change based on feedback or other considerations we might run into.

Thanks for looking into it so thoroughly!

aledavila (approved):
lgtm. are there concerns for this or was there a reason this was added before ? @dakahn

dakahn:
just to clarify our intent here, we're considering this button a visual/redundant affordance for either our timeout or our escape to dismiss. That said there has been some recent W3C discussion around [_what 'focusable' means_ ](https://act-rules.github.io/rules/6cfa84) and that's where we're getting this new error.

I think the move here based on recent guidance and proposed standards changes is to change this components tabindex to -1.

Alternatively if the notification had a focusable button we'd need to steal and return our users focus here instead

Issue #10717, opened 2022-02-11
Replace  usages of old `Tooltip` with new `Toggletip`, `Tooltip`, `Popover`, etc.


tay1orjones:
In relation to the new Popover/Toggletip, we need to find internal usage of the old Tooltip and replace it with the new one. Alternatively we can document the existing usages to be refactored later.

jnm2377:
- Tabs (IconTab)
- Icon button
- CodeSnippet (copy btn feedback tooltip)
- CopyButton
- Pagination (next page buttons)
- PaginationNav
- UI Shell actions btns
- UI shell switcher btn
- stories:
    - FormLabel
    - Modal
    - ProgressIndicator
    - Tooltip?

abbeyhrt:
Just a small update: Tabs(IconTab) is using the new Tooltip. @jnm2377 I'm down to do 

- UI Shell actions btns
- UI shell switcher btn
- stories:
  - FormLabel
  - Modal
  - ProgressIndicator
  - Tooltip?

If you want to take on the top half?

dakahn:
Jumped in and updated the UIShell for us! Tried to see if you had something up already @abbeyhrt and I couldn't see a PR 🏄🏾 🌈

PR #10947, opened 2022-03-09
feat(Pagination, PaginationNav): migrate v11 versions to use new IconButton

jnm2377:
Ref #10717

#### Changelog

- updates v11 `Pagination` to use `IconButton`
- updates v11 `unstable_Pagination` to use `IconButton`
- updates v11 `PaginationNav` to use `IconButton`

#### Testing / Reviewing

- Make sure all tests pass
- Make sure components look and work as expected
- Make sure tooltip is showing up on hover


tay1orjones (commented):
There's an odd edge case with Pagination in `carbon-react-next` where if the previous button gets disabled while the tooltip is visible, it sticks and doesn't dismiss. I found that if I hover over the stuck tooltip, it'll then dismiss.


https://user-images.githubusercontent.com/3360588/157977156-0fe3ca66-cf0e-4330-b5d8-afc9e4e14f98.mov

tay1orjones (approved):

dakahn (approved):

PR #10953, opened 2022-03-10
feat(react): switches close/open control to chevron if interactive content is present

### LINKED ISSUE #9442: Expandable Tile - no interactive elements inside interactive elements
### Discussed in https://github.com/carbon-design-system/carbon/discussions/9003

TO-DO:
- [ ] either update usage guidelines and examples for this to clearly state NO interactive elements nested within expandable tile
- [ ] or refactor component so that the whole tile isn't a button, rather just one part of it, and the interactive elements aren't nested within the button

Current workaround is to not use interactive elements within expandable tiles.

<div type='discussions-op-text'>

<sup>Originally posted by **dbollinger** June 23, 2021</sup>
### Usage questions

I am currently implementing some expandable tiles using the [react carbon component library (links to storybook)](https://react.carbondesignsystem.com/?path=/story/components-tile--expandable).  The [design system documentation](https://www.carbondesignsystem.com/components/tile/usage/#expandable)  mentions support for including "CTAs (like links)" in the content area below the fold. This seems a little misleading because the entire Tile component is implemented as a button, which means certain types of CTAs (like buttons) explicitly aren't supported, as button elements should not be nested within other button elements.

A specific use case I was looking for was to include filter tags on an expandable tile that could be clicked to remove a given filter from the item represented on the expandable tile.

Are my assumptions about CTAs correct here? It may be helpful to update the documentation to clarify these constraints, especially with respect to accessibility.

### General accessibility concerns

This might be more appropriate as a separate discussion, but in addition to the concern about which elements are semantically appropriate, I am also concerned about how screen readers might interact with static content like lists. In the design system documentation, the expandable example shows a recipe for guacamole, with a semantic layout that guides the user through the steps.

When the content is all inside a button, some screen readers (VoiceOver, in particular) will not interpret any of the semantic meaning of the button's inner HTML. For example, list metadata (i.e. "list, 3 items") and arial-labels won't be read by the screen reader, nor will heading levels or other element roles.  Instead, the screen reader will read the entire contents of the Button's inner HTML with no pauses or breaks.

It also seems that it is not possible to use a mouse to select text in the expandable tiles, both above and below the fold.  (I am using MacOS and trackpad, and this is the case in Safari, FF, and Chrome)

### Related issues and PRs

- https://github.com/carbon-design-system/carbon/issues/7552
- https://github.com/carbon-design-system/carbon/issues/6314
- https://github.com/carbon-design-system/carbon/issues/5557
- https://github.com/carbon-design-system/carbon/pull/5609
</div>

abbeyhrt:
Closes #9442 

Updates `ExpandableTile` to switch modes if it has interactive children. If there are interactive children, the control to toggle open switches exclusively to the chevron button. If there is no interactive content, the tile stays as it is now. This switch happens automatically. 

Question for devs: Should this switch between the two be done in a `useLayoutEffect`? 

#### Changelog

**New**

- Interactive content variant where only the chevron is controlling open/close 

**Changed**

- Updates the stories so the tiles aren't so insanely big. 

**Removed**

-`onBeforeClick` this seemed like a workaround for interactive content since the whole tile is a click area, with this change, I don't believe this prop is needed. 

#### Testing / Reviewing

Ensure that ExpandableTile with no interactive content looks and acts the same as before. 

Look at the ExpandableTile example with interactive content and ensure that it looks okay and works as expected, the button receives focus and controls toggle. Make sure the focusable content in the tile is reachable and that there an no violations present. 
 
For designers: I made the decision to remove the hover and focus for the whole tile when there is interactive content present. Since the whole tile is no longer clickable, it felt appropriate to remove the click indicators. Let me know if a different solution is preferred, I'm happy to change it!


packages/react/src/components/Tile/next/Tile.js
```diff
@@ -436,6 +437,12 @@ export function ExpandableTile({
     setPrevTilePadding(tilePadding);
   }
 
+  let interactiveContent;
+
+  if (belowTheFold.current) {
+    interactiveContent = getInteractiveContent(belowTheFold.current);
+  }
+
```
joshblack, line 445: @abbeyhrt I think there may need to be an effect used here because what can happen is that `belowTheFold.current` will be `undefined` on the first render. It isn't until later renders that this value will potentially be defined.

Here's a demo I made to try and isolate this, let me know if it relates or not!
https://codesandbox.io/s/spring-haze-ol99tt?file=/src/App.js

joshblack: @abbeyhrt I think there may need to be an effect used here because what can happen is that `belowTheFold.current` will be `undefined` on the first render. It isn't until later renders that this value will potentially be defined.

Here's a demo I made to try and isolate this, let me know if it relates or not!
https://codesandbox.io/s/spring-haze-ol99tt?file=/src/App.js

abbeyhrt: @joshblack Would something like this work better? I added aboveTheFold since I realized it should also change if there is interactive content there. 

``` 
useIsomorphicEffect(() => {
    const belowInteractiveContent = getInteractiveContent(belowTheFold.current);
    const aboveInteractiveContent = getInteractiveContent(aboveTheFold.current);

    if (belowInteractiveContent || aboveInteractiveContent) {
      setInteractive(true);
    }
  });
```

joshblack (approved):
Left an optimization note, feel free to ignore it if it's not relevant!

thyhmdo (approved):

Issue #10428, opened 2022-01-13
Radio button: v11 audit


aagonzales:
## Audit
A designer and developer will pair on this audit, some tasks may be discipline focused. Follow the checklist below for the features and items to review. If there are no issues with the item then check the item as complete. If any problems or bugs come up when auditing add a comment to this issue with the problem and how to correct it (if you know how). Bugs do not need to be fixed while auditing. 

## Resources
- [Design spec in Sketch](https://ibm.box.com/s/v4hk50eol2wkurvtxz295kw6pz094lq5)
- [V11 Website](https://v11.carbondesignsystem.com/)
- [SCSS](https://github.com/carbon-design-system/carbon/tree/next/packages/components/src/components)
- [Storybook](https://carbon-react-next.netlify.app/?path=/story/getting-started-welcome--welcome)

## Checklist

**Visual** (in React code) 
_Design and dev check_
- [ ] Component is using the correct design tokens (color and type), see design spec and scss code.
- [ ] Component is rendering correctly across themes and layers, see design spec and storybook.
    - [ ] White theme
    - [ ] Gray 10 theme
    - [ ] Gray 90 theme
    - [ ] Gray 100 theme
- [ ] Component is rendering correctly across browsers (check themes across browsers as well), see storybook.
    - [ ] Firefox
    - [ ] Safari
    - [ ] Chrome
- [ ] Component sizes are rendering and named correctly ([if applicable](https://v11.carbondesignsystem.com/migrating/guide/develop#size-prop-updates-breaking))

 **Website** (v11)
 _Design checks_
- [x] Style tab has correct design tokens listed 
- [x] Style and usage tabs are using the correct [size props](https://v11.carbondesignsystem.com/migrating/guide/design#sizing) names (if applicable)
- [ ] Usage tab is up-to-date for with any [v11 behavioral](https://v11.carbondesignsystem.com/migrating/guide/develop#changelog) changes (if any)
- [x] Live Demo has applied v11 changes
   - [x] Themes are rendering correctly
   - [x] Size props are named correctly
   - [x] No `light` props included

**Design Kits** (Sketch only)
 _Design checks_
- [x] Correct design tokens (type and color) are used
- [x] Component using the correct size prop names
- [x] Any additional v11 behaviors have been added

**Storybook**
 _Dev checks_
- [x] Confirm that prop table is populating 
- [ ] Take note of missing examples/stories (if any)

**React package**
 _Dev checks_
[...]

PR #10960, opened 2022-03-11
fix(react): update styles and adjust selected logic

abbeyhrt:
REF #10428 

Noticed while doing the audit that `defaultSelected` wasn't working. This PR adjusts how selected is set so that `defaultSelected` and `valueSelected` work as expected. Weirdly enough, the test for this passed even though the prop value was false  🤷‍♀️.   Also changes background-color from transparent to `$background` 

#### Changelog

**Changed**

- Changes comments because storybook tries to read <RadioButton> in comments for the proptypes as a real component and throws a warning. 

#### Testing / Reviewing

Ensure that `valueSelected` and `defaultSelected` in the v11 storybook are respected


packages/styles/scss/components/radio-button/_radio-button.scss
```diff
@@ -86,7 +86,7 @@ $radio-border-width: 1px !default;
     flex-shrink: 0;
     border: $radio-border-width solid $icon-primary;
     margin: rem(1px) $spacing-03 rem(2px) rem(2px);
-    background-color: transparent;
+    background-color: $background;
```
joshblack: I noticed that when we set the background explicitly this component starts to act oddly in places like Data Table. I wasn't sure if this was intended or if its background transparency was needed to correctly work in places like this

image

abbeyhrt: hmm yeah. I'm not sure if this is intended either. @thyhmdo with this being the effect when the background-color is `$background` and the RadioButton used in components, would we want to keep the RadioButton's background `transparent` to not have this effect or is this okay/intended?

abbeyhrt: I think I misread the audit issue, I don't think the component needed to be changed to use `transparent`, rather the website needs to be updated. Reverted the background change to the button should be good to go when used in other components 👍

PR #10999, opened 2022-03-16
chore(DataTable): size prop audit

sstrubberg:
REF #10295 

- Changed the storybook config in `main.js` to include subfolders in the `next` folder so that the Dynamic and Expansion stories could be found.
- Added the old v10 DataTable stories to the block-list in storybook config.
- Added FeatureFlag for `size` prop to only accept `xs`,`sm`, `md`, `lg`, and `xl`
- Added FeatureFlag for `lg` as a default for the `size` prop.
- Added Controls functionality to Storybook.
- Removed the "Small" story from the DataTable-Batch-Actions story in favor of adding the `size` controls for the "Default" story.
- Removed the "Small" story from the DataTable-Toolbar story in favor of adding the `size` controls for the "Default" story.
- Removed "Xl", "Sm", "Xs" stories and renamed "DefaultLg" -> "Default" in favor of adding `size` controls to the now "Default" Basic story.
- Removed Zebra story from the DefaultLg (now Default) Basic story in favor of Storybook Controls.
- Removed the redundant size-related stories in the Expansion story in favor of adding the `size` controls for the "Default" story.
- Removed old `compact`, `tall`, and `short` styles from `@carbon/styles`

#### Testing
- Ensure that the size controls for all DataTable stories are working as expected.


tay1orjones (approved):
Thanks for removing the deprecated styles, too!

dakahn (approved):

PR #11002, opened 2022-03-16
Use Akamai CDN for Plex

tay1orjones:
This removes the reliance on google fonts, instead loading Plex from Akamai CDN

We can cut this as an RC and check the website as a testing ground.

#### Changelog

**Changed**

- replaced gstatic links with those from [Akamai](https://github.ibm.com/webstandards/plex-cdn)

#### Testing / Reviewing

- All usage of plex across storybook, the elements previews, etc should now resolve from the Akamai server. You'll need to check in the dev tools if the fonts are being loaded from there and not locally.
- Double check to ensure that the font-faces included are the same and we're not missing any


tay1orjones:
I just pushed an update to ensure we're using `font-display: $carbon--font-display;` where we were previously. @joshblack any idea why we use this for mono, sans, and serif, but not for sans condensed?

joshblack:
@tay1orjones not sure on my end, doesn't hurt to add!

packages/type/scss/font-face/_mono.scss
```diff
@@ -11,420 +11,1156 @@
...
     font-display: $carbon--font-display;
     font-family: 'IBM Plex Mono';
-    font-style: italic;
-    font-weight: 300;
-    src: local('IBM Plex Mono Light Italic'), local('IBMPlexMono-LightItalic'),
-      url(https://fonts.gstatic.com/s/ibmplexmono/v4/-F6sfjptAgt5VM-kVkqdyU8n1ioSflV1gMoW.woff)
+    font-style: normal;
+    font-weight: 700;
+    src: local('IBM Plex Mono Bold'), local('IBMPlexMono-Bold'),
+      url('fonts/IBM-Plex-Mono/fonts/complete/woff2/IBMPlexMono-Bold.woff2')
```
joshblack: Quick question, should these URLs be using the akamai host? Just am seeing `fonts/` at the moment

packages/type/scss/font-face/_mono.scss
```diff
@@ -11,420 +11,1156 @@
...
+        format('woff2'),
```
joshblack, line 21: I think since this is the IE11 fallback we would just use the woff one instead of loading in woff2. I believe the rationale for this is that if they can't load the unicode-range versions it would fall back to this woff version, but if the woff2 is specified it may load that instead of the splits (not 100% sure though)

packages/type/scss/font-face/_mono.scss
tay1orjones: @joshblack Yes I've already pushed an update for it

packages/type/scss/font-face/_mono.scss
tay1orjones: @joshblack If that's the case, should we update this in the Plex repo? What I've done here is taken the [direct output](https://1.www.s81c.com/common/carbon/plex/mono.css) that is on the server and copied it over.

If I understand correctly, [the repo](https://github.ibm.com/webstandards/plex-cdn) behind this is simply spitting out the generated stuff from the Plex package without modification.

jeffchew: Not sure if IE11 even matters anymore, it is reaching EOL in June and usage is incredibly low at this point.

joshblack: @tay1orjones I went through and tested over in: https://codepen.io/joshblack/pen/jOYWBJe?editors=1100 and confirmed that it won't impact loading so don't sweat it!

tay1orjones:
@joshblack @dakahn status checks are now sorted and passing, ready for re-review

joshblack (approved):

dakahn (approved):

Issue #10879, opened 2022-02-28
Create `v11` branch


tay1orjones:
To support the efforts of the release, a new public `v11` branch needs to be created. It will hold all future development leading up to the v11 release.

- [x] Create the branch
  - Will/How does this impact our release process? Document the answer
- [x] Make it the default branch for PRs?
  - Do we want this branch to become the default PR branch at this point? 
  - It could be beneficial to have as little as possible reconciliation to do between `main` and `v11`. 
  - What would the workflow be for bugfixes to v10 during this time?
- [x] Set up `carbon-components` and `carbon-components-react` to correctly re-export from `@carbon/styles` and `@carbon/react`, respectively.

joshblack:
Going to close this out as I believe it's been addressed, feel free to re-open if that's not the case @tay1orjones

PR #11012, opened 2022-03-17
chore: remove workflows from v10 branch

tay1orjones:
Ref https://github.com/carbon-design-system/carbon/issues/10879

This removes github action workflows from the `v10` branch. Instead, `v10` workflows will be updated/added to `main`. It appears there are outstanding problems with initiating workflows from non-default branches. https://github.community/t/workflow-files-only-picked-up-from-master/16129/53

#### Changelog

**Removed**

- remove all workflows from v10 branch

#### Testing / Reviewing

- We'll probably just have to use admin rights to merge this in since status checks will all fail.


abbeyhrt (approved):

tay1orjones:
Next phase of this work is in https://github.com/carbon-design-system/carbon/pull/11013

joshblack (approved):

tay1orjones:
It looks like I was wrong. None of the workflows are triggering for `v10`. I think we need to have them in the `v10` branch in addition to `main`, https://github.com/carbon-design-system/carbon/pull/11021

PR #11013, opened 2022-03-17
chore(ci): update workflows for `v10` branch

tay1orjones:
Ref https://github.com/carbon-design-system/carbon/issues/10879

This updates our workflows to support the v11 release.

#### Changelog

**New**

- Add new workflows explicitly scoped to v10, on the `v10` branch events, using the `v10` branch instead of `main`

**Changed**

- Update the workflows triggered on/for `main`/`v11` to ignore tags that start with `v10*`

**Removed**

- Remove existing unneeded workflows from `v10` branch in https://github.com/carbon-design-system/carbon/pull/11012

#### Testing / Reviewing

- I don't think there's a reasonable way to test these unfortunately.
- Double check my syntax, that it makes sense and follows the proper config


.github/workflows/v10-ci.yml
```diff
@@ -0,0 +1,142 @@
...
+  #     - name: Install dependencies
+  #       run: yarn install --immutable --immutable-cache
+  #     - name: Build project
+  #       run: yarn build --ignore '@carbon/sketch'
+  #     - name: Run VRT
+  #       if: github.repository == 'carbon-design-system/carbon'
+  #       env:
+  #         PERCY_TOKEN: c9a21a3fde4fda0a0f822d633426ab26e2ab2c1cba55221d342d4047744c8c24
+  #       run: |
+  #         yarn run percy exec -- yarn workspace carbon-components-react test:e2e
```
joshblack:
```suggestion
```

I bet we could just remove this whole block if you want to

.github/workflows/v10-sync-generated-files.yml
```diff
@@ -0,0 +1,40 @@
+name: Sync generated files
```
joshblack: I think you can remove this one honestly if you want to, I think it's disabled ever since a change to the permissions for the token

.github/workflows/codeql-analysis.yml
```diff
@@ -2,9 +2,9 @@ name: 'codeql'
 
 on:
   push:
-    branches: [main]
+    branches: [main, v10]
   pull_request:
-    branches: [main]
+    branches: [main, v10]
```
joshblack, line 7: Would you want to just turn this off for the v10 branch?

.github/workflows/v10-deploy-vanilla-devenv.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     runs-on: macOS-latest
```
joshblack: I think you can remove this file since this workflow is disabled

joshblack (approved):

.github/workflows/codeql-analysis.yml
tay1orjones: 🤷 I don't mind leaving it on for peace of mind?

dakahn (approved):

PR #11022, opened 2022-03-18
fix(notification): adds `title`, `subtitle`, and `caption` props and removes closeOnEscape

abbeyhrt:
REF #10422 

Adds back `title`, and `subtitle`, and `caption`, and the associated styles,  to `Inline` and `Toast` notification. 

Updates spacing for `ActionableNotification` per the audit. 

Removes `closeOnEscape` for `Inline` and `Toast` and makes the notification close button non-focusable. 

Updates propTypes so that props with a default value are no longer required. My understanding is that, if the prop is required, then it should be provided by the user, not by the defaults but let me know if I'm mistaken!

#### Changelog

**Changed**

- updates tests, removes no longer needed `closeOnEscape` tests and adds required `title` prop. 

**Removed**

- `closeOnEscape` 

#### Testing / Reviewing

Verify that everything works as expected and looks as intended.


tay1orjones (approved):
Do we want the same title/subtitle props and styling for ActionableNotification?

image

abbeyhrt:
@tay1orjones I wasn't sure! I figured since it's a new component, we could move it more towards the ideal. Like what you were expressing here  "Choosing to migrate these to children was sort of a halfway point. Like, we won’t explicitly block you from using a node and including whatever you want, but we’re not going to formally encourage semantic styling via keeping the title subtitle caption props."  so I thought to keep it as just taking in children as a node.

tay1orjones (commented):
I know it's a bit counter to what we've discussed previously, but after talking with @dakahn and @kingtraceyj I think having `title`, `subtitle`, and `caption` props of type `string` is a great step towards the ideal for all three notification components.

Usage of strings encourages non-semantic content. We then control the markup to ensure non-semantic elements and styling are used. Only render the elements for these if they're provided.

Additionally having `children` prop available for all of these as an escape hatch.

dakahn (approved):

laurenmrice (changes requested):
**Inline and Toast notification padding**
- [x] The padding to the left of the title needs to be 16px. The 16px padding needs to also include the left bar. I think right now it is adding the padding to the right of the bar. 
inline and toast padding


**Actionable notifications**
- Instead of saying "provide content" as the example text. We could say something like "Content for actionable notifications goes here." Just so it fills up the space more for toast and inline variant examples.

- [x] The actionable toast in low contrast should not get a border. The border is only for inline notifications. 
Screen Shot 2022-03-22 at 11 14 42 AM

- [x] The actionable toast notification button should be modified to not include the traditional 64px right padding. Instead there should be 16px padding. 
actionable toast button

- [x] The actionable inline notification bottom padding is off, there is too much padding on the bottom. For single line and multi-line, there should be an equal amount of padding on the top and bottom of the notification. The text should sit center within the container of the notification. 
actionable inline notifcation padding

tay1orjones:
Thanks @laurenmrice! I've pushed an update that should fix all of those items.

@kingtraceyj @abbeyhrt My commit also adds the `title` and `subtitle` props to `ActionableNotification` as we discussed

packages/components/src/globals/scss/__tests__/__snapshots__/css--font-face-test.js.snap
```diff
@@ -4,680 +4,1634 @@ exports[`_css--font-face.scss experimental should output @font-face blocks from
 "@font-face {
   font-display: auto;
   font-family: 'IBM Plex Mono';
+  font-style: normal;
+  font-weight: 700;
+  src: local(\\"IBM Plex Mono Bold\\"), local(\\"IBMPlexMono-Bold\\"), url(\\"https://1.www.s81c.com/common/carbon/plex/fonts/IBM-Plex-Mono/fonts/complete/woff2/IBMPlexMono-Bold.woff2\\") format(\\"woff2\\"), url(\\"https://1.www.s81c.com/common/carbon/plex/fonts/IBM-Plex-Mono/fonts/complete/woff/IBMPlexMono-Bold.woff\\") format(\\"woff\\"); }
```
tay1orjones: Not quite sure why this was committed, I'll remove. It should only be in `v10` branch for now

laurenmrice (changes requested):
@tay1orjones 
Looking great! Just two things:

**Actionable**

Toast:
- The action button should be flush aligned vertically with the text above it. It is a tiny bit off and is too far right.
Screen Shot 2022-03-23 at 4 02 48 PM

- I thought we were not supposed to show rich text in the Actionable variants of inline and toast notifications? We need to remove rich text styling from the Actionable variants.

tay1orjones:
@laurenmrice I just pushed up a fix for the alignment. I forgot to take into account the component's left border when calculating the button's `margin-left` 🤦 

After discussing the rich text situation with @dakahn and @kingtraceyj, we felt a good middle ground would be to remove all semantic markup (so everything is now a `div`) but leave the props for `title` and `subtitle` and require these only be plain strings.

This ensures we're not providing any semantic markup or rich text that screen readers would omit. I think @kingtraceyj was going to add some guidance around it as well to ensure folks know screen readers essentially read the `title` and `subtitle` as one long sentence. This way we can keep the minor visual styling and provide more continuity between Notifications in v10 and v11.

dakahn:
@laurenmrice we can leave the rich text in there! As long as its understood that the type hierarchy/bold/italics or anything else wont be communicated to screen reader users 👍🏾

laurenmrice (approved):
Looks good to me 👍

kingtraceyj (approved):
good to go!

PR #11040, opened 2022-03-23
chore(sass): require `sass@>=1.33.0`

### LINKED ISSUE #10600: Update sass dependency version in v11
Mentioned in https://github.com/carbon-design-system/carbon/issues/9857#issuecomment-945894019, `@carbon/styles` v11 will support `sass@>=1.33.0`

For this we need to:
- [x] Update the peer dep range for the package
- [x] Remove any conditionals around `math.div` usage inside `packages/styles`
- [x] Add a note to migration guidance specifying this sass dependency version is a requirement and how to update

tay1orjones:
Closes #10600

For the files with a mirrored `filename.import.scss`, I compared the two files and ensured that the correct up to date changes were brought over to the file that remains. In some cases we had added stuff to `filename.scss`, but not to `filename.import.scss`.

#### Changelog

**Changed**

- update sass dependency ranges to be `sass@>=1.33.0`
- update migration guide to include `sass@>=1.33.0` requirement

**Removed**

- math.div conditionals
- `.import.scss` files supporting dart-sass/libsass that are no longer needed

#### Testing / Reviewing

- Double check to make sure there's no remaining math.div conditionals in the codebase
- Read migration guide changes to ensure they're clear, correct, etc.


joshblack:
@tay1orjones what do you think is going on with the layout package? Those errors were so odd to see around the `@return` expression 🤔 

It may be that the tests are using the old sass renderer in test-utils which is using `node-sass` and that's what is causing the problems (specifically `createSassRenderer` is now `SassRenderer.create`

joshblack:
@tay1orjones made a reduced PR over at: https://github.com/carbon-design-system/carbon/pull/11061 if it's helpful for only the peer deps range

tay1orjones:
Yeah closing in favor of https://github.com/carbon-design-system/carbon/pull/11061

PR #11044, opened 2022-03-23
docs(contributing): add link to windows instructions

tay1orjones:
Surfaced in https://github.com/carbon-design-system/carbon/discussions/11042


joshblack (approved):
Looks great, just needs prettier to run to format the file, I think

dakahn (approved):

PR #11047, opened 2022-03-23
feat(icons-react): swap next for es, lib

joshblack:
This PR swaps out the `next` directory in our `icons-react` for the default `es` and `lib entry points. It also updates our

#### Changelog

**New**

**Changed**

- The upgrade transform for icons has been updated to use `@carbon/icons-react` instead of `@carbon/icons-react/next`
- The react package has been updated with the results of the icon codemod
- The public API for icons has been updated with the new icon exports
- The icons package has been updated to no longer ship with `next` and now ships with `es` and `lib`

**Removed**

#### Testing / Reviewing

- Inspect the Public API snapshot for icons to see if each "sized" icon has a corresponding entry in the new snapshot
- Inspect icons in the storybook to verify that they are rendering as expected


tay1orjones (approved):

jnm2377 (approved):
🔥 phew!! what a massive set of changes! everything is working great on storybook & snapshots!

dakahn (approved):

PR #11055, opened 2022-03-23
fix(UIShellHeader): update action buttons to use TooltipIcon

### LINKED ISSUE #10717: Replace  usages of old `Tooltip` with new `Toggletip`, `Tooltip`, `Popover`, etc.

dakahn:
partially closes #10717 

updates uishell header action buttons and switcher to use TooltipIcon component. also updates styling so the icons sit right in the new component

### DIFF
--- a/packages/react/src/components/UIShell/HeaderGlobalAction.js
+++ b/packages/react/src/components/UIShell/HeaderGlobalAction.js
@@ -9,7 +9,7 @@ import cx from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { AriaLabelPropType } from '../../prop-types/AriaPropTypes';
-import Button from '../Button';
+import TooltipIcon from '../TooltipIcon';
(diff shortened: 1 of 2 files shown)


tay1orjones (changes requested):
There is a very small visual difference. The icons have moved upwards and are darker in color.


https://user-images.githubusercontent.com/3360588/160013646-7c47cfab-7fe2-4865-be6e-126651632767.mov

jnm2377 (changes requested):
I think these be using `IconButton` instead. That's what we've been migrating all the `hasIconOnly` buttons to.

abbeyhrt:
@dakahn I think with the changes from this PR: #11106, these buttons will automatically use IconButton under the hood. Not sure how this impacts this PR but just wanted to let you know in case it does!

PR #11070, opened 2022-03-25
feat(type): update default type resolver to use Akamai CDN

### LINKED ISSUE #11008: Use Akamai CDN for Plex in v11
In https://github.com/carbon-design-system/carbon/pull/11002 the v10 codebase was updated to use the Akamai CDN by default instead of google fonts. 

We need to update the default font resolver in `packages/styles` to ensure this change is applied for v11 as well.

https://github.com/carbon-design-system/carbon/blob/f99361d232a7bd610e95c7b990c2535458e8c21c/packages/styles/scss/fonts/_src.scss#L99

tay1orjones:
Closes #11008 

Updates v11 to use the [Akamai CDN](https://github.ibm.com/webstandards/plex-cdn) to load Plex.

#### Changelog

**New**

- Added a new snapshot that logs out the `src` values of the `@font-face` declarations so we can track them and more easily understand the output of the resolver.

**Changed**

- Cherry picked https://github.com/carbon-design-system/carbon/commit/418681e64742e532b99e2fc33b964a87c8fe7fb1 from the `v10` branch to ensure we have no usages of gstatic links.
- Updated the default resolver to use Akamai instead of Google Fonts.
- Updated the global config variable name to `$use-akamai-cdn`

#### Testing / Reviewing

- Double check all the CDN links in the snapshot work and resolve
- The snapshot should be loading the same number, type, name etc files with and without the `$use-akamai-cdn: true;`


joshblack (approved):

joshblack:
bump @aledavila @jnm2377 when you get a chance to review

aledavila:
@joshblack to test we just change the akamai to true and run the snapshot tests?

packages/styles/scss/fonts/__tests__/fonts-test.js
```diff
@@ -91,6 +96,56 @@ describe('@carbon/styles/scss/fonts', () => {
...
+    expect(emitted).toMatchSnapshot();
+  });
+
+  it('should emit src from akamai cdn if $use-akamai-cdn is true', async () => {
+    const { result } = await render(`
+      @use '../../config' with (
+        $use-akamai-cdn: true,
+      );
+      @use '../' as fonts;
+    `);
```
tay1orjones, line 109: @aledavila the tests actually already configure the flag on - the output is in the `.snap` files.

So from a testing perspective it should just be making sure that the snapshots make sense and the links resolve

tay1orjones: You totally could turn the flag on globally though in your local environment and load up storybook as a smoke test to see if everything looks right and the akamai assets are being pulled in.

aledavila: Thanks Taylor. That’s easier let me check that

aledavila (approved):
Lgtm

jnm2377 (approved):

Issue #8418, opened 2021-04-15
Remove deprecated code from React package


tw15egan:
Part of #8410 

This is an issue to track the effort to remove deprecated code from the React codebase

- Remove props that have been marked as deprecated
- Remove deprecated components
- Remove any code noted for removal for next major release

tay1orjones:
Any prop marked deprecated in a `next/` should be double checked against the v10 version to understand if we should remove the deprecation in both or just v10. 

There are props we've marked as deprecated in v11, we're aiming to remove only the props deprecated currently in v10.

- [ ] `Accordion/Accordion.Skeleton.js`
- [ ] `Accordion/AccordionItem.js`
- [ ] `Button/Button.js`
- [ ] `Checkbox/Checkbox.js`
- [ ] `ContentSwitcher/ContentSwitcher.js`
- [ ] `DataTable/Table.js`
- [ ] `DataTable/TableExpandHeader.js`
- [ ] `DataTable/TableToolbarSearch.js`
- [ ] `DatePickerInput/DatePickerInput.js`
- [ ] `DatePickerInput/next/DatePickerInput.js`
- [ ] `Dropdown/Dropdown.js`
- [ ] `Dropdown/Dropdown.Skeleton.js`
- [ ] `InlineLoading/InlineLoading.js`
- [ ] `Modal/Modal.js`
- [ ] `Modal/next/Modal.js`
- [ ] `NumberInput/NumberInput.js`
- [ ] `NumberInput/next/NumberInput.js`
- [ ] `OverflowMenu/OverflowMenu.js`
- [ ] `OverflowMenuItem/OverflowMenuItem.js`
- [ ] `RadioTile/RadioTile.js`
- [ ] `Search/Search.js`
- [ ] `Search/next/Search.js`
- [ ] `Select/Select.js`
- [ ] `Slider/Slider.js`
- [ ] `StructuredList/StructuredList.js`
- [ ] `StructuredList/next/StructuredList.js`
- [ ] `Tab/Tab.js`
- [ ] `Tile/Tile.js`
- [ ] `Tile/next/Tile.js`
- [ ] `TimePicker/next/TimePicker.js`
- [ ] `TimePickerSelect/TimePickerSelect.js`
- [ ] `TimePickerSelect/next/TimePickerSelect.js`
- [ ] `Tooltip/next/DefinitionTooltip.js`

tay1orjones:
### Remove deprecated components

- [ ] Icon
[...]

PR #11092, opened 2022-03-28
feat(React): remove component props that were deprecated in v10

jnm2377:
Rel #8418 

#### Changelog

- remove props that were deprecated as part of v10
- props that were deprecated in v11 (such as `light` among others) were **not** removed

#### Testing / Reviewing

- ensure all tests pass
- ensure components are working as expected:
    - NumberInput
    - OverflowMenuItem 
    - RadioTile
    - Search
    - Select
    - Slider
    - StructuredList
    - SelectableTile
    - TimePickerSelect


packages/react/src/components/NumberInput/next/NumberInput.js
```diff
@@ -77,7 +75,6 @@ const NumberInput = React.forwardRef(function NumberInput(props, forwardRef) {
     [`${prefix}--number--light`]: light,
     [`${prefix}--number--nolabel`]: hideLabel,
     [`${prefix}--number--nosteppers`]: hideSteppers,
-    [`${prefix}--number--mobile`]: isMobile,
```
tay1orjones, line 80: Did we get the styles deleted for this one too?

packages/react/src/components/Select/Select.js
```diff
@@ -222,7 +212,13 @@ Select.propTypes = {
    * `true` to use the light version. For use on $ui-01 backgrounds only.
    * Don't use this to make tile background color same as container background color.
    */
-  light: PropTypes.bool,
+  light: FeatureFlags.enabled('enable-v11-release')
+    ? deprecate(
+        PropTypes.bool,
+        'The `light` prop for `Select` is no longer needed and has ' +
+          'been deprecated in v11. It will be moved in the next major release.'
```
tay1orjones: I think the `light` prop warnings should include a link or instruction to use `Layer`. Maybe we link to somewhere in the migration guide?

tay1orjones (approved):
A couple comments. Are these all documented in the migration guide?

packages/react/src/components/NumberInput/next/NumberInput.js
jnm2377: @tay1orjones I checked the v11 styles and they had already been removed. 👍🏽

packages/react/src/components/Select/Select.js
jnm2377: I'll update! 👍🏽

packages/react/src/components/OverflowMenuItem/OverflowMenuItem.js
```diff
@@ -195,7 +182,10 @@ export default class OverflowMenuItem extends React.Component {
               <TagToUse
                 {...other}
                 {...{
-                  'data-floating-menu-primary-focus': primaryFocus || null,
+                  'data-floating-menu-primary-focus':
+                    (FeatureFlags.enabled('enable-v11-release') &&
+                      other.primaryFocus) ||
+                    null,
```
joshblack: With `primaryFocus` being removed, should this just be removed?

joshblack (approved):
Looks great! Just left one note, wasn't sure if it was relevant or not

dakahn (approved):

packages/react/src/components/OverflowMenuItem/OverflowMenuItem.js
jnm2377: @joshblack I wasn't sure if I should remove it completely bc there's no `/next` component for this, as in this file is used for v10 and v11, and it only needs to be removed for v11.

joshblack: @jnm2377 if it helps, you can treat this file at the moment as "v11". So if you want to remove it for v11 go for it 👍  All the files for v10 are under the v10 branch now

Issue #10427, opened 2022-01-13
Progress indicator: v11 audit


aagonzales:
## Audit
A designer and developer will pair on this audit, some tasks may be discipline focused. Follow the checklist below for the features and items to review. If there are no issues with the item then check the item as complete. If any problems or bugs come up when auditing add a comment to this issue with the problem and how to correct it (if you know how). Bugs do not need to be fixed while auditing. 

## Resources
- [Design spec in Sketch](https://ibm.box.com/s/v4hk50eol2wkurvtxz295kw6pz094lq5)
- [V11 Website](https://v11.carbondesignsystem.com/)
- [SCSS](https://github.com/carbon-design-system/carbon/tree/next/packages/components/src/components)
- [Storybook](https://carbon-react-next.netlify.app/?path=/story/getting-started-welcome--welcome)

## Checklist

**Visual** (in React code) 
_Design and dev check_
- [ ] Component is using the correct design tokens (color and type), see design spec and scss code.
- [ ] Component is rendering correctly across themes and layers, see design spec and storybook.
    - [ ] White theme
    - [ ] Gray 10 theme
    - [ ] Gray 90 theme
    - [ ] Gray 100 theme
- [ ] Component is rendering correctly across browsers (check themes across browsers as well), see storybook.
    - [ ] Firefox
    - [ ] Safari
    - [ ] Chrome
- [x] Component sizes are rendering and named correctly ([if applicable](https://v11.carbondesignsystem.com/migrating/guide/develop#size-prop-updates-breaking))

 **Website** (v11)
 _Design checks_
- [ ] Style tab has correct design tokens listed 
- [x] Style and usage tabs are using the correct [size props](https://v11.carbondesignsystem.com/migrating/guide/design#sizing) names (if applicable)
- [ ] Usage tab is up-to-date for with any [v11 behavioral](https://v11.carbondesignsystem.com/migrating/guide/develop#changelog) changes (if any)
- [ ] Live Demo has applied v11 changes
   - [ ] Themes are rendering correctly
   - [ ] Size props are named correctly
   - [x] No `light` props included

**Design Kits** (Sketch only)
 _Design checks_
- [ ] Correct design tokens (type and color) are used
- [x] Component using the correct size prop names
- [x] Any additional v11 behaviors have been added

**Storybook**
 _Dev checks_
- [ ] Confirm that prop table is populating 
- [ ] Take note of missing examples/stories (if any)

**React package**
 _Dev checks_
[...]

PR #11094, opened 2022-03-28
feat(react): removes support for renderLabel prop and adjust progress logic

abbeyhrt:
REF #10427 

This PR temporarily removes support for `renderLabel` on ProgressStep. It was an inaccessible pattern and is currently broken in v11 due to changes to `Tooltip`. We can address replacing it in the future but I wanted to get this breaking change in since I don't think we'll be able to fix it before the deadline. 

Separately, I noticed that the `complete` prop on `ProgressStep`, if put on the first step didn't work so I adjusted the logic  for that. 

#### Changelog

**New**

- {{new thing}}

**Changed**

- how ProgressIndicator determines the "current" step when on the first step.
- tokens used have been updated

**Removed**

-  renderLabel prop

#### Testing / Reviewing

Verify that all instances of `renderLabel` have been removed. 

Verify that `complete` and `current` props work on `ProgressStep` as expected.


tay1orjones (approved):
@abbeyhrt the public API snap needs updated

joshblack (approved):

PR #11116, opened 2022-03-30
fix(ExpandableSearch): update to work with functional Search component

### LINKED ISSUE #10749: [Bug]: ExpandableSearch story not working
### Package

@carbon/react

### Browser

Chrome

### Package version

10.53.0

### Description

Surfaced in https://github.com/carbon-design-system/carbon/pull/10643, the `expandable` story for `Search` isn't working in the `carbon-react-next` storybook for `@carbon/react`.

### Steps to reproduce

https://user-images.githubusercontent.com/3360588/153900365-c0f6906c-b853-4cc2-9faf-76b6b68c3d9e.mov

### Code of Conduct

jnm2377:
Closes #10749 

#### Changelog

- makes `Search` forward a ref
- combines `Search` input forwarded ref with internal ref
- adds `onExpand` prop to `Search` so that `ExpandableSearch` can focus the input ref when the magnifier is clicked

#### Testing / Reviewing

- test `ExpandableSearch component via keyboard and mouse
    - component should expand on click/keyboard focus
    - if there is no text in the field, component should collapse onBlur
    - if there is text in the field, component should remain open onBlur
- compare behavior and styles to [v10 ExpandableSearch](https://react.carbondesignsystem.com/?path=/story/components-search--expandable) to make sure it's working as expected 
- ensure all tests are passing


packages/react/src/components/Search/next/Search.js
```diff
@@ -95,7 +100,17 @@ function Search({
 
   return (
     <div role="search" aria-labelledby={searchId} className={searchClasses}>
-      <div className={`${prefix}--search-magnifier`} ref={magnifier}>
+      {/* the magnifier is used in ExpandableSearch as a click target to expand, 
+      however, it does not need a keyboard event bc the input element gets focus on keyboard nav and expands that way*/}
+      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
+      <div
```
joshblack, line 106: Is it possible for you to change this element to a native `button`? I think `role="button"` implies a lot of semantics that we don't really provide here

joshblack (changes requested):

packages/react/src/components/Search/next/Search.js
jnm2377: @joshblack it doesn't function as a button in our default `Search`, only in the `ExpandableSearch`, which is why I didn't change it. It's kinda tricky. We also don't use it as a button in keyboard expandable search, it gets completely overlooked and the input gets focus.

jnm2377: In terms of screen reader a11y, I tested it with VoiceOver, and it's announcing as expected. 

It seems like we have a few different use cases for Search:

- persistent search mouse navigation
- persistent search keyboard/sr navigation
- expandable search mouse navigation
- expandable search keyboard/sr navigation

and out of these 4 scenarios, only the 3rd one uses this  `div` as a `button`, so what would the best thing be here?

joshblack: @jnm2377 sounds good  👍

joshblack (approved):

abbeyhrt (approved):

PR #11118, opened 2022-03-30
feat(react): deprecate handleClick and handleKeydown, update story with href

abbeyhrt:
Adds an href to ClickableTile story so it receives focus as expected. Also removes deprecated prop. 

#### Testing / Reviewing

Check that ClickableTile is focused as goes to carbondesignsystem.com and that all instances of `handleClick` and `handleKeyDown` were removed.


jnm2377:
@abbeyhrt I'm already handling the prop deprecation here https://github.com/carbon-design-system/carbon/pull/11092/files

joshblack (approved):

joshblack:
This seems to be the correct diff for Tile (removing the old props). Currently, with #11092 I'm still seeing `handle*` props present, is this change still valid or is there something missing?

dakahn (approved):

jnm2377 (approved):

PR #11164, opened 2022-04-06
feat(styles): emit layer tokens in theme mixin

joshblack:
This PR updates our theme mixin and zones to include the level-one layer tokens in order to work around the behavior where a layer token is defined at the root level:

```scss
:root {
  --layer-one: #ffffff;
  --layer: var(--layer-one);
}
```

Later on, a zone may try to redefine what `layer-one` is:

```scss
.zone {
  --layer-one: #000000;
}
```

However, this will **not** re-evaluate `layer` to be the updated color. Instead, we need to re-emit the `--layer` CSS Custom property in order for it to have the correct value. This is currently what our layer-one helper class does, this PR adds support for this out-of-the-box.

#### Changelog

**New**

- {{new thing}}

**Changed**

- Update theme to wrap the themes `theme` mixin to add-on level one layer set tokens
- Update zone to emit level one layer set tokens

**Removed**

- Removed outdated files in the React package

#### Testing / Reviewing

- Verify that the theme mixin emits the level one layer set tokens at the top-level
- Verify that zones emit level one layer set tokens


joshblack:
cc @jnm2377 @abbeyhrt @tay1orjones when you get a sec to review today 🙏 Would be great to get in the release

abbeyhrt (approved):

jnm2377 (approved):

PR #11209, opened 2022-04-12
chore(project): remove node-sass as a dependency

joshblack:
This PR removes `node-sass` as a dependency as it is no longer needed.

#### Changelog

**New**

**Changed**

**Removed**

- `node-sass` is no longer listed as a dependency across the project
- The default `test-utils` for Sass have been removed

#### Testing / Reviewing

- Verify that styles still render as expected in elements and storybook


aledavila (approved):
storybook and elements are working as expected

tay1orjones (approved):

PR #11227, opened 2022-04-13
feat(styles): emit component tokens in zones

### LINKED ISSUE #11180: Add component tokens to zones / inline themes
We currently emit component tokens through the `theme` mixin if a component is used, but we do not currently do that for the zones that are emitted. At the moment, they only emit values from the themes.

This issue is to update our zone implementation to correctly identify which component tokens are used and make sure that they are also emitted within a zone.

joshblack:
Closes https://github.com/carbon-design-system/carbon/issues/11180

This PR updates our zone strategy to emit component tokens. This will unfortunately emit all component tokens but we could update this to conditionally include based on usage in the future. The core challenge is that if we want to include component tokens based on usage this import must happen _after_ components are included in order for them to be registered appropriately.

#### Changelog

**New**

**Changed**

- Update `_zone.scss` to emit component tokens

**Removed**

#### Testing / Reviewing

- Verify the Theme story in Storybook emits component tokens for button, tag, and notification


dakahn (approved):

joshblack:

tay1orjones (approved):

PR #11246, opened 2022-04-18
docs(codesandbox): add empty yarn.lock file for codesandbox to resolve

joshblack:
This PR adds in an empty `yarn.lock` file so that CodeSandbox correctly installs packages using Yarn. We unfortunately cannot have this lockfile list correct dependencies given that this is a workspace in a monorepo

#### Changelog

**New**

- An empty yarn.lock file for the codesandbox example

**Changed**

**Removed**


tay1orjones (commented):
@joshblack Do we need this package to be built in the build process, or should we exclude it to get around the build failure?

PR #5822, opened 2020-04-08
feat(DataTable): option to show overflow menu on hover

asudoh:
This change introduces `overflowMenuOnHover` prop to `<DataTable>` that determines whether or not to show overflow menu on hover, by changing CSS class of `<Table>` as follows:

* `overflowMenuOnHover={true}` _unsets_ `bx--data-table--visible-overflow-menu`
* `overflowMenuOnHover={false}` _sets_ `bx--data-table--visible-overflow-menu`

That CSS class changes the behavior of overflow menu in `<TableCell className="bx--table-column-menu">`; The former above shows overflow menu only on hover, the latter above shows overflow menu always.

Though our latest design dictates that showing overflow menu on hover should be optional, this change sets `overflowMenuOnHover={true}` as the default. This is because of the following reasons:

1. `<DataTable>` has never set `bx--data-table--visible-overflow-menu` before and thus setting it by default will be a breaking change
2. Showing overflow menu always can alternately be (and has been in our stories) achieved by _not_ setting `className="bx--table-column-menu"` to `<TableCell>`

Fixes #5804.

#### Changelog

**New**

- `overflowMenuOnHover` prop to `<DataTable>`

#### Testing / Reviewing

Testing should make sure `<DataTable>` with overflow menu isn't broken.

Issue #5804, opened 2020-04-03
Overflow menu for data table rows should be visible on hover


ljitendra:
### Summary

Based on storybook link for data table component, the overflow menu is visible for every row. Based on UX discussions, there should be a prop to make the menu appear only on hover or visible for each row.

<img width="1384" alt="Screen Shot 2020-04-03 at 3 11 29 PM" src="https://user-images.githubusercontent.com/15458935/78396652-bdb13f00-75bd-11ea-961c-b64eff7d4239.png">

asudoh:
Hi 👋 we used to do what you suggested, but we switched to make the current way because users seem to have difficulties to discover the overflow menu icon. CC @carbon-design-system/design

ConradSchmidt:
Like @ljitendra said, the design team decided to specified the UX so that the menu is shown on row hover. Maybe this can be discussed with the Design team - to avoid overwriting Carbon, it would be nice to have this behaviour available as a prop (like "actionsOnHover").
If we would provide a PR would it be more feasible to get this implemented?

If you changed this recently, please update your Documentation:
https://www.carbondesignsystem.com/components/data-table/usage/
**Persistent overflow menu**
By default, the overflow menu icon appears **on hover or focus**. This reduces visual clutter and drives user interactions with clear context and results.

laurenmrice:
I think we could add a prop for overflow hover. 👍🏻

ConradSchmidt:
But just to be clear, is the Carbon site/guideline outdated, when it comes to this topic? Because the guideline says something different than @asudoh as quoted [here](#issuecomment-609653982)

laurenmrice:
Yes, we need to update that to what we have as default (persistent overflow icons)! A designer on our team is actually working on docs for data table this month, so we will update it.

joshua-vaughn:
So, I'm confused since this issue is closed but the documentation doesn't seem to have been updated yet.
- Is there a hover/persistent property?
- If so, is it now "persistent" by default in the code, and this just hasn't been updated in the data table docs yet?
- Could my application suite decide that "hover" is more appropriate for our data scientist users?

Thanks!

asudoh:
Hi 👋 #5822 introduces `overflowMenuOnHover`. The default is _not_ changed. Please see #5822 for more details.

andreagmann:
[...]

PR #11247, opened 2022-04-19
chore(data-table): update overflow menu storybook

### LINKED ISSUE #10747: [Bug]: DataTable overflowMenuOnHover is broken
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v0.13.0

### React version

v17.0.2

### Description

###Expected

`DataTable` sets `overflowMenuOnHover` to true by default. This should hide `OverflowMenu`s in `DataTableRow`s. 

### Actual
`OverflowMenu`s are never hidden.

### Reason
Documentation is rather poor on this behaviour, but digging through the commit history turned up [this commit](https://github.com/carbon-design-system/carbon/commit/6795904e0875165dcfb4be11e252b107b807163f). Even when `.cds--table-column-menu` is set on the correct `TableCell`, the `OverflowMenu` within it never disappears, because the SVG inside the Button that triggers the Menu is lacking the `.cds--overflow-menu__icon` class.

### CodeSandbox example

https://carbon-react-next.netlify.app/?path=/story/components-datatable-toolbar--with-overflow-menu

### Steps to reproduce

Go to https://carbon-react-next.netlify.app/?path=/story/components-datatable-toolbar--with-overflow-menu and note that the overflow menu is always visible, not just on hover.

### Code of Conduct

aledavila:
Closes #10747

Overflow menu wasn't hiding.

#### Changelog

**Changed**

- fixed classname in stories

#### Testing / Reviewing

Check data table overflow menu


aledavila:
@jnm2377 & @tay1orjones I wanted your take on this if possible. So the issue as it stands is that the overflow menu wasn't hiding when `overflowMenuOnHover` is true. There are 2 places where that exists I'm not even sure if the prop in Table works. The other is in DataTable component I think that is the one that works. (we can look into that another time)

So to my issue as you see in my change it wasn't working because the prefix was incorrect. In order for this prop to work the user needs to pass in `className="cds--table-column-menu"` to the `TableCell` containing the overflow menu. We don't exactly surface this or tell people they need to pass it in unless theres other ideas on how to integrate it into our code(the css might be too intertwined to do that). Do you guys have any ideas on the best way to do that ? So its not exactly a bug just a lack of documentation.

tay1orjones:
@aledavila Based on https://github.com/carbon-design-system/carbon/pull/5822 and https://github.com/carbon-design-system/carbon/issues/5804, `DataTable` should be passing `overflowMenuOnHover` down to `Table`. `Table` is the only spot where the `cds--data-table--visible-overflow-menu` class is toggled on/off. 

If `overflowMenuOnHover` requires placing `cds--table-column-menu` on a `TableCell` (and rendering an actual OverflowMenu), I agree we need to at the very least document it. A blurb in storybook docs could be enough.

On the bigger picture of how to integrate this more into our code, I wonder if the `TableCell` should have a `containsOverflowMenu` prop that toggles the `cds--table-column-menu` class so we avoid this problem in the future?

jnm2377:
@aledavila @tay1orjones I'm wondering if there is a different selector that can be used to toggle on/off the overflow menu so that users don't have to forcibly add `cds--table-column-menu` (since I'm assuming that classname is part of the selector that hides the overflow menu)?

aledavila:
@jnm2377 yes the way the styles were written it expects it to have that. And I thought about removing it but then other styles use that class

jnm2377:
I don't mean necessarily removing it from the stylesheet, but adding a selector that will support the use case of a column w/o that class so that it'll still work if a user doesn't add that class.

tay1orjones:
Would it work if the selectors using `.#{$prefix}--table-column-menu` were reconfigured to instead apply the styles if a table cell has an overflow menu as a descendant/child?

 Something maybe like `td:has(> .#{$prefix}--overflow-menu)`

tay1orjones:
Bah, nope, that's a [level 4 selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) not available in browsers yet. Bummer.

We might not be able to get out of having to apply the class to the table cell.

jnm2377 (approved):

tay1orjones (approved):

PR #11255, opened 2022-04-19
feat(styles): set background color on body

joshblack:
This use-case was previously captured under `$css--body` in: https://github.com/carbon-design-system/carbon/blob/v10/packages/components/src/globals/scss/_css--body.scss

This will set the background and text color of the page to the default values

#### Changelog

**New**

**Changed**

- Update our reset to include `background-color` and `color` on the `body` element

**Removed**

#### Testing / Reviewing

- Verify in storybook that the `body` element is correctly setting background-color and color to the background and text-primary CSS Custom properties


joshblack:
One question for reviewers: should we flag this behavior similarly under `css--body` or should we just have this as a default in the reset?

jnm2377:
> One question for reviewers: should we flag this behavior similarly under `css--body` or should we just have this as a default in the reset?

in v10, is `css--body` a feature-flag/opt-in?

joshblack:
@jnm2377 yes, it was a configuration option set to true by default

jnm2377:
> @jnm2377 yes, it was a configuration option set to true by default

Ah, ok. I think it would make sense to have this be a config option in v11 as well, then. Even if it's still true by default, in the case that someone doesn't want it on, they can just turn it off. But honestly, it's not a huge change either. I don't feel particularly strongly about it 😅 what are your thoughts?

jnm2377 (approved):

abbeyhrt (approved):

joshblack:
Just updated to a config flag! Thanks for the feedback @jnm2377 🙏

PR #11260, opened 2022-04-21
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,9 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
+permissions:
+  contents: read
+
 jobs:
   dedupe:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


tay1orjones:
@naveensrinivasan From my understanding this is unnecessary because these workflow actions are all ran in secure environments and not triggered via `pull_request_target`. These are ran under the context of the event that triggered them, which in these cases is only possible by someone who already has write access to the repository (inherently secure).

The CI workflow is ran on every PR, but by default is not provided secrets anyway since they always come from forks.

@joshblack I'm curious on your thoughts here

joshblack:
Going to close this out due to inactivity. Feel free to respond to the comments above and we can re-open this to continue the conversation!

PR #11266, opened 2022-04-21
docs(button): clearer usage instructions for renderIcon prop

### LINKED ISSUE #11168: [Bug]: Carbon 11 Button's renderIcon does not position correctly for functions
### Package

@carbon/react

### Browser

Firefox

### Package version

v11

### React version

v18

### Description

With Carbon 11, you can pass in a size to an icon such that instead of doing this in v10

```
import { Add24 } from '@carbon/icons-react';
```

you can do this in v11

```
import { Add } from '@carbon/react/icons';
...
<Add size={24} />
```

So here is my issue, before I could pass a Icon object into the `renderIcon` of a `Button` component as such 

```
 <Button renderIcon={Add24} />
```    

But if we wanted to use the new carbon 11 way passing in a size as such, you have to use a function for this to work 

```
 <Button renderIcon={() => <Add size={24}>} />
```    

Now I noticed these two different ways the icon position is off. 

Using an object (passing in something like `Add24`)

image

Passing in a function (`() => <Add size={24}>`)

image

The pos is off. I know we can just import bot h`Add` and `Add24` to solvethis , but wanted to ask if I was doing something wrong or if this was ab ug?

### CodeSandbox example

N/A

### Steps to reproduce

See Description

### Code of Conduct

tay1orjones:
Closes #11168 

Update the button storybook docs to show proper icon configuration for v11, `renderIcon` in particular. Needing to pass `props` when configuring `size` is mentioned briefly in the [migration docs](https://github.com/carbon-design-system/carbon/blob/c2c856ccca31ba2ef5bf55d2c13abcfecc7f706b/docs/migration/v11.md#L1415), but I've expanded on it here

#### Changelog

**Changed**

- update button storybook docs

#### Testing / Reviewing

- View the button storybook docs, new info should be present and render appropriately.


packages/react/src/components/Button/Button.mdx
```diff
@@ -213,15 +213,39 @@ with screen-reader support if you use the `renderIcon` prop. If you are trying
...
+  kind="danger"
+  iconDescription="TrashCan">
   Delete
 </Button>
 
 ```jsx
-<Button renderIcon={Add16} iconDescription="Add">Add</Button>
-<Button renderIcon={Delete16} kind="danger"  iconDescription="Delete">Delete</Button>
+<Button renderIcon={Add} iconDescription="Add">Add</Button>
+<Button renderIcon={TrashCan} kind="danger"  iconDescription="TrashCan">Delete</Button>
```
joshblack:
```suggestion
<Button renderIcon={TrashCan} kind="danger" iconDescription="TrashCan">Delete</Button>
```

packages/react/src/components/Button/Button.mdx
```diff
@@ -2,7 +2,7 @@ import { Story, Props, Source, Preview } from '@storybook/addon-docs/blocks';
 import { withKnobs, boolean, select, text } from '@storybook/addon-knobs';
 import Button from '../Button';
 import ButtonSet from '../ButtonSet';
-import { Add16, Delete16 } from '@carbon/icons-react';
+import { Add, TrashCan } from '@carbon/react/icons';
```
joshblack:
```suggestion
import { Add, TrashCan } from '@carbon/icons-react';
```

I think we'll need this to stay `icons-react` on our end, unfortunately, since technically `@carbon/react` shouldn't be defined at this level (since we're authoring the library)

I think this will work in storybook if it looks upwards for our local package.json but may not always work out, potentially.

jnm2377 (approved):

packages/react/src/components/Button/Button.mdx
tay1orjones: Ohh interesting. Yeah it works now but I didn't even think about it. Might be a webpack local resolution feature or something? Either way I agree and would rather not rely on it.

joshblack (approved):

PR #11270, opened 2022-04-22
chore(progress-bar): move to stable

aledavila:
Progress Bar was moved to stable but still exported as unstable. Renamed export to just stable. 

#### Testing / Reviewing

Make sure progress bar works as intended


jnm2377 (approved):

joshblack (approved):

joshblack:
For the merge commit, could the scope of the commit be updated to `fix` or `feat`? I think either of those more clearly reflects the change since the Public API of the package is being updated as a result of this diff.

PR #11275, opened 2022-04-22
fix(Pagination): disabled icon button tooltip & focus state for a11y

### LINKED ISSUE #11032: [a11y]: v11 Tooltip/IconButton behavior for Pagination
### Package

@carbon/react 

### Browser

Chrome

### Operating System

MacOS

### Package version

0.16.2

n/a

### Assistive technology

n/a

### Description

While migrating v11 pagination (https://github.com/carbon-design-system/carbon/pull/10947/) to use our new v11 tooltip, we realized the new tooltip implementation was behaving differently, specifically in the case for disabled IconButtons and we had some a11y questions around it. Really, there are 2 behaviors I have questions about. Explaining the issues in detail below:

1. Pagination buttons have what may be an edge case scenario, where a user will perhaps be hovered on an IconButton just as it becomes disabled. This is because the next & previous buttons will become disabled when there are no longer more pages to go forward to or backward to. In v10, when the next/previous icon button becomes disabled and a user is hovering on it, the tooltip immediately disappears. In v11, this does not happen, the tooltip remains visible even once the button is disabled until you hover over the tooltip.

After a discussion about it last week, Josh helped me come up with a list of possible implementations that would make v11 match v10 behavior (tooltip disappears as soon as the button is disabled). However, we were wondering in terms of a11y if there was any argument for keeping a tooltip on a disabled button or what the best practice would be for this instance. Listing the options we came up with below:

```js
 // Option 0: tooltip on disabled is okay
 // Option 0.5: aria-disabled for pagination buttons

 // Option 1: directly sniff child props
 // Cons: It assumes that the direct child has the `disabled` prop
 // You could have any number of layers between this component and the disabled button
 // A -> B -> C -> Button{disabled}
   useEffect(() => {
     if (child.props.disabled) {
       setOpen(false);
     }
   }, [child.props.disabled]);

 // Option 2: let's scan our tree to see if there is something interactive or not
   useEffect(() => {
     const interactiveContent = getInteractiveContent(containerRef.current);
    console.log(interactiveContent);
     if (!interactiveContent) {
       setOpen(false);
     }
  });

 // Option 3: tell us when the underlying button changes
  const buttonRef = useRef(null);
  const triggerProps = {
    ref: buttonRef,
    onFocus: () => setOpen(true),
    onBlur: () => setOpen(false),
    // This should be placed on the trigger in case the element is disabled
    onMouseEnter,
  };

   useEffect(() => {
     function subscriber(mutationList) {
       mutationList.forEach((mutation) => {
         if (
           mutation.type === 'attributes' &&
           mutation.attributeName === 'disabled' &&
           mutation.target.disabled
         ) {
           setOpen(false);
         }
       });
     }

     const observer = new MutationObserver(subscriber);

     observer.observe(buttonRef.current, {
       attributes: true,
       childList: false,
       subtree: false,
     });

     return () => {
       observer.disconnect();
     };
   }, []);
```

2. In addition to that, I noticed that there is another behavior that is inconsistent in our v10 and v11 IconButton . In v10 pagination, when you are hovered on any of the next/prev buttons, the tooltip disappears after the first click. So even if you stayed hovered over it, and are clicking forward to move pages, it doesn’t show up again after the first activation of the button. However, in v11, the tooltip stays up the whole time. I’m wondering in terms of a11y, what is expected to happen?

### WCAG 2.1 Violation

_No response_

### CodeSandbox example

see videos below

### Steps to reproduce

Check v10 & v11 storybook Pagination stories to compare behaviors

### Issue 1 videos

https://user-images.githubusercontent.com/32556167/159340338-6a6d85f9-486e-4965-a739-3e52742d25d0.mov

https://user-images.githubusercontent.com/32556167/159340325-b8fd339c-b5bb-4963-a29c-99936ce1f425.mov

### Issue 2 videos

https://user-images.githubusercontent.com/32556167/159340333-1bef0540-4b82-483f-969e-14c6ddc509ad.mov

https://user-images.githubusercontent.com/32556167/159340531-f54ba95c-211e-49e0-8502-fa87585acd04.mov

### Code of Conduct

jnm2377:
Closes #11032 

#### Changelog

- fixes an issue with the pagination next/prev icon buttons where the tooltip would stay visible even if the button would become disabled
- also fixes an a11y issue that arises by the fact that the disabled icon button loses focus and the `main` tag gets focused when this happens. This is a bad experience for keyboard and screen reader users. When discussed in CAG, the solution we came up with was that upon a prev/next button becoming disabled, the other button would gain focus. 

#### Testing / Reviewing

Go to pagination story
  - test next/previous buttons
    - click next all the way to the end until it becomes disabled
        1. tooltip should disappear immediately
        2. the previous btn should gain focus once the next becomes disabled
     - click the prev button all the way to the beginning
         1. tooltip should disappear immediately
         2. the next btn should gain focus when the prev btn becomes disabled
     - the same should happen when navigating via keyboard


tay1orjones (approved):
ooo very cool solution! Makes total sense and looks really awesome ✨

aledavila (approved):
nice work. i jst learned a neat trick!

PR #11276, opened 2022-04-22
fix(AspectRatio): Accept 3x2 ratio prop type

francinelucca:
Adding '3x2' ratio as a valid prop type, prevent this error: 
image

This is already supported: https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/aspect-ratio/_aspect-ratio.scss#L25

#### Changelog

**Changed**

- packages/react/src/components/AspectRatio/AspectRatio.js: added 3x2 prop type


packages/react/src/components/AspectRatio/AspectRatio.js
```diff
@@ -62,7 +62,16 @@ AspectRatio.propTypes = {
...
    */
-  ratio: PropTypes.oneOf(['16x9', '9x16', '2x1', '1x2', '4x3', '3x4', '1x1']),
+  ratio: PropTypes.oneOf([
+    '16x9',
+    '9x16',
+    '2x1',
+    '1x2',
+    '4x3',
+    '3x4',
+    '3x2',
```
joshblack, line 72:
```suggestion
    '3x2',
    '2x3',
```

Would be awesome to include `2x3` in there too if you have the chance!

francinelucca: Done!

joshblack: 🙌

joshblack (approved):

aledavila (approved):

PR #11303, opened 2022-04-27
docs(examples): adds a working example for switching between light and dark mode. 

abbeyhrt:
This was requested a couple of times in our Slack channels: 

https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1651069247477039
https://ibm-studios.slack.com/archives/C046Y0YUD/p1651080265908999

So I added an example with the help of @joshblack of a way to implement light/dark mode using a nextjs example. This examples tries to provide a way to switch between light/dark mode throughout in an service but I have no idea if we can provide a "recommended" way, since there are a couple of different ways teams can go about this and some may work better for certain teams. 

#### Testing / Reviewing

Go to `examples/light-dark-mode` and run `yarn` and `yarn dev` and test out the theme to see how if it toggles between g10 and g90 as expected! Also check out the implementation to see if this would be a way we would want teams to implement this.


joshblack (approved):

jnm2377 (approved):
This is great! Thanks for doing this. I am curious about the other ways that people can do light/dark mode. It'd be great to build more examples in the future and maybe talk about pros and cons so people can decide on their own.

PR #11312, opened 2022-04-28
chore(release): v10.57.0

tay1orjones:
Release PR for v10.57.0

### DIFF
--- a/examples/codesandbox-styles/package.json
+++ b/examples/codesandbox-styles/package.json
@@ -1,7 +1,7 @@
 {
   "name": "codesandbox-styles",
   "private": true,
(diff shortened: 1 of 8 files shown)


tay1orjones:
@joshblack any problems you can think of continuing to publish `@carbon/react` and `@carbon/styles` to v0.x like this?

joshblack:
@tay1orjones I'd just flip both of those packages to `private` for now in `package.json` just in case

tay1orjones:
@joshblack makes sense, updated 👍

joshblack (approved):

jnm2377 (approved):

PR #11326, opened 2022-05-02
fix(codesandbox): update url to use version tag ref instead of main branch

### LINKED ISSUE #11282: Codesandbox broken, can not fetch dependencies
When loading the `@carbon/react` codesandbox, you're shown an error

image

This problem does appear to go away if you fork the sandbox, but can we fix it so the initial load doesn't have the error?

- [ ] Investigate and fix
- [ ] Ensure the same update, if needed, is provided to [@carbon/styles](https://codesandbox.io/s/github/carbon-design-system/carbon/tree/main/examples/codesandbox-styles) sandbox
- [ ] Ensure the same update, if needed, is provided to [carbon-components-react](https://codesandbox.io/s/github/carbon-design-system/carbon/tree/v10/packages/react/examples/codesandbox)
- [x] https://github.com/carbon-design-system/carbon-website/issues/2872

tay1orjones:
Closes #11282

Updates the codesandbox links to use a version tag ref instead of `main` to fix an issue where dependencies couldn't be downloaded.

This appears to be a caching issue with codesandbox related to github downtime, see https://github.com/codesandbox/codesandbox-client/issues/6443

#### Changelog

**Changed**

- update codesandbox links

#### Testing / Reviewing

- Confirm that the new links work for you, the sandbox should build and work as intended.


abbeyhrt (approved):
Oh yay! Was having such trouble with code sandbox. Is the downside for this that we will have to continually update it every time we release?

dakahn (approved):

tay1orjones:
@abbeyhrt Yes, unfortunately we'll have to update the links when we release new versions until the upstream issue is fixed.

PR #11328, opened 2022-05-02
feat(TreeView): support additional navigation and selection behavior

emyarod:
Related #6792

This PR adds preliminary support for <kbd>Home</kbd> and <kbd>End</kbd> navigation as well as <kbd>Shift + Home/End</kbd> and <kbd>Ctrl + A</kbd> selection according to the tree view APG

additional testing will be done to confirm the selection model for different platforms and operating systems

#### Changelog

**New**

- <kbd>Home</kbd> and <kbd>End</kbd> navigation
- <kbd>Ctrl + Shift + Home/End</kbd> and <kbd>Ctrl + A</kbd> selection 
- arm64 build artifacts

**Changed**

- support `event.code` in `match` utility function

#### Testing / Reviewing

- <kbd>Home</kbd> should move focus to the first enabled node in the tree
  - <kbd>Ctrl + Shift + Home</kbd> should select the current node and every selectable node up until the first node in the tree
- <kbd>End</kbd> should move focus to the last enabled node in the tree
  -   <kbd>Ctrl + Shift + End</kbd> should select the current node and every selectable node up until the last node in the tree
- <kbd>Ctrl + A</kbd> should select all selectable nodes in the tree


tay1orjones (approved):
For those on mac with keyboards without the home/end key:

<kbd>Home</kbd> = <kbd>fn</kbd> + <kbd>left arrow</kbd>
<kbd>End</kbd> = <kbd>fn</kbd> + <kbd>right arrow</kbd>

https://support.apple.com/en-us/HT201236

So the full sequence that worked for me was <kbd>fn</kbd> + <kbd>shift</kbd> + <kbd>left/right arrow</kbd>

mbgower:
Home/End navigation is a good addition.
Ctrl+A is a typical select all feature
Shift+Home is not a key I've ever heard used in a tree context before. I'm not even sure I've heard of it used in a select element. I have seen Ctrl+Sh+Home/End in a select, but not in a tree. So I do question you introducing this key combination. It doesn't seem to be in the APG.

joshblack (approved):
🔥

PR #11329, opened 2022-05-02
feat(progress-bar): port support for statuses from v10 to v11

tay1orjones:
This mirrors the changes from #10843 from the `v10` branch into `main`. I attempted to actually cherry pick the commit, but the histories are so diverged the merge was a nightmare. I manually ported the changes over.

@janhassel it would be awesome if you could give this a glance to ensure I'm on the mark here.

#### Changelog

**New**

- add support for statuses in progress bar
- add progress bar playground story

#### Testing / Reviewing

From the original PR:

- Run storybook
- Compare different statuses with the design guidance in https://github.com/carbon-design-system/carbon-website/pull/2688
- Status "finished" should set its value to `max`
- Status "error" should set its value to `0` and emit `aria-invalid`
- In the inline variant, the bar should be visually hidden (though accessible) with only the icon and label being shown
  - This is because the inline variant is supposed to be used when space is tight, e.g. in a data table cell and @thyhmdo and I agreed that priorizing the label and icon over the 100%-full bar would be the best option


janhassel (approved):
LGTM!

abbeyhrt (approved):
LGTM!

joshblack (approved):

Issue #8864, opened 2021-06-09
<Dropdown /> doesn't render title tooltip on hover in firefox


SwapnilAhire:
BUG -

    [Firefox 88.0.1] Tool tip is not rendered when I hover over any of the option in drop down .
    [Chrome/Firefox] There's a blue bottom line shows when I select any option from the drop down, instead it should show blue rectangle outline.

Description -

    I am on Firefox 88.0.1 and @carbon/ibm-security is at 1.41.0
    Attached screenshots that explains the issues.

![image](https://user-images.githubusercontent.com/6366084/121298707-ef383f80-c911-11eb-98eb-3584b57899cd.png)
![image](https://user-images.githubusercontent.com/6366084/121298729-feb78880-c911-11eb-9f31-31173194a334.png)

Actual implementation:
![image](https://user-images.githubusercontent.com/6366084/121298783-155ddf80-c912-11eb-8fb2-3e97da914d30.png)
![image](https://user-images.githubusercontent.com/6366084/121298802-1abb2a00-c912-11eb-9489-9a55294dae1e.png)
![image](https://user-images.githubusercontent.com/6366084/121298833-227ace80-c912-11eb-9b3e-901296a76aa2.png)

Expected behavior -

    Tool tip should render while hovering over any of the options in the drop down on Firefox browser
    On Chrome/Firefox, the blue outline rectangle should display for the selected option in the drop down

Actual behavior -

    Tool tip does not render on Firefox version 88.0.1, for Chrome browser it works as expected.
    On Chrome/Firefox, it shows just blue bottom outline below the selected option in the drop down

Steps for reproducing -

(TRY it in both Chrome and Firefox and hover over ellipsised text option)
Sandbox URL - https://codesandbox.io/s/stupefied-bush-ntvn0?file=/src/index.js

Browsers

    Firefox 88.0.1
    Chrome 90.0.4430.212

Optional information
 - none

joshblack:
Hi @SwapnilAhire! 👋 Thanks so much for taking the time to make this issue. Can confirm the text truncation browser tooltip issue, having a hard time re-creating the focus border issue though. Could you share more steps to reproduce?

SwapnilAhire:
@joshblack Thanks for you valuable feedback. After global css fix for dropdown, I can confirm that the focus border issue has been resolved and its not related to component at all but the tooltip issue still exists.

SwapnilAhire:
[...]

PR #11331, opened 2022-05-02
chore(Downshift): update dependency to latest

dakahn:
Updates our Downshift dependency to latest (v6.1.7)

_from downshift_

BREAKING CHANGES

    Update TS typings for selectedItem to accept null in both useSelect and useCombobox.

To migrate to the new change, update your types or code if necessary. selectedItem, defaultSelectedItem and initialSelectedItem now have Item | null instead of Item type. PR with the changes: https://github.com/downshift-js/downshift/pull/1090

    Update TS typings for itemToString to accept null for the item parameter, in useSelect and useCombobox + in Downshift where this was missing. useMultipleSelection type for itemToString stays the same as it can't receive null as item.

To migrate to the new change, update your types or code if necessary. itemToString: (item: Item) => string -> itemToString: (item: Item | null) => string}. PR with the changes: https://github.com/downshift-js/downshift/pull/1075 https://github.com/downshift-js/downshift/pull/1105

    Pass type to the onChange (onInputValueChange, onHighlightedIndexChange, onSelectedItemChange, onIsOpenChange) handler parameters, as specified in the documentation. Also updated the TS typings to reflect this + onStateChange - the type parameter was passed but it was not reflected in the TS types.

To migrate to the new change, update your types or code if necessary, better to view the changes in the PR: https://github.com/downshift-js/downshift/pull/985. Important: please update to the 6.0.2 version since it contains a couple of fixes for the changes in this Breaking Change. Final changes:

  stateReducer?: (
    state: UseComboboxState<Item>,
    actionAndChanges: UseComboboxStateChangeOptions<Item>, // UseComboboxStateChangeOptions has the correct typings
  ) => Partial<UseComboboxState<Item>> // this now reflects the correct return, which is the partial state.
  onSelectedItemChange?: (changes: UseComboboxStateChange<Item>) => void // changes have partial state + type
  onIsOpenChange?: (changes: UseComboboxStateChange<Item>) => void 
  onHighlightedIndexChange?: (changes: UseComboboxStateChange<Item>) => void
  onStateChange?: (changes: UseComboboxStateChange<Item>) => void
  onInputValueChange?: (changes: UseComboboxStateChange<Item>) => void

where

export interface UseComboboxStateChangeOptions<Item>
  extends UseComboboxDispatchAction<Item> { // type and optional parameters needed to compute the next state
  changes: Partial<UseComboboxState<Item>> // partial state changes proposed by the hook
}

export interface UseComboboxDispatchAction<Item> {
  type: UseComboboxStateChangeTypes
  shiftKey?: boolean
  getItemNodeFromIndex?: (index: number) => HTMLElement
  inputValue?: string
  index?: number
  highlightedIndex?: number
  selectedItem?: Item | null
  selectItem?: boolean
}

export interface UseComboboxStateChange<Item>
  extends Partial<UseComboboxState<Item>> {
  type: UseComboboxStateChangeTypes
}

BREAKING BEHAVIOURS

    [useCombobox]: When an item is highlighted by keyboard and user closes the menu using mouse/touch, the item is not selected anymore. The only selection on Blur happens using either Tab / Shift+Tab. PR with the changes: https://github.com/downshift-js/downshift/pull/1109

    [useCombobox & downshift]: When pressing Escape and the menu is open, only close the menu. When the menu is closed and there is an item selected and/or text in the input, clear the selectedItem and the inputValue. PR with the changes: https://github.com/downshift-js/downshift/issues/719

closes #11331


packages/react/package.json
```diff
@@ -50,7 +50,7 @@
     "@carbon/telemetry": "0.1.0",
     "classnames": "2.3.1",
     "copy-to-clipboard": "^3.3.1",
-    "downshift": "5.2.1",
+    "downshift": "6.1.7",
```
joshblack: Are there any user-facing breaking changes from doing a major update of `downshift`?

dakahn: updated the PR description with their changelog 👀

jnm2377:
How should we review this?

dakahn:
@jnm2377 sorry, for the lack of clarity there. Basically, we're updating a dependency for Dropdown, Combobox and Multiselect. So running through basically functionality, testing, and keyboard/screen reader functionality is preferable

dakahn:
Looks like updating breaks tooltips for truncated menu items in Chrome/Chromium browsers like [we're currently seeing in Firefox](https://github.com/carbon-design-system/carbon/issues/8864), which is interesting.

Could be related to the ref weirdness we're seeing in our `useIsTruncated` hook

dakahn (commented):
Looks like updating breaks tooltips for truncated menu items in Chrome/Chromium browsers like [we're currently seeing in Firefox](https://github.com/carbon-design-system/carbon/issues/8864), which is interesting.

Could be related to the ref weirdness we're seeing in our `useIsTruncated` hook

jnm2377:
> Looks like updating breaks tooltips for truncated menu items in Chrome/Chromium browsers like [we're currently seeing in Firefox](https://github.com/carbon-design-system/carbon/issues/8864), which is interesting.
> 
> Could be related to the ref weirdness we're seeing in our `useIsTruncated` hook

@dakahn is this blocked until we figure that out, then? not being able to read truncated text fully seems like an important issue, even though I think essential functionality (selecting items) still works. Idk. thoughts?

jnm2377:
@dakahn just wanted to follow up on this. Do we think we should defer updating since it breaks truncated tooltips?

dakahn:
yeah, until we get a handle on the existing bug we're aware of this is officially blocked id say

tw15egan:
Did a quick run-through of all the downshift components, and the only issue I saw was that I was able to keyboard navigate to and select a disabled item in `Multiselect`. All other elements seemed like they behaved correctly. Also noticed an issue with the `With Initial Selected Items` story, with a `disabled` initially selected. Not sure if that is intended. 

Tabbing / selecting disabled item
Screen Shot 2022-08-25 at 1 34 57 PM

Initially selected item is disabled
Screen Shot 2022-08-25 at 1 35 38 PM

PR #11337, opened 2022-05-03
test(vrt): generate vrt tests from storybook

joshblack:
This PR generates Visual Regression Tests based on storybook stories in `e2e/components`. These snapshots will serve as our baseline as we start to evaluate how effective this setup is for catching visual regressions.

#### Changelog

**New**

- Generate tests for components in `e2e/components`

**Changed**

**Removed**

#### Testing / Reviewing

- Go through the Percy status check and check out the screenshots, see if there are any that don't make sense to include or are rendering incorrectly


joshblack:
bump @tay1orjones @aledavila @abbeyhrt when you get a sec today!

tay1orjones (changes requested):
@joshblack Just a couple things I noticed

- [x] `theme: white, component: Modal, id: components-modal--with-state-manager` is kind of silly imo, really for all the themes not just white.
- [ ] `theme: g100, component: DatePicker, id: components-datepicker--range-with-calendar-with-layer` has an error screen for all themes
- [ ] `theme: g90, component: DatePicker, id: components-datepicker--single-with-calendar-with-layer` also has an error for all themes

joshblack:
@tay1orjones just removed the state manager story, not sure how the datepicker ones since they are working in prod so I'll see if they show up again on the rebuild 🤔

joshblack:
Seems like this should be addressed by: https://github.com/carbon-design-system/carbon/pull/11395

abbeyhrt (approved):

aledavila (approved):
lgtm except the ones taylor pointed out.

tay1orjones (approved):

tay1orjones:
For some reason I can't update this with latest from `main`, the button isn't there...

but I agree this should be good now that https://github.com/carbon-design-system/carbon/pull/11395 is in. I think https://github.com/carbon-design-system/carbon/pull/11416 would've also fixed it

PR #11345, opened 2022-05-04
chore(react): update cloud foundry manifest

### LINKED ISSUE #11343: Update deployment for v10 React storybook
The deployment process for the v10 React Storybook currently deploys to the production storybook: https://github.com/carbon-design-system/carbon/blob/v10/.github/workflows/v10-deploy-react-storybook.yml

We need to update the `manifest.yml` to point to the v10-react storybook, instead: https://github.com/carbon-design-system/carbon/blob/v10/packages/react/manifest.yml

joshblack:
Closes #11343 

#### Changelog

**New**

**Changed**

- Update CF manifest to point to v17 react cloud foundry app

**Removed**

### DIFF
--- a/.github/workflows/v10-deploy-react-storybook.yml
+++ b/.github/workflows/v10-deploy-react-storybook.yml
@@ -45,6 +45,6 @@ jobs:
         run: |
           cd packages/react
           yarn build-storybook
-          ibmcloud cf blue-green-deploy carbon-storybook \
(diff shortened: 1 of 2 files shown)


tay1orjones (approved):
The manifest should override any configs provided via the `ibmcloud` cli, but I pushed a commit updating the workflow to use the proper name just in case.

dakahn (approved):

PR #11350, opened 2022-05-05
fix(ComposedModal): wrap modal content in div to get VO in iOS working

### LINKED ISSUE #11046: [a11y]: ComposedModal - VO on iPhone, focus isn't trapped when modal is open
### Package

carbon-components-react

### Browser

Safari

### Operating System

MacOS

### Package version

10.55.0

### React version

16.9.0

Manual testing

### Assistive technology

iPhone 12 - VoiceOver

### Description

When testing the `ComposedModal` with VoiceOver on iPhone on Safari, the focus goes outside of the modal when user has swiped through all the items within the modal. Focus should remained trapped within modal.

https://user-images.githubusercontent.com/54281166/159734912-fd062c11-61a2-4fe9-8d1b-f4cb2d9ea793.mov

### WCAG 2.1 Violation

_No response_

### CodeSandbox example

NA

### Steps to reproduce

Go to storybook: https://react.carbondesignsystem.com/?path=/story/components-composedmodal--default on iPhone - Safari

1. Using VO, swipe through all modal content
2. see how VO then focuses on elements outside of the modal

### Code of Conduct

annawen1:
Closes [#11046 ](https://github.com/carbon-design-system/carbon/issues/11046)

VO on iOS - focus was not getting trapped in the `ComposedModal`. When user swipes through modal content, expected behavior is the focus will loop back to start of modal instead of going out into the browser tool bar.

BEFORE:

https://user-images.githubusercontent.com/54281166/167473026-cd76c4f4-4d0f-486a-a1c8-17b5e73a62a0.mov

AFTER: (I'm swiping back and forth to ensure both start and end sentinels redirect focus back within the modal)

https://user-images.githubusercontent.com/54281166/167473061-7af95c5e-97a8-4c69-bcac-a6b92a7ff468.mov

#### Changelog

**Changed**

- wrap modal content in another div - weird, but doing so gets VO to focus on the start/end sentinel buttons and redirect focus.
- create a `cds--modal-container-body` class for styling

#### Testing / Reviewing

- using VO, make sure focus in ComposedModal is trapped within modal until user closes the modal


annawen1:
> @annawen1 one thing that may be happening with VO on iOS is that `ComposedModal` needs to set inert and/or `aria-hidden` on content outside of the modal dialog, if I remember right. This is something we ran into when trying to build up `Dialog` for v11 where we set `aria-hidden` and `inert` on content that was not the modal dialog.
> 
> This could also totally be wrong haha, figured I'd share in case it helped out at all!

@joshblack yeah that was the #1 solution that I've seen when searching online, but we do have similar focus wrap logic for our web components locale modal that's working on VO iOS so I'm trying to see why it' doesn't do the same 🤔

packages/react/src/components/ComposedModal/next/ComposedModal.js
```diff
@@ -240,31 +240,32 @@ const ComposedModal = React.forwardRef(function ComposedModal(
...
         aria-label={ariaLabel ? ariaLabel : generatedAriaLabel}
         aria-labelledby={ariaLabelledBy}>
-        {childrenWithProps}
+        {/* Non-translatable: Focus-wrap code makes this `<button>` not actually read by screen readers */}
+        <button
+          type="button"
+          ref={startSentinel}
+          tabIndex="0"
+          className={`${prefix}--visually-hidden`}>
+          Focus sentinel
```
joshblack, line 255: Feel free to remove the contents here (Focus sentinel), I believe this was something added in the past to try and help out with DAP violations but I don't think is actually needed in terms of content.

packages/react/src/components/ComposedModal/next/ComposedModal.js
```diff
@@ -240,31 +240,32 @@ const ComposedModal = React.forwardRef(function ComposedModal(
...
       <div
-        ref={innerModal}
         className={containerClass}
         role="dialog"
         aria-modal="true"
         aria-label={ariaLabel ? ariaLabel : generatedAriaLabel}
         aria-labelledby={ariaLabelledBy}>
-        {childrenWithProps}
+        {/* Non-translatable: Focus-wrap code makes this `<button>` not actually read by screen readers */}
+        <button
```
joshblack, line 251: Just curious, will the buttons need to be `aria-hidden`?

annawen1: hmm I think setting `aria-hidden` would cause the focus wrapping to stop working because we do want the buttons to be accessible to VO. Only when VO focuses on the button does the focus wrapping logic get triggered to redirect focus.

packages/react/src/components/ComposedModal/next/ComposedModal.js
annawen1: If it's not doing any harm, I don't mind leaving it in 😄 It's easier to spot and helps with debugging

andy-blum: I've had issues in the past with the VO cursor not going to elements that don't have text inside them. I think having the text contents here is _critical_

packages/react/src/components/ComposedModal/next/ComposedModal.js
```diff
@@ -240,31 +240,32 @@ const ComposedModal = React.forwardRef(function ComposedModal(
...
+          type="button"
+          ref={startSentinel}
+          tabIndex="0"
```
andy-blum: As a button, this element should have a native tabindex, this line is probably not needed.

packages/react/src/components/ComposedModal/next/ComposedModal.js
```diff
@@ -240,31 +240,32 @@ const ComposedModal = React.forwardRef(function ComposedModal(
...
+          Focus sentinel
+        </button>
+        <div ref={innerModal} className={`${prefix}--modal-container-body`}>
+          {childrenWithProps}
+        </div>
+        {/* Non-translatable: Focus-wrap code makes this `<button>` not actually read by screen readers */}
+        <button
+          type="button"
+          ref={endSentinel}
+          tabIndex="0"
```
andy-blum: also this tabindex

packages/react/src/components/ComposedModal/next/ComposedModal.js
annawen1: Yup, just removed thanks!

tay1orjones (approved):

joshblack (approved):

PR #11361, opened 2022-05-06
fix(react): adjust validation logic for files uploaded

### LINKED ISSUE #11003: [Bug]: FileUploader case sensitive extensions
### Package

carbon-components-react

### Browser

Chrome

### Package version

v7.55.1

### React version

_No response_

### Description

I noticed the file extensions that are passed into the `accept` property are case sensitive. This may have been by design but using a mac, not sure about windows, the file browser filters files regardless of case. So if I specify .jpg, then upload via clicking through the file browser, all files except .jpg and .JPG are disabled. If I select a .JPG though it will then flag the file as an invalid type. I think the two should be consistent.

### CodeSandbox example

https://codesandbox.io/s/gifted-surf-8yz5z1?file=/src/FileUploaderExample.js

### Steps to reproduce

1. Ensure you have a file locally that has the extension .JPG
2. Click the file uploader to bring up the file browser, note it filters all files except .jpg & .JPG
3. Select the allowed .JPG file
4. Note that the file item is showing as invalid

### Code of Conduct

abbeyhrt:
Closes #11003 

This PR updates the validation logic so that the file extension is checked case-insensitively. 

#### Changelog

**New**

- {{new thing}}

**Changed**

- updates proptypes comments so that storybook doesn't warn inaccurately.

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Pull down PR, Go to Drag and drop container story example, update the "accepts" to be ".jpeg" and then upload a file that ends with ".JPEG" and make sure that the file is seen as valid.


joshblack (approved):
Looks great! Would it be helpful to add a test case for this or do you think that's already covered?

dakahn (approved):

joshblack (approved):

PR #11415, opened 2022-05-13
feat(react): added input change callback to multiselect filterable

TannerS:
Closes #

N/A

#### Changelog

**New**

Added a new callback for users can get the input as it is being typed.

Due to a carbon 10 -> carbon11 update on the filterable component, it broken some `ref` info for our service in production and this needs to be fixed sooner than later 

#### Testing / Reviewing

Wanting to test in storybook for current action detection but currently having issues with that


TannerS:

TannerS:

TannerS:
recheck

tay1orjones (approved):

joshblack (approved):

packages/react/src/components/MultiSelect/next/FilterableMultiSelect.js
```diff
@@ -165,6 +166,10 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
   }
 
   function handleOnInputValueChange(inputValue, { type }) {
+    if (onInputValueChange) {
+      onInputValueChange(inputValue, { type });
```
joshblack: One quick question, do you know if you need the `type` parameter here for your use case?  Just was wondering if we could drop that in case we ever need to swap out Downshift in the future

TannerS: @joshblack i thought about that and just thought,.. eh lets keep it up to the user to have full access

PR #11390, opened 2022-05-11
fix(ComposedModal): set focus back to trigger element on close

annawen1:
Closes #11059

When closing the `ComposedModal` in VO on iOS, the focus remains on the no longer visible close button is the user is not able to swipe out of the focus, leaving them stuck. This PR sets the `aria-hidden` attribute and provides an example within the storybook to set focus back to trigger element.

BEFORE:

https://user-images.githubusercontent.com/54281166/159930948-f1d44af0-1a38-41bf-8a1d-9d4899caccfb.mov

AFTER:

https://user-images.githubusercontent.com/54281166/167885002-e5387e8d-9fdc-4764-80fb-c563f3945947.mov

#### Changelog

**Changed**

- set focus to trigger element `onClose` in the storybook `ComposedModal - with state management` story
- set `aria-hidden` attribute so that the modal content isn't read on close

#### Testing / Reviewing

{{ Add descriptions, steps or a checklist for how reviewers can verify this PR works or not }}

Issue #11059, opened 2022-03-24
[a11y]: ComposedModal - VO - when closing modal, user becomes stuck on close button


annawen1:
### Package

carbon-components-react

### Browser

Safari

### Operating System

MacOS

### Package version

10.55.0

### React version

16.9.0

### Automated testing tool and ruleset

Manual testing

### Assistive technology

iPhone 12 - VoiceOver

### Description

When using VO on iPhone in Safari, user gets stuck on the close button when they close the `ComposedModal`

One thing I did to try to fix this issue on my end, was to pass a function to the `buttonOnClick` prop of the `ModalHeader` component that would set the focus back to the trigger element on modal close. However, the `onBlur` of the `ComposedModal` interferes with this, triggering the `wrapFocus()` and sets the focus to an item within the modal (which at that point is closed).

https://user-images.githubusercontent.com/54281166/159930948-f1d44af0-1a38-41bf-8a1d-9d4899caccfb.mov

### CodeSandbox example

N/A

### Steps to reproduce

1. Go to `ComposedModal` story: https://react.carbondesignsystem.com/?path=/story/components-composedmodal--with-state-manager on mobile with VO on.
2. Open modal, and swipe to close button on top right corner
3. Close modal
4. try to swipe out of close button

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

PR #11418, opened 2022-05-16
fix(ComposedModal): fix focus trap and refocus to trigger element when closed

annawen1:
Closes  
- #11046 
- #11059

Gets changes from https://github.com/carbon-design-system/carbon/pull/11350 and https://github.com/carbon-design-system/carbon/pull/11390 into the `v10` branch.

https://user-images.githubusercontent.com/54281166/168654043-4186899a-04b1-4549-a181-6ac60041da5f.mov

#### Changelog

**Changed**

- wrap children in another div to get focus wrapping to work
- set `aria-hidden` when modal is closed to prevent VO from reading out modal content when closed
- pass logic to re-focus to trigger element `onClose` in the ComposedModal story

#### Testing / Reviewing

test using VO on iOS
- go to ComposedModal, and swipe through modal content, making sure focus is trapped within modal
- also to go `State Manager` story to ensure when modal is closed, focus goes back to the "Launch composed modal" button

### DIFF
--- a/packages/components/src/components/modal/_modal.scss
+++ b/packages/components/src/components/modal/_modal.scss
@@ -106,6 +106,10 @@
     @include carbon--breakpoint(xlg) {
       width: 48%;
     }
+
+    .#{$prefix}--modal-container-body {
+      display: contents;
+    }
   }
 
   // -----------------------------
--- a/packages/styles/scss/components/modal/_modal.scss
+++ b/packages/styles/scss/components/modal/_modal.scss
@@ -108,6 +108,10 @@
     @include breakpoint(xlg) {
       width: 48%;
     }
+
+    .#{$prefix}--modal-container-body {
+      display: contents;
+    }
   }
 
   // -----------------------------
--- a/packages/react/src/components/ComposedModal/ComposedModal-story.js
+++ b/packages/react/src/components/ComposedModal/ComposedModal-story.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, { useState } from 'react';
+import React, { useState, useRef } from 'react';
 import ReactDOM from 'react-dom';
 import { action } from '@storybook/addon-actions';
 import {
@@ -275,6 +275,7 @@ export const PassiveModal = () => {
 };
 
 export const WithStateManager = () => {
+  const closeButton = useRef();
   /**
    * Simple state manager for modals.
    */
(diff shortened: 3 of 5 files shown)


packages/styles/scss/components/modal/_modal.scss
```diff
@@ -108,6 +108,10 @@
     @include breakpoint(xlg) {
       width: 48%;
     }
+
+    .#{$prefix}--modal-container-body {
```
annawen1, line 112: @tay1orjones should the new styles be added to both `packages/styles/scss/components/modal/_modal.scss` and `packages/components/src/components/modal/_modal.scss`?

packages/components/src/components/modal/_modal.scss
```diff
@@ -106,6 +106,10 @@
     @include carbon--breakpoint(xlg) {
       width: 48%;
     }
+
+    .#{$prefix}--modal-container-body {
+      display: contents;
```
joshblack, line 111: Just wanted to double-check, do we need to offer any fallbacks for IE11 support or should this work as expected?

joshblack (approved):

packages/styles/scss/components/modal/_modal.scss
joshblack: You may be able to get away with only `packages/components` in v10 at this point since the styles package is private, curious what you think @tay1orjones 👀  Definitely doesn't hurt having it in both places though (especially if that makes porting things back easier)

packages/components/src/components/modal/_modal.scss
annawen1: @joshblack oh I didn't think about that, are you guys still supporting IE11? the dotcom team has dropped IE from our supported browser list.

joshblack: @annawen1 it shouldn't be a problem in v11 but technically we do still support it for v10. If the file is already using non-IE11 supported features though it's definitely okay to ship as-is which is why I was curious if that was already in place

packages/styles/scss/components/modal/_modal.scss
tay1orjones: Yeah, same thought. Having the same in both spots is preferred/nice but not necessary.

tay1orjones (approved):
Looks good pending the comments from @joshblack 👍 

Once merged, does this need to go out in a v10 patch now, or can it wait until our release next week?

annawen1:
> Looks good pending the comments from @joshblack 👍
> 
> Once merged, does this need to go out in a v10 patch now, or can it wait until our release next week?

@tay1orjones I think we can sit on it till the release next week!

PR #11419, opened 2022-05-16
docs(release): add release guide

### LINKED ISSUE #11368: Release v11.3.0
- [x] Release
- [x] Present updated workflow to team
- [x] Update documentation to restart release rotation

joshblack:
Closes #11368 

This PR adds the first draft of a release guide with our new automation workflow

#### Changelog

**New**

- Add documentation for the release process

**Changed**

**Removed**


tay1orjones (approved):

docs/release.md
```diff
@@ -0,0 +1,187 @@
...
+| [Post release](#postrelease)     | Support the latest stable release and address any issues that may come up as a result of promoting the release to stable |
+
+### Prerelease
+
+The prerelease occurs on the first Tuesday of a sprint. During this stage, the
+release team will need to do the following:
+
+- [ ] Run the
+      [Version Workflow](https://github.com/carbon-design-system/carbon/actions/workflows/version.yml)
+      to automatically generate the prerelease versions for packages
```
jnm2377, line 74: This is all great! The only thing I think could maybe use more clarification is this step. What does running the Version Workflow mean or look like? This is the automated part, right? I just wouldn't know what this entails from just reading it.

joshblack: @jnm2377 great question, when you visit the link it should take you to this page where you can click "run workflow"

image

jnm2377: Ah, this is great. Could we add this screenshot to the docs as an example?

tay1orjones:
@joshblack could we add a couple things so we can completely do away with using box for this?

1.  Docs for v10
    1. How to determine if v10 needs released
    2. Steps to release v10 minor/patch
1.  Release rotation

Here's some things for these that I've been putting together in box that we could put here instead:

# How to determine if v10 needs released

To help determine if there has been code merged into the v10 branch since the last v10 release:
- [ ] Go to https://github.com/carbon-design-system/carbon/compare/
- [ ] Open the "base" ref dropdown
- [ ] Choose the "tags" tab
- [ ] Select the most recent v10.x tag
- [ ] Open the "compare" ref dropdown
- [ ] Select the `v10` branch
- [ ] View the diff
  - [ ] If the diff is empty, v10 does not need released
  - [ ] If the diff contains a list of commits, we need to release a new v10.x version


Additionally, check [the PR queue against the v10 base branch](https://github.com/carbon-design-system/carbon/pulls?q=is%3Apr+base%3Av10+sort%3Aupdated-desc). It might be that there is a PR open and waiting to merge that we could get in to include in the release.

# Release v10 version
 - Basically the contents of this doc https://ibm.ent.box.com/notes/925120581152
 
# Release Rotation

Slot | Lead | Sidekick
-- | -- | --
1 | Abbey | Josefina
2 | Josefina | TJ
3 | TJ | Taylor
4 | Taylor | Josh
5 | Josh | D.A.
6 | D.A. | Abbey

<div data-node-type="line" id="magicdomid24" class="ace-line"></div>

joshblack:
@tay1orjones just updated with some of the content you shared. The only thing I wasn't sure about was release rotation. I think if we want that documented in the project my preference would be to use something like a wiki page, if that makes sense

docs/release.md
joshblack: Done!

jnm2377 (approved):
This looks great! Thanks so much for doing this! 🥇

PR #11435, opened 2022-05-18
fix(form-group): remove default `hasMargin` prop

joshblack:
This PR removes the `hasMargin` prop from `defaultProps` in `FormGroup`. This prop used to be available in v10 but is no longer used as part of the `className` in v11

#### Changelog

**New**

**Changed**

**Removed**

- Remove the defaultProp value for `hasMargin`

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -3259,7 +3259,6 @@ Map {
   },
   "FormGroup" => Object {
     "defaultProps": Object {
-      "hasMargin": true,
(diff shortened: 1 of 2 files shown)


amtoussam:
@joshblack Should the story also get updated?

https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/FormGroup/FormGroup-story.js#L21

abbeyhrt (approved):

joshblack:
@amtoussam it should be good to go since that is the old v10 story, let me know if I'm misunderstanding though!

joshblack:
bump @dakahn when you get a sec

dakahn (approved):

PR #11452, opened 2022-05-20
feat(react): add Zone component for inline theming

joshblack:
This PR adds in a `Zone` component to accompany a common use-case of people using the `Theme` component. In particular, when trying to zone or use inline theming, it can catch people off guard that the background and text color are not being set for the given theme.

The `Zone` component has the same component API as `Theme` but will add an additional class, `.cds--zone`, which will set the background and text color. This will enable situations like:

```jsx
<Zone theme="g100">
  <Accordion>
    ...
  </Accordion>
</Zone>
```

To work as-expected without needing to add custom styles

#### Changelog

**New**

- Add `Zone` component to React

**Changed**

- Update _zone.scss to include a zone class
- Update story files for theme

**Removed**

#### Testing / Reviewing

- What do you think about this component? Is it a worthwhile addition to the project?


dakahn (approved):
Super quick! And super useful

jnm2377:
Hey @joshblack I have some questions:
- how is `Zone` different than `Theme` in terms of use case? I thought `Theme` was used for zoned theming.
- would it make more sense to just add this functionality to `Theme` rather than introducing a new component that is very similar? I think it could create confusion around when to use what or what each is for (but this may be from my own confusion about them).

joshblack:
@jnm2377 great questions, I think adding this into `Theme` could definitely be an option (and could make things less confusing)

Basically, we have two things going on:

- I want to set the theme for a part of the page
- I want to set the theme _and_ I want to initialize the background/text color of the container

Right now, using `Theme` will set the CSS Custom Properties to the correct value, e.g.

```css
.g10 {
  --cds-background: #f4f4f4;
  // ...
}
```

But it does _not_ use those values to set anything up (like background or text color). What zone would do would make this the default (setting the theme and adding background, text color), for example:

```css
.g10 {
  --cds-background: #f4f4f4;
  // ...
}

.zone {
  background: var(--cds-background);
  color: var(--text-primary);
}
```

For global themes, this is what is available behind the `$css--body` flag.

**Approaches**

I think there are several approaches we can take with this:

- Make this the default for theme classes
  - Advantages
    - Would support this use-case out of the box with no changes in React
  - Disadvantages
    - There would not be a way to opt-out of this behavior
- Add a prop to `Theme`
  - Advantages
    - Fewer components
    - Would support this case by default (if prop is set to true initially)
  - Disadvantages
    - Really would just be logistical (how do we ship this in a non-breaking way)
- Add a `Zone` component
  - Advantages
    - Clearly separates out `Theme` and `Zone` use-cases
  - Disadvantages
    - Can be confusing when to use what, or guidance may just be "you probably want to use Zone"

We might be able to go down the prop route, maybe some approaches could be:

- `background` and `textColor` props
  - These would be true by default
  - You can turn them off to not emit these in the zone
- ???

jnm2377:
@joshblack thanks for that thorough explanation! I understand now. I am a fan of going the prop route and trying to figure out how to make it work the existing `Theme` component. In terms of the cons:
> Really would just be logistical (how do we ship this in a non-breaking way)

Would another non-breaking way be to have it be `false` by default? Maybe there's like a `isZone` prop (or another bool name) that would allow users to use the zoning functionality by toggling the zone className. I know having it `true` by default probably makes more sense bc I think people will want the zoning more often than not, but since we don't want to introduce a breaking change, maybe that's an option?

joshblack:
@jnm2377 definitely down to add in a `zone` prop and we could set it to `true` by default, let me update the PR with that instead 👀 

```jsx
// Default, zone={true}
<Theme theme="white" />

// Turn off
<Theme theme="white" zone={false} />
```

PR #11473, opened 2022-05-23
feat(styles): export font weights and font family

### LINKED ISSUE #11459: Add missing exports to `styles/scss/type` entrypoint
We currently explicitly forward exports from the `@carbon/type` package in: https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/type/_index.scss

This entrypoint is currently missing the following exports:

- `$font-families`
- `@function font-family`
- `@mixin font-family`
- `$font-weights`
- `@function font-weight`
- `@mixin font-weight`

abbeyhrt:
Closes #11459


joshblack:
@abbeyhrt could you update the [Public API](https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/__tests__/type-test.js#L18) tests for this entry point to make sure it includes what has been added?

joshblack (approved):

jnm2377 (approved):
Thanks for adding this!

Issue #11367, opened 2022-05-09
Add tooltip (title) to Clear icon button in Search component


anhtuan-pham1:
_(note for maintainers, this was opened as a bug and converted to an enhancement proposal)_

### Package

carbon-components-react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v10.57.1

### React version

v18.2.0

### Description

There should be a tooltip for clear icon in the Search component as describe at:
https://github.com/carbon-design-system/carbon/issues/10233

### CodeSandbox example

Use the sandbox from live demo https://carbondesignsystem.com/components/search/usage#live-demo

### Steps to reproduce

1. Launch Sandbox from live demo (or use live demo).
2. Enter something in search input.
3. Hover over X (or clear).

When you hover over x (or clear), you should see a tooltip (title).

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

PaulJThompson:
This bug affects our product too.

dakahn:
To my knowledge the clear button in the search bar [has never provided a toolitp](https://v10.carbondesignsystem.com/components/search/usage/). A tooltip isn't a necessary visual affordance. It has a programmatic name accessible via an aria attirbute and is technically redundant since the user can hit escape in the search to clear the entered text.

Closing as a misreport, but I can reopen if theres something I missed 👍🏾

PaulJThompson:
Hi @dakahn , thanks for responding. Are you saying that you wouldn't consider a PR that implements a tooltip here, or that you just don't consider this to be a bug?

dakahn:
Of course we'd consider any PR addressing an accepted enhancement proposal if the case was made. But this isn't a bug. Triaged as a bug it was closed as a misreport.

dakahn:
Reopening as a enhancement proposal

PR #11474, opened 2022-05-23
feat: add title to clear button Search

anhtuan-pham1:
Adding title to clear button in Search Component. The clear button now has the same type of of title that was created in the NumberInput component.

Changed

* Add title ="Clear" in button element
* Update test case for clear button

Testing / Reviewing
* Based on issue https://github.com/carbon-design-system/carbon/issues/11367


abbeyhrt (approved):

packages/react/src/components/Search/next/Search.js
```diff
@@ -135,6 +135,7 @@ const Search = React.forwardRef(function Search(
         className={clearClasses}
         disabled={disabled}
         onClick={clearInput}
+        title="Clear"
```
joshblack: One idea for this field could be to use `closeButtonLabelText` which is currently being used to label the `<button>`.

```suggestion
        title={closeButtonLabelText}
```

One thing we would want to make sure is that `title` is configurable (for example, someone could specify an alternate title) given that people could use this component in different situations or would like to localize this text 👍

joshblack (commented):
Thanks so much for doing this! Left a quick question, curious what you think 👀

packages/react/src/components/Search/next/Search.js
anhtuan-pham1: Hi @joshblack I have updated the title to title={closeButtonLabelText} and the test cases. Thanks for your recommendation.

aledavila (approved):

dakahn (approved):

PR #11500, opened 2022-05-27
ci: add promote packages workflow

### LINKED ISSUE #11434: Automate updating versions to `latest` on `npm`
One of the last remaining manual steps of our process is swapping tags on `npm` from `next` to `latest`. This issue will tackle taking this step from a manual process to an automated process through a custom GitHub Action.

This custom GitHub action should support the following requirements:

- It should be triggered through a `workflow_dispatch` event
- It should be able to load the current workspaces in a monorepo
- It should be able to view the information available on the `npm` registry for the `latest` and `next` tags
- If the two tags are not the same version, make sure `latest` is updated to the version that `next` is set to (as long as it's a newer version) using `npm dist-tag add`

joshblack:
Closes #11434

This PR adds in a promote workflow that is manually dispatched to update packages to their latest version on `npm`

#### Changelog

**New**

- Add promote workflow
- Add promote action

**Changed**

**Removed**

#### Testing / Reviewing

Sadly we'll only be able to try this out in the release process itself 😞

We did use this in our last release run which gave us confidence to try it out in this format


tay1orjones (approved):
Looks awesome! Having a dry run option is really fantastic.

dakahn (approved):

PR #11508, opened 2022-05-31
feat(react): add `autoAlign` feature to popover

### LINKED ISSUE #11385: Adjust Popover to be visible on first render
Ideal: 

Popover moves into view if original positioning is outside of viewport when the Popover is opened.

abbeyhrt:
Closes #11385 

Adds `autoAlign` prop and corresponding feature to Popover component.  Popover with `autoAlign={true}` will only work on first render so resize and scroll events will currently have no impact on the positioning. 

#### Testing / Reviewing

Review code to make sure the approach is okay and go to the 'Automate' story for Popover to see if it is visible in all four instances


packages/react/src/components/Popover/index.js
```diff
@@ -23,20 +32,140 @@ const Popover = React.forwardRef(function Popover(props, ref) {
...
+
+    const rect = floating.current.getBoundingClientRect();
+
+    const conditions = {
+      left: rect.x < -16,
+      top: rect.y < -16,
+      right: rect.x + (rect.width - 16) > document.documentElement.clientWidth,
+      bottom:
+        rect.y + (rect.height - 16) > document.documentElement.clientHeight,
+    };
```
jnm2377, line 81: just wanna make sure I'm understanding correctly. this is comparing the popover measurements to the page margins? (-16?)

jnm2377: wondering if you could maybe do a one or two liner comment summarizing what `conditions` is for, for maintainers?

jnm2377:
This looks great! 🔥  Left some comments about maybe adding a few more contextual comments for maintainers, but it's not a dealbreaker. The story seems to be working great. I'm wondering if we should add some tests around this functionality, though?

packages/react/src/components/Popover/index.js
abbeyhrt: Exactly! The 16 is to account for popover's internal padding. I added a comment to explain, let me know if that helps :)

abbeyhrt:
@jnm2377 Thanks for looking through it! I think the only way to test it is through a VRT test.  Since it's just a visual change, there's no way really to meaningfully test this with RTL. There aren't any VRT tests for Popover yet and I asked @joshblack and he said to hold off on adding one for this variant until it does

jnm2377 (approved):

packages/react/src/components/Popover/next/Popover.stories.js
```diff
@@ -127,3 +127,61 @@ Playground.story = {
     (story) => <div className="mt-10 flex justify-center">{story()}</div>,
   ],
 };
+
+export const Automate = () => {
```
joshblack:
```suggestion
export const AutoAlign = () => {
```

Would this kind of story title make sense? `Automate` wasn't immediately clear to me but let me know if you feel differently!

packages/react/src/components/Popover/index.js
```diff
@@ -111,9 +247,14 @@ const PopoverContent = React.forwardRef(function PopoverContent(
   ref
 ) {
   const prefix = usePrefix();
+  const { floating } = React.useContext(PopoverContext);
+
+  const mergedRef = useMergedRefs([floating, ref]);
```
joshblack: Super small nitpick, the convention I think I've been using is:

```suggestion
  forwardRef
) {
  const prefix = usePrefix();
  const { floating } = React.useContext(PopoverContext);
  const ref = useMergedRefs([floating, forwardRef]);
```

packages/react/src/components/Popover/index.js
```diff
@@ -23,20 +32,142 @@ const Popover = React.forwardRef(function Popover(props, ref) {
...
+      setAutoAligned(true);
+      setAutoAlignment(alignment);
+    }
+  }, [autoAligned, align, autoAlign, prefix]);
+
   return (
-    <BaseComponent {...rest} className={className} ref={ref}>
-      {children}
-    </BaseComponent>
+    <PopoverContext.Provider value={{ floating }}>
```
joshblack: Since this is a context provider, it might help to either `useMemo` the object here to prevent unnecessary child re-renders or pass in `floating` directly since it will have a stable object identity (and prevent re-renders)

packages/react/src/components/Popover/index.js
```diff
@@ -7,13 +7,22 @@
...
+  floating: {
+    current: null,
+  },
+});
+
+const Popover = React.forwardRef(function Popover(props, forwardRef) {
   const {
     align = 'bottom',
     as: BaseComponent = 'span',
+    autoAlign = false,
```
joshblack, line 25: Did we ever document the different strategies we were looking at for auto alignment? e.g. `align="auto"` was another approach and I wasn't sure what the process was for `autoAlign`.

Are you hoping to ship this in stable next release or were you hoping to guard this behind a flag / indicate it's unstable?

packages/react/src/components/Popover/index.js
```diff
@@ -23,20 +32,142 @@ const Popover = React.forwardRef(function Popover(props, ref) {
...
+      return;
+    }
+
+    if (autoAligned === true) {
+      return;
+    }
+
+    const rect = floating.current.getBoundingClientRect();
+
+    // The conditions, per side, of when the popover is not visible, excluding the popover's internal padding(16)
```
joshblack, line 74: Will the popover always have this internal padding? I had thought this was specific to the story and wanted to check if that padding changed if this still would work

packages/react/src/components/Popover/index.js
```diff
@@ -23,20 +32,142 @@ const Popover = React.forwardRef(function Popover(props, ref) {
...
+      const option = options[i];
+
+      if (isVisible(option)) {
+        alignment = option;
+        break;
+      }
+      alignment = align;
+    }
+
+    if (alignment) {
```
joshblack, line 165: Will `alignment` always be defined due to L156 above? Or is there a case where it's still `null`?

joshblack (changes requested):
Looking great! Left some comments/questions

packages/react/src/components/Popover/next/Popover.stories.js
abbeyhrt: Agreed! I like AutoAlign as a title much better

packages/react/src/components/Popover/index.js
abbeyhrt: Ugh, dang I hadn't noticed that. I think i would still want to keep it though. The way that rect.x and rect.y work, it will be negative if it goes passed the padding of the page. I wanted to account for that a little bit so I figured 1rem was sensible.

packages/react/src/components/Popover/index.js
abbeyhrt: I did not document the different potential strategies. I can document what I ended up choosing  somewhere if you think that's necessary!  

 With `autoAlign` being off by default, I thought it would be okay to include it in the next release without a guard since people will have to opt in to it. I would prefer to not make it officially experimental but I'm down to put a note in the prop description for it

laurenmrice (approved):
Looks great Abbey! 🎉

tay1orjones (approved):

joshblack (approved):
🔥

PR #11514, opened 2022-05-31
chore(Copy, CopyButton): update tests to RTL

### LINKED ISSUE #11509: Use RTL for testing in `Copy/Copy-test.js`

### LINKED ISSUE #11510: Use RTL for testing in `CopyButton/CopyButton-test.js`

jnm2377:
Closes #11509 
Closes #11510 

#### Changelog

- updates `Copy` & `CopyButton` tests to RTL

#### Testing / Reviewing

- make sure tests pass


abbeyhrt (approved):

packages/react/src/components/CopyButton/CopyButton-test.js
```diff
@@ -5,97 +5,112 @@
...
-    it('should call the click handler', () => {
-      const onClick = jest.fn();
-      const clickWrapper = mount(<CopyButton onClick={onClick} />);
-      clickWrapper.simulate('click');
-      expect(onClick).toHaveBeenCalled();
-    });
+    expect(button).toHaveClass('cds--copy-btn--animating');
+    setTimeout(() => {
+      expect(button).not.toHaveClass('cds--copy-btn--animating');
+    }, 5000); // custom feedback timeout
```
joshblack: For timers, could you update this to use [timer mocks](https://jestjs.io/docs/timer-mocks) (just so we don't have to wait 5s for the test to complete)

joshblack (changes requested):

packages/react/src/components/CopyButton/CopyButton-test.js
jnm2377: ooh, I hadn't heard of this. Looking at the example, I'm wondering how that might apply here (sorry, if it's obvious). Would it literally just be adding 
```js
jest.useFakeTimers(); 
jest.spyOn(global, 'setTimeout');
```
 before the timeout or before all the tests? Just trying to figure out what it looks like in this situation bc it's different than their example. I'm not just trying to test that the timeout was called, more so trying to check that the animation changes correctly & for the amount of time designated (feedback animation during and after the timeout).

joshblack: @jnm2377 I think in this case you could:

- At the top-level write `jest.useFakeTimers()`
- In the test, you can do:

```jsx
expect(button).toHaveClass('cds--copy-btn--animating');
jest.runAllTimers();
expect(button).not.toHaveClass('cds--copy-btn--animating');
```

joshblack (approved):

PR #11525, opened 2022-06-01
fix(react): remove focus on click for panels with interactive content

### LINKED ISSUE #11522: [Bug]: Tabs - panel is focused on click even when there is interactive content
### Package

@carbon/react

### Browser

_No response_

### Package version

1.02

### React version

_No response_

### Description

When the panel area is clicked, even if there is interactive content in the panel, the panel is focused and can mess with interactive elements click events. This only happens on click, not on key down. 

### Reproduction/example

In storybook 

### Steps to reproduce

Go to any tabs example and click the tab with an interactive content. 

### Code of Conduct

abbeyhrt:
Closes #11522 

Adds or removes a new interactive class if a panel has interactive content. This is to remove the visual focus that was still showing up on click. Also adds more consistent and better examples for interactive content in tab panels. 

#### Testing / Reviewing

I added a story called "InteractiveContentUpdates` where you can click a button that will toggle interactive content in the first panel. Try it out and make sure that there is no focus on the panel when there is interactive content, and that the inputs/buttons work as expected and then remove the interactive content and make sure there is now focus on the panel itself.


jnm2377 (approved):

dakahn (approved):

abbeyhrt:
Don't merge yet! What I have here works for fixed content but the TabPanel content can be updated to be interactive so it needs to cover that use case as well

packages/react/src/components/Tabs/next/Tabs.js
```diff
@@ -637,18 +637,56 @@ const TabPanel = React.forwardRef(function TabPanel(
...
+    function callback() {
+      const content = getInteractiveContent(node);
+      if (content) {
+        setInteractiveContent(true);
+        setTabIndex('-1');
+      }
+
+      setInteractiveContent(false);
+      setTabIndex('0');
+    }
```
joshblack: Should this be in an `else` block? Looking at this in the diff view, this seems like it would set interactive content / tab index if content is truthy and then would set it to false/0 afterwards

joshblack (changes requested):

joshblack (approved):

PR #11533, opened 2022-06-02
chore(project): add initial typescript setup and config

### LINKED ISSUE #11317: [Bug]: @carbon/react/icons module does not provide IDE import auto-complete
### Package

@carbon/react

### Browser

_No response_

### Package version

1.1.0

### React version

### Description

image

image

image

image

image

### CodeSandbox example

n/a

### Steps to reproduce

### Code of Conduct

joshblack:
Closes #11317 

This PR creates an initial TypeScript setup within the project for us to experiment with for the `@carbon/icons-react` package. It also includes an initial implementation of the types for this package based on what exists in DefinitelyTyped.

#### Changelog

**New**

- Add initial TypeScript configuration to the project and initial CI workflow
- Add `Icon.d.ts` to the icons-react package

**Changed**

- Update icon-build-helpers to emit a root `index.d.ts` file in a package
- Update icons-react to specify typings file and include in published package

**Removed**

#### Testing / Reviewing

- Go over TypeScript configuration to see if it:
  - Makes sense for our monorepo setup
  - Has relevant compiler option rules enabled
  - Is being correctly used by our `typecheck` command to identify type errors


tay1orjones (approved):

jnm2377 (approved):

packages/icons-react/types/Icon.d.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright IBM Corp. 2019, 2019
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import * as React from 'react';
+
+type IconSize = 16 | 20 | 24 | 32 | '16' | '20' | '24' | '32' | number | string;
```
SimeonC, line 10: If this is changed to `... | '32' | (number & {}) | (string & {})` then autocomplete in IDE's will prefer the actual values rather than a generic string/number but still won't force it. See https://github.com/microsoft/TypeScript/issues/29729#issuecomment-505826972

jharvey10: @joshblack Are there numbers outside of 16, 20, 24, and 32 that an Icon can use?

metonym:
`@types/carbon__icons-react` has been updated to [v11.5.0](https://www.npmjs.com/package/@types/carbon__icons-react). The DefinitelyTyped maintainers were kind enough to merge the [updated Carbon icons types](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61229).

Are there plans to revive this PR? One of the maintainers recommended [generating a single `index.d.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61229#issuecomment-1205852622), which is not dissimilar to the approach in this PR.

 If type definitions are shipped with the package, `@types/carbon__icons-react` can be removed for good.

peterblazejewicz:
There was another suggestion on DT, that could be even better, using inline JSDoc comments (`@type` and the family), so there no work (no artefacts build steps) required for TS support for all of those symbols, thx!
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52664#issuecomment-834800128

Issue #8854, opened 2021-06-07
[DatePicker]: autocorrection of dates is inconsistent and provides no controls over its value


davidicus:
## What package(s) are you using?

- [x] `carbon-components`
- [x] `carbon-components-react`

## Detailed description

When entering a value into the date picker that is not a valid date and then tabbing focus away from the input the incorrect date is replaced with another properly formatted but widely different and probably incorrect date. 

**Examples**:
Typing in `33` will result in a date of `09/01/2031`
Typing in `99/9`  will result in a date of `03/09/2061`

Additionally, this behavior is inconsistent. If you look at [this](https://react.carbondesignsystem.com/?path=/story/components-datepicker--range) first "range" story and type `33` into first input then tab to the second and do the same and then tab out of the second input you will see that the first one was replaced while the second one was not. 

The "simple" story also seems to leave the date as entered. 

> Is this issue related to a specific component?

This issue is related to the `DatePicker` component. 

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I would expect either the value to be left alone and my own validation can handle the error state or there to  be a prop that allows me to enter in a default date that can work instead. Ideally it would be for this to be opt in and applications can handle their own error handling. 

> What browser are you working in?

This issue appears in both latest Chrome and FF

> What version of the Carbon Design System are you using?

This is the latest Carbon that is published to storybook. `v7.36.0`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

This question was raised to AI Applications PAL repo

## Steps to reproduce the issue

1. Go to [this](https://react.carbondesignsystem.com/?path=/story/components-datepicker--range) story
2. Type `33` into first input and tab to second
3. Type `33` into second input and tab away
4. Notice first input adds a questionable correction of `09/01/2031` and the second input leaves the number as was entered by user. 

## Additional information

![image](https://user-images.githubusercontent.com/1590966/121075348-13c1d980-c7a3-11eb-8fb3-8de2482a6851.png)

tw15egan:
Seems like this behavior is the default behavior by Flatpickr, which would add another reason to look into alternative external date picker libraries. I'm able to improve it slightly by using the `parseDate` [function](https://flatpickr.js.org/examples/#custom-parsing-and-formating), but it's still not great. 

```js
parseDate: (datestr) => {
            return new Date(datestr);
},
```

[...]

PR #11568, opened 2022-06-06
chore(Datepicker): update Flatpickr dependency

dakahn:
updates Flatpickr to latest as a pre-req to some work we're doing this sprint

### test

make sure datepicker works as expected and all functionality is as it was before updating 👍🏾 

Related: https://github.com/carbon-design-system/carbon/issues/8854


abbeyhrt (approved):

tay1orjones (changes requested):
Some funky stuff going on with manual input. The existing state in prod is also broken/odd, but a new bug has been introduced with this. `main` on the left, this PR on the right:


https://user-images.githubusercontent.com/3360588/172872384-54b327c3-5b60-42d9-8e56-2bb961060cb7.mov

tay1orjones:
@dakahn @abbeyhrt let's re-review w/ latest from `main` now that we've gotten a few of the other DatePicker fixes in

PR #11575, opened 2022-06-07
feat(react): update wrapComponent to bring in usePrefix

### LINKED ISSUE #11504: [Bug]: Prefix not working for TableToolbarContent
### Package

@carbon/react

### Browser

Edge

### Package version

v1.4.0

### React version

v18.1.0

### Description

Applied css prefix is not being applied to TableToolbarContent component.

### Reproduction/example

https://codesandbox.io/s/carbon-toolbar-yjh716

### Steps to reproduce

Class applied to toolbar content element is still cds--toolbar-content.

### Code of Conduct

joshblack:
Closes #11504

#### Changelog

**New**

- Add tests for `wrapComponent`

**Changed**

- Update the following components that use `wrapComponent` and `className`
  - TableActionList
  - TableToolbarContent
  - HeaderGlobalBar
- Update `wrapComponent` to support static and dynamic class names

**Removed**

#### Testing / Reviewing

- Verify that components that use `wrapComponent` and `className` can have their selectors correctly prefixed


packages/react/src/tools/wrapComponent.js
```diff
@@ -8,22 +8,30 @@
...
       // and so we have empty `class` attributes in the resulting markup
       // eslint-disable-next-line no-extra-boolean-cast
       className: !!componentClass ? componentClass : undefined,
     });
-  };
+  }
+
   Component.displayName = name;
   Component.propTypes = {
     className: PropTypes.string,
```
tay1orjones, line 32: Does this need to be updated to accept a string or function?

tay1orjones (approved):
Cool idea to update `className` to accept a function 👍  One question, otherwise LGTM

packages/react/src/tools/wrapComponent.js
joshblack: @tay1orjones I think it should be fine since it's the function that accepts the string/function but the component will always accept only the string 👍

dakahn (approved):

PR #11600, opened 2022-06-12
fix(styles): include CSS files in side effects

### LINKED ISSUE #11599: [Bug]: `@carbon/styles` should include `.css` in side effects
### Package

@carbon/styles

### Browser

Chrome

### Package version

1.4.0

### React version

_No response_

### Description

When using `styles.css` in a webpack app, the styles were not included because they are not listed in the side effects in `package.json`. The side effects currently include `.scss` files. I believe it should also include `.css` files for cases where the consumer uses the generated StyleSheet with an app bundler.

Specifically, I was attempting to use the StyleSheet with Carbon Charts. Note that `@carbon/styles` is a transitive dependency of `@carbon/charts`.

```js
<script>
  import "@carbon/styles/css/styles.css";
  import "@carbon/charts/styles.css";
</script>

```

### Reproduction/example

N/A

### Steps to reproduce

The bug can be reproduced by running this example: https://github.com/metonym/carbon-charts-svelte-examples/tree/webpack/webpack

### Code of Conduct

metonym:
Closes #11599

This PR includes the CSS files in the side effects of `@carbon/styles`.

#### Changelog

**Changed**

fix(styles): include CSS files in side effects

#### Testing / Reviewing

I tested this locally in my webpack app. Without the CSS files included in the side effects, webpack was ignoring the styles. With the CSS files included, the styles appeared as expected.

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -56,6 +56,6 @@
   "sideEffects": [
     "index.scss",
     "scss/**/*.scss",
-    "scss/**/*.css"
+    "css/**/*.css"
   ]
 }


metonym:
Closing in favor of #11608

metonym:
@joshblack Not a problem; let's go with yours.

PR #11608, opened 2022-06-13
feat(styles): add css folder to sideEffects

### LINKED ISSUE #11599: [Bug]: `@carbon/styles` should include `.css` in side effects

joshblack:
Closes #11599

Add the `css` folder and files to `sideEffects` in `package.json`

#### Changelog

**New**

**Changed**

- The `sideEffects` field in `package.json` now includes the `css` folder

**Removed**

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -56,6 +56,7 @@
   "sideEffects": [
     "index.scss",
     "scss/**/*.scss",
-    "scss/**/*.css"
+    "scss/**/*.css",
+    "css/**/*.css"
   ]
 }


tw15egan (approved):

metonym:
I was thinking about this and I now wonder if `scss/**/*.css` is a typo?

`grep`ing for `.css` files in `packages/styles/scss` yields no results.

```bash
find ./packages/styles/scss |grep -e "\.css"
```

dakahn (approved):

abbeyhrt (approved):

PR #11609, opened 2022-06-13
fix(ComposedModal): close modal with esc key when no element has focus

### LINKED ISSUE #10193: [Bug]: Pressing the escape key wont close composed modal
### Package

### Browser

Chrome

### Package version

10.45.0

### Description

The composed modal won't close when you hit the escape key when you don't have any focus elements within the modal

### CodeSandbox example

https://codesandbox.io/s/dry-shadow-7o0s7?file=/src/index.js

### Steps to reproduce

1. Open the modal with the button
2. De-focus the input field (by pressing out of the textInput)
3. Hit the escape button
4. The modal does not close

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/10193

After doing a bit of digging between `Modal` and `ComposedModal` it seems like the `div` needs a `tabIndex` in order for `onKeyDown` events to fire.

#### Changelog

**Changed**

- Added a `tabIndex` to the outer `ComposedModal` div so that all keypress events are captured

#### Testing / Reviewing

Open `ComposedModal with State Manager` example. Open the Modal, click on the text (to lose focus from the input), and try closing the modal with the <kbd>esc</kbd> key


jnm2377 (approved):
Sorry for the late review! It's working as expected for me! 🔥

joshblack (approved):

PR #11617, opened 2022-06-14
fix(datepicker): call setDate when value is updated

tay1orjones:
[Surfaced in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1655131902714149), the DatePicker wasn't setting the `value` when it was passed one. Small error using flatpickr's `set()` vs the intended `setDate()`

#### Changelog

**New**

- tests covering the use case of a controlled date picker

**Changed**

- Fix DatePicker to set `value` for flatpickr when given one

#### Testing / Reviewing

- I made a 'ForTesting' story that you can see is a simple example of controlling the input by tracking it's state change via onChange and passing new state via the 'clear' button. This should work as you would expect
- All other DatePicker stories should work as intended

Do not merge until I've removed the ForTesting story


packages/react/src/components/DatePicker/DatePicker-test.js
```diff
@@ -225,4 +227,47 @@ describe('DatePicker', () => {
...
+        </>
+      );
+    };
+
+    render(<DatePickerExample />);
+    expect(screen.getByLabelText('Date Picker label')).toHaveValue('');
+
+    userEvent.type(
+      screen.getByLabelText('Date Picker label'),
+      '01/20/1989{enter}'
```
jnm2377, line 264: woah, this is cool! never tried this one before!

jnm2377 (approved):
It's working as expected! 🔥

packages/react/src/components/DatePicker/DatePicker-test.js
jnm2377: Thanks for adding these tests!!

tw15egan (approved):
LGTM 👍 ✅

abbeyhrt (approved):

PR #11620, opened 2022-06-15
test(react): update stories to fix failing avt

abbeyhrt:
This PR updates `accessibility-checker` and resolves a violation that was showing up in some stories (but not the component itself, that seems to be good)

#### Testing / Reviewing

Make sure AVT ci check passes


joshblack (approved):

achecker.js
```diff
@@ -10,7 +10,7 @@
 const path = require('path');
 
 module.exports = {
-  ruleArchive: 'latest',
+  ruleArchive: '09June2022',
```
dakahn, line 13: we can't use `latest` anymore? is this an update on their end?

dakahn (approved):

achecker.js
joshblack: @dakahn I think when we use `latest` we run into an issue where the ruleset may change and PRs start failing all of a sudden, even with no related change 😞 

One idea was to specify the archive and then we update to the latest archive as they get released. This way, we can still stay up to date but not have to worry about our AVT job being flaky / failing unexpectedly.

PR #11623, opened 2022-06-15
docs(select): add playground story, controls

### LINKED ISSUE #11466: components/Select/next/Select.stories.js
**Checklist**

- [ ] Move `.stories.js` file out of `next` folder, or create `.stories.js` file and copy contents of `*-story.js` file over.
- [ ] Remove deprecated `knobs` package and usage.
- [ ] Ensure there is a playground story
- [ ] Add `argTypes` to playground story
- [ ] `bind()` it as a template as described in the storybook docs
- [ ] Pass `...args` to playground story, spread them on the applicable component(s)
- [ ] Remove deprecated props from being visible in controls of storybook

**Note: if the story has sub-components, document and share with @sstrubberg**

**Links & Resources**

- [DataTable example](https://github.com/carbon-design-system/carbon/pull/10999)

tay1orjones:
Closes #11466 

#### Changelog

**New**

- Add defaults for `bool` props in `Select`

**Changed**

- remove args/controls from non-playground stories
- add args and playground story
- provide defaults for controls via `args`

**Removed**

- {{removed thing}}

#### Testing / Reviewing

- Select stories should look and function, controls should only work on Playground story
- Playground story controls should function as expected when changed/interacted


joshblack (commented):
Playing around with the playground and I ran into a couple of quirks:

- When using `invalid`, I cannot set `invalidText` because it is of type `node`, trying to set it as a string breaks the playground and I have to refresh
- Same issue as above with `labelText`
- Same issue as above with `warnText`

aledavila:
Setting those 3 that @joshblack mentioned as below should fix that

`invalidText: {
      control: 'text',
 },`

tay1orjones:
@aledavila @dakahn @joshblack Fixed 👍 , also pulled out the unnecessary non-ui-impacting props as we discussed

joshblack (approved):

dakahn (approved):

PR #11627, opened 2022-06-16
feat(styles): add docs to index

alisonjoseph:
Closes #https://github.com/carbon-design-system/carbon-platform/issues/890

Add usage, style, code, and accessibility documentation to each component in Carbon styles index following [asset schema](https://github.com/carbon-design-system/carbon-platform/blob/main/docs/resource-schemas.md#asset-schema)

#### Changelog

**New**

- Add `docs` to each component that has corresponding docs inside carbon website repo


mattrosno (commented):
This looks great to me. Follows the spec.

Some background - this allows us to show component docs in platform while those docs still reside in the `carbon-website` repo. In the future, after platform is on Carbon's TLD, we can move these *.mdx files to this repo and in doing so, all of these paths would be relative.

tay1orjones (approved):

dakahn (approved):

francinelucca (approved):
this LGTM! niceee @alisonjoseph 🎉

PR #11634, opened 2022-06-16
fix(upgrade): move jscodeshift to external hard dep

tay1orjones:
[Surfaced in slack](https://ibm.enterprise.slack.com/archives/C2K6RFJ1G/p1655235017711649?thread_ts=1655235017711649&cid=C2K6RFJ1G)

#### Changelog

**Changed**

- move jscodeshift to external hard dep

#### Testing / Reviewing

Pull this down, run `yarn build` in `packages/upgrade`, inspect the built cli.js file and cmd+f for `jscodeshift`, it should be listed as a dependency


joshblack:
Screen Shot 2022-06-17 at 9 27 01 AM


Do we need to worry about these warnings at all? I tried pulling it down and running it and didn't see any issues 🤔 

To be safe, we could totally move jscodeshift to deps and keep as it external. Curious what your preference would be

tay1orjones:
@joshblack Yeah as I look into it more I think it's safer to have a hard dep and mark it external.

joshblack (approved):

abbeyhrt (approved):

PR #11665, opened 2022-06-22
chore(CodeSnippet): update tests to RTL

### LINKED ISSUE #10272: Use RTL for testing in `CodeSnippet/tests/CodeSnippet-test.js`

jnm2377:
Closes #10272 

#### Testing / Reviewing

- ensure all tests make sense and are passing

Note: the last test for `Show more btn` kept failing, even though it should be working since the code content is well over 15 lines. I'm unsure what is going wrong. This was a test that was originally skipped as well, so it seems like it was an issue in past tests too. If anyone has any ideas, lmk :)


packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
```diff
@@ -6,102 +6,186 @@
...
-      wrapper.setProps({
-        children: '1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15\n16',
-      });
+  it.skip('should have show more button when more than 15 rows', () => {
+    render(<CodeSnippet type="multi">{multiLong}</CodeSnippet>);
 
-      expect(wrapper.find(Button).length).toBe(1);
-    });
+    expect(
+      document.querySelector('button.cds--snippet-btn--expand')
```
tay1orjones: Instead of the querySelector, could this (and the others in lines above) be `screen.getByTitle('Show more');` or `screen.getByText('Show more');`? Would it maybe fix the issue of the test not passing?

jnm2377: @tay1orjones So I tried using that for the first 2 tests but it fails bc `show more` isn't in the document.
Screen Shot 2022-06-22 at 2 53 06 PM


I also tried using that for the last test that I skipped, but even with that selector, it fails. It seems to not show up at all, even if I manually add `maxCollapsedNumberOfRows={15}`.

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
```diff
@@ -6,102 +6,188 @@
...
+    ).not.toBeInTheDocument();
+  });
 
-    it.skip('when more then 15 rows', () => {
-      wrapper.setProps({
-        children: '1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15\n16',
-      });
+  it.skip('should have show more button when more than 15 rows', () => {
```
joshblack:
```suggestion
  it('should have show more button when more than 15 rows', () => {
```

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
```diff
@@ -6,102 +6,188 @@
...
+      document.querySelector('button.cds--snippet-btn--expand')
+    ).not.toBeInTheDocument();
+  });
+
+  it('should not have show more button when exactly 15 rows', () => {
+    render(<CodeSnippet type="multi">{multi15}</CodeSnippet>);
+
+    expect(
+      document.querySelector('button.cds--snippet-btn--expand')
+    ).not.toBeInTheDocument();
```
joshblack: Would it be possible to query for this without `querySelector`? Maybe `screen.queryByText()` and use the text for "show more" and assert that it is null?

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
```diff
@@ -6,102 +6,188 @@
...
+  it('should allow hiding the copy button', () => {
+    render(
+      <CodeSnippet type="single" hideCopyButton>
+        {single}
+      </CodeSnippet>
+    );
 
-      expect(wrapper.find(Button).length).toBe(0);
-    });
+    expect(document.querySelector('button')).not.toBeInTheDocument();
```
joshblack, line 129: Would it make sense to use `screen.queryByText()` with the copy text and assert that it is null? This could help to avoid the `querySelector` call

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
```diff
@@ -6,102 +6,188 @@
...
+    render(
+      <CodeSnippet type="inline" data-testid="code-5" className="custom-class">
+        {inline}
+      </CodeSnippet>
+    );
 
-    beforeEach(() => {
-      wrapper = mount(<CodeSnippet type={'multi'} />);
-    });
+    expect(screen.getByTestId('code-5')).toHaveClass('custom-class');
```
joshblack: I think for `className` placement we should be explicit in heirarchy and use `container`:

```jsx
  it('should allow custom classes to be applied when passed in via className', () => {
    const { container } = render(
      <CodeSnippet type="inline" className="custom-class">
        {inline}
      </CodeSnippet>
    );

    expect(container.firstChild).toHaveClass('custom-class');
```

This can help us catch if `className` is moved to a different part of the component

joshblack (changes requested):

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
jnm2377: @joshblack I tried this but was running into errors bc it's not in the document at all, so it kept failing with that selector
https://github.com/carbon-design-system/carbon/pull/11665#discussion_r904173752

joshblack: @jnm2377 it seems like you were using the `get*` matcher, is that correct? I believe the difference between `get*` and `query*` is that `get` will throw if it cannot find it, but `query*` will return `null` if it cannot find it.

jnm2377: ahh thank you! fixed it 👍🏽

packages/react/src/components/CodeSnippet/__tests__/CodeSnippet-test.js
jnm2377: @joshblack @tay1orjones this test still fails even with the different query selector. I'm not sure why. I've tried manually setting the number of max collapsed rows, passing the actual code as a child, making the code longer. When it logs the component, the show more button is just not there at all, even though it should be.

Screen Shot 2022-06-22 at 5 11 19 PM

joshblack: @jnm2377 my guess is that the layout checks that `CodeSnippet.js` is making (with like `useResizeObserver` and other logic) are using different values than what is showing up in browsers since the component is being rendered in JSDOM.

I think what this means is that `useResizeObserver` is never called so `setShouldShowMoreLess` is never set to true.

This may be something that can only be checked in the browser / using VRT for the layout to be placed as intended so dropping this makes sense to me if that is already covered

tay1orjones (approved):

joshblack (approved):

PR #11670, opened 2022-06-23
fix(styles): add user-select: none to toggle

### LINKED ISSUE #10881: [Bug]: Clicking on Toggle component selects nearby text
### Package

carbon-components-svelte

### Browser

Chrome

### Package version

v10.47.0

### React version

_No response_

### Description

If you click quickly on a Toggle component, this will cause the browser to select text on the page. This happens when clicking on the graphical part of the toggle, not the label beside it.

The linked demo is written in React, but also applies to Svelte and presumably other libraries.

### CodeSandbox example

https://codesandbox.io/s/relaxed-drake-vdo7k8

### Steps to reproduce

Double-click or triple-click the graphical element of the Toggle component. 
### Code of Conduct

joshblack:
Closes #10881 

Add `user-select: none` to toggle to prevent text selection.

#### Changelog

**New**

**Changed**

- Update `.cds--toggle` select to set `user-select: none`

**Removed**

#### Testing / Reviewing

- Verify that text selection no longer occurs when quickly clicking the toggle
- Review the approach of setting `user-select: none` to see if this is too broad of a change and if this should be more specific


tay1orjones (approved):
Can confirm this fixes it. The only potential downside I see is that the label is no longer selectable/highlight-able


https://user-images.githubusercontent.com/3360588/175393394-f2981f69-c35d-4d18-926d-350f219d70a2.mov

I'm not aware of any a11y requirement or similar that would require the text be selectable, so I think it's a fine tradeoff? I also can't think of a strong use case for someone to need to select/highlight the label for copy/paste 🤔

abbeyhrt (approved):

PR #11686, opened 2022-06-27
feat(filterablemultiselect): add `onInputValueChange` prop

tay1orjones:
[Surfaced in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1656056168466059). This brings the change from #11415 to v10. #11415 included this change to the non-`next` component source, but because it was made in `main` it never was released.

#### Changelog

**New**

- Add `onInputValueChange` to FilterableMultiSelect from #11415


tay1orjones:
I found a cleaner way to do this by cherry picking the original commit - opened a new PR https://github.com/carbon-design-system/carbon/pull/11690

Closing this in favor of https://github.com/carbon-design-system/carbon/pull/11690

PR #11697, opened 2022-06-28
fix(TreeView): use prespecified active and selected on mount only

emyarod:
This PR updates the `onSelect` signature to provide information on the active node and updates the effects to set active and selected nodes using preselected values only on mount

#### Testing / Reviewing

The storybook demos should continue functioning as expected, to test these changes a code sandbox would be better to see event handler output when handling selection


tw15egan (approved):
LGTM 👍 ✅

abbeyhrt (approved):

PR #11698, opened 2022-06-28
fix(ProgressBar): improve screen reader output

### LINKED ISSUE #11420: [a11y]: Progress Bar accessibility challenges
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

Carbon on storybook https://react.carbondesignsystem.com/?path=/story/components-progressbar--indeterminate

### React version

https://react.carbondesignsystem.com/?path=/story/components-progressbar--indeterminate

n/a

### Assistive technology

JAWS

### Description

Both the indeterminate and determinate versions of the Progress Bar have challenges with a screen reader.

In react, the Indeterminate progress bar does not have good JAWS output.
image

First, it is not announced as a live region (at least in the implementation), which results in a failure of Status Messages.

If navigating by arrow key, for each of the 3 lines, it reads:
> Progress bar label
> 0 Progress bar partially checked
> Optional helper text

The first and last lines are fine, however "0 Progress bar partially checked" is not just unhelpful but is actually confusing. Since the indeterminate is never going to resolve to another number, the 0 is meaningless. "Partially checked" is not a state I would associate with  a progress bar. (Note that a [similar output message](https://github.com/nvaccess/nvda/issues/10644) is reported with NVDA. This issue suggests it's only occurring in Chrome, but although I got an improved message in FF "50 progress bar partially checked" it's the same basic issue.)
Reviewing the [ARIA specification](https://www.w3.org/TR/wai-aria-1.2/#progressbar), the issue may be that the div with the role of progressbar has not been given `aria-busy="true"`. With this set, the progress bar is entirely bypassed and JAWS reads only the label and optional text, which is arguably a better experience.
When the loading completes, the aria-busy should be set to "false". I did not have success using any of the aria-value attribute variants (max, now, text), so the best bet may be to just convey the results via the optional helper text.

---

The [progress bar example](https://react.carbondesignsystem.com/?path=/story/components-progressbar--example) also has challenges with meaningful status info.
image

The output is variable, but something like this (after a refresh):

> Export data
> 0 Progress bar partially checked
> Fetching assets...
> 28.4MB
> 490.588

Basically, the load value (in MBs) is announced once or (as above) twice, with the same odd 'partially checked' info. Here, despite the text update, there is never an announcement of "Done". 
image

The code is properly using aria-value, but the output is not very successful.
`<div class="cds--progress-bar__track" role="progressbar" aria-labelledby="progress-bar-18" aria-describedby="progress-bar-helper-19" aria-valuemin="0" aria-valuemax="728" aria-valuenow="728"><div class="cds--progress-bar__bar" style="transform: scaleX(1);"></div></div>`

The 2 key things that need to be conveyed to a screen reader are 1) that the download (or whatever) is taking progress and 2) that it is done. As well, interim information on progress is welcome, if not too verbose.

It's possible that putting an `aria-live="polite"` on the following div (that contains the content of the optional text) would resolve this. Although this text is optional, my sense is that it is a rare progress bar that wouldn't expose information.

`<div id="progress-bar-helper-27" class="cds--progress-bar__helper-text">Optional helper text</div>`

### WCAG 2.1 Violation

https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html

### Reproduction/example

See storybook

### Steps to reproduce

1. Go to the storybook implementation examples
2. Launch JAWS
3. Listen to output

### Code of Conduct

dakahn:
Closes #11420 

JAWS was reading every helper text update (which numbered in the hundreds) and had a super verbose output. This PR hides the hleperText from JAWS and instead has a visually hidden live region that describes the ProgreesBar as 'loading' or 'done' depending on the state of `isFinished`. 

#### Testing / Reviewing

With a screen reader reload the page. The progress bar should describe itself and then tell the user it's loading. when it's finished it should then update the user by saying 'done'

#### Notes

@mbgower let me know if this is what you were shooting for in that issue ticket. I'm hiding the progress bar via `aria-busy` like you suggested, but its worth noting that in my research it seems like that attribute might be intended for a loading _region_ of a page--which i guess this progressbar could count as.


mbgower:
Thanks, @dakahn I'll have a look.
I do see aria-busy cited by mozilla
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role

It sounds like maybe you were pushing valuenow a lot.  I haven't looked behind the scenes to see the cause, but in my experience screen readers usually announce a few percentage points when loading, which tends to be adequate for most situations.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow

jnm2377:
@dakahn Is this what I should be seeing? I may be testing it wrong.

https://user-images.githubusercontent.com/32556167/176789044-565581a2-772f-41d2-9ca3-7b624ef3a860.mov

dakahn:
@jnm2377 Pretty much! make sure you're testing VO in Safari though as it can get a little whacky in other browsers

jnm2377 (approved):

joshblack (approved):

PR #11723, opened 2022-06-30
style(styles): update border values for tile

### LINKED ISSUE #11351: [a11y]: Contrast on tiles/cards is below 3:1
### Package

carbon-components

### Browser

Chrome

### Operating System

MacOS

### Package version

Take your pick :)

### React version

take your pick :)

n/a

### Assistive technology

n/a

### Description

As discussed on some recent cards, there are a variety of implementations of cards in IBM:
- Some are actionable
- some are sortable
- some are selectable
- some are just holding static text
- some are themselves non-actionable but contain a link or other actionable item
- some have multiple actionable items

Especially when we get into Carbon flavours, similarly functioning cards share very few attributes, in regard to placement, stroke width, stroke strength, spacing, etc., etc.
I think overall there are challenges for all users trying to discriminate between actionable cards/tiles and non-actionable content holders. But for users with disabilities, the problems of discernment and interaction become more pronounced.

One clear way to improve this situation is to undertake a clearer division between when a card is actionable, and when it is just a visual container for other content. There is a requirement called **Non-text Contrast** that I think can help with this.
### Visual information required to identify user interface components and states
[1.4.11 Non-text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) requires that controls (AKA [user interface components](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html#dfn-user-interface-component)) have a 3:1 contrast against adjacent colors for visual information **required to identify** them. So, if a card is actionable -- a control -- it should have something like a stroke around it that contrasts 3:1 against the page background so that users can understand it is actionable. There are other ways besides strokes to do this, for instance the entire card could be 3:1 against the background.

And of course it follows that although there is no requirement saying this, a tile/card that is NOT actionable shouldn't have a 3:1 contrast. That way everyone would be able to better tell when there is an intended interaction on the page, and when it is just 'decorative'.

This isn't a magic bullet to solve all the challenges making consistent and beautiful cards, but I recommend this basic rule of 3:1 contrast as a way of better defining intended interaction, while improving the accessibility of components for everyone :)

### WCAG 2.1 Violation

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-tile--default

### Steps to reproduce

1. Look at just about any card/tile in Carbon, including https://react.carbondesignsystem.com/?path=/story/components-tile--default
2. Use a contrast checker to check the card for 3:1 contrast against adjacent colors.

### Code of Conduct

abbeyhrt:
Closes #11351 

Adds borders to Selectable/Clickable/Expandable(no interactive) tiles to increase visual contrast. 

The borders should match this spec: 
image (3)

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Verify that the new borders match the above specs 
 Clickable/Selectable/Expandable(non-interactive) stories

### DIFF
--- a/packages/react/src/components/Tile/next/Tile.stories.js
+++ b/packages/react/src/components/Tile/next/Tile.stories.js
@@ -150,7 +150,10 @@ export const Radio = () => {
       <RadioTile value="standard" style={{ marginBottom: '.5rem' }}>
         Option 1
       </RadioTile>
-      <RadioTile value="default-selected" id="tile-2">
+      <RadioTile
+        value="default-selected"
+        style={{ marginBottom: '.5rem' }}
+        id="tile-2">
         Option 2
       </RadioTile>
       <RadioTile value="selected" id="tile-3">
--- a/packages/styles/scss/components/tile/_tile.scss
+++ b/packages/styles/scss/components/tile/_tile.scss
@@ -48,12 +48,11 @@
 
   .#{$prefix}--tile--clickable,
   .#{$prefix}--tile--selectable {
-    border: 1px solid $border-subtle;
+    border: 1px solid $border-strong;
     cursor: pointer;
     transition: $duration-moderate-01 motion(standard, productive);
 
     &:hover {
-      border: 1px solid $border-subtle-selected;
       background: $layer-hover;
     }
   }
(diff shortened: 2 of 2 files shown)


aledavila (approved):

tay1orjones (approved):

thyhmdo:
It's actually still a WIP, I will update this once it's done.

PR #11724, opened 2022-06-30
feat(TreeView): implement keyboard shortcut modal

emyarod:
Refs https://github.com/carbon-design-system/carbon/issues/6792

draft PR to collect design feedback

notification
- can app developers opt out of showing the notification?
- is the notification persistent or should it have a timeout?
  - if the notification is dismissible, what conditions are required to bring the notification back (if any)?
- notification location/positioning options?
- finalize notification content

modal
- finalize appearance

shortcut list
- finalize appearance/formatting

### DIFF
--- a/packages/react/src/components/TreeView/TreeView.js
+++ b/packages/react/src/components/TreeView/TreeView.js
@@ -8,10 +8,15 @@
 import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { Keyboard } from '@carbon/icons-react';
+import * as FeatureFlags from '@carbon/feature-flags';
 import { keys, match, matches } from '../../internal/keyboard';
 import uniqueId from '../../tools/uniqueId';
-import * as FeatureFlags from '@carbon/feature-flags';
 import { usePrefix } from '../../internal/usePrefix';
+import { Tooltip } from '../Tooltip/next';
+
+// TODO: OS dependent
+const defaultMultiselectHelpText = `Use Cmd + Enter to make a selection`;
 
 export default function TreeView({
   active: prespecifiedActive,
@@ -20,6 +25,7 @@ export default function TreeView({
   hideLabel = false,
   label,
   multiselect,
+  multiselectHelpText = defaultMultiselectHelpText,
   onSelect,
   selected: preselected = [],
   size = FeatureFlags.enabled('enable-v11-release') ? 'sm' : 'default',
@@ -197,12 +203,22 @@ export default function TreeView({
   useActiveAndSelectedOnMount();
 
   const labelId = `${treeId}__label`;
-  const TreeLabel = () =>
-    !hideLabel && (
-      <label id={labelId} className={`${prefix}--label`}>
-        {label}
-      </label>
-    );
+  const TreeLabel = () => (
+    <div className={`${prefix}--tree__label`}>
+      {!hideLabel && (
+        <label id={labelId} className={`${prefix}--label`}>
+          {label}
+        </label>
+      )}
+      {multiselect && (
+        <Tooltip align="bottom" description={multiselectHelpText}>
+          <button className="sb-tooltip-trigger" type="button">
+            <Keyboard />
+          </button>
+        </Tooltip>
+      )}
+    </div>
+  );
 
   return (
     <>
--- a/packages/styles/scss/components/treeview/_treeview.scss
+++ b/packages/styles/scss/components/treeview/_treeview.scss
@@ -22,6 +22,13 @@
     overflow: hidden;
   }
 
+  .#{$prefix}--tree__label {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    max-width: 16rem;
+  }
+
   .#{$prefix}--tree-node {
     padding-left: $spacing-05;
     background-color: $layer-01;


jeanservaas:
Not sure if it would be easier to have another review with @oliviaflory and @mbgower or if we just try to hash this out here.
I think michael will have a more informed answer to some of your questions. After reading these questions, I tried to do a little bit of quick visual research around this pattern and I'm coming up empty handed. Below I mentioned a few images Michael showed on our last call... it would probably help to post those and any relevant links here... this might require a bit more visual research to get Andrew over the finish line.

> **can app developers opt out of showing the notification?**

I would defer to Michael on this, but wouldn't this kind of be like opting out of an accessibility best practice? Seems like they shouldn't be able to opt out.

> **is the notification persistent or should it have a timeout?**
if the notification is dismissible, what conditions are required to bring the notification back (if any)?

This is a good question. Gut instinct, it seems intuitive that a user would be able to close the notification (rather than having it time out) — but then what do users do if they want a notification to return... would they refresh the page? Or would we need to have some kind of persistent collapsed state of the notification (which also might be a little weird)? I'm not sure what best practice would be on this. 

> notification location/positioning options?

So it looks like you've just defaulted it to the upper right hand corner, again, in a normal situation, i.e. in a web app with a notification center, this would make sense... but this is a very specific situation where we're telling the user a key command to operate a specific component on the page... So it seems like proximity to that component would be important. I know Michael had some visual examples that he showed us on the previous call @mbgower I'm wondering if you can post some of those visual examples here because I'm having trouble finding anything similar.

> finalize notification content

I guess I wasn't necessarily thinking that the notification had to trigger a modal... I was thinking that the notification would just tell you the specific key command that you needed to multi-select with the tree component... i.e. in this case, I'm not sure a modal would be required. What's odd in the example is that, you're giving the user a notification with a key command to trigger a modal that then displays key commands... 

- If you had enough key commands on a page to warrant a dialog, I think the notification would say something like (SUPER stupid example) "View the key commands you need to navigate this experience"

- But if the purpose of the notification is JUST to tell people the key command they'd need to multi-select with tree, I don't think you'd need the dialog... you'd just say something like "To select multiple items in this tree component, Command click..." or whatever

> Modal / shortcut list formatting

First let's see if we need a modal... If we did do a key command modal something really straight like this would be nice

image

mbgower:
> I'm wondering if you can post some of those visual examples here because I'm having trouble finding anything similar.

On this [page](https://w3c.github.io/aria-practices/examples/grid/dataGrids.html), the visual cue for arrow navigation doesn't appear until a user tries hitting an arrow key, which I wouldn't regard as a really great implementation (i also don't find the position BELOW the table instead of beside it a great feature), but this is one of the ones I just pointed out as an example of surfacing keyboard cues in the gutter.
image

mbgower:
> can app developers opt out of showing the notification?

As @jeanservaas says, if you allow folks to opt out of an accessibility, we seem to be setting ourselves up for other failures. This is like the example of letting folks remove underlines from links. Suddenly you need to create a bunch of secondary designs to prevent authors from failing when they decide to ditch the establish (accessible) convention.

If the keyboard assistance is done properly, I don't see it as something that would fall into the 'distraction' category (i.e., making _another_ problem while trying to solve this).

mbgower:
> the challenge with non-persistent information is that it becomes a timed event, which can itself be a challenge. What if someone with magnification takes a while to happen to see that guidance in their viewport, and by that point it's gone?

I don't think this was intended to be interactive, so making it dismissible raises the question 'how does a keyboard user (the only people who will actually SEE this) dismiss it?'

mbgower:
> notification location/positioning options?

I agree with Jean that proximity is pretty key here. I like the gutter beside the focus, or the top or bottom of the tree.

oliviaflory:
@emyarod a few alignment things I noticed while looking at this PR:

### 1. Caret alignment
The caret looks like it might have some extra bottom padding, or something that makes it sit higher in the container in comparison with the folder icon. Both caret and folder icon should be vertically centered with the text label, see last image with spec

Screen Shot 2022-07-05 at 2 29 28 PM
Screen Shot 2022-07-05 at 2 29 49 PM
Screen Shot 2022-07-05 at 2 43 26 PM


### 2. Text wrap
In the With icon version, we see text wrap happening under the "Business automation" folder. This causes the icon to not be aligned with the text. 
Screen Shot 2022-07-05 at 2 31 33 PM


However, in the original spec file I found the example below that shows the tree view should utilize `...` to maintain the height of a single item, and not wrap the text. I think this is the desired design approach. cc @mbgower I want to double check with you that there is no a11y issue with truncation of the label utilizing ellipse? 

Screen Shot 2022-07-05 at 2 34 57 PM

jeanservaas:
@emyarod @mbgower 

There is A LOT of good discussion here but in order to unblock/ship tree view... I'm wondering if we can:

1) focus solely on exposing the hot key to operate the multi-select modal, rather than triggering a modal with a list of command keys to navigate through the tree

2) Open an issue to carry over some of this discovery on a larger persistent global pattern for a keyboard interaction dialog (maybe in the footer as Michael suggested)

If we can agree on that... we'd like to explore Andrew's idea of the help (question mark) icon next to the tree's label, that triggers a tooltip describing the hot key interaction for multi-select). 

I know this is not a dynamic solution, so it has some limitations when the user scrolls — but it's a very simple one that we can proceed with and test. That way we can get tree out there... 

Opinions on this?

Issue #11259, opened 2022-04-20
[a11y]: loading and inline loading not consistently providing status programmatically


mbgower:
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-inlineloading--inline-loading

### Automated testing tool and ruleset

n/a

### Assistive technology

JAWS, keyboard

### Description

There are a number of considerations around loading patterns in Carbon that need to be addressed, ranging from clear accessibility bugs to inconsistencies. Several are listed here; I will add a few more after reviewing with CAG.

### Default inline loading
The [default implementation of inline loading](https://react.carbondesignsystem.com/?path=/story/components-inlineloading--inline-loading) on storybook uses aria-live to surface status (good!); however, the implementation is sometimes causing overly verbose output from the JAWS screen reader (not good). Note that any changes to the text can get read 3 times! (Haven't been able to consistently replicate this problem.) This may be partly a result of the storybook container; testing needs to be carried out on an actual implementation
![image](https://user-images.githubusercontent.com/14298245/164316694-a6609b30-9cc0-4acd-a2d6-5d8edf7f02ff.png)

   1. part of this may be due to aria-live being set to assertive (instead of polite) OR
   3. by aria-live assigned to multiple elements (the div and its child), or
   4. by having the child attribute assigned both a live region and what I suspect is an unnecessary label ("Active loading indicator") or
   5. some combination of these

I recommend considering a simple label for the symbol (for instance "loading") that can provide sufficient information whether the "Loading data..." text is also displayed or not.

### UX example inline loading
The second working example "Ux Example" is an interaction comprising a set of Cancel and Submit buttons. 
![image.png](https://zenhub.ibm.com/images/5cab73f442b8dc5ede7b84cc/1b816b0c-326a-4f38-a137-57e32a3639a6)
When you activate Submit, the text changes to "Submitting..." and then "Submitted" before returning to "Submit"
[...]

PR #11759, opened 2022-07-06
fix(Loading): improve screenreader support

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/11259

Work to improve screen reader support with `Loading` and `InlineLoading`

#### Changelog

**Changed**
- Reduced default descriptive text to "Loading"
- With `InlineLoading`, if an `iconDescription` is not provided, it will use the status as the `title` text. 
- If text is provided to `InlineLoading` via `description`, this will be read by the screenreader. If no text is provided, the `title` text is picked up (either the `iconDescription` if provided, or the `status`)

**Removed**
- The `label` element has been removed, as it was adding to the noise for screenreaders.
- `aria-labelledby` has been removed, since the `label` it was referring to no longer exists
- The `id` prop is deprecated as it is no longer used.

#### Testing / Reviewing

Open up the stories for `Loading` and `InlineLoading`. 

- Use the Accessibility checker and ensure there are no violations. 
-  Try to remove/change the `iconDescription` and `description` props for `InlineLoading` and test the changes with the MacOS VoiceOver tool.
- Try to remove/change the `description` prop for `Loading` and test the changes with the MacOS VoiceOver tool.


joshblack (approved):

mbgower (approved):
I didn't build and test with screen reader -- assuming someone will do so -- but a quick review of the code looks right. Thanks!

dakahn:
Tested with JAWS latest on Edge and Firefox on Windows 11. Works as expected 👍🏾

dakahn (approved):

mbgower:
Note, one thing I'll mention if this fix is considered to resolve #11259 is that there are other considerations captured in the original issue that are not addressed by this technical update, and probably need to be assessed by @laurenmrice or another designer:
>### Why are inline loading and loading treated as separate components?
>Inline loading is a separate component from Loading on the Carbon Design System website and storybook; however in Sketch and Figma, inline loading is a variant on Loading. I think the latter makes sense, overall, and I suggest the two components be combined. If desired to keep separate, then they should be separated in the Design kit, IMO.
>### Sketch vs Figma vs implementation
>Note that there are some differences in how Figma and Sketch show the inline loading that should be addressed. Sketch includes inline loading on text inputs; Figma does not. Figma includes a icon-less version (which I think is the 'inactive' one); Sketch does not. Neither show the above implemented variant of a button that updates its text. I consider that almost a different variant.

PR #11767, opened 2022-07-07
feat(themes): add unstable_metadata export

### LINKED ISSUE #11748: Export token metadata from `@carbon/themes`
**Background**

There are several use-cases for publishing metadata for themes and tokens, including:

- Generating pages on the website instead of manually authoring sections
- Authoring tools (like stylelint plugins) to help with usage

At the moment, token metadata is internal to the `@carbon/themes` package. It would be helpful to make this metadata available in the package itself (potentially first under `unstable`) to be used for these use-cases.

If possible, it would be helpful for tools to also access v10 token information as a part of these changes.

**Checklist**

- [ ] Export token metadata as a part of `@carbon/themes`
- [ ] Include component tokens in the output
  - [ ] Button
  - [ ] Notification
  - [ ] Tag

joshblack:
Closes #11748

Adds in an `unstable_metadata` export which contains metadata around tokens in v11 and v10. This change includes bringing in component token groups in v11. The structure is a list of tokens with a `name` field for the token name along with a `type` which corresponds to the type of token (color, type, layout)

#### Changelog

**New**

- Add `unstable_metadata` to `@carbon/themes`

**Changed**

**Removed**


packages/themes/src/tokens/index.js
```diff
@@ -11,5 +11,50 @@ import { TokenGroup } from './TokenGroup';
...
 import { group } from './v11TokenGroup';
 import { set } from './v11TokenSet';
+import * as components from './components';
+import { type } from './type';
+import { layout } from './layout';
+import { v10 } from './v10';
 
 export { Token, TokenFormat, TokenGroup, TokenSet, group, set };
+
+const v11 = [
```
lee-chase, line 21: @josh is this theme and color token? I need to be able to distinguish between theme and the color palette colors.

joshblack: @lee-chase in this case it will be the theme tokens versus anything from the IDL colors. I think colors can be pulled from `@carbon/colors` with the `colors` export IIRC (let me know if I can help out at all with that!)

lee-chase: Happy to take from @carbon/colors which is where I currently retrieve the palette from. Otherwise looks good.

tay1orjones (approved):

abbeyhrt (approved):

PR #11777, opened 2022-07-11
fix(FileUploader): add filename to screenreader output, return focus to button after deletion

### LINKED ISSUE #11198: [a11y]: File uploader file delete button needs more context
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

n/a

### React version

_No response_

inspect

### Assistive technology

_No response_

### Description

As discussed at CAG, files uploaded through the File Uploader component can be removed, but the file's name is not part of the delete mechanism.
When multiple files are uploaded, all the delete mechanisms ('x') are just called "Delete" (or sometimes "clear').

Since there is no confirmation prompt before deleting, it is especially important that a user who cannot see has an ability to know which file is being deleted. 
image

image

There are several ways of resolving, this, including constructing each uploaded file as an object. However, the simplest thing to do is:

1. Consistently change the default aria-label to "delete" instead of 'clear'.
2. provide the file name through aria-describedby

image

`<span class="cds--file__selected-file cds--file__selected-file--md"><p id="file1" class="cds--file-filename">button_usage_25.png</p><span class="cds--file__state-container"><button aria-label="Delete file" class="cds--file-close" aria-describedby="file1" type="button" tabindex="0"><svg focusable="false" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" fill="currentColor" width="16" height="16" viewBox="0 0 32 32" aria-hidden="true"><path d="M24 9.4L22.6 8 16 14.6 9.4 8 8 9.4 14.6 16 8 22.6 9.4 24 16 17.4 22.6 24 24 22.6 17.4 16 24 9.4z"></path></svg></button></span></span>`

### WCAG 2.1 Violation

1.3.1 Info and relationships

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader

### Steps to reproduce

n/a

### Code of Conduct

### LINKED ISSUE #11204: [a11y]: File uploader focus handling after file delete
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application

### React version

_No response_

n/a

### Assistive technology

JAWS and keyboard

### Description

After a file is deleted from the upload list, the focus does not go anywhere. (I'll mention that the delete action does not have a confirmation, but set that aside as a design decision.) 
https://ibm.box.com/s/scxudnpq13sqvnv0jbqxamlcyp31r8kf

Prior CAG discussions on focus handling are relevant to this problem. Because the item with focus is being deleted, where should the focus go?
In prior discussions, the thought was to go the former item in the tab order. Until the delete buttons are distinguishable programmatically (#11198) it definitely makes more sense to put it on the File add button. And even once that is resolved, I think it may be a more consistent experience to always go to the File upload button regardless of which file has been delete 

We'd previously discussed the notion of a group of controls. Perhaps an interactive parent should be part of the algorithm. This same algorithm could work for table row deletions, menu item deletions and the file uploaded group of files situation.

### WCAG 2.1 Violation

https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application

### Steps to reproduce

1. Go to https://react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application
2. Upload a few files
3. Delete a file
  - the focus will not be anywhere visually

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/11198
Closes https://github.com/carbon-design-system/carbon/issues/11204

Changes all instances of `Clear file` to `Delete file`, and adds the filename as the `id`  and `aria-describedby` to improve screenreader support

#### Changelog

**New**

- `aria-describedby` with the `filename` passed down to `Filename` component
- After an item is deleted, focus now returns to the `FileUploader` button

**Changed**

-  "Clear file" now reads as "Delete file"
  - Updated `FileUploaderButton` and `FileUploaderDropContainer` to accept an `innerRef`, which is used to set focus after a file is deleted

#### Testing / Reviewing

FileUploader --> Drag and Drop Example --> Upload a file. After the file is uploaded, turn on the MacOS screenreader and tab to the close button. It should read "Delete file, button" followed by the filename.

After the file is deleted, the focus should return to the Button / Drop Container


tay1orjones (approved):
Is there anything to do about this gray focus ring I'm seeing after deleting the item? It's only present when voiceover is on. Moving focus around and/or clicking doesn't seem to get it to go away. It eventually does go away though. Maybe it's a feature of Voiceover I'm not familiar with?


https://user-images.githubusercontent.com/3360588/178495310-27c1c676-e8bb-4b79-974e-b1dd1c8d4964.mov

tw15egan:
@tay1orjones I noticed that as well, at least the voiceover reads that the focus is back on the file uploader button though. I was able to move the focus with voiceover controls (<kbd>Control</kbd> <kbd>Option</kbd> <kbd>Shift</kbd> <kbd>⬆️</kbd>), but thats it

mbgower:
yeah, I don't know what's going on there. The VO focus should have moved to the item with focus. May be a VO bug. If something similar doesn't happen with JAWS, I'd just notch it up to an Apple curiosity. I don't think you need to remediate.

dakahn (approved):
Tested with JAWS 2022 on Windows 11 in Firefox. Read really nicely! 🏄🏾

PR #11786, opened 2022-07-12
feat(HelperClasses): add breakpoint helper classes

### LINKED ISSUE #2435: [Grid] Add visibility classes
From @DusanMilko:

I would like to request carbon team to add visibility classes for breakpoints so you could do:
bx--col-xs-1 bx--col-lg-hidden <- this hides for lg and up, could also add bx--col-lg-only-hidden to hide for that breakpoint only. I can make a pr if you would prefer, let me know

Like foundation has done:
https://foundation.zurb.com/sites/docs/visibility.html

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/2435

This PR adds in classes that can be used to hide elements _only_ at specific breakpoints. This can already be done with the 0 column span grid classes, but this specifically addresses the need for this _outside_ of a grid. 

>When the element is not in a grid, I wouldn't want to use grid classes. 

This is a PoC for a potential "Helper Classes" file that lives inside the `utilities` section of our `scss` package. We've gotten other requests for simple helper classes in the past, and this would give them a place to live. Would love to know what everyone thinks.

None of the classnames/filenames are set in stone and would love feedback.

For now, you can view the example in the storybook under the `Helpers` section at the bottom of the list. 

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


tay1orjones (commented):
I think it makes sense to have a place like this for helper classes to live. A few thoughts:

* Does it make sense to document these in an elements example instead of storybook?
* Do we need to expose these via React Components? (Assuming no)
* Should these be within a mixin (or maybe a mixin per-class?) so they're easier to compose within existing classes a product might have?
```scss
// use a single helper via mixins?
@use '@carbon/styles/scss/utilities/helper-classes';

.my-class {
  @include helper-classes.hide-at-sm();
}
```
```scss
// emit "hidden" helper classes and make them all available to this module via one big mixin?
@use '@carbon/styles/scss/utilities/helper-classes';

@include helper-classes.hidden();
```

tw15egan:
> Does it make sense to document these in an elements example instead of storybook?
I don't have a preference here, but since it lives in the style package I just put it in the storybook. 

> Do we need to expose these via React Components? (Assuming no)
I don't think so

> Should these be within a mixin (or maybe a mixin per -class?) so they're easier to compose within existing classes a product might have?

Good thinking, I think making them available as mixins fits better with the way we ship other helper utilities. I prefer the mixin per class approach, I can push up a PR with those changes so we can take a look

packages/styles/scss/utilities/_index.scss
```diff
@@ -10,6 +10,7 @@
 @forward 'component-reset' as component-*;
 @forward 'convert' show em, rem;
 @forward 'focus-outline';
+@forward 'hide-at-breakpoint';
```
tay1orjones, line 13: IIRC we also need to be sure this file is included in the build via `packages/styles/files.js` 

https://github.com/carbon-design-system/carbon/blob/775ab4c132ee42e4ba39b2fbee1d0ce13475a1b7/packages/styles/files.js#L180-L197

tay1orjones (changes requested):
One small thing, otherwise looks good to me!

packages/styles/scss/utilities/_index.scss
tw15egan: Good catch 😅

tay1orjones (approved):

aledavila (approved):

PR #11802, opened 2022-07-14
chore(DatePicker): update tests to RTL

### LINKED ISSUE #11512: Use RTL for testing in `DatePicker/DatePicker-test.js`

jnm2377:
Closes #11512 

#### Changelog

- updates tests to RTL

#### Testing / Reviewing

- ensure all tests pass


packages/react/src/components/DatePicker/DatePicker-test.js
```diff
@@ -7,228 +7,182 @@
...
-          <DatePickerInput
-            id="date-picker-input-id-finish"
-            placeholder="mm/dd/yyyy"
-            labelText="End date"
-          />
-        </DatePicker>
-      );
+  // eslint-disable-next-line
+  it('should add the date format as expected', () => {
+    //how to test for date format? it's used for a flatpickr config but not a dom attribute or anything...
```
tay1orjones: I think the material impact `dateFormat` has is just changing the format of the date  string put in the input after you select a date from the dropdown.

One way to do this would be

1. Choose a date from the datepicker (literally any, the first in the dropdown would work)
2. Get the value of the input
3. Split value on the slashes `const dateValues = value.split("/");`
4. Assert dateValues[0] is a number between 1-12 (month)
5. Assert dateValues[1] is a number between 1-31 (day)
6. Assert dateValues[2] is a number >=2022 (year)

It's not perfect but it's a decent baseline.

packages/react/src/components/DatePicker/DatePicker-test.js
```diff
@@ -7,228 +7,182 @@
...
-            id="date-picker-input-id-finish"
-            placeholder="mm/dd/yyyy"
-            labelText="End date"
-          />
-        </DatePicker>
-      );
-    });
+  // eslint-disable-next-line
+  it('has the value as expected', () => {
+    //how to test for date value? it's used for a flatpickr config but not a dom attribute or anything...
```
tay1orjones: Maybe mock downshift's [`onValueUpdate`](https://flatpickr.js.org/events/#onvalueupdate)? Could assert that it's called when the value changes after a new selection in the dropdown, or a new string is put in the input.

Otherwise this is probably covered by other tests that cover the base functionality of selecting a date and the date string changes in the input.

packages/react/src/components/DatePicker/DatePicker-test.js
```diff
@@ -271,3 +225,95 @@ describe('DatePicker', () => {
...
+        />
+        <DatePickerInput
+          id="date-picker-input-id-finish"
+          placeholder="mm/dd/yyyy"
+          labelText="End date"
+        />
+      </DatePicker>
+    );
+
+    //how to test for date locale? it's used for a flatpickr config but not a dom attribute or anything...
```
tay1orjones: I think you might be able to set a date of 01/01/2022, open the datepicker and assert that "Enero" is in the dom. The full list of what's provided for `es` is [here](https://github.com/flatpickr/flatpickr/blob/master/src/l10n/es.ts)

packages/react/src/components/DatePicker/DatePicker-test.js
```diff
@@ -271,3 +225,95 @@ describe('DatePicker', () => {
...
+          labelText="Start date"
+        />
+        <DatePickerInput
+          id="date-picker-input-id-finish"
+          placeholder="mm/dd/yyyy"
+          labelText="End date"
+        />
+      </DatePicker>
+    );
+    //how to test for date min and max? it's used for a flatpickr config but not a dom attribute or anything...
```
tay1orjones: Might be best to render the datepicker with an initial value that is equal to the minimum date so then if you open the dropdown you already have the minimum date in the dom. Try to select a date before the minimum date, assert it wasn't selected. Dates outside the min/max should get a disabled class or something IIRC. You could assert on that too.

tay1orjones (commented):
I came up with a few ideas for the ones you had comments on. Let me know if you wanna walk through them!

tay1orjones (approved):
Looks great!

aledavila (approved):

PR #11850, opened 2022-07-24
chore(project): upgrade prettier to use cache option

falsepopsky:
**Summary**

In the version 2.7 prettier introduced a new feature in the current CLI [--cache](https://prettier.io/blog/2022/06/14/2.7.0.html) / [docs](https://prettier.io/docs/en/cli.html#--cache) . With this upgrade, you can save a lot of time:

#### Small reproduction

No Cache
without cache

Cache  option enabled

cache

**New**

- Upgrade prettier

**Changed**

- added `--cache` flag in `package.json` file in the following scripts: `--format` and `format:diff`

**Removed**

- nothing


falsepopsky:
If this upgrade it's okay, some scripts that use prettier should be modified, like `lint-staged`

abbeyhrt (approved):

abbeyhrt:
@falsepopsky Thanks for doing this! I don't foresee running into any problems with this and we'll look into adding it to the other scripts that should be modified!

aledavila (approved):

PR #11893, opened 2022-08-01
fix(react): IconButton prop types for danger kind

### LINKED ISSUE #11892: [Bug]: IconButton with danger kind
### Package

@carbon/react

### Browser

Chrome

### Package version

1.8.0

### React version

18.2.0

### Description

The following code:
```
...
<Button
    hasIconOnly
    className={canDelete === false ? "hidden" : ""}
    disabled={canDelete === false}
    iconDescription={txt.DELETE}
    kind="danger--ghost"
    onClick={() => onDelete(index)}
    renderIcon={DeleteIcon}
    size="sm"
/>
...
```
Outputs the following warning in the console:
```
Warning: Failed prop type: Invalid prop `kind` of value `danger--ghost` supplied to `ForwardRef(IconButton2)`,
expected one of ["primary","secondary","ghost","tertiary"].
```
Displays and functions as intended otherwise. 

### Reproduction/example

https://stackblitz.com/edit/github-mddqrk?file=src%2FApp.jsx

### Steps to reproduce

Using Chrome, click the link to the example and open the developer tools console.
You should see the following warning:
```
Warning: Failed prop type: Invalid prop `kind` of value `danger--ghost` supplied to `ForwardRef(IconButton2)`,
expected one of ["primary","secondary","ghost","tertiary"].
```
Displays and functions as intended otherwise.

### Code of Conduct

stanley-regnard:
Closes https://github.com/carbon-design-system/carbon/issues/11892

Use `ButtonKinds` as possible values for `kind` property of IconButton component (just like how Button does), instead of the hardcoded string values that do not include any of the `danger` kind.

#### Changelog

**Removed**

- False alarm when using IconButton as a `danger` button

#### Testing / Reviewing

Using an IconButton with `kind` set to any of those values `['danger',  'danger--primary',  'danger--ghost',  'danger--tertiary']` should no longer trigger a warning in the console

### DIFF
--- a/packages/react/src/components/IconButton/index.js
+++ b/packages/react/src/components/IconButton/index.js
@@ -8,6 +8,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import Button from '../Button';
+import { ButtonKinds } from '../../prop-types/types';
 import { Tooltip } from '../Tooltip/next';
 import { usePrefix } from '../../internal/usePrefix';
 import cx from 'classnames';
@@ -85,7 +86,7 @@ IconButton.propTypes = {
   /**
    * Specify the type of button to be used as the base for the IconButton
    */
-  kind: PropTypes.oneOf(['primary', 'secondary', 'ghost', 'tertiary']),
+  kind: PropTypes.oneOf(ButtonKinds),
 
   /**
    * Provide the label to be rendered inside of the Tooltip. The label will use


stanley-regnard:

stanley-regnard:
recheck

tay1orjones (changes requested):
Thanks for submitting this! Unfortunately danger-ghost is not a valid option per design's guidance https://github.com/carbon-design-system/carbon/pull/10095#pullrequestreview-807656822

> **For `kind`** The following buttons do not come as icon only buttons and should be removed from the dropdown:
> 
> * [x]  danger
> * [x]  danger--primary
> * [x]  danger--ghost
> * [x]  danger--tertiary

stanley-regnard:
Thanks for the quick answer. So, by design, icon only buttons should not be usable as danger buttons? That's a curious choice. 
What would be the suggested alternative then?

abbeyhrt:
@stanley-regnard I think the idea is that a "danger"-ous action should be called out more prominently than through an icon. So in your case of the trash can icon, ideally that would become a "Delete" button so that we can feel more confident that the user knows they are about to perform a dangerous action

stanley-regnard:
Thank you for the clarification!

Closing this and the related issue since it's working as intended then.

sbuys:
I know this is closed but this seems like an arbitrary design decision.

Signaling danger via symbol, color, etc. is pretty ubiquitous in the physical world (see traffic signs, hazardous chemicals warning, etc.). Text does not always make signal processing clearer....

craigyu:
Petition to reopen this PR

PR #11931, opened 2022-08-08
docs: update styles and react indexed descriptions

### LINKED ISSUE #1054: Remove unwanted code pages in Carbon Styles indexing
**Description**

We previously determined with @carbon-design-system/carbon-platform-designers that if an asset's code page just links to other frameworks and has a live demo, then we don't need to include that as a page tab because both of those are now surfaced in the overview page.

For example, I don't think we want a code page tab for Accordion: https://carbondesignsystem.com/components/accordion/code.

We should look through where we specify `codePath`s here and remove `codePath`s where we don't want code tabs.

https://github.com/carbon-design-system/carbon/blob/main/packages/styles/carbon.yml

mattrosno:
Closes https://github.com/carbon-design-system/carbon-platform/issues/1054

References https://github.com/carbon-design-system/carbon-platform/issues/1109

This PR removes indexed component code documentation where code tabs are no longer needed. It also cleans up the Carbon config file formatting and order, as well as removes "unstable" from some `id`s and SVG names.

#### Changelog

**New**

- `UI shell left panel` and `UI shell right panel` Carbon React assets, so the three UI shell components in that library can get the code tabs that have Codesandbox embeds of the React components. This means the three UI shell asset documentation pages in the React library should get Usage and Style from Carbon Styles because of inheritance, and Code tab from the Carbon React library.

**Changed**

- Carbon Styles and Carbon React indexed descriptions per new design spec
- Formatting of the Carbon React `carbon.yml` to clean up white space in the file
- Alphabetized the assets in the Carbon Styles and Carbon React `carbon.yml`s so they are easier to audit
- Renamed the `id`s and thumbnail SVG names so they no longer have "unstable" in them, as "unstable" designation isn't permanent like identifiers

**Removed**

- `codePath` from indexed assets that no longer need a code tab. If the code tab previously just had links to other frameworks and live demos, since those are now shown in the asset overview tab, this PR removes the code tab from the index so the platform's asset detail pages no longer includes an empty code tab (page).

#### Testing / Reviewing

I tested this using https://github.com/carbon-design-system/carbon-platform/pull/1127 to verify that the changes to thumbnail file names and paths work. The visual changes:

You can now see UI shell left and right panels in the catalogs.
image

Also, these two libraries have new descriptions. (`INDEX_ALL=1` environment variable set for this screenshot.)
image


mattrosno:
@carbon-design-system/carbon-platform-devs can you review?

jnm2377 (approved):

tay1orjones (approved):

alisonjoseph (approved):

mattrosno:
@tay1orjones not sure why Percy and the AVT CI failed when this PR only changes `carbon.yml` files, and a few SVGs referenced in those files.

PR #11940, opened 2022-08-11
fix(ToolTip): don't pass in align as an option to tooltip div

crhuff-ibm:
Previously Tooltip was getting `align={...}` passed in as a prop on the parent div. This caused the ugly behavior seen below.
Remove `align` key word by destructuring it from `other`.

image


crhuff-ibm:

crhuff-ibm:
recheck

abbeyhrt (approved):
Looks good to me! Thank you for contributing!

jnm2377 (approved):

Issue #10184, opened 2021-11-28
React 18 Migration: Refactor the React test suite to use React Testing Library


tay1orjones:
This issue is to track the refactoring of our React test suite to use React Testing Library. 

<details>
  <summary>Links and resources</summary>
  
  - https://www.notion.so/carbondesignsystem/Audits-ca3fe640ffe242ac91bf3ab0749cb499?p=b1b911fa52a147d3af63d177c0a0ec3b&showMoveTo=true
  - https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/
  - https://kentcdodds.com/blog/avoid-nesting-when-youre-testing
</details>

<details>
  <summary>How to find enzyme usages</summary>
  
  The list below was generated by searching for `enzyme` within files having names containing our test suffix:
  
  https://user-images.githubusercontent.com/3360588/145886255-373f50c4-d3d3-4322-9963-bbea094192b4.mov
</details>

### Please use [the script to scaffold out new test files](https://github.com/carbon-design-system/carbon/issues/10184#issuecomment-1282894448)

```[tasklist]
### Wave 2
- [ ] https://github.com/carbon-design-system/carbon/issues/12665
- [ ] https://github.com/carbon-design-system/carbon/issues/12666
- [ ] https://github.com/carbon-design-system/carbon/issues/12667
- [ ] https://github.com/carbon-design-system/carbon/issues/12668
- [ ] https://github.com/carbon-design-system/carbon/issues/12669
- [ ] https://github.com/carbon-design-system/carbon/issues/12670
- [ ] https://github.com/carbon-design-system/carbon/issues/12671
- [ ] https://github.com/carbon-design-system/carbon/issues/12672
- [ ] https://github.com/carbon-design-system/carbon/issues/12673
- [ ] https://github.com/carbon-design-system/carbon/issues/12674
- [ ] https://github.com/carbon-design-system/carbon/issues/12675
- [x] https://github.com/carbon-design-system/carbon/issues/11736
- [ ] https://github.com/carbon-design-system/carbon/issues/12153
- [x] https://github.com/carbon-design-system/carbon/issues/12155
- [x] https://github.com/carbon-design-system/carbon/issues/12157
- [x] https://github.com/carbon-design-system/carbon/issues/12158
- [x] https://github.com/carbon-design-system/carbon/issues/12159
- [x] https://github.com/carbon-design-system/carbon/issues/12160
[...]

PR #11941, opened 2022-08-11
build(react): add a script that scaffolds out RTL tests

abbeyhrt:
REF #10184 

Since we've been updating tests to use RTL, I noticed that just scaffolding out the tests for all the props was taking a while and I thought that creating the test file for the new RTL tests could be automated. 

If it works (lol) this script should prompt you to choose a component to write tests for and then create a test file for that component with tests in place for its props and the appropriate imports. It's definitely pretty basic since I wanted it to work for every component so it's more of a jumping off point for tests than anything else. 

Since I like to refer to the other tests for help sometimes, this creates a new test file instead of overriding the previous one. 

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

1. Pull down the PR
2.  Run `yarn` to install `enquirer`
3.  In `packages/react`, run `node tasks/build-test-rtl.js` and follow the prompt. 
4. After selecting a component, verify that a new test file has been made with the name `${component}-test__copy.js`


tw15egan:
This is really cool and works really well! Nice job.

I think the only quirk I came across is when two components live in the same file, it only scaffolds out one. I tested `ProgressIndicator`, but it didn't create any tests for `ProgressStep`.

Regardless, this is super helpful and will help us make sure we can add tests for each prop 🎉

abbeyhrt:
@tw15egan Thanks for trying it out! And yeah :/ there are some special cases that I don't know how to get it to work for without just adding a bunch of `if()` statements, like Tile since that folder actually contains like 5 different components in one file and one test file. 

After you comment, I realized it wasn't working for Skeleton components so I updated it to work for those

tay1orjones (approved):
This is really awesome, thanks for putting it together!

tw15egan (approved):

PR #11932, opened 2022-08-09
new pictograms from backlog

dudley-ibm:
Closes #

{{closes issues 129, 132, 133, 137, 138, 145, 146, 147 and 149}}

#### Changelog

**New**

- {{New additions and their categories
Issue #129
chart--two-y-axis	(Data analytics)

Issue #132
flexibility	(Data storage)
reliability	(Data storage)
availability	(Data storage)
virtualization	(Data storage)
zero-trust	(Data storage)
cloud--infrastructure-management	(Data storage)
flexible--compute	(Data storage)
data--privacy--key	(Data storage)
resilience	(Data storage)

Issue #133
technical-owner	(Learning)

Issue #137
green-energy	(Industry)

Issue #138
sustainability	(Industry)

Issue #145
devices-at-ibm	(Device interaction)

Issue #146
ibm--z16--single-frame	(Data storage)
ibm--z16--multi-frame	(Data storage)
ibm--z16	(IBM)

Issue #147
decision-velocity	(Learning)
flexible--infrastructure	(Data storage)
quantum--safe	(Security)
ibm--z16--plus-cloud	(Data Storage)

Issue #149
cookie	(Lifestyle)
}}

{{ These pictograms should show up on both the carbon and IBM Design Language sites }}

PR #11950, opened 2022-08-12
replace z linux one pictos with z16

dudley-ibm:
**Removed**

- {{removed both `ibm--z-and-linuxone-multi-frame` and `ibm--z-and-linuxone-single-frame` pictograms from the Data Storage category since they are to be replaced by `ibm--z16--multi-frame` and `ibm--z16--single-frame` respectively, which were both added in PR #11932}}

#### Testing / Reviewing

{{ These removals should also be reflected on the IBM Design Language site }}


tw15egan:
@tay1orjones can we remove these immediately, or do we need to deprecate them first?

tay1orjones:
@tw15egan Yeah we can't delete these as they haven't been deprecated and removals would be a breaking change. This file needs updated instead:

https://github.com/carbon-design-system/carbon/blob/f26d4a0fd47ecf94d22e67c4cf14ad6cdefedb94/packages/icons/deprecated.yml#L1-L3

packages/elements/src/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -305,7 +305,6 @@ Array [
   "textSecondary",
   "themes",
   "toggleOff",
-  "unstable_metadata",
```
tay1orjones, line 308: This doesn't look correct. I'd try updating snapshots again after pulling in latest from `main`

tay1orjones (changes requested):
@tw15egan Yeah we can't delete these as they haven't been deprecated and removals would be a breaking change. This file needs updated instead:

https://github.com/carbon-design-system/carbon/blob/f26d4a0fd47ecf94d22e67c4cf14ad6cdefedb94/packages/icons/deprecated.yml#L1-L3

tay1orjones:
Hey, I'm sorry - in trying to update this PR I really messed with the history and there's a bunch of uneccesary files. Going to close this one in favor of a new PR: https://github.com/carbon-design-system/carbon/pull/11973

PR #11969, opened 2022-08-17
feat: add contained-list component

janhassel:
#10263

Implements two new components: `ContainedList` and `ContainedListItem`

#### Changelog

**New**

- Added `ContainedList` component
- Added `ContainedListItem` component
- Added related styles

#### Testing / Reviewing

- Verify with the specs shared by @laurenmrice in #10263


janhassel:
@aledavila @jnm2377

Quick note: I marked this PR as "ready for review" but it would be great if you could add the label **"do not merge"** to it as the variant names are obviously not final yet.

I still thought it would be good to start the conversation around the implementation, stories and tests. Then, once the variant names have been chosen, I can update this PR and we can remove the "do not merge" label again.

Let me know if this works for you!

janhassel:
@aledavila @jnm2377 With the docs merged and the variant names updated, this PR is ready for review.
Let me know if you have any questions!

laurenmrice (changes requested):
@janhassel Looks great, just a couple of things:

**Actions story**
- For the list title/header you are currently using the word `Heading` and in the website usage docs we are using the word “List title”. Would you rather we use "Heading" instead? Let me know and I can make those changes accordingly in the website documentation.

**Actions and Actions interactive stories**
- When entering long amounts of text that wrap to multiple lines in a row that has inline actions, the text sometimes gets too close to the icon and/or appears underneath the icons hover state. We need to add 16px padding after the text and before the container of the interactive states of the inline actions.
Artboard

laurenmrice (approved):
The "Heading Decorators" and "Playground" stories still say `Heading` instead of List title. Once that is updated it should be good to go.

Thanks so much for contributing this, @janhassel ! It looks awesome. 🚀 ⭐️

janhassel:
@laurenmrice Good catch, I missed those. Should be fixed now.
Btw, I think we can remove the "do not merge" label 😄

abbeyhrt (approved):

Issue #11785, opened 2022-07-12
Fluid inputs: TextInput


tay1orjones:
This issue is for providing the first experimental fluid input variant, TextInput. This will span both design and development to lay the groundwork for additional fluid input components to be added to the system, #5416. 

**Checklist**

- [x] https://github.com/carbon-design-system/carbon-website/issues/3091
  - [x] Create website branch to house updates
  - [x] Update "style" and "usage" tabs of component page to include Fluid variant
- [ ] Fluid inputs: Provide design tooling updates for TextInput
   - [x] https://github.com/carbon-design-system/carbon-design-kit/issues/557 (priority) 
- [ ] Fluid inputs: Add TextInput experimental implementation
  - [x] Determine if the existing FluidForm is needed, how it fits into the picture
    - FluidTextInput renders a FluidForm instead of it being behind the `isFluid` prop
  - [ ] ~Place implementation behind `isFluid` prop~ Place implementation within a new component, `FluidComponentName`
  - [ ] Add VRT tests via storybook story e2e setup
  - [ ] Add unit tests (likely simple, eg. "does the classname get added")
  - [ ] Add storybook example
  - [ ] Ensure styles are properly scoped to their own file
    - [ ] We could make a new fluid-text-input.scss that imports text-input.scss and just contains the fluid specific styles, sort of how we do MultiSelect and the other ListBox components. https://github.com/carbon-design-system/carbon/pull/11971#issuecomment-1219812870
- [x] #12018
- [x] #12080
- [x] #12081

## Spec

[Figma file](https://www.figma.com/file/9oQamCzUmVPw2jTiFkJ3Qw/Fluid-inputs_v11?node-id=0%3A1) | [Sketch file](https://ibm.box.com/s/4zcxjq750pbk0e0odl8vas0ff77fbko7)

![Text-input - White](https://user-images.githubusercontent.com/11670886/178830463-84cfe1bf-f777-4c2f-b30e-c8c9d03e6736.png)

aagonzales:
Additional missing features that need to be added to Fluid TextInput
- https://github.com/carbon-design-system/carbon/issues/12080
- https://github.com/carbon-design-system/carbon/issues/12081

PR #11971, opened 2022-08-17
feat(FluidTextInput): create `unstable__FluidTextInput` component

### LINKED ISSUE #11994: Use RTL for testing in `FluidTextInput/__tests__/FluidTextInput-test.js`

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/11785
Closes #11994

Utilizes existing `FluidForm` component to create a new `unstable__FluidTextInput` variant. 

#### Changelog

**New**

- `unstable__FluidTextInput` component, mimics the current usage of the `Fluid` `TextInput` example, but requires less work for end-users to configure

**To Do**
- [x] Write tests (#11994)  (will be completed after this is merged in, so we can utilize https://github.com/carbon-design-system/carbon/pull/11941)

#### Testing / Reviewing

Go to the `FluidTextInput story and ensure all variants are to spec and render correctly.


tay1orjones (commented):
This looks good so far! A few thoughts below.

Previously we had said the reasons for breaking something out a new component would be if it not only modifies styles but also behavior or accessibility. What behavior would be different between `TextInput` and `FluidTextInput`? Could we write separate tests to cover `FluidTextInput` specific behavior or would `TextInput` tests cover it? 

Overall, I do think the conflict of props between fluid vs. non-fluid support these being their own component.

Having all fluid inputs as their own components has a few advantages I can think of too:

* All fluid input stories will be co-located in the storybook left panel
* Each fluid input will have their own "folder" and set of stories that will be snapshotted and tracked with VRT/Percy.
* Makes it more clear for consumers the API difference between Fluid vs non-fluid

Do you think the mismatch between component name vs style file is a problem? In order to use `FluidTextInput` the user has to ensure `text-input.scss` styles are included. Conversely, I don’t think we’d want to duplicate all the styles in `text-input.scss` to a new `fluid-text-input.scss`, so the existing situation is probably fine.

What should we do with the existing `isFluid` props out there? Deprecate them with a note to use the new component could be a good first step.

tw15egan:
>Previously we had said the reasons for breaking something out a new component would be if it not only modifies styles but also behavior or accessibility. What behavior would be different between TextInput and FluidTextInput?

I agree with this in most circumstances, but since this seems to be a bundle of new components and not just a one-off variant, I tend to lean towards bunching them together. You've accurately summarized my thoughts on the advantages of creating a new component, and to me, they outweigh the negatives. This should also be coordinated with Design, to ensure they are on the same page and that they are planning to bundle Fluid components together, rather than have them co-exist inside the `TextInput` or other form element design files. 

>Do you think the mismatch between component name vs style file is a problem? In order to use `FluidTextInput` the user has to ensure text-input.scss styles are included. Conversely, I don’t think we’d want to duplicate all the styles in text-input.scss to a new fluid-text-input.scss, so the existing situation is probably fine.

I hadn't thought of that, that could be confusing. We could make a new `fluid-text-input.scss` that imports `text-input.scss` and just contains the `fluid` specific styles, sort of how we do `MultiSelect` and the other `ListBox` components. 

>What should we do with the existing isFluid props out there? Deprecate them with a note to use the new component could be a good first step.

I think that makes sense. Once this is merged (or included in this PR) we can mark the `isFluid` prop as deprecated noting that the new variant is available as a standalone component.

tay1orjones (approved):

tay1orjones:
> We could make a new fluid-text-input.scss that imports text-input.scss and just contains the fluid specific styles, sort of how we do MultiSelect and the other ListBox components.

@tw15egan Nice! That sounds like a good idea. Could include it here, or do it in a separate PR. I added it as an item on https://github.com/carbon-design-system/carbon/issues/11785

I also made https://github.com/carbon-design-system/carbon/issues/12003 for the `isFluid` deprecation

aagonzales (approved):
All the styles look good to me!! 🎉

tw15egan:
@tay1orjones you may want to check it over again (the last two commits). In #12003 I realized it was wrapping each `TextInput` in its own `form` wrapper. To avoid this, I removed the `FluidForm` wrapper in `FluidTextInput` and referenced the `isFluid` context directly. We still need that context value due to the way the elements are rendered in `TextInput`. We *may* need to use that in other `fluid` form elements that we create, but I'm not sure yet. 

Instead of scoping styles to `form.cds--form-fluid` they are scoped one level lower and I've added a new class `.cds--text-input--fluid`. I've ported all relevant styles to the new `fluid-text-input.scss` file but left the old styles in place until V12.

aagonzales:
All the styles still look good 👍

packages/react/src/index.js
```diff
@@ -66,6 +66,7 @@ export FileUploader, {
 export { FilterableMultiSelect } from './components/FilterableMultiSelect';
 export Form from './components/Form';
 export FluidForm from './components/FluidForm';
+export FluidTextInput from './components/FluidTextInput';
```
tay1orjones: I think we want these to be experimental from the start

```suggestion
export FluidTextInput as unstable__FluidTextInput from './components/FluidTextInput';
```

packages/react/src/components/FluidTextInput/FluidTextInput.stories.js
```diff
@@ -0,0 +1,168 @@
...
+  ToggletipLabel,
+  Toggletip,
+  ToggletipButton,
+  ToggletipContent,
+} from '../Toggletip';
+import { Information } from '@carbon/icons-react';
+import './test.scss';
+
+export default {
+  title: 'Components/FluidTextInput',
```
tay1orjones:
```suggestion
  title: 'Experimental/unstable__FluidTextInput',
```

From a quick search, it looks like some story files include the `unstable__` prefix, and others don't. Either way is good with me. `FluidForm` does not.

aledavila (approved):

PR #11972, opened 2022-08-17
fix(Toggle): add aria-labelledby to inner button

tw15egan:
Resolves an issue that surfaced in [Slack](https://ibm-studios.slack.com/archives/C046Y0YUD/p1660748194553729) 

Screen Shot 2022-08-17 at 5 12 05 PM

#### Changelog

**New**

- Added an `aria-labelledby` to the inner `Toggle` button element to resolve a11y violation

#### Testing / Reviewing

Ensure voiceover functionality is still functional, and a11y violations are resolved.

### DIFF
--- a/packages/react/src/components/Toggle/Toggle.js
+++ b/packages/react/src/components/Toggle/Toggle.js
@@ -75,6 +75,7 @@ export function Toggle({
         aria-checked={checked}
         disabled={disabled}
         onClick={handleClick}
+        aria-labelledby={id}
       />
       <label htmlFor={id} className={`${prefix}--toggle__label`}>
         <span className={labelTextClasses}>{labelText}</span>


aledavila (approved):

janhassel:
@tw15egan The button's accessible name is listed correctly in the accessibility tree in Firefox, Chrome and Safari and is being read accordingly in VoiceOver. Have we checked if the issue raised could this be a false flag from the a11y checker? According to the HTML spec, `<button>` is a labelable element and therefore the relation with the adjacent `<label for="button-id">` should be sufficient I believe.

tw15egan:
@janhassel I was thinking the same thing... I tested both with VoiceOver and didn't see any issues either. @dakahn is this something we should report to the a11y team?

PR #11981, opened 2022-08-19
fix(Notification): ensure correct background set in g100 low contrast

### LINKED ISSUE #11538: [Bug]: InlineNotification in Dark mode does not use proper text color
### Package

@carbon/react

### Browser

Safari

### Package version

1.4.0

### React version

18.1.0

### Description

When using lowContrast InlineNotifications in g100 the color for text is wrong. It is using the page text color (white) which makes it not readable.
Screen Shot 2022-06-02 at 10 20 31 AM

### Reproduction/example

https://stackblitz.com/edit/github-kkpgmp?file=src/App.jsx

### Steps to reproduce

Using Dark mode g100 add inline notification with lowContrast.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/11538

Ensures the correct background is set when zoning using `Theme` component

#### Changelog

**Changed**

- `$theme.layer` to `colors.gray-(one step up)`

#### Testing / Reviewing

There are two test stories for `ActionableNotification` and `InlineNotification` that set the `g100` theme in a small area of the storybook. The components should be readable. Stories will be removed before merging.

### DIFF
--- a/packages/react/.storybook/styles.scss
+++ b/packages/react/.storybook/styles.scss
@@ -12,18 +12,17 @@
   @include styles.theme(styles.$white);
 }
 
-[data-carbon-theme='g10'] {
-  @include styles.theme(styles.$g10);
-}
+// [data-carbon-theme='g10'] {
+//   @include styles.theme(styles.$g10);
+// }
 
-[data-carbon-theme='g90'] {
-  @include styles.theme(styles.$g90);
-}
+// [data-carbon-theme='g90'] {
+//   @include styles.theme(styles.$g90);
+// }
 
-[data-carbon-theme='g100'] {
-  @include styles.theme(styles.$g100);
-}
-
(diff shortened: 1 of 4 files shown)


packages/react/.storybook/styles.scss
```diff
@@ -12,18 +12,17 @@
...
-[data-carbon-theme='g100'] {
-  @include styles.theme(styles.$g100);
-}
+// [data-carbon-theme='g90'] {
+//   @include styles.theme(styles.$g90);
+// }
 
+// [data-carbon-theme='g100'] {
+//   @include styles.theme(styles.$g100);
+// }
```
tay1orjones: Also should remove this before merging - percy status check will fail until it's put back in

tay1orjones (approved):

abbeyhrt (approved):

PR #11985, opened 2022-08-22
fix(slider): prevent step value from causing error in input

### LINKED ISSUE #10171: [a11y]: Slider step size incrementation creating false negative most likely to be experienced by keyboard user
### Package

carbon-components-react

### Browser

Chrome

### Operating System

MacOS

### Package version

whichever one is currently live on carbondesignsystem.com

keyboard use

### Assistive technology

This issue will be encountered by anyone attempting to enter a value into the slider's text input . Keyboard-dependent users are the most likely group

### Description

There is an[ internal discussion](https://ibm-studios.slack.com/archives/G01GCBCGTPV/p1633455667022200) that provides more background. I will quickly summarize here:

The step size increment should not be used as a means of disallowing inputs within the range of the slider. This leads users to encounter errors which make no sense, due to input rules that are not exposed to the user. i also think the error messages do not address the underlying problem. It disproportionately penalizes users who cannot operate a slider or prefer to use the text input.

image
A slider showing an allowed range of 0-100, but an entry of 55 is showing as an error

- A slider has a visible range, by default 0-100.
- A slider's value can be set throughout this range by moving the slider or by entering a value directly into the input
- There is an ability in Carbon for an implementer to set a "step size increment" which determines what each increment on the thumb slider will do to the set value (e.g., if the slider has a step size of 10, the slider will move from 0 to 10, then to 20, etc)
- This step size is not displayed in the UI and will never be detected by someone who enters a value in the input

I think there could also be an interesting design discussion on the purpose of the step size increment.
Regardless of the range displayed, if the slider is always the same physical length, then there is no more or less effort for a pointer user to adjust the value. So I would assert that the underlying reason for the step increment was to support better keyboard interaction, and that it only need to be factored into the ability for the user to move both by 1s (using the arrow keys) and by the step increment (in Carbon, achieved with Shift+arrow)

IMO, the simple solution to this is to disallow changes of the step (or at the least, to not allow them unless the instructions and errors support that usage), and have a default stepMultiplier of 10.

### WCAG 2.1 Violation

3.3.1, 3.3.2, 3.3.3

### CodeSandbox example

https://www.carbondesignsystem.com/components/slider/code

### Steps to reproduce

1. using the code sample, https://www.carbondesignsystem.com/components/slider/code, change the step value from 1 to 10
2. put the focus on the slider then tab to the text input and enter a range of 55
3. tab off the input. the input will now be in an error state

The visible range allowed is 0-100. 55 is clearly in that range.
So there are multiple problems here:
1) The value should be allowed by all labelling and instructions visible
2) There is no error message describing the problem (already opened with https://github.com/carbon-design-system/carbon/issues/9772)
3) The error relies on color (already opened with https://github.com/carbon-design-system/carbon/issues/9772)

Looking at the code, it seems that the step value is currently used to determine the increments in the slider, and the stepMultiplier is used to determine what happens when a keyboard user presses Shift+arrow to increment the slider (i.e., Shift+arrow will cause the slider to move by 40).

```
<Slider
  ariaLabelInput="Label for slider value"
  id="slider"
  labelText="Slider label"
  max={100}
  min={0}
  step={10}
  stepMultiplier={4}
  value={50}
/>
```

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/10171

When an input is provided with `Slider`, all values inputted inside the slider range will no longer throw errors. Numbers below and above the min-max will still throw errors (as they are outside the allowed range).

#### Changelog

**Changed**

- - Previously, if the `step` was 10 and the range was `0-100`, and a user entered `55` inside the `Slider` input, it would error. This was wrong because the 55 is still inside the allowed value range of 0-100. The Slider input step will always be `1`, and the `step` will only apply to the slider handle itself. 

#### Testing / Reviewing

Take a look at the `Slider` `Step Test` story. The handle should still increment via step/step multiplier, but the input should allow any value inside the range.

### DIFF
--- a/packages/react/src/components/Slider/Slider.js
+++ b/packages/react/src/components/Slider/Slider.js
@@ -545,7 +545,7 @@ export default class Slider extends PureComponent {
       maxLabel,
       formatLabel = defaultFormatLabel,
       labelText,
-      step,
+      step, // eslint-disable-line no-unused-vars
       stepMultiplier, // eslint-disable-line no-unused-vars
       inputType,
       required,
@@ -664,7 +664,7 @@ export default class Slider extends PureComponent {
                   required={required}
                   min={min}
                   max={max}
-                  step={step}
+                  step={1}
                   onChange={this.onChange}
                   onBlur={this.onBlur}
                   onKeyUp={this.onInputKeyUp}
--- a/packages/react/src/components/Slider/next/Slider.stories.js
+++ b/packages/react/src/components/Slider/next/Slider.stories.js
@@ -46,6 +46,33 @@ export const Default = (args) => (
   />
 );
 
+export const StepTest = (args) => (
+  <>
+    <Slider
+      {...args}
+      labelText="Slider Label"
+      value={50}
+      min={0}
+      max={100}
+      step={5}
+      stepMultiplier={2}
+      noValidate
+    />
+    <br />
+    <Slider
+      {...args}
+      hideTextInput
+      labelText="Slider Label"
+      value={50}
+      min={0}
+      max={100}
+      step={5}
+      stepMultiplier={2}
+      noValidate
+    />
+  </>
+);
+
 Default.story = {
   name: 'Slider',
 };
--- a/packages/react/src/components/Slider/next/Slider.stories.js
+++ b/packages/react/src/components/Slider/next/Slider.stories.js
@@ -46,32 +46,37 @@ export const Default = (args) => (
   />
 );
 
-export const StepTest = (args) => (
-  <>
-    <Slider
-      {...args}
-      labelText="Slider Label"
-      value={50}
-      min={0}
-      max={100}
-      step={5}
-      stepMultiplier={2}
-      noValidate
-    />
-    <br />
-    <Slider
-      {...args}
-      hideTextInput
-      labelText="Slider Label"
-      value={50}
-      min={0}
-      max={100}
-      step={5}
-      stepMultiplier={2}
-      noValidate
-    />
-  </>
-);
+export const StepTest = (args) => {
+  const [val, setVal] = useState(0.5);
+  return (
+    <>
+      <Slider
+        {...args}
+        labelText="Slider Label"
+        value={0.5}
+        min={0}
+        max={1}
+        step={0.1}
+        stepMultiplier={2}
+        noValidate
+      />
+      <br />
+      <p>Slider value: {val.toFixed(1)}</p>
+      <Slider
+        {...args}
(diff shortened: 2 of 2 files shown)


packages/react/src/components/Slider/Slider.js
```diff
@@ -664,7 +664,7 @@ export default class Slider extends PureComponent {
                   required={required}
                   min={min}
                   max={max}
-                  step={step}
+                  step={1}
```
tay1orjones, line 667: I caught this in the original issue:

> ...the step value should be required to be "1" anytime an input field is provided

Does this mean when the input field is not provided it can be anything?

```suggestion
                  step={hideTextInput ? step : 1}
```

tw15egan: That's what I originally had but noticed that it was always `1` no matter what since the step/step multiplier on the slider handled is calculated elsewhere, not handled through the input. Since the input element won't be visible / doing any step calculation, I don't even think we need that level of logic. Let me know if I'm missing a use case, though!

tay1orjones: If `step` is not configurable on the `input`, would it block float values or the special `any` value that can be given? Based on the [MDN entry for `step`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step), it sounds like having it always `1` would only allow integer values.

tw15egan: Just pushed an update that uses `float` values, but I didn't see any difference between the two. Let me know if I'm still missing something

tay1orjones: Thanks, that example is helpful - it shows that the values are calculated correctly for the slider, and the input updates correctly as well. The problem is when the input gets focus and then blurred the validation runs, there is a `stepMismatch`.


https://user-images.githubusercontent.com/3360588/185998650-6d1a69cf-4f79-4f8a-a3ac-a7fdbc6cefd4.mov

In this case I think the user would need to specify `step=".1"`.

tw15egan:
After talking with Taylor, it seems like this needs more of a documentation update rather than a fix to the internal workings of `Slider` / `NumberInput` based on the MDN definition of the `step` attribute. If a `step` other than one is specified, all other values should be considered invalid and treated as such. If the `step` is changed (not equal to 1), it must be conveyed to the users

Reference:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step#accessibility_concerns

PR #12000, opened 2022-08-23
Convert packages/react storybook to typescript. Convert checkbox story to tsx

### LINKED ISSUE #11587: Support TypeScript in our Storybook to set a baseline for types
This issue is connected to our TypeScript Working Group. As part of this group, we would like to support TypeScript in the React storybook of the project in order to establish a baseline for types. The storybook itself was chosen due to its coverage of usage scenarios for the components in the system which would be helpful for making sure types are correct.

**Checklist**

- [ ] Update Storybook to support compiling stories with TSX
- [ ] Bring in initial set of types and verify TS can be run against the storybook stories

mbarrer:
Closes #11587 

Changes have been made to allow the use of typescript within `packages/react` so that storybook stories can start to be written in typescript. As part of this change, the main.js for storybook has now been converted to main.ts and has some initial typings. Typing files have been added for Checkbox that we're sourced from the DefinitelyTyped repo and adjusted for carbon v11

#### Changelog

**New**

-  `"typescript": "^4.7.4"`
-  `"@typescript-eslint/eslint-plugin": "^5.30.5"` and `"@typescript-eslint/parser": "^5.34.0"` for linting
- `"ts-node": "^10.9.1"` for supporting imports in storybook main.ts 
- `typings/shared.d.ts` common types file pulled from DefinitielyTyped
- `components/Checkbox/Checkbox.d.ts` new typing file for Checkbox (adjusted DT typings)
- `custom.d.ts` ambient module to allow imports of .mdx files in .tsx
- `tsconfig.json`

**Changed**

- update `mini-css-extract-plugin` to support native ts typings
- converted `.storybook/main.js` to typescript and allow stories to be written in typescript
- converted `components/Checkbox/Checkbox.stories.js` to tsx using the new typings
- Added linting for .ts, .tsx to `eslint-config-carbon/plugins/react.js` 

#### Testing / Reviewing

In order to test:
- fresh yarn install
- new build
- launch storybook and verify that the Checkbox story is working and has the same functionality as it had before


mbarrer:

packages/react/.storybook/main.ts
```diff
@@ -92,9 +89,10 @@ module.exports = {
   },
   framework: '@storybook/react',
   stories,
-  webpack(config) {
-    const babelLoader = config.module.rules.find((rule) => {
-      return rule.use.some(({ loader }) => {
+  // TODO: Fix typings for this function
```
sierrawetmore: Just want to check whether this was a todo that you meant to do before this was merged or just sometime in the future.

packages/react/src/typings/shared.d.ts
```diff
@@ -0,0 +1,104 @@
...
+export type ReactAnchorAttr<T = HTMLAnchorElement> =
+  React.AnchorHTMLAttributes<T>;
+export type ReactButtonAttr<T = HTMLButtonElement> =
+  React.ButtonHTMLAttributes<T>;
+export type ReactDivAttr = ReactAttr<HTMLDivElement>;
+export type ReactInputAttr<T = HTMLInputElement> = React.InputHTMLAttributes<T>;
+export type ReactLabelAttr<T = HTMLLabelElement> = React.LabelHTMLAttributes<T>;
+export type ReactLIAttr<T = HTMLLIElement> = React.LiHTMLAttributes<T>;
+
+// export type ShapeOf<
```
sierrawetmore: Is this meant to be left in?

sierrawetmore (approved):

tay1orjones:
@mbarrer Are you using Yarn v1 by chance? The project is configured to need `yarn@3.2.1`. I believe this is why the deploy previews are failing. I think if you update/install to v3.2.1 and re-run `yarn install` from the root things should start working. 

Since you added new dependencies there should also be some additions to `.yarn/cache` as well for you to commit and push up.

packages/react/src/typings/shared.d.ts
mbarrer: ah no I can remove these 2 commented out types

packages/react/.storybook/main.ts
mbarrer: Yeah this was meant to be built upon in the future. I wasnt able to fully flesh out strict typings for that function due to some complex union typings used by storybook so I left a TODO here when I get some more time to dig into the types

mbarrer:
Ah yes I have yarn 1.22.18 installed, I can update this on my end and run a fresh yarn install

tay1orjones:
@mbarrer Hey so out of curiosity I pulled down this PR, reset `yarn.lock` to what's currently in `main`, and then re-ran yarn install. This pulled the changed files down from ~1800+ to just 26. I think the snafu with yarn v1 inadvertently reset all the hashes on the yarn cache, which is why it was such a huge change set.

Anyway, should be fixed now 👍

packages/react/.storybook/main.ts
```diff
@@ -92,9 +89,10 @@ module.exports = {
...
+  webpack(config: any) {
```
metonym: If statically typing this is desired, there are two options:

1. Only use `Configuration` type from Webpack

```ts
import type { Configuration } from 'webpack';
...
webpack(config: Configuration) {
```

2. Type the entire config using `StorybookConfig`

```ts
import type { StorybookConfig } from '@storybook/core-common';

const config: StorybookConfig = { /** */ }

module.exports = config;
```

mbarrer: Yep nice call, there was a few other changes needed in the body of the config to account for the union types used in the `webpackFinal` prop. The use of the typings also caught the fact that the `webpack` property should have actually been renamed to `webpackFinal` according to the storybook 6.5 docs

mbarrer:
Looks like the build task also needs to be adjusted to output a main.js file for use in netlify, will be taking a look at this

erifsx (approved):
This looks good to me.  

I was concerned about what would happen if we released this without all the other types, or at least empty `export declare let XX = unknown` declarations for everything so I looked in the compiled lib and es folders and it doesn't look like it's actually shipping the types there yet and carbon doesn't ship the src folder.  When we want to "go live" with the feature, I believe they'll need to be copied to the appropriate place along with an index.d.ts file.

Given all that, this looks like a great way to dark release our work.

mbarrer:
Just letting you guys know that I haven't forgot about this PR, have just been busy this week with story work. I'll aim to have the build issue resolved by the end of the week

packages/react/.storybook/main.ts
```diff
@@ -4,19 +4,17 @@
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
+import fs from 'fs';
```
falsepopsky: I think it would be a good practice to put `node:` before using each node core modules/lib, etc.

example:

`import fs from 'node:fs'` [fs Docs](https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#file-system)
`import path from 'node:path'` [Path docs](https://nodejs.org/dist/latest-v16.x/docs/api/path.html#path)

mbarrer: oh cool I didnt know you could reference these like this. good to know!

mbarrer:
Seems like the issue I was hitting was something that @erifsx brought up in a thread on the [workgroup slack](https://ibm-systems-storage.slack.com/archives/C03C8VASVED/p1657912847828949?thread_ts=1657141301.263859&cid=C03C8VASVED). We found that we didnt need ts-node to get things running initially but for some reason this is now required in order to run the main.ts. Adding this to the dev dependancies seems to resolve the issues with netlify

metonym (approved):

config/eslint-config-carbon/plugins/react.js
```diff
@@ -44,6 +44,11 @@ module.exports = {
     ],
   },
   overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      plugins: ['@typescript-eslint'],
+      extends: ['plugin:@typescript-eslint/recommended'],
+    },
```
jharvey10: Assuming carbon is using the `no-unused-vars` JS rule, you'll want to turn that off in favor of the typescript-eslint one. [Here's an example](https://github.com/carbon-design-system/carbon-platform/blob/4bd974df5d0a89ad7d68a00068a3f346b6067668/.eslintrc.js#L98)

packages/react/src/components/Checkbox/Checkbox.Skeleton.d.ts
```diff
@@ -0,0 +1,7 @@
+import { FunctionComponent, HTMLAttributes } from 'react';
```
jharvey10, line 8: Missing copyright statement

packages/react/src/components/Checkbox/Checkbox.d.ts
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react';
```
jharvey10, line 8: Missing copyright statement

packages/react/src/components/Checkbox/index.d.ts
```diff
@@ -0,0 +1,5 @@
+import Checkbox from './Checkbox';
```
jharvey10, line 8: Missing copyright statement

packages/react/src/components/Notification/Notification-test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-standalone-expect */
```
jharvey10: Which part of this file is violating this rule? Should it instead be updated to conform to it?

packages/react/src/custom.d.ts
```diff
@@ -0,0 +1,5 @@
+// Ambient module to allow the importing of .mdx files in storybook .tsx files
+declare module '*.mdx' {
+  const content: any;
```
jharvey10: We should definitely aim for zero warnings. This should either be `unknown` or if working around the lack of explicit `any` is too onerous to devs (since this is a gradual conversion), perhaps we consider disabling the `no-explicit-any` rule to start.

mbarrer: Good point, i should have changed this to `unknown`

packages/react/src/components/Notification/Notification-test.js
mbarrer: Hmm not sure why this was added, removing does not seem to cause any warnings to show

mbarrer:
@jdharvey-ibm updated PR based off your recent comments

tay1orjones:
Closing in favor of https://github.com/carbon-design-system/carbon/pull/12222

PR #12012, opened 2022-08-26
feat: add controls and example for context menu

### LINKED ISSUE #11719: Ensure ContextMenu-story.js has working controls
In addition to a playground and using controls, this component story needs to be refactored so the story is more readable and doesn't use the generator function pattern it currently has.

jnm2377:
Closes #11719 

#### Changelog

- adds a playground story for the unstable `ContextMenu` stories, though there aren't a lot of props that are appropriate for controls
- adds a hard coded example of the component as well instead of relying on the story util like the other context menu stories, which makes them unfriendly to copying and pasting code

#### Testing / Reviewing

- test Playground story for `ContextMenu` and ensure the selection is changing as appropriate
- ensure the story example makes sense


abbeyhrt (approved):

tay1orjones (approved):

packages/react/src/components/ContextMenu/ContextMenu-story.js
```diff
@@ -37,6 +54,43 @@ const Story = (items) => {
...
+        <ListItem>onClose</ListItem>
+        <ListItem>className</ListItem>
+        <ListItem>id</ListItem>
+        <ListItem>target</ListItem>
+      </UnorderedList>
+    </p>
+    <p>
+      The
+      <CodeSnippet type="inline">{`<MenuGroup>`}</CodeSnippet> and
+      <CodeSnippet type="inline">{`<MenuItem`}</CodeSnippet>
```
tay1orjones:
```suggestion
      <CodeSnippet type="inline">{`<MenuItem>`}</CodeSnippet>
```

packages/react/src/components/ContextMenu/ContextMenu-story.js
```diff
@@ -37,6 +54,43 @@ const Story = (items) => {
...
+      <CodeSnippet type="inline">{`<MenuRadioGroup>`}</CodeSnippet>) do not
+      accept children. The{' '}
+      <CodeSnippet type="inline">{`<MenuRadioGroup>`}</CodeSnippet> accepts an
+      array of items to display as a group of single choice selection.
+    </p>
+  </div>
+);
+
 export const _ContextMenu = () =>
   Story([
```
tay1orjones: Could we refactor these two stories (ContextMenu, MultipleGroups) to follow the normal pattern instead of passing an array of items to a function? The story source panel doesn't even load for me on these stories.

I think using a function to generate the items via `buildMenu(items)` is fine, it's just calling the `Story` function that seems to be gunking things up.

tay1orjones (changes requested):
Looks great! A couple things I found below. Would be nice to get the story source working in some fashion if we can?

packages/react/src/components/ContextMenu/ContextMenu-story.js
jnm2377: @tay1orjones yeah, it's not great. I can update these 👍🏽

tay1orjones (approved):
Awesome!

PR #12013, opened 2022-08-26
fix(modal): update aria label functionality and storybook controls

### LINKED ISSUE #11868: [Bug]: aria label is only applied to modal if modelLabel is passed
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

@carbon/react 1.8.0

### React version

17.0.2

### Description

https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Modal/Modal.js#L158

The logic is wrong, it should be `rest['aria-label']` as the remaining props are put into the rest variable, what's happening now is if you do not provide the `modalLabel ` prop then it tries to use `['aria-label']` which will just pass in `aria-label`

### Reproduction/example

n/a

### Steps to reproduce

n/a

### Code of Conduct

aledavila:
Closes #11868

Updates aria labels variable to add `rest` and added storybook controls

#### Changelog

**Changed**

- aria-label const
- added storybook controls to modal

#### Testing / Reviewing

Make sure aria label is applied to modal without the `modalLabel` being present. Use the playground for testing and make sure its applied to the modal container. 

Make sure stories and controls are working as intended


tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #12024, opened 2022-08-29
feat(Toggletip): update `align` accepted props

tw15egan:
Noticed while working on the `ToggleTip` inside `FluidTextInput` that `ToggleTip` currently only allowed 4 `align` directions, but the underlying component (`Popover`) supports 12. I've updated the accepted props for `ToggleTip` to allow these extra variations.

#### Changelog

**Changed**

- `ToggleTip` now supports all directions provided by the underlying `Popover` component
- Updated the playground story to preview all possibilities. 
- Updated snapshots

#### Testing / Reviewing

Go to `Toggletip` playground story and try testing the different `align` variations.


abbeyhrt:
I think it _might_ have been intentional to limit ToggleTip to only those 4 alignments, I'm not sure. I added design as a reviewer to see if it was on purpose.

abbeyhrt (approved):
If the intent is to have all alignments available, then this looks good to me!

laurenmrice (changes requested):
Nice catch @tw15egan ! 

The `left-bottom` and `left-top` alignments do not have the 16px padding before or after the caret. Other than that, everything else looks great!

tw15egan:
@laurenmrice nice catch! Updated 👍🏻

tay1orjones (approved):

laurenmrice (approved):

PR #12034, opened 2022-08-30
chore(carbon__icons-react): ship TypeScript definitions

### LINKED ISSUE #11317: [Bug]: @carbon/react/icons module does not provide IDE import auto-complete

metonym:
Closes #11317

This PR is based on #11533. I simply copied and rebased the PR on `main`.

---

Currently, `@types/carbon__icons-react` has to be manually updated for TypeScript support.

The goal for this PR is to better support TypeScript for the `@carbon/icons-react` package without rewriting it completely. By generating and shipping definitions with each release, `@types/carbon__icons-react` can be removed.

Related discussions in `DefinitelyTyped`:

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61229
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52664#issuecomment-834800128

#### Changelog

**New**

- chore(carbon__icons-react): ship TypeScript definitions

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "sync": "carbon-cli sync",
     "test": "cross-env BABEL_ENV=test jest",
     "test:e2e": "cross-env BABEL_ENV=test jest -c jest.e2e.config.js",
+    "typecheck": "tsc --skipLibCheck --noEmit",
     "postinstall": "husky install"
   },
   "resolutions": {
@@ -73,7 +74,8 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "rimraf": "^3.0.0",
-    "stylelint": "^14.3.0"
+    "stylelint": "^14.3.0",
+    "typescript": "^4.7.2"
   },
   "commitlint": {
     "extends": [
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "sync": "carbon-cli sync",
     "test": "cross-env BABEL_ENV=test jest",
     "test:e2e": "cross-env BABEL_ENV=test jest -c jest.e2e.config.js",
-    "typecheck": "tsc --skipLibCheck --noEmit",
+    "typecheck": "tsc -b",
     "postinstall": "husky install"
   },
   "resolutions": {
--- a/packages/icons-react/types/Icon.d.ts
+++ b/packages/icons-react/types/Icon.d.ts
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import * as React from 'react';
+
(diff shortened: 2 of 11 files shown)


tay1orjones (approved):
Thanks for resubmitting this! I don't recall any blockers from the original PR

tsconfig.base.json
```diff
@@ -0,0 +1,26 @@
...
+    "noUnusedParameters": true,
+
+    "alwaysStrict": true,
+    "preserveConstEnums": true,
+
+    "jsx": "preserve",
+
+    "types": []
+  }
+}
```
jharvey10, line 26: I'm not sure this is the PR in which we want to define the overall monorepo's tsconfig file. I'd be more comfortable with each respective package defining their own for now until we have a global baseline that is defined by the TS working group.

Separately, It might be worth generating the basic tsconfig files using the `tsc --init` command since it outputs all *(read: most)* possible settings and gives nice context to each of them. Then from there, we can make changes to the configs above and beyond the defaults as needed.

package.json
```diff
@@ -26,6 +26,7 @@
     "sync": "carbon-cli sync",
     "test": "cross-env BABEL_ENV=test jest",
     "test:e2e": "cross-env BABEL_ENV=test jest -c jest.e2e.config.js",
+    "typecheck": "tsc -b",
```
jharvey10, line 29: This script name implies type-checking and not necessarily building. Would it make more sense for this to be `tsc --noEmit`?

packages/icons-react/types/Icon.d.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright IBM Corp. 2019, 2019
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import * as React from 'react';
+
+type IconSize = 16 | 20 | 24 | 32 | '16' | '20' | '24' | '32' | number | string;
```
jharvey10, line 10: [There's an outstanding comment](https://github.com/carbon-design-system/carbon/pull/11533/files#r890708661) relating to this that I think needs some discussion. You can ignore my comment on it because the answer to what I asked over there is "yes", but the comment about IDE autocompletion is worth exploring further.

Also, I'm not sure why strings are allowable on this prop. @tay1orjones Do react icons allow you to specify string versions of the number sizes?

tay1orjones: @jdharvey-ibm Yes, both strings or number literals [are allowed](https://stackblitz.com/edit/github-b36b4u?file=src%2FApp.jsx)

PR #12050, opened 2022-09-02
fix(styles): add display property for custom HTML elements

### LINKED ISSUE #11579: [Bug]: Need to update styles to work with custom HTML elements
### Package

@carbon/styles, carbon-components

### Browser

Chrome, Safari, Firefox, Edge

### Package version

@carbon/styles - v1.4.0, carbon-components - 10.56.0

### React version

v17

### Description

This issue can also be considered as a feature request/suggestion. I'm currently working on [carbon-components-angular](https://github.com/IBM/carbon-components-angular) to add support for carbon v11. The table wrapper class `.bx--data-table-content` & `cds--data-table-content` don't have `display: block` properties. This results in horizontal scrolling not working, normally we would bind the class to the custom element created (In this case `<ibm-table></ibm-table>`). But in this case, we ALSO have to explicitly go in & set the style for display to block.

The downside of this is that we have to maintain this style and make changes to it if newer versions of @carbon/style effect the component. This can possible cause version mismatch issues for some of the consuming teams.  There are other approaches that we can take to address this, but pollutes the DOM with unnecessary elements.

**There are other components as well where we had to explicitly set the style for the component.**

If you think this is something that we can address, I can contribute to this issue.

### Reproduction/example

https://angular.carbondesignsystem.com/?path=/story/components-table--with-dynamic-content

### Steps to reproduce

1. Go to the angular storybook
2. Click the `Add columns` button 25 - 30x. 
3. Observe the horizontal scrolling.
4. Inspect element & find `<ibm-table>` element (There should only be 1)
5. Add following class `bx--data-table-content` to the element (May need to Edit as HTML in some browsers)
6. Notice that this doesn't fix horizontal scrolling
7. Add `display: block` style to class `bx--data-table-content` after selecting `<ibm-table>`  with `.bx--data-table-content` class

### Code of Conduct

Akshat55:
Closes #11579

Specify `display` CSS property for specific wrapper classes. 

In Angular, we have custom tags `<ibm-table></ibm-table>` unlike React, which uses traditional HTML tags (`div`, `span`, etc).

This results in browsers not knowing what the display style property of the tag should be (block, inline, etc). Where applicable, we have assigned a carbon class to the host e.g. `<ibm-table class="bx--data-table-content"></ibm-table>`. But there are cases where those classes have not specified the display property, table is one of them. (Not assigning `display: block` breaks horizontal scrolling)

This change will allow custom HTML elements such as `<ibm-table class="bx--data-table-content"></ibm-table>` (in Angular) to instruct the browser to use specific display property instead of `inline` by default. It will also allow us (carbon-components-angular team) to not specify styles within the component template.

#### Changelog

**New**
- Specified `display` property for wrapper classes of accordion item, table, & context menu components.

#### Testing / Reviewing
1. [ ] Build the styles package
2. [ ] Comment out or remove **ALL** styles in React package storybook `styles.scss` (`packages/react/.storybook/styles.scss`)
3. [ ] Import the built styles in React package storybook styles file:
```scss
@import '../../styles/css/styles.css';
```
4. [ ] Run storybook & verify if there are any visual changes to Accordion, Data Table, & Context Menu components. 

I have reviewed React package code & verified via storybook to ensure these changes do not impact the package.

You can also verify if the display properties make sense to the element they are applied to, for example, `.cds--accordion__item` class is applied to list (`<li></li>`) which the browser assigns the default display property of `list-item`. 

Another example would be, `.cds--data-table-content`. This class is applied to a `<div>`, which browsers assign `block` as the display value.

Hence, these changes should not impact React package but will greatly benefit custom HTML elements (like the one's used in carbon-components-angular).


Akshat55:

tay1orjones:
It would be good to run VRT against multiple browsers [locally ](https://github.com/carbon-design-system/carbon/blob/2270ef70d32e8e634c919399ad52ecc704c83877/docs/testing.md#L107) for this one. I anticipate that adding `display: block` may cause some visual issues we won't be able to catch just by eye.

tay1orjones:
@Akshat55 It appears Angular has addressed this via the `--displayBlock=true` flag. 
- https://github.com/angular/angular-cli/issues/12244 
- https://github.com/angular/angular-cli/pull/16299

It seems like that might be a more appropriate solution rather than duplicating the default `display` properties on the components here. Let me know what you think.

Akshat55:
@tay1orjones We are already using the approach listed, but this is applying `inline styles` to the element. This seems unprofessional as the same inline style property is applied to all instances of that component, as shown in the image below.

image

Another issue with inline styles is that if the users want to make changes to the style by applying a class to the element, then they will have to use `!important` to override.

We hope to achieve something similar to what [Angular Material has done](https://github.com/angular/components/blob/e3adcfe69abc9f997a777d93ea2b9297fdee97e8/src/material/card/card.scss#L44), where the display properties are defined in the styles. Now we can override the classes on our end, but since we are already applying the wrapper classes from carbon, it makes sense to apply this change directly to the styles package.

Also, I don't think this is duplicating the display property since that `display` property isn't there in the styles package by default. This ensures that the display property stays consistent regardless of the HTML element used.

tay1orjones (approved):
That makes sense, thanks @Akshat55!

Akshat55:
Thanks @tay1orjones!

We've discovered a few more, should I include them as well in this PR or should I create another? Specifically for setting display properties for UI-shell side navigation items (`display: list-item`).

tay1orjones:
@Akshat55 A separate PR would be great!

tay1orjones:
bump for review @abbeyhrt

abbeyhrt (approved):

PR #12053, opened 2022-09-02
feat(FluidTextArea): adds new fluid text area

### LINKED ISSUE #12018: Fluid inputs: Add `FluidTextInput` to `FluidForm` story
For the FluidInputs we need to update the FluidForm story to include FluidTextInput

**Checklist**

- [ ] Update FluidForm story to proper format (remove info, refactor knobs to controls)
- [ ] Configure `default` story to use FluidTextInput

### LINKED ISSUE #12020: Fluid inputs: Add `FluidTextArea` experimental implementation
This issue is for the primary implementation of the FluidTextArea component.

**Checklist**

  - [ ] Place implementation within a new component, `FluidComponentName`
  - [ ] Export as `unstable__FluidComponentName`
  - [ ] Ensure this uses the Context's `isFluid` properly (see https://github.com/carbon-design-system/carbon/pull/11971 for example)
  - [ ] Add VRT tests via storybook story e2e setup
  - [ ] Add unit tests (likely simple, eg. "does the classname get added")
  - [ ] Add storybook file and examples
  - [ ] Ensure styles are properly scoped to their own file, `fluid-text-area-input.scss` (again, see https://github.com/carbon-design-system/carbon/pull/11971 as an example)

aledavila:
Closes #12020
Closes #12018 

Adds fluid text area


aagonzales (commented):
Looking good just a few things

**Component**

- [x] Should the resize handle feature also be disabled in the disabled state?
- [x] Error + focus state: when an input with an error state receives focus the border should turn blue/`$focus` to indicate its been given focus. This is what we do across invalid states with the default inputs. 
   image
- [x] As of Friday, I added in a Skeleton state for all the fluid input. It was an oversite in the original spec, so can we add that in now? 
   image


---

**Storybook**

- [ ] Can we add a `$layer-01`background behind the second example?
   image

packages/react/src/components/FluidTextArea/FluidTextArea.stories.js
```diff
@@ -0,0 +1,183 @@
...
+  ToggletipLabel,
+  Toggletip,
+  ToggletipButton,
+  ToggletipContent,
+} from '../Toggletip';
+import { Layer } from '../Layer';
+import { Information } from '@carbon/icons-react';
+
+export default {
+  title: 'Experimental/unstable__FluidArea',
```
tay1orjones:
```suggestion
  title: 'Experimental/unstable__FluidTextArea',
```

tay1orjones (approved):
One small change, otherwise looks great! That `playgroundWidth` control is a super cool idea 🔥

aagonzales (approved):
I didn't see the storybook change for the background color in the layers example but the component itself looks good to go.

jnm2377 (approved):

aledavila:
@aagonzales yea i think theres an issue to look at with layers and adding backgrounds unrelated to text area. so we will take a look at that in a separate issue

PR #12092, opened 2022-09-13
fix(type): update expressive-heading-04(fluid-heading-04)

### LINKED ISSUE #3149: [bug]$expressive-heading-04:xlg breakpoint size is wrong
The type stays the same size between the XLG and large breakpoint (28px) but goes from light to regular weight.

According to our specs the font size is 32px at the XLG breakpoint

image

Anytime the font is at 32px OR GREATER it should stay at light weight. It should only go back to Regular weight when it jumps down to 28px at the large breakpoint

$expressive-heading-04 (v11) > XLG breakpoint should be 32px, light weight
$fluid-heading-04 (v10) > XLG breakpoint should be 32px, light weight

You can see in the video that the type is staying the same size, but changing weights only.

design kit library page

design kit library page

## Detailed description

> What did you expect to happen? What happened instead? What would you like to see changed?

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

tay1orjones:
Closes https://github.com/carbon-design-system/carbon-website/issues/3149

#### Changelog

**Changed**

- correct the font size to be 2rem (32px) at XLG breakpoint

#### Testing / Reviewing

View the type elements preview and ensure this value has been updated in both the source as well as the collapsed token information output


jeanservaas:
It looks in the preview link like it's just jumping from regular to light all at once. I'm wondering if it's because it goes to regular weight like immediately beyond 1312px and the light jumps to regular weight all at once @oliviaflory do you think?



https://user-images.githubusercontent.com/43144260/190005265-be6afb0b-8cfa-42ae-aa04-713d675aaf38.mp4

jeanservaas:
@oliviaflory mentioned that the fluid tokens have a script (or something) to make them scale progressively when they change.

So at 1056 it should = 28px (but it’s not FYI)
between 1057–1311 it should range between 28-32px (30.99, etc)
And at 1312 = 32px

So it basically tweens between the two sizes... ideally, the weight would do that tweening too. But if we can't do that, we may have to reconsider this weight change.

tay1orjones:
@jeanservaas @oliviaflory If you're using the variable font file, you should be able to animate `font-weight` which would provide a smooth transition from one weight to another. This is outside the scope of what we provide through the monorepo though. The tokens are statically defined and we only provide the default font family/weight files.

jeanservaas (approved):
Let's actually just move forward on the type size fix for the XLG breakpoint. The systems and .com designers got together to talk about the weight change for expressive/fluid-heading-04 tokens, and we may actually make a separate issue to revert those changes... because Kenny is saying that the variable fonts do not actually animate between weights.

tw15egan (approved):
LGTM 👍 ✅

jnm2377 (approved):

PR #12094, opened 2022-09-14
fix(Toggle): fix issue with side label when hideLabel is true

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/6567

Fixes an issue I noticed when working on #6567 

It seems like the voice-over issue no longer exists (🎉), but when `hideLabel` is set to true, the label is shown instead of the toggle text.

Screen Shot 2022-09-14 at 9 58 48 AM
Screen Shot 2022-09-14 at 9 58 43 AM

#### Changelog

**Removed**

- Removed some extra code that was placing the label text incorrectly

#### Testing / Reviewing

Test with VoiceOver in Chrome / Safari and ensure the label and toggles are read correctly. In the playground story, enable `hideLabel` and test voice-over again. Ensure the correct labels are shown.


jnm2377 (approved):
VO is working for me for both stories 👍🏽

abbeyhrt (approved):
🎉

PR #12095, opened 2022-09-14
fix(Icons): update incremental-migration example to use v11 icons

tw15egan:
Fixes an issue where the icon build process was resolving to the v10 icon package.

#### Changelog

**Changed**

- Updated the `incremental-migration` example project to use the v11 icon package

#### Testing / Reviewing

Ensure the `incremental-migration` example still renders as expected


tw15egan:
cc @abbeyhrt just wanted to make sure this is still valid from the example standpoint

PR #12098, opened 2022-09-14
fix(RadioButton): fix HCM styles

tw15egan:
Fixes an issue with `RadioButton` when using Windows HCM

#### Changelog

**New**

- Added some styles so that the RadioButton is visible when selected 

#### Testing / Reviewing

Activate Windows HCM and ensure the RadioButton is visible in all states


abbeyhrt (approved):
Looks good to me!

aledavila (approved):

PR #12105, opened 2022-09-15
fix(Filename): remove tabindex when status is complete

### LINKED ISSUE #12101: [a11y]: File uploader item icon is tabbable & has incorrect role.
### Package

@carbon/react

### Browser

Chrome, Firefox

### Operating System

Windows, MacOS

### Package version

1.12.0

### React version

17/18

IBM equal access accessibility checker - Latest Deployment

### Assistive technology

_No response_

### Description

The tabbable element's role 'img, graphics-document' is not a widget role. The checkmark--filled icon is tabbable (`tabindex="0"`) and the role assigned to the svg is img (`role="img"`).

This accessibility issue can be observed via [storybook](https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader-item&args=status:complete). Set status to `complete` & open the canvas in a new [tab (With only File uploader item story)](https://react.carbondesignsystem.com/iframe.html?args=status:complete&id=components-fileuploader--file-uploader-item&viewMode=story) & run IBM equal accessibility checker.

[Permalink to icon in code.](https://github.com/carbon-design-system/carbon/blob/70eefeee38cf2516ea742d7e70b82e8b4ffde4f4/packages/react/src/components/FileUploader/Filename.js#L36)

image

Have ran the checker on firefox & chrome only. 

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?args=status:complete&id=components-fileuploader--file-uploader-item&viewMode=story

### Steps to reproduce

* Set File uploader item status to `complete`
* Run checker

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12101

Removes the `tabindex` when the status is `complete`, as this is just a status icon, and no action can be taken by clicking the icon in this state

#### Changelog

**Changed**

- Removed the default tabIndex when the status is `complete`
- Removed cursor styles when the status is `complete`

#### Testing / Reviewing

Run a11y checker on the `Filename` example when the status is `complete`. No violations should appear

### DIFF
--- a/packages/react/src/components/FileUploader/Filename.js
+++ b/packages/react/src/components/FileUploader/Filename.js
@@ -36,7 +36,8 @@ function Filename({ iconDescription, status, invalid, ...rest }) {
         <CheckmarkFilled
           aria-label={iconDescription}
           className={`${prefix}--file-complete`}
-          {...rest}>
+          {...rest}
+          tabIndex={null}>
           {iconDescription && <title>{iconDescription}</title>}
         </CheckmarkFilled>
       );
--- a/packages/styles/scss/components/file-uploader/_file-uploader.scss
+++ b/packages/styles/scss/components/file-uploader/_file-uploader.scss
@@ -271,7 +271,6 @@
   }
 
   .#{$prefix}--file__state-container .#{$prefix}--file-complete {
-    cursor: pointer;
     fill: $interactive;
 
     &:focus {


jnm2377 (approved):

packages/react/src/components/FileUploader/Filename.js
```diff
@@ -36,7 +36,8 @@ function Filename({ iconDescription, status, invalid, ...rest }) {
         <CheckmarkFilled
           aria-label={iconDescription}
           className={`${prefix}--file-complete`}
-          {...rest}>
+          {...rest}
+          tabIndex={null}>
```
abbeyhrt, line 40: Out of curiosity, what would the difference, if any, be between having tabIndex equal null or tabIndex equal to -1?

abbeyhrt (approved):

packages/react/src/components/FileUploader/Filename.js
tw15egan: When a prop is set to `null` it shouldn't render that prop on the component at all. Since it's an SVG and they don't support `tabIndex`, I figured I'd just remove it in that case. Usability wise they accomplish the same thing!

tw15egan: I think the main use case of this would be a `disabled` prop where you would do something like `disabled={disabled ? true : null}`

PR #12109, opened 2022-09-16
fix(storybook): update each folder to have a "Default" story

tay1orjones:
This updates every storybook component folder to have a story named "Default" where possible. This is in support of the platform beginning to consume our stories.

#### Changelog

**Changed**

- update any component story file where ComponentName/ComponentName instead to be ComponentName/Default

#### Testing / Reviewing

For each component modified:
- Ensure the Default story renders
- Ensure there are no regressions on the docs tab


abbeyhrt (approved):
The ones here look good! I ran through the components and a couple are missing a "default". Mainly, I think CodeSnippet, OverflowMenu, DatePicker(?), FormLabel, and Slider. There are a couple of others like Notification or DataTable, but I'm not sure if those really have a "default"

francinelucca (approved):
LGTM! thanks @tay1orjones  🎉 
My only notes is `Error boundary` and `Unordered list` have a lower case default in the story name:
image
image

And maybe we can rename to Default for `Form Label` and `Slider`?
image
image

tay1orjones:
@francinelucca @abbeyhrt Thank you both for the catches! I forgot to look through some of the `*-story.js` files we still have laying around. Those should now be updated.

As was noted, there are some that don't fit the pattern of having a `Default`. Notification components are a good example and for now won't be refactored to have a `Default` story.

francinelucca:
looks great!

PR #12140, opened 2022-09-23
fix(Search): fix event passed to `onChange` when clearing input (#12133)

### LINKED ISSUE #12133: [Bug]: Search - clear button does not call `onChange` with proper event
### Package

@carbon/react

### Browser

Chrome

### Package version

1.12.0

### React version

18.2.0

### Description

When clicking the close button after inputting text and clicking outside the component to trigger a blur/defocus event, the component is not collapsed even though the text input is empty.

In the Search component, when the Close button is clicked, `clearInput` is called. The event passed to the handler is for the button click. 

When the Search component's `value` prop is undefined/null, then the `onChange` prop is called with the button click event.

When the Search component's `value` prop is defined, then the `onChange` prop is called with a modified event:
```js
{
  ...event.target,
  target: { value: '' },
}
```
This event is not constructed properly and should either be a simple mock event:
```js
{
  target: { value: '' }
}
```
or overwriting the value:
```js
{
  ...event,
  target: {
    ...event.target,
    value: '',
  },
}
```

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-search--expandable

### Steps to reproduce

https://react.carbondesignsystem.com/?path=/story/components-search--expandable

### Code of Conduct

gi:
* fix(Search): when clearing, pass a pseudo input change event instead of
  the mouse click event with which the internal event handler is called.

Closes #12133

When the close button is clicked, the internal event handler is called with the mouse click event.

Previously, the `onChange` prop was called with this event or a modified version. However, that event callback normally expects an input change event.

This fixes the issue by creating and sending a pseudo input change event to the `onChange` prop.

#### Changelog

**Changed**

- `Search`: `onChange` is called with proper data shape when clearing input

#### Testing / Reviewing

* Verified in Storybook
* Added unit tests.


gi:

packages/react/src/components/Search/next/Search-test.js
```diff
@@ -13,6 +13,23 @@ import { mount, shallow } from 'enzyme';
...
+  const render = (props) => {
+    if (wrapper) {
+      return wrapper.setProps(props);
+    }
+
+    wrapper = mount(<Search labelText="testlabel" {...props} />);
+
+    return wrapper;
+  };
+
```
tay1orjones, line 32: Should this be in a `beforeEach` or `beforeAll`?

tay1orjones (changes requested):
Thanks for noticing this and putting up a PR! This is a quite old snippet of code - 5+ years at this point. I think the original intent of this code was probably to replace the existing `event.target.value` with an empty string.

```diff
 function clearInput(event) {
   if (!value) {
     inputRef.current.value = '';
     onChange(event);
   } else {
-    const clearedEvt = Object.assign({}, event.target, {
+    const clearedEvt = Object.assign({}, event, {
       target: {
         value: '',
       },
     });
     onChange(clearedEvt);
   }

   onClear();
   setHasContent(false);
   focus(inputRef);
 }
```

This would be the smallest change to rectify the root issue and I think the least likely to introduce a breaking change. I'd prefer not to change the signature and call structure if possible, or at least as little as possible, as these could inadvertently break folks who have relied on this since v8/v9.

gi:
@tay1orjones The actual culprit is the first `if` block:
```js
   if (!value) {
     inputRef.current.value = '';
     onChange(event);
   }
```
This clears the input's value, but it sends the `click` event from the button, which does not contain the `target.value`.

Thus, I doubt any code would be properly relying on this because it doesn't work.

Additionally, the `else` clause was sending an "event" such that the event data was really the target button element and the "target" was this manufactured object:
```js
{
  target: {
    value: '',
  },
}
```

Given that in both cases, the result was not anything which resembled a proper event, I opted to change both.

In both cases from this `clearInput` handler, the `onChange` callback does not receive the same type of event or target which it otherwise receives from the normal `change` events when the user inputs text.

packages/react/src/components/Search/next/Search-test.js
gi: It could be a `beforeAll`, but that would only be doing the same thing as this but with less clarity: the functions would need to be declared as `let` variables and then assigned in the `before` routine.

This code solves for an issue that I see a lot with front-end tests and which I see with the tests in this suite: if you render the element under test in a `beforeEach` routine, then it makes any other setup in nested contexts very difficult. You end up with multiple renders of the element.

In fact, I believe that some of the tests in this suite are finding the wrong element rendered.

If you move the mounting or rendering into each test/`it`, then you avoid this issue. 

To avoid duplicating the possibly complex mount/render declaration, I have adopted this paradigm which works quite well in my experience.

Each nested context/describe can build on the props:
* default: props reset
  * with size large: props.large = true
  * with size small: props.small = true
* with onChange: props.onChange = jest.fn();
* etc.

Each test simply becomes
```js
render(props);
expect(...);
```

tay1orjones (approved):
That makes sense, thanks!

abbeyhrt (approved):

PR #12143, opened 2022-09-23
fix(ExpandableSearch): compose prop handlers with internal (#12135)

### LINKED ISSUE #12135: [Bug]: ExpandableSearch - onChange is not composed with internal handler
### Package

@carbon/react

### Browser

Chrome

### Package version

1.12.0

### React version

18.2.0

### Description

The ExpandableSearch component defines and passes the `onChange` callback to the Search component for tracking its internal state.

This means that parent components have no way of specifying an `onChange` callback.

This component seems otherwise useless without the ability to use an `onChange` callback to get the current value.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-search--expandable

### Steps to reproduce

https://react.carbondesignsystem.com/?path=/story/components-search--expandable

### Code of Conduct

gi:
* fix(ExpandableSearch): compose prop handlers with internal

Closes #12135

The `ExpandableSearch` component passes a few callbacks from the underlying `Search` component in order to control its own state.
* `onBlur`
*  `onChange`
*  `onExpand`
*  `onFocus`

These are not composed with any props on the `ExpandableSearch` component, meaning that any of those props specified will have no effect.

This fixes the issue.

#### Changelog

**Changed**

- `ExpandableSearch`: compose event callbacks with internal handlers

#### Testing / Reviewing

* Verified in Storybook
* Added unit tests.


tw15egan:
Hey @gi, thanks for the contribution and adding all the tests! I'm running into an issue where it seems like the `ExpandableSearch` no longer opens on click. Are you seeing this same behavior?

gi:
@tw15egan Thanks for catching that. The `composeEventHandlers` function was throwing an error because no event was being passed to the `onExpand` call in `Search`. This is fixed now.

tw15egan (approved):
Working great, LGTM 👍 ✅

Thanks for contributing 🎉

abbeyhrt (approved):

Issue #5871, opened 2020-04-16
UI shell left nav: Provide a mechanism to expand width panel + section for additional links


tifftea:
### Summary

1. There should be a toggle or some other mechanism to expand or reduce the width size of the UI shell left panel. 

2. We propose to have a section at the bottom of the panel where additional icons and links that are less relevant to users are placed.

Example:
![Navigation-info](https://user-images.githubusercontent.com/63752979/79409182-1a174580-7f52-11ea-9d09-2701f416f9b2.png)

_(In this example, that toggle is shown as a hamburger menu. It is understood that a hamburger icon is not the best icon to use as it is already being used in the UI shell header.)_  

### Justification

A hamburger button was added to sit right above the navigation. Clicking on the hamburger icon expands the left panel and reveals the full menu. This panel is unique to WOC and should not be interfered with CPD UI shell.

Additional icons that sit at the bottom of the panel includes: help, documentation, API, and account info. Because these links are most irrelevant to users, it is necessary to distance them.

The Db2 Warehouse on Cloud ("WoC") navigation strategy was guided by user insights and human factor cognitive psychology in order to  assure the following goals:
- simplify and streamline users' application workflows
- scalability  
- an integration into an enterprise platform 

The hamburger icon denotes the key interaction point to reveal the nomenclature of the primary navigation to reinforce users’ cognition on the information hierarchy and structure. This functional hamburger button instantly guides users to the primary category and ties the relationship between the primary and the secondary navigations.

 
When the hamburger menu is clicked, the navigation bar slides out and displays the primary navigation menu on top and tertiary navigation menu towards the bottom. 

 
The tertiary navigation menu disassociates from the primary menu by a great distance and a line to clearly disclose the organizational hierarchy between two menus. 

 
WoC is a highly functional SaaS solution with streamlined workflows, and that, the primary navigation enables users to instantly access the secondary menu by directly clicking on the primary navigation icons. 

### Desired UX and success metrics

-A more define hierarchy and easier navigation

### "Must have" functionality

--

### Specific timeline issues / requests

--

### Available extra resources

--

asudoh:
Hi 👋 thank you for reporting - One quick question. Would you have a summary of the difference between your to-be scenario and https://codesandbox.io/s/header-with-sidenav-1u7su?from-embed?

tifftea:
@asudoh 
The current design has two states:
1. open at full width
2. closed

The proposed design has three states:
1. open at full width (text + icon)
[...]

PR #12152, opened 2022-09-26
fix(style/`--side-nav__navigation`): remove css `height` to allow components at bottom (#5871)

gi:
* fix(style/`--side-nav__navigation`): remove css height to allow components at bottom (#5871)

This removes the CSS height style from `--side-nav__navigation`. The resulting style for this element when composed with `--side-nav` is:
```css
position: fixed;
top: #{header-height};
bottom: 0;
```

With fixed positioning and top and bottom specified, the CSS `height` is unnecessary. 

Additionally, the CSS `height: 100%` calculates the height from the body, which is more than the viewport height. Thus, the height of the element will extend beyond the viewport, 'below the fold".

The element `--side-nav__items`, included as a child element, has `flex-grow` set to `1`, causing it to fill the entire height. 

Any element included after the side nav items in the side nav are pushed to the bottom. They should be expected to show entirely within the viewport, yet since the bottom is below, they either are not shown or are chopped off.

This fixes this issue.

Note: This may or may not solve #5871. It is not intended as a full solution, but it does fix this styling issue which would prevent properly supporting such a feature.

#### Changelog

**Removed**

- style/`--side-nav__navigation`: height removed

#### Testing / Reviewing

* Tested visually


tw15egan (approved):
LGTM, thanks for contributing another fix 👍 ✅

aledavila (approved):

PR #12167, opened 2022-09-27
fix(numberinput): correct stepper behavior, typecast internal value to number

### LINKED ISSUE #12165: [Bug]: NumberInput Steppers are behaving odd
### Package

@carbon/react

### Browser

Safari

### Package version

1.13.0

### React version

18.2.0

### Description

When using steppers on Number props. Sometimes I get string addition ('3' + 1 = 31) when I hit plus.  Sometimes I get it returning the MAX or NaN.

Going down seems fine.

This also seemed like a breaking change. Prior we used event to get the data even for step clicks.  But now I don't see value coming back on the stepper when the component is first initiated (it works after second try. But first one (like if set to 5; pressing sets to empty no usable event data).  Because of this moved to second param to get the value from it.

android2

For now in house we are hiding steppers due to this bug.

To note; We are only using controlled components.

### Reproduction/example

https://codesandbox.io/s/4kmjci?file=/index.js

### Steps to reproduce

- Type in number `4`
- Press plus Get `41`. Expect 5

Weirder results with more complex usages.  See GIF above.  But most likely fix basic and will be good.

### Code of Conduct

tay1orjones:
Closes #12165

#### Changelog

**Changed**

- use `parseInt` to ensure internal value is always a number

#### Testing / Reviewing

- Check out the `Testing` story
  1. Enter `44` in the input via keyboard
  2. press the increment button, it should increment as expected and not go to the max value or NaN

~🛑 Don't merge until the `Testing` story is removed~


tw15egan (approved):
Bug is fixed 👍🏻  LGTM once test and `console` statements are removed

aledavila (approved):
works as expected

abbeyhrt (approved):

PR #12173, opened 2022-09-27
test(Layer): add layer backgrounds to layer class

tw15egan:
Noticed that the `field` token may not be correct when working on adding `Layer` backgrounds with @aledavila 

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/packages/react/.storybook/styles.scss
+++ b/packages/react/.storybook/styles.scss
@@ -28,3 +28,10 @@ body {
   background: styles.$background;
   color: styles.$text-primary;
 }
+
+.cds--layer-two,
+.cds--layer-three {
+  padding: 1rem;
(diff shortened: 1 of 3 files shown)


tw15egan:
cc @aagonzales I put up a PR here so you can visualize the layer tokens... It seems like `layer` and `field` are set to the same color. Wondering if this is am implementation issue (we need to move the `Layer` component around in the stories) or if the field should be the same color as the `layer` background

Layer 2

Screen Shot 2022-09-27 at 10 06 38 AM

Layer 3
Screen Shot 2022-09-27 at 10 05 57 AM

PR #12185, opened 2022-09-28
fix(styles/ui-shell/header): fix header menu z-index (#12184)

### LINKED ISSUE #12184: [Bug]: UI Shell - header menu z-index renders behind side-nav
### Package

@carbon/styles

### Browser

Chrome

### Package version

1.13.0

### React version

17.0.2

### Description

A header menu when opened renders behind the left side-nav.

The header and side-nav both have a z-index of 8000.

The header menu, when expanded, is defined to be the same as the header. However, this causes the menu to render behind the side-nav.

This is related to #6662, but the problem there is an overflow menu within the side-nav.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-ui-shell--header-base-w-navigation-actions-and-side-nav

### Steps to reproduce

1. Ensure that the width shows the header nav.
2. Move the "Link 4" header menu into the first position of the header menu bar/items. You can drag the element in the developer tools panel in Chrome.
3. Click the menu to expand it.

https://user-images.githubusercontent.com/7407551/192696107-358ae509-20ea-4ba8-9d8c-6ac16bb0ff36.mov

### Code of Conduct

gi:
* fix(styles/ui-shell/header): update z-index to be above the side-nav

Closes #12184

This updates the ui shell header menu z-index to be +1 than the header and thus the side-nav.

The header title is also bumped up one due to a comment in the file:
> needs to be higher than menu

#### Changelog

**Changed**

- styles/ui-shell/header: header menu z-index increased by 1

#### Testing / Reviewing

* Manual


tay1orjones (approved):
Looks fine to me 👍 I want to be sure we get a review from design on this one though. I agree the header dropdowns should not be behind the sidenav, but want to be sure.

gi:
Alternatively, the side-nav z-index could be decreased by 1:

https://github.com/carbon-design-system/carbon/blob/797157dcc1f11cd903c9965a8320d2a7f8c07dfe/packages/styles/scss/components/ui-shell/side-nav/_side-nav.scss#L31

aagonzales (approved):
I think this all seems good

gi:
Perhaps it would make more sense to decrease the z-index of the side nav instead. That seems more “correct”. 

Keep as is or change?

gi:
@aagonzales @tay1orjones 
#12208 from @TannerS looks like a better solution.

jnm2377 (approved):

PR #12192, opened 2022-09-28
fix: add missing pages to the carbon react index

mattrosno:
As found in https://github.com/carbon-design-system/carbon-platform/issues/907, this PR fixes some Carbon React indexing issues.

#### Changelog

**New**

Carbon React:

- Get started page
- Tutorial FAQ
- Ordered list and Unordered list docs (this will eventually get cleaned up in https://github.com/carbon-design-system/carbon-platform/issues/1374)

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

See comments below.


mattrosno:
Usage, Style, and Accessibility page tabs now show up for `Ordered list` and `Unordered list` React components. For now, the page headers still say `List` here because the remote MDX source says `List`:
image

It's a known issue that there is a large gap above the Color heading here:
image

Other Accessibility page tabs don't render as well because of invalid imports. We're okay with this as it's a public beta. If we want, we should remove the imports in carbon-website. For example: https://next.carbondesignsystem.com/libraries/carbon-react/latest/assets/code-snippet/accessibility
image

mattrosno:
React tutorial FAQ page:
image

mattrosno:
Carbon React get started content:
image

alisonjoseph (approved):

tw15egan (approved):

francinelucca (changes requested):
Confirmed can render library pages and see asset tabs, cannot render accessibilityPath for either component
image

francinelucca:
why do we want to index a tab that won't render? 🤔 wouldn't it be better to hold off on it for now if we can't fix it? or keep it in but commented out

francinelucca (approved):

PR #12207, opened 2022-10-01
fix(react): added condition check before calling onblur

### LINKED ISSUE #12187: [Bug]: Slider, onRelease and onBlur
### Package

@carbon/react

### Browser

Chrome

### Package version

1.13.0

### React version

18.2.0

### Description

When typing in the TextInput attached to a Slider, each character typed will trigger the onRelease function.
Exiting the same TextInput will result in a error logged in the browser console:
```
Uncaught TypeError: _this.props.onBlur is not a function
    at @carbon_react.js?v=0c3423c8:65038:19
    at HTMLUnknownElement.callCallback2 (chunk-HSVJWAOT.js?v=7bb9af7f:3793:22)
    at Object.invokeGuardedCallbackDev (chunk-HSVJWAOT.js?v=7bb9af7f:3818:24)
    at invokeGuardedCallback (chunk-HSVJWAOT.js?v=7bb9af7f:3852:39)
    at invokeGuardedCallbackAndCatchFirstError (chunk-HSVJWAOT.js?v=7bb9af7f:3855:33)
    at executeDispatch (chunk-HSVJWAOT.js?v=7bb9af7f:6484:11)
    at processDispatchQueueItemsInOrder (chunk-HSVJWAOT.js?v=7bb9af7f:6504:15)
    at processDispatchQueue (chunk-HSVJWAOT.js?v=7bb9af7f:6513:13)
    at dispatchEventsForPlugins (chunk-HSVJWAOT.js?v=7bb9af7f:6521:11)
    at chunk-HSVJWAOT.js?v=7bb9af7f:6645:20
```

### Reproduction/example

https://stackblitz.com/edit/github-h8unbj?file=src%2FApp.jsx

### Steps to reproduce

If you use the Slider, onChange and onRelease values should update as expected.
While typing in the TextInput, you would notice that both onChange and onRelease values will be edited live.
Exiting the TextInput should trigger an error in the console.

### Code of Conduct

TannerS:
REF https://github.com/carbon-design-system/carbon/issues/12187

My attempt to fix this bug

#### Changelog

**Changed**

- Added condition to check if props.onblur is a function before invoking it

#### Testing / Reviewing

Verified in storybook


packages/react/src/components/Slider/Slider.js
```diff
@@ -445,7 +445,10 @@ export default class Slider extends PureComponent {
     const { value } = evt.target;
 
     this.setState({ isValid: validity });
-    this.props.onBlur({ value });
+
```
abbeyhrt: Since `onBlur` is a function in proptypes, we don't need to check whether or not it is a function here, just whether or not it's defined

```suggestion
if (this.props.onBlur) {
  this.props.onBlur({ value });
}
```

abbeyhrt (dismissed):
Thank you for taking this on! Just made a small suggestion to change the specificity of the `if` to be more in line to how we do it in other components with event handlers

tay1orjones (approved):
Hope you don't mind, I pushed a small update here based on @abbeyhrt's feedback. LGTM!

tw15egan (approved):

Issue #10751, opened 2022-02-14
Ensure playground stories have working controls


tay1orjones:
Every component in the storybook sidebar should have a Playground story, those stories should have _operable_ controls.

High level checklist for each component:

- Move `.stories.js` file out of `next` folder, or create `.stories.js` file and copy contents of `*-story.js` file over.
- Remove deprecated `knobs` package and usage.
- Ensure there is a playground story
- Add `argtypes` to playground story
- Pass `...args` to playground story, spread them on the applicable component(s)
- Make sure values are defined to enable controls view by default 
- https://github.com/carbon-design-system/carbon/issues/11249

FAQ:
- What if there's no playground story but other stories have controls enabled on them?
  - Ignore these components for now, we need to prioritize getting controls working for components that have nothing right now.

**Checklist**

- [x] #11467
- [x] #12351
- [x] #11468
- [x] #11463
- [x] #11469
- [x] #11471
- [x] #11472
- [x] https://github.com/carbon-design-system/carbon/issues/11894
- [x] #11564
- [x] #11719
- [x] #11565
- [x] #11716
- [x] https://github.com/carbon-design-system/carbon/issues/11954
- [x] #11717
  - [x] components/Dropdown/next/Dropdown.stories.js
- [x] #12161
- [x] #12224
- [x] components/FluidForm/FluidForm-story.js
- [x] components/Form/next/Form.stories.js
 - [x] #12339
- [x] #12250
- [x] #12248
- [x] #12249
- [x] ~components/Heading/Heading-story.js~
- [x] #12584
- [x] #12352
- [x] #12255
- [x] components/Layer/next/Layer.stories.js
- [x] #11464
- [x] #12162
- [x] components/Modal/next/Modal.stories.js
- [x] #11718
- [x] #11562
- [x] #11823
  - [x] components/NumberInput/next/NumberInput.stories.js
- [x] #12145
  - [x] #12146
- [x] #12226
- [x] #12271
- [x] #12272
- [x] #12273
- [x] components/Popover/next/Popover.stories.js
- [x] components/ProgressBar/ProgressBar-story.js
  - [x] components/ProgressBar/next/ProgressBar.stories.js
- [x] #12227
- [x] components/RadioButton/RadioButton-story.js
  - [x] components/RadioButton/next/RadioButton.stories.js
- [x] #12353
- [x] components/Select/Select-story.js
  - [x] #11466
- [x] #12448
[...]

PR #12209, opened 2022-10-01
feat(react): first round of playground updates

TannerS:
Closes 

Hopefully multiple sub issues in https://github.com/carbon-design-system/carbon/issues/10751

Side note: I had questions about these issues which i asked in a comment in https://github.com/carbon-design-system/carbon/issues/10751 which may affect/cause me to have to do changes in this pr. I have been having issues understanding what to keep, what to add and how to add it. So please take a look at those questions and let me know what I can fix/change  in this PR

#### Changelog

**New**

- Added some new playgrounds for some components

**Changed**

- Change some new playgrounds for some components and text and id fixes

#### Testing / Reviewing

- Reviewed in storybook


abbeyhrt:
Thanks for doing this @TannerS, this is amazing and so helpful for us! It looks like we have an open PR for loading storybook controls, do you remind removing that one to avoid conflicts?

TannerS:
@abbeyhrt can you point me in direction of what your referring too? 

Also anyway y'all can check on my questions :)

packages/react/src/components/FileUploader/next/FileUploader.stories.js
```diff
@@ -20,6 +20,44 @@ import {
 import './FileUploader-story.scss';
 
 const filenameStatuses = ['edit', 'complete', 'uploading'];
+const argTypes = {
```
abbeyhrt, line 23: FileUploader is a bit tricky since it has such robust examples but I think we would still want it to follow the pattern of the other components in terms of controls, where only the "Playground" story has controls. To do so, if you don't already know, we would want to remove the passed in `args` from every story besides that one and only define the argtypes for Playground as well.

abbeyhrt (commented):
This looks great! I only have one comment about the FileUploader story and loading but everything else is great!!

aledavila (changes requested):
Hey tanner I actually have a PR with the correct playground story. We are cleaning up the stories to where only Playground stories will have the controls and Scott has a Loading PR already up for that as well. Also could you change the `-story.js` files to be `.Stories.js` so that they all match and are updated to the file type. And the files must be removed from the next folder and into the root Component file replacing the necessary files. the index files for those components will also need to be updated accordingly so that the Feature flag is removed since it is no longer necessary.

packages/react/src/components/FileUploader/next/FileUploader.stories.js
TannerS: @abbeyhrt  can you explain this a bit more `from every story besides that one and only define the argtypes for Playground as well.`

TannerS:
Closing to start over with new instructions and to prevent overlap with Loading story and file uploader story

PR #12219, opened 2022-10-03
docs(template): added user story template

sstrubberg:
Adding a user story template. Going to use it when I create epics and see how it feels.

### DIFF
--- a/.github/ISSUE_TEMPLATE/USER_STORY.yaml
+++ b/.github/ISSUE_TEMPLATE/USER_STORY.yaml
@@ -1,0 +1,56 @@
+name: (Carbon core team ONLY) User story
+description:
+  Write a user story to begin solving their needs.
+title: '[YOUR TITLE]: Brief description'
+body:
+  - type: markdown
+    attributes:
+      value: "Avoid any type of solutions in this user story."
+  - type: markdown
+    attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/USER_STORY.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - Each product backlog item or user story should have at least one Acceptance criteria.
+
+      - Acceptance criteria defines a deliverable that can be completed in a single sprint.
+
+      - Each Acceptance criterion is independently testable.
+
+      - Include functional as well as non-functional criteria – when relevant.
+
+      - Team members write Acceptance criteria and the Product Owner verifies it."
+  - type: textarea
```
tw15egan, line 22: Would it make sense to break this into 3 different text areas? 

- `As a [user role]`
- `I need to`
- `so that I can`

Not sure if GitHub would format it nicely, but just an idea

sstrubberg: image

Here's an example of what i'm shooting for. I like that the different text areas are all contained with in the same text box. Makes it easier to be more malleable with the story.

tw15egan (approved):
LGTM 👍 ✅

abbeyhrt (approved):

PR #12221, opened 2022-10-04
fix(Modal): scope tooltip, overflow popup styles to inside modal

### LINKED ISSUE #12205: Button hasIconOnly bug with Modal
There is a bug when using the following Carbon components: Button, DataTable, and Modal. Loading a Button that utilizes the hasIconOnly property into the DataTable seems to be what creates the bug. When the Modal is opened in this configuration, the Button in the DataTable row that utilizes the hasIconOnly is still clickable and has a higher z index than the Modal and the Modal overlay. If the still accessible Button opens a different Modal with its onClick event, opening and then closing the other Modal, will cause the Buttons with the hasIconOnly property to then operate as expected (becoming inaccessible and having a lower z index than the Modal). When the Button is changed to hasIconOnly={false}, the Button then responds as expected (becomes inaccessible and has a lower z index than the Modal). Another thing that causes the Buttons to operate as expected is to change the DataTable property of stickyHeader={false} to stickyHeader={true}. For some reason, when the DataTable is set to having a sticky header, the Buttons loaded into the DataTable rows operate as expected when the Modal opens.
Has_Icon_Only_False_With_Sticky_Header
Has_Icon_Only_No_Sticky_Header
Has_Icon_Only_With_Sticky_Header
HigherZindex

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12205

Scopes `Modal` styles so that they do not leak out to elements not contained within the `Modal`

#### Changelog

**Changed**

- `added `.cds--modal` to the selector so that only `Tooltip` and `OverflowMenu` components _inside_ the `Modal` are effected. 

#### Testing / Reviewing

Added a test story to the `Default` `Modal` story. The story will be removed once approved.


abbeyhrt (approved):
Looks good to me!

tay1orjones (commented):
This breakpoint is being goofy and part of the top of the tooltip is cut off. I'm not sure why though 🤔 


https://user-images.githubusercontent.com/3360588/194091997-23dd2827-1ce0-445e-b9b5-4dc351546100.mov

tay1orjones:
Ah, yeah. I suppose the most clear-cut fix would be to change the tooltip alignment to be bottom? Not a blocker here though imo

tay1orjones (approved):

PR #12231, opened 2022-10-04
fix(text-input): add class for inline validation

### LINKED ISSUE #10234: [Question]: Should TextInput inline invalidMessage appear somewhere in the DOM?
### Question for Carbon

The TextInput when used with `inline`, `invalidMessage` and `Invalid` does not place the invalid message anywhere in the DOM. Is this correct it? The internal design for https://pages.github.ibm.com/cdai-design/pal/patterns/edit#inline-edit shows this below the text, another possibility would be to surface this data fro the error/invalid icon.

### Code of Conduct

aledavila:
Closes #10234

Fix validation on inline variant

#### Changelog

**New**

- added new class to target inline validation

#### Testing / Reviewing

Make sure validation props still work. Make sure validation props inline are showing.


packages/react/src/components/TextInput/TextInput.js
```diff
@@ -113,6 +113,8 @@ const TextInput = React.forwardRef(function TextInput(
       [`${prefix}--text-input-wrapper--readonly`]: readOnly,
       [`${prefix}--text-input-wrapper--light`]: light,
       [`${prefix}--text-input-wrapper--inline`]: inline,
+      [`${prefix}--text-input-wrapper--invalid`]:
```
tw15egan: If this only applies to the `inline` variant, can we rename the class to `${prefix}--text-input-wrapper--inline--invalid`?

aledavila: yes! will do

tw15egan (approved):

thyhmdo:
Small ask @aledavila  but can we change that text from "Invalid text" to "Error message goes here"? I think we'll try to be more consistent with our guidance languages and since we have this issue open so I just wanted to add that.

aledavila:
@thyhmdo updated now

jnm2377 (approved):

thyhmdo (approved):
thanks @aledavila looks good

Issue #12125, opened 2022-09-20
Fluid inputs: FluidDatePicker


tay1orjones:
This issue is for the primary implementation and design of the FluidDatePicker component.

**Checklist**

- [x] https://github.com/carbon-design-system/carbon-website/issues/3190
- [x] https://github.com/carbon-design-system/carbon-design-kit/issues/579
- [x] #12403
- [x] #12404

## Spec

[Sketch file]() | [Figma file]()

![DatePicker - White](https://user-images.githubusercontent.com/11670886/194885938-ab220f17-5162-4114-8e04-541ccc8973e7.png)

tw15egan:
@aagonzales do you have a spec image for this component?

PR #12247, opened 2022-10-05
feat(FluidDatePicker): implement `FluidDatePicker`

### LINKED ISSUE #12403: Fluid inputs: Add FluidDatePicker experimental implementation
- [x] Place implementation within a new component, `FluidComponentName`
  - [x] Add VRT tests via storybook story e2e setup
  - [x] Add unit tests (likely simple, eg. "does the classname get added")
  - [x] Add storybook example
  - [x] Add it to the FluidForm story
  - [x] Ensure styles are properly scoped to their own file
  - [ ] Add `component: FluidDatePicker ` label to repo for triaging issues

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/12125
Closes https://github.com/carbon-design-system/carbon/issues/12403

Adds in `unstable__FluidDatePicker` and `unstable__FluidDatePickerInput`

#### Changelog

**New**

- `unstable__FluidDatePicker` component
- Styles for `FluidDatePicker` (`_fluid-date-picker.scss`)
- Stories for `FluidDatePicker` (`FluidDatePicker.stories.js`)
- Tests for `FluidDatePicker`

**Changed**

- Fixed some styles I noticed when adding in `FluidDatePicker` to the `FluidForm` story

#### Testing / Reviewing

Go to `FluidDatePicker` and check out the stories. For now, I've got all the different states in (they will be removed before merging). Ensure all playground controls are correct and stories render as expected.


aagonzales:
@tw15egan yeah we should probably allow just triggering invalid on one side ... I'm not yet sure how that will look visually. We might have to come back to that.

Also kind of on the same thread, I forget to spec out `timePicker` in fluid and will have a sam joint input problem.

tw15egan:
@aagonzales, so would the invalid outline span the entire input, but just keep invalid text to one input (as-is) for now? Since it's experimental we can come back and refine this interaction as well

aagonzales:
@tw15egan I'm going to talk about this in design crit tomorrow and hopefully get a final answer. The bottom+left is how in forms, we originally said we would treat inline fluid inputs with one invalid state. 

image

aagonzales:
@tw15egan This is what we decided, is it doable?

image

tw15egan:
Yeah, we should be able to do that; we'll just need to add a prop of `invalid` to the `DatePicker` in addition to the `DatePickerInput`. Would this also be a valid use case outside of Fluid components? i.e., the entire `DatePicker` is invalid? If so, we may want to add that to the normal `DatePicker` as well

tw15egan:
I believe AVT tests are failing due to the duplicate ids on the stories showcasing all the different warning/error variants. Before merging, these will all be removed and set to a single story

aagonzales (commented):
The componet styles all look correct but there are some bugs in the Playground.
- I'm getting some weird bugs when I switch between Range and Single.
   - Switch from single back to range and I can't get the end date calendar menu to open
   - Switch from range (with a range selected) to single, the range is still showing when I open the menu.
- When invalid is turned on the width of the inputs grow. The error message should be wrapping and not effecting the width. (Same for warning)
Oct-28-2022 13-49-45

tw15egan:
That switcher is a bit temperamental, so I switched it to show two components. Also, I set a width on the container, so the component expands downwards now.

packages/react/src/components/FluidDatePicker/FluidDatePicker.stories.js
```diff
@@ -0,0 +1,220 @@
...
+      disable: true,
+    },
+  },
+  locale: {
+    table: {
+      disable: true,
+    },
+  },
+  onChange: {
+    action: 'clicked',
```
tay1orjones, line 167:
```suggestion
    action: 'clicked',
    table: {
      disable: true,
    },
```

packages/react/src/components/FluidDatePicker/FluidDatePicker.stories.js
```diff
@@ -0,0 +1,220 @@
...
+  },
+  onClose: {
+    action: 'clicked',
```
tay1orjones, line 173:
```suggestion
    action: 'clicked',
    table: {
      disable: true,
    },
```

packages/react/src/components/FluidDatePicker/FluidDatePicker.stories.js
```diff
@@ -0,0 +1,220 @@
...
+  },
+  onOpen: {
+    action: 'clicked',
```
tay1orjones, line 179:
```suggestion
    action: 'clicked',
    table: {
      disable: true,
    },
```

tay1orjones (approved):
Looks great to me! 🎉 Just some minor suggestions for the playground story

aagonzales (approved):
Looks good to go! 👍 ✅

alisonjoseph (approved):

PR #12261, opened 2022-10-07
fix(FluidTextInput): label overflow scroll

### LINKED ISSUE #12082: Fluid inputs: Label content overflow
Currently, Fluid Input labels wrap to a second line and start running into the input text inside the field. Can we change the label text overflow in the same manner as the input text (which triggers a horizontal scroll). 

Example of what the label currently is doing:
image

Example of input texts overflow behavior (new intended behavior for label)
Sep-09-2022 15-27-02

jnm2377:
Closes #12082 

#### Changelog

- input label should now overflow on scroll

#### Testing / Reviewing

- go to the playground story for fluid text input, update the label text to have a really long label that should overflow, check that the label scrolls instead of wraps


aledavila (approved):

aagonzales:
The scroll bar covers up the label some times and it does fade away eventually. I guess there's probably nothing we can do about that though.

image

aagonzales:
Actually, the field text doesn't have the scroll bar either. Is there a way to fix this in the label then?

Oct-11-2022 10-32-38

tay1orjones:
@aagonzales Sometimes the scroll bar doesn't fade away depending on the user settings

image

image


I suspect we might see similar problems with the scroll bars in windows.

The textbox scrolling is something the browser does by default. It's possible to hide the scroll bar on the label, but WCAG recommends against it citing usability and accessibility concerns.

It looks like there's a new [draft spec for `scrollbar-width: thin`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width) that would make the scroll bar thinner for cases like this, but it's not available in browsers yet.

All said, I think hiding the scrollbar in this case could be the most beneficial option.

aagonzales:
The good thing is, either way, I think label scrolling is a very side edge-case. If we can remove it without causing violations, then lets do it but if we have to keep it then its fine. @jnm2377 @tay1orjones

tay1orjones:
@mbgower do you think in this instance it's acceptable to remove the scrollbar?

mbgower:
The primary requirement for WCAG is covered under [Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html).
Basically, the system needs to be able to zoom up to 400% _without_ requiring horizontal scrollbars to make content visible.

So it really depends on how you're handling breakpoints and text resizing. IMO, a text input area is likely an acceptable place to use horizontal scrolling without violating this because the input width is constrained by design -- although I assume Carbon has some guidance saying that where the character count allows more characters than can be displayed, use a text area!

I know Carbon does have guidance saying input labels should be short, so I think your design guidance is making this outcome very unlikely.

That said, there are a couple of things to note...
Unless the input itself is short and only the label is long, I do not actually think it's going to matter for a keyboard user, because when I enter the input itself, I can enter and edit text, and read it, but just using my cursor keys or home/end to read the content. No "scrolling" is needed. I simply move around in the text.

The place where you create a potential issue if you don't offer scrolling is actually the mouse user. But I believe the interaction workaround would be that they too would click in the input and use arrow keys, or the OS affordance for panning within text, to read the information. If that's acceptable to you folks, I don't think it violates accessibility (which tends to be worried about keyboard more than mouse).

So that only leaves the label as a potential problem. I think I was reading above that the label will line break, rather than scroll? So I'm not sure I understand a scenario where you'd _need_ scrollbars.

mbgower:
I thought up an example... Use of read-only inputs. In such a situation, if you are constraining the horizontal width, you could potentially be in a situation where the user would be unable to see both a very long label and any input text that exceeds the horizontal spacing of the input. @quarryboy, have you considered this?

quarryboy:
@mbgower would the label not be seen in that instance? i thought it was the opposite of disabled and would be seen/processed for read-only.

i think it's a bit concerning that the label would be more than a few words long. what's the actual use case? it seems like the emphasis should be on putting all that text into a tooltip instead.

@aagonzales out of curiosity, how does this test with browsers that use a two-finger horizontal scroll in conjunction with it's history? e.g.—two-fingers to the left causes you to go "back". seems like it would be very frustrating to accidentally go back in history and have to start the entire form process over again. i saw somethign about violations above? how do violations fit in with this particular use case?

aagonzales:
@mbgower What we're really asking is can we hide the scroll bar because the scrollbar when active is covering up the label content.

See https://github.com/carbon-design-system/carbon/pull/12261#issuecomment-1274890371

mbgower:
@aagonzales it's a question of balancing considerations in the edge case where the label is long. On the one hand, the scrollbar somewhat obscures the label. But in the example you show, it is only obscuring while the user is actively employing it. So, there is a way to see the text. Whereas, if you remove the scrollbars entirely, you seem to be removing a users ability to reposition the label to even see it?
If that is a correct summary, then I think it is less risky to leave the scrollbars in place. However, if I'm wrong about removing the scrollbars removing functionality -- if it merely removes the visual cue for scrolling but still allows a user to do it -- then I think hiding the scrollbars is fine.

tay1orjones:
> if it merely removes the visual cue for scrolling but still allows a user to do it -- then I think hiding the scrollbars is fine.

Yes, we'll only be hiding the visual scrollbar. The ability to scroll will be retained.

tay1orjones (approved):
The scrollbar on the label is now hidden - the scrolling interaction between both the label and the input is consistent


https://user-images.githubusercontent.com/3360588/197269206-63bf9bdb-360a-4440-ab84-932b48910c61.mov

PR #12263, opened 2022-10-07
chore(ModalWrapper): deprecate for ComposedModal

### LINKED ISSUE #12118: ModalWrapper deprecation
We export it but recommend not to use it, so should probably find a way to deprecate or figure out what to do with this component

- We recommend using ComposedModal instead

**Checklist**
- [x] Identify the differences between the this and ComposedModal
- [x] Write migration guidance
- [x] Figure out deprecation strategy, messaging
  - [x] Add the deprecation section back to story, add ComposedModal to this section
  - [x] Add a deprecation message to component .mdx docs

jnm2377:
Closes #12118 

#### Changelog

- adds deprecated section to storybook
- adds deprecation warning to modal wrapper
- adds migration doc for modal wrapper

#### Testing / Reviewing

- ensure 
    - all tests pass
    - modal wrapper is in the deprecated section 
    - migration doc makes sense

### DIFF
--- a/packages/react/src/components/ModalWrapper/ModalWrapper.js
+++ b/packages/react/src/components/ModalWrapper/ModalWrapper.js
@@ -18,7 +18,7 @@ export default class ModalWrapper extends React.Component {
   if(__DEV__) {
     warning(
       didWarnAboutDeprecation,
-      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in v12 of `@carbon/react`'
+      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in the next major version, `@carbon/react@v2.x`'
     );
     didWarnAboutDeprecation = true;
   }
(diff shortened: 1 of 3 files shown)


packages/react/src/components/ModalWrapper/ModalWrapper.js
```diff
@@ -10,8 +10,19 @@ import React from 'react';
...
 import { ButtonKinds } from '../../prop-types/types';
+import { warning } from '../../internal/warning';
+
+let didWarnAboutDeprecation = false;
 
 export default class ModalWrapper extends React.Component {
+  if(__DEV__) {
+    warning(
+      didWarnAboutDeprecation,
+      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in v12 of `@carbon/react`'
```
tay1orjones: Since this is a message to devs, should we clarify the version number to match the real version number of the package?

```suggestion
      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in the next major version, `@carbon/react@v2.x`'
```

tay1orjones (approved):
One small suggested change but otherwise looks good!

packages/react/src/components/ModalWrapper/ModalWrapper.js
jnm2377: that makes sense 👍🏽

aledavila (approved):

PR #12288, opened 2022-10-11
feat(DefinitionTooltip): add `openOnHover` prop

### LINKED ISSUE #3104: DefinitionTooltip docs on hover do not match implementation
DefinitionTooltip is unique in that it does not have the option to appear on hover. 

The documentation suggests otherwise:

https://github.com/carbon-design-system/carbon-website/blob/4a0433bfbf8b95186a8a96294ef03bde706617d7/src/pages/components/toggletip/usage.mdx#L71-L72

https://github.com/carbon-design-system/carbon-website/blob/4a0433bfbf8b95186a8a96294ef03bde706617d7/src/pages/components/tooltip/usage.mdx#L337-L340

We need to determine if
1. The implementation is correct and the docs are wrong.
2. The implementation is wrong and the docs are correct.

I think it's the docs that are wrong, but we should make sure by looking back on the flurry of activity on this component leading up to v11.

Checklist:

- [x] Determine if implementation or docs are incorrect
- [x] Depending on the findings, update the docs or update the implementation

tw15egan:
Closes https://github.com/carbon-design-system/carbon-website/issues/3104

Adds an `openOnHover` prop that allows the `DefinitionTooltip` to open on hover

#### Changelog

**New**

- `openOnHover` prop added to `DefinitionTooltip`

**Changed**

- Updated the `align` prop to accept all values that `Popover` supports

#### Testing / Reviewing

Go to the `DefinitionTooltip` playground story and toggle the `openOnHover` prop to `true`. The tooltip should open on mouse enter and close on mouse leave.


aledavila (approved):
lgtm

jnm2377 (approved):

PR #12291, opened 2022-10-12
Update BUG_REPORT.yaml

sstrubberg:
**New**

Added a drop-down where users making bug issues can assign the severity level based on the descriptions our squad came up with and documented in [GH](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/support.md#severity).

#### Intended outcome

To see if the Carbon maker and our squad align on the severity level of a given bug. Ultimately, the person on triage makes the final call, but this could expedite the triaging process or be a conversation starter with the person reporting the bug.


.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
```diff
@@ -70,6 +70,19 @@ body:
...
+      label: Severity
+      description:
+        'Read more to understand our [severity levels](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/support.md#severity)'
+      options:
+        - 'Severity 1 = Must be fixed ASAP. The response must be swift. Someone from the team must drop all current work and be immediately reassigned to address the issue.'
+        - 'Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.'
+        - 'Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.'
+        - 'Severity 4 = Unrelated to a user task, has a workaround or does not need a workaround.'
+    validations:
+      required: true
```
tay1orjones, line 85: The more required fields we have, the more friction and higher the bar for reporting a bug. Over time I think we should be vigilant about keeping it as low as possible. This increases the required fields to 7.

I do think this should be required, but just want to highlight it as something to be mindful of.

tay1orjones (approved):

.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
sstrubberg: Totally hear you. After reviewing the 7, they all seem suuuuper important to me, but definitely something to keep in mind for anything we choose to add in the future.

tay1orjones:
One thing that this can potentially impact negatively is the triage process. 

If the user-reported severity is different than what the core team decides as severity, we'll need to add the severity label _and update the issue body_. If we don't, there will be a mismatch of severity in the issue body and the label applied.

One way around this would be to modify the field to be "Suggested severity" or something to highlight that a user reporting a bug is merely suggesting their perspective of how severe the bug is. The core team will take it from there and determine if the suggested severity stands, or if it should be classified as a different severity and why.

sstrubberg:
> One thing that this can potentially impact negatively is the triage process.
> 
> If the user-reported severity is different than what the core team decides as severity, we'll need to add the severity label _and update the issue body_. If we don't, there will be a mismatch of severity in the issue body and the label applied.
> 
> One way around this would be to modify the field to be "Suggested severity" or something to highlight that a user reporting a bug is merely suggesting their perspective of how severe the bug is. The core team will take it from there and determine if the suggested severity stands, or if it should be classified as a different severity and why.

I dig it. I suppose when I wrote it up, I was thinking that the suggestion was assumed and that ultimately, the labels and the person on triage determines the sev level, not what the user reported in the body. By no means do I want our people on triage going back and fixing the body. I'll update it to suggested priority.

tw15egan (approved):

PR #12297, opened 2022-10-13
feat(react): deprecate light prop everywhere

abbeyhrt:
I think due to confusion on whether or not we would deprecate `light` before or after v11 was released, we were inconsistent with deprecating it. This PR deprecates `light` in all components and removes it from the props table in storybook. Lemme know if we would prefer to have it show up in the props table but just with a note saying it's deprecated instead! 

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Verify that the light prop is deprecated as expected and that the docs changes are consistent across the components.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -563,9 +563,7 @@ Map {
       "hideCopyButton": Object {
         "type": "bool",
       },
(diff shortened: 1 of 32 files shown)


tay1orjones (approved):
👏 Thank you! 👏

aledavila (approved):

PR #12298, opened 2022-10-13
fix(tile): use layer in story

### LINKED ISSUE #11744: [Bug]: Tile with interactive element, text input background not showing
### Package

@carbon/react

### Browser

Chrome

### Package version

Latest

### React version

Latest

### Description

In Tile - expandable with interactive element demo, 
https://react.carbondesignsystem.com/?path=/story/components-tile--expandable-with-interactive
the text input inside expanded region is not showing correct field color. Expecting the field of input to be white, perhaps through using the correct layer or inline theming. 
image

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tile--expandable-with-interactive

### Steps to reproduce

1. Go to https://react.carbondesignsystem.com/?path=/story/components-tile--expandable-with-interactive
2. Click on tile's chevron to expand the tile 

### Code of Conduct

tay1orjones:
Closes #11744 

#### Changelog

**Changed**

- update tile story to wrap input in a `Layer`

#### Testing / Reviewing

View the tile--expandable-with-interactive story, the input should have the proper background color across all 4 themes.


packages/react/src/components/Tile/Tile-story.js
```diff
@@ -165,7 +166,9 @@ export const Expandable = () => (
     <TileBelowTheFoldContent>
       <div style={{ height: '400px' }}>
         Below the fold content here
-        <TextInput id="test2" invalidText="A valid value is required" />
+        <Layer>
+          <TextInput id="test2" invalidText="A valid value is required" />
+        </Layer>
```
tw15egan: We'll need to add this to the `next` story as well, not currently showing correctly in the preview

tw15egan (changes requested):

packages/react/src/components/Tile/Tile-story.js
tay1orjones: Thanks for the catch! Updated

tw15egan (approved):
LGTM 👍 ✅

sstrubberg (approved):

PR #12299, opened 2022-10-13
fix(overflowmenu): convert icon description to sentence case

### LINKED ISSUE #12293: [Bug]: The title attribute of OverflowMenu icon "open and close list of options" should be proper cased or should be in sentence case.
### Package

carbon-components-react

### Browser

_No response_

### Package version

7.37.0

### React version

16.14.0

### Description

The title attribute of OverflowMenu icon (the 3 dots menu icon) "open and close list of options" should be proper cased or should be in sentence case. We can use iconDescription attribute to override the hover text however, as this is a generic issue we would require some default setting in this carbon component for the proper casing of the text.

### Reproduction/example

https://codesandbox.io/s/magical-glitter-03v3mk

### Steps to reproduce

Please hover on the 3 dot icon (OverflowMenu icon). The hover text "open and close list of options" is in small case. It should be either proper cased or sentence cased.

### Code of Conduct

tay1orjones:
Closes #12293 

#### Testing / Reviewing

- Go to an overflow menu, hover over it, the title tooltip should now be in sentence case


tw15egan:
@tay1orjones I don't think we render a `title` on the `OverflowMenu` as-is. Should we add a `title` attribute to the underlying `button` that renders the `iconDescription` content as default?

tay1orjones:
@tw15egan looking closer at this you're right, it's not a `title` on the OverflowMenu. It looks like in v10 we were placing a `<title>` _within_ the svg icon. In v11 we no longer do that.

I don't think we necessarily need to apply a `title` anywhere, but it is worth noting that there is no longer a title tooltip rendered when hovering on the OverflowMenu in v11.

Bigger thing I noticed though is that one of these changes is behind the v11 flag and so I need to make this change in v10 as well.

tw15egan (approved):

aledavila (approved):

Issue #11949, opened 2022-08-12
[Feature Request]: Export color codes (hex/rgba) for Component colors


dabrad26:
### Summary

For theme colors it is possible to get the hex/rgba colors used by the SCSS via JavaScript modules. This is handy for using these items in other setups. Like React Native.

We currently get theme colors from `@carbon/themes/src/[g100/g10].js` which provides an object of tokens to hex/rgba syntax (camelCase version of the kebab style mixing found on Carbon website).

Would be nice to be able to get the component colors as well (button-, tag-, notification-...)

Currently we can get the main theme but not the components which forces us to hardcode them in our app. But ideally we could get all from Carbon project and keep Carbon as source of truth without having to upgrade in parallel.

### Justification

For the React Native library we are working on we would like to get theme color data from Carbon directly instead of duplicating items and missing changes in Carbon not propagating into our library.

### Desired UX and success metrics

People using Carbon React Native library in future won't have to wait on us if colors change in Carbon as updating `@carbon/themes` would do the work for them.

### Your team

Carbon Mobile Design and IBM Aspera

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

tay1orjones:
After discussing with the team, we can duplicate the values from the sass into the js for now. 

At a later date we can automate the creation of component tokens in sass from the js.

PR #12300, opened 2022-10-13
feat(themes): generate button tokens 

abbeyhrt:
REF #11949 

Creates a component token file in JS and generates scss based on those values. Adjusts `button/tokens.scss` in `@carbon/styles` to use those generated files. This is just for the Button component, if this approach is okay with everyone, I'll make the changes for Tag and Notification as well 

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Verify that `Button` in all themes and variants, looks the same as before (hopefully VRT should catch if it doesn't) and confirm that this is the approach we want for component token generation


tw15egan (approved):
This is awesome 🎉 👍🏻 ✅

tay1orjones (approved):
Having a single source of truth like this is so much better than duplicating the values into the js like I originally proposed. Thank you! 💪 🎉 🙏

PR #12304, opened 2022-10-14
ci(release-notifications): add missing inputs

tay1orjones:
A couple of the inputs were missed in the action.yml config, causing the workflow to fail.


tw15egan (approved):
LGTM 👍 ✅

aledavila (approved):

PR #12306, opened 2022-10-14
Use quoted strings for color names in the swatch map

neo:
According to the [Sass Docs](https://sass-lang.com/documentation/values/maps#:~:text=%E2%9A%A0%EF%B8%8F-,Heads%C2%A0up,-!):
> Most of the time, it’s a good idea to use [quoted strings](https://sass-lang.com/documentation/values/strings#quoted) rather than [unquoted strings](https://sass-lang.com/documentation/values/strings#unquoted) for map keys. This is because some values, such as color names, may look like unquoted strings but actually be other types. To avoid confusing problems down the line, just use quotes!

Although this is most likely a breaking change... 😳


neo:

tay1orjones:
> Although this is most likely a breaking change... 😳

Actually, I don't think it is! Based on this from [the sass docs](https://sass-lang.com/documentation/values/strings#quoted:~:text=When%20a%20quoted%20string%20is%20injected%20into%20another%20value%20via%20interpolation%2C%20its%20quotes%20are%20removed!%20This%20makes%20it%20easy%20to%20write%20strings%20containing%20selectors%2C%20for%20example%2C%20that%20can%20be%20injected%20into%20style%20rules%20without%20adding%C2%A0quotes.):

> When a quoted string is injected into another value via interpolation, [its quotes are removed](https://sass-lang.com/documentation/interpolation#quoted-strings)! This makes it easy to write strings containing selectors, for example, that can be injected into style rules without adding quotes.

tay1orjones (approved):
This looks okay to me. Thank you!

tw15egan (approved):

patricklx:
this broke carbon/charts  as there its still using unquoted keys to get the colors

neo:
Yes, it's probably indeed a breaking change – that when doing `map-get`, it won't try to add the add the quotes to the key to find the match.

tay1orjones:
Sorry, I didn't anticipate that use case. We can revert this 👍

tay1orjones:
I just published [`v11.17.0-rc.1`](https://github.com/carbon-design-system/carbon/releases/tag/v11.17.0-rc.1) that includes the reverted fix.

PR #12309, opened 2022-10-16
chore(react): updated formgroup playground

### LINKED ISSUE #12339: Storybook controls: add working controls for FormGroup

TannerS:
Closes https://github.com/carbon-design-system/carbon/issues/12339

### New

    New storybook via formgroup.stories.js

### Removed

    deleted formgroup-story.js

### Testing / Reviewing

Reviewed in storybook


packages/react/src/components/FormGroup/FormGroup.stories.js
```diff
@@ -0,0 +1,95 @@
...
+import FormGroup from '../FormGroup';
+import TextInput from '../TextInput';
+import RadioButtonGroup from '../RadioButtonGroup';
+import RadioButton from '../RadioButton';
+import Button from '../Button';
+import { Stack } from '../Stack';
+
+export default {
+  title: 'Components/FormGroup',
+  component: FormGroup,
```
tay1orjones, line 18: One pattern we've started to follow is that if a prop does not impact the component visuals at all, we hide/remove it from the controls panel. Also if a prop does not have a meaningful control (ones with type of node, for instance like `children`) should also be hidden/removed from the controls panel.

Putting these up here at the component-level makes these apply to all stories, instead of just one. So these will be hidden from the controls pane for all stories. They will still show up in the component docs tables though on the `Docs` tab!

```suggestion
  component: FormGroup,
  argTypes: {
    legendId: {
      table: {
        disable: true,
      },
    },
    className: {
      table: {
        disable: true,
      },
    },
    children: {
      table: {
        disable: true,
      },
    },
  },
```

packages/react/src/components/FormGroup/FormGroup.stories.js
```diff
@@ -0,0 +1,95 @@
...
+  legendId: {
+    control: {
+      type: 'text',
+    },
+  },
+  invalid: {
+    control: {
+      type: 'boolean',
+    },
+    defaultValue: false,
```
tay1orjones: This control doesn't seem to do anything? We might need to apply `args.invalid` to a sub component in the playground story or something maybe?

tay1orjones (changes requested):
Just a couple things to change, this is looking good so far!

packages/react/src/components/FormGroup/FormGroup.stories.js
TannerS: > One pattern we've started to follow is that if a prop does not impact the component visuals at all, we hide/remove it from the controls panel. Also if a prop does not have a meaningful control (ones with type of node, for instance like `children`) should also be hidden/removed from the controls panel.
> 
> Putting these up here at the component-level makes these apply to all stories, instead of just one. So these will be hidden from the controls pane for all stories. They will still show up in the component docs tables though on the `Docs` tab!

I wont lie, i see these in a bunch of components

```
export default {
  title: 'Components/FormGroup',
  component: FormGroup,
  parameters: {
    docs: {
      page: mdx,
    },
  },
};
```

and even ones that have `subComponents` such as 

```
export default {
  title: 'components/TreeView',
  component: TreeView,
  subcomponents: {
    TreeNode,
  },
  args: {},
  argTypes: {
    children: {
      table: {
        disable: true,
      },
    },

    className: {
      table: {
        disable: true,
      },
    },

    label: {
      table: {
        disable: true,
      },
    },

    onSelect: {
      table: {
        disable: true,
      },
    },
  },
};
```

i dont actually know how this works or what it does. I understand that, default will be the default thing in the storybook, and im slowly learning passing arg types and args pass in controls, but why set it up this way vs just exporting a component like others do, or what or how does  `subcomponents` work or what is it for?

maybe if i understand this better I can understand these stories better

TannerS: > argTypes: {

to also add to more questions than above

```
argTypes: {
    legendId: {
      table: {
        disable: true,
      },
    },
    className: {
      table: {
        disable: true,
      },
    },
    children: {
      table: {
        disable: true,
      },
    },
  },
```

setting the arg types with the prop name followed by table than disabled, that is a storybook thing to tell all components within this particular components storybook, to disable them from showing up?

TannerS: @tay1orjones  please see the above, thank you

packages/react/src/components/FormGroup/FormGroup.stories.js
TannerS: looks like the code for this is simply 

```
  <fieldset
      {...(invalid && { 'data-invalid': '' })}
```

is setting in the dom, but i guess it doesnt affect anything visually. maybe removing it is best? no idea what this is suppose to do in thiscomponent

TannerS: @tay1orjones

tay1orjones: @TannerS Yeah that's super old code we might be able to remove but for now let's just remove that prop from the controls panel since it doesn't change anything visually

packages/react/src/components/FormGroup/FormGroup.stories.js
tay1orjones: @TannerS Yeah, I know it's confusing. I will say reading through the storybook docs can help, but I'll annotate one of those examples with some descriptions:

```js
export default {
  title: 'components/TreeView', // Place all stories in this file in a folder "TreeView" inside the section "components"
  component: TreeView, // Adds this component to the API/prop tables in the "Docs" tab
  subcomponents: {
    TreeNode, // Also adds this component to the API/prop tables in the "Docs" tab
  },
  args: {}, // Define "component level" args that will apply to every story in this file
  argTypes: { // Define "component level" argTypes that will apply to every story in this file
    children: {
      table: {
        disable: true,
      },
    },
```

---

> setting the arg types with the prop name followed by table than disabled, that is a storybook thing to tell all components within this particular components storybook, to disable them from showing up?

Yes, exactly. It tells storybook that for every story in this component/file, `disable` the prop from appearing in the controls `table`

Hope this helps!

tay1orjones (approved):
Thank you! 🎉

tw15egan (approved):

PR #12331, opened 2022-10-18
chore(react): add exports to package.json

jharvey10:
Closes (None)

## Changelog

Add an `"exports"` section to the `@carbon/react` package.json file.

From the Node.js docs:

> The ["exports"](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#exports) provides a modern alternative to ["main"](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#main) allowing multiple entry points to be defined, conditional entry resolution support between environments, and preventing any other entry points besides those defined in ["exports"](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#exports). This encapsulation allows module authors to clearly define the public interface for their package.

### Why?

In certain node environments (namely [ts-node](https://typestrong.org/ts-node/)), the `@carbon/react/icons` import is considered as a "directory import" and causes the ts-node/esm loader to fail with an error like this:

```
  Error: ERR_UNSUPPORTED_DIR_IMPORT /Users/jdharvey/code/carbon-design-system/carbon-platform/node_modules/@carbon/react/icons /Users/jdharvey/code/carbon-design-system/carbon-platform/packages/mdx-components/src/main/article-card/article-card.tsx
```

This behavioral inconsistency is documented in the Node docs [over here](https://nodejs.org/api/esm.html#customizing-esm-specifier-resolution-algorithm).

> The current specifier resolution does not support all default behavior of the CommonJS loader. One of the behavior differences is automatic resolution of file extensions and the ability to import directories that have an index file.

**Note:** I can also get around this issue by specifying the `--experimental-specifier-resolution=node` flag when invoking ts-node, but it seems like it might be a good idea to sprinkle in the newer exports syntax for even wider compatibility.

## Noteworthy stuff

- The order of the keys in the exports object matters! Earlier entries take precedence over later entries. This applies to the values within each export key as well (e.g. "import" takes precedence over "default").
- Sass exports are included to make JS-based loaders (such as css-loader) happy.

## v12 breaking consideration

I added a "new" export of `./scss`, which would be used via `@use '@carbon/react/scss';`. This is in addition to what you can currently do via `@use '@carbon/react';`, however to get the latter to work, I had to add two entries to the `"."` export as "fallbacks". To me, this is a bit of an anti-pattern to have the same export path mean something different depending on whether you're using JS or SCSS.

I'd advocate for deprecating the `@use '@carbon/react';` variant and instead moving users towards `@use '@carbon/react/scss';`. Once this is done (as a breaking change), it would mean that the existing top-level index.scss file could instead be renamed to something like `styles.scss` and [part of] the exports could be re-defined as:

```json
"exports": {
  ".": {
    "node": "./lib/index.js",
    "import": "./es/index.js",
    "require": "./lib/index.js",
    "default": "./lib/index.js",
  },
  "./scss": "./styles.scss",
  "./scss/*": "./scss/*",
  ...
},
```

## Testing / Reviewing

- Ensure the usual `@carbon/react` import still works in both a commonjs and esm import environment
- Ensure the `@carbon/react/icons` import still works in both a commonjs and esm import environment

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -5,6 +5,22 @@
   "license": "Apache-2.0",
   "main": "lib/index.js",
   "module": "es/index.js",
+  "exports": {
+    ".": {
+      "node": "./lib/index.js",
+      "import": "./es/index.js",
+      "require": "./lib/index.js",
+      "default": ["./lib/index.js", "./index.scss"]
+    },
+    "./scss": "./index.scss",
+    "./scss/*": "./scss/*",
+    "./icons": {
+      "node": "./icons/index.js",
+      "import": "./icons/index.esm.js",
+      "require": "./icons/index.js",
+      "default": "./icons/index.js"
+    }
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/carbon-design-system/carbon.git",


packages/react/package.json
```diff
@@ -5,6 +5,22 @@
   "license": "Apache-2.0",
   "main": "lib/index.js",
   "module": "es/index.js",
+  "exports": {
+    ".": {
+      "node": "./lib/index.js",
```
joshblack, line 10: Hey Joe! 👋 

This is super random so feel free to disregard 😂 Was just working on this with something else and wanted to share.

You might be able to get away with only `import` and `require`. From what I understand, `node` is the most specific and will match in the Node.js environment even if calling from an ESM context (`"type": "module"`). Switching over to `import` and `require` allow you to retain the import semantics (ESM or CommonJS) regardless of environment (Node.js, bundler)

Hope I'm understanding that right 🤔 Would love to hear what you think!

jharvey10: Hey @joshblack!!

Alright, so you got me digging deeper into this lol, but I'll also give you a little backstory as well.

My read of the docs was definitely the same as yours... `import` and `require` *should be sufficient*, and you can include `default` as a best practice for ultra portability. (`default` is also the one that the css-loader seems to pick up in webpack. It's possible it'd also use one of the higher-priority keys, but it seems to make more sense to keep the scss import in the default block instead of one that's really meant for ES/Node.)

But in practice, this was not behaving as expected 🤔. I get errors like this:

```
SyntaxError: The requested module '@carbon/react' does not provide an export named 'Accordion'
```

I can verify that `ts-node` grabs the ES import that it finds from the `exports` key, but because the es file's extension inside of `@carbon/react` isn't `mjs` and because the package isn't marked as `"type": "module"`, ts-node appears to treat this file as a commonjs module instead of an ES module. It's basically using Node to load the module (accepting ES6-style import syntax along the way), but then resolving its imports as commonjs-style imports when the module is used in a different file 🤷‍♂️.

Things move further if I set `"type": "module"` in the `carbon/react` pacakge.json or I start changing file extensions to `.mjs` (and I exclude the `"node":...` from the exports).

I'm actually not sure whose problem this is to solve 😂. I found this in [the proposal](https://github.com/jkrems/proposal-pkg-exports/) for the exports functionality:

> Resolution should not depend on file extensions, allowing ESM syntax in .js files.

But does that mean that those files are *treated* as ESM even if the package isn't marked as `"type": "module"`? The answer is actually "no" lol. The ES-style syntax is *allowed* in those files, but the module itself is still considered as a cjs module. Here's an example from a minimum repro:

image

All of this is to say that because the chain of ES module files within `carbon/react` aren't **Node-compatible** ES modules (i.e. .mjs file extensions or package.json files with type:node in them), they won't work as-is when imported into a Nodejs environment, but they **will** still work in a browser environment, and there's probably some bundlers/runtimes that'll tolerate the discrepancy and figure out what to do. node/ts-node just isn't one of those.

Since I have a workaround, I think I'll close this PR for now and investigate my options for avoiding ts-node

PR #12358, opened 2022-10-20
feat(number input): added a prop to disable the wheel functionality

### LINKED ISSUE #11982: [Feature Request]: Provide an easy way to disable wheel events in `NumberInput`
### Summary

If I want to disable wheel events in `NumberInput`, I have to add an event listener manually.

Could we have an easier way to disable wheel events in `NumberInput`?

```javascript
function MyNumberInput = () => {
  return (
    <NumberInput
      {...rest}
      ref={inputRef}
      onFocus={() => {
        if (inputRef.current) {
          inputRef.current.addEventListener("wheel", disableWheel);
        }
      }}
      onBlur={() => {
        if (inputRef.current) {
          inputRef.current.removeEventListener("wheel", disableWheel);
        }
      }}
    />
  );
};

/**
 * It prevents any wheel event from emitting.
 *
 * We want to prevent this input field from changing by the user accidentally
 * when the user scrolling up or down in a long form. So we prevent the default
 * behavior of wheel events when it is focused.
 *
 * Because React uses passive event handler by default, we can't just call
 * `preventDefault` in the `onWheel` event handler. So we have to get the input
 * ref and add our event handler manually.
 *
 * @see https://github.com/facebook/react/pull/19654
 * @param {WheelEvent} e A wheel event.
 */
function disableWheel(e) {
  e.preventDefault();
}
```

### Justification

When user editing a long form with a number input field, it is very easy to change the number input accidentally. User complains about saving the wrong value multiple times.

_No response_

### Required functionality

_No response_

_No response_

### Your team

@Superbil

### Available extra resources

https://stackoverflow.com/questions/9712295/disable-scrolling-on-input-type-number

### Code of Conduct

mgueyraud:
Closes #11982 

Added a prop to disable the wheel functionality that inputs with type number have.

#### Changelog

**New**

-  Added a disableWheel prop in the component


mgueyraud:

sstrubberg:
@mgueyraud Looks like one of the tests is failing. From the root of the carbon folder in your dev environment, run `yarn test packages/react/__tests__/PublicAPI-test.js -u`. That should take care of it.

mgueyraud:
Hey @sstrubberg, thank you very much!. I already did the commit, hope it works.

tay1orjones (changes requested):
Thanks for putting this together! Based on the original issue, this is a bit off mark from what was discussed. 

- The wheel event should only be disabled when the input has focus.
- The input should not be blurred.
- The `onWheel` event can not be used due to limitations with React described in the original issue. 

On focus of the input, a manual event handler needs to be applied to the `input` element that calls `evt.preventDefault();` when fired. This manual handler should be removed on blur of the input.

mgueyraud:
Hey @tay1orjones , changed the approach as you said, let me know if it's good.

sstrubberg:
Ok, now I think you need to run `yarn format` or `yarn format:diff` to fix the next error :)

mgueyraud:
Hey @sstrubberg , I did as you said but it seems it's not something from formatting, if you check the action it looks like it didn't install the dependencies correctly.

tw15egan:
@mgueyraud, have you pulled the latest changes from `main` and run `yarn`? May need to run `yarn build` afterward, there have been a few PRs merged this afternoon. Afterward, you can run `yarn test -u` if any snapshots still need updating.

mgueyraud:
@tw15egan , did the merge and added me as contributor also.

tay1orjones (approved):
Looks good! Wheel events do not change the value of the input when the input is focused and the pointer is hovered over the current value. There is no change when the input is not focused.

https://user-images.githubusercontent.com/3360588/198167230-9ac4314e-4ed9-44eb-b48b-3a09d79f3f07.mov



Thanks so much for the contribution! 🎉 

It would be nice if there was a test to cover this functionality, but it's not totally necessary here for this contribution. We'd probably have to use `fireEvent()` instead of something from `user-event`. `user-event` doesn't support wheel/scroll events.

tw15egan (approved):
LGTM, thanks for contributing! 🎉 ✅ 👍🏻

tw15egan:
@mgueyraud Looks like there are some merge conflicts, once these are fixed we're good to go 👍🏻

mgueyraud:
Hey @tw15egan , did the merge and resolved the conflicts!

PR #12374, opened 2022-10-24
fix(TextInput): remove duplicate class when custom class passed down

### LINKED ISSUE #12373: [Bug]: className applied twice in TextInput
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

1.15

### React version

18.3.0

### Description

The margin-bottom of the CSS is applied twice in the TextInput component: 

1. between the text input and the helper label
3. after the helper label 

This breaks what I'd understand the margin-bottom of the component TextInput to do.

```
<TextInput
      className="some-margin"
      id="text-input-1"
      name="textInput"
      invalidText={"invalid"}
      labelText={"textinput label"}
      placeholder={"textinput placeholder"}
      helperText={"textinput helperText"}
      value={"value"}
    ></TextInput>

```

### Suggested Severity

### Reproduction/example

https://codesandbox.io/s/textarea-textinput-inconsistent-zpnr2e

### Steps to reproduce

See codesandbox

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12373

Removes some code causing a custom class to be passed down to the input itself. In V11, custom classes are always applied to the outer wrapper. 

#### Changelog

**Removed**

- Removed a ternary that was causing a custom class always to be passed to the input if using V11
- Removed an unused `${prefix}--text__input` that had no attached styles. `${prefix}--text-input` is the `input` class.

#### Testing / Reviewing

Change the `className` of the `TextInput` component. Ensure the class is only added to the external wrapper.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -1313,7 +1313,6 @@ Map {
   "ControlledPasswordInput" => Object {
     "$$typeof": Symbol(react.forward_ref),
     "defaultProps": Object {
-      "className": "\${prefix}--text__input",
       "disabled": false,
       "helperText": "",
       "invalid": false,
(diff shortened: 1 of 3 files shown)


alisonjoseph (approved):
Confirmed custom class is only getting added to the wrapper div 👍

tay1orjones (approved):

PR #12380, opened 2022-10-24
chore(search): add playground story and remove next folder

### LINKED ISSUE #12353: Storybook controls: update controls for Search, remove /next and export toggle

alisonjoseph:
Closes #12353

#### Changelog

**New**

- Add playground search story
- add displayName to search

**Removed**

- Removed /next folder for search

#### Testing / Reviewing
http://localhost:3000/?path=/story/components-search--playground


aledavila (approved):

packages/react/src/components/Search/next/Search-test.js
```diff
@@ -1,335 +0,0 @@
-/**
- * Copyright IBM Corp. 2016, 2018
- *
- * This source code is licensed under the Apache-2.0 license found in the
- * LICENSE file in the root directory of this source tree.
- */
-
-import React from 'react';
-import { Search as SearchIcon, Close } from '@carbon/icons-react';
-import Search from './Search';
```
tay1orjones, line 10: IIUC this path means the `next` test has been running against the classic/original `Search`, not `next/Search`? This could be why tests are failing now all of the sudden.

packages/react/src/components/Search/Search.js
```diff
@@ -5,283 +5,251 @@
...
-
-  static getDerivedStateFromProps({ value }, state) {
-    const { prevValue } = state;
-    return prevValue === value
-      ? null
-      : {
-          hasContent: !!value,
-          prevValue: value,
-        };
+const Search = React.forwardRef(function Search(
```
tay1orjones, line 20: IIRC due to the forwardRef there isn't a name associated with this component. Adding a `displayName` to this might fix some of the tests that are searching for the name "Search"

tay1orjones: https://github.com/carbon-design-system/carbon/pull/9591/

packages/react/src/components/Search/next/Search-test.js
alisonjoseph: ahh yea, that makes sense

packages/react/src/components/Search/Search.js
alisonjoseph: That worked!

sstrubberg (approved):

PR #12382, opened 2022-10-24
fix: remove icon from readonly text input

lee-chase:
Contributes to #2177

Adjusts the readOnly textInput based on #2177. 

- This removes the EditOff icon.
- Removes requirement from test for icon.

#### Changelog

**Changed**

- TextInput

#### Testing / Reviewing

Reviewed the result in Storybook, updated test.


tay1orjones (approved):

aagonzales (approved):
Approving the changes list. But I had a follow up, is the read-only input supposed to be focusable? I see its receiving focus currently and that makes it seem like its actionable.

alisonjoseph (approved):

lee-chase:
@aagonzales the standard HTML input is focusable when readonly https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_readonly

tw15egan (approved):

PR #12387, opened 2022-10-25
test(containedlist): move to stable - add vrt, avt, storybook docs

### LINKED ISSUE #12319: Move `ContainedList`, `ContainedListItem` to stable
`ContainedList` and `ContainedListItem` have been flagged as ready for a review to move to stable. Based on [the docs](https://github.com/carbon-design-system/carbon/blob/main/docs/experimental-code.md), the following criteria need to be met when moving a component from experimental to stable:

- [x] All files have a copyright banner
- [x] All components exported in `src/index.js` and should not be `unstable_`
      prefixed
- [x] Component has a label in the github repository
- [ ] Component should be documented on the website
  - [x] Component should have a usage tab
  - [x] Component should have a style tab
  - [x] https://github.com/carbon-design-system/carbon-website/issues/3198
  - [x] https://github.com/carbon-design-system/carbon-website/issues/3153
  - [x] https://github.com/carbon-design-system/carbon-website/issues/3152
- [x] For each component exported:
  - [x] Component is written as a function declaration or uses `forwardRef`
  - [x] Component has `propTypes` defined
    - [x] Each prop type has a comment (used in storybook)
    - [x] Prop types are as specific as needed, prefer `PropTypes.shape` over
          `PropTypes.object` if possible
  - [x] Default props are listed as default args in the function definition (not
        in defaultProps)
    - [x] Note: default props should be stable, in other words props like
          `onClick = () => {}` can cause re-renders since the function identity
          is not stable
  - [x] Component has a story in `<ComponentName>.stories.js`
    - [ ] Component has an mdx document that follows our outline
    - [ ] mdx document coverages at least common use-cases and provides a prop
          table
    - [x] Stories cover at least common use-cases
    - [x] Stories may include a `Playground` story for controls
    - [x] Stories should mirror intended usage of the component
  - [x] Component has unit/integration tests written in RTL for testing the
        component API
  - [ ] Component is tested via VRT for at least the initial render state
  - [ ] Component is tested via AVT for at least the initial render state
 

Most of these above are complete already from the initial delivery in https://github.com/carbon-design-system/carbon/pull/11969 🎉 

---

Additionally I think we need to address the following in the storybook:
- [x] On the playground story, hide `className`, `actions`, and `children` from the controls table
- [x] Suggestions for updated story names
  - [x] Following the storybook convention, it needs a "Default" story
    - [x] Rename `On page` -> Default
  - [x] Disclosed 
  - [x] With actions
  - [x] With icons
  - [x] With interactive items
  - [x] With interactive items and actions
  - [x] With list title decorator
  - [x] Playground
  - [x] would it be worthwhile to have a `with layer` story? I figure that might be a common use case
  - [x] Could we reduce the repetition/duplication in the stories down to a single `ContainedList` with `ContainedListChildren` in each story?

tay1orjones:
Closes #12319

Final steps to mark ContainedList as stable

#### Changelog

**New**

- Component has an mdx document that follows our outline
- mdx document provides a prop table
- Component is tested via VRT for all stories
- Component is tested via AVT for the initial render state
- Added platform metadata `carbon.yml`, added live demo .mdx

**Changed**

- Move story file out of `/next`
- Move story out from under `Experimental` to `Components` section
- Hide `action`, `children`, `className` from all stories' controls pane

**Removed**

- `unstable_` prefix from ContainedList

#### Testing / Reviewing

- View all stories, ensure they render and function correctly
- Make sure exports look correct - `unstable_` prefix should be gone from both ContainedList and ContainedListItem


janhassel (approved):
LGTM, looking forward to having this component promoted to stable 👍 

One question / comment: I noticed your removed the second list from both the "Default" and "Disclosed" story. The intent was to show their difference in spacing ("on page" having a top spacing if adjacent to another list while "disclosed" is flush). Do you think it would be worth to bring these back or does the Playground story cover this enough?

e2e/components/ContainedList/ContainedList-test.e2e.js
```diff
@@ -0,0 +1,101 @@
...
+        });
+      });
+
+      test('default @vrt', async ({ page }) => {
+        await snapshotStory(page, {
+          component: 'ContainedList',
+          id: 'components-containedlist--default',
+          theme,
+        });
+      });
```
tay1orjones, line 23: Duplicate here I need to fix

packages/react/src/components/ContainedList/ContainedList.mdx
```diff
@@ -0,0 +1,34 @@
+import { Props, Story, Preview } from '@storybook/addon-docs';
+
+# ContainedList
+
+[Source code](https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/ContainedList)
+&nbsp;|&nbsp;
+[Usage guidelines](https://www.carbondesignsystem.com/patterns/common-actions/#copy)
```
janhassel:
```suggestion
[Usage guidelines](https://carbondesignsystem.com/components/contained-list/usage/)
```

tay1orjones:
> One question / comment: I noticed your removed the second list from both the "Default" and "Disclosed" story. The intent was to show their difference in spacing ("on page" having a top spacing if adjacent to another list while "disclosed" is flush). Do you think it would be worth to bring these back or does the Playground story cover this enough?

@janhassel Sorry, I totally missed this detail. I just pushed a commit bringing it back so the VRT will cover that styling and prevent regressions there.

e2e/components/ContainedList/ContainedList-test.e2e.js
tay1orjones: Was missing `disclosed`

alisonjoseph (approved):
LGTM 🚀

tw15egan (approved):
Looks great!

PR #12462, opened 2022-11-01
feat(Modal): add isFullWidth prop

### LINKED ISSUE #11933: [Feature Request]: Modal to allow full width & full height content
### Summary

This feature request is to add an option to remove the margin left, right and bottom and give the creative freedom to team using it. This should give us a modal content area with no styles. This would allow proper usage of container components inside modal, such as data table, tabs, fluid text input, etc, to look edge to edge, and achieve text alignment.

### Justification

Current Carbon Modal has margin left right and bottom built in, making it impossible to achieve these following designs documented on [Carbon Modal page](https://carbondesignsystem.com/components/modal/usage/) and [Carbon for IBM Products](https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage/)
image
image

The column customization modal actually looks like this in implementation. Notice how margin left, right and bottom is limiting the data table inside to reach full width.
image

This feature request is to add an option to remove the margins and give the creative freedom to teams to implement contained components inside modal in the expected way. We are going to be adding it for Carbon for Salesforce to ensure Column customization modal look as expected. You can reference [our implementation here](https://github.ibm.com/carbon-for-salesforce/carbon-for-salesforce/pull/109#pullrequestreview-11056373).

_No response_

### Required functionality

_No response_

_No response_

### Your team

Carbon for Salesforce / EPX (Ecosystem Platform Experiences)

### Available extra resources

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/11933

Adds a new `isFullWIdth` prop that removes any margin and padding placed on the `modal-content` class. 

#### Changelog

**New**

- `isFullWidth` prop and associated styles for `Modal` and `ComposedModal` that removes margin and padding. 
- e2e tests added to catch regressions
- updated snapshots

#### Testing / Reviewing
Was unsure if we also wanted to remove the padding/margin from the headers; for now they are still there because they match the spec images

- Go to `ComposedModal` and `Modal` and ensure the `Full Width` example matches the spec images.

### DIFF
--- a/e2e/components/ComposedModal/ComposedModal-test.e2e.js
+++ b/e2e/components/ComposedModal/ComposedModal-test.e2e.js
@@ -37,6 +37,13 @@ test.describe('ComposedModal', () => {
           theme,
         });
       });
+      test('full width modal @vrt', async ({ page }) => {
+        await snapshotStory(page, {
+          component: 'ComposedModal',
+          id: 'components-composedmodal--full-width',
+          theme,
+        });
+      });
     });
   });
 
--- a/e2e/components/Modal/Modal-test.e2e.js
+++ b/e2e/components/Modal/Modal-test.e2e.js
@@ -29,6 +29,14 @@ test.describe('Modal', () => {
           theme,
         });
       });
+
+      test('full width modal @vrt', async ({ page }) => {
+        await snapshotStory(page, {
+          component: 'Modal',
+          id: 'components-modal--full-width',
+          theme,
+        });
+      });
     });
   });
 
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -1254,6 +1254,9 @@ Map {
       "danger": Object {
         "type": "bool",
       },
+      "isFullWidth": Object {
+        "type": "bool",
+      },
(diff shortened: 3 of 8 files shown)


aagonzales:
The full-width stories look good, but I had a follow-up question after seeing the Playground example. Would a user be able to mix full width and the indented style? Or is it an either-or situation? Like below

image

tw15egan:
I think they would still be able to do that; they would just have to add padding independently. This just removes any padding we ship to prevent our styles from blocking any designs since we only place padding on the modal body container.

shixiedesign (approved):
Thank you! Looks great and thanks for the ⚡ speed. I'm wondering if this creates additional work for @carbon-design-system/carbon-platform-designers to add guidance on when to use the `isFullwidth` prop...? This PR looks good to me though! Thank you!

aagonzales:
You bring up a good point @shixiedesign we need to add guidance on the website now. It would actually be the system squad who adds it (platform is pulling markdown from the original site atm). I've made an [issue](https://github.com/carbon-design-system/carbon-website/issues/3239) so we don't forget.

aagonzales (approved):

alisonjoseph (approved):

tay1orjones (approved):
Looks great!

PR #12472, opened 2022-11-03
feat(time-picker): add storybook controls

### LINKED ISSUE #12356: Storybook controls: update controls for TimePicker, remove `/next` and export toggle

aledavila:
Closes #12356

Make sure stories work as intended.
Make sure controls work as intended and only showing in playground


packages/react/src/components/TimePicker/TimePicker.stories.js
```diff
@@ -89,3 +81,74 @@ export const WithLayer = (args) => {
     </>
   );
 };
+
+export const Playground = (args) => {
+  return (
+    <TimePicker id="time-picker" labelText="Select a time" {...args}>
+      <TimePickerSelect id="time-picker-select-1">
```
tw15egan:
```suggestion
      <TimePickerSelect id="time-picker-select-1" disabled={args.disabled}>
```

Can we add this so that the `select` element gets disabled when the playground control is set?

packages/react/src/components/TimePicker/TimePicker.stories.js
```diff
@@ -89,3 +81,74 @@ export const WithLayer = (args) => {
...
+        <SelectItem value="AM" text="AM" />
+        <SelectItem value="PM" text="PM" />
+      </TimePickerSelect>
+      <TimePickerSelect id="time-picker-select-2">
```
tw15egan:
```suggestion
      <TimePickerSelect id="time-picker-select-2" disabled={args.disabled}>
```

Can we add this so that the `select` element gets disabled when the playground control is set?

tw15egan (approved):
Small comment but otherwise looks great 👍🏻

alisonjoseph (approved):

PR #12501, opened 2022-11-04
fix(button): text alignment and using rem for padding

### LINKED ISSUE #9478: [Button] Icon should top align when button label wrap to 2 lines
### Button icon should top align to 1st line of button label when the label wraps

Currently in some field height has this behavior, but not all. Especially missing from default field height (48px). This behavior (when label wrap, top align icon) should apply to all heights of the button.

Storybook: [button playground ](https://react.carbondesignsystem.com/?path=/story/components-button--playground)

image

Expecting icon to align to first line of text:

image

### LINKED ISSUE #10318: [Bug]: Button padding is not using relative units
### Package

carbon-components

### Browser

_No response_

### Package version

10.50.0

### React version

_No response_

### Description

The button’s left and right padding use pixel units while top and bottom use rem so they're not scaling correctly when changing the font size settings.

_Chrome’s font size setting: Very Large vs Medium_
image
image

https://github.com/carbon-design-system/carbon/blob/main/packages/components/src/globals/scss/_theme-tokens.scss#L97-L130

### CodeSandbox example

https://codesandbox.io/s/7t5hz

### Steps to reproduce

Increase/decrease browser font size setting.

### Code of Conduct

IgnacioBecerra:
Closes #9478
Closes #10318

#### Changelog

**New**

- aligning items at the top of a button set no matter what

**Changed**

- using `rem` over `px` in button padding styles

#### Testing / Reviewing

The buttons should have the same padding as before, they will just scale better in different screens.

Also inspect one of the buttons and add more than two or three lines of text, the other button's text will remain at the top.


tw15egan:
It seems like the icon is still center-aligned when text wraps to multiple lines

Screenshot 2022-11-08 at 11 11 51 AM

tay1orjones:
Hey @IgnacioBecerra - just following up here. I noticed on the "set of buttons" story that the text vertical alignment isn't correct. It's a bit too high - not center/middle.

image

tay1orjones (changes requested):
Hey @IgnacioBecerra - just following up here. I noticed on the "set of buttons" story that the text vertical alignment isn't correct. It's a bit too high - not center/middle.

image

Issue #12465, opened 2022-11-02
[Bug]: textarea cols not working as expected


ariellalgilmore:
### Package

@carbon/react

### Package version

latest

### React version

latest

### Description

When adjusting the `cols` in `<TextArea>` nothing is changing. I would expect the width of the `<TextArea>` to shrink or grow depending on the number set to `cols`. 

`rows` works as expected.

when removing the `min-width` and `width: 100%` then the `<TextArea>` adjusts as expected
<img width="457" alt="Screen Shot 2022-11-02 at 7 39 08 AM" src="https://user-images.githubusercontent.com/20210594/199518703-b9a4cd8f-7294-4b2e-9f05-436b9351c498.png">

### Suggested Severity

Severity 4 = Unrelated to a user task, has a workaround or does not need a workaround.

### Reproduction/example

https://codesandbox.io/s/vigorous-knuth-vsff4p?file=/src/App.js

### Steps to reproduce

Go to the example and try changing the number of `cols`.

Then try adjusting the styles like mentioned above and see how it then does shrink/grow

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
Thanks for opening this @ariellalgilmore! I'm wondering if this is intentional, to override the `columns` value provided to the input. 

@laurenmrice @tw15egan do either of you recall if there's reasoning for the `width: 100%` on the TextArea?

tw15egan:
@tay1orjones I would imagine it's just to conform with other form elements that they expand to fill their container. We could probably update this to be an inline style and not apply the inline style if the `col` prop is passed down.

ariellalgilmore:
@tw15egan would it make sense to still keep the `min-width` applied as well?

TannerS:
@tay1orjones @tw15egan  I want to warn against inline styles

ibm cloud/console is moving to implementing CSP header on browsers and inline styles will be blocked from displaying which means any cloud app using carbon will not be following this but instead throwing an error in the console

tw15egan:
@TannerS yikes, that may be a bigger issue across the repo, as I believe we do this is a few places. Might be worth stubbing out a new issue to take a deeper look into it.

[...]

PR #12502, opened 2022-11-04
fix(text-area): cols attribute

ariellalgilmore:
Closes #

#12465 

- cols was not working as expected. The `width: 100%` style was not letting user's set the amount of `cols`

#### Changelog

**Changed**

- Change the `width: 100%` to be an inline style, so now it's a conditional on if the `cols` value is set or not

**Removed**

- removed the default value of `cols: 50` because if there is no number assigned to `cols` then it should just be 100% width

#### Testing / Reviewing

- Check that all the TextArea stories show initially 100% width with no `cols` value.
- then go to the playground story and try updating the number set to `cols`
Nov-04-2022 15-41-40

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -7636,7 +7636,6 @@ Map {
   "TextArea" => Object {
     "$$typeof": Symbol(react.forward_ref),
     "defaultProps": Object {
-      "cols": 50,
       "disabled": false,
       "enableCounter": false,
       "helperText": "",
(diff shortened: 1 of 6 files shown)


alisonjoseph (changes requested):
This looks great! 

Looks like Percy is complaining because we have  `cols: 50` set in the `Form.stories` file. I think if we remove that it should show up back at the default of 100% which is what we want.

ariellalgilmore:
> This looks great!
> 
> Looks like Percy is complaining because we have `cols: 50` set in the `Form.stories` file. I think if we remove that it should show up back at the default of 100% which is what we want.

fixed!

alisonjoseph (approved):
🥳

tw15egan (approved):
LGTM 👍 ✅

PR #12512, opened 2022-11-07
fix(input): override Chrome autocomplete styles

### LINKED ISSUE #9645: [Bug]: autofill breaks dark themes
### Package

carbon-components

### Browser

Chrome

### Package version

10.37

### Description

Chrome has this presumptuous `!important` CSS that overrides both the color and background-color of autofilled form fields:

Screen Shot 2021-09-11 at 8 42 35

(Other browsers may have something similar.)

Naturally, that `!important` browser CSS "breaks" Carbon's `g90` and `g100` themes for example:

Screen Shot 2021-09-11 at 8 51 07

You might think the solution is to set `autocomplete="off"` or similar on the `<form>`, but that doesn't work reliably.  Chrome and other browsers decided to give users what they want (i.e. autofill) rather than doing what most HTML pages say to do (i.e. no autofill).

To make matters worse there seems to be a Chrome bug that sometimes sets the auto-fill pseduo-class even when there's no prefilled value, in which case hovering and then unhovering the field clears it.

But basically, the issue is that Chrome's built-in CSS is overriding Carbon's.

### CodeSandbox example

https://codesandbox.io/s/carbon-autofill-7fkq6?file=/public/index.html

### Steps to reproduce

In our app, the autofill is happening just by loading the page.  I couldn't get a good example of that.  But I did create https://codesandbox.io/s/carbon-autofill-7fkq6?file=/public/index.html.  If you fill in your first name from Chrome's autofill dropdown, then Chrome will fill in all the other fields for you.   And all the fields will all have a light blue background

If you switch the theme to `g90` or `g100` (not sure how to do this in CodeSandbox, maybe you can show me), then the light blue background will still be there.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9645

Fix found via comments in [this article](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/). 

This change overrides Chrome's default `autofill` styles to reset the component to normal. 

Question for @carbon-design-system/design - do we want to use a custom "autocomplete" color to differentiate when this is active? 

#### Changelog

**New**

- selector to change `autocomplete` styles on `input` and `textarea` elements

#### Testing / Reviewing

Updated the `TextInput` `Toggle Password Visibility` story to enable `autocomplete` functionality. Ensure the Chrome autocomplete styles are not visible and are correctly overridden.


alisonjoseph (approved):
LGTM!

aledavila (approved):

PR #12568, opened 2022-11-08
fix(Icons): story icons no longer shrink to fit container

jnm2377:
The current icons story has a bug where the icons are much smaller than their actual sizes: https://react.carbondesignsystem.com/?path=/story/elements-icons--default this is because the wrapper has a set width, so the svgs are shrinking to fit that width.

#### Changelog
- increase wrapper width

#### Testing / Reviewing
- go to icons story example > icons should now be the correct size
- note: unsure if this story has a VRT, so tests may break bc of the visual change


alisonjoseph (approved):

packages/react/src/components/Icons/Icons.stories.scss
```diff
@@ -6,7 +6,7 @@
 //
 .demo-icon-example {
   display: flex;
-  width: 10rem;
+  width: 15rem;
```
tw15egan, line 9: Would it also be helpful to set a `min-width` on the `svg`?

jnm2377: @tw15egan that's a good call. better safe than sorry. 👍🏽

jnm2377: @tw15egan pushed an update. lmk if you think I should revert the original width to 10rem.

tw15egan: Nah, can't hurt to have both. Thanks for fixing this!

tw15egan (approved):

Issue #8405, opened 2021-04-14
Provide useNotification hook


tay1orjones:
We offer a variety of notification styles but unfortunately leave much of the animation and logic around display notifications to the end-user. We would like to codify this behavior/logic into a single hook, `useNotification`, that would encapsulate this behavior and make it easier for product teams to display/render notifications in their UI. This may look like:

```jsx
function MyComponent() {
  const send = useNotification();

  function sendNotification() {
    send({ title: 'This is the notification title' });
  }

  return <button type="button" onClick={sendNotification}>Send test notification</button>;
}
```

Packages impacted
- `@carbon/react`

#### Acceptance Criteria

- [ ] TBD

tay1orjones:
A few additional opportunities to look into with this:

1. Improve the "escape to close" functionality proposed in https://github.com/carbon-design-system/carbon/pull/8560
   - Regarding multiple/stacked notifications, this hook could assist in keeping track of how many notifications have been rendered, and how to close them in order of appearance using <kbd>esc</kbd>. 

2. Keyboard navigation improvements
   - It would also be beneficial if focus could be handled to move from one notifcation to the next and also when closed via <kbd>esc</kbd> or the close button. The hook could potentially inform tab order.

3. Notification Panel interop
   - Could this hook can assist with the development of the [notification panel](https://www.carbondesignsystem.com/patterns/notification-pattern/#notification-panel) pattern by sending notification data to a central store when a notification is sent via the hook?

johnbister:
Here’s a rough pass on how multiple inline notifications could stack and be dismissed.  I kept it simple for now to make sure we like the overall mechanics. 

@mjabbink @jeanservaas @tay1orjones Let me know if you have any feedback I should take into the next pass. Thanks!

https://user-images.githubusercontent.com/70543333/120850559-752c4300-c52c-11eb-9606-de8e26c54cbf.mp4

johnbister:
We might want to limit how many notifications are visible in the stack at a time.  We could accomplish this by some kind of gradient on the maximum oldest notification being displayed. As newer notifications are dismissed, older ones become visible.

https://user-images.githubusercontent.com/70543333/121598410-a73f1880-c9f6-11eb-88a9-a0e1e51e84f9.mp4

Also, for any notifications that timeout by default (might be more relevant to the toast notifications) we could show the user that this notification will dismiss automatically by using the thick stroke on the left as a timer.


[...]

PR #12572, opened 2022-11-08
feat(notificationstack): New hooks and components for NotificationStack

cgirani:
Reference: #8405

New hooks and components to render NotificationStack. 

#### Components/hooks and their use/purpose:
**`NotificationStack`** - Component to be used in begin of the application, it's a wrapper of all notifications
**`notification-store`** - Function where we hold the state of notifications, along with functions that act like dispatch/reducer
**`notification`** - Export a function to trigger the notification, this function can be imported in any file in the application
**`useNotification`** - Hook with notifications info from the store and some handlers (actions/dispatch) to be used, this one is used in the component `NotificationStack`

Even files are aligned about Notifications hooks, trigger function and store are generic enough for future uses of any other component...

#### Changelog

**New**

- New files to support notification stack headless
   - NotificationStack
   - notification-store
   - notification
   - useNotification
   - css file for NotificationStack

**Changed**

- `Index.css` to add new css files

**Removed**

- None

#### Testing / Reviewing

At the storybook test the NotificationStack component.
- If opens multiple notifications
- If closes correctly with the timeout
- Animation worksas expected

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -895,6 +895,15 @@
         "code",
         "doc"
       ]
+    },
+    {
+      "login": "cgirani",
+      "name": "Claudia Tejos",
+      "avatar_url": "https://avatars.githubusercontent.com/u/22034430?v=4",
+      "profile": "https://github.com/cgirani",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
--- a/README.md
+++ b/README.md
@@ -200,6 +200,9 @@ check out our [Contributing Guide](/.github/CONTRIBUTING.md) and our
     <td align="center"><a href="http://jakubfaliszewski.github.io/portfolio/"><img src="https://avatars.githubusercontent.com/u/25402419?v=4?s=100" width="100px;" alt=""/><br /><sub><b>Jakub Faliszewski</b></sub></a><br /><a href="https://github.com/carbon-design-system/carbon/commits?author=jakubfaliszewski" title="Code">💻</a></td>
     <td align="center"><a href="https://github.com/francinelucca"><img src="https://avatars.githubusercontent.com/u/40550942?v=4?s=100" width="100px;" alt=""/><br /><sub><b>Francine Lucca</b></sub></a><br /><a href="https://github.com/carbon-design-system/carbon/commits?author=francinelucca" title="Code">💻</a> <a href="https://github.com/carbon-design-system/carbon/commits?author=francinelucca" title="Documentation">📖</a></td>
   </tr>
+  <tr>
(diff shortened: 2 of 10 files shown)


cgirani:
> **DCO Assistant Lite bot:** Thanks for your submission! We ask that you sign our [Developer Certificate of Origin](https://github.com/carbon-design-system/carbon-dco/blob/main/dco.md) before we can accept your contribution. You can sign the DCO by adding a comment below using this text:
> 
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request

tay1orjones (commented):
Thanks so much for submitting this! It's a great first step towards landing something like this in the project.

I'd love to schedule some time to review this with you further and see how we could nail down the experience and API a bit more.

Some high level thoughts from an initial glance:

> `NotificationStack` - Component to be used in begin of the application, it's a wrapper of all notifications

Great approach here, I think it's a very common pattern to expect consumers to wrap part or all of their application tree with a context-provider-type of component like this.

> `notification-store` - Function where we hold the state of notifications, along with functions that act like dispatch/reducer

I'd like to see this use `useReducer` instead of a vanilla reducer pattern. Also, have you ran into any issues storing these in state? I could forsee a situation where someone needs to persist the notification store across sessions - we might want to think about how to open up this to enable "saving" to an external store, if even just localstorage.

> `notification` - Export a function to trigger the notification, this function can be imported in any file in the application

I'd like to see this provided by the hook, instead of it's own function to import. Also we should look into using `useLayoutEffect` instead of a `setTimeout` for being able to calculate the height.

> `useNotification` - Hook with notifications info from the store and some handlers (actions/dispatch) to be used, this one is used in the component NotificationStack

I'd really like to refocus this on the experience of the consumer. I think in an ideal world we'd have the `useNotification` hook as a central entrypoint for all of this behavior. `trigger` would be destructured off the hook in consumer apps or in storybook examples.

Will a consumer ever need to `dispatch` manually? If we need a second hook for the internal `dispatch`, etc., that would be fine, but we should probably keep those concerns separate from `useNotification` if there's no need for a consumer to call `dispatch`.

Additional thoughts:

- This only works with ToastNotification. Ideally, InlineNotification and ActionableNotification support would be included and provide the animation defined in the root issue.
- I see you haven't added any of this to the root index.js exports yet. We'll want to really hone in on what we intend to expose to consumers and think about limiting that API surface area so the underlying implementation can shift without introducing breaking changes to consumers.
- As this firms up we can first introduce it as experimental, prefixed as `unstable__useNotification`.
- Respecting the user's `prefers-reduced-motion` setting will be important to provide support out of the gate

Again thank you for this initial drop! 🙏 🎉  I can reach out to schedule some time to discuss this.

cgirani:

tay1orjones:
Sorry this has languished - our team has had other priorities and been unable to pick up this workstream and devote the time and attention it deserves. If anyone is willing to continue this work based on the suggestions and feedback provided above, we'd love to chat

PR #12600, opened 2022-11-11
fix(typescript): provide .d.ts files at the root

tay1orjones:
Currently in the build output there is not an `index.d.ts` that references all the component's `.d.ts` files. Without this intellisense, etc does not work.

I've modified the root index.js to be `.ts`, which should get rollup to add the root `index.d.ts`. Unfortunately this has highlighted another issue - anywhere we're using 

```js
export default from './Component';
```
needs to be refactored to 
```js
import Component from './Component';
export default Component;
```

#### Testing / Reviewing

- cd into `packages/react`
- run yarn build
- view `es` folder output - an `index.d.ts` should be present alongside the compiled `index.js`


tay1orjones:
Closing in favor of https://github.com/carbon-design-system/carbon/pull/12787

Issue #12241, opened 2022-10-05
[Read-only Inputs]: Epic


sstrubberg:
### User story

> As an `IBM Maker`:

> I need to: build forms that allow Read-only component content to be parsed

> so that: my users can more semantically and accessibly mark-up their forms.

### Additional information

- _#2177_
- Read-only textinput has the work done. Use it as a reference. Make sure the visuals line up.

#### Acceptance Criteria

```[tasklist]
### Update website docs
- [x] https://github.com/carbon-design-system/carbon-website/issues/3248
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3263
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3264
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3265
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3266
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3267
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3268
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3269
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3262
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3270
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3271
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3272
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3273
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3274
```

```[tasklist]
### Update design kits
- [x] https://github.com/carbon-design-system/carbon-design-kit/issues/601
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/632
```

```[tasklist]
### Primary implementation
- [x] https://github.com/carbon-design-system/carbon/issues/12243
- [x] https://github.com/carbon-design-system/carbon/issues/12401
- [x] https://github.com/carbon-design-system/carbon/issues/12409
- [x] https://github.com/carbon-design-system/carbon/issues/12432
- [x] https://github.com/carbon-design-system/carbon/issues/12613
- [x] https://github.com/carbon-design-system/carbon/issues/12419
- [x] https://github.com/carbon-design-system/carbon/issues/12415
- [x] https://github.com/carbon-design-system/carbon/pull/12382
- [x] https://github.com/carbon-design-system/carbon/issues/12407
[...]

PR #12601, opened 2022-11-11
[DO NOT MERGE]: Read only rollup

lee-chase:
This relates to #12241 

Added this to enable reviewers to get an understanding of the full set of read-only components

#### Changelog

**New**

Added a new story `Experimental/ReadOnlyComparrisons

#### Testing / Reviewing

Checked to story works locally.


tay1orjones:
@lee-chase what did you want to do with this one? Should these changes be merged or marked ready for review?

tay1orjones:
@lee-chase I think the value derived from the `All` page will be encompassed by the docs updates to the website listed in https://github.com/carbon-design-system/carbon/issues/12241 These docs will show the comparison between the different states as it relates to the read-only variant.

Besides that, the `readOnly` prop was added to all the Playground stories, so consumers do have a way to preview and understand the prop and how it impacts the component's visual and behavior. So overall I think we could close this for now.

I do believe that we should resurrect this once we have a way to add stories that are included for development purposes only, https://github.com/carbon-design-system/carbon/issues/12971. I'd love to see the `All` story visually snapshotted so we can avoid regressions in the styles across disabled and readonly!

Issue #12475, opened 2022-11-03
Fluid inputs: FluidTimePicker


tay1orjones:
https://github.com/carbon-design-system/carbon/issues/12475#tasklist-block-b5fb7b8c-69a9-4516-9023-9dc0c2cf80c3

## Spec

[Sketch](https://ibm.box.com/s/658frh5yaeyzh6ymrplqtbrllwp7vbdw) | [Figma](https://www.figma.com/file/9oQamCzUmVPw2jTiFkJ3Qw/Fluid-inputs_v11?node-id=329%3A1710)

![Time-picker - White](https://user-images.githubusercontent.com/11670886/200693481-f3248aa0-e5fb-4ce0-87bc-3a107da7f6c6.png)

aagonzales:
Spec update after development concerns.
<img width="1001" alt="image" src="https://user-images.githubusercontent.com/11670886/202216378-bc36968f-b662-4da9-8874-58cbe7b1f75f.png">

PR #12603, opened 2022-11-11
feat(FluidTimePicker): add `FluidTimePicker` variant

### LINKED ISSUE #12478: Fluid inputs: Add FluidTimePicker experimental implementation

### LINKED ISSUE #12479: Fluid inputs: FluidTimePickerSkeleton

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/12475
Closes https://github.com/carbon-design-system/carbon/issues/12478
Closes https://github.com/carbon-design-system/carbon/issues/12479

-Adds in the `FluidTimePicker` component as `unstable`
- It also fixes an issue with `FluidSelect` that caused the focus/invalid styles to disappear when next to an adjacent fluid element
- It also fixes an issue with skeleton styles for all fluid inputs

#### Changelog

**New**
- `unstable__FluidTimePicker` component
- - `unstable__FluidTimePickerSelect` component
- Styles for `FluidTimePicker` (`_fluid-time-picker.scss`)
- Stories for `FluidTimePicker` (`FluidTimePicker.stories.js`)
- Tests for `FluidTimePicker`

**Changed**

- Updates index, snapshots
- Aligns all Fluid input skeleton styles

#### Testing / Reviewing

Go to `FluidTimePicker` and ensure all stories look correct (will prune down once reviewed)
Check all other Fluid component skeleton styles an ensure they all look the same
Check FluidForm story to ensure there are no overlaping focus styles


tw15egan:
@aagonzales most of the styles are there now if you want to take a look. The entire input error/warning states look good, but the single-element variants (when just one element is invalid/warning) have a bit of a funky interaction when hovering. 

Wondering if there is an alternative there; perhaps the borders should stay? Is this likely to be a widely used scenario? Curious what your thoughts are after seeing it in action.

aagonzales:
@tw15egan hmm that does get funky ... My cut instinct is to say let's not allow invalid and warning on individual items and only do the group ones. 

I actually don't know what the use-case would be for only have one of the invalid and not the others, besides to be hyper specific about which part has the error. But I think since its packaged as one component then maybe its ok if these state only have the grouped form.

tw15egan:
@aagonzales I could remove quite a bit of somewhat hacky code if we got rid of the single element invalid, so I'm all for it. If it becomes something that users want, we can always add it in later since it's experimental. 

So should we move forward with only allowing the whole element invalid for now?

aagonzales:
@tw15egan Yes, let's move forward with only allowing the whole element invalid for now. I agree we can always add the individual validation in later if there's a user need.

alisonjoseph (approved):
LGTM!

aagonzales (commented):
When the invalid state is triggered it looks like the type is moving in and down by a few pixels. This will cause issues with type alignment in a form. I'm guessing its because its making room for the invalid + focus state? Is there anything we can do to not make the type jump?
Nov-22-2022 10-41-43

tw15egan:
@aagonzales just pushed up some changes that should resolve any text jumping  👍🏻

aagonzales:
@tw15egan hmm I still see it happening.

tw15egan:
@aagonzales hmm looks like the deploy preview didn't redeploy... might be related to all these VRT checks failing we're seeing on all the PRs

aagonzales:
@tw15egan that looks much better but it still looks like the selects are shifting like a pixel left and right.
Nov-22-2022 16-42-57

tw15egan:
I've manually run the build again, and it should be showing the right commit now 🙄 Have another PR up that should fix this issue as well @aagonzales

aagonzales (approved):
Nice! That did it! Ship it 🚀

tay1orjones (approved):

Issue #12330, opened 2022-10-18
[Feature request] add enhancements to contained list


laurenmrice:
## Enhancements for Contained list

I spoke with Diana Tran from the Sustainability Software Design PAL about some differences between their List and our - Contained list components to try to incorporate some of their use cases into our guidance and specs.

- [PAL Figma examples](https://www.figma.com/file/DXsHD2rqGTcHRfsRKMkW86/PAL-Components-%E2%80%93%C2%A0Sustainability-Software?node-id=10418%3A401657)

- Note: Functionality like multi-select, expandable rows, nesting lists, and batch action bars will remain at the PAL level as an extension.

---

These are the following enhancements we could add to the Contained list component:

### Sizes
For on page list variant
- [x] Add an `Extra large` 64px size (with description text). 
- [x] Add a `Medium` 40px size.
- [x] Add a `Small` 32px size. 

![contained-list-usage-6 copy 4](https://user-images.githubusercontent.com/43969356/203086397-e64b0efb-3540-4cc8-a7c7-f3f2d0b58e05.png)

### Modifiers
For on page list variant and disclosed list variant
- [x] Extended rules: Add a modifier to turn on extended rules versus indented rules.
- [x] Header icon button: Have the ability to add an icon button in the header. (primary and ghost) 

For on page list variant
- [x] Overflow menu: Add overflow menu capabilities in rows. 
- [x] Multiple columns for rows: Include functionality to be able to add multiple columns for rows.
  - [x] For additional row column text, make sure the text says` List item details`.

```[tasklist]
### Tasks
- [ ] #12801
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/583
```

![Contained list—On page—Modifiers](https://user-images.githubusercontent.com/43969356/228867535-e5888085-19ec-47f8-9a3f-d705123e83a0.png)

----

![Contained list—On page—Modifiers (specs)](https://user-images.githubusercontent.com/43969356/198126033-b70c227a-93c6-473b-a49d-2747af0565a3.png)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

PR #12609, opened 2022-11-14
feat(contained-list): add support for sizes

janhassel:
Contributes to #12330 

#### Changelog

**New**

- Added `size` prop to `ContainedList`

#### Testing / Reviewing

Test in playground story
- The default size should be `lg`
- The `props.size` should be applied to all items within a list
- The `props.size` should be applied to the `on-page` variant list header
- The `disclosed` list header should stay at its fixed size regardless of `props.size`


tay1orjones (commented):
LGTM! 🎉

aledavila (approved):

tay1orjones (approved):

Issue #11661, opened 2022-06-22
Plex is not up to date, package is too large


tay1orjones:
We're working to update how Plex is published - https://github.com/IBM/plex/discussions/452. It's likely Plex will start to publish individual packages per-family.

Currently we're blocked by this. The compressed package binary is too large, https://github.com/IBM/plex/issues/453. Once this has been addressed, we need to update packages:

**Checklist**

- [x] Update Plex package(s) and versions
- [x] Determine how we would need to revise the sass font resolver, and then update it
- [x] Ensure no regressions
- [ ] Put together an e2e/percy snapshot for the Plex families Carbon includes (nice to have)

**Workaround**

To use the most up to date Plex assets with Carbon, you can configure the sass to use the Akamai server. It can be configured via the `$use-akamai-cdn` scss config variable.

mjabbink:
@tay1orjones @alisonjoseph Making the note here that the “ﬁ” ligature should be the default setting. Whatever Carbon for IBM.com is doing and what Carbon did with v10 is the correct expectation.

sstrubberg:
@tay1orjones can you remind me where we're at with this? Is this the one we're punting to Brand?

tay1orjones:
@sstrubberg This issue is for the work we need to do in this `carbon` repo, no matter who updates the Plex repo. We can't do this work until the Plex repo is updated and publishes an npm package for each Plex font family.

tay1orjones:
Plex has been coverted to a monorepo, https://github.com/IBM/plex/pull/577, and is now publishing a collection of per-family packages. These packages should be individually small enough to work around this issue.

We need to update the default font resolver to work with the new packages and directory structure.

tay1orjones:
Closed via https://github.com/carbon-design-system/carbon/pull/17025

Thanks again @tw15egan! 🎉

tay1orjones:
The original fix was reverted in https://github.com/carbon-design-system/carbon/pull/17211, we need to revisit this and determine a way to introduce this change without breaking projects that are already configuring `$font-path`

s100:
I'm interested in seeing this fixed. Currently our application has several direct and indirect dependencies which pull in `@ibm/plex@6.4.1`, meanwhile we also use `@carbon/styles` which pulls in `@ibm/plex@6.0.0-next.6`. The IBM Plex font package is large in itself, but having two of them is quite inconvenient.

tay1orjones:
[...]

PR #12614, opened 2022-11-14
Update `@carbon/styles` package.json to use a non-beta version of IBM Plex

### LINKED ISSUE #12608: [Feature Request]: Update `@carbon/styles` to a more recent `@ibm/plex`
### The problem

We're on Carbon 10 and our entire application is standardised on `@ibm/plex@6.0.2` - the latest version of this package is `@ibm/plex@6.1.1` which we could probably also use.

We recently added a dependency which makes use of `@carbon/styles` to our application and we discovered that this added a dependency on `@ibm/plex@6.0.0-next.6`:

https://github.com/carbon-design-system/carbon/blob/4b9b09a55304717307f288d8b588d23294318793/packages/styles/package.json#L42

This dependency is 50.6MB on disk. This is admittedly is a huge improvement over our current `@ibm/plex` which is 194MB on disk... but because we have two Plex installs in play at once, means this isn't an improvement for us, it's a regression of 50.6MB. This increases our install times and disk usage.

### The solution

I suggest that `@carbon/styles` should use the same Plex version as the rest of the Carbon ecosystem, a recent, supported version. If you want to experiment with new beta versions of Plex, I think that should be done using a beta version of `@carbon/styles` itself.

### Examples

_No response_

### Application/PAL

IBM App Connect Enterprise (don't know what a PAL is)

### Business priority

### Available extra resources

I can open a PR to change this version number but I don't know if that suffices.

### Code of Conduct

s100:
Closes #12608.

#### Changelog

**Changed**

- Version range of `@ibm/plex`.

#### Testing / Reviewing

I imagine the existing unit tests for `@carbon/styles` will cover this.

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -39,7 +39,7 @@
     "@carbon/motion": "^11.5.0",
     "@carbon/themes": "^11.12.0",
     "@carbon/type": "^11.11.0",
-    "@ibm/plex": "6.0.0-next.6"
+    "@ibm/plex": "^6.0.2"
   },
   "devDependencies": {
     "@carbon/test-utils": "^10.26.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2279,7 +2279,7 @@ __metadata:
     "@carbon/test-utils": ^10.26.0
     "@carbon/themes": ^11.12.0
     "@carbon/type": ^11.11.0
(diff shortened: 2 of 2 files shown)


s100:

tay1orjones:
Hey thanks for opening this! Unfortunately, updating to the latest version is blocked by #11661

PR #12616, opened 2022-11-14
chore(plex): update plex package, add git lfs

### LINKED ISSUE #12608: [Feature Request]: Update `@carbon/styles` to a more recent `@ibm/plex`

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/12614
Closes https://github.com/carbon-design-system/carbon/issues/12608

Updates `@ibm/plex` to the latest and configures Git Large File Storage to handle the large `.yarn-cache` Plex file.

#### Changelog

**New**

- Git LFS configured

**Changed**

- Bumped `@ibm/plex` to `v6.1.1`

#### Testing / Reviewing

@tay1orjones is there a reason why there is such a large file size increase in the package? ANy way we can reduce this or is this okay as-is

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,9 @@
 # Auto detect text files and perform LF normalization
 * text=auto
-
 # These files should have specific line endings
 *.sh	text eol=lf
 *.bat	text eol=crlf
-
 # These files are binary and should be left untouched
 # (binary is a macro for -text -diff)
 *.jpeg	binary
+.yarn/cache/@ibm-plex-npm-* filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,11 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: '16.x'
+          lfs: true
+      - name: Fetch LFS objects
+        run: git lfs fetch
+      - name: Checkout LFS objects
+        run: git lfs checkout
       - name: Install dependencies
         run: yarn install --immutable --immutable-cache
       - name: Check formatting of project files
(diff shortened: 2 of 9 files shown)


tay1orjones:
> is there a reason why there is such a large file size increase in the package? ANy way we can reduce this or is this okay as-is

@tw15egan Plex turned on a number of font features which caused the file sizes to increase just slightly but enough to get over the 100mb limit git imposes on files like the yarn cache file.

I opened https://github.com/carbon-design-system/carbon/issues/11661 outlining a path forward to resolve this one. There's no way to _decrease_ the size, but we certainly can split it up. The idea being we convert the plex repo to a monorepo and publish packages per font family so users can opt in to each one individually or use the single monolith package.

I'm unclear on if yarn formally supports LFS, but it seems like you've got it working here! Unfortunately our Netlify plan does not include LFS so I don't think we can enable this right now 😞

PR #12621, opened 2022-11-15
docs(contained-list): add usage examples

janhassel:
Contributes to #12330 

Adds a "Usage Examples" story showcasing the flexibility of the "Contained List" component based on what @laurenmrice shared in #12330.

#### Changelog

**New**

- Add "Usage Examples" story to "Contained List"

### DIFF
--- a/packages/react/src/components/ContainedList/ContainedList.stories.js
+++ b/packages/react/src/components/ContainedList/ContainedList.stories.js
@@ -9,17 +9,25 @@ import React from 'react';
 
 import { action } from '@storybook/addon-actions';
 import {
+  Add,
   Apple,
+  ArrowUpRight,
   Fish,
   Strawberry,
   SubtractAlt,
   Wheat,
 } from '@carbon/icons-react';
+
+import { Layer } from '../Layer';
 import { VStack } from '../Stack';
 import Button from '../Button';
 import ExpandableSearch from '../ExpandableSearch';
-import { Layer } from '../Layer';
+import OverflowMenu from '../OverflowMenu';
+import OverflowMenuItem from '../OverflowMenuItem';
 import Tag from '../Tag';
+
+import { usePrefix } from '../../internal/usePrefix';
+
 import mdx from './ContainedList.mdx';
 
 import ContainedList, { ContainedListItem } from '.';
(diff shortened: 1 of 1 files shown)


alisonjoseph (approved):

laurenmrice (changes requested):
Thanks Jan! This is looking great. ⭐️ Just a couple of comments:
* We are not going to support the Medium(40px) size, so we can take that one out.
* For the overflow menu hover states, can you make the hover state background go on top of the divider rules? This seems to be working right in the “With Actions” story https://deploy-preview-12621--carbon-components-react.netlify.app/?path=/story/components-containedlist--with-actions.
* For the usage example with the diagonal arrow/icon button in the header, instead could we just show a ghost icon button with an `add` icon? I know that this is a use case in the wild that people currently need so I would like to show that as another vetted example.

janhassel:
@laurenmrice 

> We are not going to support the Medium(40px) size, so we can take that one out.

Could you fill me in on why the `md` would be skipped? I honestly think it'd be one of the most used sizes as it matches nearby input fields. Also from a consumer perspective if a component supports `sm` and `lg` it would catch me off guard when passing in `md` would not work as expected.

> or the overflow menu hover states, can you make the hover state background go on top of the divider rules?

They are, actually. Just looked into it and the OverflowMenu button uses `background-hover` which has 12% opacity. The ghost button instead uses `layer-hover` which has 100% opacity. You can see the OverflowMenu button overlays the divider rules as they appear slightly darker.
image

> For the usage example with the diagonal arrow/icon button in the header, instead could we just show a ghost icon button with an add icon?

Done ✅

laurenmrice:
@janhassel

**Medium 40px size:**

- We were thinking about not adding the medium 40px size at this time because we have not seen a specific use case for it yet and thought the large 48px size would suffice as a middle ground between the small and extra large sizes. It would also be one less size to maintain in code and in various design tools. If you feel that people would benefit from also using this size we could add it.

**Overflow menu button hover**

- Oh okay, thanks for clarifying. I believe this is a bug in our own code and the overflow menu hover should be updated to use `$layer-hover` instead. I will make a separate issue for this problem. One of our devs can pick it up and the Contained list will inherit the changes eventually.

Thank you!

tw15egan (approved):
LGTM 👍 ✅

laurenmrice:
Ok, let's add it! @janhassel

janhassel:
Great! 👍 

In this case I believe the PR is ready to be merged.
@tay1orjones

laurenmrice (approved):
Looks good to me! Thanks Jan. ✅

PR #12658, opened 2022-11-15
fix(icon-tabs): fix tooltip styles icon only tab

### LINKED ISSUE #12463: [Bug]: `IconTab` tooltip is not visible
### Package

@carbon/react

### Browser

Chrome

### Package version

11.16.0

### React version

17

### Description

The tooltips are not visible on the IconTab stories https://react.carbondesignsystem.com/?path=/story/components-tabs--icon-20-only

If I remove `overflow-x: auto;` from the `.cds--tabs .cds--tab--list` selector, they're visible again. Unsure if this style is required for the tab scroll functionality, or if we could simply remove it.

### Suggested Severity

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tabs--icon-20-only

### Steps to reproduce

Open devtools and hover over a tab - the tab is wrapped in a popover and the popover content/tab/etc is there in the dom. The classes are being switched out with the intent to make it visible, but it's not visible.

### Code of Conduct

aledavila:
Closes #12463 

Makes tootip on Icon tab list visible 
As the issue states the issue stems from the `overflow-x: auto`  on the `tab-list`. Its necessary for scrolling the tab list when its longer than the container its in. The solution here is only based on the assumption that because icons are small there wouldn't be an instance where they are longer than the container so i removed it from icon only tab lists. I tried many other solutions that weren't working including z-index but it doesn't work. If this doesn't feel like the correct solution I'm also open to suggestions 

#### Testing / Reviewing

Make sure theres no regressions


tay1orjones (approved):
Having a ton of icon tabs that need to be scrollable is far more of an edge case than having the tooltip visible for a single tab. 

I agree we should prioritize the tooltips over the scrolling functionality if the two are incompatible.

alisonjoseph (approved):

PR #12659, opened 2022-11-15
ci(project): move add-to-project job into a separate workflow

tay1orjones:
This moves the `add-to-project` job into its own workflow file. This way we can more granularly control when it's ran. 

In the previous set up, if you had removed an issue from the project and then commented on the issue, the workflow would run again moving it back into the project.

With this change, issues will be automatically added to the project only when the issue is created or labelled.

#### Changelog

**New**

- add `add-to-project` workflow file

**Removed**

- removed `add-to-project` from issue triage workflow file


tw15egan (approved):

aledavila (approved):

.github/workflows/add-to-project.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  add-to-proposals-project:
+    name: Add issue with enhancement label to the Proposals project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v0.4.0
+        with:
+          labeled: ${{ env.LABEL_ENHANCEMENT }}
+          project-url: ${{ env.PROPOSALS_PROJECT_URL }}
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
```
tay1orjones, line 26: @sstrubberg Based on our conversations yesterday, I just added this here that will add issues with the https://github.com/carbon-design-system/carbon/labels/type%3A%20enhancement%20%F0%9F%92%A1 label to the Proposals project. These will no longer be added to the Design System project by default. Let me know what you think!

PR #12711, opened 2022-11-22
feat(tile): add experimental improved contrast tile

### LINKED ISSUE #12464: Experimental tile: code implementation
This issue is for the primary implementation of the new experimental tile variant.

At it's core, the required changes are likely just styling placed behind a new classname/selector.

There's a few ideas on the "toggle mechanism" we could use to allow consumers to flip on the new styling:
- ~via a new `hasImprovedContrast` (naming tbd) prop on `Tile`~
- ~Through a whole new component exports, `unstable__Tile`, `unstable__SelectableTile`, etc.~
- ~via a context wrapper component (feature flag) - this would enable certain sections of an app's tiles to use the new variant~
- Add a new sass feature flag, off by default

Really it depends on feedback we get:

- Is individual opt-in important?
  - A: No
- Is opting-in to an entire section (all?) of an app more desirable?
  - A: Yes
- Is it allowable from a design-consistency standpoint to split the design and have portions of the same app using both tile designs at the same time?
  - A: No - we do not want old and new styling mixed in an application. It should be one or the other.

https://github.com/carbon-design-system/carbon/issues/12464#tasklist-block-1a5a0911-1803-46ec-b64f-4f291f1d8f1d

alisonjoseph:
closes #12464

#### Changelog

**New**

- add`$border-tile` layer contextual token (`$border-tile-01`,$border-tile-02`,$border-tile-03`)
- add `enable-experimental-tile-contrast` feature flag
- add experimental tile styles
- add Experimental Improved Contrast story

#### Testing / Reviewing

https://deploy-preview-12711--carbon-components-react.netlify.app/?path=/story/components-tile-experimental-improved-contrast--clickable


packages/react/src/feature-flags.js
```diff
@@ -11,4 +11,5 @@ FeatureFlags.merge({
   'enable-css-custom-properties': true,
   'enable-css-grid': true,
   'enable-v11-release': true,
+  'enable-experimental-tile-contrast': true,
```
tay1orjones: This needs to be `false` by default, right?

packages/styles/scss/_feature-flags.scss
```diff
@@ -11,6 +11,7 @@
     'enable-use-controlled-state-with-value': true,
     'enable-css-grid': true,
     'enable-v11-release': true,
+    'enable-experimental-tile-contrast': true,
```
tay1orjones: Same here - I think this needs to be `false` by default?

packages/react/src/feature-flags.js
alisonjoseph: Yep, this was just for testing the styles.

packages/react/src/components/Tile/Tile.mdx
```diff
@@ -52,6 +52,32 @@ import { Story, Props, Source, Preview } from '@storybook/addon-docs';
...
 </Preview>
 
+## Experimental Improved Contrast
+
+In Sass, you can enable the improved contrast tiles via a feature flag.
+
+```scss
+@include feature-flags.enable('enable-experimental-tile-contrast');
+```
+
```
alisonjoseph, line 62: @tay1orjones is this the correct way to do this? I wasn't sure.

tay1orjones: Yeah I think that should work. The other way I had working for testing was

```
@use '@carbon/react/scss/feature-flags' with (
  $feature-flags: (
    'enable-experimental-tile-contrast': true,
  )
);
```

tay1orjones (approved):
Looks great to me! 🚀

tay1orjones (commented):
What do you think of adding the improved contrast stories to the e2e test file `e2e/components/Tile/Tile-test.e2e.js` so they're snapshotted w/ Percy?

alisonjoseph:
@tay1orjones ah yea, I'll do that 👍

tw15egan (approved):
LGTM 👍 ✅

thyhmdo:
Looks good to me. Just one thing I'm not sure. So we'd have the experimental tile under the tile component instead of having it under the Experimental section? 
image
image

alisonjoseph:
@thyhmdo we could certainly move it under experimental, I think the logic was that its an experimental style change, not something fundamental to how the component is built. @tay1orjones might have had some more thoughts around that too.

thyhmdo:
Thanks @alisonjoseph In term of consistency, can we move this under the experimental section for now? This may become the official style in the future. It's probably confusing if as they're variants of tiles.

alisonjoseph:
@thyhmdo updated!

thyhmdo (approved):

PR #12757, opened 2022-12-01
fix(NumberInput.js): assign min and max value to default props

### LINKED ISSUE #12695: [Bug]:  NumberInput error if min or max props are not present
### Package

@carbon/react

### Browser

Chrome

### Package version

1.17.0

### React version

17.0.1

### Description

Using the NumberInput component, the onChange function doesn't work, it returns the following error 'Warning: Received NaN for the value attribute. If this is expected, cast the value to a string'.

I looked at the following issue #11926, where a user had the same problem and in the CodeSandbox posted as a solution, it seems to work. However, I realized that in the working solution min & max props are present and are missed in the non-working solution.

Looking closely at your code, I noticed that in the NumberInput.js component if the min and max are not passed as props, they are not assigned any default value. In my case, I didn't have a value of max and so this causes the onChange to fail.

Looking at the documentation, the min and max props are optional and therefore this causes the user this error, without being able to understand the reason for the NaN.

### Suggested Severity

### Reproduction/example

https://codesandbox.io/s/upbeat-sinoussi-wgmqg1?file=/src/App.js

### Steps to reproduce

The error occurs if either min or max is not set.

### Code of Conduct

darioplatania:
Closes #12695 

In the open issue, a problem was found in the component `NumberInput.js`, when the min or max value is not passed.

#### Changelog

**New**

- assigned default value to min and max if they are not passed via props

**Changed**

- `min` has 0 and `max` has 100

#### Testing / Reviewing

As described in issue #12695 that I opened, without a default value in min or max, the `onClick` function fails. You can test the unwanted behavior without passing either value. This change I made assigns a default value to both the min and max, so the `clamp` function inside onClick does not fail and NumberInput it does not generate errors.


darioplatania:

darioplatania:
recheck

darioplatania:
Hello team, yesterday during the PR I didn't realize my `user.name` and `user.email` wasn't set in the git config but now I've put them. I read the **Developer Handbook** (in the commit conventions) and I didn't see the possibility of generating an empty commit like in this case where I just changed my global git configurations. How can I proceed? Do I create a new PR? Thanks in advance for the help.

tw15egan:
I've just updated the branch with the latest changes from `main`, and perhaps you could update your local branch with these changes and push it back up with the git config changes. If that doesn't work, I would just create a new PR and close this one.

darioplatania:
> I've just updated the branch with the latest changes from `main`, and perhaps you could update your local branch with these changes and push it back up with the git config changes. If that doesn't work, I would just create a new PR and close this one.

Thanks for the reply. Unfortunately, it doesn't work, I close this PR and create a new one.

PR #12767, opened 2022-12-02
test(Slider): add additional RTL tests

### LINKED ISSUE #12676: Use RTL for testing in `Slider/Slider-test.js`

matthewgallo:
Closes #12676 

This PR adds RTL tests, building upon the `readOnly` tests that Lee started, for the Slider component.

I ran into an issue/may have potentially found a bug while testing the `onInputKeyUp` prop, receiving an error saying `Unknown event handler property 'onInputKeyUp'. It will be ignored.` I got this error message when testing this component within the Carbon repo and also adding this prop in a sandbox environment (using latest `@carbon/react` release). Because of this I've left out the test for that prop.

#### Changelog

**Changed**

- `Slider-test.js`

**Removed**

- `Slider/__test__/Slider-test.js` (moved this into `Slider/Slider-test.js`)

#### Testing / Reviewing

Ran all tests locally
image


packages/react/src/components/Slider/Slider-test.js
```diff
@@ -1,441 +1,480 @@
...
+          ariaLabelInput: inputAriaValue,
+          value: initialValue,
+          max: 100,
+          onChange,
+        });
+        const theSlider = screen.getByRole('slider');
+        click(theSlider);
+        type(theSlider, '{arrowright}');
+        type(theSlider, '{arrowright}');
+        expect(onChange).toHaveBeenCalledTimes(3);
```
tay1orjones: @matthewgallo It looks like this one is failing in CI, `onChange` is being called 4 times.

matthewgallo: Hey @tay1orjones, was able to get tests all passing in CI now, but Percy seems to still be failing. Any suggestions there?

packages/react/src/components/Slider/Slider-test.js
```diff
@@ -1,441 +1,489 @@
...
+        const { type, click } = userEvent;
+        renderSlider({
+          ariaLabelInput: inputAriaValue,
+          onChange,
+        });
+        const theSlider = screen.getByRole('slider');
+        click(theSlider);
+        type(theSlider, '{arrowright}');
+        type(theSlider, '{arrowright}');
+        expect(onChange).toHaveBeenCalledTimes(2);
```
tw15egan: Seems like the initial click for focus is causing an `onChange`. Can we test `onKeyDown` instead? Or do we just change the expected to `3`

packages/react/src/components/Slider/Slider-test.js
matthewgallo: This got cleared up after merging main.

packages/react/src/components/Slider/Slider-test.js
matthewgallo: Ah, thanks for the suggestion! `onKeyDown` seems to be working

tw15egan (approved):

tay1orjones (approved):

PR #12768, opened 2022-12-02
fix(NumberInput.js): assign min and max value to default props

darioplatania:
Closes [#12695](https://github.com/carbon-design-system/carbon/issues/12695)

In the open issue, a problem was found in the component NumberInput.js, when the min or max value is not passed.

#### Changelog

**New**

* assigned default value to min and max if they are not passed via props

**Changed**

* min has 0 and max has 100

#### Testing / Reviewing

As described in issue [#12695](https://github.com/carbon-design-system/carbon/issues/12695) that I opened, without a default value in min or max, the onClick function fails. You can test the unwanted behavior without passing either value. This change I made assigns a default value to both the min and max, so the clamp function inside onClick does not fail and NumberInput it does not generate errors.


darioplatania:

darioplatania:
recheck

tay1orjones:
@darioplatania The bot unfortunately needs a comment only containing the text "I have read the DCO document and I hereby sign the DCO." once, without the quote. Sorry for the hassle

tay1orjones (approved):
Thanks for contributing this!

darioplatania:
> @darioplatania The bot unfortunately needs a comment only containing the text "I have read the DCO document and I hereby sign the DCO." once, without the quote. Sorry for the hassle

No problem, I'll answer as you always wrote to me in this PR. Thanks for your help

tw15egan (approved):
Thanks for adding this in!

vasharma05:
Doesn't making the max value 100 a very limiting number for the user, who wants to use the number input with unknown min and max values.

I believe this is a short term answer, and the exact solution to this would be to solve this issue handling the dependency on the min and max values.

darioplatania:
@vasharma05 your answer is correct. There're two ways in my opinion:

1. assign `Infinity` to the max value (on MDN it's an integer value greater than any other number). In this way, the number 100 is not limiting and could also not be set as props when you don't have a maximum value.

2. do not assign any default props to min and max and handle the _undefined_ case in the `clamp` function

@tw15egan @tay1orjones what do you think? Could this be more correct? Thanks in advance for your reply

dabrad26:
This is a major breaking change.  Defaulting max to 100 out of nowhere has caused us to have to revert the carbon updates to work on a solution.  Changing default props does not seem like the proper approach.  We only caught this when QA started reporting unable to input numbers higher than 100 even though we were sure we were not setting them.

hsu022210:
Opened a bug ticket at https://github.com/carbon-design-system/carbon/issues/13243

vasharma05:
Thanks @tay1orjones for re-opening this and looking into this. We at OpenMRS were also seeing some issues related to this. 

Please feel free to connect for anything.
CC: @denniskigen

PR #12131, opened 2022-09-21
refactor(react): update tests and move component from next

abbeyhrt:
Closes #12087 

This PR:

- Moves `/next` ProgressIndicator to `/ProgressIndicator` 
- updates tests to RTL
- adds to build test script to prompt user to copy test file if the selected component doesn't have a test file. 
- 
#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Ensure components are exported correctly and their API is fully tested. 

If you feel like it, pull down the PR and try out the test script and select a component that doesn't have its own test file (such as ProgressStep or any of the Tile components) and make sure that the file is copied to your clipboard

PR #12782, opened 2022-12-05
test(SelectItemGroup): convert Enzyme to RTL for SelectItemGroup

cknabe:
Closes #https://github.com/carbon-design-system/carbon/issues/12160

image


packages/react/src/components/SelectItemGroup/SelectItemGroup-test.js
```diff
@@ -7,37 +7,56 @@
...
-    );
+    const renderSelectItemGroup = (props) =>
+      render(
+        <SelectItemGroup
+          className="extra-class"
+          data-testid="select-item-group"
+          label="testLabel"
+          {...props}
+        />
+      );
```
tay1orjones: Our test files usually don't follow this pattern and instead call `render` directly in each `it()` block. It's more duplication, but the benefit we've found is that it's easier to maintain as the tests grow. Changing a single test due to future API changes is simpler when there's no abstraction.

We don't have the eslint rule `testing-library/no-render-in-setup` on yet, but it's worth noting this _will still pass_ the rule, so it's not a hard blocker.

tay1orjones (commented):
Thanks for taking this on! I see just a few style nits that we may not have clearly documented yet. One in a comment below. 

Also, we've been trying to avoid using `getByTestId` and instead [prioritize using queries accessible to everyone](https://testing-library.com/docs/queries/about#priority). In some cases this isn't possible, and `container` can be used, [like here](https://github.com/carbon-design-system/carbon/pull/12131/files#diff-74ff05bffc7367c5a1f68a010058a7bfb36ce5d5a0c88870c6a3b4ad851d98f4R15-R32).

Most of this is sort of laid out for us by [a script we've been using](https://github.com/carbon-design-system/carbon/issues/10184#issuecomment-1282894448) when refactoring these so our test files start with the same general format/template. It would put most of this in place for you.

packages/react/src/components/SelectItemGroup/SelectItemGroup-test.js
cknabe: Sure, I will refactor away all of the render* methods I have added to tests. As for the other stuff, I noticed that container was used heavily, so I started down that path. However, I felt guilty for taking a shortcut. I have been following their priority queue, and they recommend using test ids before containers.

> In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. **Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names.** Learn more about data-testids from the blog post ["Making your UI tests resilient to change"](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change)

I guess it doesn't explicitly state "don't use containers", but I think most of the container usage I've seen was to access DOM structure or class names. 

I used Abbey's script on one of the tests I wrote, but it felt like more work than if I just converted over the existing test. I can go back and scrub the tests I have done with her script.

I have a feeling most (if not all) of the container references in Carbon tests can be refactored to use test ids instead of containers.

cknabe: e.g. Abbey's script produces this:
```
    it('should support a custom `className` prop on the outermost element', () => {
      const { container } = render(<SelectItemGroup className="custom-class" />);
      expect(container.firstChild).toHaveClass('custom-class');
    });
```

Is not using container to get the firstChild, based on the DOM structure? When using getByTestId is higher in the priority queue?
`expect(screen.getByTestId('test-id')).toHaveClass('custom-class')`

tay1orjones: Hey so sorry for the delay here! Somehow this slipped out of my review queue.

Yeah the intent from that test within Abbey's script is that we want to ensure we're following the pattern we've agreed on, which is to only apply classnames to the outermost element of a component. It's one of the only test patterns where we specifically _want_ to test the DOM structure.

Technically a `data-testid` could be placed anywhere, whereas `container` is static. `container.firstChild` more confidently asserts that the outermost element is receiving the classname.

tay1orjones (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #12784, opened 2022-12-05
fix: remove inline styles

### LINKED ISSUE #12386: [Bug]: SkeletonText has an inline style that breaks CSP in SSR
### Package

carbon-components-react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v7.59.4

### React version

v17.0.2

### Description

In SkeletonText, there is a `width` property that you cannot nullify it. It is default to `100%` if not specified. This `width` prop turns out to be an inline style. When using in SSR, I will then see this violation.

```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'nonce-1ab3c34441e11a4377e7b34384251a01'". Either the 'unsafe-inline' keyword, a hash ('sha256-Q5Zbi1HJmWC2e721HDLiKbPwTXmvBEI2Dttcb1cd/1w='), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present.
```

Can we have a way to avoid setting the `width` prop which will NOT add an inline styling?

### Suggested Severity

### Reproduction/example

N/A

### Steps to reproduce

Render this piece of code:

```
                            <div className="details-page-cluster-container">
                                <div className="createArtifactFlexGroupRow form-row half-width">
                                    <SkeletonText />
                                </div>
                            </div>
```

You will see an inline style of `width: 100%` added to the DOM. When turning on CSP, this will violate the rule if it's coming from SSR because the DOM contains an inline style.

### Code of Conduct

### LINKED ISSUE #12580: [Bug]: Inline styles blocked by CSP  `style-src` header
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

^1.5.0

### React version

v17

## Background

IBM console / cloud has started to implement the CSP header ( https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP ). Part of that is it blocks inline styles, 

```
<style>
// css here
<style>

```
will be blocked unless it has a `nonce` value passed in by the consumer

where 

`<div style=....></div> `

will be blocked no matter what

this is problematic since all of ibm cloud ios moving to CSP, some inline styles (second example) will be blocked by CSP and break the component to some limit.

<details>
  <summary><h4>Here is some that I personally searched for and found in the carbon repo (click to expand)</h4></summary>
  
## DatePicker

with csp 
image

without 
image

image

image

image

image

image

image

image

This one for some reason doesnt give me a warning but if you look here theres a style property that gets added (i changed it to height to show an example)

image

but csp can affect browsers differently so it should be something to look into 

## SkeletonText / Slider / ProgressBar 

These also have inline styles that may be blocked by csp so worth a look

## Other 

image
</details>

### Suggested Severity

### Reproduction/example

N/A

### Steps to reproduce

I am testing on my local microservice for ibm cloud with csp to verify 

DatePicker

### Code of Conduct

https://github.com/carbon-design-system/carbon/issues/12580#tasklist-block-a9785d44-ab13-438b-8bb3-d2ba230a2a72

alisonjoseph:
closes #12580
closes #12386 

- [x] Remove inline styles from DataTable - TableHeader
- [x] Remove inline styles from ProgressBar
- [x] ~Remove inline styles from Stack~ This should be fine as is, no actual styles
- [x] Remove inline styles from TextArea
- [x] Remove inline styles from Tile
- [x] Remove inline styles from TooltipIcon
- [x] Remove inline styles from Fileuploader
- [x] Remove inline styles from DatePicker arrow icons
- [x] Remove inline styles from SkeletonText
- [x] Remove inline styles from Slider
- [x] Determine if third party library inline styles can be removed
  - removed inline styles from Datepicker SVGs, are there others?

#### Changelog

**Changed**

- remove inline style tags from all components and convert styles to CSSOM

#### Testing / Reviewing
Nothing should have changed for the following components
- DataTable (specifically TableHeader)
- ProgressBar
- TextArea
- Tile
- TooltipIcon
- FileUploader
- DatePicker arrows
- SkeletonText (almost complete refactor, pls double check this one)
- Slider


alisonjoseph:
fyi @TannerS hoping this works 🙏

packages/react/src/components/Accordion/__tests__/__snapshots__/Accordion.Skeleton-test.js.snap
```diff
@@ -51,14 +51,12 @@ exports[`AccordionSkeleton should render 1`] = `
...
-            className="cds--skeleton__text cds--accordion__title"
-            style={
-              Object {
-                "width": "100%",
-              }
-            }
-          />
+          <div>
+            <p
+              className="cds--skeleton__text cds--accordion__title"
```
tay1orjones: AccordionItemSkeleton doesn't look correct in the deploy preview
image

packages/react/src/components/TooltipIcon/TooltipIcon.js
```diff
@@ -13,6 +13,7 @@ import { composeEventHandlers } from '../../tools/events';
 import { keys, matches } from '../../internal/keyboard';
 import toggleClass from '../../tools/toggleClass';
 import { usePrefix } from '../../internal/usePrefix';
+import useIsomorphicEffect from '../../internal/useIsomorphicEffect';
 
 const getInstanceId = setupGetInstanceId();
 const TooltipIcon = ({
```
tay1orjones, line 19: We actually don't even export this component anymore, and it's not in the storybook. I think we forgot to remove the files after v11.

tay1orjones: https://github.com/carbon-design-system/carbon/blob/607b441688769e4c0311e8ddc00f5ba1f51e4c14/docs/migration/v11.md#tooltips:~:text=TooltipIcon%20has%20been%20removed%2C%20use%20IconButton%20instead

tay1orjones (changes requested):
I compared storybooks and everything looks great except the Accordion skeleton text

packages/react/src/components/TooltipIcon/TooltipIcon.js
tay1orjones: I opened an issue https://github.com/carbon-design-system/carbon/issues/12824

packages/react/src/components/Accordion/__tests__/__snapshots__/Accordion.Skeleton-test.js.snap
tay1orjones: Percy didn't catch this because this story is not included in the e2e tests. Only the default story is:

https://github.com/carbon-design-system/carbon/blob/005e7a595afd7edccbffe5d9fce50a370baa6615/e2e/components/accordion/accordion-test.e2e.js#L15-L25

We might want to double check every component that uses SkeletonText

alisonjoseph: I’ll take a look, I updated a lot of logic in that component so I must have missed something.

alisonjoseph: fixed, there was an extra div being rendered that didn't need to be that was throwing off styles. Should be rendering exactly as it was before now.

tw15egan:
@alisonjoseph I thin snapshots just need to be updated now

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #12787, opened 2022-12-05
fix(typescript): Refactor carbon/react index.js to index.ts to expose new typescript typings

mbarrer:
Convert the main index.js to a `ts` file to expose the new typings we are adding for each component to the consumer. 

#### Changelog

**Changed**
- Renamed `src/index.js` to `src/index.ts`
- The majority of exports in the index.ts now use `export * from`. This will automatically export every named export declared in the file. 
- Default export references maintained in each `Component/index.js` file 

#### Testing / Reviewing

All existing component imports for the storybook should be valid and rendering. In a typescript environment, you should be able to import and reference the `Checkbox` component and its typings (`CheckboxProps`)


mbarrer:
@tay1orjones Updated, is that command generally something we should be running for every PR?

tay1orjones (commented):
Trying to view the public api snapshot, and the diff is huge it won't load:

image


This file is built by looping over the keys of an object containing all the exports. My assumption is that the type of these objects has shifted some, and caused the exports to reorder, resulting in the huge diff. I think we could try to get a PR in to sort these first, then update this PR with that so the diff here might be more manageable. 

It worries me to merge this without being able to verify what has changed in the public api snapshot file.

tay1orjones:
> Updated, is that command generally something we should be running for every PR?

Yeah when adding types, I think the public api snapshot may change due to additions. It will likely need updated for every PR adding types for components.

packages/react/src/components/Toggle/Toggle.Skeleton.js
```diff
@@ -10,7 +10,7 @@ import React from 'react';
 import cx from 'classnames';
 import { PrefixContext } from '../../internal/usePrefix';
 
-export default class ToggleSkeleton extends React.Component {
+export class ToggleSkeleton extends React.Component {
```
jharvey10: What are your thoughts on if we should follow the same convention as the other files, having the `export { ToggleSkeleton }` at the bottom and just `class ToggleSkeleton ....` up here?

packages/react/src/components/ToggleSmall/ToggleSmall.Skeleton.js
```diff
@@ -10,7 +10,7 @@ import React from 'react';
 import cx from 'classnames';
 import { PrefixContext } from '../../internal/usePrefix';
 
-export default class ToggleSmallSkeleton extends React.Component {
+export class ToggleSmallSkeleton extends React.Component {
```
jharvey10: Same question/comment here about having the export statements at the bottom of the file.

packages/react/src/components/UIShell/index.js
```diff
@@ -5,35 +5,35 @@
...
+export { default as HeaderMenuButton } from './HeaderMenuButton';
+export { default as HeaderMenuItem } from './HeaderMenuItem';
+export { default as HeaderName } from './HeaderName';
 export { HeaderNavigation } from './HeaderNavigation';
-export HeaderPanel from './HeaderPanel';
-export HeaderSideNavItems from './HeaderSideNavItems';
-export Switcher from './Switcher';
-export SwitcherItem from './SwitcherItem';
-export SwitcherDivider from './SwitcherDivider';
+export { default as HeaderPanel } from './HeaderPanel';
```
jharvey10, line 19: what are your thoughts about also exporting `HeaderNavigation` and `SideNavMenu` as default exports in their respective files so that they can follow the same paradigm here in this file?

packages/react/src/index.ts
```diff
@@ -0,0 +1,167 @@
...
+export * from './components/SkeletonPlaceholder';
+export * from './components/SkeletonIcon';
+export * from './components/DataTableSkeleton';
+export * from './components/Pagination/Pagination.Skeleton';
+export * from './components/RadioButton/RadioButton.Skeleton';
+export * from './components/Tag/Tag.Skeleton';
+export * from './components/Toggle/Toggle.Skeleton';
+export * from './components/ToggleSmall/ToggleSmall.Skeleton';
+export * from './components/Icon/Icon.Skeleton';
+export * from './components/UIShell';
```
jharvey10, line 93: This is so clean. I love it.

tay1orjones (approved):
@mbarrer I hope you don't mind, I pushed an update here that will alphabetize exports and component api for the public api snapshot. This brought everything in line and now shows a much more small (and expected! 😅 ) diff. 

Pending @jdharvey-ibm's comments, I think this looks good to go.

packages/react/src/index.ts
tay1orjones: Same 🤩

mbarrer:
@jdharvey-ibm Updated with those suggestions. Overall I think this looks pretty darn clean! 😄

aledavila (approved):

PR #12789, opened 2022-12-06
fix(getInteractiveContent): add null checks

tay1orjones:
Based on some conversations with @fbarroso24: When using `react-test-renderer` to test components using this hook/function, the component will error with

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

This PR adds some null checks to avoid iterating over non-iterable Symbols passed to this function stemming from the `react-test-renderer` environment.

#### Changelog

**Changed**

- modify `getInteractiveContent` to include null checks

#### Testing / Reviewing

- Check to make sure Popover, Tooltip, IconButton stories are not be impacted and work as intended
- @fbarroso24 has already tested this fix locally to ensure it fixes the issue he's reported

### DIFF
--- a/packages/react/src/internal/useNoInteractiveChildren.js
+++ b/packages/react/src/internal/useNoInteractiveChildren.js
@@ -33,6 +33,10 @@ export function useNoInteractiveChildren(
  * @returns {HTMLElement}
  */
 export function getInteractiveContent(node) {
+  if (!node || !node.childNodes) {
+    return null;
+  }
+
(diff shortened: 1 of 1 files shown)


fbarroso24 (approved):
Thanks @tay1orjones!

tw15egan (approved):
LGTM 👍 ✅

alisonjoseph (approved):

PR #12795, opened 2022-12-06
feat(contained-list): add support for indented rulers

janhassel:
Contributes to #12330 

Makes contained list rulers extended by default and adds option to indent them.

#### Changelog

**New**

- Add prop `isInset` to `ContainedList`

#### Testing / Reviewing

- By default the lines in between list items should extended all the way to the edge of the component container
- When `props.isIndented` is passed, the lines in between list items should be indented by 16px and align with the list item's text
  - The header line is not affected by this

I'm not the biggest fan of the name `isIndented` but couldn't find anything comparable in the system. Please let me know if you have any good ideas! (We've switched it to `isInset`) - TE


tw15egan (approved):
LGTM 👍 ✅

laurenmrice:
@janhassel We could also say `isInset` if you like that terminology better?

janhassel:
@laurenmrice Yes, I like that more. Updated 👍

laurenmrice (approved):
Looks great! Thanks Jan ⭐️

PR #12700, opened 2022-11-18
test(SecondaryButton): replace Enzyme with RTL

cknabe:
Closes #12157

image

PR #12805, opened 2022-12-06
test(SecondaryButton): Enzyme->RTL - refactor out the render in setup

cknabe:
Relates to https://github.com/carbon-design-system/carbon/pull/12700
- In the above PR, I applied a pattern to the SecondaryButton-test where I created a renderSecondaryButton method. This is not the correct style. A render method should just be declared in every test statement.
- I also had removed the prefix.. and I think that may have been a bad decision. To keep consistency, I added that back in.
- I attempted to run Abbey's script to generate the tests, but because this is an inherited button type, the script doesn't find 'propTypes' and does not run. Abbey's script is more for the parent button anyway.


tw15egan (approved):
Thanks for fixing that 🙏

alisonjoseph (approved):

PR #12807, opened 2022-12-06
DO NOT MERGE: chore(carbon-platform): migrate StorybookDemo instances to new API

francinelucca:
Co-authored-by: Andrea Cardona <Andrea.Cardona@ibm.com>

Preparing this for when platform's https://github.com/carbon-design-system/carbon-platform/pull/1497 is merged and the `StorybookDemo` component's API changes :)

### DIFF
--- a/packages/react/src/components/Accordion/docs/overview.mdx
+++ b/packages/react/src/components/Accordion/docs/overview.mdx
@@ -3,10 +3,6 @@
 <StorybookDemo
   themeSelector
   url="https://react.carbondesignsystem.com"
-  variants={[
-    {
-      label: 'Default',
-      variant: 'components-accordion--default'
-    }
-  ]}
-/>
+>
+  <Variant label="Default" variant="components-accordion--default" />
+</StorybookDemo>
--- a/packages/react/carbon.yml
+++ b/packages/react/carbon.yml
@@ -24,9 +24,9 @@ library:
         - title: Step 5
           path: 'https://github.com/carbon-design-system/carbon-website/blob/carbon-platform/src/pages/developing/react-tutorial/step-5.mdx'
         - title: Wrapping up
-          path: 'https://github.com/carbon-design-system/carbon-website/blob/main/src/pages/developing/react-tutorial/wrapping-up.mdx'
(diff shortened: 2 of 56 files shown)


aledavila (approved):

packages/react/src/components/Accordion/docs/overview.mdx
```diff
@@ -3,10 +3,6 @@
...
-  variants={[
-    {
-      label: 'Default',
-      variant: 'components-accordion--default'
-    }
-  ]}
-/>
+>
+  <Variant label="Default" variant="components-accordion--default" />
+</StorybookDemo>
```
jharvey10, line 8: @francinelucca this new API looks so nice!
What are your thoughts on the prop name of `variant` since the component is named also named `Variant`? Should we find a more specific name for the prop?

francinelucca: should we name it `path` instead? 'cause it is the path the variant lives in inside the `StorybookDemo` `url`

tay1orjones: @francinelucca You might consider `id` or a more descriptive `storyId`. As a reference, [our e2e tests](https://github.com/carbon-design-system/carbon/blob/b53e3e9e02a18e4cf94b9f90e31caf4e979ac05b/e2e/test-utils/storybook.js#L30-L31) visit stories via url like this and we mirror the pattern that storybook itself references/names these as.

`storyId` (or simply `id`) consists of a `component` and `story`:

```js
const component = `accordion`;
const story = `default`;
const storyId = `components-${component}--${story}`;
```

tay1orjones (approved):

packages/react/src/components/Accordion/docs/overview.mdx
alisonjoseph: I like `storyId` for the prop name.

francinelucca: same! I think we should go with `storyId` :)

francinelucca:
this is dependent on the [rmdx stuff](https://github.com/carbon-design-system/carbon-platform/pull/1497), I can reopen when we're ready!

PR #12821, opened 2022-12-09
style(Modal): scope p element styles to top level

### LINKED ISSUE #10098: [Bug]: Inline Notifications inside Modals CSS is messed up
### Package

### Browser

Safari

### Package version

10.46.0, React: 7.46.0

### Description

Inline Notifications Inside a Modal have messed up styling due to https://github.com/carbon-design-system/carbon/blob/2da12d0d99e45563d78431a0e6a19df92ac1ff54/packages/components/src/components/modal/_modal.scss#L134 targeting the `p` tag that Inline Notifications uses.

Probably avoid targeting tags or add `:not` for all possible parent components that can render in modal and use `p` (or switch InlineNotifications to div instead of `p`...

Screen Shot 2021-11-11 at 3 32 49 PM
Screen Shot 2021-11-11 at 3 33 49 PM

Product: IBM Aspera Starter

### CodeSandbox example

https://codesandbox.io/s/youthful-framework-cp4zh

### Steps to reproduce

Add InlineNotificaiton to Modal.
See Title/Subtitle in notification styling is wrong.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/10098

Scopes styles a bit tighter so that elements further down the cascade are not inadvertently affected. 

#### Changelog

**Changed**

- added a `>` so that only top level `p` elements have styles applied

#### Testing / Reviewing

Ensure there are no visual regressions


alisonjoseph (approved):
Looks good to me, I do have a slight concern someone might be using this component and counting on these styles for any paragraphs they have inside the modal? However it makes sense not to do that. 😬

sstrubberg (approved):

PR #12835, opened 2022-12-12
feat: enable onblur method in inputs components

### LINKED ISSUE #12834: [Feature Request]:  onBlur method in TextInput component
### The problem

I have a use case that, after user fill some TextInputs, I need to make a request to calculate a value for my application. Right now, I am using onChange to call validator function and, after, request the value from API. This is slowing down our application and decreasing usability.

### The solution

We would also like to be able to listen to the blur event for TextInput component adding method onBlur to possible props.

### Examples

_No response_

### Application/PAL

Supplier and XFuel

### Business priority

### Available extra resources

I already done all necessary changes to this PR.

### Code of Conduct

niveamartins:
Closes #12834 

This PR adds onBlur method to all TextInput types.

#### Changelog

**New**

- now, dev can set a function to be called when focus out in an input setting the function on onBlur props.


tay1orjones:
https://github.com/carbon-design-system/carbon/issues/12834#issuecomment-1348735763

> `onBlur` should already be supported as we spread any extra props to the underlying input. Here is an example:
>
> https://stackblitz.com/edit/github-dodfnj?file=src/App.jsx

I'm going to close this as I think this is the appropriate resolution. Let us know if not and we can reopen!

PR #12840, opened 2022-12-13
fix(FluidInput): update fluid input labels to prevent wrapping

### LINKED ISSUE #12825: [Bug]: Fluid Input labels should scroll when text is long instead of breaking to new line
### Package

@carbon/styles

### Browser

Chrome

### Package version

Latest

### React version

_No response_

### Description

Refs https://github.com/carbon-design-system/carbon/pull/12737#pullrequestreview-1211952102

WRONG: 
image

Components that need to be updated:

- [ ] FluidComboBox
- [ ] FluidDatePicker
- [ ] FluidDropdown
- [ ] FluidMultiSelect
- [ ] FluidNumberInput
- [ ] FluidSelect
- [ ] FluidTextArea
- [ ] FluidTimePicker

### Suggested Severity

### Reproduction/example

Storybook

### Steps to reproduce

Add text to label

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12825

Applies styles to labels to truncate text and prevent wrapping

#### Changelog

**New**

- Styles to truncate text for all Fluid Inputs

#### Testing / Reviewing

Ensure text truncates on long strings and that `Tooltip` inside of a label renders properly


aagonzales:
Should there be a preview able for this one? @tw15egan

tay1orjones:
@aagonzales @tw15egan I'm not sure why the previews didn't work, but I just updated with latest from `main` and it got them to pull through.

tay1orjones (commented):
When I put super long text into the label via devtools this is what I see

image

Is it supposed to wrap like this when the tooltip is present?

tw15egan:
@tay1orjones If I applied the styles to a label with a tooltip, the tooltip would not show up. This prevents wrapping for `label` when `Tooltip` is not present. If someone uses a tooltip, the additional information would be in the tooltip rather than going to two lines. 

Do you have any workarounds to get it to work with the `Tooltip`? The `overflow-x` seems to be the issue

tay1orjones (commented):
Recap of our convo and what we found might work:

- `width: calc(100% - 2rem);` on the `label`
- `white-space: nowrap;`, `overflow-x: scroll;`, plus scrollbar hiding styles on the `span.cds--toggletip-label` inside the `label`

tw15egan:
@tay1orjones thanks for the chat, all styles should be working properly now 🎉

aagonzales (approved):
Nice! It looks good to me.

tay1orjones (approved):

PR #12873, opened 2022-12-16
Add typescript types to combo box

### LINKED ISSUE #12538: Add TypeScript types to ComboBox

sierrawetmore:
Closes #12538

This PR converts ComboBox.js to typescript.

#### Changelog

**Changed**

- ComboBox.js has been converted to TS

#### Testing / Reviewing

These changes can be tested and verified by running a fresh yarn, yarn build and yarn storybook. Verify that the behavior in the ComboBox story remains unchanged


sierrawetmore:

sierrawetmore:
recheck

tay1orjones (changes requested):
Hey, thanks for contributing this! One minor ask - instead of renaming the `.js` file to `.tsx`, could you actually rename it using git?

`git mv packages/react/src/components/ComboBox/ComboBox.js packages/react/src/components/ComboBox/ComboBox.tsx`

This way the diff will clearly show the file extension change but also the proper unified diff showing your changes within file.

Right now it's hard to tell what you've added to the file vs was was in it to begin with.

Also we should wait to merge this until https://github.com/carbon-design-system/carbon/pull/12787 is in

tay1orjones:
Here’s an example of how the diff should look after renaming via `git mv` alongside changes within the file

image

sierrawetmore:
Thank, didn't realize that was the better way of renaming it. It should look better now!

tay1orjones (commented):
@sierrawetmore Awesome yeah that looks great! https://github.com/carbon-design-system/carbon/pull/12787 is in - I think it caused some merge conflicts here

sierrawetmore:
This is on my to-do list in the next couple days. Also waiting to see where we land on the generic types conversation.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -89,23 +266,21 @@ const ComboBox = React.forwardRef((props, ref) => {
...
     readOnly,
     selectedItem,
     shouldFilterItem,
     size,
     titleText,
     translateWithId,
-    type, // eslint-disable-line no-unused-vars
     warn,
     warnText,
-    onStateChange, // eslint-disable-line no-unused-vars
```
tay1orjones: Typically these are destructured out of `props` so that when `...rest` is spread onto a child element it won't cause errors trying to apply attributes to an element that doesn't support them. 

Can you confirm adding `type` and/or `onStateChange` onto the component does not produce errors in the browser console?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -134,17 +309,17 @@ const ComboBox = React.forwardRef((props, ref) => {
...
 
   const filterItems = (items, itemToString, inputValue) =>
-    items.filter((item) =>
+    items.filter((item) => shouldFilterItem ?
       shouldFilterItem({
         item,
         itemToString,
         inputValue,
-      })
+      }) : defaultShouldFilterItem()
```
tay1orjones: Where is `defaultShouldFilterItem` defined?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -134,17 +309,17 @@ const ComboBox = React.forwardRef((props, ref) => {
   }
 
   const filterItems = (items, itemToString, inputValue) =>
-    items.filter((item) =>
+    items.filter((item) => shouldFilterItem ?
```
tay1orjones: Why is this check necessary?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -374,6 +550,7 @@ const ComboBox = React.forwardRef((props, ref) => {
                     translateWithId={translateWithId}
                     disabled={disabled || readOnly}
                     onClearSelection={handleSelectionClear}
+                    selectionCount={0}
```
tay1orjones, line 554: Could you explain this addition? Not sure why the selectionCount would need to be defined at 0 for this case.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -16,6 +16,8 @@ import {
   WarningFilled,
 } from '@carbon/icons-react';
 import ListBox, { PropTypes as ListBoxPropTypes } from '../ListBox';
+import ListBoxMenuItem from '../ListBox/ListBoxMenuItem';
+import ListBoxMenu from '../ListBox/ListBoxMenu';
```
tay1orjones: I think it's possible that this could break tests in downstream consumer's projects that rely on the component name to select nodes/components in test assertions. Is there a way to retain the usage of the namespaced exports (`ListBox.Menu`, etc.) instead of these direct ones?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -382,48 +559,49 @@ const ComboBox = React.forwardRef((props, ref) => {
...
                               highlightedIndex === index ||
-                              (selectedItem?.id &&
-                                selectedItem?.id === item.id) ||
+                              ((selectedItem as any)?.id &&
+                              (selectedItem as any)?.id === item.id) ||
                               false
                             }
                             title={
-                              itemToElement ? item.text : itemToString(item)
+                              itemToElement ? item.text : itemToString ? itemToString(item) : defaultItemToString(item)
```
tay1orjones: Why is the addition of a new `default*()` function necessary here?

tay1orjones (changes requested):
The biggest blocker here I think is the props being destructured out of `...rest`. Also a few questions below

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -89,23 +266,21 @@ const ComboBox = React.forwardRef((props, ref) => {
     light,
     onChange,
     onInputChange,
-    onToggleClick, // eslint-disable-line no-unused-vars
+    onToggleClick,
```
tay1orjones: Where is this being used now that we can remove the eslint line disable?

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: This is defined here https://github.com/carbon-design-system/carbon/blob/e3823aa2e0b14022df8ba86b91edf61ad6d4948e/packages/react/src/components/ComboBox/ComboBox.tsx#L38

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: `shouldFilterItem` is possibly undefined and results in a ts error if we do not check it.

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: `itemToString` is possibly undefined, so we needed to check; but looking at this again, I'm not sure the default is really what we want here, since we don't know what item is or what properties it has.

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: It is being used here https://github.com/carbon-design-system/carbon/blob/e3823aa2e0b14022df8ba86b91edf61ad6d4948e/packages/react/src/components/ComboBox/ComboBox.tsx#L363

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: So `selectionCount` is a required property and the component now throws a ts error without it. The reason for it being 0, is that the selection count is used to show a badge next to the component when multiple items are selected. In this case, the combobox only allows a single selection, so that badge shouldn't be present and setting it to 0 (rather than 1) accomplishes that.

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: Yeah, I don't recall the reason for changing this.

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: I didn't realize that could be an issue and checking that further shows that adding `onStateChange` would cause an error in that case. I've added these props back in to avoid this.

packages/react/src/components/ComboBox/ComboBox.tsx
sierrawetmore: Updated to be 
```
                         title={
                              itemToElement
                                ? item.text
                                : itemToString
                                ? itemToString(item)
                                : undefined
                            }
``` 
which will keep the original behavior

tw15egan (approved):
Functionality hasn't changed, no failing tests / a11y violations. LGTM 👍🏻 ✅

tay1orjones (approved):
Thank you! 🏆

PR #12881, opened 2022-12-19
fix(numberinput): remove v11 feature flag, typecast custom classname

### LINKED ISSUE #12878: [Bug]: 'undefined' className added to NumberInput
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v11.19.0

### React version

v17.0.2

### Description

In many places throughout the code base, we apply custom classNames to child components using a pattern like:

`[className]: x`

where x is typically className or !!className.
Screenshot 2022-12-18 at 2 07 48 PM
Screenshot 2022-12-18 at 2 07 57 PM

In all (almost all?) these cases, className is an optional property, meaning that it will some times be undefined. This means that we will pass `[undefined]: x` to our classNames() function when generating the list of class names to apply to an element. In most cases, we use `[className]: className`, which results in passing `[undefined]: undefined`, which evaluates to ['undefined']: false, so the "class name" "undefined" is not added (because it is mapped to false). However, this seems unexpected and likely to cause bugs in the future.

In addition, there is one case where className can be mapped to true, in the case of NumberInput. I confirmed that in the currently released version of carbon, this means that the element actually has the string "undefined" added as a class name.
Screenshot 2022-12-18 at 2 01 52 PM

Discovered this issue while adding type annotations for #12548, but guessing this will come up in many components when we try to add type annotations (easy workaround is to cast away the `| undefined` using `[className!]`, but if there's a "cleaner" solution possible that would be ideal).

### Suggested Severity

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-numberinput--default

### Steps to reproduce

Open [NumberInput storybook](https://react.carbondesignsystem.com/?path=/story/components-numberinput--default). Inspect the page and find the container div with class cds--form-item. Note that it also has a class "undefined".

### Code of Conduct

tay1orjones:
Closes #12878

#### Changelog

**Changed**

- remove the v11 feature flag from numberinput
- typescast custom class name to boolean so it does not provide a classname of `undefined`

#### Testing / Reviewing

- Go to the default numberinput story
- The `${prefix}--form-item` element should no longer have a class of `undefined`


alisonjoseph (approved):

aledavila (approved):
Lgtm

PR #12885, opened 2022-12-19
fix(code-snippet): removed polyfilled

### LINKED ISSUE #8271: [CodeSnippet] remove resize observer polyfill
### Summary

As IE11 will no longer be supported in v11 we will be able to remove the resize observer polyfill we use through `use-resize-observer`. Right now this is only used in CodeSnippet.

https://github.com/carbon-design-system/carbon/blob/ce5a810bb3aa5e802cdf00a5a9e1e83e3254b1f2/packages/react/src/components/CodeSnippet/CodeSnippet.js#L11

### Justification

The polyfill is only an additional [2.44kb in size](https://github.com/que-etc/resize-observer-polyfill) to the bundle but we won't need it anymore. It's a super small change:

```diff
- import useResizeObserver from 'use-resize-observer/polyfilled';
+ import useResizeObserver from 'use-resize-observer';
```

sstrubberg:
Closes #8271

### DIFF
--- a/packages/react/src/components/CodeSnippet/CodeSnippet.js
+++ b/packages/react/src/components/CodeSnippet/CodeSnippet.js
@@ -8,7 +8,7 @@
 import PropTypes from 'prop-types';
 import React, { useState, useRef, useEffect, useCallback } from 'react';
 import classNames from 'classnames';
-import useResizeObserver from 'use-resize-observer/polyfilled';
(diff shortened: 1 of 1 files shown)


tay1orjones:
@sstrubberg What's the best way to validate this?

sstrubberg:
lol, now i understand why you didn't just toss this in. Closing!

PR #12900, opened 2022-12-23
feat(Button): add as prop to IconButton inside of the Button component

matkrzy:
Button with flag `hasIconOnly` renders Button component without `as` prop so we are not able to pass custom render component

#### Changelog

**Changed**

- IconButton inside of the Button component has new prop `as`

#### Testing / Reviewing

1. set flag `hasIconOnly`
2. pass custom component via `as` property
3. custom component should render instead default button

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -977,6 +977,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "matkrzy",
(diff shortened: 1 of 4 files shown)


matkrzy:

tay1orjones (approved):
Great catch, thanks!

tw15egan (approved):

PR #12906, opened 2022-12-31
feat(react): enable textarea to limit text input to set amount of words

### LINKED ISSUE #12905: [Feature Request]: word limit for textArea component
### The problem

I would like for the textArea component in the react package to be able to limit the amount of words that user can enter. At the moment it can only limit the amount of characters.

### The solution

Allow textArea component in the react package to count words entered by the user and prevent user from typing more words once the word limit is reached.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/13788
- [ ] https://github.com/carbon-design-system/carbon/issues/13789
```

### LINKED ISSUE #13788: TextArea word limit: implementation

matejoslav:
Co-authored-by: SamChinellato <49278203+SamChinellato@users.noreply.github.com>
Co-authored-by: stevenpatrick009 <106097350+stevenpatrick009@users.noreply.github.com>

Closes #13788 

Added functionality to textArea component to limit the number of words the user can type. Also updated the useAnouncer to handle new wording by allowing it to either display "x characters left" or "x words left" based on args passed in.

#### Changelog

**New**

- textArea component has now word counter mode and can now limit the amount of words user types
- added more tests for the textArea component

**Changed**

- useAnouncer can now take the entityName and use that to display the "x <entity name> left"
- added tests for useAnnouncer

#### Testing / Reviewing

- run the storybook for react components
- navigate to the textArea component
- select the "WithWordCounter" textArea type
- type in few words and observe the counter increase
- try to type in more words than the limit and observe that user can not add more words
- try to paste in text with more words than the limit and observe that the text has been cut to match the word limit


matejoslav:

matejoslav:
recheck

sstrubberg:
@matejoslav thank you so much for the PR. Two things as you're building this out. 

1. Have you considered any error message handling for when a user reaches their limit?
2. Updates to the documentation on the website reflecting this enhancement.

matejoslav:
@sstrubberg Good points.

For point number 2, will make relevant doc updates and push them.

For point number 1, wanted to make the component behave in similar way to having a maximum character limit, at the moment the user cannot exceed the amount of words if they have reached the limit, the announcer message is also displayed showing 0 words left. Not sure if a more visual error handling is needed but happy to add one if you think it would benefit the user experience.

matejoslav:
Just checked the website repo, is it okay to assume that this change would have to be merged first and then we would add the feature description there or would you want them to be done at the same time?

sstrubberg:
> Just checked the website repo, is it okay to assume that this change would have to be merged first and then we would add the feature description there or would you want them to be done at the same time?

correct, merge first, then document. Just didn't want that part to get lost :)

tay1orjones:
@matejoslav Looks like all-contributors needs to be ran again to resolve the conflict with `main`

aledavila:
Bump @matejoslav

matejoslav:
Hello, I am looking into it now 👍

matejoslav:
For some reason I am getting test failure for "Simple date picker › should initialize a calendar when using react.lazy"
I did a clean clone of the carbon repository and getting the same issue.

matejoslav:
Screenshot 2023-01-29 at 17 19 09

matejoslav:
Screenshot 2023-01-29 at 17 30 01

tay1orjones:
@matejoslav It's probably something with your local environment. Try pulling latest from upstream `main` and running `yarn install` and `yarn build`. `yarn test` should pass at that point.

matejoslav:
@tay1orjones the issue is now resolved 👍 was just something strange with my local clone

tay1orjones:
@alisonjoseph @tw15egan I think this is ready for review 👍

package.json
```diff
@@ -90,7 +90,7 @@
       "body-max-line-length": [
         2,
         "always",
-        80
+        90
```
alisonjoseph: What was the reason for this change?

packages/react/src/components/TextArea/TextArea.js
```diff
@@ -183,6 +229,11 @@ TextArea.propTypes = {
    */
   cols: PropTypes.number,
 
+  /**
+   * Specify the method used for calculating the counter number
+   */
+  counterMode: PropTypes.oneOf(['default', 'words']),
```
alisonjoseph: Instead of default and words, would it make more sense to be a little more specific and use `character` and `word`

alisonjoseph (changes requested):
This is looking and working great! Couple small suggestions, I'm not sure we need to keep the story for this, leaving it in the playground should be sufficient. I would also update the comment / description for the `enableCounter` prop to just say `Specify whether to display the counter` since it is not longer only used for character count.

tw15egan (approved):
LGTM once the comments from Alison have been addressed 👍🏻 ✅

tw15egan (changes requested):
It looks like this file was recently converted to Typescript by the Typescript working group helping improve our library. The merge may be a bit messy, it may be best to close this and resubmit starting from the `TextArea.tsx` file

matejoslav:
@alisonjoseph thank you will action your comments

@tw15egan yeah might resubmit, will attempt the merge resolution locally for fun

sstrubberg:
@matejoslav Do you need any help getting this conflicts resolved? Anything you need from us?

matejoslav:
@sstrubberg Hello, just had lack of time but working on it now 👍

sstrubberg:
Hey @matejoslav closing this PR for now. Feel free to re-open once you have your conflicts resolved.

package.json
matejoslav: This was because one of the collaborators name was longer than 80 characters and could not be properly included in the commit without increasing the limit for each line. It was the first commit in the branch from December last year

matejoslav:
Just spotted https://github.com/carbon-design-system/carbon/pull/13449

Depending on which PR is merged first, there will be a requirement to update the useEffect to include the behaviour for different counterMode

commenting just in case

package.json
tw15egan: It doesn't have to do with the user's name, it's only validating the character count of the commit message. It just needed a shorter commit message.

tasks/sync.js
```diff
@@ -9,14 +9,6 @@
...
 const fs = require('fs-extra');
 const path = require('path');
-const prettier = require('prettier');
-const lerna = require('../lerna.json');
-const packageJson = require('../package.json');
-
-const prettierOptions = {
-  ...packageJson.prettier,
-  parser: 'markdown',
-};
```
tw15egan: Are these changes intentional?

package.json
matejoslav: Reverted this change

tasks/sync.js
matejoslav: ESLint complained locally that these were unused, will check again

matejoslav: Screenshot 2023-05-09 at 19 00 19

tw15egan: Okay, I was just unsure why this was being pulled in to this PR. Thanks for clarifying.

alisonjoseph: Agree, this seems unrelated to the issue/PR

matejoslav:
@tw15egan is the merge-conflict label still required?

tw15egan (dismissed):
This looks great, thanks for sticking with the PR and adding in this functionality! LGTM 👍 ✅

tw15egan:
@matejoslav apologies, but #13449 was merged in before this one and is currently causing a conflict as noted [here](https://github.com/carbon-design-system/carbon/pull/12906#issuecomment-1535459610)

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -131,6 +131,11 @@ export interface TextAreaProps
    * Provide the text that is displayed when the control is in warning state
    */
   warnText?: ReactNodeLike;
+
+  /**
+   * Specify the method used for calculating the counter number
+   */
+  counterMode?: string;
```
francinelucca:
```suggestion
  counterMode?: 'character' | 'word';
```

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -155,22 +161,67 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
-    defaultValue?.toString().length || value?.toString().length || 0
+    defaultValue
+      ? counterMode === 'character'
+        ? defaultValue.toString().length
+        : defaultValue.toString().match(/\w+/g)?.length || 0
+      : value
+      ? counterMode === 'character'
+        ? value.toString().length
+        : value.toString().match(/\w+/g)?.length || 0
+      : 0
```
francinelucca: this is really hard to read, could we offload this logic onto a `getInitialTextCount()` function and use if statements instead?

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -155,22 +161,67 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
     onChange: (evt) => {
-      if (!other.disabled && onChange) {
-        setTextCount(evt.target.value?.length);
-        onChange(evt);
+      if (!other.disabled) {
+        if (counterMode == 'character') {
+          setTextCount(evt.target.value.length);
+        } else if (counterMode == 'word') {
+          if (!evt.target.value) {
+            console.log('reset');
```
francinelucca:
```suggestion
```

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -155,22 +161,67 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
+            } else if (matchedWords && matchedWords.length > maxCount) {
+              const first_max = evt.target.value
+                .split(/\s+/)
+                .slice(0, maxCount)
+                .join(' ');
+
+              setTextCount(maxCount);
+              textareaRef.current.value = first_max;
+            }
+            // todo: add handling if no matched words were found
```
francinelucca: is a handler still necessary or can we remove this comment?

packages/react/src/internal/useAnnouncer.js
```diff
@@ -5,9 +5,12 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-export function useAnnouncer(textCount, maxCount) {
+export function useAnnouncer(textCount, maxCount, entityName) {
```
francinelucca: if we set the default of `entityName `to 'characters' here then the if statement below is not necessary and we can always construct the sentence using `${entityName}`
```suggestion
export function useAnnouncer(textCount, maxCount, entityName = 'characters') {
```

francinelucca (changes requested):
This is looking great! just some questions and refactoring suggestions

matejoslav:
@tw15egan yep, resolving the conflict is almost done just got stuck on the tests as the counter changes are using setTimeout which needs accounting for in the tests

packages/react/src/components/TextArea/TextArea.tsx
matejoslav: removed

packages/react/src/components/TextArea/TextArea.tsx
matejoslav: done

packages/react/src/components/TextArea/TextArea.tsx
matejoslav: done

packages/react/src/components/TextArea/TextArea.tsx
matejoslav: done

packages/react/src/internal/useAnnouncer.js
matejoslav: done

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -155,31 +161,103 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
-      defaultValue?.toString()?.length || value?.toString()?.length || 0
+      defaultValue
+        ? counterMode === 'character'
+          ? defaultValue.toString().length
+          : defaultValue.toString().match(/\w+/g)?.length || 0
+        : value
+        ? counterMode === 'character'
+          ? value.toString().length
+          : value.toString().match(/\w+/g)?.length || 0
+        : 0
```
francinelucca: can we call the getinitialtextcount function here as well?

francinelucca (changes requested):

packages/react/src/components/TextArea/TextArea.tsx
matejoslav: My bad must have missed it

matejoslav:
Could you also please double check there was no regression for https://github.com/carbon-design-system/carbon/pull/13449 I did check but you never know 😃

francinelucca (changes requested):
Confirmed no regression on https://github.com/carbon-design-system/carbon/pull/13449 as a result of these changes. 
However, I do see a weird behavior in that in the current state of the TextArea, setting `enableTextCounter` to false makes the component ignore `maxCount` and doesn't look like that is the case for when counter = 'word' with these new changes, see video below: 

https://github.com/carbon-design-system/carbon/assets/40550942/90f33e28-da71-4d85-9f8d-7c49371eabab

I'd expect counter=word and counter=character to behave the same when it comes to `enableCounter` being on/off and the maxCount being reached, which in production is: the `maxCount` is ignored if `enableCounter` is turned off

matejoslav:
@francinelucca Good spot, I think there might be tests missing for this case so will add them, and this is I think due to the onChange and onKeyDown implementation that does not check if counter is enabled so that should fix it

sstrubberg:
@matejoslav Just wanted to acknowledge you for sticking with this PR and making this contribution with all the back and forth. You rock!

matejoslav:
Got an update, onKeyDown fix is simple as I can basically just add a check for "enableCounter" to be true.

for the onChange function it is a bit tricky. I will post explanation once I test some things out.

francinelucca:
doesn't look like this is ready for re-review @sstrubberg

francinelucca:
Hi @matejoslav 👋🏼, looks like we still have `enableCounter` issue referenced here https://github.com/carbon-design-system/carbon/pull/12906#pullrequestreview-1453799054, will you be able to work on this? We'd love to get this merged in.
TIA

matejoslav:
@francinelucca yes, I have been away for few days but now I am back so will make a change and also post questions about some specific behaviour that I observed

matejoslav:
https://github.com/carbon-design-system/carbon/assets/8973672/7f7cbb31-36d6-407f-ada4-6102b5209179

matejoslav:
@francinelucca please let me know if behaviour in the video is what you would expect, for words when limit is reached and counter is disabled user can continue now typing more words. When counter is enabled again and limit was 5 but user typed in 10 words the next onChange trigger causes the text area's text to be trimmed to the max word count. This is different from the behaviour that was there for characters as the characters that go over the limit stay in the content.

matejoslav:
Once confirmed I will add tests and finalise changes.

laurenmrice (approved):

tay1orjones (dismissed):
I think with that this should be good to go. @francinelucca can you confirm?

tay1orjones:
@matejoslav looks like there are some conflicts here to address

matejoslav:
@francinelucca @tay1orjones correct, need to add tests, also there appears to be conflict. As far as the behaviour update goes I need to figure out how to handle it, basically the trim was there in case user would paste in text to prevent breaking the limit by pasting in content longer than limit

matejoslav:
Fyi: made some progress with the logic, now it behaves similarly to character, only bug is that the counter does not update immediately when enable counter goes from false to true and there are more words than the limit, it needs an onchange event to register new count so need to check. Also added logic to prevent pasting more words than the limit when enable counter is true

matejoslav:
And it still needs tests :)

matejoslav:
@tay1orjones @francinelucca modified functionality is now pushed and I added missing test, can you please test and let me know. I think the code needs a bit of a refactor so will see what I can do :)

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -155,32 +161,122 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
     id,
+    onKeyDown: (evt) => {
+      if (!other.disabled && enableCounter && counterMode === 'word') {
+        const key = evt.which;
+        console.log(key);
+        if (maxCount && textCount >= maxCount && key === 32) {
+          evt.preventDefault();
+        }
+      }
+    },
```
francinelucca:
```suggestion
    onKeyDown: (evt) => {
      if (!other.disabled && enableCounter && counterMode === 'word') {
        const key = evt.which;
        if (maxCount && textCount >= maxCount && key === 32) {
          evt.preventDefault();
        }
      }
    },
```

packages/react/src/components/TextArea/__tests__/TextArea-test.js
```diff
@@ -309,4 +310,146 @@ describe('TextArea', () => {
...
+
+      rerender(
+        <TextArea
+          id="input-1"
+          labelText="TextArea label"
+          enableCounter={true}
+          maxCount={5}
+          counterMode="word"
+          // defaultValue="one two three four five"
+        />
```
francinelucca:
```suggestion
      rerender(
        <TextArea
          id="input-1"
          labelText="TextArea label"
          enableCounter={true}
          maxCount={5}
          counterMode="word"
        />
```

packages/react/src/components/TextArea/__tests__/TextArea-test.js
```diff
@@ -309,4 +310,146 @@ describe('TextArea', () => {
...
+
+      const clipboardEvent = new Event('paste', {
+        bubbles: true,
+        cancelable: true,
+        composed: true,
+      });
+
+      clipboardEvent.clipboardData = {
+        getData: () => 'one two three four five six seven eight nine ten',
+      };
```
francinelucca: is this necessary? doesn't seem like it's being used

francinelucca (changes requested):
the enableCounter bug seems to be fixed 🎉. This is almost there, just noticed one detail:
In the video below, I'd expect the counter to correctly show up the number of words/characters when switching between modes


https://github.com/carbon-design-system/carbon/assets/40550942/07b0cabd-e151-494e-97af-bc8a3156f7d0


Also added some cleanup suggestions :)

matejoslav:
Ok let me take a look at the requested changes asap

packages/react/src/components/TextArea/__tests__/TextArea-test.js
matejoslav: let me check this, I know I had one working event and one random one

matejoslav:
Interestingly I didnt have conflicts in Readme and contributors file, @francinelucca would you want me to put the contributors at the end of the updated list?

packages/react/src/components/TextArea/__tests__/TextArea-test.js
matejoslav: yep you are right its not used

matejoslav:
@francinelucca on the switching of modes and counter update, I think its because it needs onChange to execute at least once to redo the counting, I can try something, potentially manually triggering onChange with the textarea ref to get the counter update or alternatively a hook I guess

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -159,32 +165,122 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
   );
   const { current: textAreaInstanceId } = useRef(getInstanceId());
 
   useEffect(() => {
-    setTextCount(
-      defaultValue?.toString()?.length || value?.toString()?.length || 0
-    );
-  }, [value, defaultValue]);
+    setTextCount(getInitialTextCount(defaultValue, value, counterMode));
+  }, [value, defaultValue, counterMode]);
```
matejoslav, line 199: @francinelucca hmm I would expect this to handle the recounting of characters

matejoslav: I think we uncovered a big bug with text area itself, value/defaultValue prop is undefined so when we type the count changes but as soon as you change the mode it tries to calculate initial again and returns 0, that useEffect runs also everytime you type so it gets set to 0 and onChange changes it back to correct value, what was the idea of the useEffect? It was implemented before and I just modified it slightly

matejoslav: Ok the issue is that if value or default value is undefined the change of mode reruns the getInitial count and returns 0, I also discovered that everytime we type the initial count function is reexecuted, that is because there is a re-render of the component (for some reason) and useState reruns the initial value, got a fix I think.

francinelucca:
> Interestingly I didnt have conflicts in Readme and contributors file, @francinelucca would you want me to put the contributors at the end of the updated list?

I don't think order matters that much as long as the file structure is correct, your call!

matejoslav:
Will push the fix tomorrow, finalised it today and also fixed bug in onPaste which allowed people to paste in any content once they had 1 word in text area beforehand

matejoslav:
@francinelucca check please if it now works as expected

matejoslav:
The contributorsrc and Readme seem to have been altered and the order of the people was changed, could you please check and advise what to do?

francinelucca (approved):
Storybook issue seems to be fixed now, LGTM!

francinelucca:
> The contributorsrc and Readme seem to have been altered and the order of the people was changed, could you please check and advise what to do?

Might be worth to copy/paste the file from main, add your changes and commit to solve conflicts, your call though!

alisonjoseph (approved):
LGTM!

PR #12969, opened 2023-01-11
fix(datepicker): ensure flatpickr instance is crated after lazy input resolves

### LINKED ISSUE #10692: [Bug]: DatePicker and DatePicker input are not clickable when loaded with React.lazy
### Package

carbon-components-react

### Browser

Chrome

### Package version

v10.52.0

### React version

v.17.0.2

### Description

When I import the DatePicker and DatePickerInput components with React.lazy, the date picker does not set up well.
Calendar does not appear when I click on the date input. 
In addition when I tested I found that in some cases the initial date value that I add to the DatePickerInput component's value property does not appear in the input.

It seems that other carbon react form components work well with React lazy.

Additional info:
The example code from the codebox works well without React lazy:
https://codesandbox.io/s/carbon-react-datepicker-non-lazy-example-ldj4x

react-flatpickr works well with React lazy:
https://codesandbox.io/s/react-flatpickr-lazy-example-gvwmy 

### CodeSandbox example

https://codesandbox.io/s/carbon-react-datepicker-lazy-example-xr50d

### Steps to reproduce

- import DatePicker and DatePickerInput components with React lazy. You can use react-lazily for lazy import. It can handle destructuring during import.
- Wrap the components in React.Suspense with any fallback property.

### Code of Conduct

tay1orjones:
Closes #10692 

If the DatePicker was brought in via React.lazy, the `ref` assigned to the start input was not being updated after the component resolved. This caused the flatpickr instance to never be assigned to the input, so the calendar would never show.

#### Changelog

**Changed**

- update datepicker to use `useCallback` instead of `useRef`, store the status of the ref in state, and use it to trigger the existing `useEffect` after react.lazy resolves

#### Testing / Reviewing

- Check out the `simple` datepicker story, it should function normally. I've modified it to use react.lazy
- ~Don't merge this PR until we've reverted the story to it's original state.~ Done


tw15egan (approved):

francinelucca (approved):
😍 LGTM!

PR #12977, opened 2023-01-13
fix(overflow-menu): accessibility navigation issues

### LINKED ISSUE #9471: Tab navigation on OverflowMenu skips the tabbing of upcoming focusable elements
## Brief description
Tab navigation on OverflowMenu causes it to skip the tabbing of upcoming focusable elements and continues tab navigation from top of the DOM.

## Environment

> Browser
Chrome

## Detailed description

Latest

After entering (opening) the overflow menu and then tabbing, it should tab to the next focusable element.

> What happened instead?
After entering (opening) the overflow menu and then tabbing, it skipped upcoming focusable elements and restarted tab navigation from top of the webpage DOM.

1. Step one: Tab to an overflow menu
2. Step two: Open it, then click tab again
3. Step three: Tab once more and it skips next (upcoming) tab stops

> Can be reproduced via the Live demo here: https://www.carbondesignsystem.com/components/overflow-menu/code/

francinelucca:
Closes #9471

- Fixes focus reset on tabbing out of open overflow menu and allows navigation of context menu items with up and down arrow keys. 
- Rewrite aria labels for overflowMenu and button
- uses `IconButton` instead of `button` to automatically display tooltip 

#### Changelog

**Changed**

- Cleanup arrow navigation logic (was not working): prevent undefined overflowItem by not destructuring next focusable element 
- Focus menu element on tab out to preserve tab order 
-  Small `closeMenuAndFocus` refactor
- Rewrite `ariaLabel` and `iconDescription` props to "Options"
- Replace overflowMenu's `button` element with an `IconButton` to display tooltip
- Adds `closeOnActivation` optional prop to tooltip
- Updates `IconButton` to close tooltip on button activation
- Adds scss styles to maintain `Breadcrumb`'s styling (accounting for new tooltip) when using overflowmenu 
- Adds tests to cover new cases for tooltip
- Updates snapshots

**Removed**

- Unused `handleKeyDown` function

#### Testing / Reviewing

On local OverflowMenu SB:
- Test keyboard navigation: ESC, Tab and Up & Down arrows
- Overflow menu button should now have a tooltip "Options"
- Expect pressing "tab" on open overflow menu to close the menu and preserve current tab position, compare to behavior in https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default
- tooltip should close on activation (click/enter/space) of the button

Review/Take into consideration:
- I had to update the public API snapshot to include the new `closeOnActivation` prop for tooltip and got a warning about semver, is there something else I need to do or is this ok?
- I had to take the overflowmenu body [outside of the button](https://github.com/carbon-design-system/carbon/pull/12977/files#diff-1b11ccdde5fc59739a0fdd9e17c5a58b9df3bed62fda2e8fee592a3438c8853cR569-R571) (was previously inside) and wrap the entire thing in a `span` to cover for the new tooltip behavior

---

TODO: 

- [x] Remove Test Data before merging


packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -233,10 +234,8 @@ class OverflowMenu extends Component {
   static contextType = PrefixContext;
 
   static defaultProps = {
-    ariaLabel: FeatureFlags.enabled('enable-v11-release')
-      ? null
-      : 'Open and close list of options',
-    iconDescription: 'Open and close list of options',
+    ariaLabel: FeatureFlags.enabled('enable-v11-release') ? null : 'Options',
```
tay1orjones: This feature flag is now permanently enabled, the toggle can be removed.

packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -400,9 +396,9 @@ class OverflowMenu extends Component {
           return nextIndex;
       }
     })();
-    const { overflowMenuItem } =
+    const overflowMenuItem =
       this[`overflowMenuItem${enabledIndices[nextValidIndex]}`];
-    overflowMenuItem?.current?.focus();
+    overflowMenuItem?.focus();
```
tay1orjones, line 394: Is this the root of the bug? There was already code attempting to focus the item but it just wasn't calling `focus()` on the right thing?

packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -564,20 +560,20 @@ class OverflowMenu extends Component {
...
-        <button
+        <IconButton
           {...other}
           type="button"
           aria-haspopup
           aria-expanded={this.state.open}
           className={overflowMenuClasses}
           onKeyDown={this.handleKeyPress}
           onClick={this.handleClick}
-          aria-label={ariaLabel}
```
tay1orjones, line 575: Does this mean we can also remove this from the props destructure at the top of `render()`?

packages/react/src/components/OverflowMenu/OverflowMenu.js
```diff
@@ -564,20 +560,20 @@ class OverflowMenu extends Component {
 
     return (
       <ClickListener onClickOutside={this.handleClickOutside}>
-        <button
+        <IconButton
           {...other}
```
tay1orjones: One piece here I'm curious about is if folks are going to want to control the positioning of the tooltip. I can imagine someone opening an issue showing the menu conflicting with the tooltip placement. 

Can they do that here via `...other`, or would a more explicit prop type be preferred? At the very least maybe an update to the `.mdx` docs stating how the tooltip position can be configured?

tay1orjones: @laurenmrice do we need to update the website or design specs to show that OverflowMenu now includes a tooltip via IconButton? For context, this was the recommendation from @mbgower in the root issue.

tay1orjones (changes requested):
A few comments below. 

I'm having a hard time figuring out how to validate this in storybook. Besides the tooltip (looks great btw!), the focus behavior seems to be the same to be between the deploy preview and prod.

What's the best way to test that the focus behavior is working correctly? Should we add another interactive element to the story to validate it moves correctly? Or is it enough to show that it moves correctly to the next focusable item outside of the storybook pane?

packages/react/src/components/OverflowMenu/OverflowMenu.js
francinelucca: yeah, there's no "overflowItem" property to destructure from the result of 400 and even then the ".current" was always coming up undefined, I tried just focusing the element directly and that worked. The value is being set here for context:
image

my theory is maybe this code used to work way back when but then as the component kept getting changed/refactored this broke ?

packages/react/src/components/OverflowMenu/OverflowMenu.js
francinelucca: yes in theory just passing in the `align` prop to the overflowMenu should forward that to the IconButton , i'm happy to add a blurb in the docs about how the tooltip alignment and any other IconButton props can be passed onto it or we could make an explicit prop, whichever seems more correct?

packages/react/src/components/OverflowMenu/OverflowMenu.js
francinelucca: so we're still using this prop to pass onto the `UL` 
image

So I'd say we still need to destructure it?

mbgower:
If my suggestion in leading to challenges, happy to discuss them.
On the question of positioning, one thing I was wondering...
Do we want the tooltip to disappear on `Enter` and `space` (i.e., activation)?
At the moment, one can discuss a tooltip by pressing Esc. That makes a lot of sense. When I'm on a button, to me it seems logical that if I activate it, the tooltip should disappear since I've made a decision to act on it (I obviously don't need to know what it was for).
If that were enabled, you'd never have a conflict between the tip and the menu shown from a menu-button (i.e., overflow menu). The tip would disappear even as the menu showed up.

Does that give us anything?

packages/react/src/components/OverflowMenu/OverflowMenu.js
francinelucca: Done!

francinelucca:
> If my suggestion is leading to challenges, happy to discuss them. On the question of positioning, one thing I was wondering... Do we want the tooltip to disappear on `Enter` and `space` (i.e., activation)? At the moment, one can dismiss a tooltip by pressing Esc. That makes a lot of sense. When I'm on a button, to me it seems logical that if I activate it, the tooltip should also disappear since I've made a decision to act on the button (I obviously don't need to know what it was for anymore). If that were put in place, you'd never have a conflict between the tip and the menu shown from a menu-button (i.e., overflow menu). The tip would disappear even as the menu showed up.
> 
> Does that give us anything?

@mbgower would that behavior be overflowMenu only or does it also apply to iconButton?

packages/react/src/components/OverflowMenu/OverflowMenu.js
francinelucca: Added doc section on how to customize tooltip alignment: https://deploy-preview-12977--carbon-components-react.netlify.app/?path=/docs/components-overflowmenu--default#customizing-the-tooltip

francinelucca:
> I'm having a hard time figuring out how to validate this in storybook. Besides the tooltip (looks great btw!), the focus behavior seems to be the same to be between the deploy preview and prod.
> 
> What's the best way to test that the focus behavior is working correctly? Should we add another interactive element to the story to validate it moves correctly? Or is it enough to show that it moves correctly to the next focusable item outside of the storybook pane?

@tay1orjones added a button element after the overflow to make it easier to test, basically on prod tabbing out of the open menu will reset the tab order (you'll have to tab twice to get to the menu button again), in the preview you should stay at the current position

mbgower:
> would that behavior be overflowMenu only or does it also apply to iconButton?

I think it would apply to both. I can't see any reason why it wouldn't. Not only does me acting on a button with a tooltip suggest I no longer need the tooltip displayed, but what I am effectively doing by pressing that button is moving focus, right? And that _should_ make the tooltip disappear at any rate.
There are some odd interactions in Carbon at the moment where the focus remains on a button after being activated -- it's effectively a togglebutton. But even then, I think it makes sense for it to go away.

https://user-images.githubusercontent.com/14298245/214708851-560155f6-cc9b-46a6-a031-11bb43d761a0.mov

BTW, i thought I'd just QA this by looking at Google Sheets where I know they do a lot of these tooltips on buttons (and even on other controls, such as unlabelled inputs). The action seems to make sense across all the ones where it exists. The tooltip goes away on user interaction, where opening or just acting on it.


https://user-images.githubusercontent.com/14298245/214709094-331eeff1-e9b4-45d6-89b4-39584c160625.mov

francinelucca:
@mbgower does this look about right in terms of keyboard behavior and the tooltip? Should have the tooltip on IconButton as well 🙏🏻
https://deploy-preview-12977--carbon-components-react.netlify.app/?path=/story/components-overflowmenu--default

mbgower:
> does this look about right in terms of keyboard behavior and the tooltip?

Yes, the default overflow button's tooltip behaviour works well. It disappears on Esc, activation (Enter/Space) or Tab (loss of focus).

 Note that in the playground, there's a focus issue when the kebab is opened. Not relevant to the tooltip, but thought I'd point it out.

francinelucca:
> Note that in the playground, there's a focus issue when the kebab is opened (menu items not getting focus). Not relevant to the tooltip, but thought I'd point it out.

@mbgower my bad, should be working now!

tay1orjones (approved):
This looks solid to me, thanks for taking the time to iterate on this so much!

tw15egan:
Looks good! Arrow keys can navigate the menu, and focus is where I expect it to be. 

There is one issue that may be unrelated to this, but the `ul` still seems to get focus. Is there any we can avoid this? 

Screenshot 2023-02-02 at 1 16 04 PM

francinelucca:
> There is one issue that may be unrelated to this, but the `ul` still seems to get focus. Is there any we can avoid this?
> 
> Screenshot 2023-02-02 at 1 16 04 PM

@tw15egan so that happens when `focusTrap` is on (which is the default), if you try it in the playground (starts at false in playground) you'll see if won't happen; so I understand it's a focusTrap thing but I'm actually not sure what is the intent for that prop's behavior 😅

tw15egan:
@francinelucca hmm, that's weird; not sure why we have the `focusTrap` on there since it isn't navigated by tabs anyways 🤔 Thanks for filling me in.

@tay1orjones, any ideas? Anyways, it seems unrelated to this specific fix, so we can create a separate issue if we want to take action on it!

tw15egan (approved):

mbgower:
I'll mention that the focus around the whole menu is not necessarily a BAD thing; it really depends on how the focus handling happens on the menu items. 
FOr instance, I've seen some implementations where highlighting is used to show which menu item has focus and there is also a focus indicator around the whole list which helps define the border (sorta analogous to how there is a cursor flashing inside this text area while there is a focus indicator around the whole thing).
I'm certainly not saying that's necessary here! Just observing that it _could_ be part of a good, intentional design when tackled holistically.

francinelucca:
@tw15egan @mbgower  I'll take these notes and create an exploration/discussion issue to look into the `focusTrap` behavior on OverflowMenu 👍

PR #13001, opened 2023-01-22
build(typescript): fix rollup placing build in nested directory

jkap:
With `declarationDir` set, Rollup places most of its output, including type definitions, in a nested folder. (ex: `lib/lib/index.d.ts` instead of `lib/index.d.ts`). This prevents TypeScript from finding the typedefinition, as by default it's looking for a `.d.ts` sibling to the main import.

Impacts `@carbon/react` only. This issue may be present in other packages, but I have only observed it in `@carbon/react`.

Rollup behaves as expected when replacing `declarationDir` with `outDir`.

#### Changelog

**Changed**

- Rollup config for `@carbon/react` has been updated to output type declarations as siblings for the main imports, rather than in a nested directory.

#### Testing / Reviewing

Run `yarn install` and `yarn build` as usual, check the output of `packages/react/es` and `packages/react/lib`.


jkap:

tay1orjones (approved):
Yes, thank you for this! Looks great 

image

tw15egan (approved):
Output looking as expected; thanks for this! 👍 ✅

Issue #6177, opened 2020-06-01
Add listbox components to public API exports


joshblack:
We currently have several internal components that we use to build up our dropdowns, including:

- ListBox
- ListBoxField
- ListBoxMenu
- ListBoxMenuItem
- ListBoxMenuIcon
- ListBoxSelection

It would be helpful to update these, make sure they're good to go, and then add them as part of our public API. This would be helpful so teams could build out dropdown-variants using the look and feel of our components without necessarily having to use the behavior.

Looking forward, this may be the recommended strategy in conjunction with downshift so we don't leak library details through the public API of our components.

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/12863
- [ ] https://github.com/carbon-design-system/carbon/issues/13054
```

sonarestis:
Writing with a personal email, but my team in IBM from Ottawa, Canada would really appreciate if the components above are exposed. Saves us time from reimplementing the wheel!

PaulJThompson:
+1 it would be super-helpful if these were exposed.

dan-cooke:
This would be incredibly useful, as the import from `lib` directly seems to break tree shaking of icons react.

This is the only lib import in our entire, app we are most likely going to have to copy the source code for the ListBox component into our repo in the meantime.

mango906:
Hello! How is this issue going?

mthaak:
Hi! Has there been any progress on this? I'm having the exact same issue as @dan-cooke

jaydee-ibm:
Hi @sstrubberg + @tay1orjones just chiming in as a dev rep from another internal IBM team (w3 carbon + IBMer Carbon Components). We are also using Listbox and its related components for the new w3 and it would be great to have them exposed sooner rather than later (noticed they've been moved from Next to Later on the roadmap as of October). 

Xoxo Jd

abdonrd:
What's the status of this? Are there plans to export these components? Thanks!

sstrubberg:
@abdonrd It's currently iceboxed. I'm recalling a conversation that @tay1orjones and I had a while ago about this one. I'll sync up with him and see where we left off on this workstream's path forward. Are you open to making a PR?

abdonrd:
@sstrubberg how does this affect https://github.com/carbon-design-system/carbon/issues/12863?

PR #13022, opened 2023-01-24
refactor(typescript): convert Dropdown and dependencies to TypeScript

### LINKED ISSUE #12542: Add TypeScript types to Dropdown

### LINKED ISSUE #13048: Add TypeScript types to FluidForm

lewandom:
Closes #12542
Closes #13048 

Convert React Dropdown component and the dependencies (FluidForm and ListBox) to add TypeScript support.

#### Changelog

**Changed**

- Converted relevant component files from *.js to *.ts/*.tsx.
- Introduced appropriate TypeScript types for components (component types, props types) and variables (e.g. event handlers arguments).
- Addressed a small issue in Dropdown click handler related to use of focus() function.

#### Testing / Reviewing

The Dropdown component imports for the storybook should be valid and rendering. In a TypeScript environment, you should be able to import and reference the Dropdown<ItemType> component and its typings (DropdownProps<ItemType>, OnChangeData<ItemType>).


lewandom:

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -5,39 +5,193 @@
...
+
+  /**
+   * Disable the control
+   */
+  disabled?: boolean;
+
+  /**
+   * Additional props passed to Downshift
+   */
+  downshiftProps: any; // UseSelectProps<ItemType>;
```
francinelucca: is there a way to type this?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -221,12 +379,15 @@ const Dropdown = React.forwardRef(function Dropdown(
...
+              const disabled = isObject && 'disabled' in item && item.disabled === true;
               const itemProps = getItemProps({
                 item,
                 index,
-                disabled: item.disabled,
+                disabled,
               });
-              const title = itemToElement ? item.text : itemToString(item);
+              const text = isObject && 'text' in item ? item.text : itemToString(item);
+              const title = itemToElement ? text : itemToString(item);
```
francinelucca:
```suggestion
              const title = isObject && 'text' in item && itemToElement ? item.text : itemToString(item);
```

packages/react/src/components/ListBox/ListBoxSelection.tsx
```diff
@@ -74,7 +114,7 @@ function ListBoxSelection({
         className={`${prefix}--tag__close-icon`}
         onClick={handleOnClick}
         onKeyDown={handleOnKeyDown}
-        disabled={disabled}
+        // disabled={disabled} TODO determine if that's always been broken
```
francinelucca: can we get more info on what's not working here? if there's an existing bug we can open a separate issue for this

packages/react/src/components/ListBox/ListBoxSelection.tsx
```diff
@@ -40,14 +80,14 @@ function ListBoxSelection({
...
-  const handleOnKeyDown = (event) => {
+  const handleOnKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
     event.stopPropagation();
     if (disabled || readOnly) {
       return;
     }
 
     // When a user hits ENTER, we'll clear the selection
-    if (match(event, keys.Enter)) {
+    if (match(event.code, keys.Enter)) {
```
francinelucca, line 90: is this currently broken? how do we test this? @tay1orjones

francinelucca (changes requested):
Thanks for getting this PR up! This is looking good, just some change requests and a couple questions.

packages/react/src/components/Dropdown/Dropdown.tsx
lewandom: @francinelucca thanks for spotting, added better type.

packages/react/src/components/ListBox/ListBoxSelection.tsx
lewandom: @francinelucca the error is:
```
Property 'disabled' does not exist on type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>'
```

I guess `disabled` is not a correct property for a `<div/>`... :)

packages/react/src/components/Dropdown/Dropdown.tsx
lewandom: @francinelucca changed.

packages/react/src/components/ListBox/ListBoxSelection.tsx
lewandom: Looks like MultiSelect uses <ListBox.Selection/> component.

packages/react/src/components/ListBox/ListBoxSelection.tsx
francinelucca: should this be a `button` instead of a `div` with `role="button` 🤔, any thoughts here? The div doesn't seem to accept a disabled attribute indeed https://www.w3schools.com/tags/att_disabled.asp. any thoughts here @carbon-design-system/developers ?

packages/react/src/components/ListBox/ListBoxSelection.tsx
francinelucca: tested and works both ways, so this should be good 👍

packages/react/src/components/ListBox/ListBoxSelection.tsx
lewandom: This seems related: https://github.com/facebook/react/issues/8977. I think it's safe to remove the `disabled` attribute entirely, given  `aria-disabled` is already set on this node. More thoughts welcome.

lewandom: @francinelucca I pushed a commit that removes the `disabled` attribute entirely. The goal of the typescript migration was to make as little changes as possible, so moving to `<button/>` is not the best idea here. Removing `disabled` attribute from that div is no harm given the link stated above, since it's not being respected anyway.

lewandom:
@francinelucca I fixed the formatting failure. Looks like I also accidentally removed @tay1orjones from the reviewers list...

jpsorensen:
> Looks like I also accidentally removed @tay1orjones from the reviewers list...

Did that happen when you re-requested review? I had a similar thing happen and I'm almost sure I didn't remove the second reviewer myself.

lewandom:
> > Looks like I also accidentally removed @tay1orjones from the reviewers list...
> 
> Did that happen when you re-requested review? I had a similar thing happen and I'm almost sure I didn't remove the second reviewer myself.

@jpsorensen yes, that's what I did and that's what's happened.

packages/react/src/components/ListBox/ListBoxSelection.tsx
tay1orjones: @lewandom @francinelucca yes, can confirm we want to remove `disabled` here. Since the ListBox components are not exported and available to the consumer yet (#6177), removing this prop is not considered a breaking change.

lewandom: @tay1orjones since the changes for typescript were supposed to be minimal, I'd rather keep the component prop - it still has uses, e.g. for classes and event handling. I only removed setting the `disabled` attribute on the `div` element inside the component. Hope that's fine.

tay1orjones (approved):
Looks great, thank you! 🏆

packages/react/src/components/ListBox/ListBoxSelection.tsx
francinelucca: I'll open a separate ticket to completely remove the `disabled` prop from `ListboxSelection`

francinelucca (approved):

packages/react/src/components/ListBox/ListBoxSelection.tsx
tay1orjones: Ah yes totally fair, let's leave it in for now with an issue to dive in separately

francinelucca: https://github.com/carbon-design-system/carbon/issues/13054

PR #13025, opened 2023-01-25
test(TreeView): updated treeview test to use rtl

### LINKED ISSUE #12685: Use RTL for testing in `TreeView/TreeView-test.js`

TannerS:
Closes #12685 

#### Changelog

Changed TreeView test to use RTL

#### Testing / Reviewing

Ran jest to test


tay1orjones (commented):
This is looking pretty solid! We just chatted about this separately but want to reflect here that this needs to be updated to follow the template outputted from [the scaffolding script](https://github.com/carbon-design-system/carbon/issues/10184#issuecomment-1282894448).

TannerS:
> This is looking pretty solid! We just chatted about this separately but want to reflect here that this needs to be updated to follow the template outputted from [the scaffolding script](https://github.com/carbon-design-system/carbon/issues/10184#issuecomment-1282894448).

hey @tay1orjones so i can generate it and then copy over my existing test to it?

tay1orjones:
@TannerS Yeah - you might copy your existing test file over to another location so the script can overwrite it and then migrate your tests into the new file.

TannerS:
Thank you

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

TannerS:
@tay1orjones do the test have to match the generated template exactly?

TannerS:
@tay1orjones  updated

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+      render(
+        <TreeView active="Node 1" label="Tree">
+          <TreeNode data-testid="Node 1" id="Node 1" label="Node 1" />
+          <TreeNode id="Node 2" label="Node 2" />
+        </TreeView>
+      );
+
+      const node = screen.getByTestId('Node 1');
+
+      expect(node).toHaveClass('cds--tree-node--active');
```
tay1orjones: We've been following a pattern in other files to put the classname prefix into a constant at the top of the file. 

```
const prefix = 'cds';


describe('renders as expected - Component API', () => {

....

    expect(node).toHaveClass(`${prefix}--tree-node--active`);

```

This way the test file reads identically as to how the prefix is used in component source files.

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+        </TreeView>
+      );
 
-  it('should render', () => {
-    expect(wrapper).toMatchSnapshot();
-  });
+      const nodeParent = screen.getByTestId('Node 1');
+      const nodeChild = screen.getByTestId('Node 2');
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
```
tay1orjones: This is covered by the label test further down in the file
```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
-    expect(wrapper.find('TreeLabel').text()).toBe('Tree view');
-    wrapper.setProps({ hideLabel: true });
-    expect(wrapper.find('TreeLabel').text()).toBeFalsy();
-  });
+      fireEvent.click(screen.getByTestId('Node 1'));
 
-  describe('Single node selection', () => {
-    it('should be able to preselect a node', () => {
-      expect(wrapper.find(`.${prefix}--tree-node--selected`).text()).toBe('1');
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
-      wrapper.find('TreeNode[value="2"]').simulate('click');
-      expect(onTreeSelect).toHaveBeenCalledTimes(1);
+    it('should respect selected prop', () => {
+      render(
+        <TreeView label="Tree View" selected={['Node 1']}>
+          <TreeNode id="Node 1" data-testid="Node 1" label="Node 1" />
+        </TreeView>
+      );
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+        <TreeView label="Tree View">
+          <TreeNode label="Node 1" />
+          <TreeNode label="Node 2" />
+        </TreeView>
+      );
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
+      expect(screen.getByText('Node 1')).toBeInTheDocument();
+      expect(screen.getByText('Node 2')).toBeInTheDocument();
+    });
```
tay1orjones: This feels like a duplicate of `should render children as expected`, do you think so or no?

```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+          <TreeNode data-testid="Node 1" label="Node 1" isExpanded={true}>
+            <TreeNode data-testid="Node 2" label="Node 2" />
+          </TreeNode>
+        </TreeView>
+      );
+
+      const nodeParent = screen.getByTestId('Node 1');
+      const nodeChild = nodeParent?.querySelector('div > span');
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+            isExpanded={true}>
+            <TreeNode isExpanded={true} data-testid="Node 3" label="Node 3" />
+          </TreeNode>
+        </TreeView>
+      );
+
+      const nodeParent = screen.getByTestId('Node 1');
+      const nodeChild = screen.getByTestId('Node 2');
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+      expect(nodeChild).toHaveClass('cds--tree-node--disabled');
```
tay1orjones: Could this instead query against the actual disabled attribute? Or aria-disabled? I'm not sure which the component uses.

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+      expect(within(nodeParent).getByText('Node 1')).toBeInTheDocument();
+      expect(within(nodeChild).getByText('Node 2')).toBeInTheDocument();
+      expect(within(nodeChild).getByText('Node 3')).toBeInTheDocument();
```
tay1orjones: I think these aren't necessary to test the disabled prop. What do you think?
```suggestion
```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+    });
+
+    it('should render tree with icons', () => {
+      render(
+        <TreeView label="Tree View">
+          <TreeNode renderIcon={Document} data-testid="Node 1" label="Node 1" />
+        </TreeView>
+      );
```
tay1orjones: Instead of using a live icon and just asserting that the icon class is applied, we can encapsulate a custom icon within a mock. Here's an example from SideNavMenu

https://github.com/carbon-design-system/carbon/blob/4cb3f0e6cce19a298044b26e977950e8fdd52d8b/packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js#L47-L60

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
```
tay1orjones:
```suggestion

```

packages/react/src/components/TreeView/TreeView-test.js
```diff
@@ -1,95 +1,219 @@
...
-    expect(wrapper).toMatchSnapshot();
-  });
+      const nodeParent = screen.getByTestId('Node 1');
+      const nodeChild = screen.getByTestId('Node 2');
+
+      expect(screen.getByLabelText('Tree View')).toBeInTheDocument();
+      expect(nodeParent).toHaveClass('cds--tree-parent-node');
+      expect(nodeChild).toHaveClass('cds--tree-leaf-node');
+      expect(within(nodeParent).getByText('Node 1')).toBeInTheDocument();
+      expect(within(nodeChild).getByText('Node 2')).toBeInTheDocument();
```
tay1orjones, line 56: What's the difference between using `within()` like this vs the native `nodeParent.contains(nodeChild)`?

tay1orjones (changes requested):
Hey thanks so much for this, and sorry for the delay in review!

Overall looks good - I found some redundancy in the tests though and a couple spots I think we could improve. Let me know what you think

Also in general see if we can reduce the reliance on classnames. If we can instead assert against `aria-selected`, `disabled`, etc. it would be preferred. Classnames are not subject to semver requirements and may change. Having as little tests as possible relying on them will help these tests be more resilient to change. In some cases it's unavoidable though.

packages/react/src/components/TreeView/TreeView-test.js
TannerS: From what i understand, `within` will return a....container?, where it can be returned from other rtl queries and search for all the queries within in like this example 

```
 takes a DOM element and binds it to the raw query functions, allowing them to be used without specifying a container
```

https://testing-library.com/docs/dom-testing-library/api-within/

where the old way seems to be more of if this container is inside a parent one without considering anything else

could this both work? maybe, i just went off the RTl way of doing it tbh

tay1orjones (approved):
LGTM!

tw15egan (approved):
LGTM 👍 ✅

PR #13028, opened 2023-01-25
fix(ExpandableTile): remove extra onClick call

### LINKED ISSUE #12963: [Bug]: ExpandableTile onClick being called twice
### Package

@carbon/react

### Browser

_No response_

### Package version

1.20.1

### React version

16.14.0

### Description

We have logic to call when expanding the tile but the onClick is being called twice, both when clicking and when hitting the Enter key.

I'm making this a Sev 4 because we could implement a workaround but I'd really rather not implement that complexity.

### Suggested Severity

### Reproduction/example

https://stackblitz.com/edit/github-odxajz

### Steps to reproduce

In the StackBlitz, click the Open in New Tab button so that you can open the dev tools to see the console panel.

Click on the tile.  Note that 'onclick" is printed twice to the console.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12963

Since we are using the `composeEventHandlers` internal tool, we do not need to manually invoke `onClick` at the end of `handleClick`. 

This also fixes an issue causing `isTileMaxHeight` to run too many times and set an incorrect value, resulting in a squished `Above The Fold` view.

#### Changelog

**Changed**

- Added `isTileMaxHeight` to the computed style  `useIsomorphicEffect` dependency array

**Removed**

- Extraneous `onClick` call

#### Testing / Reviewing

Go to `Expandable with Interactive` story, and ensure the component is fully rendered and not squished.  Next, open up dev tools. Only one `on click` should appear when the Expandable icon is clicked.


tw15egan:
@GiacomoP99 I added a few more `Tile` components to the interactive page, which seems to be working for me. Any ideas about what the difference in implementation may be? 

Screenshot 2023-01-26 at 11 21 30 AM

andreancardona (approved):
@tw15egan tested at different breakpoints looks good!

tay1orjones (approved):

PR #13031, opened 2023-01-26
test(password-input): update to rtl

### LINKED ISSUE #12678: Use RTL for testing in `TextInput/PasswordInput-test.js`

aledavila:
Closes #12678

Updates old test to RTL

#### Changelog

**New**

- all props are tested

**Changed**

- enzyme to RTL

**Removed**

- removed old test file

#### Testing / Reviewing

You can test locally by running `yarn` test packages/react/src/components/TextInput/__tests__/PasswordInput-test.js` in the root folder.

Screen Shot 2023-01-25 at 11 46 09 PM


packages/react/src/components/TextInput/__tests__/PasswordInput-test.js
```diff
@@ -0,0 +1,352 @@
...
+        />
+      );
+
+      expect(screen.getByTestId('test-id')).toHaveAttribute(
+        'value',
+        'This is default text'
+      );
+    });
+
+    it('should respect disabled prop', () => {
```
tay1orjones, line 54: In addition to the existing assertion expecting it to be disabled, should we validate that onClick and onChange are not invoked as was done in the previous test?

packages/react/src/components/TextInput/__tests__/PasswordInput-test.js
```diff
@@ -0,0 +1,352 @@
...
+    it('should spread extra props onto input element', () => {
+      render(
+        <PasswordInput
+          data-testid="test-id"
+          id="input-1"
+          labelText="TextInput label"
+        />
+      );
+
+      expect(screen.getByTestId('test-id')).toHaveClass(
```
tay1orjones: If we're specifically looking for the input element, `getByLabelText` can be used instead of `getByTestId`.

```suggestion
      expect(screen.getByLabelText("TextInput label")).toHaveClass(
```

packages/react/src/components/TextInput/__tests__/PasswordInput-test.js
```diff
@@ -0,0 +1,352 @@
...
+      render(
+        <PasswordInput
+          data-testid="test-id"
+          id="input-1"
+          labelText="PasswordInput label"
+          className="custom-class"
+        />
+      );
+
+      expect(screen.getByTestId('test-id')).toHaveClass('custom-class');
```
tay1orjones: Same here
```suggestion
      expect(screen.getByLabelText('PasswordInput label')).toHaveClass('custom-class');
```

packages/react/src/components/TextInput/__tests__/PasswordInput-test.js
```diff
@@ -0,0 +1,352 @@
...
+      render(
+        <PasswordInput
+          data-testid="test-id"
+          id="input-1"
+          labelText="PasswordInput label"
+          defaultValue="This is default text"
+        />
+      );
+
+      expect(screen.getByTestId('test-id')).toHaveAttribute(
```
tay1orjones:
```suggestion
      expect(screen.getByLabelText('PasswordInput label')).toHaveAttribute(
```

packages/react/src/components/TextInput/__tests__/PasswordInput-test.js
```diff
@@ -0,0 +1,352 @@
...
+import { render, screen } from '@testing-library/react';
+
+const prefix = 'cds';
+
+describe('PasswordInput', () => {
+  describe('renders as expected - Component API', () => {
+    it('should spread extra props onto input element', () => {
+      render(
+        <PasswordInput
+          data-testid="test-id"
```
tay1orjones:
```suggestion
```

packages/react/src/components/TextInput/PasswordInput-test.js
```diff
@@ -1,213 +0,0 @@
...
-    );
-
-    const passwordInput = () => wrapper.find('input');
-
-    describe('input', () => {
-      it('renders as expected', () => {
-        expect(passwordInput().length).toBe(1);
-      });
-
-      it('should accept refs', () => {
```
tay1orjones, line 27: Is there a test for refs in the new file? I think we'd just want to assert that it's placed on the input element. Something like this might work:

```js
  it('should support a `ref` that is placed on the <input> element', () => {
    const ref = jest.fn();
    render(
        <PasswordInput
          labelText="PasswordInput label"
          ref={ref}
        />
      );
    expect(ref).toHaveBeenCalledWith(screen.getByLabelText('PasswordInput label'));
  });
```

I found quite a few spots in the codebase we already do similar versions of this - do a global search for `` `ref` `` in `*-test.js` files for more examples.

tay1orjones (changes requested):
This looks great! There's a couple spots I'm curious if we could increase parity with the old tests - I marked them below.

Also I marked a few spots, but across the board I think we should use `getByLabelText` instead of `getByTestId` where possible. Anywhere we do this we can also remove the `test-id` that is placed on the rendered component.

aledavila:
@tay1orjones great thanks that makes sense I was having trouble figuring what the best way to grab the input was.

aledavila:
@tay1orjones updated tests

tay1orjones (approved):
Awesome! 🎉  Thanks for incorporating that feedback

andreancardona (approved):
tested and looks good!

Issue #12513, opened 2022-11-07
Add TypeScript types to `@carbon/react` components


tay1orjones:
This issue is to track the implementation of TypeScript definitions for components exported from `@carbon/react`.

The lists below organize all exports from the package that need to have types added into three categories: high, medium, and low priority. There is a [project board](https://github.com/orgs/carbon-design-system/projects/53) for this effort as well.

### How to contribute

- If you'd like to contribute, **please comment in the component issue stating your intent to work on an issue.**
- If that person has not submitted a PR within 10 days, consider it available again.

Thanks so much for assisting with this huge effort! 🙏 

### Every checklist item below needs the following:

1. Add component types for all components listed in the title by following the [adding component types documentation](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md#steps-to-provide-baseline-type-definitions-for-components)
1. Submit a pull request for the changes
    - Please keep pull requests as small as possible
    - Avoid adding additional components to a single PR unless necessary
    - [Link to close the related issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword#linking-a-pull-request-to-an-issue-using-a-keyword), e.g. `Closes #12513`

---

REF https://github.com/carbon-design-system/carbon/issues/16360

# High priority

### DataTable
- [x] #12515
- [x] #12516
- [x] #12517
- [x] #12518
- [x] #12519
- [x] #12520
- [x] #12521
- [x] #12522
- [x] #12523
- [x] #12524
- [x] #12525
- [x] #12526
- [x] #12527
- [x] #12528
- [x] #12529
- [x] #12530
- [x] #12531
- [x] #12532
- [x] #12533
- [x] #12534
- [x] #12535
- [x] #12536

### Grid
- [x] #12537

### Inputs
- [x] #12654
- [x] #12655
- [x] #12538
- [x] #12539
- [x] #12540
- [x] #12541
- [x] #12542
- [x] #12543
- [x] #12544
- [x] #12545
- [x] #12546
- [x] #12547
- [x] #12548
- [x] #12549
- [x] #12550
- [x] #12551
- [x] #12552
- [x] #12553
- [x] #12554
- [x] #12555
- [x] #12556
- [x] #12557
- [x] #12558
- [x] #12559
- [x] #12560
- [x] #12561
- [x] #12562
- [x] #12563
- [x] #12564
[...]

PR #13042, opened 2023-01-28
Add Typescript annotations to TableSelectRow

### LINKED ISSUE #12531: Add TypeScript types to TableSelectRow

jpsorensen:
Closes #12531 

Convert existing propType definitions to Typescript annotations on the TableSelectRow component. This is part of a broader effort to add Typescript annotations to components, tracked in https://github.com/carbon-design-system/carbon/issues/12513.

#### Changelog

**Changed**

- TableSelectRow.js has been converted to tsx

#### Testing / Reviewing

- In the packages/react directory, run yarn build, yarn storybook. In the storybook, navigate to DataTable>Selection and confirm that the component's behavior remains unchanged.
- Past usages of TableSelectRow at the bottom of TableSelectRow.tsx. Confirm that they compile as expected. Modify inputs so that they have the wrong types and confirm that it does not compile. Remove a required parameter and confirm that it does not compile. (I used the examples from the TableSelectRow unit tests).


packages/react/src/components/DataTable/TableSelectRow.tsx
```diff
@@ -12,6 +12,57 @@ import InlineCheckbox from '../InlineCheckbox';
...
 import { usePrefix } from '../../internal/usePrefix';
 
+export interface TableSelectRowProps {
+  /**
+   * Specify the aria label for the underlying input control
+   */
+  ariaLabel: string;
+
+  /**
+   * Specify whether all items are selected, or not
```
jpsorensen: I'm guessing that this was accidentally copied over from TableSelectAll. I decided to leave it as-is (leave it in the PropTypes docs, and to have the Typescript docs mirror the PropTypes docs) in this PR so as not to get a separate change mixed up in this Typescript conversion. Just let me know if you'd prefer that I fix this as part of this PR!

packages/react/src/components/DataTable/TableSelectRow.tsx
```diff
@@ -34,7 +85,8 @@ const TableSelectRow = ({
   };
   const InlineInputComponent = radio ? RadioButton : InlineCheckbox;
   const tableSelectRowClasses = classNames(`${prefix}--table-column-checkbox`, {
-    [className]: className,
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    [className!]: className,
```
francinelucca: think this way you can forego having to disable the eslint rule
```suggestion
    ...(className && {[className]: true}),
```

packages/react/src/components/DataTable/TableSelectRow.tsx
francinelucca: nice catch! I do believe that's a mistake, I think it'll be fine to fix here as to not carry over more mistakes, we can rename here and in the proptypes to something like "Specify whether row is selected, or not" if you don't mind adding it 🙏🏻

francinelucca (approved):
Thanks for getting this PR up! LGTM, just two minor details

packages/react/src/components/DataTable/TableSelectRow.tsx
jpsorensen: Done, thank you!

tay1orjones (approved):

PR #13057, opened 2023-01-31
fix(toggle): ensure onToggle still work with props['aria-labelledby']

### LINKED ISSUE #13004: [Bug]: Toggle labelA and labelB are still being ignored when hideLabel is tur
### Package

@carbon/react

### Browser

_No response_

### Package version

1.2.1

### React version

16.14.0

### Description

The intent of the last fix for https://github.com/carbon-design-system/carbon/issues/12896 was to allow us to still use labelA and labelB when using hideLabel and aria-labelledby to label the component in an external element.  However, there is now no label next to the toggle at all.

### Suggested Severity

### Reproduction/example

https://stackblitz.com/edit/github-jsd1gh

### Steps to reproduce

See that the toggle has no label to right of the toggle, even though I have defined labelA and labelB and have added the aria-labelledby with my hideLabel.

### Code of Conduct

janhassel:
Closes #13004

#### Changelog

**New**

- If `props['aria-labelledby']` is passed the wrapper element listens for a `click` event to trigger `handleClick` as the labeling element is no longer a `<label>` but a `<div>` and the `<button>` is visually hidden.

**Changed**

- Clarified the relationship between `props.hideLabel` and `props.labelText` in the prop description of `props.hideLabel`

#### Testing / Reviewing

- In the storybook, add a dummy `props.onClick` and `props.onToggle` function and verify that they are triggered both by clicking on the toggle and by using the keyboard **when `props['aria-labelledby']` is passed**.
  - Verify that these functions are only called once per click/keyboard interaction.
  - Verify that these functions are also called when `props['aria-labelledby']` is not passed.


francinelucca (approved):
🚀

tay1orjones (approved):
LGTM - Thanks for this!

PR #13061, opened 2023-01-31
fix(FileUploaderButton): refactor to button to remove a11y violation

### LINKED ISSUE #12725: [a11y]: FileUploader has critical violation "The tabbable element's role 'none' is not a widget role"
### Package

carbon-components-react

### Browser

Chrome

### Operating System

Windows

### Package version

n/a

### React version

n/a

### Assistive technology

_No response_

### Description

The FileUploader button has a critical a11y violation "The tabbable element's role 'none' is not a widget role" that was found in our product using the IBM Accessibility Checker tool and is re-producible on the [v10](https://v7-react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader) and [v11](https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader) storybook as well.

v10 storybook

FileUploaderCarbon
FileUploaderCarbonRef

### WCAG 2.1 Violation

_No response_

### Reproduction/example

Carbon v10 storybook - https://v7-react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader

### Steps to reproduce

On the storybook link (either v10 or v11) for the FileUploader story, run the IBM Accessibility Checker scan. The scan should report the critical violation mentioned above (scroll to the bottom of the violations list).

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12725

Refactors the `FileUploaderButton` to use a `button` instead of utilizing the `label` to handle click/keyboard interactions

#### Changelog

**New**

- `button` is rendered adjacent to the `label` element in `FileUploaderButton`

**Changed**

- Moved most props from the `label` element to the new `button` element. All functionality should remain the same. 
- Updated tests to utilize the `button` for `onClick` tests

**Removed**

- Removed `role="button"` wherever it was being set, as it is now rendering a button by default
- Removed the `tabIndex` as this is handled automatically when using a `button`

#### Testing / Reviewing

Ensure all of the following:
- All tests pass (especially a11y checker)
- Component renders as expected
- Existing props work as expected
- Functionality remains the same (`onClick`, `onChange` can be passed down)


packages/react/src/components/FileUploader/FileUploaderButton.js
```diff
@@ -78,17 +77,20 @@ function FileUploaderButton({
 
   return (
     <>
-      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
-      <label
-        ref={innerRef}
-        tabIndex={disabled ? -1 : tabIndex || 0}
```
tay1orjones, line 84: Should we deprecate `tabIndex` since it's no longer explicitly used like this? With the new setup `tabIndex` will be passed through `...other` onto `<button>`, right?

If we deprecate, what would the message be? They technically can still pass a `tabIndex`, it's just won't be included in the prop types anymore.

tay1orjones (approved):
Looks good! One question below

aledavila (approved):

PR #13092, opened 2023-02-04
fix(react): update external settings to exclude node_modules from bundle

joshblack:
Hi there! 👋 

This PR updates the `external` setting in the rollup config to better match on nested imports to modules. Currently, the setting will only match on exact imports and so certain dependencies are being brought into the package: https://unpkg.com/browse/@carbon/react@1.21.0/es/node_modules/

**Before**

```js
// Marked as external
import 'module-name';

// Is not marked as imported because it is not an exact match
import 'module-name/path/to/nested/module';
```

**After**

```js
// Marked as external
import 'module-name';

// Marked as external
import 'module-name/path/to/nested/module';
```

#### Changelog

**New**

**Changed**

- Update `external` config for rollup by transforming packages into RegEx's

**Removed**

#### Testing / Reviewing

- Run `yarn build` in `packages/react`
- Assert that `node_modules` is no longer included in the ESM and CommonJS bundles


francinelucca (commented):
Confirmed no longer see `node_modules` in the bundles 🚀, looks like CI checks are failing though 🤔

francinelucca (approved):

tay1orjones (approved):
Hey there! Thanks for this 🎉

Issue #12721, opened 2022-11-23
[a11y]: Dropdown and multi-select dropdown inconsistent and have implementation issues


mbgower:
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

React library

### React version

https://react.carbondesignsystem.com/?path=/story/components-dropdown--default

### Automated testing tool and ruleset

n/a

### Assistive technology

n/a

### Description

There are interaction differences and challenges between the React Dropdown and [Multi-select dropdown](https://react.carbondesignsystem.com/?path=/story/components-multiselect--playground) 

### Issue 0 Inconsistent categorization between Storybook and documentation
I originally just mentioned this in passing, but I think it is worth flagging as an issue: Multi-select is a separate component in Storybook (under M), but is just a variation on Dropdown in the Carbon guidance and on the Figma design tool. I think it should consistently be co-located in Storybook as well. If there is concern with the amount of samples under one topic, you could always call it Dropdown Multiselect so that it as least appears next in the left navigation.

### Issue 1 Inconsistent keystrokes
For both, they should open on `Enter`, `Spacebar`, `Down Arrow` and (optionally) `Alt+Down Arrow` (`Cmd+Down Arrow` on Mac). 
This is the behaviour for the Dropdown (good), but the Multi-select only opens on Enter and Spacebar (bad). Especially because unselected dropdowns/multiselects look the same, they should operate the same.

### Issue 2 Inconsistent behaviour on open
When they open, the currently shown option should be given focus /highlighted in the dropdown list (see the first screenshot under issue 3 for what this should look like) .
This is not the behaviour for any of them.
For the Dropdown, if the **Enter or Spacebar** is selected, the default value "Dropdown menu options" is not even displayed in the list, so nothing in the list takes focus. (Multi-select has the same behaviour.)
![image](https://user-images.githubusercontent.com/14298245/203630075-7e7159cc-67a4-4151-ad7e-d453be3d1455.png)
#### Issue 2b
[...]

PR #13097, opened 2023-02-06
docs(dropdown): add initial selected item to dropdown storybook demo

alisonjoseph:
Ref #12721 "Issue 3 There is no placeholder in a dropdown; Every shown value should be an option."

This adds Add initial selected item for dropdown stories to align with correct guidance that there shouldn't be a placeholder in a dropdown.

#### Changelog

- Add initial selected item for dropdown stories

#### Testing / Reviewing

Storybook demos for dropdown should now show the initial selected item and not the label field.


tw15egan (approved):
LGTM 👍 ✅

aledavila (approved):

PR #13108, opened 2023-02-06
fix(date-picker): correct deleted end range behavior

### LINKED ISSUE #12734: [a11y]: DatePicker range variety - clears start date after tabbing off end date
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.17.0

### React version

_No response_

Manual

### Assistive technology

Keyboard

### Description

Typing an end date and tabbing off of the field clears the start date

### WCAG 2.1 Violation

_No response_

### Reproduction/example

Storybook

### Steps to reproduce

See https://react.carbondesignsystem.com/?path=/docs/components-datepicker--range-with-calendar
* Tab to the Range Datepicker 
* Type a valid date into the "Start Date"
* Press tab (focus now moves to the "End Date" and keeps the start date)
* Type a valid date into the "End Date"
* Press tab (focus now moves off the widget. The "End Date" is kept, but the "Start Date" now clears)

Note - this does not happen if you first set dates using the mouse

### Code of Conduct

francinelucca:
Closes #12734 

Prevents end date from being deleted on datepicker change when user has not made changes to input. 

#### Changelog

**Changed**

- packages/react/src/components/DatePicker/DatePicker.js: extend `onChange` behavior to prevent end date being deleted

#### Testing / Reviewing

1.  Tab to the Range Datepicker
2. Type a valid date into the "Start Date"
3. Press tab (focus now moves to the "End Date" and keeps the start date)
4. Type a valid date into the "End Date"
5.  Tab back to Start Date (shift + tab) --  you'll have to do it twice to get to the start date, see note below
6. Tab again to End Date (Tab)

You should still see both previously selected dates displayed in the calendar. Compare to react.carbondesignsystem.com where date gets deleted

### Note

There's a "double focus" issue when tabbing off the end date after changing it manually and not pressing "enter" to trigger the change, I paired up with Taylor on this and we're fine with this behavior for now until flatpickr gets more stable upstream or we switch into a different implementation


tay1orjones (approved):
Thanks for tackling this one! Looks great - the double focus issue is much less of a concern compared to the original bug. I think we can merge this even with that slight remaining nit and circle back to it if necessary.

tw15egan (approved):
It is working as expected! 🎉 ✅ 👍🏻

PR #13112, opened 2023-02-07
fix(Toggletip): general a11y / ux improvements

### LINKED ISSUE #12922: [a11y]: toggletip implementation bugs
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-toggletip--default

### React version

_No response_

n/a

### Assistive technology

_No response_

### Description
There are four primary challenges:
1) None of the toggletips should be visible in storybook until activated. Instead, they are all showing before the i icon is triggered
2) When the toggletip is expanded and has a tabstop inside it, tabbing out of the toggletip should collapse the tip. This is not happening in the [playground](https://react.carbondesignsystem.com/iframe.html?args=defaultOpen:true&id=components-toggletip--playground&viewMode=story) or default views
3) If the focus is in the toggletip when the Esc is pressed, the focus should return to the trigger. Instead, it is unclear where it goes. There is no other place that makes sense except the toggle tip, IMO.
4) There is something called a "Toggletip label" that appears _in front of_ the i icon button. AFAIK, this cannot exist and has no place in the component. It is not listed in the anatomy of the component on the Usage page, for example. Maybe it's meant to demonstrate that the toggletip icon may follow the label for another component, but it is very confusing and should be revamped.
image

### WCAG 2.1 Violation

Focus Order

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?args=defaultOpen:true&id=components-toggletip--playground&viewMode=story

### Steps to reproduce

1. Go to https://react.carbondesignsystem.com/iframe.html?args=defaultOpen:true&id=components-toggletip--playground&viewMode=story
2. Use a keyboard to navigate to it. The tip will already be open although it has yet to be activated (first issue)
3. On an expanded tip with a tabstop inside it, tab into the tip and keep tabbing until the focus leaves the tip. The tip is not collapsed (second issue)
4. On an expanded tip with a tabstop inside it, tab into the tip and press Esc. The trigger will not get focus (third issue)

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12922

Addresses general UX / a11y issues found in the `Toggletip` component

#### Changelog

**New**

- `Toggletip` will now close when the menu is tabbed through, and focus leaves the popover body.
- If <kbd>Escape</kbd> is pressed while an item inside the `Toggletip` is focused, the focus will return to the `Toggletip` trigger 

**Changed**

- Updated `ToggletipLabel` to `body-01` to match the design specs
- Labelled the stories better to note that `Toggletip` is not normally rendered `open` by default

#### Testing / Reviewing

- Open the `Toggletip` story 
- <kbd>Tab</kbd> through the component. Focus should go to the trigger, then press <kbd>Space</kbd> to open the menu. <kbd>Tab</kbd> through the component, and the menu should close when the `focus` leaves the popover body.
-  <kbd>Tab</kbd> through the component. While an item inside has `focus` (not the trigger), hit the <kbd>Escape</kbd> key. The `focus` should return to the trigger.
- Ensure all existing functionality remains


aledavila (approved):
lgtm works as expected !

packages/react/src/components/Toggletip/index.js
```diff
@@ -95,9 +95,33 @@ function Toggletip({
...
+    if (open && event.relatedTarget === null) {
+      return;
+    }
+    if (!event.currentTarget.contains(event.relatedTarget)) {
+      // The menu should be closed when focus leaves the `Toggletip`  (#12922)
+      actions.close();
     }
   }
 
+  // If the `Toggletip` is the last focusable item in the tab order, it shoudl also close when the browser window loses focus  (#12922)
```
tay1orjones, line 118:
```suggestion
  // If the `Toggletip` is the last focusable item in the tab order, it should also close when the browser window loses focus  (#12922)
```

tay1orjones (approved):

PR #13129, opened 2023-02-10
feat(Heading): add prop to override level of a section

remolueoend:
In certain situations, such as when rendering headings as part of the content of a modal, one may wish to set the start level of the first section explicitly. This PR introduces the optional property `level` on section components, allowing users to override the level of a section.

#### Changelog

The Section component accepts an optional property `level`, allowing users to set the section's level explicitly.

#### Testing / Reviewing

This feature is tested using component tests and available as storybook component.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1031,6 +1031,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "remolueoend",
+      "name": "remolueoend",
+      "avatar_url": "https://avatars.githubusercontent.com/u/7881606?v=4",
+      "profile": "https://github.com/remolueoend",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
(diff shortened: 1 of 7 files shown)


remolueoend:

remolueoend:
Hey there Carbon team, I hope this PR finds you well.

The tests are currently failing because my proposal introduces a change to the public API, requiring at least a minor version bump. I tried to fix it by running `lerna version`, but failed to do so: I was only able to bump the version of all packages or none. Do you have any tips how to proceed in this case?

Thank you very much for looking into this PR, I hope my contribution turns out to be useful.

tay1orjones (commented):
Hey there, thanks for suggesting this change!

The intent of Section and Heading is to follow the [HTML heading spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) and provide an automatic hierarchy of headings that do not skip. Adding this escape hatch feels counterintuitive to that original intent, and somewhat confusing from an API design standpoint.

Could you elaborate more on what problem this solves for you? Modal's `modalHeading` is intentionally an `h2` and `modalLabel` an `h3`.

remolueoend:
Hey there, thanks a lot for looking into my PR.

Sure thing, I may have misunderstood the overall concept. Because modal headings and modal labels are rendered as `h2` and `h3` respectively, I assumed that the first section header appearing in a modal content should always be an `h4`, independent of the structure of sections on the page that renders the modal. To achieve this, I was thinking of wrapping the content of a modal with `<Section level={4}>{children}</Section>`.

But then again, I may have misunderstood the concept of sections in modals :)

Thanks again!

remolueoend:
Hi @tay1orjones , I tried to do a bit of research about accessibility of headings in modals. It seems that most sources support the same approach as Carbon: the heading level of the title of a modal is fixed (e.g. `h2`) and does not depend on the heading structure of the underlying page. My question now is: if I need to render a heading as part of the content of a modal, what level should it have? My assumption is `h4` in case of Carbon. Setting the level explicitly on the first section in the modal content would solve this.

I'd love to hear your thoughts about this!

tay1orjones (approved):
After thinking some more about this, I think there's going to be cases like this where components need to specify a set of headings without using `Heading`. So providing a way to override `Heading` to start where we need it to like this makes sense.

Do note that in the case of `Modal` (and `ComposedModal`) the `label` (h2) and `title` (h3) are both optional. If `title` is not configured, you'd probably want to start the hierarchy with an `h3` instead of `h4`

francinelucca (approved):

remolueoend:
> Do note that in the case of `Modal` (and `ComposedModal`) the `label` (h2) and `title` (h3) are both optional. If `title` is not configured, you'd probably want to start the hierarchy with an `h3` instead of `h4`

Ah true, almost forgot about this case.

Thank you all for the discussion and accepting my PR. Hope to hear from you for another contribution!

PR #13142, opened 2023-02-13
feat(TextInput): add typescript types

### LINKED ISSUE #12561: Add TypeScript types to TextInput

bianca-sparxs:
Closes #12561 

Part of #12513 to convert the propTypes of the TextInput component into typescript type interface.

#### Changelog

**Changed**

- TextInput.js converted to .tsx
- Typescript interface added to reflect proptypes and definitelyTyped types of this component
    useNormalizedInputProps jsdocs updated to reflect more accurate input types, since Typescript will infer a Typescript definition from those docs
-  add self to contributors

#### Testing / Reviewing

- Take a storybook component that uses TextInput, paste at the bottom of the `.tsx` file and make changes to the component  (i.e add props, add incorrect types) and verify types are correctly enforced


bianca-sparxs:

bianca-sparxs:
recheck

packages/react/src/components/TextInput/TextInput.tsx
```diff
@@ -6,7 +6,7 @@
  */
 
 import PropTypes from 'prop-types';
-import React, { useContext, useState } from 'react';
+import React, { ReactNode, useContext, useState } from 'react';
 import classNames from 'classnames';
 import { useNormalizedInputProps } from '../../internal/useNormalizedInputProps';
 import PasswordInput from './PasswordInput';
```
bianca-sparxs, line 12: `import * as FeatureFlags from '@carbon/feature-flags';` gives me an error saying that the module is not found but this line is unchanged from when i converted it to .tsx, thought I would bring it up to get comments on that?

bianca-sparxs:
Hey @francinelucca thanks, merge conflicts were resolved! :)

packages/react/src/components/TextInput/TextInput.tsx
bianca-sparxs: not related to code, did not run `npm run build` at that time

packages/react/src/components/TextInput/TextInput.tsx
```diff
@@ -94,6 +220,7 @@ const TextInput = React.forwardRef(function TextInput(
     title: placeholder,
     disabled: normalizedProps.disabled,
     readOnly,
+    //eslint
```
francinelucca: is this line necessary?
```suggestion
```

packages/react/src/internal/useNormalizedInputProps.js
```diff
@@ -44,6 +44,7 @@ import { usePrefix } from './usePrefix';
  * @param {InputProps} props - The props passed to the component
  * @returns {NormalizedInputProps}
  */
+/** @type any */
```
francinelucca: you can probably get rid of this by documenting the `helperId` property in `NormalizedInputProps` above; something like `* @property {string} helperId - id used for helper text` should work

packages/react/src/components/TextInput/TextInput.tsx
```diff
@@ -220,7 +346,9 @@ const TextInput = React.forwardRef(function TextInput(
 });
 
 TextInput.displayName = 'TextInput';
+// @ts-expect-error types not yet provided for this component
 TextInput.PasswordInput = PasswordInput;
+// @ts-expect-error types not yet provided for this component
```
francinelucca: I think this would still error out even if the components had been migrated, the error seems to be that those properties are not defined in `TextInput`
image

francinelucca (changes requested):
Thanks for putting this together @bianca-sparxs! 

Mostly LGTM, just some minor change requests and looks like there's some merge conflicts 🙏🏻

packages/react/src/components/TextInput/TextInput.tsx
bianca-sparxs: that was a typo, thanks for the catch

francinelucca (approved):

alisonjoseph (approved):

PR #13144, opened 2023-02-13
Add TypeScript annotations to MultiSelect

### LINKED ISSUE #12547: Add TypeScript types to MultiSelect

awarrier99:
Closes #12547 

Convert existing propType definitions to TypeScript annotations on the MultiSelect component. This is part of a broader effort to add TypeScript annotations to components, tracked in #12513.

#### Changelog

**Changed**

- `MultiSelect.js` has been converted to `MultiSelect.tsx`
- Added awarrier99 to contributor list
- Changed a `MultiSelect` story to include required prop `label`
- Changed a `MultiSelect` test to return an empty span instead of an empty string for `itemToElement` (a string is not valid as a JSX constructor, which is how the prop is used in the component)
- Changed `MultiSelect` sorting options to use `locale` prop, rather than hardcoded 'en'

#### Testing / Reviewing

- In the `packages/react` directory, run `yarn build`, `yarn storybook`. In the storybook, navigate to `MultiSelect` and confirm that the component's behavior remains unchanged.
- In `MultiSelect.stories.js` and `MultiSelect-test.js`, add `// @ts-check` as the first line in the file (temporarily, for the sake of this test). Confirm that the files do not have compilation errors (related to the `MultiSelect` component). Make changes to the components in those files - change number to string, etc - and confirm that compilation errors appear.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -4820,7 +4820,127 @@ Map {
         "type": "bool",
       },
       "downshiftProps": Object {
-        "type": "object",
+        "args": Array [
+          Object {
+            "children": Object {
+              "type": "func",
+            },
+            "defaultHighlightedIndex": Object {
+              "type": "number",
+            },
+            "defaultIsOpen": Object {
+              "type": "bool",
+            },
+            "environment": Object {
+              "args": Array [
+                Object {
+                  "addEventListener": Object {
+                    "type": "func",
+                  },
+                  "document": Object {
+                    "args": Array [
+                      Object {
+                        "activeElement": Object {
+                          "type": "any",
+                        },
+                        "body": Object {
+                          "type": "any",
+                        },
+                        "getElementById": Object {
+                          "type": "func",
+                        },
+                      },
+                    ],
+                    "type": "shape",
+                  },
+                  "removeEventListener": Object {
+                    "type": "func",
+                  },
+                },
+              ],
+              "type": "shape",
+            },
+            "getA11yStatusMessage": Object {
+              "type": "func",
+            },
+            "getItemId": Object {
+              "type": "func",
+            },
+            "highlightedIndex": Object {
+              "type": "number",
+            },
+            "id": Object {
+              "type": "string",
+            },
+            "initialHighlightedIndex": Object {
+              "type": "number",
+            },
+            "initialInputValue": Object {
+              "type": "string",
+            },
+            "initialIsOpen": Object {
+              "type": "bool",
+            },
+            "initialSelectedItem": Object {
+              "type": "any",
+            },
+            "inputId": Object {
+              "type": "string",
+            },
+            "inputValue": Object {
+              "type": "string",
+            },
+            "isOpen": Object {
+              "type": "bool",
+            },
+            "itemCount": Object {
+              "type": "number",
+            },
+            "itemToString": Object {
+              "type": "func",
+            },
+            "labelId": Object {
+              "type": "string",
+            },
+            "menuId": Object {
+              "type": "string",
+            },
+            "onChange": Object {
+              "type": "func",
+            },
+            "onInputValueChange": Object {
+              "type": "func",
+            },
+            "onOuterClick": Object {
+              "type": "func",
+            },
+            "onSelect": Object {
+              "type": "func",
+            },
+            "onStateChange": Object {
+              "type": "func",
+            },
+            "onUserAction": Object {
+              "type": "func",
+            },
+            "scrollIntoView": Object {
+              "type": "func",
+            },
+            "selectedItem": Object {
+              "type": "any",
+            },
+            "selectedItemChanged": Object {
+              "type": "func",
+            },
+            "stateReducer": Object {
+              "type": "func",
+            },
+            "suppressRefError": Object {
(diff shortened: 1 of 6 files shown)


awarrier99:

andreancardona:
@awarrier99 Thank you so much for the contribution! Do you mind resolving the merge conflicts please :)

awarrier99:
> @awarrier99 Thank you so much for the contribution! Do you mind resolving the merge conflicts please :)

Sure, they should be resolved now!

awarrier99:
> @awarrier99 Thank you so much for the contribution! Do you mind resolving the merge conflicts please :)

@andreancardona Sure, they should be resolved now!

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -266,13 +478,12 @@ const MultiSelect = React.forwardRef(function MultiSelect(
...
       <ListBox
-        onFocus={isFluid ? handleFocus : null}
-        onBlur={isFluid ? handleFocus : null}
+        onFocus={isFluid ? handleFocus : undefined}
+        onBlur={isFluid ? handleFocus : undefined}
         type={type}
         size={size}
         className={className}
         disabled={disabled}
-        readOnly={readOnly}
```
tw15egan, line 275: What's the reasoning for removing this prop?

awarrier99: It appears that this prop is not actually used by the `ListBox` component, I visually confirmed in the file that no such prop existed and I was also receiving a TypeScript error for passing a non-existent prop

awarrier99: Although I suppose that prop could've been contained in the `...rest` group of props in the `ListBox` component which is passed to the underlying `div`, in which case I think a change needs to be made explicitly support the prop (or add an index signature), as the only way to suppress the error is with a `@ts-ignore`

tw15egan: Thanks for clarifying. I'm not seeing any behavioral changes with removing it, so I'll defer to you on TS best practices and what you would like to do here 👍🏻

tw15egan (approved):
LGTM once the `all-contributors` merge conflict is fixed! 👍 ✅

packages/react/src/components/MultiSelect/MultiSelect.tsx
awarrier99: No problem, and sounds good - I think if the prop isn't being used for anything meaningful it would make the most sense to just remove it rather than still pass it while telling TypeScript to ignore it (just for the `ListBox` component, it's still used in other places in this component)

awarrier99:
> LGTM once the `all-contributors` merge conflict is fixed! 👍 ✅

Resolved!

awarrier99:
@andreancardona @tw15egan I noticed with the previous run of the workflows that there was an error with a snapshot saying that the public API shouldn't be changed without a new semver bump - do I need to do something to resolve that?

tw15egan:
@awarrier99 usually means snapshots need to be updated due to the changes made. If the changes are intentional (in this case, yes we are modifying the component footprint quite a bit), you can update snapshots with `yarn test -u`

awarrier99:
> @awarrier99 usually means snapshots need to be updated due to the changes made. If the changes are intentional (in this case, yes we are modifying the component footprint quite a bit), you can update snapshots with `yarn test -u`. None of these changes should be breaking so we are good to go 👍🏻

Got it, sounds good! I'll leave the snapshots as is then

awarrier99:
@tw15egan sorry I misunderstood, I thought you were saying we were good to go without needing to update the snaps - thanks for taking care of that

tw15egan:
@awarrier99, no worries, I could have worded it better! I just meant it was okay to update them because we are knowingly making changes to the API 👍🏻 Thanks for contributing 🎉

andreancardona (approved):

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -4820,127 +4820,10 @@ Map {
         "type": "bool",
       },
       "downshiftProps": Object {
-        "args": Array [
```
tw15egan: @awarrier99 circling back, is there a way to retain the `downshiftProps` here? Looking at the Dropdown ts conversion, I do not see any snapshot changes to the `downshiftProps`. I'm trying to parse the difference in implementation to see what may be causing it 🤔 

Reference PR: https://github.com/carbon-design-system/carbon/pull/13022/files

tw15egan (changes requested):

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
tw15egan: The other snapshot changes are intentional and are fine 👍🏻

awarrier99: Sure, let me take a look at this

awarrier99: @tw15egan this should be resolved now - I had changed the prop type to match what was in the `Dropdown` component originally, but I've now reverted that change

tw15egan (approved):

Issue #11770, opened 2022-07-08
ContextMenu: Bring to `stable` ☂️ 


tay1orjones:
Originally proposed in https://github.com/carbon-design-system/carbon/issues/6830, it's been over a year since the original submission and likely this is ready to move from `unstable` to `stable`

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/674
- [ ] https://github.com/carbon-design-system/carbon/issues/13514
```

tay1orjones:
@janhassel thanks for meeting today about this one! Capturing some notes below - feel free to add or reclarify any pieces:

> Needs storybook updates, remove generator config, replace with usage examples, figure out left nav hierarchy/naming

- [ ] Move OverflowV2 out from under "Menu" to be it's own root level item in the storybook left nav
- [ ] Add some basic stories for Menu showing off the Menu components in isolation
  - [ ] Add these basic stories to the e2e tests so they're covered by Percy/Visual regression tests
- [ ] Rename "ContextMenu" story folder to "useContextMenu"
- [ ] Add a .mdx docs page for Menu and include there a blurb about `useContextMenu` and how it can be a useful augmentation w/ Menu components. Provide a link to the useContextMenu default story.

janhassel:
@tay1orjones With the recent changes, would you mind re-assessing this checklist and the completion status?
Irt the two points "Add to Figma kit" and "Finish website docs" I will need some assistance from a Carbon designer to tackle this.

sstrubberg:
@janhassel I stub out the work. We'll have a designer on our team handle the website docs and figma kit. You'll just take care of the implementation :)

sstrubberg:
Need to talk to design and dev about the naming of this.

sstrubberg:
Documentation is being tracked in [Menu and Friends](https://github.com/carbon-design-system/carbon-website/issues/3563).

PR #13146, opened 2023-02-14
refactor(menu): use react.context and simplify logic

janhassel:
Contributes to #11770 

- Refactors `Menu` to use `React.Context` in order to avoid direct DOM manipulation like before
- Simplifies positioning logic
- Aligned component namings
- Removed storybook generator code
- Reorganized stories

#### Changelog

**Changed**

- Structure
  - Renamed `MenuGroup` to `MenuItemGroup`
  - Renamed `MenuDivider` to `MenuItemDivider`
  - Renamed `MenuRadioGroup` to `MenuItemRadioGroup`
  - Renamed `MenuSelectableItem` to `MenuItemSelectable`
  - Consolidated MenuItem components into single `MenuItem.js` file to avoid cyclic dependency
  - Removed default export in favor of named `Menu` export
  - Added `forwardRef` to all components
- `Menu`
  - Full refactor of `Menu` to use `React.Context` instead of direct DOM manipulations
  - Added support for `xs` size
- `MenuItemRadioGroup`
  - Added `props.itemToString` for custom functions to convert items to strings
  - Added option to control state via `props.selectedItem`
- `MenuItemSelectable`
  - Added option to control state via `props.selected`

**Removed**

- Storybook generators
- Direct DOM manipulation utilities

#### Testing / Reviewing

- Storybook
- Automatic tests


packages/react/src/components/Menu/MenuContext.js
```diff
@@ -0,0 +1,36 @@
...
+      return {
+        ...state,
+        items: [...state.items, action.payload],
+      };
+  }
+}
+
+const MenuContext = React.createContext({
+  state: menuDefaultState,
+});
```
francinelucca: would it also make sense to export an empty `dispatch` property here? took me a while to figure out where the dispatch function was coming from in `MenuItem`

packages/react/src/components/Menu/MenuItem.js
```diff
@@ -1,53 +1,433 @@
...
+    } else {
+      context.state.requestCloseRoot();
+
+      if (onClick) {
+        onClick(e);
+      }
+    }
+  }
+
+  function handleMousEnter() {
```
francinelucca:
```suggestion
  function handleMouseEnter() {
```

packages/react/src/components/Menu/MenuItem.js
```diff
@@ -1,53 +1,433 @@
...
-      {children}
-    </MenuOption>
+      ref={ref}
+      className={classNames}
+      tabIndex="-1"
+      aria-disabled={disabled}
+      aria-haspopup={hasChildren || null}
+      aria-expanded={hasChildren ? submenuOpen : null}
+      onClick={handleClick}
+      onMouseEnter={hasChildren ? handleMousEnter : null}
```
francinelucca:
```suggestion
      onMouseEnter={hasChildren ? handleMouseEnter : null}
```

packages/react/src/components/Menu/MenuItem.js
```diff
@@ -1,53 +1,433 @@
...
+          open={submenuOpen}
+          onClose={() => {
+            closeSubmenu();
+            menuItem.current.focus();
+          }}
+          x={boundaries.x}
+          y={boundaries.y}>
+          {children}
+        </Menu>
+      )}
```
francinelucca:
```suggestion
 {hasChildren && (
        <>
          <div className={`${prefix}--menu-item__shortcut`}>
            <CaretRight />
          </div>
          <Menu
            label={label}
            open={submenuOpen}
            onClose={() => {
              closeSubmenu();
              menuItem.current.focus();
            } }
            x={boundaries.x}
            y={boundaries.y}>
              {children}
            </Menu>
          </>
      )}
```

packages/react/src/components/Menu/MenuItem.js
```diff
@@ -1,53 +1,433 @@
...
+   */
+  renderIcon: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
+
+  /**
+   * Provide a shortcut for the action of this MenuItem. Note that the component will only render it as a hint but not actually register the shortcut.
+   */
+  shortcut: PropTypes.string,
+};
+
+const MenuItemSelectable = React.forwardRef(function MenuItemDivider(
```
francinelucca:
```suggestion
const MenuItemSelectable = React.forwardRef(function MenuItemSelectable(
```

francinelucca (changes requested):
Mostly LGTM, some cleanup suggestions.

Two more things:
1. I'm experiencing some errors when opening a menu for the second time and trying to use keyboard navigation (see console. I am arrowing down in this example):

https://user-images.githubusercontent.com/40550942/219105336-dbfa1a4b-4916-423e-bfff-ec56d37b25f3.mov


2. Should bullet points #2.5 & #4 be addressed in this PR or is that outside of the scope?
https://github.com/carbon-design-system/carbon/issues/11770#issuecomment-1315488039
4. image

packages/react/src/components/Menu/MenuContext.js
janhassel: I used the "missing" `dispatch` to identify whether or not a menu was the root menu, but I can see how it's a bit opaque. I just pushed an updated with a dummy `dispatch` in the default state and a different method of identifying the root menu. Let me know what you think!

janhassel:
@francinelucca
1. good catch! should be fixed now
2. yes, I'd rather separate the implementation work from the documentation work in order to prevent PRs getting to big

tay1orjones (approved):
LGTM!

packages/react/src/components/Menu/MenuContext.js
francinelucca: nice! thanks!!

francinelucca:
I'm still noticing some unexpected keyboard behaviors:

- Based on https://react.carbondesignsystem.com/?path=/story/experimental-unstable-menu-overflowmenuv2--custom-icon looks like the menu should close on `tab` or `Esc`
- Still seeing console errors when tabbing off the menu, coming back , closing it and reopening via `Enter` and trying to `ArrowDown`, see attached

https://user-images.githubusercontent.com/40550942/219411197-71be8c4d-fe18-4668-ae3e-f68e6730b697.mov

janhassel:
@francinelucca Thanks for catching that! It should be fixed now. Let me know if you find anything else!

packages/react/src/components/OverflowMenuV2/OverflowMenuv2.stories.js
```diff
@@ -6,68 +6,63 @@
...
-  component: Menu,
+  title: 'Experimental/unstable__OverflowMenuV2',
+  component: OverflowMenuV2,
 };
 
-const Story = (items, props = {}) => (
-  <StoryFrame>
-    <OverflowMenuV2 {...props}>{buildMenu(items)}</OverflowMenuV2>
-  </StoryFrame>
-);
```
francinelucca, line 25: I think we still need this so that the Experimental notification shows up on the storybook, we should add it to the new Menu story as well, see https://react.carbondesignsystem.com/?path=/story/experimental-unstable-menu-overflowmenuv2--overflow-menu-v-2

francinelucca (changes requested):
Behavior LGTM now 🔥. Just one more thing I found on the story 🙏🏻

packages/react/src/components/OverflowMenuV2/OverflowMenuv2.stories.js
janhassel: The banner stems from a time when there were almost no experimental components and I think I added it to follow the `unstable_Tree` back then which had a notification stating the component is till undergoing an a11y review before stable.

Today, I checked the other experimental components in the storybook and none had a banner or similar. Therefore I thought I'd align with the rest and remove it. Let me know if you'd rather have it added again!

francinelucca: yeah we can remove for consistency, thanks for clarifying!

francinelucca (approved):

PR #13150, opened 2023-02-14
test(table-select-row): migrate TableSelectRow test to RTL

### LINKED ISSUE #12670: Use RTL for testing in `DataTable/__tests__/TableSelectRow-test.js`

francinelucca:
Closes #12670 

#### Changelog

**Changed**

- packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js: migrate to RTL
- packages/react/src/components/DataTable/__tests__/__snapshots__/TableSelectRow-test.js.snap: update snapshot

#### Testing / Reviewing

CIs should still pass


andreancardona (approved):
😎

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
```diff
@@ -23,45 +23,49 @@ describe('DataTable.TableSelectRow', () => {
...
-          <TableRow>
-            <TableSelectRow {...mockProps} className={customClassName} />
-          </TableRow>
-        </TableHead>
-      </Table>
-    );
-    const elements = wrapper.find(`td.${customClassName}`);
-    expect(elements.length).toBe(1);
+    it('should render with the provided class name', () => {
+      const customClassName = 'custom-table-select-row-classname';
```
tay1orjones: It looks like there's already a custom classname provided through `mockProps`.

Honestly I'm an advocate for removing abstractions like `mockProps` in favor of explicitly placing the props on the component in render. To me it feels like a hasty abstraction that reduces readability and doesn't optimize for change.

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
```diff
@@ -23,45 +23,49 @@ describe('DataTable.TableSelectRow', () => {
...
+        <Table>
+          <TableHead>
+            <TableRow>
+              <TableSelectRow {...mockProps} className={customClassName} />
+            </TableRow>
+          </TableHead>
+        </Table>
+      );
+      const elements = container.querySelectorAll(`td.${customClassName}`);
+      expect(elements.length).toBe(1);
```
tay1orjones: Two things:

1. Instead of using querySelector, use `screen` coupled with a [query](https://testing-library.com/docs/queries/about#using-queries).
2. All the [jest-dom matchers](https://github.com/testing-library/jest-dom#tohaveclass) are available and can make things more readable

There's [a recommended priority](https://testing-library.com/docs/queries/about#priority) for the queries to skew towards using more accessible ones. I think this is one of the few cases where `getByTestId` actually makes sense, despite it being at the bottom of that list.

This test could look something like

```
      render(
        <Table>
          <TableHead>
            <TableRow>
              <TableSelectRow {...mockProps} data-testid="test-row" />
            </TableRow>
          </TableHead>
        </Table>
      );
      expect(screen.getByTestId('test-row').firstChild).toHaveClass(className);
```

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
```diff
@@ -23,45 +23,49 @@ describe('DataTable.TableSelectRow', () => {
...
+      const { container } = render(
+        <Table>
+          <TableHead>
+            <TableRow>
+              <TableSelectRow {...mockProps} />
+            </TableRow>
+          </TableHead>
+        </Table>
+      );
+      container.querySelector('input.cds--checkbox').click();
```
tay1orjones: Same thing here with using a query instead of querySelector. I think `screen.getByRole('checkbox')` would give you the same element to click. Also events should be fired using `user-event`.

`userEvent.click(screen.getByRole('checkbox'));`

tay1orjones (commented):
This looks pretty solid! 💪  I have some thoughts below on some improvements that could be made. Did you find [the script](https://github.com/carbon-design-system/carbon/issues/10184#issuecomment-1282894448) helpful in stubbing out the file?

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
francinelucca: Now using the custom className frpm `mockProps`. I didn't remove 'cause most of the props are required so I would've just duplicated the values in every test. But I'm happy to do that if you feel that makes sense!

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
francinelucca: done!

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
francinelucca: done! thanks so much for the RTL references 🙏🏻

francinelucca:
@tay1orjones you caught me! I forgot to run the script 😳, added all the tests that came from that an implemented those changes you requested :)

packages/react/src/components/DataTable/__tests__/TableSelectRow-test.js
francinelucca: I ended up going with `expect(screen.getByRole('row').firstChild).toHaveClass(mockProps.className);` btw because `TableSelectRow` is not spreading props so it's not passing off the datatestId prop

tay1orjones (approved):
@francinelucca oh no worries! Just was genuinely interested if it was or was not helpful. This looks great!

Issue #12696, opened 2022-11-18
[Bug]: Data Table - Sorting and null


JOwens818:
### Package

@carbon/react

### Browser

Firefox

### Package version

1.15.0

### React version

17.0.2

### Description

DataTable component with 'isSortable' = true...  Sorting works as expected as long as there are no null values in a given column.  However if there are null values present in the column an error gets thrown in the console and sort functionality does not do anything.  See attached for error and reproduction example.  In the Stackblitz example I tried with both 1.15.0 and 1.17.0.  

![dataTableError](https://user-images.githubusercontent.com/36175299/202724456-bbefa168-bb84-44f2-a8ce-5727ad3d3d0b.png)

### Suggested Severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Reproduction/example

https://stackblitz.com/edit/github-9h9nvq?file=src/App.jsx

### Steps to reproduce

Run the app and try to sort on a column that has null values (DEPTNAME).  You will notice that sorting on DEPTCODE still works.

Interestingly enough, if I remove the TableSelectRow functionality the sort appears to work again on DEPTNAME...

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tw15egan:
I see the bug as well. As a workaround, it seems to work if you pass in an empty string: https://stackblitz.com/edit/github-9h9nvq-lvtpgo?file=src%2FApp.jsx

jsehull:
Hey team - I've been working on a fix and would like to be assigned to this issue

tw15egan:
@jsehull excellent, we would welcome any PR to fix this. Thanks for taking this on

jsehull:
@tw15egan my PR is mentioned above this comment, but do I need to manually link the PR to the Issue or the Issue to the PR under Development (right sidebar)?

PR #13155, opened 2023-02-15
fix(DataTable): add sorting null values

jsehull:
Closes #

Fix for Issue [#12696](https://github.com/carbon-design-system/carbon/issues/12696)

#### Changelog

**New**

- fix ability to sort data even if it has null values in rows

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

My basic findings BEFORE fix
- A: single null value in one column did NOT break sorting
- B: multiple null values in one column DID break sorting

1. open `packages/react/src/components/DataTable/stories/shared.js`
2. under `const rows` set various values to `null` (i.e. multiple `name` keys set to null for column sorting)
3. open the Storybook site to test in browser: `Storybook > DataTable > Sorting > Default` (or anywhere that has sorting enabled)
4. test the sort feature: hover over column headers and cycle clicks at least **3 times** (3 different sorting comparison types), but also a bunch of times 💪 
5. confirm both A and B findings (above) sort as expected now

### DIFF
--- a/packages/react/src/components/DataTable/tools/sorting.js
+++ b/packages/react/src/components/DataTable/tools/sorting.js
@@ -20,6 +20,10 @@ import { sortStates } from '../state/sortStates';
  * @returns {number}
  */
 export const compare = (a, b, locale = 'en') => {
+  // prevent multiple null values in one column (sorting breaks)
+  a === null ? (a = '') : null;
+  b === null ? (b = '') : null;
+
   if (typeof a === 'number' && typeof b === 'number') {
     return a - b;
   }


jsehull:

francinelucca:
Hey @jsehull 👋🏼
sorry, I think your best bet here will be to close out and do a new PR with the correct git configurations in place, it's the same thing as last time like you said.

jsehull:
NEW PR created to address DCO error: https://github.com/carbon-design-system/carbon/pull/13161

tw15egan:
Closing in favor of #13161

PR #13162, opened 2023-02-15
refactor(search): refactor tests to use RTL instead of enzyme

### LINKED ISSUE #12153: Use RTL for testing in `Search/Search-test.js`

tay1orjones:
Closes #12153

#### Changelog

**Changed**

- Refactor `Search` and `ExpandableSearch` tests to use React Testing Library
- Fix a bug in Search - the `renderIcon` prop was not working

#### Testing / Reviewing

- Tests should pass


packages/react/src/components/Search/Search-test.js
```diff
@@ -1,333 +1,157 @@
...
-        });
+    it('should respect labelText prop', () => {
+      render(<Search labelText="test-search" />);
 
-        it('is called', () => {
-          expect(onChange).toHaveBeenCalledWith(target.value);
-        });
-      });
+      expect(screen.getByRole('searchbox')).toBeInTheDocument();
+    });
```
francinelucca: is testing that 'searchbox' is in the document sufficient to test for the labelText? 😅

francinelucca (changes requested):

packages/react/src/components/Search/Search-test.js
tay1orjones: Zoinks! Great catch. I just updated it with a more thorough test

francinelucca (approved):
🔥

tw15egan (approved):
LGTM 👍 ✅

Issue #11095, opened 2022-03-29
[a11y]: Element "button" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'


xiepingp:
### Package

carbon-components-react

### Browser

Chrome, Firefox

### Operating System

Windows, MacOS

### Package version

v10.39.0

### React version

v7.56.0

### Automated testing tool and ruleset

IBM Equal Access Accessibility Checker - Latest Deployment

### Description

When scanning the datatable with batch actions using Accessibility Checker, below violation is returned: Element "button" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'.
Same violations are happening when batch action bars is displayed (checkbox is selected).

### WCAG 2.1 Violation

IBM 1.3.1 Info and Relationships & IBM 4.1.2 Name, Role, Value

### CodeSandbox example

(not sure)

### Steps to reproduce

Scan the page using Accessibility Checker: https://react.carbondesignsystem.com/?path=/story/components-datatable-batch-actions--default

![Screen Shot 2022-03-28 at 3 41 04 PM](https://user-images.githubusercontent.com/63960906/160520195-6612c6a7-1cff-45c7-9d9d-51813cf6d70c.png)
![Screen Shot 2022-03-28 at 3 41 34 PM](https://user-images.githubusercontent.com/63960906/160520205-01982a52-b4fc-40b1-ac07-5ee0c3d7f69d.png)
![Screen Shot 2022-03-28 at 3 41 41 PM](https://user-images.githubusercontent.com/63960906/160520209-d0489ee8-b1f2-4e58-a4c1-7b1bbe804902.png)
![Screen Shot 2022-03-28 at 3 41 46 PM](https://user-images.githubusercontent.com/63960906/160520213-edf23f94-e872-467b-8701-9e4805ef0cf9.png)

Select the checkbox and the batch actions bar is displayed:
![Screen Shot 2022-03-29 at 3 25 00 PM](https://user-images.githubusercontent.com/63960906/160557392-2c8a4527-78d4-4a41-af17-f22ae24ff00e.png)
![Screen Shot 2022-03-29 at 3 25 47 PM](https://user-images.githubusercontent.com/63960906/160557405-0e3334e1-abda-44d7-b9e9-1948a23410b8.png)
[...]

Issue #6423, opened 2020-07-09
Remove `tabIndex` from default props of `Button`


vladbalanescu:
## What package(s) are you using?
Carbon-components-react@^7.1.0
Carbon-components@^10.11.1

## Detailed description

> Describe in detail the issue you're having.
[This line](https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/Button/Button.js#L164) needs to be taken out so that if we have a use case in which we don't want to have the button focusable, we can define `tabIndex={undefined}` and the user won't be able to select that button via tabbing through

> Is this issue related to a specific component?
`Button`

> What browser are you working in?
Chrome, Firefox, Safari

## Additional information
Button is focusable even when the panel is shut:
![image](https://user-images.githubusercontent.com/16047402/87031621-4baf4e00-c1db-11ea-9e5e-b10c8a78fa7d.png)

emyarod:
rather than `undefined` does setting it to a negative value or `null` work for you?

vladbalanescu:
@emyarod if you mean on the consumer side, than the answer is no, if I set it to any of those values this component would use the default prop supplied, since the line above needs taking out for us to be able to have the ability to not focus it.

emyarod:
here is the behavior when I use a negative tabindex https://codesandbox.io/s/frosty-bose-3yehv

can you create a reproducible example of negative tabindex being overwritten by the default prop value?

vladbalanescu:
Works brilliant with a negative index, thank you

PR #13171, opened 2023-02-16
fix(Button): remove default tabIndex

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/11095
Refs https://github.com/carbon-design-system/carbon/issues/6423

Removes the default `tabIndex` value from `Button`, as it was causing a11y violations when buttons were placed inside of an element with `aria-hidden="true"`

#### Changelog

**Removed**

- Removed the default `tabIndex` value

#### Testing / Reviewing

Ensure all Buttons still retain proper focus and tab order. `Button`, `DataTable` batch actions, etc..


tay1orjones:
I think this makes sense. In particular, after looking at [the MDN topic](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) on this, we probably shouldn't be adding 0 as a default:

> Some focusable HTML elements have a default tabindex value of 0 set under the hood by the [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent). These elements are an [`<a>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [`<area>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area) with href attribute, [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), ... Developers shouldn't add the `tabindex` attribute to these elements unless it changes the default behavior (for example, including a negative value will remove the element from the focus navigation order).

tay1orjones:
IBMa [also thinks](https://ibm-studios.slack.com/archives/G01GCBCGTPV/p1676569885403589?thread_ts=1676565828.500929&cid=G01GCBCGTPV) this is the right direction to take.

I'm going to mark this as ready for review.

tay1orjones (approved):
The following areas look correct to me on the deploy preview:
- [x] DataTable/Batch actions
- [x] ButtonSet
- [x] ActionableNotification
- [x] Toggletip
- [x] UIShell
- [x] Pagination

andreancardona (approved):
tested and all looks good also ran the checker!

PR #13174, opened 2023-02-16
refactor: convert Slider to TypeScript

### LINKED ISSUE #12557: Add TypeScript types to Slider

sierrawetmore:
Closes #12557 

This PR converts Slider.js to TypeScript.

#### Changelog

**Changed**

- Slider.js has been converted to TS

#### Testing / Reviewing

These changes can be tested and verified by running a fresh yarn, yarn build and yarn storybook. Verify that the behavior in the Slider story remains unchanged


packages/react/src/components/Slider/Slider.tsx
```diff
@@ -624,6 +779,7 @@ export default class Slider extends PureComponent {
                 <span className={`${prefix}--slider__range-label`}>
                   {formatLabel(min, minLabel)}
                 </span>
+                {/* @ts-ignore onBlur + onChange types are incompatible*/}
```
francinelucca, line 784: is it correct that the `onBlur` and `onChange` functions are passed onto the top `div` with `{...other}` in line 794 @tay1orjones? seems like  we're calling those prop functions explicitly in our`onBlur` and  `componentDidUpdate` functions and the types do seem to be incompatible 🤔. should these props be deleted from `other` like we're doing with `onRelease` and `invalid` on line 731/732?

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -563,7 +718,7 @@ export default class Slider extends PureComponent {
       formatLabel = defaultFormatLabel,
       labelText,
       step,
-      stepMultiplier, // eslint-disable-line no-unused-vars
+      stepMultiplier: _stepMultiplier,
```
francinelucca, line 723: what's the purpose of destructuring this prop here @tay1orjones ? doesn't seem like we're using it for rendering purposes, should we `delete other.stepMultiplier` as well in lines 731-732

francinelucca (commented):
Mostly LGTM, just some implementation questions for the team

packages/react/src/components/Slider/Slider.tsx
tay1orjones: This prop is destructured off `this.props` so that it is not included in `...other`. `other` is spread onto a `div` and would error if this property was applied to it.

This is the only file in which the `delete` pattern is used, and it's quite old code (4+ years). By and large in the rest of the codebase we follow the pattern of destructuring out any props we don't want included in `other`, and turn off the unused variable eslint rule.

What's the purpose of adding this rename instead of retaining the `eslint-disable-line `?

packages/react/src/components/Slider/Slider.tsx
tay1orjones: Yep looks like those two are passed onto the `div` via `other`.

It's possible someone is relying on those for some type of functionality. I'd say it's out of scope of this PR to remove them without investigating how `onBlur` and `onChange` functions currently. It might be being called twice or something.

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -484,11 +637,12 @@ export default class Slider extends PureComponent {
...
    */
-  calcValue = ({ clientX = null, value = null, useRawValue = false }) => {
+
+  calcValue = ({ clientX, value, useRawValue = false }: CalcValueProps) => {
     const range = this.props.max - this.props.min;
-    const boundingRect = this.element.getBoundingClientRect();
-    const totalSteps = range / this.props.step;
-    let width = boundingRect.right - boundingRect.left;
+    const boundingRect = this.element?.getBoundingClientRect();
+    const totalSteps = range / (this.props.step ?? 1);
```
tay1orjones: I don't think this is necessary. `step` already has a default of `1` set

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -484,11 +637,12 @@ export default class Slider extends PureComponent {
...
    * @param {boolean} [params.useRawValue=false] `true` to use the given value as-is.
    */
-  calcValue = ({ clientX = null, value = null, useRawValue = false }) => {
+
+  calcValue = ({ clientX, value, useRawValue = false }: CalcValueProps) => {
     const range = this.props.max - this.props.min;
-    const boundingRect = this.element.getBoundingClientRect();
-    const totalSteps = range / this.props.step;
-    let width = boundingRect.right - boundingRect.left;
+    const boundingRect = this.element?.getBoundingClientRect();
```
tay1orjones: This will ensure `this.element` is not null, but should we also protect for if the method isn't found on `this.element`?

```suggestion
    const boundingRect = this.element?.getBoundingClientRect?.();
```

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -393,16 +544,18 @@ export default class Slider extends PureComponent {
     // If shift was held, account for the stepMultiplier
     if (evt.shiftKey) {
       const stepMultiplier = this.props.stepMultiplier;
-      delta *= stepMultiplier;
+      delta *= stepMultiplier ?? 1;
```
tay1orjones: This shouldn't be needed. `stepMultiplier` has a default of `4` set already. Using `1` if `stepMultiplier` is null seems unnecessary and could cause unwanted behavior.

tay1orjones (changes requested):
I have a few questions below.

Could you help clarify why numerical props in the cases below need a nullish coalescing operator when they're already provided default values?

packages/react/src/components/Slider/Slider.tsx
sierrawetmore: We have `@typescript-eslint/no-unused-vars` enabled, so unused vars must match /^_/u

packages/react/src/components/Slider/Slider.tsx
sierrawetmore: You're right, I had forgotten about the default values.

packages/react/src/components/Slider/Slider.tsx
sierrawetmore: Yep, forgot about the defaults.

packages/react/src/components/Slider/Slider.tsx
sierrawetmore: Sounds good to me

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -517,8 +673,9 @@ export default class Slider extends PureComponent {
       };
     }
 
-    let steppedValue = Math.round(leftPercent * totalSteps) * this.props.step;
-    let steppedPercent = this.clamp(steppedValue / range, 0, 1);
+    let steppedValue =
+      Math.round(leftPercent * totalSteps) * (this.props.step ?? 1);
```
francinelucca:
```suggestion
      Math.round(leftPercent * totalSteps) * (this.props.step ?? Slider.defaultProps.step);
```

francinelucca (changes requested):

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -379,12 +531,11 @@ export default class Slider extends PureComponent {
...
     }
 
-    const which = Number.parseInt(evt.which);
     let delta = 0;
-    if (matches(which, [keys.ArrowDown, keys.ArrowLeft])) {
-      delta = -this.props.step;
-    } else if (matches(which, [keys.ArrowUp, keys.ArrowRight])) {
-      delta = this.props.step;
+    if (matches(evt.which, [keys.ArrowDown, keys.ArrowLeft])) {
+      delta = -(this.props.step ?? Slider.defaultProps.step);
```
tay1orjones, line 536: @sierrawetmore Sorry if I'm being dense here on these default props - why is this needed?

What error is given if this is reverted to it's original version without the nullish coalescing operator providing the default as a fallback?

sierrawetmore: No worries, in this case it gives the following error: `let delta: number
Type 'number | undefined' is not assignable to type 'number'.
  Type 'undefined' is not assignable to type 'number'.ts(2322)`

tay1orjones: Ah okay, yeah @francinelucca and I just looked at it as well. I think it's ultimately due to this being a class component and the props are not being destructured, leading to the compiler believing `step` is potentially undefined because it isn't aware that React will supply the default value.

tay1orjones (approved):

francinelucca (approved):

PR #13206, opened 2023-02-22
feat(TimePicker): add typescript typings

### LINKED ISSUE #12563: Add TypeScript types to TimePicker, TimePickerSelect

GalvinGao:
Closes #12563

Adds TypeScript typings to `TimePicker` and `TimePickerSelect`.

#### Changelog

**New**

- Adds TypeScript typings to `TimePicker` and `TimePickerSelect`.

**Changed**

`N/A`

**Removed**

`N/A`

#### Testing / Reviewing

By using the storybook, the maintainer may navigate to `/?path=/story/components-timepicker--default` on their local storybook instance to see that the component successfully renders with the provided stories. A screenshot is attached for reference.

image


packages/react/src/components/TimePickerSelect/TimePickerSelect.tsx
```diff
@@ -0,0 +1,116 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
```
francinelucca:
```suggestion
 * Copyright IBM Corp. 2016, 2023
```

francinelucca (changes requested):
Hey @GalvinGao 👋🏼
 Thanks for getting this PR up!

Mostly LGTM, looks like `TimePickerSelect` changes are being picked up as a file deletion/addition, @GalvinGao did you use the `gh mv` script as per described in the [Steps to provide baseline type definitions for components ](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md#steps-to-provide-baseline-type-definitions-for-components)? if not, could you? 🙏🏻

francinelucca (approved):
LGTM

packages/react/src/components/TimePicker/TimePicker.tsx
```diff
@@ -124,8 +242,13 @@ const TimePicker = React.forwardRef(function TimePicker(
...
     const mappedChildren = React.Children.map(children, (pickerSelect) => {
-      return React.cloneElement(pickerSelect, {
-        ...pickerSelect.props,
+      const item = pickerSelect as React.ReactElement<TimePickerSelectProps>;
+      if (item.type !== TimePickerSelect) {
+        return pickerSelect;
+      }
+
+      return React.cloneElement(item, {
+        ...item.props,
```
tay1orjones: I don't think we can do this without a breaking change. Right now it is not a requirement that a child be a `TimePickerSelect`. Consumers can use their own version of a TimePickerSelect-like component that may differ in signature from our `TimePickerSelect`.

tay1orjones (changes requested):

packages/react/src/components/TimePicker/TimePicker.tsx
GalvinGao: I see... then the only way I think is to make item `any` to let TS happy about the props we used in the second argument of `React.cloneElement`. I've tried to do a global search on the usage of `React.cloneElement` and see if there's existing solutions for this but it seems there's no existing ones already migrated to TS. Any thoughts?

GalvinGao: Used `as any` instead of type assertion in 1309904711bf91de11e7b3d1a0a6e42a2baac2b6

tay1orjones (approved):
Looks good, thanks for making that change 👍

Issue #12971, opened 2023-01-12
Add ability to hide dev-only stories from the production storybook


tay1orjones:
In some instances, it's helpful to add a story to the storybook for development and testing purposes. These stories are not helpful to the consumer though, and would promote confusion if included in the production site/build.

For these cases, it would be nice if there was a way to add these stories, but have them in a special section that is not included in the production build.

https://github.com/carbon-design-system/carbon/issues/12971#tasklist-block-ce3faa63-639a-433c-98d1-7e9479ec4a5f

<details>
  <summary><h3>A few potential approaches</h3></summary>
  
  ### Use a [environment flag](https://storybook.js.org/docs/react/configure/environment-variables) to turn on/off dev mode
  - This dev mode would need to be applied in the netlify deploy preview environment. It is currently considered a 'production' environment, but we would want to show the dev stories for testing.
  - Ideally our CI and e2e tests would run against all these stories as well. This could be a good way to increase our visual snapshot coverage without introducing a bunch of stories that are only there for snapshotting purposes (e.g. input variants - readonly, disabled, etc.)
  
#### Author stories in `*.dev.stories.js` files, and exclude these from the build in `.storybook/main.js`
  - Benefits: clear separation from production stories
  - Cons: does not live alongside other stories, more difficult to copy paste and maintain as component APIs change

#### Author stories in the same file as production stories, but add a flag/prefix to their names, e.g. `DEV_ReadOnlyInputExamples`
  - We could use this prefix to exclude these stories from the build in via the [`excludeStories` option](https://storybook.js.org/docs/react/api/csf#non-story-exports)
  - https://github.com/storybookjs/storybook/discussions/15134
  - Maybe something like `excludeStories: process.env.DEV_MODE ? /.*DEV$/ : [],` in each story file
  - Benefits: dev stories live alongside production ones, maintenance is easier
  - Cons: less clear separation, but would be easy to find/identify dev stories with an all-caps `DEV_` (or similar) prefix
  
I think ultimately the best approach right now could be to add the environment flag, and use option B
</details>

### Concerns
- This could become a wasteland dumping ground of unhelpful stories that are only used once for bug-repo/resovle.
  - To address, could we define some parameters around why/when we add a dev-only story vs not?
- Visually snapshotting production and dev-only stories could quickly run up our snapshot count, exceeding the quota.
  - To address, could we provide a way for stories to be exempt from snapshotting to avoid running up our snapshot count?

[...]

PR #13207, opened 2023-02-22
test(menu): add vrt and avt

janhassel:
Contributes to #11770

Adds VRT and AVT for `Menu` component

#### Changelog

**New**

- Added `Menu-test.e2e.js` to enable VRT and AVT

**Changed**

- Provided default `false` to `MenuItemSelectable` `selected` state to avoid `aria-checked` not being populated when neither `defaultSelected` nor `selected` is passed

---

@tay1orjones Is there a way to create stories that are not indexed by storybook but can be picked up by percy? That way we could have separate tests for each of the subcomponents (`MenuItem*`). Or would you say covering the Playground story is sufficient?

### DIFF
--- a/e2e/components/Menu/Menu-test.e2e.js
+++ b/e2e/components/Menu/Menu-test.e2e.js
@@ -1,0 +1,37 @@
+/**
+ * Copyright IBM Corp. 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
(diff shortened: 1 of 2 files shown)


alisonjoseph (approved):
LGTM!

tay1orjones (approved):
> Is there a way to create stories that are not indexed by storybook but can be picked up by percy?

@janhassel No, not yet - #12971. For now we're only snapshotting existing stories showing a single state. Using just the playground story for now is great 👍

PR #13209, opened 2023-02-22
docs: add docs for menu and useContextMenu

janhassel:
Contributes to #11770 

#### Changelog

**New**

- Added documentation for both `Menu` (and its subcomponents) and the `useContextMenu` hook

#### Testing / Reviewing

Storybook

---

Note: these docs assume the changes submitted in #13210.

PR #13210, opened 2023-02-22
feat(useContextMenu): add ref support

janhassel:
When working on #13209 I noticed that the `useContextMenu` currently only supports DOM elements (or `window` or `document`) as its trigger which means developers would need to wrap `useContextMenu` in a `useEffect` in order to pass it the correct `ref.current` at all times.

Since this feels a bit cumbersome and the `useContextMenu` hook already uses a `useEffect` with `trigger` as its dependency, I updated the hook to directly support react refs.

#### Changelog

**New**

- Added support for react refs as `trigger` argument in `useContextMenu`

#### Testing / Reviewing

```jsx
function TestComponent() {
  const el = useRef(null);
  const menuProps = useContextMenu(el);

  return (
    <>
      <div ref={el}>
        Right click this element
      </div>
      <Menu {...menuProps}>
        <MenuItem label="Edit" />
        <MenuItem label="Delete" kind="danger" />
      </Menu>
    </>
  );
};
```

Menu should appear when right-clicking on the `div`, but not when right-clicking anywhere else on the screen.


francinelucca (approved):
lgtm

tay1orjones (approved):

PR #13223, opened 2023-02-24
fix(menuitem): prevent disabled and danger styles where it doesn't make sense

janhassel:
Fixes some illogical uses of `disabled` and `danger` for `MenuItem`.

#### Testing / Reviewing

- A `MenuItem` that is disabled should not emit `onClick`
- A `MenuItemSelectable` that is disabled should not change on click
- A `MenuItem` that is disabled and danger should not receive a red background hover
- A `MenuItem` with children (submenu) can never be disabled
- A `MenuItem` with children (submenu) can never be danger


tay1orjones (approved):
Nice! These make a ton of sense 👍

andreancardona (approved):
looks good!

PR #13235, opened 2023-02-26
Add Typescript annotations to Select, SelectItem and SelectItemGroup

### LINKED ISSUE #12555: Add TypeScript types to Select, SelectItem, SelectItemGroup

jpsorensen:
Closes #12555 

Convert existing propType definitions to Typescript annotations on Select, SelectItem and SelectItemGroup.

#### Changelog

**Changed**

- Select.js, SelectItem.js and SelectItemGroup.js converted to tsx
- Removed label attribute from unit test

#### Testing / Reviewing

- In the packages/react directory, run yarn build, yarn storybook. In the storybook, navigate to Select and confirm that the component's behavior remains unchanged.
- Find usages of Select (along with SelectItem and SelectItemGroup) throughout the code base. Add them to the bottom of Select.tsx and confirm that they compile. (In this case, one did not and I eventually ended up changing the test - removing label. See notes below).


packages/react/src/components/Select/__tests__/Select-test.js
```diff
@@ -174,7 +174,6 @@ describe('Select', () => {
       render(
         <Select
           id="select-1"
-          label="Select label"
```
jpsorensen, line 177: This does not pass compilation, as there's no label attribute on Select.

Spent some time digging around and eventually concluded it should not be there, but happy to change this if someone else knows something I missed.

There's no explicitly defined label attribute. It does pass its ...other attributes on to the <select> element, but according to my searching (both looking at the Typescript definition of select and looking at docs for <select> online, I don't think it has a label attribute - it can be associated with a <label> element, but as far as I could tell setting a label= attribute is not a shorthand for an actual label element.

andreancardona: I'll defer to @tay1orjones on this one 👀

tw15egan: I believe it should be using the `labelText` prop, not `label`

jpsorensen: Great, thanks for weighing in! Updated to use `labelText` instead of `label`.

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+   */
+  invalid?: boolean;
+
+  /**
+   * Message which is displayed if the value is invalid.
+   */
+  invalidText?: ReactNode;
+
+  /**
+   * Provide label text to be read by screen readers when interacting with the
```
andreancardona:
```suggestion
   * Provide label text to be read by screen readers when interacting with the control.
```

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+   * control
```
andreancardona:
```suggestion
```

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+   */
+  labelText?: ReactNode;
+
+  /**
+   * `true` to use the light version. For use on $ui-01 backgrounds only.
+   * Don't use this to make tile background color same as container background color.
+   *
+   * @deprecated The `light` prop for `Select` is no longer needed and has
```
andreancardona:
```suggestion
   * @deprecated The `light` prop for `Select` is no longer needed and has been deprecated in v11 in favor of the new `Layer` component.
```

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+   *    been deprecated in v11 in favor of the new `Layer` component. It
```
andreancardona:
```suggestion
   * It will be moved in the next major release.
```

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+   *    will be moved in the next major release.
```
andreancardona:
```suggestion
```

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+  light?: boolean;
+
+  /**
+   * Reserved for use with <Pagination> component. Will not render a label for the
+   * select since Pagination renders one for us.
+   */
+  noLabel?: boolean;
+
+  /**
+   * Provide an optional `onChange` hook that is called each time the value of
```
andreancardona:
```suggestion
   * Provide an optional `onChange` hook that is called each time the value of the underlying `<input>` changes.
```

packages/react/src/components/Select/Select.tsx
```diff
@@ -18,6 +25,109 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+   * the underlying `<input>` changes
```
andreancardona:
```suggestion
```

andreancardona:
@jpsorensen just cleaning up a few of the white spaces in the comments. Looks all good to me though - thank you so much for the contribution 🎉

andreancardona (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #13236, opened 2023-02-27
feat(DataTable): add rowIds to the sortRow extra params list

### LINKED ISSUE #13208: [Feature Request]: DataTable sortRow - pass rowIds to sortRow
### The problem

Occasionally, the sorting of a column depends on more than just the cell display value. Other row/state data may be needed to perform the sort.

The product I work on supports the notion of "acceptable" status values. This allows customers to ignore anticipated hardware/software errors during maintainence windows. To indicate this, our user interface displays an "acceptable" icon in status cells - a red 'x' indicates the status is unacceptable and a green checkmark indicates the status is acceptable.

image

We historically sort our status columns by acceptability and then by status text. However, the existing carbon DataTable `sortRow` function does not provide the information needed to do this. We need `sortRow` to also provide the IDs of the rows being sorted so we can check row/object acceptability.

### The solution

Add the rowIds (`[a, b]`) to the object param passed to `sortRow` functions - https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/tools/sorting.js#L97

### Examples

````
  rowIds.slice().sort((a, b) => {
    const cellA = cellsById[getCellId(a, key)];
    const cellB = cellsById[getCellId(b, key)];
    return sortRow(cellA && cellA.value, cellB && cellB.value, {
      key,
      sortDirection,
      locale,
      sortStates,
      compare,
      **rowIds: [a, b]**
    });
  });
````
- https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/tools/sorting.js

### Application/PAL

zSystems Hardware Management Console

### Business priority

### Available extra resources

_No response_

### Code of Conduct

cknabe:
Closes #13208

The request is to return the rowIds in the custom sortRow hook for DataTable. 
I looked for any place to update documentation and did not see any.
I looked at the existing tests, and am not sure it is warranted or needed to add a test that would cover this addition.
I do not see a possibility for side effects to the component with this addition.
I do not see that the props for the sortRow hook are documented for the user, so this addition would not be visible without inspecting the code.

#### Changelog

- Added rowIds as a property on the extra params list for the sortRow hook

#### Testing / Reviewing

- Setup a DataTable with a custom sortRow prop that prints out the extra params returned in the hook. You will see the current id values that are set on the rows that are currently being sorted by sortRow, in the rowIds field.


cknabe:
I was looking for an issue to pick up, and my team as well has dealt with needing this over the last few years.

tw15egan (approved):
LGTM 👍 ✅

tw15egan:
cc @aledavila

aledavila (approved):

PR #13242, opened 2023-02-27
fix(modal): aria-label a11y bug

### LINKED ISSUE #13166: [a11y]: The ARIA attributes "aria-label" are not valid for the element <div> with ARIA role "presentation"
### Package

@carbon/elements

### Browser

Chrome

### Operating System

Windows

### Package version

V10.*.*

### React version

_No response_

### Assistive technology

_No response_

### Description

Problem: The ARIA attributes "aria-label" are not valid for the element <div> with ARIA role "presentation" 

===== DEFECT DESCRIPTION ===== 
IBM 4.1.2 Name, Role, Value

Functionally, ARIA roles, states, and properties are analogous to CSS for assistive technologies. For screen reader users, ARIA controls the rendering of their non-visual experience. Incorrect ARIA misrepresents the visual experience, with potentially devastating effects on their corresponding non-visual experience. Incorrect ARIA misrepresents the intended experience to all assistive technology users, with potentially devastating effects.

Many HTML elements have native semantics that map, by default, to implied ARIA semantics (roles). Other elements specified as having No corresponding role do not have implicit ARIA semantics. Although they still have meaning it's not provided by ARIA. In order for assistive technologies to predictably represent the context and behavior of these elements (UI controls), these implicit semantics should not be overridden as defined in the ARIA in HTML - Document conformance requirements for use of ARIA attributes in HTML specification. It is therefore recommended that developers only add a role attribute to a semantically neutral and allowed HTML element such as a
or , rather than incorrectly attempting to override the semantics of the detected element. No ARIA is better than bad ARIA for assistive technology users.

===== EXPECTED RESULT =====
ARIA attributes must be valid for the element and ARIA role to which they are assigned

### WCAG 2.1 Violation

_No response_

### Reproduction/example

Reproduce

### Steps to reproduce

1. Go to https://ondeck.console.cloud.ibm.com/networking/vlans/2588291
2. Edit tag
3. Use IBM Equal Access Accessibility Checker to scan
image

### Code of Conduct

alisonjoseph:
Closes #13166 

#### Changelog

- adds aria-label to props so its not spread into ...rest on the wrapper div

#### Testing / Reviewing

Run accessibility checker on modal story, _"The ARIA attributes "aria-label" are not valid for the element `<div>` with ARIA role "presentation"_ error should no longer appear.

Aria label should still appear on the modal div (not the wrapper div)

https://deploy-preview-13242--carbon-components-react.netlify.app/iframe.html?args=aria-label:test-label&id=components-modal--playground&viewMode=story


francinelucca (approved):
Confirmed no longer see 4.1.2 violation
image

tw15egan (approved):

PR #13244, opened 2023-02-27
fix(numberinput): remove defaults from min and max

### LINKED ISSUE #13243: [Bug]: Number input component accidently has max set to 100 by default
### Package

@carbon/react

### Browser

_No response_

### Package version

1.21.0

### React version

18.2.0

### Description

Number input has 100 as maximum value even when `max` is not set.

Screenshot 2023-02-27 at 12 30 57 PM

### Reproduction/example

https://codesandbox.io/s/boring-fog-m498gt?file=/index.js

### Steps to reproduce

Remove `max` of ` <NumberInput>`

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Reverts #12768 

Closes https://github.com/carbon-design-system/carbon/issues/13243


packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -209,8 +209,8 @@ const NumberInput = React.forwardRef(function NumberInput(
     invalid = false,
     invalidText,
     light,
-    max = 100,
-    min = 0,
+    max,
```
tw15egan, line 212:
```suggestion
    max = Infinity,
```

Would this be a safe middle ground between the original issue and the breaking change?

tw15egan (approved):

packages/react/src/components/NumberInput/NumberInput.tsx
tay1orjones: Kind of. Even with this the `clamp` function will still return `NaN` if `min` is undefined.

andreancardona (approved):

dabrad26 (approved):

cknabe (approved):

francinelucca (approved):

hsu022210 (approved):

tay1orjones:
This fix was included in [1.24.0-rc.0](https://www.npmjs.com/package/@carbon/react/v/1.24.0-rc.0)

PR #13255, opened 2023-02-28
feat: adds types for Copy & CopyButton

### LINKED ISSUE #13227: Add TypeScript types to CopyButton

AlexanderMelox:
Closes #13227 

This PR adds types to `CopyButton` and `Copy`;

#### Changelog

**Changed**

```shell
packages/react/src/components/Copy/Copy.js -> packages/react/src/components/Copy/Copy.tsx
packages/react/src/components/CopyButton/CopyButton.js -> packages/react/src/components/CopyButton/CopyButton.tsx
```

#### Testing / Reviewing

Created a dummy component to verify that the types are working correctly.


AlexanderMelox:
This is my first typescript conversion PR so feedback would be appreciated! I am stuck on some issues, when changing Copy over to tsx I now get these new TS errors but it's for components that aren't TS.
Screenshot 2023-02-28 at 3 34 16 PM

Screenshot 2023-02-28 at 3 34 31 PM

What do we do in these situations?

francinelucca:
@AlexanderMelox looks like it's complaining because we're defaulting aria-label to null and it's expecting it to be `string | undefined`, so changing `aria-label` from this: 
```jsx
aria-label={
        (!children && (animation ? feedback : other['aria-label'])) || null
      }>
```
to this: 
```jsx
aria-label={
        (!children && (animation ? feedback : other['aria-label'])) || undefined
      }>
```
will hopefully take care of that.

The other `IconButton` issue is a bit more complex, I think it's due to `IconButton` not being typed yet, we can maybe get around that with this for now until we type `IconButton`. Not actually sure if this is the best way to go 😅
```jsx
  // TODO: remove these two lines when IconButton is properly typed
  type IconButtonType = React.ButtonHTMLAttributes<HTMLButtonElement> & {align: string, label: string | undefined}
  const TypedIconButton = IconButton as  React.FC<IconButtonType>

  if (FeatureFlags.enabled('enable-v11-release')) {
    return (
      <TypedIconButton
        align="bottom"
        className={classNames}
        label={animation ? feedback : other['aria-label']}
        onClick={composeEventHandlers([onClick, handleClick])}
        onAnimationEnd={composeEventHandlers([
          onAnimationEnd,
          handleAnimationEnd,
        ])}
        {...other}
        aria-live="polite"
        aria-label={
          (!children && (animation ? feedback : other['aria-label'])) || undefined
        }>
        {children}
      </TypedIconButton>
    );
  }
```

PR #13264, opened 2023-03-01
feat: migration examples prefix selectors

### LINKED ISSUE #12342: Migration example: Customizing the prefix of selectors in Sass and React
This issue is for creating a new example app showing how to configure a custom prefix for class names as well as auto-generated Ids.

Checklist
- [x] Create the example/directory based on the existing vite example (not nextjs, nextjs templates can't be run in codesandbox/stackblitz)
- [x] Example should use `<ClassPrefix>`, `usePrefix()`
- [x] Example should use `<IdPrefix>`, `useIdPrefix()`
- [x] Example should configure sass to have a custom class prefix

andreancardona:
Closes #12342
- [x] add prefix example
- [x] add id-prefix example

Testing / Reviewing
- checkout PR locally and cd into `examples/class-prefix` and `examples/id-prefix`  run yarn dev


packages/react/src/internal/useIdPrefix.js
```diff
@@ -7,7 +7,7 @@
 
 import React from 'react';
 
-export const IdPrefixContext = React.createContext(null);
+export const IdPrefixContext = React.createContext('cds');
```
andreancardona: @tay1orjones do we want to update this to `null` I'm doing it here, but want to double check before committing

packages/react/src/components/IdPrefix/index.js
```diff
@@ -7,7 +7,7 @@
 
 import PropTypes from 'prop-types';
 import React from 'react';
-import { IdPrefixContext } from '../../internal/useIdPrefix';
+import { IdPrefixContext } from '../../internal/';
```
francinelucca: is this change intentional? seems to me like the previous path was correct

examples/id-prefix/src/App.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { useIdPrefix } from '@carbon/react';
+import { IdPrefix } from '@carbon/react';
+
+function ExampleComponent() {
+  const idPrefix = useIdPrefix();
+
+  return (
+    <p>The current prefix is: {idPrefix}</p>
```
francinelucca:
```suggestion
    <p>The current id prefix is: {idPrefix}</p>
```

francinelucca (changes requested):
mostly LGTM, some questions.

Should we add a `README.md` to the examples with info on the app and basic install/run instructions ? See https://github.com/carbon-design-system/carbon/tree/main/examples/incremental-migration#readme

packages/react/src/components/IdPrefix/index.js
andreancardona: Oh weird don't know how that happend

andreancardona:
```suggestion
import { IdPrefixContext } from '../../internal/useIdPrefix';
```

francinelucca (commented):
LGTM! pending Taylor's reply on `packages/react/src/internal/useIdPrefix.js` changes.

Also looks like some of the CI checks were not successful @andreancardona

packages/react/src/internal/useIdPrefix.js
tay1orjones: No I think we need to keep it as `null`. IIRC this is because we weren't configuring an ID prefix by default before `useIdPrefix` was added. 

Adding a prefix after the fact like this would probably need to be a breaking change

tay1orjones: @andreancardona No I think we need to keep it as `null`. IIRC this is because we weren't configuring an ID prefix by default before `useIdPrefix` was added. 

Adding a prefix after the fact like this would probably need to be a breaking change

tay1orjones:
> Also looks like some of the CI checks were not successful @andreancardona

I think it just needs a `yarn install` at the root?

francinelucca (approved):

tay1orjones (approved):
LGTM! 🎉

PR #13173, opened 2023-02-16
fix(ListBox): a11y - move filter tag as sibling of menu trigger, esc key improvements

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12596

Refactors `MultiSelect` so that the Filter Tag is an adjacent element to the trigger `button`, rather than a descendant of the `button`. Previously, this was causing issues as an element with `role="button"` cannot be a descendant of a `button` element, causing an a11y violation.

This PR also includes changes to the `esc` behavior for the `MultiSelect` / `ComboBox` variants. The `x` to clear input for `FilterableMultiSelect` and the `x` for the tag selection are not accessed by tabbing. Instead, the actions a user could perform with these have been rolled into the input itself. For example:

**Multiselect**:
Open and select an item with the keyboard. One `esc` press should close the menu. The second `esc` press should clear the selection.

**FilterableMultiselect**:
Open and select an item with the keyboard. One `esc` press should close the menu AND **clear** the input. The second `esc` press should clear the selection.

**ComboBox**:
Open and select an item with the keyboard. One `esc` press should close the menu AND **retain** the input. The second `esc` press should clear the input.

#### Changelog

**New**
- A wrapper (`cds--list-box__field--wrapper`) for the Filter Tag and the trigger Button.
- `onFocus` and `onBlur` handlers attached to the wrapper to apply a focus outline when the inner button is focused/blurred. 
- Styles for the new wrapper component, which mimics the filterable styles. 
- Keyboard handlers to capture `escape` presses inside `MultiSelect`, `FilterableMultiSelect` and `Combobox`

**Changed**

- The trigger button and `ListBox.Selection` (the filter tag) are now adjacent siblings under the new wrapper, rather than the filter tag being a child of the trigger button.
- Adjusted some styles based on the new structure.
- `ListBox.Selection` (the `x` or filter `x`) no longer is in the tab order, and events are captured by the element itself. 

#### Testing / Reviewing

- Open the `MultiSelect` story and select an item. Run the Accessibility Checker and ensure no violations are found.
- Compare with the latest deployed storybook and ensure the component functions the same and there are no regressions.

**Multiselect**:
Open and select an item with the keyboard. One `esc` press should close the menu. The second `esc` press should clear the selection.

**FilterableMultiselect**:
Open and select an item with the keyboard. One `esc` press should close the menu AND **clear** the input. The second `esc` press should clear the selection.

**ComboBox**:
[...]

PR #13268, opened 2023-03-01
fix(ListBox): Standardize `ListBox` menu open behavior

### LINKED ISSUE #13185: [Feature Request]: Align Dropdown / Combobox / Multiselect `enter` key (open) behaviors
### The problem

Originally brought up in https://github.com/carbon-design-system/carbon/issues/12596

Our 4 `ListBox` variants behavior is different to open the menu. Some open on enter, some on space, some on down arrow. Comment from @mbgower from the original issue:

> Say, while doing so, and comparing interaction across all these things I noticed that pressing Enter was NOT opening the filtering multiselect listbox when it had focus. I thought it should, so went and looked at the others just to make sure everything was consistent. While doing so, noticed a couple of other things (sorry!):

> On your dropdown, Enter works fine to toggle the list open (and put focus on the item selected) and closed. I noticed that Down Arrow on a closed dropdown is opening but ALSO advancing past the currently selected item. I don't think that should happen. We definitely want the down arrow to open the listbox, but my expectation is that the action is identical to the Enter key: opens listbox and puts focus on the currently selected item. I think the APG [select-only comboxbox](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) reflects this pattern pretty accurately. BTW, I think I had an existing ticket on this one.
> 
> Your default multiselect is opening on Enter, but not opening on Down Arrow. (the complete opposite of the filerable variant.) Any reason for that? My expectation would be that the Enter would reveal the listbox, as would the Down Arrow. Unfortunately, there's really nothing directly parallel to this in the APG.
> 
> Still talking about the multiselect, the behaviour on open has some considerations. Normally I'd expect the list to open with the focus going to the first item (And BTW, the behaviour with multiselect is that where something is already selected it becomes the first item). However, because Enter is also used to select, this gets a little tricky. If the focus goes to something in the list, a user trying to toggle the list closed by pressing Enter could inadvertently either select the first unselected item or unselect the first selected item. So, my inclination is to say that opening a multiselect with Enter should open the list, but not put focus inside the list. The user would have to arrow down to select items. I'd invite discussion on this. IF the multiselect is opened with the Down Arrow, I think I'd expect regular behaviour: open and move focus to first item.
> 
> And just to return to the top of this comment, I confirm my expectation that the filtering multiselect should open on Enter

### The solution

We need to identify and implement the correct behaviors for opening all ListBox variants and align them as best as possible.

The keys that can be used:
- <kbd>Space</kbd>
- <kbd>Enter</kbd>
- <kbd>Down Arrow</kbd>

The variants:
- `ComboBox`
- `Dropdown`
- `FilterableMultiselect`
- `Multiselect`

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13185
Refs https://github.com/carbon-design-system/carbon/pull/13173

Aligns `ComboBox`, `Dropdown`, `MultiSelect`, and `FilterableMultiSelect` so that they all open on the same keys.

The keys that can be used:
- <kbd>Space</kbd>
- <kbd>Enter</kbd>
- <kbd>ArrowDown</kbd>

They can all be closed via <kbd>Escape</kbd> or by making a selection.

#### Changelog

**New**

- `MultiSelect`: Add open on  <kbd>Space</kbd> and  <kbd>ArrowDown</kbd>
- `FilterableMultiSelect`: Add open on  <kbd>Enter</kbd>

**Changed**

- Lots of whitespace formatting changes due to TS updates. (Anyway we can automate these?)

**Removed**

- Updated `Dropdown` stories to not have an initial item set, as it was making the open with <kbd>ArrowDown</kbd> seem like it was progressing to the 2nd item on the first enter. It should now go to the first item when opened with <kbd>ArrowDown</kbd> 

#### Testing / Reviewing

Go to `ComboBox`, `Dropdown`, `MultiSelect`, and `FilterableMultiSelect` and make sure they can all open with <kbd>Space</kbd>,<kbd>Enter</kbd>, <kbd>ArrowDown</kbd> and that they can all close with <kbd>Escape</kbd>

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.stories.js
+++ b/packages/react/src/components/Dropdown/Dropdown.stories.js
@@ -121,7 +121,7 @@ Playground.argTypes = {
     control: {
       type: 'text',
     },
-    defaultValue: 'this is an example label',
+    defaultValue: 'This is an example label',
   },
   warn: {
     control: {
@@ -139,7 +139,7 @@ Playground.argTypes = {
     control: {
       type: 'text',
     },
-    defaultValue: 'this is an example title',
+    defaultValue: 'This is an example title',
   },
   size: {
     options: ['sm', 'md', 'lg'],
@@ -158,7 +158,6 @@ export const Default = () => (
       id="default"
       titleText="Dropdown label"
       helperText="This is some helper text"
-      initialSelectedItem={items[0]}
       label="Dropdown menu options"
       items={items}
       itemToString={(item) => (item ? item.text : '')}
@@ -171,7 +170,6 @@ export const Inline = () => (
     <Dropdown
       id="inline"
       titleText="Inline dropdown label"
-      initialSelectedItem={items[0]}
       label="Dropdown menu options"
       type="inline"
       items={items}
@@ -186,7 +184,6 @@ export const WithLayer = () => (
       id="default"
       titleText="First Layer"
       helperText="This is some helper text"
-      initialSelectedItem={items[0]}
       label="Dropdown menu options"
       items={items}
       itemToString={(item) => (item ? item.text : '')}
@@ -196,7 +193,6 @@ export const WithLayer = () => (
         id="default"
         titleText="Second Layer"
         helperText="This is some helper text"
-        initialSelectedItem={items[0]}
         label="Dropdown menu options"
         items={items}
         itemToString={(item) => (item ? item.text : '')}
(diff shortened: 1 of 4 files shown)


andreancardona (approved):
tested and they all look good!

alisonjoseph (approved):
LGTM!

PR #13272, opened 2023-03-01
refactor(Dropdown): ariaLabel to aria-label

jsehull:
#12802 

Update `Dropdown` to use the proper `aria-label`

#### Changelog

**New**

- n/a

**Changed**

1. added `deprecated()` to the old propType
2. added new usage for `aria-label`
3. kept backwards compatibility so that both aria versions can work with `ariaLabel` (deprecated) and `aria-label` (new).
4. updated tests API

**Removed**

- n/a

#### Testing / Reviewing

1. Open Storybook > Dropdown in browser for Dev Tool testing
2. Open `Dropdown.stories.js` in code
3. Add one aria label type at a time to the component in story code. 
 - 1. Only one version can be active at a time.
   - `aria-label` (new)
   - `ariaLabel` (deprecated))
 - 2. Be sure to find the correct usage for the given component as explained in code or on the story itself
 - 3. Test examples below. 
   - Note: For my testing, I duplicate the lines and then comment out the one I am not using. The image blow does NOT yet comment out one of the test lines:
image

4. confirm that both propTypes still display in Dev Tools as intended
5. confirm that Storybook > Docs > Component API contains both propTypes as options


francinelucca (approved):

tw15egan (approved):

jsehull:
I don't think the above conflict above is solely TS related, as it seems to duplicate a section, but since it is TS I'm not familiar with what needs to happen to resolve. Any suggestions?

francinelucca:
Hi 👋🏼 @jsehull. Typesript files' conflicts can be resolved the same way as in a `.js` file, shouldn't make a difference

Issue #8076, opened 2021-03-11
Selection in Menus should wrap based on platform norms


carmacleod:
Interestingly, native menus wrap selection on Windows, and they don't wrap on Mac. By "wrap" I mean that if the user keeps typing down arrow [up arrow], they will eventually wrap to the top [bottom] menu item and continue selecting from there.

It would be nice to special-case whether or not to wrap Carbon menus (including context menu and overflow menu) based on the platform, and the behavior needs to be consistent across components.

Combobox wraps, Dropdown doesn't, OverflowMenu wraps, ContextMenu doesn't.

Need a proper inventory of components (instead of just my 2-second test ;) and need to be sure how the equivalent platform components behave by testing more thoroughly before deciding how to proceed.

I couldn't decide whether to open this as a bug or an enhancement, but I decided on bug, because as a Windows keyboard user, the context menu not wrapping felt really wrong to me.

emyarod:
in the case of dropdown I believe it's also mimicking Select so that's why it currently does not wrap. should we tackle this on a component by component basis or should we adopt the existing platform behaviors?

carmacleod:
Question: Do people use Carbon Dropdown for menus? If so, that complicates things.

Personally, I would tackle all of the menu-related components first, and make them match the platform's menu wrapping behavior.

Then look at the dropdowny/comboboxy/select-like components on a case by case basis. Maybe they are ok? There was some inconsistency - maybe that should be looked at? What do the platforms do?

Mostly, I care about the menu behavior. I only mentioned combobox/dropdown because @janhassel asked about dropdowns, which at first I thought meant "dropdown menus". Still not sure what was meant... ;)

janhassel:
@carmacleod No, Dropdowns shouldn't be used for menus I think. I only brought that up because of the similar interaction (list of clickable elements), maybe that was a bad reference.

I agree, we should try to follow the respective system's behaviour for the individual components. Do you have a list of what native components wrap on which system and which don't?

carmacleod:
> Do you have a list of what native components wrap on which system and which don't?

Not really. I only really noticed that menus (including context menus) wrap on Windows (I'm mostly a Windows user).
Then I checked Mac and saw (to my surprise) that menus (and context menus) don't wrap there.

----

Regarding other popup-listy things, looking in Settings - General on **Mac**, native pop-up buttons don't wrap.
[...]

PR #13277, opened 2023-03-02
fix(menu): focus first item when opened

### LINKED ISSUE #13265: OverflowMenuV2: Should focus first item on open
### Package

@carbon/react

### Browser

Chrome

### Package version

1.23.1

### React version

v17.0.2

### Description

[OverflowMenuV2](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-overflowmenuv2--custom-icon) should immediately focus the first `MenuItem` after being opened, compare to [OverflowMenu](https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default)'s behavior, which is correct.

Note: the fix for this should be at the `Menu` component level which is where this logic is abstracted and what `OverflowMenuV2` is implementing under the hood

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/experimental-unstable-overflowmenuv2--custom-icon

### Steps to reproduce

1. Tab into the button that opens the overflowMenu
2. press Enter
3. Notice the focus is on the container element of the `MenuItem`s , focus should be in the first `MenuItem` element.
4. Repeat same steps in https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default
5. Notice the focus is shown correctly in the first `MenuItem`

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

janhassel:
Closes #13265
Contributes to #8076

#### Changelog

**Changed**

- When a menu is opened, the first item will receive focus by default instead of the entire menu widget
- Focus cycles now (when the last menu item is focused and the user presses <kbd>Arrow down</kbd> the first item will be focused)

#### Testing / Reviewing

Storybook

---

This affects everything that consumes `Menu` which for now is `OverflowMenuV2` and `Menu` itself (likely in combination with `useContextMenu`). With #13224 this will also trickle down to `ComboButton` and `MenuButton`.


francinelucca (approved):
LGTM, thanks! 🔥

tay1orjones (approved):

PR #13281, opened 2023-03-02
refactor(Notification): ariaLabel to aria-label

jsehull:
#12802 

Update `Notification` to use the proper `aria-label`

#### Changelog

**New**

- n/a

**Changed**

1. added `deprecated()` to the old propType
2. added new usage for `aria-label`
3. kept backwards compatibility so that both aria versions can work with `ariaLabel` (deprecated) and `aria-label` (new).
4. updated tests API

**Removed**

- n/a

#### Testing / Reviewing

1. Open Storybook > Notification in browser for Dev Tool testing
2. Open `ActionableNotification.stories.js` in code
- 1. as well as `InlineNotification.stories.js` and `ToastNotification.stories.js`
4. Add one aria label type at a time to the component in story code. 
 - 1. Only one version can be active at a time.
   - `aria-label` (new)
   - `ariaLabel` (deprecated))
 - 2. Be sure to find the correct usage for the given component as explained in code or on the story itself
 - 3. Test examples below. 
   - Note: For my testing, I duplicate the lines and then comment out the one I am not using. The image blow does NOT yet comment out one of the test lines:
image

5. confirm that both propTypes still display in Dev Tools as intended
6. confirm that Storybook > Docs > Component API contains both propTypes as options


packages/react/src/components/Notification/stories/InlineNotification.stories.js
```diff
@@ -23,6 +23,7 @@ export default {
     kind: 'error',
     lowContrast: false,
     hideCloseButton: false,
+    ['aria-label']: 'closes notification',
```
francinelucca, line 26: should we remove the `ariaLabel` arg here? seems weird to have both. Same comment for the other notifications

francinelucca (changes requested):

packages/react/src/components/Notification/stories/InlineNotification.stories.js
jsehull: Good question. I think you're right, I may have done it out of habit for the _component_ prop types, but this is just the default given for use on a _story_ instead. I'll remove the old one and keep the new.

packages/react/src/components/Notification/Notification.js
```diff
@@ -396,6 +419,8 @@ ToastNotification.defaultProps = {
 };
 
 export function InlineNotification({
+  ['aria-label']: ariaLabel,
+  ariaLabel: deprecatedAriaLabel,
```
francinelucca: I'm thinking since we are adding this prop to `InlineNotification` for the first time now, we can probably get away with not adding `ariaLabel` at all and just adding `aria-label` here. I don't think it's necessary to add a deprecated prop here that wasn't supported previous to this PR 🤔

packages/react/src/components/Notification/stories/InlineNotification.stories.js
jsehull: @francinelucca added the changes

packages/react/src/components/Notification/Notification.js
jsehull: Yeah that makes sense. The only reason I could think to have it in there is so that the notification types are uniform. Would you like for me to remove `ariaLabel`?

francinelucca: Yeah I say let's go ahead and remove it. thanks! @jsehull

francinelucca (approved):
🚀

tay1orjones (approved):

Issue #10616, opened 2022-02-02
Add "xs" size variant


janhassel:
### Summary

Related: #6202, #7704, #10617

*This issue is to track progress of a new size variant (`xs/1.5rem/24px`) for components and is based on a discussion with @tw15egan and @aagonzales.*

---

The need for a size variant below `sm` has been identified by various teams building information-rich layouts. The goal is to offer a new size variant: `xs` which would equate to 1.5rem/24px in component height. The left and right paddings remain the same as for any other size variant to ensure alignment with other components of the UI.

### Available extra resources

Our UX Engineering team will be able to support the development of this enhancement

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

laurenmrice:
We will be working on this effort in this related issue [#6202](https://github.com/carbon-design-system/carbon/issues/6202?reload=1), so going to close.

Issue #7704, opened 2021-02-01
Need for compact inputs / components


janhassel:
### Summary

The current component designs are very limiting in how interface can be designed. Particularly produtivity (web)applications are suffering strongly from these limitations. There is strong need for more compact versions of the components (whether it be through a prop or an overall feature flag).

### Justification

Picture a sidepanel with number inputs – a fairly common UX pattern. Take a look at Keynote, Sketch or Figma for example.

When using the smallest available number inputs (32px high, 150px wide) and a spacing of 8px in between, the width of the 
sidepanel is 213% of the size of sketch's sidepanel. A sidepanel with this width is not feasible and wouldn't even fit in the `sm` breakpoint.

![image](https://user-images.githubusercontent.com/28265588/106435125-bf17f600-6472-11eb-9686-37c484ae2944.png)

### "Must have" functionality

Offer compact versions of components, reducing their inner paddings and min-widths. Potentially find altermative solutions for validation messages to keep the width down).

aagonzales:
Yeah we should evaluate if we need min-widths on input components. Or maybe change it depending on component size. I've brought this topic up with the team today so hopefully we can get some traction on it soon.

Related issue: https://github.com/carbon-design-system/carbon/issues/6202

tw15egan:
I agree, unless the visuals of the component are changed under a certain width, input components should simply fill 100% of their container. That way, a small width can be specified on the container and the inputs will shrink to fit, without a `min-width` preventing that. I think this can be broken into two different issues

- Removing `min-width` on form components
- Provide an `xs` size variant to reduce the overall height

janhassel:
@tw15egan for the `xs` size it would be helpful to reduce the component's padding. The number input currently uses spacing-05 I think, which could be reduced to spacing-03 maybe?

wnm3:
We are very limited in space provided in an iframe to 266x699 so having absolute control over spacing is necessary. Overriding CSS becomes excessive and seems to undo the benefits of using a design framework. I'd even go as far as asking for xxs where minimum padding and margins are used allowing us to add back what is needed to improve the view and giving us control over component placement, alignment.

Issue #6202, opened 2020-06-05
Condensed components ☂️


pascal-potvin:
# Acceptance criteria

Systematically support condensed sizes across core components so teams can build more compact and dense layouts without breaking from the system.

### Justification

Many IBM product teams build content-heavy, data-dense applications where screen real estate directly impacts user efficiency. Also, as teams transition legacy UIs to the latest version of Carbon, they can require smaller component heights and denser layouts to accommodate the amount of information users need to see at once. Acquisitions may also face having the need to translate their smaller size components into ours.

Users working with complex pipelines, models, tables, and rules rely on high information visibility to complete tasks effectively. However, density is not consistently supported across Carbon components. This forces teams to implement custom spacing, type, and sizing overrides, increasing maintenance costs and reducing design–development parity.

While Carbon’s geometry assumes web-based patterns like scrolling and pagination, there are power users in certain products that operate in sustained, tool-like workflows where maximum on-screen visibility is essential. Without systematic support for smaller component sizes, teams struggle to meet these functional needs within the system.

### Why this matters
- Speeds up delivery and reduces maintenance for designers and developers
- Preserves user efficiency in complex workflows
- Improves Carbon adoption in data-heavy products

# Phases

#### Audit, find use cases, and make a final list of component sizes needed.
- [x] Audit the design system components for where we can introduce new potential sizes.
- [x] Get real use cases from products teams of where they need smaller component sizes.
- [x] Make final decisions on which new component sizes we need to add to the system.
- [x] Determine which components to start adding sizes too. Which ones are higher and lower priority. 

#### Add 5-10 high priority components sizes to the system

**Dev:**
- **Contextual layout tokens:** This needs to be done first and separately from this workstream.
- **React/WC:** Implement in both frameworks
- **Figma Code Connect:** This will be done at a later date depending on the outcome of how Figma will be structured. 

**Design:** Add to the Figma kit and Website documentation.

> [!Important]
>  Because of v12 priorities, only complete Phase 1 first. Phase 2 will continue after v12 is released.

Phase 1
- Select `xs`
- Search `xs`
- Pagination `xs`
- Text input `xs`
- Password input `xs`
- Dropdown `xs`
- Combobox `xs`
- Multiselect `xs`
- Data table toolbar `xs`
- Data table pagination `xs`
- Contained tabs `sm`
- Line tabs `sm`
- Vertical tabs `lg`, `md` , `sm`

## After v12 is released

Phase 2
- Number input`xs` 
[...]

PR #13287, opened 2023-03-03
feat: add experimental contextual layout token system

### LINKED ISSUE #10318: [Bug]: Button padding is not using relative units

### LINKED ISSUE #10617: Condensed container component
### Summary

Related: #6202, #7704, #10616

This issue is to further discuss the addition of a utility component which would locally override spacing tokens to support condensed UI zones like sidebars based on a discussion with @tw15egan and @aagonzales.

---

In situations where many components (especially form elements) are placed in a narrow container such as a sidebar, the need for more condensed components has been identified. The current `sm` size or proposed `xs` size variants do not fully cover this use case as the horizontal space / padding inside the components is critical.

The idea @tw15egan, @aagonzales and I came up with is to offer a utility component that locally overrides the spacing token scale to only affect its children.
This would mean that (theoretically) no additional prop or individual component work would be needed and all components *should* adapt correctly and consistently.

```scss
:root {
  --cds-spacing-05: 1rem;
}

.cds--condensed-zone {
  --cds-spacing-05: 0.5rem; // can be hardcoded or dynamically calculated with a predefined factor
}
```

Any component using `spacing-05` (like for `padding-left` and `padding-right`) would then render as a narrow / condensed variant. Obviously, the entire spacing scale should be redefined instead of just individual values.

This utility component should only be used in contexts with no alignment to other components, meaning that it should not be used in the main area of a UI for example but only in something like a slide-in or slide-over sidebar. In this context all components should be children of the utility component to achieve consistency and alignment within the zone. This needs to be clearly stated in the design guidance / documentation.

The type scale would also likely need to be adjusted a bit.

### Justification

_No response_

_No response_

### Required functionality

_No response_

_No response_

### Available extra resources

Our UX Engineering team will be able to support the development of this enhancement

### Code of Conduct

### LINKED ISSUE #13423: [Bug]: IconButton `size` prop default should be `lg`
### Package

@carbon/react

### Browser

Chrome

### Package version

1.25.0

### React version

17

### Description

IconButton's `size` prop has an incorrect default of `md`. Buttons have a default size of `lg`, and this includes the IconButton. `Button` with `hasIconOnly` works correctly.

Some background on this:
- In a v11 beta the IconButton initially did not set the `size` (https://github.com/carbon-design-system/carbon/discussions/10247#discussioncomment-1798312)
- This defect was fixed in https://github.com/carbon-design-system/carbon/pull/10339 and a default was added, but it was `md`.

```[tasklist]
- [ ] Update the default `size` in IconButton to be `lg`
```

janhassel:
Implementation of #10619

Closes #10318 
Closes #10617
Closes #13423 

Contributes to #6202
Contributes to #7704
Contributes to #10616

---

This PR adds an experimental system for contextual layout tokens. These allow for layout contexts to change specific rendering settings that are applied to all components within a context. At the moment, these settings are `size` and `density`.

The system consists mainly of the sass module, but also offers a convenience component in React which can be used to create a new layout context.

The following example will set both `Button` as well as `TextInput` to use `size = lg`

```jsx
<Layout size="lg">
  <Button />
  <TextInput />
</Layout>
```

The system is backwards compatible and components accepting a `size` prop will still render in the appropriate size. In addition, any component explicitly setting a `size` will use that over the layout context's setting:

The following example will render Button as `sm` and `TextInput` as `lg`.

```jsx
<Layout size="lg">
  <Button size="sm" />
  <TextInput />
</Layout>
```

"Density" is an experimental new layout setting which can be set to `condensed` (defaults to `normal`). It will provide a preferred `padding-inline` value which is `spacing-05` for `normal` and `spacing-03` for `condensed`.

---

Components can use the sass functions to more easily access the layout context settings:

```scss
.cds--component {
  @include layout.use('size', $default: 'md');
  @include layout.use('density', $default: 'normal');

  height: layout.size('height');
  padding-inline: layout.density('padding-inline');
}
```

Components that use non-standard sizes can still participate in layout contexts by locally overriding the tokens. An example can be seen in the `Tag` styles.

See the updated `packages/styles/docs/sass.md` document for the full details.

In order to apply specific constraints to children components that might differ from their own preference, the `<LayoutConstraint>` utility component can be used. Like `<Layout>` all it does is setting class names.

An example of this use case is the icon-only content switcher. It uses IconButton under the hood which supports all sizes and defaults to `lg`. The ContentSwitcher though should only support `sm`->`lg` and defaults to `md`.

```jsx
<div className="cds--component">
  <LayoutConstraint size={{ default: 'md', min: 'sm', max: 'lg' }}>
    {/* IconButton is now locked to the sizes sm->lg and defaults to md */}
    <IconButton … />
  </LayoutConstraint>
</div>
```

---

This PR includes migration for the following components:
- Accordion
- Button
  - percy highlights a bunch of visual differences between this PR and the current button implementation. This is because the current button implementation was actually differing from the visual specs (most noticeable the `padding-inline` was changing for each step below `lg`)
- ContainedList
- ContentSwitcher
- Tag
- TextInput


janhassel:
@tay1orjones Yes, I just pushed the discussed update which also makes IconButton render as `lg` by default. Through percy I noticed that this will actually also fix the `PaginationNav` component which seems to have been affected by the regression as well since the ghost arrow buttons were using `md` while the number buttons were using `lg`.

I will update the original description of this PR (hopefully tomorrow) and mark it as "ready for review" afterwards.

janhassel:
@tay1orjones I marked this PR as "ready for review" now and linked the associated issues it would close or contribute towards.

@alisonjoseph @andreancardona Please let me know if you'd like to set up a quick call going over this PR and the concept in person!

alisonjoseph (commented):
This is all working as intended and looks great! 🥳 My one question is around surfacing the docs for this. A little worried someone will see the new links in storybook and have no idea what they mean and know to go look under `unstable__Layout` in the Experimental section?
Screenshot 2023-04-06 at 8 48 15 AM

Would it be more obvious if we used the unstable name for the storybook button? 
Screenshot 2023-04-06 at 8 52 09 AM

andreancardona:
@aagonzales Do you mind taking a quick review at the Percy Snapshots - just want to make sure all looks good on your end! :)

aagonzales (commented):
I agree with Alison, from a technical perspective, it is working as intended (its super cool to see it live!) but how it is presented in Storybook is also concerning to me. Just because these settings are now possible doesn't mean they are designed-approved variants (or maybe they are we just have done the work to validate it). I think there may also be confusion between when to use these new controls vs just the size prop. Also when a density or layout step is not available in a component but still shows in the menu but doesn't change anything could also be confusing to users. 

Is there a way to show these controls on the experimental layout section for no, until we get better design usage documentation on how we want these controls to be used per component and it what situations? Or if that's not possible, I think we should de-emphasise the controls. The beta icon and the label draw attention to these icons above all the other controls in the toolbar. If we can use just icons here and have them fly a little more under the radar for now that would be better.

aagonzales:
This is my read on the Percy alerts @andreancardona 

**Things to be addressed:**
- It looks like the icon in the `button` component is 1px too high now with an incorrect padding-top or bottom.
- This looks like it might be true for all icons showing in the Percey screens, they have all moved up 1px. Effects `containedList`, `accordion`
   
 **Acceptable changes**  
- For file uploader, looks like this actually fixes a bug in main. The padding for the button was off line main 🤔 not sure how that happened. So that update is good.
   - Also true for the button in `toggleTip`, that change can stay.
- `toggle` differences have just changed to show the `md` size as default instead of `sm` which is fine. 
- `copyButton` has also changed the default size showing which fine
- `paginationNav` looks like the external padding around the story maybe and not about the component itself?

andreancardona:
@janhassel Hey Jan! would love your thoughts on Anna's review. Thank you :)

janhassel:
@aagonzales
I updated the button component, will take a look at `Accordion` later.

Most of the items you listed under "Acceptable changes" are fixes for regressions that were caused by the `IconButton` size being wrong (ref #13423). For PaginationNav, this meant that the icon buttons with arrows were smaller than the buttons with numbers, hence the change in percy. You can see this in main when you hover over one and focus the other:
image

@aagonzales @alisonjoseph I changed the name to "unstable__Layout <group>" and only made them visible in the dev storybook (`yarn storybook`) so that we can test components for support without confusing end users. What do you think about this?

aagonzales:
> I changed the name to "unstable__Layout " and only made them visible in the dev storybook (yarn storybook) so that we can test components for support without confusing end users. 

Yes, that's an excellent solution for this.

janhassel:
@aagonzales Sorry, it took a bit longer than I had hoped but it's ready for review again. A new visual regression happened with the addition of icon-only content switchers.
Would you mind checking this one specifically? With this PR, the individual switches are 40px x 40px, whereas I think in the current main implementation, they're slightly smaller. Here's a comparison between the two with 3 40px squares overlayed:

Current main:
image

This PR:
image

Maybe I'm misreading the visual spec though, so let me know which one is the "correct" one.

packages/styles/scss/_layout.scss
```diff
@@ -6,46 +6,80 @@
...
           $value
         );
       }
+
+      // CSS max() doesn't accept '0' as a value without a unit
+      // stylelint-disable length-zero-no-unit
+      @include custom-property.declaration(
+        'layout-#{$group}-#{$property}-min',
+        0px
+      );
```
andreancardona, line 31:
```suggestion
      );
      
```

janhassel:
@alisonjoseph @andreancardona I think the labels "status: merge conflict" and "status: waiting for author's response" can be removed now.

Let me know if you have any questions!

alisonjoseph:
I'm seeing some visual breaking changes with this PR in Percy 👀 

Tag
Screenshot 2023-05-03 at 9 24 41 AM

Copy Button
Screenshot 2023-05-03 at 9 27 17 AM

Button in File Uploader
Screenshot 2023-05-03 at 9 28 31 AM

Also in the icon content switcher, but looks like you confirmed above with Anna that is correct. 👍

janhassel:
@alisonjoseph

- Tag
  - The default size for `<Tag />` is `md`, but the default story was showing them in `sm`. This PR removed the explicit size override which is why they render in `md` now.
- CopyButton
  - Do you remember if `CopyButton` used to be `md` in v10 as well or is this also a regression by the difference in size for IconButton (#13423)?
- Button in FileUploader
  - The implementation of buttons in main currently differ from the visual spec: the padding-inline decreases for sizes `sm` and `md`. This PR includes a fix for these defects. Percy shows it in this story as the button in FileUploader is set to `md`

janhassel:
@alisonjoseph Ok, I wrapped `CopyButton` in a `LayoutConstraint`. Let me know if this works for you!

andreancardona (approved):
reviewed on my end and looks good - would just want a last  pass / visual review

janhassel:
@alisonjoseph @aagonzales Let me know if you have any questions!

alisonjoseph (approved):

aagonzales (approved):
Approved! Awesome work Jan!!! 🚀 🎉

janhassel:
@alisonjoseph @aagonzales I think one of you still need to approve the changes in percy for all tests to pass

PR #12924, opened 2023-01-05
test(eslint): update eslint to error if style prop is used

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12803

Updates our `eslint-config` to forbid using the `style` prop/dom node. 

#### Changelog

**New**

- `react/forbid-component-props`
- `react/forbid-dom-props`

#### Testing / Reviewing

Try adding a `style` prop and make sure an error appears in your editor.

PR #13289, opened 2023-03-03
docs(style): add section on authoring inline styles

tay1orjones:
I realized in https://github.com/carbon-design-system/carbon/pull/13224#discussion_r1124591659 today that we never documented this after https://github.com/carbon-design-system/carbon/pull/12784 and https://github.com/carbon-design-system/carbon/pull/12924

#### Changelog

**New**

- adds docs for how to author inline styles correctly for CSP


francinelucca (approved):
🚀

tw15egan (approved):
LGTM 👍 ✅

Issue #12525, opened 2022-11-07
Add TypeScript types to TableExpandRow


tay1orjones:

mbarrer:
Starting work on this today

Issue #12522, opened 2022-11-07
Add TypeScript types to TableCell


tay1orjones:

mbarrer:
Have also started conversion on this component due to some knock on changes from other table components

PR #13297, opened 2023-03-06
refactor(typescript): Convert TableCell, TableExpandRow, TableExpandedRow, TableRow to tsx

### LINKED ISSUE #12529: Add TypeScript types to TableRow

mbarrer:
Closes #12529, #12522, #12525, #12526

Converts TableCell, TableExpandRow, TableExpandedRow and TableRow to tsx in preparation for the main Table conversion

#### Changelog

**Changed**

- Added JSDoc typings on `wrapComponent.js`
- Converted TableCell.js -> TableCell.tsx
- Converted TableExpandRow.js -> TableExpandRow.tsx
- Converted TableExpandedRow.js -> TableExpandedRow.tsx
- Converted TableRow -> TableRow.tsx

#### Testing / Reviewing

To verify these changes are working, we can check to see if the existing storybook behavior is still working correctly. Should be able to properly import converted components in a TS environment


packages/react/src/tools/wrapComponent.js
```diff
@@ -10,7 +10,16 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { usePrefix } from '../internal/usePrefix';
 
+/**
+ * @param {{ name: string, className?: string, type: string }} param0
```
francinelucca:
```suggestion
 * @param {{ name: string, className?: string | Function, type: string }} param0
```

francinelucca (changes requested):
LGTM, one minor change suggestion

packages/react/src/tools/wrapComponent.js
mbarrer: Ah ty I missed the function usage. I would prefer something more strict than `Function` though. Would you settle for this instead? 

```
@param {{ name: string, type: string, className?: string | (prefix: string) => string }} props
```

francinelucca: sounds good! though technically the `getClassName` function could return an object with classNames as keys and it'd still work 🤔: 
image

mbarrer: Ah right because most components consume the className with that `classnames` lib. I didnt see any current usages in the code where the getClassName returns an obj, I could change it to this though. might be a bit too wordy for a jsdoc typing and better served when we convert the internal files to ts. Let me know if I should change it

image

francinelucca (approved):

packages/react/src/tools/wrapComponent.js
francinelucca: you're right, doesn't seem like it's being used anywhere returning an object, I'm fine keeping it as-is. thanks!

tay1orjones (approved):

PR #13301, opened 2023-03-07
fix(NumberInput): set defaults min/max prop

### LINKED ISSUE #12695: [Bug]:  NumberInput error if min or max props are not present

jsehull:
Closes #12695

As requested in the [latest comment](https://github.com/carbon-design-system/carbon/issues/12695#issuecomment-1447124989) for #12695, added "bound" limit to the clamp function and not directly on the component itself.
___
Let me know if the documentation is too verbose

#### Changelog

**New**

- add new prop defaults for the min/max values in the `clamp` function

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

1. Open Storybook > NumberInput
2. Open `NumberInput.stories.js` in code
3. Confirm the component works as expected if:
   - `min` and `max` have specific values 
   - `min` and `max` are NOT used
   - NOTE: warnings for incorrect #'s only show up IF a min/max is given (i.e. going over the initial `boundLimit` will NOT be flagged with an inline error unless specified as a min/max)


jsehull:
What do errors with Percy require on my end?

tw15egan:
@jsehull I think the branch is just out of date, there should be no visual regressions with this change

tw15egan (approved):
Not getting any errors if I do not set a `max` or `min` value in the `Default` story

alisonjoseph (approved):

Issue #12360, opened 2022-10-20
[a11y]: Search component labelling causing some potential failures


mbgower:
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-search--default

### Automated testing tool and ruleset

IBM Accessibility Checker, latest deployment

### Assistive technology

JAWS

### Description

The search component implementation -- as evidenced by topics such as https://ibm-studios.slack.com/archives/C036P1CTN/p1664802238686169 and https://ibm-studios.slack.com/archives/C0M053VPT/p1660229567172769 -- can result in a failure of IBM's current checker rules, as well as create a verbose screen reader experience.

Here are the main problems:
- Verbosity: Because the search input is within a search region, both of which are called "search" by default, the screen reader output is a bit verbose "Search SearchRegion Search Search edit"
- Duplicate labels: Because Carbon provides this default label for all search components, when a second search mechanism is used on a page, it causes an automatic fail of a checker rule which requires multiple regions of the same type to have unique labels

I think a solution to this may be to make the placeholder text in any search input into its name, with "search" as a fall back if no placeholder text exists.
I also think Carbon could consider giving the UI Shell Header search region a default name of "site" -- or potentially remove the search region entirely. I'm waiting for some clarification from the ARIA authoring group (https://github.com/w3c/aria-practices/issues/2516) which may help guide Carbon resolution.

### WCAG 2.1 Violation

4.1.2 Name, Role, Value

### Reproduction/example

use react storybook

### Steps to reproduce

Open https://react.carbondesignsystem.com/?path=/story/components-search--default

Inspect the aria-label given to both the search role and the input

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
> I think a solution to this may be to make the placeholder text in any search input into its name, with "search" as a fall back if no placeholder text exists.

Yeah, This seems really reasonable for now until if/when we get a firm idea on removing the region entirely or not.

gcattan:
If you have two Search with the same placeholders, then you still have an accessibility issue:

[...]

PR #13306, opened 2023-03-07
fix(Search): add default placeholder (a11y)

jsehull:
#12360 

Adding temporary solution as suggested in #12360 until greater decision is made.

#### Changelog

**New**

- add default placeholder value to `Search`

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

Visit Storybook > Search to confirm:
1. default placeholder is "search" when no placeholder is given
2. if a placeholder IS used, the placeholder is visible and NOT "search"
   - Search Playground is a good place for this


packages/react/src/components/Search/Search.js
```diff
@@ -32,7 +32,7 @@ const Search = React.forwardRef(function Search(
     onClear = () => {},
     onKeyDown,
     onExpand,
-    placeholder = '',
+    placeholder = 'search',
```
andreancardona:
```suggestion
    placeholder = 'Search',
```

andreancardona: Small suggestion - seems like custom is to capitalize - was browsing different examples:  http://web-accessibility.carnegiemuseums.org/code/search/

All else looks good 🔥

jsehull: Good eye, can do!

tw15egan (approved):

francinelucca:
@mbgower is this what you meant by "make the placeholder text in any search input into its name, with "search" as a fall back if no placeholder text exists." ? or do we need to add the `name` html attribute to the Input tag? I seem to have confused myself here

mbgower:
HI @francinelucca, I find I need to navigate to the Search component when I follow the react build link (so if I'm doing that wrong, let me know!).

I see a default value in the search edit called "A default value". I want to make sure that no one thinks this is what I'm talking about (and I don't understand why that string exists, as I've never encountered such a thing in the real world).
image

If I clear this default value, then I see the placeholder text for this search input, which is "search"
image

In my experience, most teams do not use "Search" as their placeholder value, but this could be the fall back in Carbon where a team doesn't specify anything.  All Carbon needs to do is make it so whatever the string for the placeholder is, is also the (non-visible) label for the region. Remember, we already have region of _type_ search, so it's not like there isn't already context provided.

Does that answer your question?

tw15egan:
I just pushed up some changes that @francinelucca and I were messing around with. It will set the placeholder value as the `aria-label` on the outer `div` that has a role of `search`. It seems like this way, we can have multiple Search instances with the same `labelText` as long as they have differing `placeholder` text. The screenreader reads the placeholder text, although it is still a bit verbose. 

Check the `with layer` stories and see if you come across any violations, and test with a screenreader

francinelucca:
> I just pushed up some changes that @francinelucca and I were messing around with. It will set the placeholder value as the `aria-label` on the outer `div` that has a role of `search`. It seems like this way, we can have multiple Search instances with the same `labelText` as long as they have differing `placeholder` text. The screenreader reads the placeholder text, although it is still a bit verbose.
> 
> Check the `with layer` stories and see if you come across any violations, and test with a screenreader

Thanks @tw15egan , @mbgower!
@mbgower is this what you would expect screenreader-wise in these instances with the corrected changes in place? https://deploy-preview-13306--carbon-components-react.netlify.app/?path=/story/components-search--with-layer

mbgower:
There is a slight inconsistent oddity on the first (and mabye third?) input not happening on the second one. But I think either is still an improvement.
I made a short video that hopefully demonstrates.

https://user-images.githubusercontent.com/14298245/224431887-b4fc3dff-e5a5-4413-bd5a-5cd890059688.mov

francinelucca (approved):

tw15egan (approved):

PR #13308, opened 2023-03-08
Contained Tabs grid POC

alisonjoseph:


tay1orjones:
We landed this in https://github.com/carbon-design-system/carbon/pull/13927

PR #13310, opened 2023-03-08
fix(CodeSnippet): a11y tabstop + attributes

### LINKED ISSUE #10527: [a11y]: add required tab stop to multiline Codesnippet textbox
multiline codesnippet has an inaccessible scrollable text box. this text box should be able to receive focus and have a tab stop

jsehull:
Closes #10527

**The following changes are only for the `Singleline` and `Multiline` snippets, but do not include `Inline`.

Primary fix: add tab stop for `Multiline` `CodeSnippet`
I also added a few more `aria` attributes as requested in #10527 comments.
___

I would like confirmation on `aria-multiline`. Documentation suggests that without it, screen readers assume it is a single line. Obviously it's now `readonly` so they won't edit, but I thought it may help inform users that they can/should scroll since it is not one line. 

Thoughts on keeping/not?

#### Changelog

**New**

- add tabstop to `Multiline` snippets
- add `role='textbox'` to multi (single currently has)
- add `aria-readonly` to single/multi (users cannot interact with the "textarea")
- add `aria-multiline` for multi (single will never have)

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

Visit Storybook > `CodeSnippet` to compare `aria` changes for Single and Multi line snippets. Check for the following:
1. Multiline tab stop is working (tab selects + allows keyboard scrolling)
2. Single/Multi both show a `role` of 'textbox'
3. Single/Multi have are `aria-readonly`
4. Multi now has `aria-multiline` added

Multiline example:
image

### DIFF
--- a/packages/react/src/components/CodeSnippet/CodeSnippet.js
+++ b/packages/react/src/components/CodeSnippet/CodeSnippet.js
@@ -226,10 +226,14 @@ function CodeSnippet({
     <div {...rest} className={codeSnippetClasses}>
       <div
         ref={codeContainerRef}
-        role={type === 'single' ? 'textbox' : null}
-        tabIndex={type === 'single' && !disabled ? 0 : null}
+        role={type === 'single' || type === 'multi' ? 'textbox' : null}
+        tabIndex={
+          type === 'single' || (type === 'multi' && !disabled) ? 0 : null
+        }
         className={`${prefix}--snippet-container`}
         aria-label={ariaLabel || 'code-snippet'}
+        aria-readonly={type === 'single' || type === 'multi' ? true : null}
+        aria-multiline={type === 'multi' ? true : null}
         onScroll={(type === 'single' && handleScroll) || null}
         {...containerStyle}>
         <pre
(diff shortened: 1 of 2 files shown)


tw15egan:
Looks great. Can we adjust the `focus` styles so that it aligns with Carbon?

`_code-snippet.scss:L197`
```scss
.#{$prefix}--snippet--multi .#{$prefix}--snippet-container:focus {
    @include focus-outline('outline');
    outline-offset: 0;
}
```

packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -226,10 +226,14 @@ function CodeSnippet({
     <div {...rest} className={codeSnippetClasses}>
       <div
         ref={codeContainerRef}
-        role={type === 'single' ? 'textbox' : null}
-        tabIndex={type === 'single' && !disabled ? 0 : null}
+        role={type === 'single' || type === 'multi' ? 'textbox' : null}
+        tabIndex={
+          type === 'single' || (type === 'multi' && !disabled) ? 0 : null
```
francinelucca: should this be the following instead? looks like the original condition for `single` is changing
```suggestion
          (type === 'single' || type === 'multi') && !disabled ? 0 : null
```

jsehull: It must have auto formatted it. I think if we use your suggestion it _won't_ next time. I'll work on getting that added!

francinelucca (approved):

tw15egan (approved):

PR #13311, opened 2023-03-08
feat(PasswordInput): add typescript types to PasswordInput

### LINKED ISSUE #12550: Add TypeScript types to PasswordInput

shryoo-ibm:
Closes #12550 

Part of #12513, Adding TypeScript types to carbon react components

#### Changelog

**Changed**

- Adds type declarations to the props of the PasswordInput component

#### Testing / Reviewing

Types can be tested by making a <PasswordInput /> component and trying out the different property values.


shryoo-ibm:

packages/react/src/components/TextInput/PasswordInput.tsx
```diff
@@ -308,7 +468,7 @@ PasswordInput.propTypes = {
    */
   size: FeatureFlags.enabled('enable-v11-release')
     ? PropTypes.oneOf(['sm', 'md', 'lg'])
-    : PropTypes.string,
+    : PropTypes.oneOf(['sm', 'md', 'lg', 'xl']),
```
tay1orjones: This feature flag will always `true` now that we've released v11. This toggle could be removed entirely. The type definition should be updated as well. `xl` is no longer a valid value.

tay1orjones (changes requested):
Just one suggestion, otherwise I think this looks good!

packages/react/src/components/TextInput/PasswordInput.tsx
shryoo-ibm: Got it, I was a bit confused about that so the clarification is a big help!

tay1orjones (approved):
Looks great, thanks! 🎉

tay1orjones:
@shryoo-ibm looks like there's a minor conflict to resolve - just regen the allcontributors probably

alisonjoseph (approved):

aledavila:
bump @shryoo-ibm looks like you have some minor conflicts to fix :) then we can review thanks!

tay1orjones (approved):
Looks good to me 👍 thanks!

PR #13320, opened 2023-03-09
fix(Modal): align button text to baseline

### LINKED ISSUE #13319: [Bug]: Modal button label wraps with incorrect alignment
### Package

@carbon/styles

### Browser

Chrome

### Package version

1.24.0

### React version

_No response_

### Description

When text inside the modal buttons are long and wraps to two lines, the alignment of text labels are thrown off. Currently seeing below. Please note this bug is happening on modal size `xs` and `sm` as well so the button labels are really not that long, so the error is obvious. 

image

Expecting button labels to be 16px from top of button:

image

### Reproduction/example

This is what the modal looks like in Partner Portal right now:

image

### Steps to reproduce

1. Go to: https://react.carbondesignsystem.com/?path=/story/components-modal--playground
2. Type in a longer text in any footer button label in inspect

### Suggested Severity

### Application/PAL

IBM Partner Portal / Carbon for Salesforce

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13319

Aligns the button text to the button baseline so the content will flow down instead of pushing up into the modal

#### Changelog

**Changed**

- set `align-items` to `baseline`

#### Testing / Reviewing

Ensure when extra text wraps, it does not cause misalignment at the top of the text string. 

cc @shixiedesign / @carbon-design-system/design  Note: This will cause all text in Modals to be pushed down ~1px to align them all to the baseline. Percy should catch this; let me know if that is an okay tradeoff


tay1orjones (approved):

tay1orjones:
@tw15egan Percy didn't catch it 🧐

tw15egan:
@tay1orjones hmm, thought it would; it seems to have caught more minor changes...

thyhmdo (approved):

PR #13328, opened 2023-03-10
fix(batchToolTips): make tooltips on batch action visible

AustinGitHub:
Closes # very minor fix and issue, didn't think a whole issue needed to be opened for this. If it does then I will open one upon request. This solves an issue on our IBM DataStage repo where we need tooltips for the batch action icons

Enables tooltips on batch action icons

#### Changelog

**New**

Enables tooltips on batch action icons

**Changed**

Tool tips to be visible when batch action items are in view. 

**Removed**

overflow 

#### Testing / Reviewing

Screenshot 2023-03-09 at 6 38 25 PM
Screenshot 2023-03-09 at 6 54 53 PM

Screenshot 2023-03-09 at 6 45 04 PM

Only issue I saw was with `tooltipPosition` the value `top` won't work because of how it clips. But I think this is ok because the `bottom` will be used more in usage. `left` and `right` will work just fine, I tested and didn't see issue. Thanks!

Batch actions with tooltips should be able to be visible now.

### DIFF
--- a/packages/react/src/components/DataTable/stories/DataTable-batch-actions.stories.js
+++ b/packages/react/src/components/DataTable/stories/DataTable-batch-actions.stories.js
@@ -7,7 +7,7 @@
 
 import React from 'react';
 import { action } from '@storybook/addon-actions';
-import { TrashCan, Save, Download } from '@carbon/icons-react';
+import { TrashCan, Save, Download, Add } from '@carbon/icons-react';
 
 import Button from '../../Button';
 import DataTable, {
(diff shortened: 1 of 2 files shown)


AustinGitHub:

packages/react/src/components/DataTable/stories/DataTable-batch-actions.stories.js
```diff
@@ -89,6 +89,18 @@ export const Default = () => (
...
+                tooltipPosition="bottom"
+                tabIndex={batchActionProps.shouldShowBatchActions ? 0 : -1}
+                renderIcon={Add}
+                onClick={batchActionClick(selectedRows)}>
+                Delete
+              </TableBatchAction>
+              <TableBatchAction
+                hasIconOnly
+                iconDescription="Save"
+                tooltipPosition="bottom"
```
AustinGitHub, line 103: this is to add another example of the tooltip in use

AustinGitHub:
Hi @alisonjoseph and @francinelucca when you get the chance can you review this today? Would like to get this in if possible, thanks!

AustinGitHub:
Screenshot 2023-03-10 at 9 00 06 AM

francinelucca (approved):
LGTM

alisonjoseph (approved):
LGTM

AustinGitHub:
Thank you both!

Issue #13014, opened 2023-01-24
Layer stories: Define a new design specification/template for "with layer" stories


tay1orjones:
- [x] Explore possible ways to arrange a story that best demonstrates the correct usage of layering.
- [x] Review with dev
- [x] Create a final spec for dev

REF https://github.com/carbon-design-system/carbon/pull/13331

PR #13331, opened 2023-03-10
docs: redesign with-layer stories

### LINKED ISSUE #13015: Layer stories: Update "with layer" stories' implementation
Update the implementation to use a template for `with layer` stories based on the outcome of https://github.com/carbon-design-system/carbon/issues/13014

janhassel:
Closes #13015 

Ref #13014

This is a design proposal for the "with layer" stories.

#### Changelog

**New**

- Added a template component `WithLayer` to be used in storybook only

#### Testing / Reviewing

The template component renders `props.children` in all three layers and places them onto the respective layer's background. A label with the layer name is added for clarification.

For simple components, the markup just requires a demo setup to be wrapped in `<WithLayer>`:
```jsx
<WithLayer>
  <ProgressBar label="Progress bar"  />
</WithLayer>
```

In case the demo requires an id or other unique attribute per layer (like for radio button groups), `props.children` can also be a function that will receive the layer index (0, 1, 2) and returns the child with all necessary props set:
```jsx
<WithLayer>
  {(layer) => (
    <ProgressBar id={`progress-bar-${layer}`} label="Progress bar" />
  )}
</WithLayer>
```


tay1orjones (approved):

aagonzales:
@janhassel Thank you for doing this! These needed so much help. Honestly, your design was such a dramatic improvement it would have been fine shipping as is. But we did do a design iteration pass at it and this is what we decided on. 

We thought use purple as an annotation color would help abstract the layer area from the component more. There are different colors for light vs dark themes but since we're not using the theme tokens, is it still possible to switch between colors? If not then we can try and find a fixed purple that will work across themes. 

image

image

aagonzales:
> What we can also do is use the tag tokens for the label text and background, but use a fixed color for the border. Perhaps purple-50 would work?

@janhassel I was just thinking about this solution as I was reading through your comments. I think that works nicely! 

---

I also have one other comment, the layer names as you have them are a little confusing because they are the same a the color token name but they don't actually line up. `$layer-01` would be used on the second layer as the background color and the first layer just uses `$background`. I think it might be clear is we just label them as "Layer 1" / "Layer 2" / "Layer 3" so they wont be confused with tokens.

janhassel:
@aagonzales Good point. I was thinking in terms of like "on layer-01", but I think changing it to something less token-looking is good. Should it still use `code-01` or do we maybe also wanna use sans-serif here to further differentiate from tokens?

aagonzales:
I wasn't surer about that one so I checked with the other designers and we think you can keep`code-01` to help it look more like an annotation and not part of the component.

aagonzales (approved):
This looks awesome! Thanks for doing this 🎉

alisonjoseph (approved):
This is awesome!

tw15egan:
Just wanted to chime in and say how awesome this is, @janhassel! Just used it in https://github.com/carbon-design-system/carbon/pull/13469 and it looks / works great 🔥 💯

janhassel:
@tw15egan Glad to hear! 🙂 Was there a problem with using the Accordion as a child directly without using the function? Hope I didn't miss anything.

```jsx
<WithLayer>
  <Accordion>
    ...
  </Accordion>
</WithLayer>
```

as opposed to

```jsx
<WithLayer>
  {() => (
    <Accordion>
      ...
    </Accordion>
  )}
</WithLayer>
```

PR #13350, opened 2023-03-15
refactor(Tooltip): convert Tooltip and DefinitionTooltip to TypeScript

### LINKED ISSUE #13231: Add TypeScript types to Tooltip

lewandom:
Closes #13231.

Add TypeScript typings to Tooltip and DefinitionTooltip components.

#### Changelog

**New**

- added TooltipProps and DefinitionTooltipProps interfaces

**Changed**

- added TypeScript types to Tooltip and DefinitionTooltip
- added TypeScript types to useDelayedState helper function to allow for generic types usage
- adjusted existing type of `useFallbackId` helper function parameter to allow undefined value
- adjusted existing type of `match` helper function parameter to allow React.KeyboardEvent value

#### Testing / Reviewing

The Tooltip component imports for the storybook should be valid and rendering. In a TypeScript environment, you should be able to import and reference the Tooltip component and its typings (TooltipProps).


packages/react/src/internal/useId.js
```diff
@@ -65,7 +65,7 @@ export function useId(prefix = 'id') {
 
 /**
  * Generate a unique id if a given `id` is not provided
- * @param {string} id
+ * @param {string|undefined} id
```
francinelucca, line 68: `useId` signature says it returns a string, under what circumstances would this return undefined?

lewandom: @francinelucca the change pertains to function `id` argument type, not the return type.

francinelucca: my bad! I read that wrong

francinelucca (approved):

aledavila (approved):

PR #13359, opened 2023-03-16
style(Listbox): remove background on active

### LINKED ISSUE #13120: [a11y]: MultiSelect selected items do not need a gray fill when it is redundant
Original guidance from @mbgower in https://github.com/carbon-design-system/carbon/issues/12723:

> ### Issue 2 Highlighting should only be on items with focus, not on selected items
> In the multi-select, items that are selected are both checked AND have a fill added. This is needlessly redundant AND causes difficulties navigating since the same fill is added to show the item with focus. A user cannot see _any_ evidence of focus moving when moving through multiple selected items. This is shown in the following screen snippet where the item with focus is option 2. This may be resolved if the issue on increasing contrast on the item with focus, https://github.com/carbon-design-system/carbon/issues/12720,  is addressed.
> image

We're considering this for the future, but it has a large design impact and will need exploration.

pratikkarad:
Closes #13120 

**Removed**

- Removed additional style

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1076,6 +1076,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "pratikkarad",
+      "name": "Pratik Karad",
+      "avatar_url": "https://avatars.githubusercontent.com/u/32093370?v=4",
+      "profile": "https://github.com/pratikkarad",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1083,7 +1083,8 @@
       "avatar_url": "https://avatars.githubusercontent.com/u/32093370?v=4",
       "profile": "https://github.com/pratikkarad",
       "contributions": [
-        "code"
+        "code",
+        "a11y"
       ]
     }
   ],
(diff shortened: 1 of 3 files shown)


pratikkarad:

tw15egan:
@pratikkarad, thanks for contributing! Looks like you linked the wrong issue, could you update that? Thanks!

pratikkarad:
Hi @tw15egan, updated the description. Thanks for pointing that out. 😄

tw15egan:
Seeing an issue with the selected items, I believe if we do not have a `selected` background, we would want to retain the border between items

Screenshot 2023-03-16 at 10 00 51 AM

pratikkarad:
Addressed the mentioned issue in the latest commit. PTAL, Thanks!

tw15egan:
Question for @laurenmrice: Should we use the same `hover` color for all items if we remove the selected background color? Right now, they differ



Also, I see a border when navigating via the keyboard on a selected item:

laurenmrice:
Hi @pratikkarad 👋 in the [issue](https://github.com/carbon-design-system/carbon/issues/13120) for this, we noted that the design team needs to do some visual explorations around this change first before development can start because it will have a large design impact, so this PR will have to be put on hold for now.

pratikkarad:
Sure @laurenmrice. Please update the issue or PR once you finalize the design.

tay1orjones:
Thanks for your help getting this one started! In aim to avoid PRs going stale, I'm going to close this out for now. Once we've got a design direction established we can reopen this and make the necessary updates.

PR #13360, opened 2023-03-16
fix(button): don't overwrite rest['aria-pressed'] when passed

### LINKED ISSUE #13357: [Bug]: Button props.isSelected doesn't apply aria-pressed to element
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.24.0

### React version

_No response_

### Description

When using the `<Button kind="ghost" hasIconOnly isSelected … />`, the correct classname (`cds--btn--selected`) is applied to the `<button>` element. However, it should also mark the element as `aria-pressed` which is not happening anymore.

Ref #7900

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-button--icon-button&args=isSelected:true;kind:ghost

### Steps to reproduce

1. Visit storybook link above
2. Inspect `<button>` element
3. `aria-pressed` is not present in the attributes

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

janhassel:
Closes #13357 

The issue is happening because the `<IconButton>` renders a `<Button>` underneath and already sets `aria-pressed`. The `<Button>` component then overwrites `rest['aria-pressed']` because the constraint of `hasIconOnly` is not met anymore and the objects are merged with the `...rest` before `...otherProps`.

#### Changelog

**Changed**

- Don't overwrite `rest['aria-pressed']` anymore when it is passed to `<Button>`

#### Testing / Reviewing

Follow reproduction steps from linked issue.

This solution has the drawback that if a consumer is wrongfully setting `aria-pressed`, the component won't correct it anymore but take that over its own logic. Please let me know what you think about that!


alisonjoseph (approved):
LGTM!

tay1orjones (approved):

PR #13362, opened 2023-03-16
fix(MultiSelect): remove enter key select

### LINKED ISSUE #13119: [a11y]: Revise MultiSelect keyboard selection behavior to not use `enter`
```[tasklist]
### Tasks
- [ ] The `Enter` key should no longer select items within MultiSelect, only `space` should select items
- [ ] Add tests to cover the behavior change
```

Original guidance from @mbgower in https://github.com/carbon-design-system/carbon/issues/12723:

> ### issue 1 Only Spacebar should be used to select checkboxes in multi-select
> 
> For a single-select dropdown, both Enter and Spacebar can be used to select the item. However, with multi-select it is recommended to only make Spacebar the selection mechanism. There are a number of reasons:
> 1. Spacebar is the traditional mechanism for selection. It is the key that toggles a checkbox, and the multi-select appears as a list of checkboxes.
> 2. It allows Enter to have the function of completing the selections and closing the dropdown. A user _can_ collapse the dropdown by tabbing to the next component, but it is recommended to use a typical 'do' command for this.
> 4. It helps prevent the user from accidentally submitting an incomplete form. There is a slight danger that in a form with a default Submit button, pressing Enter _may_ cause the form to be submitted.

pratikkarad:
Closes #13119 

**Removed**

- Removed additional `Enter` key from MultiSelect keyboard selection


prakarsh70:

prakarsh70 (approved):
LGTM.

tw15egan (changes requested):
It looks like there is a linting error (`MenuKeyDownEnter` can be removed if it is not used), and the tests need to be updated to account for the new behavior (can change the failing tests' truthiness)

pratikkarad:
Hey @tw15egan, I have corrected the lint error and test cases.
I had a question regarding running scoped test cases because it takes too much time to run the whole test suite.
Any suggestion on how do I run test suite only for a specific package?

tw15egan:
@pratikkarad yeah, running the whole test suite is time-consuming. You can run specific tests by running this command from the root of the monorepo: 

`yarn test packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js --watch`

If you use VS Code, you can also install the `Jest` extension, which lets you run specific tests inside the editor.

pratikkarad:
Thanks for the suggestion. I have updated the PR. PTAL, Thanks!

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -192,7 +192,9 @@ describe('MultiSelect', () => {
     });
     pressEnter();
 
-    expect(itemNode.getAttribute('data-contained-checkbox-state')).toBe('true');
+    expect(itemNode.getAttribute('data-contained-checkbox-state')).toBe(
+      'false'
+    );
```
tw15egan:
```suggestion
```

We can probably just remove this. I didn't see that the block below it is the same. We should also rename this test to say `should not toggle selection with enter`

tw15egan (approved):

tay1orjones (approved):

Issue #12528, opened 2022-11-07
Add TypeScript types to TableHeader


tay1orjones:

imp-dance:
Seems to me like this one can be closed / checked.

imp-dance:
> Seems to me like this one can be closed / checked.

Sorry for going on about this, but still seems like this is merged yet still open. Would help with the overview if it's closed to properly reflect it's state.

PR #13367, opened 2023-03-17
refactor(typescript): Convert TableHead, TableHeader to tsx

### LINKED ISSUE #12527: Add TypeScript types to TableHead

pratikkarad:
Closes #12527, #12528

Converts TableHead and TableHeader to tsx in preparation for the main Table conversion

#### Changelog

**Changed**

- Converted TableHead.js -> TableHead.tsx
- Converted TableHeader.js -> TableHeader.tsx

#### Testing / Reviewing

To verify these changes are working, we can check to see if the existing storybook behavior is still working correctly. Should be able to properly import converted components in a TS environment


packages/react/src/components/DataTable/TableHead.tsx
```diff
@@ -5,9 +5,12 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import React, { ThHTMLAttributes } from 'react';
 import wrapComponent from '../../tools/wrapComponent';
 
-const TableHead = wrapComponent({
+export type TableHeadProps = ThHTMLAttributes<HTMLTableCellElement>;
```
francinelucca: should this be an `HTMLTableSectionElement` instead? https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableSectionElement. Doesn't look like `thead` is considered an HTMLTableCellElement https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableCellElement
```suggestion
export type TableHeadProps = ThHTMLAttributes<HTMLTableSectionElement>;
```

francinelucca (changes requested):
Thanks for putting this together @pratikkarad! Just one change request regarding the `thead` interface. also looks like some of the CI checks are failing

packages/react/src/components/DataTable/TableHead.tsx
pratikkarad: Thanks for pointing it out. I have updated the code.

packages/react/src/components/DataTable/TableHeader.tsx
```diff
@@ -91,7 +158,7 @@ const TableHeader = React.forwardRef(function TableHeader(
     [`${prefix}--table-sort--descending`]:
       isSortHeader && sortDirection === sortStates.DESC,
   });
-  const ariaSort = !isSortHeader ? 'none' : sortDirections[sortDirection];
+  const ariaSort = !isSortHeader ? 'none' : sortDirection ? sortDirections[sortDirection] : sortDirections[sortStates.NONE];
```
pratikkarad: I observed It is still failing here. I tried debugging but got no luck. Any help or suggestion is appreciated. 😃

francinelucca: let's change this line to ` const ariaSort = (!isSortHeader || !sortDirection) ? 'none' : sortDirections[sortDirection];`

and change the type of `sortDirections` in the top to 
```
const sortDirections: {[key: string]: 'none' | 'ascending' | 'descending'} = {
  [sortStates.NONE]: 'none',
  [sortStates.ASC]: 'ascending',
  [sortStates.DESC]: 'descending',
};
```


That should fix this issue

pratikkarad (commented):
Hi @francinelucca, I'm facing a problem with declaring the static properties on the functional component in typescript.
For eg.
`TableHeader.translationKeys = Object.values(translationKeys);`
In TableHeader, `translationKeys` is giving the compilation error.
I tried creating a separate type as suggested on StackOverflow/blogs, but still got no luck.

francinelucca:
> Hi @francinelucca, I'm facing a problem with declaring the static properties on the functional component in typescript. For eg. `TableHeader.translationKeys = Object.values(translationKeys);` In TableHeader, `translationKeys` is giving the compilation error. I tried creating a separate type as suggested on StackOverflow/blogs, but still got no luck.

can you try doing `(TableHeader as any).translationKeys = Object.values(translationKeys);`

francinelucca (approved):

aledavila (approved):

PR #13375, opened 2023-03-20
feat(HeaderMenuItem): update prop

### LINKED ISSUE #13369: [Feature Request]: Update HeaderMenuItem to use `isActive`, like SideNavMenuItem
### The problem

HeaderMenuItem uses `isCurrentPage` to trigger the "selected" state for the component, whereas in `SideNavMenuItem` the prop `isActive` is used. This is sort of confusing to have different named props doing functionally similar things, all within the UIShell.

### The solution

Updating this component to use the same prop would be a nice unification of the component APIs.

```[tasklist]
- [ ] Add `isActive` to HeaderMenuItem
- [ ] Infer it's value based on `isCurrentPage` (if provided)
- [ ] Deprecate `isCurrentPage`
- [ ] Remove `isCurrentPage` in v12
```

### Examples

https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1679052333819769

pratikkarad:
Closes #13369 

#### Changelog

**Changed**

- Added `isActive` prop
- Deprecated `isCurrentPage` in HeaderMenuItem

#### Testing / Reviewing

To verify these changes are working, we can check to see if the existing storybook behavior is still working correctly.


packages/react/src/components/UIShell/HeaderMenuItem.js
```diff
@@ -66,7 +70,13 @@ HeaderMenuItem.propTypes = {
...
 
   /**
-   * Applies selected styles to the item if a user sets this to true and aria-current !== 'page'.
+   * Applies selected styles to the item if a user sets this to true and `aria-current !== 'page'`.
+   */
+  isActive: PropTypes.bool,
+
+  /**
+   * Applies selected styles to the item if a user sets this to true and `aria-current !== 'page'`.
+   * @deprecated Please use `isActive`. This will be removed soon in v12.
```
tay1orjones:
```suggestion
   * @deprecated Please use `isActive` instead. This will be removed in the next major release.
```

packages/react/src/components/UIShell/HeaderMenuItem.js
```diff
@@ -66,7 +70,13 @@ HeaderMenuItem.propTypes = {
...
    */
   isCurrentPage: PropTypes.bool,
```
tay1orjones: This needs the `deprecate()` prop types function helper applied here.

Here's an example of how it's used: https://github.com/carbon-design-system/carbon/blob/b0f39f3d34152057c1bd3f972ecd24e0c3269982/packages/react/src/components/ComboBox/ComboBox.tsx#L736-L740

tay1orjones (changes requested):
Thanks for this! Just a couple minor things to address

pratikkarad:
Hey @tay1orjones, I have updated the PR. PTAL, Thanks! 😃

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -189,12 +201,15 @@ class HeaderMenu extends React.Component {
       [`${prefix}--header__submenu`]: true,
       [customClassName]: !!customClassName,
     });
+    if (isCurrentPage) {
+      isActive = isCurrentPage;
+    }
```
tw15egan:
```suggestion
let isActivePage = isActive ? isActive : isCurrentPage;
```

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -189,12 +201,15 @@ class HeaderMenu extends React.Component {
...
     const linkClassName = cx({
       [`${prefix}--header__menu-item`]: true,
       [`${prefix}--header__menu-title`]: true,
-      // We set the current class only if `isCurrentPage` is passed in and we do
+      // We set the current class only if `isActive` is passed in and we do
       // not have an `aria-current="page"` set for the breadcrumb item
-      [`${prefix}--header__menu-item--current`]: isCurrentPage,
+      [`${prefix}--header__menu-item--current`]: isActive,
```
tw15egan:
```suggestion
      [`${prefix}--header__menu-item--current`]: isActivePage,
```

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -180,6 +191,7 @@ class HeaderMenu extends React.Component {
       focusRef, // eslint-disable-line no-unused-vars
       ...rest
     } = this.props;
+    let { isActive } = this.props;
```
tw15egan:
```suggestion
```

tw15egan: I would add `isActive` above `isCurrentPage` in the above section, so it is not spread onto the component via the `...rest` spread

pratikkarad:
Hi @tay1orjones, as suggested by @tw15egan, I tried doing the same thing to `HeaderMenu` and updated the test cases but it's still failing for some reason. Any suggestion?

tw15egan (changes requested):

packages/react/src/components/UIShell/HeaderMenu.js
pratikkarad: The above props and the `isCurrentPage` prop are being destructured as `const`, so we won't be able to infer the value to `isActive` from `isCurrentPage` as we cannot re-assign value to const.
As we will remove the `isCurrentPage` in the next major release, then only we can import it along with other props. I guess, we will need this let variable till then.

tw15egan: You need to destructure them both, and then assign them to a new `let` further down: `let isActivePage = isActive ? isActive : isCurrentPage;` and then use `isActivePage` handle all the `isActive/isCurrentPage` functionality. Otherwise `isActive` will be spread onto the `HeaderMenu` and cause the test failure that you are seeing.

pratikkarad: Sorry. I didn't read all the comments 😅. Will work on the changes.

tw15egan (approved):
LGTM 👍 ✅

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -37,9 +38,19 @@ class HeaderMenu extends React.Component {
...
-    isCurrentPage: PropTypes.bool,
+    isActive: PropTypes.bool,
+
+    /**
+     * Applies selected styles to the item if a user sets this to true and `aria-current !== 'page'`.
+     * @deprecated Please use `isActive` instead. This will be removed in the next major release.
+     */
+    isCurrentPage: deprecate(
+      PropTypes.bool,
+      'The `isCurrentPage` prop for `HeaderMenuItem` has ' +
```
tay1orjones:
```suggestion
      'The `isCurrentPage` prop for `HeaderMenu` has ' +
```

tay1orjones (approved):

PR #13378, opened 2023-03-20
feat(ContentSwitcher): add icon-only ContentSwitcher

### LINKED ISSUE #12811: Icon content switcher: Make a POC
This issue is for making a proof of concept to experiment with the existing a11y makeup of ContentSwitcher

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/12811

Adds in a new icon-only `ContentSwitcher`

#### Changelog

**New**

- `IconSwitch` component was created, and rendered as a child of `ContentSwitcher`. When this is used instead of `Switch`, it will render an `IconButton`.

```jsx
<ContentSwitcher>
    <IconSwitch name="one" text="Table of Contents">
      <TableOfContents />
    </IconSwitch>
    <IconSwitch name="two" text="Workspace Test">
      <Workspace />
    </IconSwitch>
    <IconSwitch name="three" text="View Mode">
      <ViewMode_2 />
    </IconSwitch>
  </ContentSwitcher>
```

- The `name` passed into `IconSwitch` will render the `Tooltip` text. This will also set the `aria-label` on the `IconButton`
- Tests and snapshots added for the new `IconSwitch`
- Styles for the new icon-only `ContentSwitcher`

**Changed**

- `ContentSwitcher` now passes along `size` to its children.
- Updated the `size` prop types to valid v11 sizes.
- Adjusted styles for the new icon-only `ContentSwitcher`
- Added a `wrapperClasses` prop to `IconButton`. I needed a way to surface selected + disabled states to the outermost wrapper, and adding one in `IconSwitch` was causing the `role="tab"` be too far removed from its parent, and causing screenreader issues (Each button read `Tab 1 of 1`). Since it is now just 1 level removed, the screen reader announces the number of tabs correctly. 

**Removed**

- Removed the `with layer` snapshot and replaced it with the new icon-only story. The `with layer` story has no visual differences from the default story. 

#### Testing / Reviewing

Go to `ContentSwitcher` and check out the icon-only story. 
- Ensure it matches the spec, and tooltips appear. 
- Ensure sizes are able to change, and supports both selection modes. 
- Try disabling an icon button and ensure it works. 

Make sure there are no regressions to the standard `ContentSwitcher`


packages/react/src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -141,25 +141,44 @@ export default class ContentSwitcher extends React.Component {
...
+              onClick: composeEventHandlers([
+                this.handleChildChange,
+                child.props.onClick,
+              ]),
+              onKeyDown: this.handleChildChange,
+              selected: index === this.state.selectedIndex,
+              ref: this.handleItemRef(index),
+            });
+          }
+        })}
```
tw15egan: @tay1orjones Do you think this is necessary? The only reason I've done this is so we don't pass `size` down to the standard `Switch` component, but I do not see any issues by passing it down, and it makes it harder to read. It ensures no extra props are sent down, but I think `size` is the only prop that needs to be sent down. The rest can be passed directly to `IconSwitch`.

tay1orjones: @tw15egan If `size` is passed down to `Switch`, it'll get lumped into `...other` which is spread onto a `button`. Would that generate one of those "invalid attribute on dom element" warnings?

tw15egan: I didn't see any, but that may be because it isn't camel-cased?

tay1orjones: Yeah I bet you're right. That, and `size` _is_ a valid dom attribute (albeit only for `select` and `input`)

packages/react/src/components/ContentSwitcher/ContentSwitcher.js
```diff
@@ -141,9 +140,14 @@ export default class ContentSwitcher extends React.Component {
       ...other
     } = this.props;
 
+    const isIconOnly = React.Children.map(children, (child) => {
+      return child.type.displayName === 'IconSwitch';
```
tay1orjones, line 144: Just want to make a note here that this locks the API so consumers can only use IconSwitch. It could prove problematic if someone were to to roll their own IconSwitch component and names it something different. Let's cross that bridge if we come to it though - no need to over optimize for it right now imo

tay1orjones (approved):
Looks great! 🎉 🏆

laurenmrice (changes requested):
Looking really good!! Just a couple of comments.

---

- I noticed there is a slight shadow or gray color bug on the left side of the first section when its selected



- When the last section is selected and has focus, the focus border does not completely overlap the selected color layer on the right side. 



- The icon color in the sections that are unselected should be `$icon-primary`.


- The Large size content switcher should have 20px icons instead of 16px. 
Screen Shot 2023-03-30 at 3 07 09 PM

packages/styles/scss/components/content-switcher/_content-switcher.scss
```diff
@@ -218,4 +231,93 @@
...
+    border-top-left-radius: rem(4px);
+  }
+
+  .#{$prefix}--content-switcher--icon-only
+    .#{$prefix}--content-switcher-popover__wrapper:first-child
+    .#{$prefix}--content-switcher--selected,
+  .#{$prefix}--content-switcher--icon-only
+    .#{$prefix}--content-switcher-popover__wrapper:last-child
+    .#{$prefix}--content-switcher--selected {
+    border-color: $layer-selected-disabled;
```
aledavila, line 257: This is adding the grey border that lauren was mentioning.

aledavila (approved):
LGTM pending lauren's style issue

laurenmrice (changes requested):
Everything looks good except for the focus state in Safari only. The inset padding becomes inconsistent in some content switcher sections in the different sizes. 
Apr-04-2023 17-28-14

laurenmrice (approved):
Looks great to me! Thanks TJ! ✅

PR #13381, opened 2023-03-21
fix(RadioTile): positioning bug when grid aligned

### LINKED ISSUE #13046: [Bug]: TileGroup scrolls to top when RadioTile is clicked in grid layout
### Package

@carbon/react

### Browser

Chrome

### Package version

@carbon/react@1.21.0

### React version

react@17.0.2

### Description

When using `display: grid` on the wrapper element around `RadioTile`s, the scroll position is reset / set to top every time the user selects an option.

See reproduction link for detailed implementation.

### Suggested Severity

### Reproduction/example

https://stackblitz.com/edit/github-jml6ye?file=src%2FApp.jsx,package-lock.json

### Steps to reproduce

https://user-images.githubusercontent.com/28265588/215439371-021628f8-36eb-4158-aebb-6d71b1a89f54.mov

### Code of Conduct

alisonjoseph:
Closes #13046

#### Changelog

**New**

- add wrapper div around each radio tile input/label

#### Testing / Reviewing

Ensure that nothing changed with RadioTiles, check demo story to make sure it doesn't jump to top when you select tiles.
https://deploy-preview-13381--carbon-components-react.netlify.app/?path=/story/components-tile--radio-grid-demo

**DON'T MERGE UNTIL DEMO FILES ARE DELETED**


packages/react/src/components/RadioTile/RadioTile.js
```diff
@@ -56,7 +56,7 @@ const RadioTile = React.forwardRef(function RadioTile(
   }
 
   return (
-    <>
+    <div>
```
tay1orjones, line 59: Adding this wrapper is causing the tests to fail since the tests use `container.firstChild` to assert if the tile is checked or not. Could probably just be updated to `screen.getByRole('radio')` instead I think?

tay1orjones (approved):
Looks good, just a couple tests failing though

packages/react/src/components/RadioTile/RadioTile.js
alisonjoseph: Somehow I forgot about this PR, I'll get it updated. Thanks @tay1orjones

francinelucca (commented):
confirm no scroll on bottom tiles, if I scroll all the way down and click on the mid tiles I still get scrolled up, is that expected?

https://user-images.githubusercontent.com/40550942/231249689-bc96b8f0-5eba-4b7e-a86e-16f11c2b8e9b.mov

alisonjoseph:
@francinelucca I think its just scrolling up to the one you clicked on correct? If so, that's expected.

francinelucca (approved):

Issue #13013, opened 2023-01-24
Disabled tiles: Provide spec of disabled styling for all Tile components, and for experimental tile


tay1orjones:
Disabled state design spec for each tile variant

#### Acceptance Criteria

- spec for old tile
- spec for experimental tile

PR #13382, opened 2023-03-21
feat(tiles): add disabled story and update styles

### LINKED ISSUE #13198: Disabled tiles: Implementation
#### Acceptance Criteria

- [x] Add `disabled` prop to all tile variants that don't have it
- [x] Add disabled styles for components that don't have it already (Expandable Tile)
- [x] Add new disabled styles that are properly scoped to the experimental tile variant
- [x] Add `disabled` props to storybook controls
- [x] Add a new story to showcase disabled styling/state
- [ ] Cover the disabled state story in VRT/Percy
- [x] Add unit tests to assert that the tiles are `.toBeDisabled()`

aledavila:
Closes #13198
Design ref: #13013

- Add icons and new `renderIcon` prop for custom icon to Clickable tile
- fixes some disabled styles and add disabled controls to stories. 
- Add error icon to Clickable Tile on disabled


francinelucca (changes requested):
is the `disabled`  control supposed to be in every story? I don't see it in `Default`, `Expandable`, `Multiselect` or any of the `with xxxx` variants.

I also don't see a story showcasing the disabled state 😅

aledavila:
@francinelucca No it's not. They don't have disabled states. Let me also link the design. 

It's only Clickable, Selectable, and Radio. I wasn't sure if I should add specific disabled stories because it would add 3 more stories to the Tiles (plus we don't have prop specific stories in most of our components). It seemed overcrowded. I can add them in each individual story instead of adding the controls though. I'm not sure what's better

francinelucca:
@aledavila you're right, probably don't need to add 3 more stories. I didn't realize the controls at the bottom where interact-able for these ones 🤦🏽‍♀️. Should multiselect have it as well though? looks like there's a spec for it

alisonjoseph (approved):
Tile, Clickable, Radio and Expandable tiles and improved contrast tiles disabled styles all look good!

francinelucca (approved):
🚀

thyhmdo:
Chatted with Ale. She's going to add a default icon (arrow-right) for the other states, rather than disabled. Here is the specs
 
image_720

aledavila:
Hello @francinelucca @alisonjoseph I added a new prop and added an icon to the Clickable tile per my convo with Thy if you don't mind reviewing again! Thank you

francinelucca (approved):
🚀

thyhmdo (approved):
looks good! thanks

alisonjoseph (approved):

alisonjoseph (dismissed):
@aledavila @thyhmdo @francinelucca 

This looks great, but on second thought I'm wondering if this is a breaking change to drop an icon in by default? I'm worried users might be adding their own icons already and/or have content placed in the lower right? Should this be opt-in behavior for v11, and default for v12? Thoughts?

francinelucca:
> @aledavila @thyhmdo @francinelucca
> 
> This looks great, but on second thought I'm wondering if this is a breaking change to drop an icon in by default? I'm worried users might be adding their own icons already and/or have content placed in the lower right? Should this be opt-in behavior for v11, and default for v12? Thoughts?

I was trying to find something relating to this in the versioning doc, it says adding a new prop should be a minor https://github.com/carbon-design-system/carbon/blob/main/docs/guides/versioning.md#a-new-prop-is-added-to-a-component but I do agree maybe the default icon is troublesome here 🤔. I'd side with Alison to make the default an opt-in in v11

thyhmdo:
@alisonjoseph @francinelucca I see. Maybe I don't understand that enough. I thought once users use our components, it means that it replaces the current behaviors/info. But it's a requirement for enhancing the accessibility and we do call it out in [the doc](https://carbondesignsystem.com/components/tile/usage/#alignment). It only applies to Clickable tiles. So maybe we can just have this for the tiles under the Experimental/Improved contrast tile and not for the current tiles.

alisonjoseph:
@thyhmdo Correct, it would replace what they have, however we don't control what they place inside the tile, so having an icon pop up where they might have already placed their own icon, pictogram, text, button etc could be problematic. 

Adding it to experimental only for now could work or we add it as an opt-in behavior for the regular tile.

thyhmdo:
@alisonjoseph I like that. Thank you! 

@aledavila hey Ale, seems like there's another update with Alison and Francine's suggestion. 

We'll add the new behavior for Clickable experimental tiles only for now (which you already did).  And we could add it as an opt-in behavior for the current clickable tile.

tay1orjones:
@thyhmdo @alisonjoseph @francinelucca I'll take this on and finish it out.

The experimental disabled tile styling is not a separate component and is just additional styling behind a feature flag. So we can't add things to one and remove it from the other - everything is technically an opt-in on the existing tiles.

I'll look at placing the default icon behavior behind a new feature flag and updating all the stories to make sense alongside the experimental disabled flag stories.

alisonjoseph:
Styles seem off for Clickable with custom icon story, maybe the icon styles don't need to be behind the feature flag?

https://deploy-preview-13382--carbon-components-react.netlify.app/?path=/story/components-tile--clickable-with-custom-icon

Screenshot 2023-04-17 at 2 28 33 PM

tay1orjones:
@alisonjoseph good catch, thanks! The styles are no longer behind a flag and should be fixed.

I also updated the .mdx docs for the two new sections in storybook to properly reflect how to configure the feature flags

francinelucca (dismissed):
LGTM!! I like the feature flags section, my only concern would be maybe it hides stuff a bit?
question: is it possible to test that a default icon is added when the feature flag is turned on and vice versa? if so, should we?
Also should the new stories be added into VRT?

francinelucca:
Also looks like the Clickable Tile (default) story is now less wide than prod, is that intended?
image
image

thyhmdo:
Same thought with Francine. It's very confusing to look for the Experimental tile. Is that common for devs to understand the term "Featured Flags"? We have stable tiles vs experimental tiles but hidden under the Featured Flags. 
@tay1orjones is this what you meant it's impossible to have them separated so Clickable tiles have to have its own folder within tiles? 

Is it possible to simplify this one like this?

image

tay1orjones:
You all bring up good points, great feedback - thank you!

@francinelucca I do think there's a way we could test this, I'll see what I can put together.

@thyhmdo Like we discussed today, I think having a location within the storybook that _all_ feature flags are turned on makes sense. That way we can effectively have integration test stories of components with all the flags turned on at the same time. Additionally we can organize stories under a single component folder, to avoid the duplication. We can document the feature flags in their own docs-only stories as well.


High level hitlist for my next rev on this is:
- [ ] Add tests for default icon feature flag
- [ ] Include experimental/feature-flag stories in VRT
- [ ] Document feature flags in their own docs-only stories
- [ ] Put all experimental tile stories under a single folder `Feature Flags/Tile
- [ ] Revert the width on the prod stories

janhassel:
coming over from https://github.com/carbon-design-system/carbon/pull/13678#issuecomment-1536207571

@tay1orjones I like this approach and the documentation of the feature flags together with how to enable them in the individual packages. The only thing I'd point out is that I had a hard time to understand which feature flags are enabled now in the "Feature Flags / Tile" stories and only got clarification through your comment here.
I wonder if this can be improved.

My first take would be to make the feature flags controllable through the story decorators. Unfortunately the toolbar doesn't allow for multi-select without any plugins.
image

Alternatively, the bottom actions panel could be used but might bring in confusion as to why some global types are in the toolbar and some in the actions.

Or, without making the feature flags user-controllable, an annotation would be helpful.

This could potentially follow the style of the Layer annotations with a different color and icon:
image

I just hope it doesn't get too busy with stacking annotations then.
image

The advantage with this is the consistency in story annotations though in case there comes up a need for another one, like deprecation for example?

image

Each annotation kind could be identified with a unique icon + unique color (from tag component tokens for out-of-the-box theme support).

Anyways, just wanted to throw this in here, let me know if that makes sense!

tay1orjones:
@janhassel I just pulled in the changes from your branch for the annotations on the stories - thank you again!

@alisonjoseph @francinelucca This is ready for review 👍

alisonjoseph (approved):
LGTM! 🔥 Love the new feature flags docs

francinelucca (approved):
🔥🔥🔥

PR #13385, opened 2023-03-21
fix(UIShell): add bg color to selected states

### LINKED ISSUE #10132: [Bug]: update selected menu item in UI Shell
### Package

carbon-components

### Browser

Chrome

### Package version

current

### Description

Need to update the selected state of in the header menu in the UI Shell. 
From slack comment: https://ibm-studios.slack.com/archives/C046Y0YUD/p1636956609126800

### Screenshots

As-is state:
image

To be state:
image

### Steps to reproduce

- Go to ui shell + header base + navigation 
- Open menu
- Set an item to selected

### Code of Conduct

jsehull:
Closes #10132 

Add desired background colors for **selected** menu-items

@aagonzales helped confirm the story above just needs proper background color to bring it up to spec. Let me know if there are any other design requirements!

#### Changelog

**New**

- add background colors for both selected states ("selected" and "selected + hover")

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing
1. open `UIShell.stories.js` and find `HeaderBaseWNavigation`
2. add `isCurrentPage` to a `HeaderMenuItem` under both `HeaderNavigation`(desktop) and `SideNav`(mobile) 
image
3. visit Storybook > UI Shell > Header Base w/Navigation to confirm updates for desktop and mobile now display as designed

Please reference the issue for design details.

### DIFF
--- a/packages/styles/scss/components/ui-shell/header/_header.scss
+++ b/packages/styles/scss/components/ui-shell/header/_header.scss
@@ -254,12 +254,11 @@
   a.#{$prefix}--header__menu-item[aria-current='page']::after,
   .#{$prefix}--header__menu-item--current::after {
     position: absolute;
-    top: 0;
-    right: 0;
     bottom: -2px;
-    left: 0;
-    width: 100%;
-    border-bottom: 3px solid $border-interactive;
+    left: -2px;
+    width: calc(100% + 4px);
+    height: 3px;
+    background-color: $border-interactive;
     content: '';
   }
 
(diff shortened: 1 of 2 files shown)


aagonzales:
@jsehull My dev skills aren't good enough to test the change locally. Could you post some screenshots of the selected state? The token look right in code, I just want to double-check the visuals

aagonzales:
@jsehull  Thank you! Just one thing, it doesn't look like the blue vertical bar is extending the full height of the container... I think this was probably an existing bug, but can you fix it?

image

jsehull:
👍 I can have a fix on the SUBMENU items by removing the border for submenu items ONLY (which is `2px` all around).

But now I'm seeing the same is true for the MENU items (both horizontal and vertical layouts). 
image
image

@aagonzales Is this an issue as well? If so, do we need borders on these? I'm not familiar with all of the downstream impact on states or other IBM implementation this would have. Maybe some developer input is good too.

Options
- static changes to absolute positioning/width if needed
- dynamic - remove border so blue line is always 100% length

If this is simple, I guess we can add it on. If it requires a longer discussion, I'd be glad to help put in a PR on a new issue. Let me know what you'd like!

tw15egan:
You can also fake it with the pseudo element, I just bumped up the width to `calc(100% + 4px)` to account for the `2px` focus border on each side, and set `left` to `-2px` to set it equal to the transparent border. 

Screenshot 2023-03-22 at 2 36 26 PM


The same can probably be done with the vertical `current` indicator, just adjusting the `height` and the `top` values

jsehull:
They should be full length "border highlights" now @aagonzales 

image

image

aagonzales (approved):
Looks good to me! Thanks for working on this!

aagonzales (changes requested):
Oh wait, I just saw the other comments and doubled check some things. It looks like the selected state is not remaining when the shell responds down the to left nav. 

The selected state in left nav menu should also still have the vertical highlight bar and not a horizontal one. 

Mar-23-2023 11-22-04

jsehull:
Thanks for clarifying, I'll get that sorted out

jsehull:
Sorry for the confusion team, this PR continues over at #13402

PR #13390, opened 2023-03-22
fix(ExpandableTile): support forward ref

### LINKED ISSUE #13340: [Bug]: ExpandableTile does not forward a ref
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.17.0

### React version

v18.2.0

### Description

 Looks like ExpandableTile, and maybe Tile too, does/do not forward a ref:
 
 I set ref={ref}, on ExpandableTile (also tried on Tile) but ref.current is still "null" 

### Reproduction/example

https://stackblitz.com/edit/github-5hp6sj?file=src/App.jsx

### Steps to reproduce

1. get a ref with React useRef: const ref=useRef(null)
2. set Component's ref like ref={ref}
3. debug and check ref.current (i.e in a button handler function)
4. ref.current is always null

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

pratikkarad:
Closes #13340 

#### Changelog

**Changed**

- Used existing ref to support the forward ref


packages/react/src/components/Tile/Tile.js
```diff
@@ -360,7 +363,7 @@ export function ExpandableTile({
   const aboveTheFold = useRef(null);
   const belowTheFold = useRef(null);
   const tileContent = useRef(null);
-  const tile = useRef(null);
+  const tile = useRef(ref);
```
tay1orjones: Unfortunately this won't work. The refs need to be merged - we have a helper function hook you can use for it

```
const tile = useRef(null);
const ref = useMergedRefs([forwardRef, tile]);
```

What I would recommend is this:
- Rename the forward ref `ref` to `forwardRef`
- Use the above snippet
- Apply `ref` to the DOM element instead of `tile` - `ref={ref}`

tay1orjones (dismissed):

packages/react/src/components/Tile/Tile.js
tay1orjones: You can see how we use this elsewhere in other situations by searching the codebase for `useMergedRefs`

francinelucca (approved):
LGTM

aledavila (approved):

aledavila:
@tay1orjones oh you have changes requested hold on this

PR #13392, opened 2023-03-23
feat: add eslint-plugin-jest-dom

### LINKED ISSUE #13312: Add the `eslint-plugin-jest-dom` package to our eslint config
This plugin can help solidify our testing practices by ensuring we use jest-dom matchers where possible.

```[tasklist]
- [ ] Add the dependency and configure it with the recommended defaults
- [ ] Fix any errors that result from new rules
- [ ] If there are a ton of errors, we can turn off problematic rules and make sub-issues to tackle those rules one by one to fix errors and turn the rules on.
```

falsepopsky:
Closes #13312

I added the dependency `eslint-plugin-jest-dom` and enabled the recommended configuration by default. I also fixed a ton of errors resulting from this change.

#### Changelog

**New**

- add eslint-plugin-jest-dom 

**Changed**

- fixed eslint errors in tests

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Make sure there are no errors when running lint, and that the tests pass successfully.


falsepopsky:
trackings errors from tests:

https://github.com/carbon-design-system/carbon/blob/394debed5021de7cdc0af22ed3e79ba982eee4e1/packages/react/src/components/RadioTile/RadioTile-test.js#L87



```
FAIL packages/react/src/components/RadioTile/RadioTile-test.js
  ● RadioTile › renders as expected - Component API › should respect value prop

    input with type=checkbox or type=radio cannot be used with .toHaveValue(). Use .toBeChecked() for type=checkbox or .toHaveFormValues() instead
    
      85 |       render(<RadioTile value="standard" />);
      86 |
    > 87 |       expect(screen.getByRole('radio')).toHaveValue('standard');
         |                                         ^

```

https://github.com/carbon-design-system/carbon/blob/394debed5021de7cdc0af22ed3e79ba982eee4e1/packages/icon-build-helpers/src/builders/react/components/__tests__/Icon-test.js#L72

```
FAIL packages/icon-build-helpers/src/builders/react/components/__tests__/Icon-test.js
  ● Icon › should be focusable if an aria label and tab index is used

    expect(received).toBeInTheDocument()

    received value must be an HTMLElement or an SVGElement.
    Received has type:  string
    Received has value: "Mock icon"

      70 |     );
      71 |
    > 72 |     expect(getContainer().getAttribute('aria-label')).toBeInTheDocument();
         |                                                       ^
```

[Summary of all failing tests](https://github.com/carbon-design-system/carbon/actions/runs/4495781469/jobs/7909782469#step:8:389)

config/eslint-config-carbon/plugins/jest.js
```diff
@@ -8,7 +8,8 @@
 'use strict';
 
 module.exports = {
-  plugins: ['eslint-plugin-jest'],
+  plugins: ['eslint-plugin-jest', 'jest-dom'],
+  extends: ['plugin:jest-dom/recommended'],
   overrides: [
     {
```
falsepopsky: Maybe move extends to overrides scope...

```
  overrides: [
    {
      extends: ['plugin:jest-dom/recommended']
      files: ['*-test.js', '*.test.js', '*-spec.js', '*.spec.js'],

```

falsepopsky:
Fixes:

### RadioTile

- why fails:

  [tohavevalue](https://github.com/testing-library/jest-dom#tohavevalue)
  > It accepts input, select and textarea elements with the exception of `<input type="checkbox">` and `<input type="radio">`, which can be meaningfully matched only using [toBeChecked](https://github.com/testing-library/jest-dom#tobechecked) or [toHaveFormValues](https://github.com/testing-library/jest-dom#tohaveformvalues).

- doesn't work w/:

  [toHaveFormValues](https://github.com/testing-library/jest-dom#tohaveformvalues)
  >It is important to stress that this matcher can only be invoked on a [form](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement) or a [fieldset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFieldSetElement) element.

   toBeChecked() ???

-  Fix test pass:
  [ByDisplayValue](https://testing-library.com/docs/queries/bydisplayvalue/)
expect(screen.getByDisplayValue('standard')).toBeInTheDocument()


### Icon

wip

config/eslint-config-carbon/plugins/jest.js
tay1orjones: Yeah I think moving the `extends` to the `overrides` scope is a good idea so this recommended rule set is scoped to files only with those prefixes.

tay1orjones (changes requested):
Hey this looks really fantastic! Thanks for fixing all the associated errors. 

I do think your suggestion is a good idea - let's move that, but otherwise I think this looks good to go since CI is passing 🎉

sstrubberg:
Hey @falsepopsky! thanks for the contribution. Curious if you are an IBMer or if you are using Carbon externally for a project. Let us know!

tay1orjones (approved):

falsepopsky:
@sstrubberg 

Hi!. I'm not an IBMer but I have used Carbon in the past. I mostly use icons or pictograms package. Have a great day!

aledavila (approved):

PR #13401, opened 2023-03-24
fix(radiobutton): shows helper text

aledavila:
Helper text wasn't showing up in Radio Group


aledavila:
Hey @francinelucca here's the fix for the helper text. You can lose out your PR if need be I added the helper id that you added in your PR if this is blocking you and we can work off of this. If you are getting an error it may also be cause the helper text is outside of the fieldset. I have a fix for that as well locally that I can push up instead.

francinelucca (approved):
LGTM!

andreancardona (approved):
looks good!

tay1orjones:
@aledavila there's a conflict on this one

aledavila:
updated @tay1orjones

PR #13402, opened 2023-03-24
fix(UIShell): add bg color to selected states (NEW)

### LINKED ISSUE #10132: [Bug]: update selected menu item in UI Shell

jsehull:
Closes #10132 

Continuation of closed PR (fork error 😬 ): #13385
Please reference this PR for comments and progress

THIS PR finishes CSS details for `HeaderNavigation` and `SubMenu` `selected` states
___

@aagonzales Example of full height border for SideNav in latest fix request 
image

#### Changelog

**New**

- add background colors for both selected states ("selected" and "selected + hover")
- ensured full height pseudo-borders (previously stopped short at border width)

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing
3. visit Storybook > UI Shell > Header Base w/Navigation to confirm updates for desktop and mobile now display as designed

Please reference the above issues for design details/comments.

___

### DIFF
--- a/packages/react/src/components/UIShell/HeaderMenu.js
+++ b/packages/react/src/components/UIShell/HeaderMenu.js
@@ -193,6 +193,10 @@ class HeaderMenu extends React.Component {
       ...rest
     } = this.props;
 
+    const hasActiveChildren =
+      React.Children.map(children, (child) => {
+        return child.props.isActive || child.props.isCurrentPage;
+      }).length > 0;
     const accessibilityLabel = {
       'aria-label': ariaLabel,
       'aria-labelledby': ariaLabelledBy,
@@ -208,6 +212,8 @@ class HeaderMenu extends React.Component {
       // We set the current class only if `isActive` is passed in and we do
       // not have an `aria-current="page"` set for the breadcrumb item
       [`${prefix}--header__menu-item--current`]: isActivePage,
+      [`${prefix}--header__menu-item--current`]:
+        isActivePage || (hasActiveChildren && !this.state.expanded),
     });
 
     // Notes on eslint comments and based on the examples in:
--- a/packages/react/src/components/UIShell/HeaderMenu.js
+++ b/packages/react/src/components/UIShell/HeaderMenu.js
@@ -193,9 +193,9 @@ class HeaderMenu extends React.Component {
       ...rest
     } = this.props;
 
-    const hasActiveChildren = React.Children.toArray(children).some((child) => {
-      return child.props.isActive || child.props.isCurrentPage;
-    });
+    const hasActiveChildren = React.Children.toArray(children).some(
+      (child) => child.props.isActive || child.props.isCurrentPage
+    );
 
     const accessibilityLabel = {
       'aria-label': ariaLabel,
--- a/packages/react/src/components/UIShell/HeaderMenu.js
+++ b/packages/react/src/components/UIShell/HeaderMenu.js
@@ -193,10 +193,10 @@ class HeaderMenu extends React.Component {
       ...rest
     } = this.props;
 
-    const hasActiveChildren =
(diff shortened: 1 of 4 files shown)


jsehull:
recheck

tay1orjones:
recheck

tw15egan (dismissed):
Styles look great, but it seems like it is not changing the parents state if the child is active. I think we can add something like this to `HeaderMenu.js` to handle that

`HeaderMenu.js L205`
```js
    const hasActiveChildren =
      React.Children.map(children, (child) => {
        return child.props.isActive || child.props.isCurrentPage;
      }).length > 0;
```

and then conditionally render the active class on the parent if a) its child has the active state and b) it is not expanded

`HeaderMenus.js L218`
```js
      [`${prefix}--header__menu-item--current`]:
        isActivePage || (hasActiveChildren && !this.state.expanded),
```

With that we get:

jsehull:
Ty for the feedback, will look into this sometime soon

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -193,6 +193,10 @@ class HeaderMenu extends React.Component {
       ...rest
     } = this.props;
 
+    const hasActiveChildren =
+      React.Children.map(children, (child) => {
+        return child.props.isActive || child.props.isCurrentPage;
+      }).length > 0;
```
francinelucca: slight performance improvement. you might have to do `.toArray()` before calling `.some`, not sure
```suggestion
    const hasActiveChildren =
      React.Children.some(children, (child) => {
        return child.props.isActive || child.props.isCurrentPage;
      });
```

packages/react/src/components/UIShell/HeaderMenu.js
```diff
@@ -208,6 +212,8 @@ class HeaderMenu extends React.Component {
       // We set the current class only if `isActive` is passed in and we do
       // not have an `aria-current="page"` set for the breadcrumb item
       [`${prefix}--header__menu-item--current`]: isActivePage,
+      [`${prefix}--header__menu-item--current`]:
+        isActivePage || (hasActiveChildren && !this.state.expanded),
```
francinelucca: looks like this class is defined twice?

francinelucca (changes requested):
Some refactoring suggestions.

Also looks like Percy is throwing out some changes in the bottom borders of the nav items, wondering if we're cool with those @aagonzales 

https://user-images.githubusercontent.com/40550942/235253112-6cceb73f-76b9-4b8e-84e4-d9d783722eb3.mov

aagonzales:
@francinelucca I think that's fine, looks like the original one was not including the internal border area. It should extend 16px on both sides of the text.

packages/react/src/components/UIShell/HeaderMenu.js
tw15egan: Updated 👍

tw15egan:
@aagonzales This one should be good to go for the final visual review, check out [this story](https://deploy-preview-13402--carbon-components-react.netlify.app/?path=/story/components-ui-shell--header-base-w-navigation) for testing 👍🏻

francinelucca (approved):
🚀 LGTM

aagonzales (approved):
Nice! I think that's all working as intended now 🎉

tay1orjones (approved):

PR #13406, opened 2023-03-27
refactor(Button): align icon top for wrap alignment 

jsehull:
#9478
NOT COMPLETE , currently seeking feedback.
New location for a closed PR #13370 
___

During Carbon Dev Office Hours, it was suggested to submit a format to compare all sizes so others can test/tinker 🧪.

Thank you for your insight!

___

**USE CASE** (from Design Team office hours): user changes language on the page, which can have unintended changes in length/wrapping
- image

**Findings**
- ONLY need to test sizes `sm`, `md`, `lg`
  - does NOT apply to `xl` or `2xl` who meet design requirements (already set to align `baseline`)
- `btn` is set to `align-items: center`, which is why icons stay centered.
- manipulating `line-height` had various results for me. It seems the core issue exists: text and icon have different alignment needs, so even if we alter line-height, icons require a new value
- icon background temporarily highlighted to display spacing since icons vary in size 
- Idea 1 (ugly): align all button icons with `baseline`, which looks great on wrap, but bad on single line (yanks text up)
- Idea 2 (dynamic): use JS to monitor the `max-width` of the button via `offsetWidth` sometime around page load (i.e. after language updated)
    - if `width >= 320px`, set alignment from center => baseline.
- Idea 3 (static): find a static distance for `top` since icons have `absolute` positioning
    - I tried matching `top` spacing of icon with button `padding`, but that did not align properly.
    - using `pixel` values for now because padding had calculations (below)
      - SIZE = `padding-top` = CLASS
      - sm = calc(0.375rem - 3px) = .cds—btn—sm
      - md = calc(0.675rem - 3px) = .cds—btn—md
      - lg =  calc(0.875rem - 3px) = .cds—btn
      - xl = 1rem = .cds--btn--xl:not(.cds--btn--icon-only)
      - 2xl = 1rem = .cds--btn--2xl:not(.cds--btn--icon-only)
___

**The following solution attempts Idea 3**, though alignment isn’t pixel perfect.

Are there more considerations beyond language selection that could alter button state/wrapping? Would that change the correct solution? Is it desired for all CSS to stay within the SCSS  file and not rely on JS? What’s the typical way to handle that?

#### Changelog

**New**

- add static `top` distance for the button icons

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

Open Storybook > Button > Default
- Visit `Button.stories.js` to alter `Default` example. Currently all 5 sizes are displayed
image

- Visit bottom of `/styles/scss/components/button/_button.scss` 
image

    - test area setup with pixel amounts to demonstrate how single line and wrap do not easily share a value
- Test various CSS values per button `size` and see what kind of alignment you can get


packages/styles/scss/components/button/_button.scss
```diff
@@ -419,4 +419,27 @@
   .#{$prefix}--btn--ghost.#{$prefix}--btn--icon-only .#{$prefix}--btn__icon {
     @include high-contrast-mode('icon-fill');
   }
+
+  // TEST for #9478
+  // align icons manually to account for line wrap (see: #9478)
+  .#{$prefix}--btn.#{$prefix}--btn--sm .#{$prefix}--btn__icon {
+    top: 7px;
```
alisonjoseph:
```suggestion
    top: rem(7px);
```

jsehull: added rem wrap to each instance

tay1orjones:
> Idea 2 (dynamic): use JS to monitor the max-width of the button via offsetWidth sometime around page load (i.e. after language updated)

Also this can be really expensive performance-wise. An additional wrinkle with this approach is that browser size changes can impact the wrapping, as can other elements on the page. It would probably require using ResizeObserver to detect all changes and recalculate the wrapping.

tay1orjones (commented):
> Idea 2 (dynamic): use JS to monitor the max-width of the button via offsetWidth sometime around page load (i.e. after language updated)

Also this can be really expensive performance-wise. An additional wrinkle with this approach is that browser size changes can impact the wrapping, as can other elements on the page. It would probably require using ResizeObserver to detect all changes and recalculate the wrapping.

andreancardona:
@alisonjoseph any thoughts on this? should we close this 👀

alisonjoseph:
Yea, I will go ahead and close for now, its linked in the design issue so we can reference it / reopen once we have a final spec.

PR #13409, opened 2023-03-27
fix(Modal): add proper right padding to header

### LINKED ISSUE #13228: Add TypeScript types to ActionableNotification, ToastNotification, InlineNotification, NotificationActionButton, NotificationButton

### LINKED ISSUE #13288: [Bug]: padding-right of modal-header is different to the guidance
### Package

@carbon/react

### Browser

Chrome

### Package version

v10

### Description

In the [guidance](https://v10.carbondesignsystem.com/components/modal/usage/#alignment) it is stated that modal header and description should have 20% padding-right. The actual code component does have 20% padding for the description but not for the header.

I guess either the guidance or the padding needs to be updated.

### Screenshots

image

image

### Steps to reproduce

See guidance on alignment on [carbondesignsystem.com](https://v10.carbondesignsystem.com/components/modal/usage/#alignment) 
Inspect modal in the [storybook](https://react.carbondesignsystem.com/?path=/story/components-modal--playground)

### Code of Conduct

jsehull:
Closes #13288 

Ensure Heading has proper padding alignment to match design specs shown here
image

Currently, only the paragraph text has the 20% right padding layout

#### Changelog

**New**

- add 20% `padding-right` for the Header text to match the `p` text

**Changed**

- n/a

**Removed**

- n/a

#### Testing / Reviewing

1. Open Storybook > Modal > Default (or any version)
2. Inspect the `h3` Header and confirm there is now right padding
image
3. Also note that the padding values match the `p` tag below since they maintain an equal cutoff point (only IF there is a description area on that modal variation)


francinelucca (changes requested):
Looks like the header is more padded than the content here
image

Should the padding-right be reduced by `3rem` since the modal header is padded-right by 3rem?
image

I also think the content is over-padded by `1rem`. Since it is right-padded `1rem` but then the `p` tag is reducing `2rem` from the 20%
image
image

jsehull:
@francinelucca Good eye, it does need to account for that right side padding. 

Your question had me dig for how the `padding-right` percentages work currently. Here's what I found when testing a consistent length:

___
Calculation info
type: (total `padding-right`)/full div length = right padding % | calculation used
**i.e. `padding-right` sum of both the item (h3) and outer div
__

`h3`
current: (99.2 + 3rem)/720 = 20.44% | `20% - 2rem` (current as of this PR)
perfect: (96 + 3rem)/720 = 20% | `20% - 2.2rem`

`p`
current: (105.7 + 1rem)/720 = 16.88% | `20% - 2rem`
perfect: (128 + 1rem)/720 = 20% | `20% - 0.6rem`


The header was close, but the `p` text area was further off, which is probably why you caught it so fast.
___
Shall I update them both to use the "perfect 20%" or use a simpler rem calculation? (i.e. p: `20% - 1rem` = 19.1%, h3: `20% - 2rem` = 20.44%) I didn't know if that makes it easier to understand/maintain.

packages/styles/scss/components/modal/_modal.scss
```diff
@@ -222,6 +222,8 @@
   .#{$prefix}--modal-header__heading {
     @include type-style('heading-03');
 
+    // padding should take into account the left and right padding of modal container
+    padding-right: calc(20% - $spacing-07);
```
francinelucca:
```suggestion
    // padding should take into account the left and right padding of modal container
    padding-right: calc(20% - $spacing-09);
```

francinelucca (changes requested):
Thanks for that detailed explanation @jsehull. let's just deduct an extra rem in the header and that should be good enough.

packages/styles/scss/components/modal/_modal.scss
jsehull: Simple - sounds good

francinelucca (approved):

aledavila (approved):

PR #13442, opened 2023-03-31
feat: contained list persistent search

### LINKED ISSUE #12801: [ContainedList]: Persistent search: Have a persistent search bar above list rows.
This issue is for implementing search functionality in the contained list.

Design spec: https://github.com/carbon-design-system/carbon/issues/12330

### Acceptance criteria
- [x] Search bar is persistent, below title and above items
- [x] Typing in the search bar would filter the children list items based on a simple text match search (like FilterableMultiSelect)
- [x] Removing/clearing the value from the search bar would "reset" the component to show all children rows
- [ ] Provide `onChange` handlers for search so async actions can be taken
- [ ] Support search initial state - render filtered items based on initial search value
- [ ] Should it provide a `filterItems` type of function to allow for overriding the search behavior? What if it's not a simple text node in the list item? Do we allow non-text nodes as children items/rows?

andreancardona:
Closes #12801 

- This PR includes contained list persistent search and the `Search` story mock-up
- After chatting with @laurenmrice it seems we don't want to define the search functionality for the teams - and so instead I included a story to users can see what the default design would look like

Testing: 

- [x] Control


francinelucca (changes requested):
This LGTM, since we're not adding a prop specifically to enable the search in `ContainedList`, wondering if it would make sense to add a blurb in the [docs](https://deploy-preview-13442--carbon-components-react.netlify.app/?path=/docs/components-containedlist--with-search) on how to implement the search

laurenmrice (changes requested):
Looking good @andreancardona ! ⭐️ Just a couple things:

- It looks like the `With Actions` story used to have the collapsible search option, which we want to keep but now it only has the persistent search option. In storybook, somehow we need to be able to have both options of search available to use: having it collapsible in the header/title area and having it persistent in the row. People should not be able to use both at once though in one list. I don’t know if this should be prop options, or just different story examples? Let me know if that makes sense.
- [x] Have both options (search in title/ header) &  persistent in the row (but not both at the same time)
- [x] There should not be a divider line in the search between the icon button and the search field. 
- [x] Update focus / hover state styling 

Screen Shot 2023-04-07 at 9 24 30 AM

- Typically, we try to hide or overlap these horizontal rule dividers of the rows when there is a hover state. (For example: the dismiss icon buttons below it)
Screen Shot 2023-04-07 at 9 24 58 AM

andreancardona:
@laurenmrice ooo good catch I'll fix that!

packages/react/src/components/ContainedList/ContainedList.mdx
```diff
@@ -23,6 +23,17 @@ import { Props, Story, Preview } from '@storybook/addon-docs';
   <Story id="components-containedlist--default" />
 </Preview>
 
+### Search
+
+We have added in the capability to pass down the `Search` component within the `ContainedList` action prop.
+The `Search` component itself is not a prop but does render with specific styling such that the `Search` bar appears below the Contained List title as opposed to rendering inline.
+The `Search` component will also remain sticky under the title, so that it remains visible on scroll, when there are many list items passed in. 
+## Render Custom Icon
```
francinelucca: is this header here intended? 
image

packages/react/src/components/ContainedList/ContainedList.mdx
```diff
@@ -23,6 +23,17 @@ import { Props, Story, Preview } from '@storybook/addon-docs';
   <Story id="components-containedlist--default" />
 </Preview>
 
+### Search
```
francinelucca, line 30:
```suggestion
## Search
```

packages/react/src/components/ContainedList/ContainedList.mdx
```diff
@@ -23,6 +23,17 @@ import { Props, Story, Preview } from '@storybook/addon-docs';
...
+
+We have added in the capability to pass down the `Search` component within the `ContainedList` action prop.
+The `Search` component itself is not a prop but does render with specific styling such that the `Search` bar appears below the Contained List title as opposed to rendering inline.
+The `Search` component will also remain sticky under the title, so that it remains visible on scroll, when there are many list items passed in. 
+## Render Custom Icon
+
+<Preview>
+  <Story id="components-containedlist--with-search" />
+</Preview>
+
```
francinelucca: Could we add some sample code before the preview? either whole or broken into pieces

```
const [searchTerm, setSearchTerm] = useState('');
const [searchResults, setSearchResults] = useState([]);
const handleChange = (event) => {
    setSearchTerm(event.target.value);
  };

useEffect(() => {   
  const listItems = [
      'List item 1',
      'List item 2',
      'List item 3',
      'List item 4',
      'List item 5',
      'List item 6',
      'List item 7',
      'List item 8',
  ];

  const results = listItems.filter((listItem) =>
    listItem.toLowerCase().includes(searchTerm.toLowerCase())
  );
    setSearchResults(results);
}, [searchTerm]);

return (
    <ContainedList
      label="List title"
      kind="on-page"
      action={
        <Search
          placeholder="Search"
          value={searchTerm}
          onChange={handleChange}
          size="lg"
        />
      }>
      {searchResults.map((listItem, key) => (
        <ContainedListItem key={key}>
          {listItem}
        </ContainedListItem>
      ))}
    </ContainedList>
```

francinelucca (changes requested):

laurenmrice (commented):
@andreancardona 

- I can see both the Expandable Search and Persistent Search options in different stories now. 👍

- I was wondering if we could move the `With Search` story to be right below the `With Expandable Search` story so they are closer together?
- [x] Can you also change the name of the story from `With Search` to say `With Persistent Search`to be more clear. This is the terminology we will be using in our design docs/design kit

Other than that everything looks great!

andreancardona:
> * Can you also change the name of the st

I don't think we can change the order since storybook renders alphabetically - unless we put it under one umbrella `Search` folder? Thoughts. Otherwise, just made all the other changes!

laurenmrice (approved):
@andreancardona Oh that is right, we can leave the positioning of the stories as is. Everything looks good. Thanks Andrea! ✅

packages/react/src/components/ContainedList/ContainedList.js
```diff
@@ -33,17 +33,39 @@ function ContainedList({
...
   );
 
+  const filteredChildren = children?.filter(
+    (child) =>
+      !['Search', 'ExpandableSearch'].includes(child?.type?.displayName)
+  );
+
+  const isActionSearch = ['Search', 'ExpandableSearch'].includes(
+    action?.type?.displayName
+  );
```
tay1orjones: The only caveat I see is that this is locked to components specifically named `Search` and `ExpandableSearch`. If a user wants to use their own component similar to `Search` but named something else, the component won't get any of the styling or rendering help we've provided. The children won't be filtered, etc.

I kind of wish we would've kept ContainedList experimental a little bit longer 😅 

If we had the ability to make a breaking change, I'd suggest making the header its own component, ContainedListHeader. This way we'd have options to do things like
```jsx
<ContainedList>
    <ContainedListHeader action={<ExpandableSearch/>}/>
    <ContainedListItem/>
    <ContainedListItem/>
</ContainedList>
```

or

```jsx
<ContainedList>
    <ContainedListHeader>
        <Search/>
    </ContainedListHeader>
    <ContainedListItem/>
    <ContainedListItem/>
</ContainedList>
```

`ContainedListHeader` could then always wrap any provided children with the class used for styling the sticky header, enabling more flexibility without having to snoop children `displayName`s. Also from my testing, the persistent Search doesn't appear to need to be rendered within the `ul`, it works as a sibling of `.${prefix}--contained-list__header`

---

Could we achieve this without a breaking change via conditional rendering?

* Make the new ContainedListHeader component
* Deprecate props previously used in the header section of ContainedList, point users to ContainedListHeader
* Leave the existing header code in ContainedList, but if ContainedList is given a child of `ContainedListHeader`, don't render the existing code. Instead render the ContainedListHeader. If it's not given a ContainedListHeader, render the existing code.
* Update documentation and stories to use ContainedListHeader
* In v12, remove the existing header code and condition looking for ContainedListHeader as a child

tay1orjones (commented):
This is a big dump of information 😅 let's pair up and chat, I'm curious what you think.

packages/react/src/components/ContainedList/ContainedList.js
francinelucca: that'd work for applying the search classes , but we'd still need to use the component's `DispalyName`s to filter out the children, correct?

andreancardona: @tay1orjones Great feedback - thank you :)   @francinelucca 's comment was the same thought I had, curious your thoughts!

tay1orjones: Yes good point - we'd still need to use `displayName` to distinguish between ContainedListHeader and ContainedListItem(s).

After a convo with @andreancardona though, I think it makes sense to disregard this feedback of mine altogether. Due to the rigidity of the design spec around this I think we could merge this as is and wait for feedback from consumers. If there's a strong need for customization of the header, we can always provide an isolated header component in the future.

tay1orjones (approved):
A CI check is failing, but this looks good otherwise! 👍

packages/react/src/components/ContainedList/ContainedList.mdx
```diff
@@ -23,6 +23,116 @@ import { Props, Story, Preview } from '@storybook/addon-docs';
   <Story id="components-containedlist--default" />
 </Preview>
 
+### Search
+
+#### Expandable Search: 
+Within the Contained List title you can pass in an `ExpandableSearch` component within the `action` prop as seen below.
```
francinelucca:
```suggestion
Within the Contained List you can pass in an `ExpandableSearch` component within the `action` prop as seen below. 
```

francinelucca (changes requested):
LGTM, one typo suggestion and I also think the `WithExpandableSearch` story should be added to VRT? other than that this should be good to go!

francinelucca (changes requested):
LGTM, one typo suggestion and I also think the `WithExpandableSearch` story should be added to VRT? other than that this should be good to go!

francinelucca (commented):
🎉

francinelucca (approved):

PR #13445, opened 2023-03-31
fix(NumberInput): have less padding if number is valid

### LINKED ISSUE #10155: [Bug]: NumberInput clips input value when component width is small-ish
### Package

carbon-components

### Browser

Chrome, Safari

### Package version

demo site

### Description

When number input width is small, the input values get clipped.
I see this in both angular and react implementation.

image

It looks like we reserve a big space for the error icon. 
I am wondering if we could only clip the value when error icon is shown.

image

### CodeSandbox example

https://codesandbox.io/s/zen-field-vjyhe?file=/src/index.js

### Steps to reproduce

Just look, you will see behavior of number getting clipped.
image

### Code of Conduct

pmweeks98:
Closes #10155

Change padding for number input depending on whether icon is showing using the data-invalid property

#### Changelog

**Changed**

- padding for number input 

#### Testing / Reviewing

- Go to storybook playground for number inout
- set width for #root to 200px
- Set a max value to be very high
- Input number - Observe number goes right up to the edge of the decrement button
- Decrease the max value so that icon shows
- Observe number is clipped at the same level it is today

The new padding is right up to the decrement button when the icon is not showin, see below. This may not be desirable so let me know if the padding should be increased
Screenshot 2023-03-31 at 21 54 43


francinelucca (approved):
this LGTM, @carbon-design-system/design can we get signoff on these new `padding-right` values for the input? 
```
Default: 80px
Default + Invalid: 128px

Large: 96px
Large + invalid: 144px

Small: 64px
Small + invalid: 112px
```
Essentially we're removing `48px` from the normal state and re-adding it to account for the invalid icon when needed
image
image

andreancardona (approved):

laurenmrice (changes requested):
The invalid state padding looks correct for the different sizes. But the enabled state needs some adjustments. There should be 16px padding before the container of the subtract control. 
Artboard Copy

francinelucca (changes requested):
I don't think this looks right according to Lauren's spec. looks like it's still missing the extra 16px right padding on the default state 
image
image

pmweeks98:
Sorry - made the wrong changes and didnt check with storybook cause I was having issue with nvm

image

laurenmrice (changes requested):
@pmweeks98 

- The enabled state spacing for each size looks good now. ✅
- I did notice though that the `warn` prop does not get the same spacing as the `invalid` prop. The warn prop should have the same spacing as the invalid prop.

Screen Shot 2023-04-10 at 9 07 30 AM
----

The size + invalid state should match the spacing listed below:

Large: `112px`
Large + Invalid: `144px`
Large + Warn: `144px`

Medium: `96px`
Medium (default) + Invalid: `128px`
Medium (default) + Warn: `128px`

Small: `80px`
Small + Invalid:`112px`
Small + Warn:`112px`

francinelucca (approved):
LGTM!

laurenmrice (approved):

pmweeks98:
@andreancardona @aubrey-oneal can one of ye have a look at this please?

alisonjoseph (approved):

PR #13460, opened 2023-04-03
fix(CopyButton): adjust aria-live, add closeOnActivation

### LINKED ISSUE #13163: [a11y]: CopyButton announces "Copied!" multiple times when pressing the copy button
### Package

carbon-components-react

### Browser

Chrome

### Operating System

_No response_

### Package version

3.3.3

### React version

^16.12.0

manual testing with NVDA

### Assistive technology

NVDA

### Description

The `<CopyButton>` component announces "Copied!" message more than once when pressed.

### WCAG 2.1 Violation

1.3.1 

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-copybutton--default

### Steps to reproduce

- Using Google Chrome, open any of the below links where CopyButton is present:
  - https://carbondesignsystem.com/components/code-snippet/usage/#live-demo
  - https://codesandbox.io/s/c681ih
  - https://react.carbondesignsystem.com/?path=/story/components-copybutton--default
- Open NVDA and its speech viewer tool
- Using Tab navigation, navigate until reaching an instance of `<CopyButton>` component and press space bar to act on it.
- Notice that the "Copied!" message is announced more than once.

Related issues:
https://github.com/carbon-design-system/carbon/issues/5385
https://github.com/carbon-design-system/carbon/issues/5553

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13163

removes the `aria-live` that was causing the voiceover to read the copy text twice. This seems like a holdover from before we used `IconButton` under the hood. Since `IconButton` has proper aria roles and labels, it can handle the voiceover itself. 

Also noticed an issue where the popover content was being closed immediately upon activation, rendering the `feedback` text useless

To fix this, I've added the `closeOnActivation` prop as configurable in `iconButton`, rather than setting it to always be false. It is passed in as `false` from `Copy`

#### Changelog

**Changed**

- `closeOnActivation` can be configured in `iconButton`

**Removed**

- `aria-live` was removed from `CopyButton`

#### Testing / Reviewing

Ensure there are no a11y violations. Ensure the voiceover only reads the feedback text once.


aledavila (approved):
lgtm. no a11y violations and it doesn't keep repating copied

andreancardona (approved):
😎

PR #13467, opened 2023-04-04
feat(OverflowMenu): add `tooltipAlignment` prop as part of OverflowMenu API

### LINKED ISSUE #13446: [Feature Request]: OverflowMenu missing ability to control tooltip position
### The problem

OverflowMenu has a requirement for an icon description which triggers a tooltip but there is no way to control the tooltips position.  This defaults to top center which can cause render issues depending on the use case of the overflow menu and its position on the page.

### The solution

Add tooltipPosition similar to other components 

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13446

`align` could already be passed to `IconButton` inside `OverflowMenu` via the `other` prop. This change will surface as `tooltipAlignment` on the `OverflowMenu` API so users will know it is available to use in this component.

#### Changelog

**New**

- `tooltipAlignment ` prop is added in `OverflowMenu`

#### Testing / Reviewing

Go to the playground story inside `OverflowMenu` and change the `tooltipAlignment ` prop

### DIFF
--- a/packages/react/src/components/ComboButton/index.js
+++ b/packages/react/src/components/ComboButton/index.js
@@ -36,7 +36,7 @@ const ComboButton = React.forwardRef(function ComboButton(
     label,
     onClick,
     size = 'lg',
-    tooltipAlign,
+    tooltipAlignment,
     translateWithId: t = defaultTranslateWithId,
     ...rest
   },
@@ -105,7 +105,7 @@ const ComboButton = React.forwardRef(function ComboButton(
         label={t('carbon.combo-button.additional-actions')}
         size={size}
         disabled={disabled}
-        align={tooltipAlign}
+        align={tooltipAlignment}
         aria-haspopup
         aria-expanded={open}
         onClick={handleTriggerClick}
@@ -163,7 +163,7 @@ ComboButton.propTypes = {
   /**
    * Specify how the trigger tooltip should be aligned.
    */
-  tooltipAlign: PropTypes.oneOf([
+  tooltipAlignment: PropTypes.oneOf([
     'top',
     'top-left',
     'top-right',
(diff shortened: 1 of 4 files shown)


alisonjoseph (approved):
LGTM!

francinelucca (approved):
This LGTM, would we want to remove this https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--default#customizing-the-tooltip then or should we keep it?

janhassel:
@tw15egan Hey, just chiming in as I'd port this over to `OverflowMenuV2` as well. When working on `ComboButton` I called this prop `tooltipAlign` in order to not confuse it with the alignment of the options themselves.

Now that I look through the code of other components `tooltipAlignment` seems to be the established name as it is used in:
- `Button`
- `ControlledPasswordInput`
- `PasswordInput`
- `HeaderGlobalAction`

Should this all be aligned and therefore be added to `OverflowMenuV2` and updated in `ComboButton`? Both of these components are marked as experimental, so no breaking change needed. For the `OverflowMenu` component I guess it would have to be decided within this PR?

tw15egan:
That's a great point @janhassel, I hadn't considered those use-cases. I think aligning them to be `tooltipAlignment` across the repo makes sense, I can update `OverflowMenu`, `OverflowMenuV2`, and `ComboButton` here to follow that norm.

tw15egan:
@alisonjoseph @francinelucca I've made some changes based on Jan's feedback and have updated this prop to be `tooltipAlignment` so that it aligns with the way we do this in other parts of the codebase. I also went ahead and removed that doc section, let me know what you think 👍🏻

tw15egan:
I think I'm actually going to close this, as this can already be achieved using the `align` prop, and has documentation (https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--default#customizing-the-tooltip). If we change this to `tooltipAlignment, it could break users currently using `align`, unless we add `align` as a prop as well, and pass it as `tooltipAlignment`, which seems messy. 

`ComboButton` should probably be updated to align with the other components in a separate PR

PR #13469, opened 2023-04-04
fix(layer): add border-subtle-00 to layerset map

### LINKED ISSUE #13465: Border-subtle: color bugs
Something is happening with the `$border-subtle` token in the light themes that is causing the wrong border color to appear on the wrong background. I think maybe its because `border-subtle-00` is not accounted for in contextual tokens maybe? 

Using accordion as an example these are the expected border and background combinations for the the white theme. However, that is not what is seen in [storybook](https://react.carbondesignsystem.com/?path=/story/components-accordion--default&globals=theme:g10). 

image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13465

Added `$border-subtle-00` to `layer-sets.scss` so that it properly renders the correct value when no layer element is present

#### Changelog

**New**

- `$border-subtle-00` to `layer-sets.scss` 

#### Testing / Reviewing

Go to the `Accordion`, `DataTable`, and other stories that contains a `$border-subtle` value and ensure it matches the spec


alisonjoseph (approved):

francinelucca (approved):

aagonzales (approved):
Yes! That fixes it! 🎉

Issue #4338, opened 2019-10-15
[DataTable] Performance: Select All takes 5 - 30 s for tables with 1000+ rows


tomkdgun:
[DataTable] Performance: Select All takes 5 - 30 s for tables with 1000+ rows

https://codesandbox.io/embed/codesandbox-1lumy

For DataTable with 5000 rows and **only 10 rendered**, "Select All" (checkbox selection) takes **5 seconds**. You can play with MyDataTable.js (contructor) to change number of variables.
For 30000 rows it hangs for over **30 seconds**.

I'm assuming that this could be related with DataTable.setAllSelectedState implementation.
Maybe cloning using `JSON.parse(JSON.stringify(obj));` will be faster.

## What package(s) are you using?

    "carbon-components": "10.7.0",
    "carbon-components-react": "7.7.0",
    "carbon-icons": "7.0.7",
    "@carbon/layout": "10.5.0",
    "@carbon/themes": "10.7.0",

## Steps to reproduce the issue

1. Go to codesandbox I've created 
2. Try to select or unselect "Select all" checkbox (table header)

> Please create a reduced test case in CodeSandbox

https://codesandbox.io/embed/codesandbox-1lumy

asudoh:
Great if somebody can try simple `for` loop or a like that doesn't require rest/spread for each row.

tomkdgun:
Any update on this issue ?

ljitendra:
For a DataTable with radio selection having multiple columns having their own cell renderers, it takes more than 60 seconds to load 15000 rows. And the page becomes responsive. Need to check this scenario as well when defect is being addressed.

tymerica:
This is happening on production builds on our system, 4.5k items causes 15 seconds on my computer but on weaker computers it just locks up for minutes

dezkareid:
did you try react-virtual?

tomkdgun:
We had pagination with only 30 items displayed, so problem is with some internal state handling inside DataTable.

PR #13525, opened 2023-04-13
fix(DataTable): align items top on text wrapping

### LINKED ISSUE #12610: [Bug]: Data table content inside cells should top align when with text wrap in some cells
### Package

@carbon/styles

### Browser

_No response_

### Package version

Latest

### React version

_No response_

### Description

Currently, when some cells' content wrap to 2 lines, the text alignment is middle/center aligned.
Expecting all the content to top align:

image

### Suggested Severity

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datatable-basic--default

### Steps to reproduce

1. Go to https://react.carbondesignsystem.com/?path=/story/components-datatable-basic--default
2. Narrow down browser width to see text wrapping and alignment bug

### Code of Conduct

francinelucca:
Closes #12610 

Programmatically determine if a `Table` has a cell that is wrapping text content inside of it, if so apply a new `cds--data-table--top-aligned` class that will make the table align items top.

My approach here is for any given table cell, measure the textWidth inside a canvas when it is not being wrapped and compare it to the cell's width, if the rendered text is wider than the cell then the text is being wrapped. This was the only way I found to do this, definitely open to suggestion on a better implementation.

Shoutout @guidari  for pairing up with me on this 🚀

#### Changelog

**Changed**

- Call to new `setTableAlignment` function in `Table` component on first render and on resize to determine wether table should be top aligned.
- Add/Modify `Table` css to align items top and appropriate padding when necessary
- Enforce default `lg` size in `Table`

#### Testing / Reviewing

https://deploy-preview-13525--carbon-components-react.netlify.app/?path=/story/components-datatable-basic--default

1. Go to a DataTable `Playground` story
2. turn `experimentalAutoAlign` control on
3. resize window until text is wrapped, confirm items align to the top on text wrapping and that when no text is being wrapped datatable still aligns center and looks as expected. 
4. Turn `experimentalAutoAlign` control off and confirm items do not align to the top when text is wrapping
5. I know it's a lot but please test different variations and sizes 🙏🏻🙏🏻🙏🏻
Sizes: `xs`,`sm`,`md`,`lg`,`xl`
Variations:
- Expansion
- Selection (Checkbox, RadioButton)
- Actions
- With OverflowMenu


alisonjoseph:
I haven't dug into the code yet, but should we hide this functionality behind a prop? it visually changes things, so technically adding this in now could be classified as breaking. Thoughts @francinelucca @tay1orjones

tay1orjones:
@alisonjoseph We did the feature flag for the new disabled styling for Tile because it was changes to styling of existing selectors.

In this case it's all behind a new classname. Yes in theory there could still be conflicts with downstream consumers' styling but I think the surface area of this is relatively minimal and doesn't warrant a feature flag.

We can always add a feature flag in the future if we receive feedback of it causing unintended consequences downstream.

packages/react/package.json
```diff
@@ -95,6 +95,7 @@
     "babel-preset-carbon": "^0.4.0",
     "browserify-zlib": "^0.2.0",
     "browserslist-config-carbon": "^11.1.0",
+    "canvas": "^2.11.2",
```
tay1orjones, line 98: Is this needed? I don't see it being used

packages/react/src/components/DataTable/Table.tsx
```diff
@@ -66,13 +74,69 @@ export const Table = ({
...
+      ).some((td) => {
+        if (td.children.length > 0) {
+          return;
+        }
+        const computedStyles = window.getComputedStyle(td);
+        context.font = computedStyles.font
+          ? computedStyles.font
+          : `${computedStyles.fontSize}" "${computedStyles.fontFamily}`;
+
+        let textWidth = context?.measureText(td.textContent ?? '').width ?? 0;
```
tay1orjones: Looking into the strategy of using canvas to measure things like this I found this

https://erikonarheim.com/posts/canvas-text-metrics/

> The `measureText.width` Is A Lie
> You probably want to add `actualBoundingBoxLeft` and `actualBoundingBoxRight`, this will be a more accurate than just `width`. The reason is some fonts have glyphs that hang or stretch over their width used for text placement. This is especially true if you want to ensure text renders within certain confines, like UI for a game.

This is my general concern about this strategy of using canvas to measure. Do you know if there's any other pitfalls around it? Thinking about emoji unicode glyphs that can sometimes take up to character widths and things like that.

packages/react/src/components/DataTable/Table.tsx
```diff
@@ -66,13 +74,69 @@ export const Table = ({
...
+      const isMultiline = Array.from(
+        tableRef.current.querySelectorAll('td')
+      ).some((td) => {
+        if (td.children.length > 0) {
+          return;
+        }
+        const computedStyles = window.getComputedStyle(td);
+        context.font = computedStyles.font
+          ? computedStyles.font
+          : `${computedStyles.fontSize}" "${computedStyles.fontFamily}`;
```
tay1orjones: Will this work properly if the user has font settings in the browser turned up? or if the browser is zoomed?

packages/react/src/components/DataTable/Table.tsx
```diff
@@ -66,13 +74,69 @@ export const Table = ({
...
+  const setTableAlignment = useCallback(() => {
+    const fragment = document.createDocumentFragment();
+    const canvas = document.createElement('canvas');
+    fragment.appendChild(canvas);
+    const context = canvas.getContext('2d');
+
+    if (tableRef.current && context) {
+      const isMultiline = Array.from(
+        tableRef.current.querySelectorAll('td')
+      ).some((td) => {
```
tay1orjones: Iterating over every table cell on every render seems really expensive. Do you think this will be performant enough to use on tables with large datasets that don't use pagination? Might be worth testing to assess the usability impact.

packages/react/src/components/DataTable/Table.tsx
```diff
@@ -66,13 +74,69 @@ export const Table = ({
...
+
+      if (isMultiline) {
+        tableRef.current.classList.add(`${prefix}--data-table--top-aligned`);
+      } else {
+        tableRef.current.classList.remove(`${prefix}--data-table--top-aligned`);
+      }
+    }
+  }, [prefix]);
+
+  useEvent(window, 'resize', setTableAlignment);
```
tay1orjones: Similar concern here with performance, iterating over every table cell for every resize event could really impact usability if the updates are blocking.  I don't think `useEvent` debounces or throttles the events either.

packages/react/src/components/DataTable/Table.tsx
```diff
@@ -66,13 +74,69 @@ export const Table = ({
     [`${prefix}--data-table--sticky-header`]: stickyHeader,
     [`${prefix}--data-table--visible-overflow-menu`]: !overflowMenuOnHover,
   });
+
+  const setTableAlignment = useCallback(() => {
+    const fragment = document.createDocumentFragment();
+    const canvas = document.createElement('canvas');
+    fragment.appendChild(canvas);
+    const context = canvas.getContext('2d');
```
tay1orjones: Does jsdom support canvas? Can we write tests against this functionality?

tay1orjones (changes requested):
Kudos for diving into this, it's complicated and difficult to figure out something that works! I'm a bit concerned about the reliability of measuring with canvas and the performance hit of measuring each cell.

If this method proves to be reliable and performant though, I could see it being useful in other situations where we need to intelligently test for truncation when we can't rely on `text-overflow: ellipsis`. What do you think about moving the measurement logic to a reusable `useTextMeasurement` hook? 

```js
// something like this maybe?
const hasOverflow = useTextMeasurement(myRef);
```

packages/react/package.json
francinelucca: yeah looks like the test suite needs it, texts were failing because the getContext API was not available, adding this dev dependency fixed the issue

packages/react/src/components/DataTable/Table.tsx
francinelucca: I think it should with the added canvas dependency, the tests that were failing are passing since adding that

packages/react/src/components/DataTable/Table.tsx
francinelucca: yeah I agree it's not the most elegant code out there 😅. I couldn't figure out any other way to do it though, any one cell could be wrapping content, so we need to check all of them - or stop when we find one that is wrapping, which I'm doing but if none of them are wrapping then yeah, it'll run through all-.
I can add a large datatable to test, how many rows/columns are you thinking will be a good enough test?

packages/react/src/components/DataTable/Table.tsx
francinelucca: zoom seems to work, will test browser font!

packages/react/src/components/DataTable/Table.tsx
francinelucca: I can definitely refactor it to use the bounding boxes, I honestly don't have a whole lot of experience with the canvas API but I can do some research

packages/react/src/components/DataTable/Table.tsx
tay1orjones: 1000+ rows is where we've seen other performance issues https://github.com/carbon-design-system/carbon/issues/4338

If we're not 100% confident in the performance, we could consider putting the behavior behind a an experimental prop so users can opt into it rather than it be on by default. Document that it has performance implications, etc

packages/react/src/components/DataTable/Table.tsx
francinelucca: I'm actually experiencing more precision when using `textwidth` than when using bounding boxes:
with textWidth, notice when `top-aligned` class gets applied/removed:

https://user-images.githubusercontent.com/40550942/235536574-e9618b62-411e-4706-a7ff-bf1173d6c4c2.mov


with `actualBoundingBoxLeft` and `actualBoundingBoxRight`, notice when `top-aligned` class gets applied/removed:

https://user-images.githubusercontent.com/40550942/235536598-0d2fc21e-24dd-4a9c-9a14-a8159e272a8a.mov

packages/react/src/components/DataTable/Table.tsx
francinelucca: Correctly aligns to the top on browser font size change: 

https://user-images.githubusercontent.com/40550942/235542073-3d56c835-cae4-4309-bbb4-fd2a3e4c69db.mov

francinelucca: Correctly aligns to the top on browser zoom: 

https://user-images.githubusercontent.com/40550942/235542657-f637fe1a-beb9-414d-8aee-944cc714515a.mov

tay1orjones (commented):
Thanks for providing those testing stories! The 10000 cells example looks pretty rough

image

Would it be more applicable for this code to live within TableCell or TableRow and be inside a `useLayoutEffect` instead of `useEffect`? I wonder if that would break this work up into [multiple tasks](https://web.dev/optimize-long-tasks) instead of one long running one. I also wonder if react might then be able to assist in yielding to the main thread since each task is contained within a multiple components instead of in the `Table`.

packages/react/package.json
```diff
@@ -50,6 +50,7 @@
     "@carbon/telemetry": "0.1.0",
     "classnames": "2.3.2",
     "copy-to-clipboard": "^3.3.1",
+    "debounce": "^1.2.1",
```
tay1orjones: @francinelucca Could this use the existing `lodash.debounce` that's already installed, or is this package better?

francinelucca: done! thankss

tay1orjones (dismissed):
Looks great! 🎉

laurenmrice (changes requested):
@francinelucca This is looking great! 🌟 Just found a couple of small buggy things:

**Across the board:** the `Md size` column header text needs to stay vertically center in the container, right now it's a bit too high. Looks like it's working okay for the other sizes.

Screen Shot 2023-05-19 at 9 52 13 AM

-----

**Expansion story:** For the` Md size`, the chevron icon needs to be aligned with the first row of text in each row. Looks like it is hanging a little low. 

Screen Shot 2023-05-19 at 10 02 28 AM


----

**Selection (Checkbox):** For the `Md size`, the checkbox icon is too high in the rows and should be aligned with the first line of text in the row.

Screen Shot 2023-05-19 at 10 06 15 AM

----

**Selection (Radio):** For `Xs`, `Sm`, `Md sizes`, the radio icon is too high and should align with the first line of text in the row. I also noticed the column header still has a checkbox, and it should be removed, but that seems to be unrelated to this specific PR.

Screen Shot 2023-05-19 at 10 17 43 AM

---

**Actions:** The checkbox is a little high in the `Sm` and `Md sizes`, and is a little low in the `Lg size`. 

Screen Shot 2023-05-19 at 10 30 02 AM


---

**With OverflowMenu:** For the `Lg size`, the overflow menu icon should be aligned with the first line of text in the row. Right now its a little low.

Screen Shot 2023-05-19 at 10 25 53 AM

francinelucca:
@laurenmrice this should be ready for re-review, pending header wrapping confirmation

alisonjoseph (dismissed):
This looks great! pending design approval 😄

francinelucca:
This should be ready for re-rev @laurenmrice

francinelucca:
@alisonjoseph @tay1orjones re-requesting revs 'cause I added a decent amount of code to handle TableHeaders independently of the body 😬

laurenmrice (changes requested):
**Expansion story**

- When I am on the `Xs` size, without text wrapping to multiple lines in rows, I notice the focus around the chevron icon is the correct size (32 x24), but when the text starts wrapping to multiple lines in rows, the focus box becomes larger in size. The focus box should always remain the same size (32 x 24) in both cases so that the chevron is in the center of the focus box. 

Screen Shot 2023-05-30 at 12 13 31 PM


- This also happens with the `Lg` size, but it's at first showing a very small focus box around the chevron icons when text lines do not wrap in rows, but once the text does start to wrap, the focus box is the correct size (32 x 32). It should always be 32 x 32 for both cases.

Screen Shot 2023-05-30 at 12 14 06 PM


**With Overflow Menu story**
- Looks like there is some added padding to rows in the LG size than the XL size.

May-30-2023 12-12-27

francinelucca:
This should be good for re-rev @laurenmrice 🙏🏻

laurenmrice (approved):
Looks good to me! ✅

tay1orjones (approved):

alisonjoseph (approved):

PR #9103, opened 2021-07-02
feat(eslint): add eslint-testing-library-plugin; add config

dakahn:
Closes #9048 

Adds eslint-testing-library and configs for said eslint testing 🏄🏽

#### Testing / Reviewing

- `yarn install` in Carbon root
- run `yarn lint` and there should be no errors 👍🏽

PR #13528, opened 2023-04-14
fix(eslint): add eslint-plugin-testing-library

### LINKED ISSUE #9048: Add the `eslint-plugin-testing-library`package to our eslint config
We would like to add support for the testing library ESLint plugin to our eslint config: https://github.com/testing-library/eslint-plugin-testing-library

This issue will cover adding the plugin, enabling relevant rules, and scoping them to specific test files th at are currently using React Testing Library. This will be the tricky part of this change as most files are currently using enzyme. The strategy that we use to bring this in will need to allow us to incrementally update our files over to RTL and the eslint config should be able to reflect that.

falsepopsky:
Closes #9048

Following this pull request [feat(eslint): add eslint-testing-library-plugin; add config](https://github.com/carbon-design-system/carbon/pull/9103), only the configuration was introduced, but the necessary dependency was never added and the configuration was never imported in order to load it.

#### Changelog

**New**

- feat: add `eslint-plugin-testing-library` and config rules.

**Changed**

- deps: bump `eslint`, `eslint-config-prettier`, `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin`

#### Testing / Reviewing

The CI should pass before merging this pull request


falsepopsky:
### files

The previous configuration only lints a specific file. I think it should only lint React packages,
but I'm unsure whether the "-link" suffix is intended for this purpose.

### rules

I removed all the rules that are enabled by default and only left the ones that are not included in
the previous developer's recommended configuration. I suggest taking a look at these rules.

| Rules                           | eslint-plugin-testing-library | internal    |
| ------------------------------- | ----------------------------- | ----------- |
| await-fire-event                | off                           | error       |
| no-manual-cleanup               | off                           | error       |
| prefer-explicit-assert          | off                           | error       |
| prefer-user-event               | off                           | error       |
| prefer-wait-for                 | off                           | error       |
| render-result-naming-convention | error                         | off         |
| no-global-regexp-flag-in-query  | off                           | never added |
| prefer-query-by-disappearance   | error                         | never added |

For reference to the rules, please check
https://github.com/testing-library/eslint-plugin-testing-library#supported-rules.

This is a work in progress, so I can't move on until the rules and files are approved by the core
maintainers.

tay1orjones:
@falsepopsky Sorry for the delay here! I wouldn't place too much emphasis on decisions made by the previous developer a while back in the initial PR. Let's come at it fresh.

For files, I think it can be scoped to `*-test.js` files inside the `packages/react` directory. Do a global search for `@testing-library/react` in the codebase and you should see most/all usages in those files and in that directory.

For rules, we're now using the latest version of `@testing-library/react` and `user-event`. I'd like to try for the recommended ruleset for now.

falsepopsky:
@tay1orjones here's a quick update:

### Config

https://github.com/carbon-design-system/carbon/blob/c3ffbe4c3f5c2564e96484770c653367bffd7164/config/eslint-config-carbon/plugins/testing-library.js#L10-L18

### Errors

Most of the errors are caused by the [no-node-access](https://github.com/testing-library/eslint-plugin-testing-library/tree/main/docs/rules/no-node-access.md) rule. It is difficult to read and follow from the workflow terminal, so I have hosted the file to make it easier to view: [ESLint Report](https://falsepopsky.github.io/carbon-eslint-errors/)

### Changes

I fixed some ESLint errors in components starting with the letters A and B. Please take a look when you have a moment and let me know if everything looks good.

packages/react/src/components/Accordion/__tests__/AccordionItem-test.js
```diff
@@ -13,9 +13,11 @@ import { render, screen } from '@testing-library/react';
...
   describe('renders as expected - Component API', () => {
     it('should spread extra props onto outermost element', () => {
-      const { container } = render(<AccordionItem data-testid="test-id" />);
-
-      expect(container.firstChild).toHaveAttribute('data-testid', 'test-id');
+      render(<AccordionItem data-testid="test-id" />);
+      expect(screen.getByTestId('test-id')).toHaveAttribute(
+        'data-testid',
+        'test-id'
+      );
```
tay1orjones: Refactoring to use `screen.getByTestId` like this doesn't test to ensure the attribute is placed on the outermost element.

So this is a case in which we want to ignore the rule since we intentionally want to access the node to test that the data-testid is placed on the outermost element.

packages/react/src/components/Accordion/__tests__/AccordionItem-test.js
```diff
@@ -29,9 +31,8 @@ describe('AccordionItem', () => {
     });
 
     it('should support a custom `className` prop on the outermost element', () => {
-      const { container } = render(<AccordionItem className="custom-class" />);
-
-      expect(container.firstChild).toHaveClass('custom-class');
+      render(<AccordionItem className="custom-class" />);
+      expect(screen.getByRole('listitem')).toHaveClass('custom-class');
```
tay1orjones: Same thing here with this not testing the outmost element as intended

tay1orjones (changes requested):
Thanks for making another revision on this. I think the bulk of the node errors are places where we intentionally want to test the node. I outlined a couple examples below.

I'm in support of placing eslint ignore comments for the `testing-library/no-node-access` rule on all the failures for now. We can always iterate towards the ideal in future PRs.

tay1orjones:
Interestingly, the rule has an option we could turn on for `firstChild` usage: https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-node-access.md#options

We could turn this on and still ignore the remaining warnings for now in aim to get this PR in.

packages/react/src/components/Accordion/__tests__/AccordionItem-test.js
falsepopsky: I'm not understanding how `outermost` should work here, are you saying the parent component of `<AcordionItem />`?

i'm using debug with the old code and i'm getting the same result:


**container.firstChild**
```
    it('should spread extra props onto outermost element', () => {
      const { container, debug } = render(<AccordionItem data-testid="test-id" />);
      
      debug(container)

      expect(container.firstChild).toHaveAttribute('data-testid', 'test-id');
    })
    
 ```
 
container_debug


**screen**
 
screen_debug

**Comment by falsepopsky** [comment]
> Interestingly, the rule has an option we could turn on for `firstChild` usage: https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-node-access.md#options
> 
> We could turn this on and still ignore the remaining warnings for now in aim to get this PR in.

i'm ok with that rule but i suggest to at least take a read at this, specially at [Not using screen](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen):

[Common mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)

File: packages/react/src/components/Accordion/__tests__/AccordionItem-test.js (hunk 1 above)
Comment (tay1orjones) [reply packages/react/src/components/Accordion/__tests__/AccordionItem-test.js]: By "outermost" I mean the top-level parent element within the component. Our convention is that data-testid's should always be placed on the root/parent element of a component and not placed on inner children elements.

Given a simple component, such as

```js
const MyComponent = (props) => (
  <div data-testid={props['data-testid']}>
    <span>sample component</span>
  </div>
);
```

If `data-testid` was moved in this component source to no longer be on the outermost element

```js
const MyComponent = (props) => (
  <div>
    <span data-testid={props['data-testid']}>sample component</span>
  </div>
);
```

The test using `firstChild` would fail, as we want it to. The test using `screen.getByTestId` would still pass, which is not what we want.

tay1orjones:
@falsepopsky Yes! That's a great resource we're familiar with. We want to default to using `screen` where possible. There are only a few limited cases (such as the "outermost element" tests for `className` and `data-testid`) where we have a requirement to use `container.firstChild`.

falsepopsky:
[ESLint Report](https://falsepopsky.github.io/carbon-eslint-errors/)

---

- bump(deps) `@typescript-eslint/eslint-plugin`, `typescript-eslint/parser`, `eslint`, to remove noisy warnings from console.
- Any changes previously made to the `outermost` tests have been reverted.
- The `allowContainerFirstChild` rule has been added

---

There are some errors that are difficult to refactor or rewrite to fix those, as many of them involve a mix of using querySelector, removeChild(), .parent, etc. on the container. In some cases, using `test-id` or `role` may be a better option, but it may not always be good because the tests are focused on one of the children, parent nodes, or other elements (outermost, etc...).

This can lead to confusion and wasted time trying to find a way to access the relevant element without breaking changes for both sides(lint,test), especially if a test for component 'A' is actually testing component 'B'.

tay1orjones:
@falsepopsky Just wanted to check in with you - did you need some help resolving the remaining errors? Looking through that last ESlint report, I think we could put eslint ignore comments on the remaining erroring lines. We can selectively improve those situations in future PRs.

I just want to be sure this PR lands. The work you've done is substantial and I don't want it to go stale.

falsepopsky:
that's it, added [eslint-disable-next-line rules](https://github.com/carbon-design-system/carbon/pull/13528/commits/e7444e94941bb330c7fbffd59afea5c980ce3011) to prevent errors in CI, so better to checkout those later...

tay1orjones (approved):
Thank you so much for all your work on this! 🏆 🎉

alisonjoseph (approved):
nice! LGTM 🔥

PR #13529, opened 2023-04-14
feat(Tabs): Dismissable tabs

### LINKED ISSUE #12813: Tabs: add dismissable functionality to tab
This issue is for the implementation of dismissible functionality for tabs.

Open questions:
- There's lots of conversation around dismissible tabs with regards to accessibility:
  - https://github.com/w3c/aria/issues/1630
  - close button probably doesn't have to be in the keyboard nav, could be aria-hidden
  - but it could be if we'd like to give the user more affordance to discovery of the delete function
  - aria `tab` role spec says children are presentational only
- Does the design allow for users to put non-operable icon tabs next to dismissasble tabs in the tablist?
  - A: No, the entire tab group/list must be all the same type
- Will this actually unmount the tab component, or will it just be a handler?
  - We'll probably want to provide users a callback that they can use to unmount the component
  - This should support consumer's ability to add a confirmation dialog or similar before actually closing/unmounting the tab.
  - This also will need to support the use case of an tab containing unsaved work, to prompt for save, and/or allow devs to save data in the background
- Does this apply to ContainedTabs? Can they be dismissed?
  

https://github.com/carbon-design-system/carbon/issues/12813#tasklist-block-abb364de-c55b-4db3-93ea-8939819821a2

francinelucca:
Closes #12813 

This adds dismissable functionality to tabs

- Adds new `dismissable` & `onTabCloseRequest` props to `Tabs`
- Adds new `Dismissable` Tabs story
- Adds Dismissable tabs docs guidance to Tabs story
- Adds VRT & RTL tests for dismissable tabs

TODO:
- [x] Confirm disabled state
- [x] Confirm hover state

#### Changelog

**New**

- `dismissable` & `onTabCloseRequest` props in `Tabs`, also passed through the `TabsContext`
- dismissable VRT story
- RTL tests for dismissable tabs
- Tabs docs guidance for dismissable tabs
- Dismissable tab story
- close button styling rule for dismissable tab
- `Dismissable` control in `Tabs` Playground story

**Changed**

- Refactored `Tabs` scroll buttons visibility logic to account for deletable tabs
- Datatable docs package import correction (use `@carbon/react` instead of `carbon-components-react`)
- Overrider renderIcon with Dismiss(close) Icon on `Tab` when `TabsContext`'s value has `dismissable`
- `onKeyDown` handle in `Tab` to call `onTabCloseRequest` when `delete` is pressed on dismissable tab

#### Testing / Reviewing

1. https://deploy-preview-13529--carbon-components-react.netlify.app/?path=/story/components-tabs--dismissable
- Verify styling matches spec
image
- Verify dismissable behavior
- Verify reset behavior
- Verify keyboard functionality

2. Verify other tab stories still work/behave as expected and no visual/behavioral regressions
3. Verify Dismissable docs are coherent and readable https://deploy-preview-13529--carbon-components-react.netlify.app/?path=/docs/components-tabs--dismissable#dismissable-tabs


francinelucca:
Not sure why `icons-vue` is failing in the e2e testing with `ReferenceError: Cannot access 'WatsonHealth3DCursor32' before initialization`, any insight would be greatly appreciated 🙏🏻

francinelucca:
Hey @laurenmrice @kingtraceyj I have some questions

1. Can contained tabs be dismissable? In my current implementation they are but I can add a condition to ignore the prop for contained if that's what we want.
2. Should the close button have a hover state? I've added one but I didn't see one in the specs, I can remove it if it shouldn't have it. Also what would the correct values be? the defaults look good to me but the contained one is definitely wrong. I'm using `$layer-hover` for both.
3. Should disabled tabs still show the close icon? In my implementation they do but I have a feeling that's wrong, just confirm and I can not show it on disabled tabs.

I've attached a screen record demonstrating all of the above for reference

https://user-images.githubusercontent.com/40550942/233253792-e6e5c6ef-bc8f-4a6f-93a8-95cf8c043dee.mov

kingtraceyj:
sorry for the delay on this @francinelucca 

1. We will have dismissible contained tabs. Instead of having layered hover states, we are going to have separate hover touch zones.
- hover states/touch targets need to be 24px X 24px for accessibility reasons
- when a user hovers over the main content of an enabled tab, the whole tab will get the hover state `$layer-accent-hover`
image
- when a user hovers over the 'x' of the tab, only the 24px touch zone will get the hover state `$layer-accent-hover`
image
- the only hover state for an active tab is the 24px X 24px touch target `$layer-hover`
image

2. Line tabs will also get the 24px hover state. checking right now if we want to go with `$background-hover` or `$layer-hover` token

3. We will allow for a disabled dismissible tab. We are leaning this way because we don't necessarily know when and how the tab becomes dismissible and would be unique to product teams. It would be up to the teams to decide whether to show that or no

francinelucca:
This should be ready for another visual review @kingtraceyj @laurenmrice 
Line tabs: http://localhost:3001/?path=/story/components-tabs--dismissable
Contained tabs: http://localhost:3001/?path=/story/components-tabs--playground&args=dismissable:true;contained:true

Attaching some videos as well

https://github.com/carbon-design-system/carbon/assets/40550942/ce9a7058-d5de-40bc-82bf-6613b5eb6b5b


https://github.com/carbon-design-system/carbon/assets/40550942/677ac0e0-bb2d-4921-ac07-440aada9284c



Still wondering if the disabled dismiss button should have an active color or not since disabled tabs are dismissable?
image

kingtraceyj (changes requested):
hover color on 'x' should be $layer-hover-accent

current:
image

intended:
image

kingtraceyj (approved):
looks go to good! 💥

andreancardona (approved):
🔥 looks amazing!

andreancardona (approved):
🔥 looks amazing!

guidari (approved):
Awesome! LGTM!!

PR #13530, opened 2023-04-15
fix(DataTable): prevent 2 rerenders when updating props

### LINKED ISSUE #13489: [Bug]: DataTable renders twice when rows change
### Package

@carbon/react

### Browser

Chrome

### Package version

1.26.0

### React version

18.2.0

### Description

When `props.rows` or `props.headers` of a DataTable is updated, the component re-renders twice.

The first render is initiated by React as it notices the props changed, the second re-render apparently happens because the DataTable uses `componentDidUpdate` to detect if the rows or headers changed. If so, it uses `this.setState` to sync the updated props with the internal state.

This causes the component to re-render first with the old state and then again with the new state.
It can lead to problems when within the render function of the DataTable one tries to access more meta information about a row, for example for additional styling or interactivity.

More concrete illustration: if I pass 6 rows into the DataTable and want to reference the original entry like so:

```jsx
// <DataTable rows={data} header={headers}>…

rows.map((row, index) => {
  const reference = data[index];

  return (
    <TableRow key={row.id}>…</TableRow>
  );
})
```

the reference would be undefined if I remove one row from the `data` array as the DataTable would render once with the old 6 rows, and only then with the new 5 rows. Obviously I can build fallbacks to ignore the issue the first time but that doesn't seem very elegant and can cause confusion when debugging.

### Reproduction/example

https://codesandbox.io/s/musing-cache-hmi9jq?file=/index.js

### Steps to reproduce

1. Click the "Include last row" button
2. A new row appears
3. Click the "Exclude last row" button
4. Error is thrown

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

pmweeks98:
Closes #

Fixes #13489 
Switched the state change to happen before the render so only one render is required on props change. 

#### Changelog

**Changed**

- switched state change  from `componentDidUpdate` to `shouldComponentUpdate`

#### Testing / Reviewing

Checkout this branch
Modify the story file to contain the reproducible example in the bug
see console logs


francinelucca (approved):
LGTM!

alisonjoseph (approved):

PR #13609, opened 2023-04-17
fix(structured-list): alignment prop, skeleton and storybook cleanup

### LINKED ISSUE #11149: Structured list: alignments
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox

### Package version

v11

### Description

- The structured list component should only offer the `flush` alignment for the `simple` variant. The states for the selectable variant with flush alignment start to get tricky visually. We need to update structured list to only support the flush alignment for the simple variant.

---

### Structured list
This is an outline of what the modifiers should be per variant:
- **Simple variant**
  - Hang alignment
    - Default size
    - Condensed size
  - Flush alignment
    - Default size
    - Condensed size
- **Selectable variant**
   - Hang alignment
      - Default size
      - Condensed size

### Screenshots

 In the newest React storybook, I do not see the props anymore.

### Steps to reproduce

Currently I do not see the props in React for the different alignments, so those should be added in. 

### Code of Conduct

alisonjoseph:
Closes #11149

#### Changelog

**New**

- clean up storybook controls / props shown

**Changed**

- update `StructuredList` so `isFlush` prop no longer affects `selection` variant
- fix alignment of checkmark icon so it is top/right with 16px padding (12px for condensed)
- add controls to default and selection story

**Removed**

- old storybook file
- unused border prop from `StructuredListSkeleton` component
- playground story

#### Testing / Reviewing

https://deploy-preview-13609--carbon-components-react.netlify.app/?path=/story/components-structuredlist--default

`StructuredListSkeleton` story should only show the `rowCount` prop
Playground story removed, selection and default stories should have controls. `isFlush` is hidden from story but should no longer affect the component when `selection` is true.


laurenmrice (changes requested):
It looks like the checkmark icon in the selected row state does not appear anymore on the right side of the row. It should still be there.
Frame 1

alisonjoseph:
Screenshot 2023-04-18 at 1 13 55 PM

@laurenmrice the selection variant is composed of different components, so it won't show up in the playground story., It requires more than adding/removing a prop. I wonder if we should remove the Playground story for this component and just enable the controls for each variant? Similar to what we do with Button variants?

Also, looks like the alignment is off for the checkbox, I can fix that in this PR too.
Screenshot 2023-04-18 at 1 16 52 PM

laurenmrice:
@alisonjoseph Oh okay I see, thanks for clarifying that.  I would be fine with removing that functionality from the playground story and enable the controls for each variant if that makes it more consistent.

alisonjoseph:
updated @laurenmrice

laurenmrice (approved):

tay1orjones (approved):

tw15egan (approved):
Looks great! 👍🏻 ✅

PR #13614, opened 2023-04-17
fix(SideNav): added esc key to close the sidenav

### LINKED ISSUE #13414: Left UI Shell: The responsive menu is not dismissable with Esc
From the parent issue:

> Problem 2: The responsive menu is not dismissable with Esc
> When the hamburger menu is activated, it becomes an X, which can be activated to close the left nav. This is an awkward implementation of a close function, which really doesn't work well with this disclosure pattern. There is no good reason not to cause the left nav to be dismissed on pressing Esc, and there are several good reasons to do it:
> 
> It allows the user to dismiss the temporarily displayed left nav from anywhere inside the menu
> Because the focus does not immediately go into the revealed left nav, but instead traverses the site, it is even more difficult for a keyboard user to reach the X to dismiss the nav
> As discussed in the next problem, because the main content is partially obscured with a light-box effect, the inability for the keyboard user to dismiss the left panel, either through Esc or by tabbing out of the left nav, becomes more problematic.

guidari:
Closes #13414 

Added an event listener for ESC key. When the ESC key is pressed inside the `HeaderContainer` or `SideNav` component the sidenav is closed

#### Testing / Reviewing

To access the hamburger menu, switch to the mobile version and use your keyboard keys (Space or Enter) to open it. Once the menu is open, you can close the sideNav by pressing the ESC key.


packages/react/src/components/UIShell/HeaderContainer.js
```diff
@@ -14,6 +14,14 @@ const HeaderContainer = ({ isSideNavExpanded, render: Children }) => {
   const [isSideNavExpandedState, setIsSideNavExpandedState] =
     useState(isSideNavExpanded);
 
+  // Closing the SideNav when the 'ESC' key is pressed
+  document.addEventListener('keydown', function (event) {
+    if (event.key === 'Escape') {
```
francinelucca: we generally like to use our `match` helper to compare against keys:
image
You can import it in `packages/react/src/internal/keyboard/match.js`

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -40,6 +40,16 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
   const [expandedViaHoverState, setExpandedViaHoverState] =
     useState(defaultExpanded);
   const expanded = controlled ? expandedProp : expandedState;
+
+  // Closing the SideNav when the 'ESC' key is pressed
+  document.addEventListener('keydown', function (event) {
+    if (event.key === 'Escape') {
```
francinelucca: same thing here about `match` util

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -40,6 +40,16 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
   const [expandedViaHoverState, setExpandedViaHoverState] =
     useState(defaultExpanded);
   const expanded = controlled ? expandedProp : expandedState;
+
+  // Closing the SideNav when the 'ESC' key is pressed
+  document.addEventListener('keydown', function (event) {
```
francinelucca: same comment here as above about `useEvent` and attaching the listener in the DOM

packages/react/src/components/UIShell/HeaderContainer.js
```diff
@@ -14,6 +14,14 @@ const HeaderContainer = ({ isSideNavExpanded, render: Children }) => {
   const [isSideNavExpandedState, setIsSideNavExpandedState] =
     useState(isSideNavExpanded);
 
+  // Closing the SideNav when the 'ESC' key is pressed
+  document.addEventListener('keydown', function (event) {
```
francinelucca: I think this eventListener needs to be cleaned up on unmount, the `useEvent` hook (packages/react/src/internal/useEvent.js) should be able to handle all that for you 
image

Also, is there an element we can attach this event to or does it need to be the whole document? Like, could we add `onKeyDown` events to the `Children` instead?

francinelucca (changes requested):

packages/react/src/components/UIShell/HeaderContainer.js
guidari: I'm gonna check if we can attach to an element to improve that

guidari: I changed to `useEvent` hook on `HeaderContainer` component and attach to an eventHandler on `SideNav` component. 
Thanks for the suggestions, now looks better!

francinelucca (approved):

tw15egan (approved):
LGTM 👍 ✅

lewandom:
@guidari this change breaks server-side rendering: 

```
info  - Creating an optimized production build .ReferenceError: window is not defined
    at HeaderContainer (/home/mlewandowski/git/ibm_github/developer-profession/guidepost/.next/server/chunks/166.js:29753:23)
```

guidari:
Hey @lewandom, thanks for letting me know!

After doing some research, I have found that switching from using `window` to `document` can be a beneficial strategy when working with server-side rendering.

How did you catch that error? I would like to know if I can repro it

lewandom:
@guidari thanks! I'm using Next.js in my project, it comes with SSR by default.

lewandom:
@guidari just a thought: could you try using `useWindowEvent` helper function instead of `useEvent`? I think it may be a suitable approach, since the implementation seems to be entirely within `useEffect` calls.

guidari:
Thanks for the suggestion @lewandom 

I open a new PR to fix this issue, if you wanna check it out.

I also did a sample code in [StackBlitz](https://stackblitz.com/edit/nextjs-cqbapt?file=pages%2Findex.js,pages%2FuseEvent.js)

Issue #13424, opened 2023-03-29
React 18 Enhancements: Refactor defaultProps to default function parameters


tay1orjones:
React 18.3 will introduce a deprecation of defaultProps. Components will need to move off defaultProps and instead use default function parameters. 

Additionally, we'll need to ensure that default values are static https://github.com/carbon-design-system/carbon/issues/10777 - this mostly includes function props

We probably should wait to tackle this until React 18.3 lands to see if the react team has additional recommendations for this work and/or a codemod available to automate this refactor.

```js
// TODO: provide code example
```

```[tasklist]
### Components to refactor
- [x] Use default function parameters in `.../icon-build-helpers/src/builders/react/components/Icon.js`
- [x] Use default function parameters in `.../react/examples/custom-data-table-state-manager/src/components/CustomDataTablejs:`
- [x] Use default function parameters in `Accordion/Accordion.Skeleton.js`
- [x] Use default function parameters in `Checkbox/Checkbox.tsx`
- [x] Use default function parameters in `CodeSnippet/CodeSnippet.js`
- [x] Use default function parameters in `ComboBox/ComboBox.tsx`
- [x] Use default function parameters in `ComposedModal/ComposedModal.js`
- [x] Use default function parameters in `ComposedModal/ModalFooter.js`
- [x] Use default function parameters in `ComposedModal/ModalHeader.js`
- [x] Use default function parameters in `ContentSwitcher/ContentSwitcher.js`
- [x] Use default function parameters in `Copy/Copy.js`
- [x] Use default function parameters in `CopyButton/CopyButton.js`
- [x] Use default function parameters in `DataTable/DataTable.js`
- [x] Use default function parameters in `DataTable/Table.js`
- [x] Use default function parameters in `DataTable/TableBatchAction.js`
- [x] Use default function parameters in `DataTable/TableBatchActions.js`
- [x] Use default function parameters in `DataTable/TableBody.js`
- [x] Use default function parameters in `DataTable/TableHeader.js`
- [x] Use default function parameters in `DataTable/TableSelectAll.tsx`
- [x] Use default function parameters in `DataTable/TableToolbar.tsx`
- [x] Use default function parameters in `DataTable/TableToolbarMenu.tsx`
- [x] Use default function parameters in `DataTable/TableToolbarSearch.js`
- [x] Use default function parameters in `DataTableSkeleton/DataTableSkeleton.tsx`
- [x] Use default function parameters in `Dropdown/Dropdown.tsx`
[...]

PR #13615, opened 2023-04-17
Types/filterable multi select

### LINKED ISSUE #12546: Add TypeScript types to FilterableMultiSelect / use `useCombobox` instead of `<Downshift>` component

kubijo:
Closes #12546

OK, gentlemen and ladies… in the words of one wise old goat head… `Ooph, this a fine mess!` so right now I'm just submitting the PR, leaving a few comments to give some context for my changes and going away from the computer as it's pretty late here...

I'm expecting some pushback here anyway, so I'll clean it all up based on the feedback.

#### Changelog

**New**

- add types to `FilterableMultiSelect`


packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -26,9 +41,179 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+
+  /**
+   * Additional props passed to Downshift
+   */
+  downshiftProps?: DownshiftProps<Item>;
+
+  /**
+   * Default sorter is assigned if not provided.
+   */
+  filterItems(
```
kubijo, line 88: This one is optional too, but it seems that the proper way to get `intellisense` working when the given prop has a `defaultProps` entry is to omit the optionality

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -26,9 +41,179 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+  /**
+   * Specify whether the title text should be hidden or not
+   */
+  hideLabel?: boolean;
+
+  /**
+   * Provide helper text that is used alongside
+   * the control label for additional help
+   */
+  helperText?: ReactNode;
```
kubijo, line 107: This was I think one of the cases where `propTypes` do not declare a prop that is expected in the component body… (I won't bother with listing the other ones)

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -26,9 +41,179 @@ import { useFeatureFlag } from '../FeatureFlags';
...
+  /**
+   * Provide the text that is displayed when the control is in warning state
+   */
+  warnText?: ReactNode;
+}
+
+function FilterableMultiSelectRenderFunction<Item extends ItemBase>(
   {
     ['aria-label']: ariaLabel,
+    // @ts-expect-error: deprecated prop
```
kubijo: My thinking was that we shouldn't propagate deprecated props to typescript interfaces, since users shouldn't use them anyway

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -75,7 +261,7 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
   const [currentSelectedItems, setCurrentSelectedItems] = useState(
     initialSelectedItems ?? []
   );
-  const textInput = useRef();
+  const textInput = useRef<HTMLInputElement>();
```
kubijo: Consumers are otherwise oblivious of the `current` element type and shout about it

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -179,15 +363,11 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
     }
   }
 
-  function handleOnInputValueChange(inputValue, { type }) {
+  function handleOnInputValueChange(inputValue: string) {
```
kubijo: I've been looking for this mysterious `type` attribute and haven't been able to find it anywhere for this handler… My hunch is that someone has mistaken this with some kind of `stateChange` handler, where it might exist and since this evidently wasn't checked by anything it slipped by

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -244,6 +424,7 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
             inputValue,
             selectedItem,
           }) => {
+            const selectedItems = (selectedItem || []) as Item[];
```
kubijo: This was another of those WTF moments where I could not for the life of me find any way to get this properly typed as an array from `Downshift` without resorting to this kind of ugly hack… 

I'm open to suggestions, but since you're using a pretty ancient version of `Downshift`, the amount of work for a proper implementation might very soon get too high…

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -293,7 +471,7 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
               },
             });
 
-            const inputProps = getInputProps({
+            const inputProps: object = getInputProps({
```
kubijo: I don't have a clue as to why, but ts computes the return type to `never`… but this cast seems good enough™ since we don't really care about the precise shape.

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -380,8 +558,8 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
                 ) : null}
                 <ListBox
                   aria-label={deprecatedAriaLabel || ariaLabel}
-                  onFocus={isFluid ? handleFocus : null}
-                  onBlur={isFluid ? handleFocus : null}
+                  onFocus={isFluid ? handleFocus : undefined}
+                  onBlur={isFluid ? handleFocus : undefined}
```
kubijo: This one generally encountered when converting to TS since it's pretty usual to specify optional prop as `prop?: type` and that doesn't cover null's… [some might say that is a good thing](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/) and I tend to agree just based on simplicity on the definition side

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -502,8 +686,32 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
       )}
     />
   );
-});
+}
+
+type PropsWithRef<P, T> = PropsWithoutRef<P> & RefAttributes<T>;
```
kubijo: So… here be dragons… My reasoning is explained in the code because future generations might reasonably get perplexed & possibly even face-slappingly angry without some explation for such fugliness...

This is, however, at least to my knowledge, one of those shining examples where typescript is just the best bad thing we have.

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -162,8 +348,6 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
     switch (type) {
       case stateChangeTypes.keyDownArrowDown:
       case stateChangeTypes.keyDownArrowUp:
-      case stateChangeTypes.keyDownHome:
-      case stateChangeTypes.keyDownEnd:
```
kubijo: These don't exist in the enum

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
tay1orjones: We're going to begin moving away from defaultProps to use default function params, #13424  

If it helps with this, feel free to refactor this to use default function params

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
tay1orjones: Yeah ideally everything expected would be declared in propTypes

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
tay1orjones: I see your [comment above](https://github.com/carbon-design-system/carbon/pull/13615/files#diff-a1249d25d8548f5c2c1a100e428df6fb0f881e7f213b96994153fc4aede57f33R146-R149) too. This mismatch could be because we're using v5.x downshift and latest is v7.x?

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
tay1orjones: Totally agree and we can always improve this in the future. How do you feel about renaming it to `_selectedItems` or something? I think it's confusing to have `selectedItems` declared twice, once in the render prop function of Selection, and then again here in this scope.

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
tay1orjones: Yep I agree this is fine

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
tay1orjones: Thanks for commenting it so well. @mbarrer or @jdharvey-ibm would either of you have thoughts on situations like this?

tay1orjones (approved):
I don't see anything that would block a merge for this one, but I do have some suggestions for changes.

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
jharvey10: Some food for thought when looking at this from a user's perspective. Here's the error you get when using the `FilterableMultiSelect` component:

image

Compare that to using `FilterableMultiSelectRenderFunction` directly:

image

This isn't a deal breaker, but it is definitely a more confusing error in the first image than the second.

It looks like this might also introduce an additional component render into the tree, which could be a minor debugging and performance hit. I didn't verify this, but the:
```tsx
const Comp = forwardRef<HTMLDivElement, Props>(
  FilterableMultiSelectRenderFunction
);
return <Comp {...props} />;
```
makes me think that might be the case.

I feel like we had a similar discussion to this for the `ComboBox` before I went on leave at the end of last year. Were there any directional decisions made on this type of thing in that PR?

kubijo: Yeah, the type printing is a mess!

Honestly, I'd rather forgo the `forwardRef` wrapper entirely and just offer a new prop in lieu of `rootRef` / `inputRef` / …

This, in my view, needlessly complicates things on many levels… some of which are the generics / extra wrappers / differences in exports

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
kubijo: Nice to hear that, but I'm not really eager to balloon this already quite complex refactor…

How does is sound to you to delegate this one to the task you linked, as far as it is not needed for the typescript addition itself (e.g. so that it doesn't produce type errors)

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
kubijo: And would you expect me to add those in too? This one is easy, obviously, but that is not a given in general & I have completely stopped using propTypes a long time ago, so I'm not eager to jump into that if you're fine with `it was wrong until now, it can die with propTypes`

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
kubijo: Dunno… your guess is as good as mine, I think.

What I know for certain is that Downshift upgrade is not a joke… also fixing one handler the other way around can quickly spiral to a huge refactor due to many call sites and the messy hooks.

francinelucca (changes requested):
I'm seeing some differences in the component's behavior in this PR (first) vs Prod (second), specifically the "Option x" input text

https://github.com/carbon-design-system/carbon/assets/40550942/36567a00-060b-4a62-84d6-2a2663f013a8

kubijo:
> Hey there, it's been a couple months since this was last updated. We'd love to get it in but there's some remaining pieces of feedback that need addressed.

I've rebased the branch to current upstream & addressed what I deemed actionable, but it was just a one nitpick ... other than that, it seems like the proverbial conversation ball is on your side guys…

kubijo:
@tay1orjones I can give it a shot... Someone will ping it here, right?

kubijo:
> #14276 has been merged now, @kubijo.

So… I've rebased the branch onto the current `upstream/main` state and resolved conflicts / new errors, however right after running `yarn run test` I'm getting the following message (twice):

```
The exported identifier "MultiSelectProps" is not declared in Babel's scope tracker
as a JavaScript value binding, and "@babel/plugin-transform-typescript"
never encountered it as a TypeScript type declaration.
It will be treated as a JavaScript value.

This problem is likely caused by another plugin injecting
"MultiSelectProps" without registering it in the scope tracker. If you are the author
 of that plugin, please use "scope.registerDeclaration(declarationPath)".
```

Any idea what is that about?

kubijo:
I've just spent 4 hours trying to resolve this fucking nightmare without much success… 
- The `Downshift` component is the same mess as it was before
- their recommended way of creating a searchable multi-select is a combination of both of their hooks (`useCombobox` & `useMultipleSelection`)
  - when I tried that according to their docs, I got an absolute mess of errors from your tests… this state was just pushed into my branch for anyone interested… I am now going for a walk and will attend to my work tomorrow because this just causes me to lose my already limited my sanity.

tay1orjones:
@kubijo let me know how you'd like to approach this moving forward. I'm happy to leave open if you'd like to continue to pull on this thread. Totally fine if not, we can put it back on the queue for someone else to pick up if you're exasperated. I really appreciate all the effort you've put into this so far 💙

kubijo:
> @kubijo let me know how you'd like to approach this moving forward. I'm happy to leave open if you'd like to continue to pull on this thread. Totally fine if not, we can put it back on the queue for someone else to pick up if you're exasperated. I really appreciate all the effort you've put into this so far 💙

Hey man... your assessment of my feeling towards this issue is quite correct, and I do appreciate the kind words of respect.

As for the way forward, I'd personally just rewrite the whole thing from the ground up with a typescript-first composable approach.
The incremental approach seems to be just too much here. One only gets tangled up in the heaps of legacy code mess, and downshift doesn't help with it in the slightest.

All that said, I will have to let this one go to someone with more energy and less frustration to keep my remaining sanity.

tay1orjones:
@kubijo I hear you, thanks again for taking a look at this. It won't all be a waste as we'll still use this PR as a point of reference for future work.

PR #13618, opened 2023-04-18
fix(grid): improve and document use-flexbox-grid feature flag

### LINKED ISSUE #13617: [Feature Request]: Improve and clarify how FlexGrid styles can be emitted via `$use-flexbox-grid`
### The problem

The styles for FlexGrid are not emitted by default. To emit them consumers must do one of the following:

- Turn on the `$use-flexbox-grid` feature flag (this consequently turns off CSS Grid)
- Explicitly `@use` the flexbox module `@use @carbon/styles/scss/grid/flexbox`
- `@use '@carbon/react/grid';` and call the `flex-grid()` mixin

Here's how the feature flag is set up currently:

https://github.com/carbon-design-system/carbon/blob/36685a0be5e62db11a60d938c76b1c49a402c99a/packages/styles/scss/grid/_index.scss#L13-L17

None of this is documented in the [storybook docs for FlexGrid](https://react.carbondesignsystem.com/?path=/docs/elements-flexgrid--auto-columns#grid) that I can see.

### The solution

I think we should modify the feature flag to only impact the emitting of flex-grid code and always include CSS Grid code.

```
@if config.$use-flexbox-grid == true {
  @include grid.flex-grid();
}

@include grid.css-grid();
```

```[tasklist]
- [x] Update the feature flag conditional
- [x] Update the documentation
```

--- 

### Why not emit FlexGrid styles by default?

- FlexGrid is a less-complete implementation of the Grid specification
- FlexGrid doesn't support nesting
- FlexGrid requires the use of extra markup via `Row`
- While FlexGrid has it's uses, the CSS Grid implementation is the intended entrypoint for new projects. 

For these reasons, I believe the FlexGrid styles should continue to require an explicit opt-in.

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/13617

#### Changelog

**Changed**

- Update `use-flexbox-grid` to only control emitting flex grid styles
- Always emit CSS Grid styles
- Update v11 migration docs
- Update FlexGrid.mdx storybook docs

#### Testing / Reviewing

- Ensure the docs make sense


alisonjoseph (approved):
Thanks so much for doing this!

andreancardona (approved):
🔥

PR #13626, opened 2023-04-19
 Add TypeScript types to OverflowMenu, OverflowMenuItem #13237 

### LINKED ISSUE #13237: Add TypeScript types to OverflowMenu, OverflowMenuItem

kubijo:
Closes #13237 

#### Changelog

**New**

- add types to `OverflowMenu`


kubijo:
In IDE and when I try to run storybook I have a problem with being unable to import `@carbon/icons-react`. Does anybody have a clue what could be the issue?

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -92,8 +116,123 @@ export const getMenuOffset = (menuBody, direction, trigger, flip) => {
...
+
+  /**
+   * The event handler for the `focus` event.
+   */
+  onFocus(): void;
+
+  /**
+   * The event handler for the `keydown` event.
+   */
+  onKeyDown(): void;
```
francinelucca: can we specify the event arg here? I would also check the other `on...` functions

packages/react/package.json
```diff
@@ -89,6 +89,7 @@
     "@storybook/manager-webpack5": "^6.5.6",
     "@storybook/react": "^6.5.6",
     "@storybook/theming": "^6.5.6",
+    "@types/react-is": "^17.0.3",
```
tay1orjones, line 92: Instead of having this listed in devDependencies, I think this can instead be placed under `resolutions` in the root `package.json` alongside the types packages for react and prop-types

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -92,8 +116,123 @@ export const getMenuOffset = (menuBody, direction, trigger, flip) => {
...
+type ChildElement = ReactElement<
+  OverflowMenuItemProps,
+  typeof OverflowMenuItem
+>;
+
+export interface OverflowMenuProps extends HTMLAttributes<HTMLElement> {
+  /**
+   * The child nodes.
+   */
+  children: MaybeArray<ChildElement>;
```
tay1orjones: We can't restrict `children` to only be OverflowMenuItem(s).

```suggestion
  children?: React.ReactNode;
```

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -369,26 +501,26 @@ class OverflowMenu extends Component {
...
+    const enabledIndices: number[] = [];
+    React.Children.forEach<ChildElement>(this.props.children, (child, i) => {
+      if (
+        isElement(child) &&
+        child.type === OverflowMenuItem &&
+        !child.props.disabled
+      ) {
+        enabledIndices.push(i);
+      }
+    });
```
tay1orjones: Same thing here - we can't require that the children be OverflowMenuItem. Why is the additional check for `isElement` necessary?

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -584,9 +718,12 @@ class OverflowMenu extends Component {
...
-  return React.forwardRef(forwardRef);
-})();
+const WithFwdRef = React.forwardRef<HTMLElement, OverflowMenuProps>(
+  (props, ref) => {
+    return <OverflowMenuComponent {...props} innerRef={ref} />;
+  }
+);
+WithFwdRef.displayName = 'OverflowMenu';
+
+export const OverflowMenu = WithFwdRef;
```
tay1orjones, line 720: Previously the named export did not forward a ref. Only the default export used forwardRef.

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -584,9 +718,12 @@ class OverflowMenu extends Component {
...
+export default WithFwdRef;
```
tay1orjones, line 721: I'd prefer to not rename the component from what it is now. It's possible consumers are relying on the existing name in tests or otherwise, even with the `displayName` configured the same. Would something like this work instead?

```suggestion
export { OverflowMenu };
export default (() => {
  const forwardRef = (props, ref) => <OverflowMenu {...props} innerRef={ref} />;
  forwardRef.displayName = 'OverflowMenu';
  return React.forwardRef<HTMLElement, OverflowMenuProps>(forwardRef);
})();
```

tay1orjones (changes requested):
Thanks for this! Just a few changes below I found. Let me know what you think.

Also, the only _required_ prop I see within propTypes is the iconDescription. Should the rest of them be marked optional with a `?` appended to them?

tay1orjones:
> In IDE and when I try to run storybook I have a problem with being unable to import @carbon/icons-react. Does anybody have a clue what could be the issue?

@kubijo A few things to try: 

* Ensure you're using `yarn`
* Run `yarn` at the root, followed by `yarn build`

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
kubijo: Yep, I have noticed it and considered it an oversight… however I haven't considered a chance of it being intetional :thinking: … is it? What would be the reason?

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
kubijo: Because otherwise the `child` variable is an expansive type including crap like `boolean`, `string`, `number` and whatnot… the `isElement` narrows the type to `ReactElement`

kubijo:
> Thanks for this! Just a few changes below I found. Let me know what you think.
> 
> Also, the only _required_ prop I see within propTypes is the iconDescription. Should the rest of them be marked optional with a `?` appended to them?

Hard to say really ... I'd say this is more for you to decide what really is required ... it's pretty hard to some conditional required logic based on props without introducing complexity that just is not worth it

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
kubijo: What is the logic behind not having the ref forwarding on named export, though? I don't really see the use-case, but I'll be the first to admit that my scope is pretty limited …

Also, why the extra function wrapper?

packages/react/package.json
kubijo: `resolutions` are, AFAIK, for version enforcing… right now, the `types` package wasn't there at all & I've been led to believe that dev-dependencies is the place for `@types/` packages since that is what yarn's typescript plugin does, and it makes an intuitive sense to me… that said I will, of course, concur to what is your way of doing this.

francinelucca (changes requested):

packages/react/package.json
kubijo: So, I have just tried what you proposed, but as expected, that doesn't work… resolutions only force versions, they don't add dependency if it's not already declared somewhere

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
kubijo: Btw, do you guys have some prepared lint/test job that runs `tsc` in check-only mode? I haven't been able to find any in the `package.json`s' scripts

tay1orjones: My recollection is that the internal typescript working group had determined that we wouldn't be able to put ci/git-hook checks in place until types were being provided for all components

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
tay1orjones: There probably is not a good reason for it honestly. It's unlikely, but someone could be relying on the named export via importing the filepath. Changing it to use forwardRef could break that usage.

```js
@import { OverflowMenu } from '@carbon/react/es/components/OverflowMenu/OverflowMenu';
```

I assume the extra function wrapper was in there simply so the forwardRef component could be given the `OverflowMenu` displayName.

PR #13627, opened 2023-04-19
feat(Storybook): Enable "buildStoriesJson" to support composition #13506

### LINKED ISSUE #13506: [Feature Request]: Storybook composition support
### The problem

I would like to use [storybook-composition](https://storybook.js.org/docs/react/sharing/storybook-composition) in our storybook instance to be able to seamlessly browse your reference instance.

When I, however, try that by adding the following section to our storybook config…

```js
module.exports = {
    ...config,
    refs: {
        'design-system': {
            title: 'Carbon Design System',
            url: 'https://react.carbondesignsystem.com',
            expanded: false,
        }
    }
}
```

…I only get this error

image

- Clicking on the first entry in the popup takes me to your storybook instance and it renders fine
- Our storybook version is `7.0.2`

### The solution

I would like the feature to work as described in https://storybook.js.org/docs/react/sharing/storybook-composition

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

kubijo:
Closes #13506

Enable the `buildStoriesJson` storybook flag so that CDS storybook can be [composed](https://storybook.js.org/docs/react/sharing/storybook-composition)

#### Changelog

**New**

- storybook: Enable the `buildStoriesJson` flag


kubijo:
I have validated that the storybook build now includes a [`stories.json` file](https://deploy-preview-13627--carbon-components-react.netlify.app/stories.json); however, I'm getting a CORS error when trying to integrate it into our storybook instance:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://deploy-preview-13627--carbon-components-react.netlify.app/index.json. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.
```

Could somebody look into this?
I'm guessing that `netlify` will have some kind of settings for this kind of thing (not sure about your public production build, though)

tay1orjones (approved):
@kubijo The production storybook is served via GitHub Pages, which has a `Access-Control-Allow-Origin` header of `*`. So it should work there.

I added a commit to this PR to update the Netlify config to specify the same for that header. I'm not sure if it'll work on the deploy preview yet. It might need to be merged into `main` for it to take effect.

In either case you should be good to go 👍 Thanks for the PR!

alisonjoseph (approved):

PR #13628, opened 2023-04-19
 Add TypeScript types to ComposedModal, ModalHeader, ModalBody, ModalFooter #13534 

### LINKED ISSUE #13534: Add TypeScript types to ComposedModal, ModalHeader, ModalBody, ModalFooter

kubijo:
Closes #13534 

#### Changelog

**New**

- add types to `ComposedModal`, `ModalHeader`, `ModalBody`, `ModalFooter`


packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -76,60 +90,136 @@ SecondaryButtonSet.propTypes = {
...
-  ref
-) {
-  const prefix = usePrefix();
-
-  const footerClass = cx({
-    [`${prefix}--modal-footer`]: true,
-    [customClassName]: customClassName,
-    [`${prefix}--modal-footer--three-button`]:
-      Array.isArray(secondaryButtons) && secondaryButtons.length === 2,
-  });
```
kubijo, line 105: This form is needlessly complicated & syntax-dense, plus it produces type errors when className is nullable (which it practically always is)

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -81,194 +109,271 @@ ModalBody.propTypes = {
...
-      wrapFocus({
-        bodyNode,
-        startSentinelNode,
-        endSentinelNode,
-        currentActiveNode,
-        oldActiveNode,
-        selectorsFloatingMenus,
-      });
+    function handleKeyDown(evt: KeyboardEvent) {
+      if (evt.key === 'Escape') {
```
kubijo: `event.which` is deprecated

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -81,194 +109,271 @@ ModalBody.propTypes = {
...
+      relatedTarget: currentActiveNode,
+    }) {
+      if (open && currentActiveNode && oldActiveNode && innerModal.current) {
+        const { current: bodyNode } = innerModal;
+        const { current: startSentinelNode } = startSentinel;
+        const { current: endSentinelNode } = endSentinel;
+        wrapFocus({
+          bodyNode,
+          startTrapNode: startSentinelNode,
+          endTrapNode: endSentinelNode,
```
kubijo, line 249: These were apparently invalid keys here & re-mapping seemed like the much easier option

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -81,194 +109,271 @@ ModalBody.propTypes = {
...
+      isFullWidth && `${prefix}--modal-container--full-width`,
+      containerClassName
+    );
+
+    // Generate aria-label based on Modal Header label if one is not provided (L253)
+    let generatedAriaLabel;
+    const childrenWithProps = React.Children.toArray(children).map((child) => {
+      switch (true) {
+        case isElement(child) &&
+          child.type === React.createElement(ModalHeader).type: {
```
kubijo, line 284: This needed addition of `@types/react-is` dependency to give us the type-safety + we can now be sure about the actual type of the child (since `ReactNode` is a broad type

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -81,194 +109,271 @@ ModalBody.propTypes = {
...
+          const el = child as ReactElement<
+            ModalHeaderProps,
+            typeof ModalHeader
+          >;
+          generatedAriaLabel = el.props.label;
```
kubijo, line 289: Here we get the exact props shape as opposed to unknown property access

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -81,194 +109,271 @@ ModalBody.propTypes = {
...
-        if (button.current) {
-          button.current.focus();
+
+        case isElement(child) &&
+          child.type === React.createElement(ModalFooter).type: {
+          const el = child as ReactElement<
+            ModalFooterProps,
+            typeof ModalFooter
+          >;
+          return React.cloneElement(el, { closeModal, inputref: button });
```
kubijo, line 299: Ditto about the exact props shape, including the (now fixed) omission of event parameter in the `ModalFooter` props

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -1,24 +1,37 @@
...
 
+interface SecondaryButtonProps {
+  buttonText: ReactNode;
+  onClick(): void;
+}
+export interface SecondaryButtonSetProps {
+  closeModal(evt: MouseEvent): void;
+  onRequestClose(evt: MouseEvent): void;
+  secondaryButtonText?: string;
+  secondaryButtons: [SecondaryButtonProps, SecondaryButtonProps];
```
francinelucca:
```suggestion
  secondaryButtons?: [SecondaryButtonProps, SecondaryButtonProps];
```

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -1,24 +1,37 @@
...
+import React, { type ReactNode, type MouseEvent, type Ref } from 'react';
 import PropTypes from 'prop-types';
 import Button from '../Button';
 import ButtonSet from '../ButtonSet';
 import cx from 'classnames';
 import { usePrefix } from '../../internal/usePrefix';
 
+interface SecondaryButtonProps {
+  buttonText: ReactNode;
+  onClick(): void;
```
francinelucca: should this take in a button click event?

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -76,60 +90,136 @@ SecondaryButtonSet.propTypes = {
...
+  ) {
+    const prefix = usePrefix();
+
+    const footerClass = cx(
+      `${prefix}--modal-footer`,
+      customClassName,
+      Array.isArray(secondaryButtons) && secondaryButtons.length === 2
+        ? `${prefix}--modal-footer--three-button`
+        : null
+    );
```
francinelucca, line 191:
```suggestion
    const footerClass = cx(
      `${prefix}--modal-footer`,
      Array.isArray(secondaryButtons) && secondaryButtons.length === 2
        ? `${prefix}--modal-footer--three-button`
        : null,
        {...(customClassName && { [customClassName]: true })}
    );
```

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -145,6 +235,7 @@ ModalFooter.propTypes = {
   /**
    * Specify an optional function that is called whenever the modal is closed
    */
+  // @ts-expect-error: Invalid derived type
```
francinelucca: making the function optional in the prop types should help get rid of this override

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -76,60 +90,136 @@ SecondaryButtonSet.propTypes = {
...
+
+  /**
+   * Specify a custom className to be applied to the Modal Footer container
+   */
+  className?: string;
+
+  /**
+   * Specify an optional function that is called whenever the modal is closed
+   */
+  closeModal(evt: MouseEvent): void;
```
francinelucca:
```suggestion
  /**
   * Specify an optional function that is called whenever the modal is closed
   */
  closeModal?(evt: MouseEvent): void;
```

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -167,12 +259,14 @@ ModalFooter.propTypes = {
    * Specify an optional function for when the modal is requesting to be
    * closed
    */
+  // @ts-expect-error: Invalid derived type
```
francinelucca: I think we want to opt for fixing the type errors in these cases instead of disabling the rule, some of these can be removed by making the interface prop optional

packages/react/src/components/ComposedModal/ModalHeader.tsx
```diff
@@ -0,0 +1,203 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { type ReactNode, type MouseEvent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { Close } from '@carbon/icons-react';
+import { usePrefix } from '../../internal/usePrefix';
+
+export interface ModalHeaderProps {
```
francinelucca: does this need to extend Div element props? it can receive them through `...rest`

packages/react/src/components/ComposedModal/ModalHeader.tsx
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * Provide an optional function to be called when the close button is
+   * clicked
+   */
+  buttonOnClick(): void;
```
francinelucca:
```suggestion
  buttonOnClick?(): void;
```

packages/react/src/components/ComposedModal/ModalHeader.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * Specify an optional className to be applied to the modal close icon node
+   */
+  closeIconClassName?: string;
+
+  /**
+   * Provide an optional function to be called when the modal is closed
+   */
+  closeModal(event: MouseEvent): void;
```
francinelucca:
```suggestion
  closeModal?(event: MouseEvent): void;
```

packages/react/src/components/ComposedModal/ModalHeader.tsx
```diff
@@ -0,0 +1,203 @@
...
+    ref
+  ) {
+    const prefix = usePrefix();
+
+    function handleCloseButtonClick(evt: MouseEvent) {
+      closeModal(evt);
+      buttonOnClick();
+    }
+
+    const headerClass = cx(`${prefix}--modal-header`, customClassName);
```
francinelucca, line 103:
```suggestion
    const headerClass = cx({[`${prefix}--modal-header`]: true, ...(customClassName && { [customClassName]: true })});
```

francinelucca: Same correction for all the custom classes below

packages/react/src/components/ComposedModal/ModalHeader.tsx
```diff
@@ -0,0 +1,203 @@
...
+    );
+  }
+);
+
+ModalHeader.propTypes = {
+  /**
+   * Provide an optional function to be called when the close button is
+   * clicked
+   */
+  // @ts-expect-error: Glitching types with the Props derivation
```
francinelucca: let's try to prevent using these as much as possible, what's the error this is generating here?

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+    },
+    ref
+  ) {
+    const prefix = usePrefix();
+    const contentClass = cx(
+      `${prefix}--modal-content`,
+      hasForm && `${prefix}--modal-content--with-form`,
+      hasScrollingContent && `${prefix}--modal-scroll-content`,
+      customClassName
+    );
```
francinelucca, line 58: we've been using this className standard to prevent undefined classes 
image

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+      </>
+    );
+  }
+);
+
+ModalBody.propTypes = {
+  /**
+   * Required props for the accessibility label of the header
+   */
+  // @ts-expect-error: Invalid derived type
```
francinelucca: is there a way to solve for the type error instead of disabling the rule?

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+   */
+  hasForm: PropTypes.bool,
+
+  /**
+   * Specify whether the modal contains scrolling content
+   */
+  hasScrollingContent: PropTypes.bool,
+};
+
+export interface ComposedModalProps {
```
francinelucca: should this extend HTML props?

packages/react/src/components/ComposedModal/ComposedModal.tsx
francinelucca: let's actually use our `match` utility instead 🙏🏻
image

image

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+        setisOpen(false);
+      }
+    }
+
+    const modalClass = cx(
+      `${prefix}--modal`,
+      isOpen && 'is-visible',
+      customClassName,
+      danger && `${prefix}--modal--danger`
+    );
```
francinelucca: we've been using this className standard to prevent undefined classes. Same comment for `containerClass` below
image

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+  /**
+   * Specify a CSS selector that matches the DOM element that should be
+   * focused when the Modal opens
+   */
+  selectorPrimaryFocus: PropTypes.string,
+
+  /**
+   * Specify the CSS selectors that match the floating menus
+   */
+  // @ts-expect-error: Invalid derived type
```
francinelucca: let's try to prevent using these as much as possible, what's the error this is generating here?

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+   * Specify an optional handler for closing modal.
+   * Returning `false` here prevents closing modal.
+   */
+  onClose: PropTypes.func,
+
+  /**
+   * Specify an optional handler for the `onKeyDown` event. Called for all
+   * `onKeyDown` events that do not close the modal
+   */
+  // @ts-expect-error: Invalid derived type
```
francinelucca: let's try to prevent using these as much as possible, what's the error this is generating here?

packages/react/src/components/ComposedModal/ComposedModal.tsx
francinelucca: let's try to prevent using these as much as possible, what's the error this is generating here?

packages/react/src/components/ComposedModal/ComposedModal.tsx
francinelucca: let's try to prevent using these as much as possible, what's the error this is generating here?

francinelucca (changes requested):
Thanks for taking this on! 
Just have some suggestions, questions and change requests.

I think mostly just implementing our standard approach to optional classNames in tsx and trying to clean up the ts-lint overrides 🙏🏻

packages/react/src/components/ComposedModal/ModalFooter.tsx
kubijo: Do you really feel that this is worth it?

It's a difference from a variable being passed to this in compiled code… when the function, that It's passed into, solves for the nullability anyway...

```javascript
_objectSpread({}, customClassName && _defineProperty({}, customClassName, true)));
```

Your house, your rules, but this really strikes me as silly from both readability & runtime cost perspective… I wouldn't go around rewriting all of those just for the heck of it, but fixing with ongoing rewrite seems rather sensible :)

The class names order is a different thing & I do agree that the one passed in should be last.

packages/react/src/components/ComposedModal/ComposedModal.tsx
kubijo:
```
TS2322: Type 'Requireable<(...args: any[]) => any>' is not assignable to type 'Validator<(evt: KeyboardEvent<Element>) => void>'.
   Types of property '[nominalTypeHack]' are incompatible.
     Type '{ type: ((...args: any[]) => any) | null | undefined; } | undefined' is not assignable to type '{ type: (evt: KeyboardEvent<Element>) => void; } | undefined'.
       Type '{ type: ((...args: any[]) => any) | null | undefined; }' is not assignable to type '{ type: (evt: KeyboardEvent<Element>) => void; }'.
         Types of property 'type' are incompatible.
           Type '((...args: any[]) => any) | null | undefined' is not assignable to type '(evt: KeyboardEvent<Element>) => void'.
             Type 'undefined' is not assignable to type '(evt: KeyboardEvent<Element>) => void'.
```

packages/react/src/components/ComposedModal/ModalFooter.tsx
francinelucca: Confirmed `undefined` & `null` values are being ignore, we can keep as-is then 👍

francinelucca (approved):

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+  selectorsFloatingMenus: PropTypes.arrayOf(PropTypes.string),
+
+  /**
+   * Specify the size variant.
+   */
+  size: PropTypes.oneOf(['xs', 'sm', 'md', 'lg']),
+};
+ComposedModal.defaultProps = {
+  selectorPrimaryFocus: '[data-modal-primary-focus]',
+};
```
francinelucca, line 444: what happened to the onKeyDown default prop here? 
image

packages/react/src/components/ComposedModal/ModalHeader.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Specify an optional className to be applied to the modal heading
+   */
+  titleClassName: PropTypes.string,
+};
+
+ModalHeader.defaultProps = {
+  iconDescription: 'Close',
+};
```
francinelucca, line 204: what happened to the `buttonOnClick` & `closeModal` default props here?
image

packages/react/src/components/ComposedModal/ModalFooter.tsx
```diff
@@ -156,6 +246,7 @@ ModalFooter.propTypes = {
   /**
    * The `ref` callback for the primary button.
    */
+  // @ts-expect-error: Invalid derived type
   inputref: PropTypes.oneOfType([
```
francinelucca, line 250: I couldn't figure this one out, is there anything we can do here to fix the type error @tay1orjones @mbarrer ?
```
  /**
   * The `ref` callback for the primary button.
   */
  // @ts-expect-error: Invalid derived type
  inputref: PropTypes.oneOfType([
    PropTypes.func,
    PropTypes.shape({
      current: PropTypes.any,
    }),
  ]),

```
image

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,457 @@
...
+
+ModalBody.propTypes = {
+  /**
+   * Required props for the accessibility label of the header
+   */
+  // @ts-expect-error: Built-in prop-types > TS logic doesn't jive well with custom validators
+  ['aria-label']: requiredIfGivenPropIsTruthy(
+    'hasScrollingContent',
+    PropTypes.string
+  ),
```
francinelucca, line 89: do we have a standard established for moving custom prop types to typescript? are we good to ignore the error? 🤔 @tay1orjones @mbarrer

francinelucca (changes requested):
Mostly there, some questions on default props and looks like some of the tests are failing @kubijo https://github.com/carbon-design-system/carbon/actions/runs/4764692624/jobs/8469588982?pr=13628

packages/react/src/components/ComposedModal/ComposedModal.tsx
kubijo: It is much more readable and makes more sense to skip the defaulting to no-op and either condition the consumer site or add the fallback in desctructuring... there are occurrences of both of these patterns elsewhere in the codebase, so I've opted for it here as well.

Also, one aspect that helps here specifically is that plays much nicer with typescript because if you supply defaultProps for a property, the pops entry must be non-nullable

kubijo: Regarding the failing test ... as far as I'm aware, it's about typescript errors around the invalid html attribute

kubijo:
> Mostly there, some questions on default props and looks like some of the tests are failing @kubijo https://github.com/carbon-design-system/carbon/actions/runs/4764692624/jobs/8469588982?pr=13628

I have just now fixed some snapshots, but this thing stumps me… any ideas on your side?

```
● ComposedModal › it renders as expected › should change size based on size prop

    Expected test not to call console.error()

    Error: Uncaught [Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.]

      19 |   describe('it renders as expected', () => {
      20 |     it('supports a custom class on the outermost div', () => {
    > 21 |       render(<ComposedModal className="custom-class" />);
         |       ^
      22 |
      23 |       expect(screen.getByRole('presentation', { hidden: true })).toHaveClass(
      24 |         'custom-class'
```

packages/react/src/components/ComposedModal/ComposedModal.tsx
tay1orjones: No we don't have a pattern established for this yet. Totally fine to ignore for now and we can improve it in the future.

packages/react/src/components/ComposedModal/ModalFooter.tsx
tay1orjones: I'm not sure, I think it's fine to merge as is with the expected error

tay1orjones (approved):
This looks fine to me other than the status checks that are failing. 

@kubijo I'm not sure what's causing that snapshot error. You might remove piece by piece of your changes to see what might be causing it.

kubijo:
Yep...

kubijo:
The good: I've got both storybook and in-IDE jest debug running and resolved the issue.
The ugly: I have no clue as to why it wasn't an issue before.

packages/react/src/components/ComposedModal/ComposedModal-test.js
```diff
@@ -179,16 +175,17 @@ describe('ComposedModal', () => {
...
-        'is-visible'
-      );
+      await new Promise((resolve) => setTimeout(resolve, 1e3));
+
+      const elementModal = screen.getByRole('presentation', { hidden: true });
+      expect(elementModal).toHaveClass('is-visible');
 
-      expect(screen.getByTestId('test-id-1')).toHaveFocus();
+      const elementInput = screen.getByTestId('test-id-1');
+      expect(elementInput).toHaveFocus();
```
francinelucca: is the timeout here necessary if you `await` the `expect` statements?
I.E.: `await expect(elementModal).toHaveClass('is-visible');` instead of `expect(elementModal).toHaveClass('is-visible');`

francinelucca (changes requested):
This LGTM except I see a small regression where the ComposedModal doesn't seem to be closing on outside click, it does seem to close on outsideClick in production: 

Prod:

https://github.com/carbon-design-system/carbon/assets/40550942/24b0319a-f96d-4658-94df-6a953fa31277


Deployment:

https://github.com/carbon-design-system/carbon/assets/40550942/61f60850-814b-4cb4-a489-ed01bdcddd7e

packages/react/src/components/ComposedModal/ComposedModal-test.js
kubijo: Oh my, sorry… the timeout is a forgotten experiment for fixing. The issue was elsewhere, but this got left behind.
I'll fix it right away

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -0,0 +1,446 @@
...
+    function handleKeyDown(evt: KeyboardEvent) {
+      if (match(evt, keys.Escape)) {
+        closeModal(evt);
+      }
+
+      onKeyDown?.(evt);
+    }
+    function handleMousedown(evt: MouseEvent) {
+      const isInside = innerModal.current?.contains(evt.target as Node);
+      if (!isInside && !preventCloseOnClickOutside) {
```
kubijo, line 234: @francinelucca here is the fix for modal not closing on outside clicks ... as a rant sidenote: I find frankly fascinating that the previous version ever worked… that was some quite creative nonsense :thinking:

francinelucca (approved):
👏🏽👏🏽👏🏽

Issue #13537, opened 2023-04-17
Add TypeScript types to SelectableTile


tay1orjones:

kubijo:
I'm taking this

Tresau-IBM:
@tay1orjones looks like this was not closed automatically from the PR #13631

PR #13631, opened 2023-04-20
 Add TypeScript types to the xTile family of components #13536 

### LINKED ISSUE #13536: Add TypeScript types to ExpandableTile

### LINKED ISSUE #13542: Add TypeScript types to Tile, TileAboveTheFoldContent, TileBelowTheFoldContent

kubijo:
Closes #13542, #13536, #13537

#### Changelog

**New**

-  Add TypeScript types to `Tile`, `SelectableTile`, `ExpandableTile`, `TileAboveTheFoldContent`, `TileBelowTheFoldContent`


packages/react/src/components/Tile/Tile.tsx
```diff
@@ -238,18 +365,18 @@ export const SelectableTile = React.forwardRef(function SelectableTile(
   return (
     <div
       className={classes}
-      onClick={!disabled ? handleOnClick : null}
+      onClick={!disabled ? handleOnClick : undefined}
       role="checkbox"
       aria-checked={isSelected}
       disabled={disabled}
```
kubijo: Raises following error:

```
TS2322: Type '{ children: Element[]; defaultChecked?: boolean | undefined; defaultValue?: string | number | readonly string[] | undefined; suppressContentEditableWarning?: boolean | undefined; suppressHydrationWarning?: boolean | undefined; ... 252 more ...; title: string; }' is not assignable to type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>'.   Property 'disabled' does not exist on type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>'.
```

What do you propose?

francinelucca: I think we should be good to remove this line since `disabled` is not a valid HTML attribute for `div` 
image
https://www.w3schools.com/tags/att_disabled.asp

kubijo: OK then, I have also removed the other invalid attributes by the same logic (namely `value` & `name`).

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -15,17 +24,26 @@ import useIsomorphicEffect from '../../internal/useIsomorphicEffect';
...
+  children?: ReactNode;
+  className?: string;
+}
+
+export const Tile = React.forwardRef<HTMLDivElement, TileProps>(function Tile(
+  {
+    children,
+    className,
+    // @ts-expect-error: Deprecated prop
+    light = false,
```
francinelucca: i think this prop should still be included in the `TileProps` above

francinelucca (changes requested):
Mostly LGTM, I'm doing some research on the `name` & `value` props to confirm we're good to remove but in the meantime, looks like some CI checks are failing and also I think we should add the `light` prop to the components' interface. See https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComboBox/ComboBox.tsx#L170
TIA!

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -166,12 +227,78 @@ ClickableTile.propTypes = {
...
+  SelectableTileProps
+>(function SelectableTile(
   {
     children,
     className,
     disabled,
     id,
+    // @ts-expect-error: Deprecated prop
     light,
     name,
```
francinelucca: confirmed we're good to remove the `name`, `value` & `disabled` attributes from the div. looks like that was the only thing we were using the `name` and `value` prop for though so I think we can either remove them from the component's destructuring or add an underline (`name: _name`) to the name to avoid these errors 
image
I'm not sure if removing it from the props since that might be considered a breaking change even if it's not used, maybe we @deprecate them instead? @tay1orjones

francinelucca (changes requested):

packages/react/src/components/Tile/Tile.tsx
tay1orjones: Yeah let's just deprecate them

tay1orjones:
@kubijo thanks for your work so far on this one. There's some remaining feedback to address here I think and status checks are failing.

packages/react/src/components/Tile/Tile.tsx
kubijo: `disabled` still seems to very much be in use for `className`s though, so I presume that only the other two should really be deprecated… which I'm doing

francinelucca (commented):
This is most of the way there, looks like there's one outstanding comment @kubijo  https://github.com/carbon-design-system/carbon/pull/13631#discussion_r1178179617. Other than that LGTM

packages/react/src/components/Tile/Tile.tsx
kubijo: Added & rebased :+1:

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -53,20 +68,67 @@ Tile.propTypes = {
...
+  ClickableTileProps
+>(function ClickableTile(
   {
     children,
     className,
     clicked = false,
     disabled,
     href,
+    // @ts-expect-error: Deprecated prop
     light,
```
francinelucca: we'll also need this one to be included in the interface, as well as in Selectable and ExpandableTile

francinelucca (changes requested):

packages/react/src/components/Tile/Tile.tsx
kubijo: asdf

francinelucca (approved):
LGTM!

tay1orjones (approved):

PR #13634, opened 2023-04-20
feat(link): adds external prop

### LINKED ISSUE #13633: [Feature Request]: Add `external` prop to `<Link />` component
### The problem

This features adds a boolean prop to Links called `external`. This would streamline the `target="_blank"`and `rel="noopener noreferrer"`

### The solution

Add the following to the carbon `Link` component.

```
<a 
     target={external ? '_blank' : undefined)}
     rel={external ? 'noopener noreferrer' : undefined} 
  >
    I'm a link
</a>
```

  

### Examples

```
<Link href="www.carbondesignsystem.com" external>Go to carbon</Link>
```

### Application/PAL

Instana

### Business priority

### Available extra resources

_No response_

### Code of Conduct

AlexanderMelox:
Closes #13633 

This PR adds an external prop to allow the link component to open in a new tab

#### Changelog

**Changed**

- `packages/react/src/components/Link/Link.js`

#### Testing / Reviewing

1. Change `external` control in storybook to true
2. When you click on the link it should open in a new tab

### DIFF
--- a/packages/react/src/components/Link/Link.js
+++ b/packages/react/src/components/Link/Link.js
@@ -20,7 +20,7 @@ const Link = React.forwardRef(function Link(
     visited = false,
     renderIcon: Icon,
     size,
-    external = false,
+    openInNewWindow = false,
     ...rest
   },
   ref
@@ -36,7 +36,7 @@ const Link = React.forwardRef(function Link(
   let rel = null;
   let target = null;
 
-  if (external) {
+  if (openInNewWindow) {
     rel = 'noopener';
     target = '_blank';
   } else {
@@ -89,11 +89,6 @@ Link.propTypes = {
   disabled: PropTypes.bool,
 
   /**
-   * Specify if the link should open in a new tab
-   */
-  external: PropTypes.bool,
-
-  /**
    * Provide the `href` attribute for the `<a>` node
    */
   href: PropTypes.string,
@@ -102,6 +97,11 @@ Link.propTypes = {
    * Specify whether you want the inline version of this control
    */
   inline: PropTypes.bool,
+
+  /**
+   * Specify if the link should open in a new tab
+   */
+  openInNewWindow: PropTypes.bool,
 
   /**
    * Optional prop to render an icon next to the link.
--- a/packages/react/src/components/Link/Link.js
+++ b/packages/react/src/components/Link/Link.js
@@ -20,6 +20,7 @@ const Link = React.forwardRef(function Link(
     visited = false,
     renderIcon: Icon,
     size,
+    external = false,
     ...rest
   },
   ref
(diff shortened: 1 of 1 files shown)


AlexanderMelox:
Note: the user can technically pass `external: true` but still set the target and rel separately and would overwrite this prop.

AlexanderMelox:
> @AlexanderMelox Thanks for making a contribution! I'm concerned this is making assumptions that all external links need to open in a new window. You should already be able to pass in `target="_blank"` to get the same functionality, but if we want to add a prop I would suggest naming it differently, possibly `openInNewWindow`

Hey! Normally I thought the pattern was if you're leading the user away from the page it should open up in a new tab?
Screenshot 2023-04-20 at 10 55 56 AM

AlexanderMelox:
I do see what you mean if the link is technically internal but want it to open in a new window 🤔

tay1orjones:
@AlexanderMelox [This article](https://css-tricks.com/use-target_blank/) outlines a stance I think is more suitable to a Link component provided by a design system, and what's been the operating assumption by our team so far when approaching this space.

tay1orjones (commented):
If I'm understanding correctly, this new prop is just a shortcut?

`<Link target="_blank" />` is equivalent to `<Link openInNewWindow/>`.

What if instead of providing a new prop, we just codified `target` as a prop in propTypes?

This will additionally need a new unit test to cover the prop.

AlexanderMelox:
@tay1orjones I was also under the impression that it's a good idea when using `target="blank` to also include `rel="noopener noreferrer"`. I always get an eslint error if I don't add that line of code

tay1orjones:
@AlexanderMelox I think that depends on if you trust or control the destination.

From [`react/jsx-no-target-blank`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md#when-to-override-it):
> For links to a trusted host (e.g. internal links to your own site, or links to a another host you control, where you can be certain this security vulnerability does not exist), you may want to keep the HTTP Referer header for analytics purposes.

---

Also, when I was on that page I found something related. Modern browsers automatically imply the `rel="noopener"`. Since we've dropped IE11 from our supported browser list in v11, we could probably remove it?

https://github.com/carbon-design-system/carbon/blob/0a606d345958f1f8677f37682f16019a2930d737/config/browserslist-config-carbon/index.js#L10



> Modern browsers (Chrome ≥ 88, Edge ≥ 88, Firefox ≥ 79 and Safari ≥ 12.2) automatically imply rel="noopener". Therefore this rule is no longer needed, if legacy browsers are not supported. See https://web.dev/external-anchors-use-rel-noopener/ and https://caniuse.com/mdn-html_elements_a_implicit_noopener for more details.


---

All this to say, I think it might be worthwhile for `rel` to be a configurable prop

tay1orjones:
Check out how Seek's Braid design system does this: https://seek-oss.github.io/braid-design-system/components/ButtonLink/props/

They've got `rel` and `referrerPolicy` as configurable options, which I think could make a lot of sense here. What do you think @alisonjoseph?

alisonjoseph:
Hi @AlexanderMelox chatted with @tay1orjones and some of the other Carbon devs offline about this issue. We've decided to close this issue/PR as you are already able to achieve this functionality without adding additional complexity to the component; it forwards all the props to the <a> tag already.

Issue #13541, opened 2023-04-17
Add TypeScript types to TabsSkeleton


tay1orjones:

tay1orjones:
Closed via https://github.com/carbon-design-system/carbon/pull/13635

Issue #10777, opened 2022-02-17
React 18 Enhancements: Ensure default props are initialized with serially-stable values


dcwarwick:
### Package

carbon-components-react

### Package version

10.53.0

### Description

I notice that several places where `defaultProps` were being used before, this has been removed and instead the default prop values are being supplied as ES6 destructuring initialisers. E.g. Button, TextInput. This is fine, of course, as this is now a preferred way of handling default prop values.

However, there is a tricky anti-pattern to watch out for. The destructuring initialisers will run EVERY TIME the render function is called, so whenever a prop is NOT supplied the default value will be inserted on EVERY CALL to the render function. This is ok if the default value is primitive, or any other serially stable value: when React checks what has changed it will consider these props to have the same values and not trigger any hooks that depend on them, including rendering. However, if the default value is something that is not serially stable, the prop will appear to React to have a new value every time the function is called, and so affected hooks and rendering will happen EVERY TIME. This is usually not harmful except that it is not optimal. Optimization is one of the motivations for switching the way that default prop values are supplied.

Some examples of values that are not serially stable:
 - []
 - () => {}
 
For example, `TextInput` now has the following two default prop values:
```
    onChange = () => {},
    onClick = () => {},
```
The `onChange` and `onClick` props, if not supplied, will be initialised to a new arrow function on every call, which React will see as a new prop value on every call, so it’ll re-render the text input on every call.

Also, `Button` has the following default prop value:
```
    size = FeatureFlags.enabled('enable-v11-release') ? 'lg' : 'default',
```
Now, this *is* serially stable, but it will check the feature flag on every invocation, which is perhaps not ideal either.

When the initialization expression is not serially stable, the initialization needs to be done outside the function parameter list. So, for example, the `TextInput` could instead do:
```
const noopFn = () => {};

// ...

  onChange = noopFn,
  onClick = noopFn,
```
and this will now be fine.

Similarly, `Button` could do:
```
const defaultSize = FeatureFlags.enabled('enable-v11-release') ? 'lg' : 'default',

// ...

    size = defaultSize,
```
and that will ensure the default value is only calculated once.

[...]

PR #13635, opened 2023-04-20
Types/tabs

### LINKED ISSUE #13540: Add TypeScript types to Tab, TabContent, TabList, TabPanel, TabPanels, Tabs

kubijo:
Closes #13540, #13541

#### Changelog

**New**

- add typescript types to the Tabs components family


packages/react/src/components/Tab/Tab.tsx
```diff
@@ -115,41 +182,39 @@ export default class Tab extends React.Component {
       handleTabClick,
       handleTabKeyDown,
       disabled,
+      // @ts-expect-error: Deprecated prop
```
francinelucca: we want to include these in the component props interface even if they're deprecated

packages/react/src/components/Tab/Tab.tsx
```diff
@@ -102,11 +170,10 @@ export default class Tab extends React.Component {
     label: FeatureFlags.enabled('enable-v11-release')
       ? undefined
       : 'provide a label',
-    selected: false,
-    onClick: () => {},
-    onKeyDown: () => {},
```
francinelucca: we're ok with removing these from here but they need to be re-added as default function params (in the props destructuing in the render method)

packages/react/src/components/Tab/Tab.tsx
```diff
@@ -76,11 +149,6 @@ export default class Tab extends React.Component {
     renderAnchor: deprecate(PropTypes.func),
     renderButton: PropTypes.func,
 
-    /*
-     * An optional parameter to allow overriding the content rendering.
-     **/
-    renderContent: PropTypes.func,
-
```
francinelucca: @tay1orjones  wondering if we're good to remove this unused prop 🤔. If not, also wondering if this one and `role` need to be added to the `TabProps` interface

packages/react/src/components/TabContent/TabContent.tsx
```diff
@@ -66,9 +87,3 @@ TabContent.propTypes = {
    */
   selected: PropTypes.bool,
 };
-
-TabContent.defaultProps = {
-  selected: false,
-};
```
francinelucca, line 72: This is good to remove but we want to re-set that default at the function param level

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -116,32 +252,39 @@ function TabList({
   contained = false,
   iconSize,
   leftOverflowButtonProps,
+  // @ts-expect-error: Deprecated prop
```
francinelucca: same comment here about adding to interface

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -601,11 +807,50 @@ Tab.propTypes = {
...
+   * Specify the duration in milliseconds to delay before hiding the tooltip
+   */
+  leaveDelayMs?: number;
+}
+
+const IconTab = React.forwardRef<HTMLDivElement, IconTabProps>(function IconTab(
   {
     children,
     className: customClassName,
-    defaultOpen = false,
```
francinelucca: does this default need to be removed?

francinelucca (changes requested):
This is looking good! Behavior works as expected, just a couple change requests

packages/react/src/components/Tab/Tab.tsx
kubijo: Not really ... what matters is type safety and if the call sites are already conditional, you don't need to create empty functions left and right

packages/react/src/components/TabContent/TabContent.tsx
kubijo: Also not true as the default "falsy" `undefined` serves the same as `false`

packages/react/src/components/Tab/Tab.tsx
tay1orjones: The broad convention within the codebase right now is that call sites to event handlers like this `on*` are not conditional, they're called unconditionally with the expectation that they're available or a default noop. So it's important we have defaults set for those cases.

Additionally, when setting defaults for callback functions like this, they need to be a static reference so a new function isn't created every time it's called #10777

packages/react/src/components/TabContent/TabContent.tsx
tay1orjones: I get that falsy is inferred and won't break anything, but doesn't it impact the information shown about the prop in the intellisense within say vscode? Or does it also infer that boolean props without defaults inherently default to undefined/false?

kubijo: The rule is `optional bool` instead of `required bool`, which means that you can do either of:
- `<Foo />`
- `<Foo bar />`
- `<Foo bar={true} />`: this one should be caught by some linter rule as extraneous, since it's the same as not setting the value
- `<Foo bar={false} />`: This would only be needed if the property was optional and had a fallback to `true`, which is a DX antipattern

tay1orjones:
@kubijo thanks again for this! There's some remaining feedback to address here. I'd like to get this in but if it's stalled out we'll close after a few days.

kubijo:
> @kubijo thanks again for this! There's some remaining feedback to address here. I'd like to get this in but if it's stalled out we'll close after a few days.

I've got a rebase conflict seemingly due to `Tab.js` being deleted here https://github.com/carbon-design-system/carbon/pull/13735
What's up with that? What do we do with my rewrites then?

packages/react/src/components/Tab/Tab.tsx
kubijo: I can understand a reasoning for this in a plain JS codebase, where you don't have much of a chance to be safe-guarded against missed conditional call site and having a default value gets you a safer code… But I would like to suggest revisiting this reasoning while migrating to Typescript, where you do get the static insurance that you're not calling a nullable object.

packages/react/src/components/Tab/Tab.tsx
kubijo: How should I interpret this? Do I add something to the typescript interface?

kubijo:
As far as I can tell, someone moved a lot of code here and as a result, we're getting **MANY** new errors in the `Tabs.tsx` module (for example)… my patience with such a refactor is fleeting, and I really don't care for it…

Is there any chance of moving this to the author of those changes (them having the context of what was changed) and using this as a source of entropy?

kubijo:
> @kubijo sorry for the churn, `packages/react/src/components/Tab/index.js` was deleted because it's not being used. It was legacy code behind a v11 feature flag that has been permanently set `true`. You've already done the typings for the correct/new `Tab` in `Tabs.tsx`
> 
> There's been some new features added to Tabs recently, but I think it would only impact typings for two new props, `dismissable` and `onTabCloseRequest`. The delta shouldn't be too large.

Can someone on your side do the rebase, then? You (as in maintainers) should have a write access, I believe.

francinelucca (approved):
🚀

tay1orjones (approved):

tay1orjones:
Thanks for your help @francinelucca!

PR #13639, opened 2023-04-21
13420 rail sidenav esc

### LINKED ISSUE #13420: SideNav with isRail cannot be dismissed with Esc
From the parent issue: 

> ### Side rail cannot be dismissed with Esc
> 
> The side rail is essentially a reveal-on-focus (or hover) side panel, which auto collapses when interaction leaves its elements, either via keyboard activation or focus or pointer location. I think it's quite nicely done, but it has one key glitch. Its current implementation fails [On Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html), since it is not possible to dismiss the side panel with Esc. This should be an easy fix (just need to decide where focus would go, which I think is the main content, probably at top of region)

> ### Steps to reproduce
> 
> #### Side rail Esc
> 1. Open browser to https://react.carbondesignsystem.com/iframe.html?args=&id=components-ui-shell--side-nav-rail&viewMode=story
> 2. Press tab to enter Side rail
> 3. Press Esc to confirm Side rail is not dismissed (BAD)
> 4. Press Tab until main content is reach (side rail collapses, GOOD)

guidari:
Closes #13420 

The `SideNav Rail` wasn't closing when clicking **ESC**. Now when you click **ESC** the user can close the `SideNav` and then reopen if **TAB** is pressed.

#### Changelog

Open the `SideNav Rail` and use **TAB** to open. Then press **ESC** to close.

The `SideNav Rail` should retract.

To reopen you can press **TAB**, if the focus is inside the navigation


francinelucca (commented):
From #13420 :

> This should be an easy fix (just need to decide where focus would go, which I think is the main content, probably at top of region)

Should the focus go to the main content after `Escape`? pressing escape seems a bit redundant here if you still have to tab (which makes the sidenav reopen) all the way into the main content

guidari:
I considered using the same behavior as `SkipToContent.js`, which utilizes a href to navigate to the main content. But I'm not sure if it would the correct solution. Because it depends on what the user would write in the id of the `main` tag.

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -107,11 +108,19 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
     eventHandlers.onBlur = (event) => {
       if (!event.currentTarget.contains(event.relatedTarget)) {
         handleToggle(event, false);
       }
     };
+
+    eventHandlers.onKeyDown = (event) => {
+      if (match(event, keys.Escape)) {
+        handleToggle(event, false);
+        window.location.href = '#main-content';
```
francinelucca: I'm not sure what the right thing to do here is 🤔, like is it worth adding an `href` prop like [skipToContent](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/UIShell/SkipToContent.js) or should we just drop it and not redirect to the main content at all?

Ref:   https://github.com/carbon-design-system/carbon/issues/13420
> This should be an easy fix (just need to decide where focus would go, which I think is the main content, probably at top of region)

francinelucca (changes requested):

packages/react/src/components/UIShell/SideNav.js
tay1orjones: Yeah that was my thought - match the setup of SkipToContent and allow it to be configurable. I think we should provide something out of the box for it as a reasonable default

guidari: I added an `href` prop just like `SkipToContent.js`. 

The `SkipToContent.js` functionality appears to be functioning properly, but I've noticed a weird behaivour when using MacOS's voice over feature. Specifically, pressing the ESC key to skip to the main content doesn't trigger the voice over to inform the user that they've reached the main content, even though they are at the main content.

Both on `SkipToContent.js` and `SideNav`

guidari: I just noticed that some tests are failing. I'm gonna take a look at it

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -187,6 +197,12 @@ SideNav.propTypes = {
    */
   expanded: PropTypes.bool,
 
+  /**
+   * Provide the `href` to the id of the element on your package that is the
+   * main content.
+   */
+  href: PropTypes.string,
```
francinelucca: could we make this `required` if we add a default or would that be a breaking change @tay1orjones ? SkipToContent has it as required with default 
image

francinelucca (changes requested):
Almost there. I think we should add a default for the href prop like in `SkipToContent`
image

packages/react/src/components/UIShell/SideNav.js
guidari: I added the default prop and set the href as required.

Making the href attribute required for this component create any issues for developers who are already using it?
I don't know if make sense what I'm thinking

tay1orjones: The only problem I could see is that if someone pulls in this change, this href is going to be specified even if it wasn't previously. Their router solution may pick up on this or not, but I think it's possible it could cause unexpected side effects. 

We could err on the side of caution and make this opt-in for now. We could wait to mark it as required in v12. This could either be behind a feature flag or we could simply open an issue and wait to mark it required until v12 lands.

francinelucca: > this href is going to be specified even if it wasn't previously
> we could simply open an issue and wait to mark it required until v12 lands.

@tay1orjones  If we make it not required won't it be specified anyway since we added a default? or you're saying we should remove the default as well? 
image

tay1orjones: Yeah, I think it needs to be optional and the default removed

guidari: Perfect! I just removed the default also

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -107,11 +109,19 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
     eventHandlers.onBlur = (event) => {
       if (!event.currentTarget.contains(event.relatedTarget)) {
         handleToggle(event, false);
       }
     };
+
+    eventHandlers.onKeyDown = (event) => {
+      if (match(event, keys.Escape)) {
+        handleToggle(event, false);
+        window.location.href = href;
```
francinelucca: I think since `href` is not guaranteed to be defined since we removed the default, let's wrap this line in an `if(href){...}`

francinelucca (changes requested):

francinelucca (approved):
🚀🚀🚀

andreancardona (approved):
Other than the merge conflicts - looks good to me!

PR #13648, opened 2023-04-24
[a11y] Search: Search Icon Label Logic Update

sjbeatle:
With the current implementation, an accessibility violation occurs when the Search Icon's container doesn't have a role. The current implementation removes the `role` attribute when the `onExpand` property is undefined, but it should also remove the `aria-labelledby`.

This update will avoid the accessibility violation: https://www.w3.org/TR/wai-aria-1.1/#roles

#### Testing / Reviewing

The error can be see in the `@carbon/react` Storybook UI. Visit the [Search component](https://react.carbondesignsystem.com/?path=/story/components-search--default) and click on the 'Accessibility' tab

Thanks!


sjbeatle:

sjbeatle:
recheck

francinelucca (approved):
Confirmed no longer see this violation:
image

@sjbeatle looks like some tests are failing other than that LGTM!

guidari (approved):
Looks good to me also. Let's fix the tests and it will be ready to merge.

Thank you

sjbeatle:
> Looks good to me also. Let's fix the tests and it will be ready to merge.
> 
> Thank you

Great! I'll work on that right away. Thanks for the speedy response! 😁

sjbeatle:
@guidari @francinelucca 
Thank you for your patience, and sorry for the delay!

I think this one should be all set, now. Please let me know if there's anything else you'd like me to address.

Cheers!

packages/react/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
```diff
@@ -483,7 +483,6 @@ exports[`DataTable behaves as expected selection should render and match snapsho
           role="search"
         >
           <div
-            aria-labelledby="custom-id"
```
francinelucca, line 486: could we actually pass the `onExpand` prop so that this keeps the aria-labelledby instead? TIA!

francinelucca (changes requested):

andreancardona:
@sjbeatle Hey Steven just waiting on a small fix on your end and we can get this in :)

packages/react/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
sjbeatle: Hello @francinelucca 

Are you asking to refactor the `TableToolbarSearch` component, removing all of the `onExpand` handling logic that exists there and simply pass it through to the `Search` sub-component?

This change doesn't feel trivial to me. We will have to ensure that there are no deleterious effects.

sjbeatle:
> @sjbeatle Hey Steven just waiting on a small fix on your end and we can get this in :)

Hello @andreancardona!

Thank you, but I'm not confident that the fix is trivial. Please see my recent comment [here](https://github.com/carbon-design-system/carbon/pull/13648/files#r1191032860).

Thank you.

packages/react/src/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
francinelucca: Hi @sjbeatle sorry I'm reading this back and realizing my explanation wasn't very clear.
I meant in the tests where `aria-labelledby` is being removed from the snapshots, if we send in an `onExpand={() => {}}` empty prop function to the `<ToolbarSearch />` in the test file, then the `aria-labelledby` won't be removed from the snapshot and we can continue to ensure that it is there when the onExpand prop is passed in.

sjbeatle: Hey @francinelucca,

All good, no worries. I actually understood it as you explain, but I guess my point is that what you’re asking to do won’t work…

Since the `onExpand` prop is de-structured in the function’s call, it won’t be present in the `…rest` variable. So, passing `onExpand` in `TableToolbarSearch` is insufficient to ensure it is passed down to `Search`.

Additionally, `TableToolbarSearch` is calling the `onExpand` method in its own internal logic handlers, so simply passing it through to the `Search` sub-component might have duplicative and/or deleterious effects.

In order to accomplish what you’re recommending, refactoring the parent component will be required.

I’m happy to try and help with that work, but I’d just ask that your team look extra carefully at whatever changes I make, since you’re more familiar with the wider implications to a non-trivial change this would require.

I’ll give it a try unless I hear otherwise.

Thanks!

francinelucca: ah I didn't realize it was being filtered out by the `TableToolbarSearch`, we can leave it as is then, thanks for explaining that!

francinelucca (approved):

PR #13670, opened 2023-04-26
13416 menu focus order

### LINKED ISSUE #13416: Left UI Shell: The Focus Order fails
Originally outlined in the parent issue: 

> Problem 1: The Focus Order fails
> When the hamburger menu is activated, it discloses a left nav panel.
> The traditional and expected behaviour is that the user's next navigation point will be the first item in this revealed panel, either via tab (if it's just revealing a list of links, etc) or via down arrow (if it is revealing a menu).
> Instead, the focus progresses across the top navigation, and does not reach the first item in the left nav until all items in the top nav have been reached.
> It appears that the same tab order established when the top and left nav were persistent, has not been altered. This seems to constitute a failure of [Focus Order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html). Combined with other issues to be detailed, this leads to a higher severity result for keyboard users.

guidari:
Closes #13416 

Redirecting the navigation to the `SideNav` component when it's expanded and is not a fixed side nav.

#### Testing / Reviewing

First switch to mobile view, then open the hamburguer menu and press tab.

You should be redirected to the `SideNav`


packages/react/src/components/UIShell/SideNav.js
```diff
@@ -119,13 +119,22 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
     eventHandlers.onMouseLeave = () => handleToggle(false, false);
   }
 
+  useEffect(() => {
+    console.log(isFixedNav);
```
francinelucca:
```suggestion
```

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -119,13 +119,22 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
+    if (expanded && !isFixedNav) {
+      document.getElementById(`${prefix}--side-nav__navigation`).focus();
```
francinelucca: I think this should work? if not you can declare your own ref and `useMergedRefs` (see screenshot from [Dialog](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Dialog/index.js#L22)) to pass onto the div instead of `ref={ref}` in line 138. This should also eliminate the need for setting an `id` in the `div` element
```suggestion
{
      ref.current?.focus();
}
```
image

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -119,13 +119,22 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
+
   return (
     <>
       {isFixedNav ? null : (
         // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
         <div className={overlayClassName} onClick={onOverlayClick} />
       )}
       <nav
+        id={`${prefix}--side-nav__navigation`}
+        tabIndex="-1"
```
francinelucca: what is this `tabIndex` change solving for?

francinelucca (changes requested):
Behavior LGTM, just some cleanup suggestions and a question on the tabIndex change, also looks like some tests are failing @guidari 
Happy to pair up on this if necessary!

packages/react/src/components/UIShell/SideNav.js
guidari: I added the tabIndex="-1" attribute to allow for direct focus on the nav tag.

packages/react/src/components/UIShell/SideNav.js
guidari: I added the tabIndex="-1" attribute to allow for direct focus on the nav tag.

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -119,14 +123,21 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
     eventHandlers.onMouseLeave = () => handleToggle(false, false);
   }
 
+  useEffect(() => {
+    if (expanded && !isFixedNav) {
+      sideNavRef.current.focus();
+    }
```
francinelucca: let's check for undefined just in case
```suggestion
    if (expanded && !isFixedNav && sideNavRef.current) {
      sideNavRef.current.focus();
    }
```

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -40,6 +41,9 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
   const [expandedViaHoverState, setExpandedViaHoverState] =
     useState(defaultExpanded);
   const expanded = controlled ? expandedProp : expandedState;
+  const sideNavRef = useRef(null);
+  const refNav = useMergedRefs([sideNavRef, ref]);
```
francinelucca: I think the name `navRef` makes more sense here, just a suggestion tho'
```suggestion
  const navRef = useMergedRefs([sideNavRef, ref]);
```

francinelucca (changes requested):

francinelucca (approved):
This LGTM @guidari I would just confirm with Gower that it's ok for the focus to leave the open/close button on "Enter" and move onto the `nav`

guidari:
> This LGTM @guidari I would just confirm with Gower that it's ok for the focus to leave the open/close button on "Enter" and move onto the `nav`

I talked to michael. He wants us to bring that to the meeting on Monday.
I'll prepare to talk about it there

guidari:
After the Gower's meeting, I made the necessary changes to the code

The toggle menu currently has focus, and if the user presses the `tab` key, the focus will shift to the navigation `tab`

For that to happen I added an `id` to the`HeaderMenuButton` component. To make this happen, I added an `id` to the `HeaderMenuButton` component. However, I'm not sure if adding `ids` to Carbon components is considered best practice, as users might prefer to add their own 'ids'. Let me know if you think this approach is appropriate

Finally, I added the `aria-expanded` attribute to ensure accessibility with voice-over functionality

packages/react/src/components/UIShell/SideNav.js
```diff
@@ -119,14 +126,30 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
...
 
+  useEvent(window, 'keydown', (event) => {
+    const focusedElement = document.activeElement;
+
+    if (
+      match(event, keys.Tab) &&
+      expanded &&
+      !isFixedNav &&
+      sideNavRef.current &&
+      focusedElement.id === `${prefix}--toggle_menu` &&
```
francinelucca: could we achieve the same thing with a className? my concern is like you mentioned that setting the Id of the button internally prevents consumers from setting their own ids, which might be a breaking change

francinelucca (changes requested):

packages/react/src/components/UIShell/SideNav.js
guidari: Of course! I think it is a better solution also

guidari: And I think I'll have to create a new PR to solve those merge conflicts. They are related to the typescript change

guidari: I opened this [new PR](https://github.com/carbon-design-system/carbon/pull/13809) to prevent the typescript merge conflicts @francinelucca

PR #13675, opened 2023-04-26
chore(deps): bump and fix prettier

falsepopsky:
## What

1. Update `prettier` to latest.
2. Remove `prettier` from `devDependencies` in [prettier-config-carbon](https://github.com/falsepopsky/carbon/blob/07b031b190a4c3ee1a1b5d553d7374316b50c517/config/prettier-config-carbon/package.json).
3. Add `--cache` flag in `lint-staged`.
4. Don't use `--cache` in `format:diff` script.

## Why

1. Previously, in a [pull request](https://github.com/carbon-design-system/carbon/pull/11850) that I created, I added the `--cache` option, which had a [bug](https://github.com/prettier/prettier/issues/13015). However, this bug was [fixed](https://github.com/prettier/prettier/pull/13016) and released on [V8](https://prettier.io/blog/2022/11/23/2.8.0.html)
2. `prettier-config-carbon` is a shareable configuration that does not require `prettier` as `devDependency`. Instead, it should be added as a `peerDependency` to indicate that this configuration requires version `v2.x` of the `prettier` package.
3. I missed adding the `--cache` flag in my old [pr](https://github.com/carbon-design-system/carbon/pull/11850#issuecomment-1193227903).
4. The primary reason for this change is that I believe it is preferable to check the files rather than depending on the cache.

#### Changelog

**Changed**

- Update `prettier` to latest.

**Removed**

- Remove `prettier` from `devDependencies` in `prettier-config-carbon`

#### Testing / Reviewing

CI should pass.


tay1orjones (approved):
This looks great, thanks for updating it!

alisonjoseph (approved):
LGTM, other than a merge conflict

falsepopsky:
conflict solved, thanks

Issue #12728, opened 2022-11-25
[a11y]: OverflowMenuItem, None of the descendent elements with "menuitem" role is tabbable


mateBe95:
### Package

carbon-components

### Browser

Chrome

### Package version

Newest

### React version

16.12.0

### Automated testing tool and ruleset

IBM Equal Accessibility Checker

### Description

<img width="1704" alt="image" src="https://user-images.githubusercontent.com/23103157/203948178-2af16297-e5cb-4fc6-8d10-9e51848620e3.png">

None of the descendent elements with "menuitem" role is tabbable

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?id=components-overflowmenu--default&viewMode=story

### Steps to reproduce

https://react.carbondesignsystem.com/iframe.html?id=components-overflowmenu--default&viewMode=story

1.Open menu.
2. Scan.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
@mbgower @tombrunet Is this a valid failure? @francinelucca just addressed the keyboard navigation in https://github.com/carbon-design-system/carbon/pull/12977. OverflowMenuItem navigation happens via arrow keys, not by tabbing.

mateBe95:
Any updates here @tay1orjones the issue still seems to be occurs https://react.carbondesignsystem.com/iframe.html?id=components-overflowmenu--default&viewMode=story

francinelucca:
> Any updates here @tay1orjones the issue still seems to be occurs https://react.carbondesignsystem.com/iframe.html?id=components-overflowmenu--default&viewMode=story

In the overflowmenu item navigation is done using the up/down arrow keys. @mbgower bumping, is there something else we should be considering here regarding this violation?

mbgower:
I don't understand the issue. I'm pulling in @tombrunet to have a look.
I conjecture it might be because you're using buttons for the menuitems and putting the negative tabindex on them, and it's setting off a rule. But since they've been given a role of menuitem, I don't see the problem; they're reachable by arrow key.

I _am_ seeing an oddity where when I press tab after opening the menu, the focus is going around the whole list of options. That suggests there's something a little dodgy in the code?

[...]

PR #13678, opened 2023-04-27
chore(menu): move to stable

### LINKED ISSUE #13514: ContextMenu: Implementation
#### Acceptance Criteria

- [ ] Go through the newly defined steps from carbon-design-system/carbon#11688 to move ContextMenu from `unstable` to `stable`
- [x] Needs storybook updates, remove generator config, replace with usage examples, figure out left nav hierarchy/naming
- [x] Namespaces within the Menu exports need to be removed
- [x] `useEffect` within the Menu needs to be revisited - ignoring deps, ensure edge cases are covered
- [x] Revisit `useContextMenu` hook to instead use React context. No longer detect the children directly, but instead communicate through this context

janhassel:
Closes #13514
Ref https://github.com/carbon-design-system/carbon/pull/13467#issuecomment-1497156970

Affects:
- Menu
  - MenuItem
  - MenuItemDivider
  - MenuItemGroup
  - MenuItemRadioGroup
  - MenuItemSelectable
- useContextMenu
- ComboButton
- MenuButton
- OverflowMenuV2
  - OverflowMenuV2 stays marked as "experimental". The component itself is moved to `OverflowMenu/next` and can be enabled with a feature flag (`enable-v12-overflowmenu`)

---

- [x] All files have a copyright banner
- [x] All components exported in `src/index.js` and should not be `unstable_` prefixed
- [x] Component has a label in the github repository
  - Currently "component: context-menu" exists
  - @sstrubberg could you either rename this into "component: menu" or add additional ones if we wanna split it up?
    - "component: menu-button"
    - "component: combo-button"
    - "component: overflow-menu-v2"
- [ ] Component should be documented on the website
  - [ ] Component should have a usage, style, and code tab
  - [ ] Component may have a component demo
- [x] For each component exported:
  - [x] Component is written as a function declaration or uses `forwardRef`
  - [x] Component has `propTypes` defined
    - [x] Each prop type has a comment (used in storybook)
    - [x] Prop types are as specific as needed, prefer `PropTypes.shape` over `PropTypes.object` if possible
  - [x] Default props are listed as default args in the function definition (not in defaultProps)
    - [x] Note: default props should be stable, in other words props like `onClick = () => {}` can cause re-renders since the function identity is not stable
  - [x] Component has a story in `<ComponentName>.stories.js`
    - [x] Component has an mdx document that follows our outline
    - [x] mdx document covers at least common use-cases and provides a prop table
    - [x] Stories cover at least common use-cases
    - [x] Stories may include a `Playground` story for controls
    - [x] Stories should mirror intended usage of the component
  - [x] Component has unit/integration tests written in RTL for testing the component API
  - [x] Component is tested via VRT for at least the initial render state
  - [x] Component is tested via AVT for at least the initial render state


francinelucca:
>  Component has a label in the github repository
Currently "component: context-menu" exists
@sstrubberg could you either rename this into "component: menu" or add additional ones if we wanna split it up?
"component: menu-button"
"component: combo-button"
"component: overflow-menu-v2"

I vote for separate component tags 🙋🏽‍♀️

sstrubberg:
> > Component has a label in the github repository
> > Currently "component: context-menu" exists
> > @sstrubberg could you either rename this into "component: menu" or add additional ones if we wanna split it up?
> > "component: menu-button"
> > "component: combo-button"
> > "component: overflow-menu-v2"
> 
> I vote for separate component tags 🙋🏽‍♀️

I'm indifferent until I understand the ramifications. I'd love for the rest of @carbon-design-system/developers to weigh in.

tay1orjones:
It would be nice to be able to distinguish issues as relating to `menu`, and/or the components built atop it, `menu-button`, `combo-button`, etc.

tay1orjones:
I'm not sure if we should mark OverflowMenuV2 as stable under that name. The ideal for v12 would be that the existing OverflowMenu would be replaced with the internals of OverflowMenuV2 without a name change. 

Would it make more sense to instead:
- make a new feature flag like `v12-enable-new-overflowMenu`
- Toggle in `overflowmenu/index.js` which OverflowMenu is used depending on the value of the feature flag
- Update OverflowMenuv2 to just be a wrapper using FeatureFlags
```
<FeatureFlags flags={v12-enable-new-overflowMenu: true}>
  <OverflowMenu {…props}
</FeatureFlags>
```
- Place a deprecation warning on `OverflowMenuV2`, pointing folks to instead import OverflowMenu and wrap it in a `<FeatureFlag>`. This can be done at each individual usage point, globally at the root of their app, or anywhere inbetween.

This way existing usages of OverflowMenuV2 aren't impacted, but we're providing a path to a more seamless upgrade where folks won't have to change imports.

Thoughts?

A few unknowns:
- How will we handle the TypeScript typings of these two components under one export?

tay1orjones:
@janhassel I'm proposing a new way to document feature flags and organize their stories over in https://github.com/carbon-design-system/carbon/pull/13382#issuecomment-1535341246

I made a new file `/components/Tile/Tile.featureflag.stories.js` to house the stories. It's title is `title: 'Experimental/Feature Flags/Tile'`.

You could mirror the same for now and we could add the `enable-v12-overflowmenu` feature flag to the feature flag overview docs story once that other PR is merged.

janhassel:
@tay1orjones I updated the OverflowMenuV2 implementation as you described above. I was also able to already use the new `WithDeprecationNotice` story template 😀 https://deploy-preview-13678--carbon-components-react.netlify.app/?path=/story/experimental-unstable-overflowmenuv2--overflow-menu-v-2

Could you look over it and let me know if there's anything I missed?

tay1orjones (changes requested):
Percy caught this - the breadcrumb overflow icon is no longer horizontal like it should be

image

francinelucca (approved):
looks great!

tay1orjones (approved):
Looks great, thanks!

tay1orjones:
I just made the following labels:
- https://github.com/carbon-design-system/carbon/labels/component%3A%20menu-button
- https://github.com/carbon-design-system/carbon/labels/component%3A%20combo-button

We'll just reuse the existing https://github.com/carbon-design-system/carbon/labels/component%3A%20overflow-menu for both versions with and without the new feature flag.

tay1orjones:
Once https://github.com/carbon-design-system/carbon/pull/13840 is in, we'll also want to make the change here to remove `'aria_child_tabbable'` from the denylist in playwright.config.js and config/jest-config-carbon/matchers/toHaveNoACViolations.js

We had to add it in to get that PR to pass due to the Menu's ` None of the descendent elements with "menuitem" role is tabbable` a11y violation #12728

janhassel:
@tay1orjones Just pushed the updates 👍

laurenmrice (changes requested):
This looks great, @janhassel ! ✨Just a couple of things:

----

**Menu story**
- The caret icon in the first row of the menu needs to be vertically centered in the row with the text for all sizes. Currently, it looks to be a couple of pixels too high.
 
Screen Shot 2023-05-25 at 9 42 46 AM

**Menu button with dividers story (Safari bug)**
- When the menu is open it overlaps the button.

Screen Shot 2023-05-25 at 9 53 28 AM

**Context menu question**
- I don't see a story with the terminology "Context menu" specifically. Is `Menu `replacing this? For some reason, I was expecting to see it as its own component like Combo button, Menu button, and Overflow menu. I also can't find a way to right-click to get the menu to appear/open and was wondering if that interaction is still needed to display here or is it fine to only specify in our documentation?

laurenmrice (approved):

francinelucca:
@tay1orjones we still see the violation when removing the rule 
image

I see the [IBMa issue](https://github.com/IBMa/equal-access/issues/1402) is still open so I don't think we can expect this false positive violation to not show up yet...

tay1orjones:
@francinelucca @janhassel Ah I'm sorry, I thought the `aria-owns` and `aria-controls` updates would fix it. Looks like we can't remove that yet until the false positive is addressed, https://github.com/IBMa/equal-access/issues/1402.

tay1orjones (approved):
I just re-added `aria_child_tabbable` to the AC denylist. This looks good to merge and should pass now 👍

PR #13707, opened 2023-05-02
feature/Theme types

### LINKED ISSUE #13579: Add TypeScript types to Theme, GlobalTheme, useTheme

mrkjdy:
Closes #13579 

Converted Theme/index.js to tsx and added appropriate prop types. Should I add myself in the `.all-contributorsrc` and the `README.md`?

#### Changelog

**New**

- Added interfaces for GlobalTheme and Theme props 

**Changed**

- Renamed `Theme/index.js` to `Theme/index.tsx`

#### Testing / Reviewing

You can check that building still works and that there are no errors when viewing the Theme stories


packages/react/src/components/Theme/index.tsx
```diff
@@ -7,15 +7,22 @@
...
 import PropTypes from 'prop-types';
-import React, { useMemo } from 'react';
+import React, { useMemo, type PropsWithChildren } from 'react';
 import { usePrefix } from '../../internal/usePrefix';
 import { LayerContext } from '../Layer/LayerContext';
 
-export const ThemeContext = React.createContext({
+interface GlobalThemeProps {
+  theme?: 'white' | 'g10' | 'g90' | 'g100';
+}
```
guidari, line 17: I have a question/suggestion. Can we also add a type `children?: React.ReactNode;` and get rid of the `GlobalTheme.propTypes`? Since the typescript will provides static type checking.
And if that make sense we could also approach the `Theme` with the same logic.

I'm curious about your thoughts? @mrkjdy @tay1orjones @andreancardona

guidari (commented):
Hey, thank you for contrubuting @mrkjdy 

LGTM.

Feel free to add yourself to the list. You can follow this doc:
https://github.com/carbon-design-system/carbon/blob/main/.github/CONTRIBUTING.md#5-add-yourself-to-the-contributor-list

packages/react/src/components/Theme/index.tsx
mrkjdy: Would keeping the propTypes be useful for developers that are still using JavaScript? The react docs [recommend using TypeScript instead of using propTypes](https://react.dev/reference/react/Component#static-proptypes).

guidari: That make sense. I just checked a few other `.tsx` files and it kept the `propTypes` also

packages/react/src/components/Theme/index.tsx
```diff
@@ -62,13 +75,12 @@ export function Theme({
...
   }, [theme]);
+  const BaseComponentAsAny = BaseComponent as any;
 
   return (
     <ThemeContext.Provider value={value}>
       <LayerContext.Provider value={1}>
-        <BaseComponent {...rest} className={className}>
-          {children}
-        </BaseComponent>
+        <BaseComponentAsAny {...rest} className={className} />
```
guidari, line 83: Should't we also add the `children` prop as it was? Despite omitting that prop, I noticed that the storybook still functions properly.

But maybe would be good to keep it there. Let me know what you think

packages/react/src/components/Theme/index.tsx
tay1orjones: Yeah good question, we do need to keep the proptypes. It's mentioned in [the docs for adding component types](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md#strategy).

tay1orjones (approved):

guidari (approved):

packages/react/src/components/Theme/index.tsx
mrkjdy: The children prop will be set by the `{...rest}`

PR #13716, opened 2023-05-02
13417 menu loss of focus

### LINKED ISSUE #13415: Left UI Shell: The main content is obscured, both through a failure to reflow and the lightbox effect.
From the parent issue: 

> Problem 5: The main content is obscured, both through a failure to reflow and the lightbox effect.
> When the hamburger menu is activated, two things happen to exacerbate the issue.
> 
> There is a lightbox effect, which dims all the main content
> The main content is not adjusted so that it reflows into the space not obscured by the left nav. Instead it continues to display across the whole width of the screen, despite being partially non-visible.
> This means that the main content is both made less visible due to contrast and partially fully obscured behind the left navigation. This will be a failure in [WCAG 2.2 (Focus Not Obscured)](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html), and is not a great experience!
> 
> These problems are compounded, and there are several ways to alleviate them, some of which solve more than one problem at once. Carbon needs to decide what exactly the responsive design interaction is supposed to be. Happy to discuss the possibilities -- both optimal and acceptable -- from an accessibility perpspective!

### LINKED ISSUE #13417: Left UI Shell: The hamburger menu disclosure does not collapse on loss of focus
From the parent issue: 

> Problem 3: The hamburger menu disclosure does not collapse on loss of focus
> If the menu is activated, the hamburger menu is displayed. The user can eventually tab through it, but it does not disappear when the tab leaves the left nav.
> This is particularly odd since Carbon's implementation of the left nav when the hamburger menu ISN'T activated is much better. In that situation, the left nav is automatically revealled after focus leaves the top menu, and automatically hidden when focus leaves the left nav and enters the main content.
> If the behaviour of the activated left nav matched the behaviour when it was just revealed on focus, it would resolve or diminish almost every problem.

guidari:
Closes #13417 
Closes #13415 

I added a new prop to act like a function, so when the focus leaves the `SideNav` this fucntion will trigger and it will switch the `expanded` prop in the `HeaderContainer`

#### Testing / Reviewing

Switch to mobile view. Open the `SideNav` by click on the hamburger menu, then enter and leave the `SideNav` to test the behaviour


tay1orjones (changes requested):
This looks pretty good, and in most cases the menu closes as expected! 

I was able to find this case though where the menu does not close as expected. 

I click to put focus on the canvas, and then move focus through the menu. I get to the end of the menu, it does not close, and continue moving focus again with tab. You can see the button being focused behind the menu and the page scroll location moving as focus moves between the button and link on the page.

https://github.com/carbon-design-system/carbon/assets/3360588/63d37b6f-200b-4c92-9b8e-b0225a82d7aa

tay1orjones (approved):
Looks great!

francinelucca (changes requested):
This LGTM! I see how this solves for #13417 , but is this also solving for #13415 ? I don't see a difference in the content being obscured 🤔
image

francinelucca (approved):
Thanks for walking me through that @guidari !

PR #13717, opened 2023-05-02
feature/HeaderContainer types

### LINKED ISSUE #13586: Add TypeScript types to HeaderContainer

mrkjdy:
Closes #13586

Converted UIShell/HeaderContainer.js to tsx and added appropriate prop types. I also replaced prop `isCurrentPage` with `isActive` in `UIShell.stories.js` to fix a warning that was being logged.

#### Changelog

**New**

- Added interfaces for HeaderContainer and render props

**Changed**

- Renamed `UIShell/HeaderContainer.js` to `UIShell/HeaderContainer.tsx`
- Replaced prop `isCurrentPage` with `isActive` in `UIShell.stories.js` to fix warning

#### Testing / Reviewing

You can check that building still works and that there are no errors when viewing the UIShell stories that have a HeaderContainer


guidari (approved):
LGTM. Thanks for adding the types and remove a deprecated prop @mrkjdy 

You just have to fix the tests, please. To fix the issue, update the snapshots by running `yarn test -u` and then commit and push again. I tested locally and it worked

packages/react/src/components/UIShell/HeaderContainer.tsx
```diff
@@ -26,17 +38,13 @@ const HeaderContainer = ({ isSideNavExpanded, render: Children }) => {
...
 HeaderContainer.propTypes = {
   /**
    * Optionally provide a custom class name that is applied to the underlying <header>
    */
   isSideNavExpanded: PropTypes.bool,
-};
 
-HeaderContainer.defaultProps = {
-  isSideNavExpanded: false,
+  render: PropTypes.elementType,
```
tw15egan: Should this be `PropTypes.func`? We should also add a description of the prop if we are adding it to the `propType` definitions

packages/react/src/components/UIShell/HeaderContainer.js
```diff
@@ -1,42 +0,0 @@
-/**
```
mrkjdy: I used `git mv` on this file, but after my last push git considers it as deleted. I guess the file is different enough now that it considers it as deleted/new rather than renamed? Maybe this is a problem on GitHubs side?

packages/react/src/components/UIShell/HeaderContainer.tsx
mrkjdy: I added a description with an example. Should doc comments be added to the HeaderContainerProps interface as well? When I use the element in this repo vscode displays the doc comments from the propTypes for it, but I'm wondering if it will do that when this package is used

Also, I believe that elementType is the correct type for the render prop because the way it's used within the HeaderContainer prop is as an element:
https://github.com/carbon-design-system/carbon/blob/b0a17c4e5db755184763eec47391628ff6fd0a34/packages/react/src/components/UIShell/HeaderContainer.tsx#L35-L40
A function that returns a JSX.Element is a component

packages/react/src/components/UIShell/HeaderContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+   *        isPersistent={false}>
+   *        <HeaderSideNavItems>
+   *          <HeaderMenuItem href="/some-path">Some label</HeaderMenuItem>
+   *        </HeaderSideNavItems>
+   *      </SideNav>
+   *    </Theme>
+   *}/>
+   * ```
+   */
+  render: PropTypes.elementType.isRequired,
```
tw15egan, line 63: If this prop wasn't marked explicitly as required before, we should not make it required now as users downstream will consider this a breaking change.

packages/react/src/components/UIShell/HeaderContainer.tsx
```diff
@@ -0,0 +1,82 @@
...
+HeaderContainer.propTypes = {
+  /**
+   * Optionally provide a custom class name that is applied to the underlying <header>
+   */
+  isSideNavExpanded: PropTypes.bool,
+
+  /**
+   * A function that is passed two parameters, `isSideNavExpanded`, and
+   * `onClickSideNavExpand`, which can then be used the returned components. For
+   * example:
```
tw15egan: The comment looks good, I don't think we need the full example though as it is displayed in the stories.

packages/react/src/components/UIShell/HeaderContainer.tsx
tw15egan: > Should doc comments be added to the HeaderContainerProps interface as well?

I don't believe we've been adding with the other typescript refactors


> A function that returns a JSX.Element is a component

Thanks for clarifying! 👍🏻

packages/react/src/components/UIShell/HeaderContainer.js
tw15egan: I wonder if removing the large example in the comment will revert it back to `git mv` diff? Its fine either way 👍🏻

packages/react/src/components/UIShell/HeaderContainer.tsx
mrkjdy: If you look at how the component was/is implemented, it would not work if no render prop is passed. Adding isRequired should not break anything for users because the component would already not be working. By not requiring the prop, developers may try to use the component and run into an error trying to render undefined. Adding isRequired to the prop will make it clearer that they need to add it

Let me know if that makes sense or if I should still remove it

packages/react/src/components/UIShell/HeaderContainer.js
mrkjdy: It did! 😄

tw15egan (approved):
LGTM 👍 ✅

PR #13720, opened 2023-05-03
fix(vite): update example to use react 18 createRoot

tay1orjones:
This fixes the following error in the vite example

```
Warning: ReactDOM.render is no longer supported in React 18. 
Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. 
Learn more: https://reactjs.org/link/switch-to-createroot
```

#### Changelog

**Changed**

- update vite example to use `createRoot`

#### Testing / Reviewing

- The vite example should work and not display the console error anymore


guidari (approved):
LGTM. I tested locally and it worked fine!

alisonjoseph (approved):

PR #13735, opened 2023-05-04
chore(enable-v11-release): cleanup enable-v11-release feature flag usage

### LINKED ISSUE #13727: [Feature Request]: Cleanup `enable-v11-release` feature flag usage
### The problem

A few of the source code files still utilize `enable-v11-release` feature flag, which should have been unnecessary with the v11 release, per #8410.

### The solution

Suggestions from @tay1orjones:
- Set the enable-v11-release flag to true in the root .yml file in packages/feature-flags
- Remove any tests that were covering v10 code (without the feature flag)
- Remove FeatureFlag usages that turn on enable-v11-release in tests
- Fix any remaining failing tests (the bulk of tests should already be “fixed” and succeeding with the feature flag on)
- Remove the feature flag toggles within the components’ source

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

lewandom:
Closes #13727

Cleans up remaining usage of the `enable-v11-release` feature flag, which is obsolete since v11 release.

#### Changelog

**Changed**

- Set the enable-v11-release flag to true in the root .yml file in packages/feature-flags
- Fixed test failures caused by behavior changes induced by setting the above flag to true by default
- Adjusted PublicAPI snapshot to match removed defaultProps of undefined value

**Removed**

- Removed `enable-v11-release` flag usage in component code and in test code
- Removed tests specific to v10

#### Testing / Reviewing

All automated tests should be passing. Changed components should render in storybook without changes.


packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -260,7 +259,6 @@ const ComboBox = React.forwardRef((props: ComboBoxProps, ref) => {
     ['aria-label']: ariaLabel,
     ariaLabel: deprecatedAriaLabel,
     className: containerClassName,
-    direction,
```
lewandom: Comment from author: I removed this line to satisfy linter (unused variable), but it might be a bigger problem - the entire `direction` prop unused / not working in v11. Please review this one with special attention.

tw15egan: Interesting, it seems to be working [here](https://react.carbondesignsystem.com/?path=/story/components-combobox--playground&args=direction:top).

I think the issue is that just the `[enabled ? null : containerClassName],` line needs to be removed, not the entire `className` block.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -382,18 +380,8 @@ const ComboBox = React.forwardRef((props: ComboBoxProps, ref) => {
...
   const showWarning = !invalid && warn;
-  const className = cx(
-    `${prefix}--combo-box`,
-    [enabled ? null : containerClassName],
-    {
-      [`${prefix}--list-box--up`]: direction === 'top',
-      [`${prefix}--combo-box--warning`]: showWarning,
-      [`${prefix}--combo-box--readonly`]: readOnly,
-    }
-  );
```
tay1orjones, line 396: I don't think all of this should be removed. It's just the `containerClassName` that's behind the v11 feature flag. This will solve the issue with `direction` not being used.

packages/react/src/components/Copy/Copy.js
```diff
@@ -55,49 +54,25 @@ export default function Copy({
...
-        ])}
-        {...other}
-        aria-label={
-          (!children && (animation ? feedback : other['aria-label'])) || null
-        }>
-        {children}
-      </IconButton>
-    );
-  }
+  const initialLabel = other['aria-label'] ?? '';
```
tay1orjones, line 57: Why add this? Is it due to `label` being a required prop on `IconButton`?

packages/react/src/components/Tab/index.js
```diff
@@ -5,11 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import * as FeatureFlags from '@carbon/feature-flags';
-import { Tab as TabNext } from '../Tabs/Tabs';
-import { default as TabClassic } from './Tab';
-
-const Tab = FeatureFlags.enabled('enable-v11-release') ? TabNext : TabClassic;
```
tay1orjones, line 12: Can the source for TabClassic be deleted?

tay1orjones (changes requested):
Just a couple changes to make. Overall this looks really great! 🎉

packages/react/src/components/Copy/Copy.js
lewandom: Yes, tests are failing on prop types validation.

packages/react/src/components/ComboBox/ComboBox.tsx
lewandom: Thanks for spotting! I pushed a commit with partial revert.

packages/react/src/components/ComboBox/ComboBox.tsx
lewandom: Good finding! I pushed a commit with partial revert.

packages/react/src/components/Tab/index.js
lewandom: @tay1orjones I can delete it if that's the desired approach. I guess keeping it would only make sense if somebody imports the "classic" implementation explicitly.

tay1orjones: Yeah I support removing it 👍

lewandom: I pushed a commit to remove the "classic" Tab implementation.

tay1orjones (approved):
Thanks for making all those changes!

tw15egan (approved):
LGTM 👍 ✅

tw15egan:
Thanks for tackling this @lewandom 🎉 💪🏻

PR #13738, opened 2023-05-04
docs(migration): prefix removal carbon-- v11

### LINKED ISSUE #13687: Compat tokens: Better document `carbon--` prefix removal
In v10, many tokens/mixins/functions were prefixed with `carbon--`. In v11, this prefix was removed and no longer used. 

> Note: This prefix was not related to the _classname_ prefix `bx` that changed to `cds` in v11.

As part of migrating to v11, consumers have to do a global find and replace to remove the `carbon--` prefix from their usages, e.g.:

```diff
- @include carbon--breakpoint(lg) {
+ @include breakpoint(lg) {
      width: 42%;
  }
```

As part of adding the new token compat modules #13661, we will not be mapping v10 token values that had the `carbon--` prefix to the v11 token names. This simple "find all and replace all" shouldn't impact the time for consumers to migrate much at all. The docs on the website need to be updated to clearly explain this situation and state "Do a find all and replace to remove `carbon--`".

```[tasklist]
- [ ] https://github.com/carbon-design-system/carbon/pull/13738
- [ ] https://github.com/carbon-design-system/carbon-website/pull/3540
```

guidari:
Part of #13687 

The documentation has been enhanced and clarified to provide a clear understanding of the steps involved in the migration process

I also think I would have to update the repo `carbon-website`

If you take a look in this [doc](https://carbondesignsystem.com/migrating/guide/develop#step-5:-update-imports) it might make sense to provide the same information that the migration file contains

Let me know if make sense that change in `carbon-website` or only the migration file is enough


alisonjoseph (approved):

tay1orjones (approved):
Nice and simple! We will need to add this to the migration guide on `carbon-website` as well.

PR #13744, opened 2023-05-05
fix: a11y-dropdown-typeahead-downshift

### LINKED ISSUE #10634: [a11y]: Dropdown typeahead needs to handle space key differently
### Package

### Browser

Chrome

### Operating System

Windows

### Package version

10.51

### React version

7.51

keyboard

### Assistive technology

keyboard

### Description

Steps for vanilla:
- https://www.carbondesignsystem.com/components/dropdown/usage/#live-demo
- focus the example Dropdown and quickly type Option 4
- this should select Option 4, but the space opens the Dropdown and Option 2 is actually selected (because of #10628)

Steps for react:
- https://react.carbondesignsystem.com/?path=/story/components-dropdown--default
- focus the example Dropdown and quickly type Option 4
- space opens the Dropdown and Option 1 is selected

If the Dropdown is open, then typing Option 4 closes the Dropdown because that's the normal behavior of the space key.

I would expect [space to be added to the search string when using typeahead](https://github.com/openui/open-ui/issues/171#issuecomment-686016533), and not used to open/close the Dropdown.

### WCAG 2.1 Violation

_No response_

### CodeSandbox example

see description

### Steps to reproduce

see description

### Code of Conduct

andreancardona:
Closes #10634

DRAFT: 

- https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect#toggle-button


tay1orjones:
Thanks for opening this! So recapping our pairing session today:

- This unfortunately won't completely resolve https://github.com/carbon-design-system/carbon/issues/10634
- This PR ensures the <kbd>space</kbd> is properly gated from downshift, and yet it still isn't properly concatenating the space to the typeahead "search"
- It could be considered an improvement to the situation overall since it ensures the menu does not open when pressing space during a "search".
  - To test this, focus the dropdown and type `option 4`, the menu should not open. 
  - The resulting item shown after the search is still not correct though
    - this is a downshift bug, and I don't see a way around it at the moment https://github.com/carbon-design-system/carbon/issues/10634#issuecomment-1536568093

If others consider this an improvement, lets merge this.

If not, I think we can take a second shot at this when we update downshift https://github.com/carbon-design-system/carbon/issues/13749

tay1orjones (approved):

alisonjoseph (approved):
I think we're good to merge this, feels like an improvement to me

tw15egan (approved):

PR #13754, opened 2023-05-07
feat: added types to Link component

### LINKED ISSUE #13561: Add TypeScript types to Link

SunnyJohal:
Closes #13561

Added typescript types to Link component.

#### Changelog

**New**

- Added typescript types to Link component.

**Changed**

-  File extensions from .js to .tsx

**Removed**

- n/a

#### Testing / Reviewing

- Run storybook, navigate to link component and check that it is still rendering.
- Run unit tests


packages/react/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,167 @@
...
+    {
+      children,
+      className: customClassName,
+      href,
+      disabled = false,
+      inline = false,
+      visited = false,
+      renderIcon: Icon,
+      size,
+      target,
```
tay1orjones, line 81: `target` needs to be added to propTypes

tay1orjones (changes requested):
Thanks! Just one small change needed

packages/react/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,167 @@
...
+  AriaAttributes,
+  ComponentType,
+  HTMLAttributeAnchorTarget,
+  PropsWithChildren,
+} from 'react';
+import { usePrefix } from '../../internal/usePrefix';
+
+interface LinkProps {
+  /**
+   * @description Indicates the element that represents the
```
tay1orjones, line 21: The `@description` JSDoc annotation is [not supported](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#unsupported-tags)

```suggestion
   * Indicates the element that represents the
```

Additional usages of `@description` in this PR will need updated as well

tay1orjones (changes requested):

packages/react/src/components/Link/Link.tsx
SunnyJohal: Will remove these now

packages/react/src/components/Link/Link.tsx
SunnyJohal: Will add this now

andreancardona (approved):
Thank you so much for this - looks good!

tay1orjones (approved):

PR #13760, opened 2023-05-08
feature/HeaderMenuItem types

### LINKED ISSUE #13591: Add TypeScript types to HeaderMenuItem

mrkjdy:
Closes #13591 

Converted UIShell/HeaderMenuItem.js to tsx and added appropriate prop types. I also converted UIShell/Link.js to tsx as HeaderMenuItem uses it and there is no issue open to convert UIShell/Link.js.

#### Changelog

**New**

- Added interface for HeaderMenuItem props
- Added type for Link props

**Changed**

- Renamed UIShell/HeaderMenuItem.js to UIShell/HeaderMenuItem.tsx
- Renamed UIShell/Link.js to UIShell/Link.tsx

#### Testing / Reviewing

- building
- viewing the UIShell stories that have a HeaderMenuItem/Link
- creating a component that uses the HeaderMenuItem/Link and testing which props are valid/invalid


tay1orjones (approved):

tw15egan (approved):

packages/react/src/components/UIShell/Link.tsx
```diff
@@ -6,22 +6,49 @@
...
+  type ElementType,
+  type ForwardedRef,
+  forwardRef,
+  type WeakValidationMap,
+  type ComponentProps,
+  Ref,
+} from 'react';
+
+// Note: Maybe we should use `as` instead of `element`? `as` appears to be
+// standard and is used in other places in this project.
```
mrkjdy, line 20: @tay1orjones What do you think about switching to `as`? I could add a comment about `element` being deprecated and add `as` to the prop type

tay1orjones: @mrkjdy I'm not opposed to adding `as`, but we'll need to keep `element` around. We have a deprecation warning helper that can be placed on `element` proptypes. The `@deprecated` ts annotation can be added as well. We'll also need to map legacy usages of `element` to be the value  for `as` so nothing breaks for folks across the minor version this PR will be included in.

alisonjoseph (approved):

tay1orjones (approved):

PR #13765, opened 2023-05-08
feature/SideNav types

### LINKED ISSUE #13596: Add TypeScript types to SideNav

mrkjdy:
Closes #13596 

Converted UIShell/SideNav.js to tsx and added appropriate prop types.

#### Changelog

**New**

- Added type for SideNav props

**Changed**

- Renamed UIShell/SideNav.js to UIShell/SideNav.tsx
- Using `isValidElement()` to check if a child is valid before cloning

#### Testing / Reviewing

- building
- viewing the UIShell stories that have a SideNav
- creating a component that uses the SideNav and testing which props are valid/invalid


packages/react/src/components/UIShell/SideNav.tsx
```diff
@@ -65,11 +102,11 @@ const SideNav = React.forwardRef(function SideNav(props, ref) {
   //   : t('carbon.sidenav.state.closed');
 
   const className = cx({
+    ...(customClassName !== undefined && { [customClassName]: true }),
```
tay1orjones: What's the reason for this change? The previous `[customClassName]: !!customClassName` is a bit more readable to me

tay1orjones (approved):

packages/react/src/components/UIShell/SideNav.tsx
mrkjdy: TypeScript does not allow keys to be possibly undefined. But I updated it and it should be much more readable now!

alisonjoseph (approved):

packages/react/src/components/UIShell/SideNav.tsx
```diff
@@ -14,35 +23,63 @@ import { keys, match } from '../../internal/keyboard';
...
-  const {
+// const translations = {
+//   'carbon.sidenav.state.open': 'Close',
+//   'carbon.sidenav.state.closed': 'Open',
+// } as const;
+
+// type TranslationId = keyof typeof translations;
+
+// type Translation = typeof translations[TranslationId];
+
```
tay1orjones:
```suggestion
```

packages/react/src/components/UIShell/SideNav.tsx
tay1orjones: Nice! looks great

packages/react/src/components/UIShell/SideNav.tsx
mrkjdy: These depend on this removed code btw:
https://github.com/mrkjdy/carbon/blob/f1eab8039fb175743d8deb9079123ba0f29cab06/packages/react/src/components/UIShell/SideNav.tsx#L34-L35

https://github.com/mrkjdy/carbon/blob/f1eab8039fb175743d8deb9079123ba0f29cab06/packages/react/src/components/UIShell/SideNav.tsx#L54-L55

PR #13772, opened 2023-05-09
feature/SkipToContent types

### LINKED ISSUE #13583: Add TypeScript types to SkipToContent

mrkjdy:
Closes #13583 

Converted UIShell/SkipToContent.js to tsx and added appropriate prop types.

#### Changelog

**New**

- Added type for SkipToContent props

**Changed**

- Renamed UIShell/SkipToContent.js to UIShell/SkipToContent.tsx
- Changed children from being required and a string to being a ReactNode and optional
- Changed href from being required to being optional
- Updated the tests

#### Testing / Reviewing

- building
- viewing the UIShell stories that have a SkipToContent
- creating a component that uses the SkipToContent and testing which props are valid/invalid


packages/react/src/components/UIShell/SkipToContent.tsx
```diff
@@ -28,28 +30,21 @@ function SkipToContent({
 
 SkipToContent.propTypes = {
   /**
-   * Provide text to display in the SkipToContent `a` tag
+   * A ReactNode to display in the SkipToContent `a` tag.
+   * `'Skip to main content'` by default.
    */
-  children: PropTypes.string.isRequired,
```
tay1orjones, line 33: I think restricting this to a string was intentional

tay1orjones (changes requested):
One small note, otherwise looks good!

packages/react/src/components/UIShell/SkipToContent.tsx
mrkjdy: What if someone wants to put their own custom button or other component there? I decided to stick with the type that React has for the `children` prop of an `a` rather than overriding it. Let me know if that makes sense, otherwise I can restrict it to a string

alisonjoseph: I don't think we would ever want anything other than a string in here, but maybe I'm missing a use-case?

mrkjdy: I just didn't want to unnecessarily restrict people, but I can change it back and adjust the type so that it only accepts string children

alisonjoseph (approved):

tay1orjones (approved):

PR #13773, opened 2023-05-09
feature/Form types

### LINKED ISSUE #13553: Add TypeScript types to Form

mrkjdy:
Closes #13553 

Converted Form/Form.js to tsx and added appropriate prop types.

#### Changelog

**New**

- Created type for Form props

#### Testing / Reviewing

- building
- viewing the stories that have a Form
- creating a component that uses the Form and testing which props are valid/invalid


packages/react/src/components/Form/index.ts
```diff
@@ -5,6 +5,4 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import Form from './Form';
-export default Form;
-export { Form };
```
tay1orjones, line 10: I think we'll need to keep the named export in the event that someone is relying on it (unlikely but possible)

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -3637,16 +3637,6 @@ Map {
...
-  "Form" => Object {
-    "propTypes": Object {
-      "children": Object {
-        "type": "node",
-      },
-      "className": Object {
-        "type": "string",
-      },
-    },
-  },
```
tay1orjones: This should be removing from this snapshot

packages/react/src/__tests__/index-test.js
```diff
@@ -66,7 +66,6 @@ describe('Carbon Components React', () => {
         "FilterableMultiSelect",
         "FlexGrid",
         "FluidForm",
-        "Form",
```
tay1orjones: Same here, this shouldn't be removed.

tay1orjones (changes requested):

packages/react/src/components/Form/index.ts
mrkjdy: Should be fixed!

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
mrkjdy: Should be fixed!

packages/react/src/__tests__/index-test.js
mrkjdy: Should be fixed!

andreancardona (approved):

tay1orjones (approved):

PR #13776, opened 2023-05-09
fix(Type): add v10 type tokens

### LINKED ISSUE #13686: Compat tokens: add type tokens module
As part of #13661, this issue is for updating `@carbon/type` to include all v10 type tokens and values alongside v11 type tokens in both sass and javascript.

```[tasklist]
- [x] Gather a list of the tokens that changed for this package from [the v11 changelog](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/v11.md#type-tokens)
- [x] Update the root token definition in js and sass and ensure the output is generated correctly for generated sass stylesheets
- [x] Write a sass test to ensure the tokens are included when the module is `@use`d and that both v10 and v11 tokens can be used in the same file (with or without namespaces)
```

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13686

Ensures all v10 type tokens can be referenced in v11, without causing any compilation errors. Token changes can be seen [here](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/v11.md#type-tokens)

#### Changelog

**New**

- `caption-01` and `caption-02` re-added to export, marked as deprecated
- Test to ensure tokens can be referenced. 

#### Testing / Reviewing

Find a story in the React storybook that has a stylesheet, and try setting an element to a v10 type token

For example:

```scss
@use '../../../scss/type';


div {
    @include type.type-style('caption-01');
}
```


tay1orjones (approved):
Awesome!

tay1orjones:
I think snaps just need updated

guidari (approved):
Cool! I tested on `_popover.scss` and LGTM

PR #13783, opened 2023-05-10
fix(Layout): add v10 layout tokens

### LINKED ISSUE #13688: Compat tokens: add layout/spacing tokens module
As part of #13661, this issue is for updating `@carbon/layout` to include all v10 layout/spacing tokens and values alongside v11 layout/spacing tokens in both sass and javascript.

```[tasklist]
- [ ] Gather a list of the tokens that changed for this package from [the v11 changelog](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/v11.md)
- [ ] Update the root token definition in js and sass and ensure the output is generated correctly for generated sass stylesheets
- [ ] Write a sass test to ensure the tokens are included when the module is `@use`d and that both v10 and v11 tokens can be used in the same file (with or without namespaces)
```

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13688

Ensures all v10 layout tokens can be referenced in v11, without causing any compilation errors. Token changes can be seen [here](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/v11.md#scssglobalsscss_spacingscss)

Note: I did NOT add any tokens that are not even currently exported in v10 (deprecated v9 tokens like $spacing-4xs and `$layout-4xs`). I also did not add any `carbon--` aliases, as this should be handled via a global find + replace to simply remove all `--carbon` instances

#### Changelog

**New**

- `$layout-01` through `$layout-07` re-exported.
- Snapshot tests for the `@carbon/styles/scss/spacing` entry point

**Changed**

- Updated tests in `@carbon/layout`
- Updated build task to generate `_layout.scss` 

#### Testing / Reviewing

Find a story in the React storybook that has a stylesheet, and try setting an element to a v10 layout token

For example:

```scss
@use '@carbon/styles/scss/spacing';


div {
    padding: spacing.$layout-01;
}
```


francinelucca (approved):
LGTM!

guidari (approved):
LGTM! I tested the style and it worked

alisonjoseph (approved):

PR #13790, opened 2023-05-11
feature/Toggletip types

### LINKED ISSUE #13543: Add TypeScript types to Toggletip, ToggletipActions, ToggletipButton, ToggletipContent, ToggletipLabel

mrkjdy:
Closes #13543  

Converted `Toggletip`, `ToggletipActions`, `ToggletipButton`, `ToggletipContent`, and `ToggletipLabel` to tsx and added appropriate prop types. I also converted `useEvent` to TypeScript as it is used in `Toggletip`. I noticed that `Popover`s type definition didn't allow it to be used as a Polymorphic component properly, so I made some improvements to it as well. If you would like me to revert the changes to Popover I can, but it should work better now. Additionally, I checked if there were any type errors with `tsc` and noticed some minor export issues in a few JS files, so I fixed those as well. If you would like me to change their extension to `.ts` or revert the changes there please let me know as well.

#### Changelog

**New**

- Created types for `Toggletip`, `ToggletipActions`, `ToggletipButton`, `ToggletipContent`, and `ToggletipLabel` props
- Typed `useEvent` and `useWindowEvent`

**Changed**

- Made `Popover`s type polymorphic
- Fixed some typescript compile errors in a few JS files caused by incorrect export syntax
- Corrected a few propTypes
- Fixed a type error in `Tooltip`

#### Testing / Reviewing

- building
- viewing the stories that have a `Toggletip`, `ToggletipActions`, `ToggletipButton`, `ToggletipContent`, or `ToggletipLabel`
- creating a component that uses the `Toggletip`, `ToggletipActions`, `ToggletipButton`, `ToggletipContent`, or `ToggletipLabel`, or `Popover` and testing which props are valid/invalid


packages/react/src/internal/useEvent.ts
```diff
@@ -0,0 +1,63 @@
...
+export function useEvent(
+  elementOrRef: HTMLElement | MutableRefObject<HTMLElement | null>,
+  eventName: keyof GlobalEventHandlersEventMap,
+  callback: EventCallback
+) {
+  const savedCallback = useRef<EventCallback | null>(null);
+
+  useEffect(() => {
+    savedCallback.current = callback;
+  }, [callback]);
```
francinelucca: per our [guidelines documentation](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md#steps-to-provide-baseline-type-definitions-for-components), we want to avoid converting internal helpers to TypeScript
image

can we revert this back to JS and add JsDocs instead? 🙏🏻

packages/react/src/components/Popover/index.tsx
```diff
@@ -393,7 +403,9 @@ const PopoverContent = React.forwardRef(function PopoverContent(
       <span className={`${prefix}--popover-caret`} />
     </span>
   );
-});
+}
+
+export const PopoverContent = React.forwardRef(PopoverContentRenderFunction);
```
francinelucca, line 408: I think we'd still want to keep the `Popover` named export here
image

packages/react/src/components/Toggletip/index.tsx
```diff
@@ -173,7 +206,7 @@ Toggletip.propTypes = {
    * Provide a custom element or component to render the top-level node for the
    * component.
    */
-  as: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
+  as: PropTypes.elementType,
```
francinelucca, line 209: will making this change still allow for strings such as "div" to be passed in?

francinelucca (changes requested):
thanks for putting this together @mrkjdy! mostly LGTM, just some comments

packages/react/src/components/Popover/index.tsx
mrkjdy: It's still exported, I moved it to where it's declared if that's alright:
https://github.com/carbon-design-system/carbon/blob/e28b49d0c84017d38f69c9c110e8dc758d21756d/packages/react/src/components/Popover/index.tsx#L303

packages/react/src/internal/useEvent.ts
mrkjdy: This should be done now! One of the benefits of the types is that it lets your editor know which `eventName`s are valid and gives you suggestions

packages/react/src/components/Toggletip/index.tsx
mrkjdy: Yes! I tested passing a string into this component in a story and no warnings were printed about it. Any `keyof JSX.InstrinsicElements` should be valid. Any other string would not be a valid.

packages/react/src/components/Popover/index.tsx
francinelucca: I see! nice, thanks!

francinelucca (approved):

tay1orjones (approved):

PR #13803, opened 2023-05-15
feat(popover): add onRequestClose prop

### LINKED ISSUE #13408: [Bug]: Popover click interaction behavior does not match spec'ed guidance
### Package

@carbon/react

### Browser

Chrome

### Package version

1.25.0

### React version

17

### Description

From the [website](https://carbondesignsystem.com/components/popover/usage/#behaviors):

> *Behavior -> Interactions -> Mouse*
> Click: Open the popover by clicking on the trigger button. Close the popover by clicking on the trigger button again, *or anywhere outside of the open popover container*.

So I see two things we need to address:

* Popover implementation does not limit only one popover open at a time ([see example here](https://react.carbondesignsystem.com/?path=/story/components-popover--tab-tip), both tab tip popovers can be open at the same time). Although, this can be easily implemented in an application via state stored in a shared context.
* Clicking outside the popover does not dismiss the popover 

While Popover doesn’t have these two details correct, [Toggletip](https://react.carbondesignsystem.com/?path=/story/components-toggletip--default) _does_ have both of these behaviors since it’s open/close is tied to focus of the trigger element. Maybe we could lift this up out of Toggletip and into the Popover so all popover-based elements could benefit from the change?

### Reproduction/example

You can see in storybook: https://react.carbondesignsystem.com/?path=/story/components-popover--tab-tip

### Steps to reproduce

Click to open both popovers, more than one is rendered at a time.

Click outside of the popover, it does not close the already open popover(s)

tay1orjones:
Closes #13408 

This mirrors the approach of `onRequestClose` from `Modal` to `Popover`

#### Changelog

**New**

- add `onRequestClose` prop to popover that is called when a click happens outside the popover and/or when the browser window is blurred

**Changed**

- update tabtip popover story to show 
  - popovers can be limited to only one open at a time
  - clicking outside the popover dismisses the popover
- update storybook docs describing the purpose of this prop

#### Testing / Reviewing

- View the Tabtib story for popover, only one popover should be allowed to be open at a time.
- Opening a popover and then clicking outside of it should close the popover.


francinelucca (approved):
LGTM! looks like there's a linting error @tay1orjones 
image

packages/react/src/components/Popover/index.tsx
```diff
@@ -370,6 +391,12 @@ Popover.propTypes = {
    */
   isTabTip: PropTypes.bool,
 
+  /**
+   * Specify a handler for closing popover.
+   * The handler should care of closing the popover, e.g. changing the `open` prop.
```
alisonjoseph:
```suggestion
   * The handler should take care of closing the popover, e.g. changing the `open` prop.
```

alisonjoseph (approved):
Nice! LGTM (minor copy update to code comment)

PR #13805, opened 2023-05-15
fix(Motion): add v10 Motion tokens

### LINKED ISSUE #13689: Compat tokens: add motion tokens module
As part of #13661, this issue is for updating `@carbon/motion` to include all v10 motion tokens and values alongside v11 motion tokens in both sass and javascript.

```[tasklist]
- [ ] Gather a list of the tokens that changed for this package from [the v11 changelog](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/v11.md)
- [ ] Update the root token definition in js and sass and ensure the output is generated correctly for generated sass stylesheets
- [ ] Write a sass test to ensure the tokens are included when the module is `@use`d and that both v10 and v11 tokens can be used in the same file (with or without namespaces)
```

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13689

Ensures all v10 motion tokens can be referenced in v11, without causing any compilation errors. Token changes can be seen [here](https://github.com/carbon-design-system/carbon/blob/main/docs/migration/v11.md#carbonmotion)

Note: I did not add any `carbon--` aliases, as this should be handled via a global find + replace to simply remove all `--carbon` instances

#### Changelog

**New**
- `$fast-01` through `$slow-02`
- Compat tests in `@carbon/styles`

**Changed**
- V11 tokens added to js export
- snapshots updated

#### Testing / Reviewing

Find a story in the React storybook that has a stylesheet, and try setting an element to a v10 motion token

For example:

```scss
@use '@carbon/styles/scss/motion';


div {
    transition: all motion.$fast-01;
}
```


tay1orjones (approved):
LGTM 🎉

andreancardona (approved):
🔥

PR #13809, opened 2023-05-16
UIShell: Adjust focus order when opening the menu

### LINKED ISSUE #13416: Left UI Shell: The Focus Order fails

guidari:
Closes #13416 

**This pull request has been opened as a resolution to the merge conflicts that could not be resolved in the [previous pull request](https://github.com/carbon-design-system/carbon/pull/13670). The conflicts happened due to a change from .js to .tsx in the main branch, which impacted the merging process.**

Redirecting the navigation to the SideNav component when it's expanded and is not a fixed side nav.

### Testing / Reviewing
First switch to mobile view, then open the hamburguer menu and press tab.

You should be redirected to the `SideNav`


francinelucca (approved):
🚀

andreancardona (approved):
😎

PR #13816, opened 2023-05-17
fix(Multiselect): prevent console error when typing and open

### LINKED ISSUE #9938: Carbon Components React: MultiSelect not filterable - bug when typing if menu open
Hello!

There is currently a bug with the `MultiSelect` component, when **not filterable**.

This can be reproduced in storybook here: https://react.carbondesignsystem.com/?path=/story/components-multiselect--default

If you open the multiselect and begin typing (note: difficult in storybook due to shortcut commands), an error occurs.
```Uncaught TypeError: items.map is not a function```

The error relates to this code:
```
var _useSelect = (0, _downshift.useSelect)((0, _createPropAdapter.mapDownshiftProps)(_objectSpread(_objectSpread({}, downshiftProps), {}, {
    highlightedIndex: highlightedIndex,
    isOpen: isOpen,
    itemToString: function itemToString(items) {
      return Array.isArray(items) && items.map(function (item) {
        return _itemToString(item);
      }).join(', ') || '';
    },
    onStateChange: onStateChange,
    selectedItem: controlledSelectedItems,
    items: items
  }))),
```

The itemToString part fails due to `items` not being an array. Currently patching with the following:
```
return Array.isArray(items) && items.map(function (item) {
        return _itemToString(item);
      }).join(', ') || '';
```

One solution might be to remove the event listener from the Select when not filterable, unless there are keys used for opening/closing the menu in which case maybe more specific key down events.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/9938

Fixes an issue that would cause a console warning if the menu was open and a user started typing

#### Changelog

**Changed**

- Checks if `items` is an Array before trying to map over it

#### Testing / Reviewing

Go to `MultiSelect`, open the menu, and try typing. Verify no console errors appear when typing.

### DIFF
--- a/packages/react/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/react/src/components/MultiSelect/MultiSelect.tsx
@@ -339,7 +339,15 @@ const MultiSelect = React.forwardRef(function MultiSelect<ItemType>(
     highlightedIndex,
     isOpen,
     itemToString: (items) => {
-      return (items as ItemType[]).map((item) => itemToString(item)).join(', ');
+      return (
+        (Array.isArray(items) &&
+          items
+            .map(function (item) {
+              return itemToString(item);
+            })
+            .join(', ')) ||
+        ''
+      );
     },
     onStateChange,
     selectedItem: controlledSelectedItems,


francinelucca (approved):
lgtm

tay1orjones (approved):

PR #13821, opened 2023-05-17
fix: make ActionableNotification `subtitle` proptypes node to allow objects

crhuff-ibm:
Make actionable notification subtitle proptypes to allow more than just a string. If the consumer wants things like bold text, hyperlinks, multiple lines for text, a string doesn't allow it without throwing warnings.

#### Changelog

**Changed**

- Notification prop types for subtitle

#### Testing / Reviewing

Test ~Inline or~ Actionable Notification with `subtitle={<div>text</div>}` instead of `subtitle="text"` and view no warnings.


tw15egan (approved):

tay1orjones (changes requested):
I think this change should probably only be made to `ActionableNotification`.

`InlineNotification` and `ToastNotification` can only have a role of `alert`, `status`, or `log`, which means:
- All semantic value (rich text, line breaks, etc) placed in the `subtitle` would not be read by screenreaders.
- These components can't contain interactive elements like hyperlinks

For those reasons I think it makes the most sense to keep `subtitle` as a plain string in `InlineNotification` and `ToastNotification`.

Opening up `subtitle` to be a node in `ActionableNotification` should be fine though because it gets a role of `alertdialog`, which preserves semantic value and can include interactive elements.

tay1orjones (approved):
@crhuff-ibm Sorry for the delayed response here. If you're using any rich text decoration you'll need to use `ActionableNotification`. 

I've updated the PR so that `ActionableNotification`'s `actionButtonLabel` is now optional and the button isn't rendered if the `actionButtonLabel` isn't provided. This makes it so you can configure just `subtitle` without an action.

With this PR, your usage would look something like this:
```jsx
export default function App() {
  const subtitle = (
    <span>
      I've got some <strong>rich</strong> <em>text</em>{' '}
      <strong>
        <em>here</em>
      </strong>
    </span>
  );
  return (
    <div>
      <ActionableNotification inline subtitle={subtitle} />
    </div>
  );
}
```

tw15egan (approved):
Ideal tradeoff 👍🏻  
LGTM 👍 ✅

crhuff-ibm:
Deal, thank you for the compromise!

PR #13831, opened 2023-05-19
perf(DataTable): speed up select-all with large tables

### LINKED ISSUE #4338: [DataTable] Performance: Select All takes 5 - 30 s for tables with 1000+ rows

### LINKED ISSUE #6146: DataTable: select all action is slow and block ui when the table has thousands of rows even with pagination
### Summary
I use DataTable to list my data, because there may have lots of rows, so i implement pagination for this table. For example 6000 rows.  The select all action is slow and block ui for about 6 seconds. If there are 50000 rows of data, the section all action can crash my browser. However, the other actions like search, sort, pagination is fast.
image
The wired thing is, select one by one is fast. 
And the cancel action is slow even there is only one selected and may crash browser if has more data.
image

Here is my pagination code. If without the pagination, the select all can be slow too. I think the select all may need optimization inside the DataTable. 
```
<TableBody>
                  {rows
                    .slice((page - 1) * pageSize, page * pageSize)
                    .map(row => (
                      <TableRow key={row.id}>
                        <TableSelectRow {...getSelectionProps({ row })} />
                        {row.cells.map(cell => getCell(cell))}
                      </TableRow>
                    ))}
                </TableBody>

```
### Justification

This issue has effect on the user experience of our product.

mrspeaker:
Rewrite `setAllSelectedState` to improve performance with large tables. In my test case on my machine with 10000 rows, selecting all rows previously took 14000ms, this fix reduces it to 8ms. It uses one instance of (local) variable mutation.

Closes #6146
Closes #4338

The `setAllSelectedState` function is called when you click the *select-all* (or *deselect all*), or select one row then hit the *cancel* button on the bulk-action toolbar (in `handleOnCancel`). It appears to be "accidentally quadratic" and starts to become an issue with more than a few thousand rows. To avoid it, I do a local mutation of the accumulator rather than re-splatting out the entire object each iteration.

For 10000 rows, this reduces the original 14 second time (this is for Vite transpiled code, 8 seconds for native), down to ~250ms. Most of that time is in the `filteredRowIds.includes` check. So I added an extra flag (`isFiltered`) to avoid calling when the filter is not in use. When there are no filtered rows, the function runs in ~8ms. 

#### Testing / Reviewing

The fix does not break any existing tests, and should not introduce any changes to the existing behavior of the DataTable: Selecting-all/deselecting all should work as expected. When using the bulk-action toolbar, clicking cancel should deselect the row. If rows are filtered, clicking select-all should select all filtered rows.

To test the performance, you need to make a table with a lot of (at least 10000) rows. With a large dataset like this, the performance improvements are obvious, but if you want to measure it then wrap calls to `setAllSelectedState` with console.time messages. E.g, inside the DataTable's `handleOnCancel` function:
```
...
console.log("Cancel button was hit - call setAllSelectedState");
console.time("setAllSelectedState");
const output = this.setAllSelectedState(state, false, this.getFilteredRowIds());
console.timeEnd("setAllSelectedState");
return {
  shouldShowBatchActions: false,
   ...output
};
```


mrspeaker:

tay1orjones (approved):
@mrspeaker Thank you, this is great! 🎉  This will also close #4338, I've already updated the issue body to reflect this.

Do you think we could speed it up even more by not relying on `reduce` and just using a for loop instead to build up the rowsByID?

tw15egan (approved):

alisonjoseph (approved):

mrspeaker:
> 



> Do you think we could speed it up even more by not relying on `reduce` and just using a for loop instead to build up the rowsByID?

@tay1orjones - strangely, I thought the same - but when I tested it the `for` loop was every-so-slightly slower! I'm not sure why (and I bet it's browser specific), but in either case the time for the loop itself is negligible compared to the loop body. I decided to keep the `reduce` there to more closely match the coding style of the code base.

tay1orjones:
@mrspeaker Very interesting, thanks for testing it out!

PR #13850, opened 2023-05-22
feat: add enterDelayMs-type prop to delay displaying SideNav component

### LINKED ISSUE #13791: Add enterDelayMs-type prop to SideNav isRail
### The problem

`<SideNav isRail>` is the most elegant nav solution Carbon offers, and it truly is great...except in one regard, where it is just plain irritating: it is wayyyyy too eager to open on mouse hover. Most of our pages contain DataTables, which consistently have expansion and checkbox components at their left edges. So very frequently, people accidentally mouse-stray a little too far left and into the rail zone, which causes the sidenav to flash open **_instantly_**...so then they have to move the mouse back to the right a significant distance to get out of the now-open sidenav area to get it to close...then CAREFULLY move the mouse back to the left edge of the page body without accidentally opening the nav again. This happens all day long, so it's become a constant frustration and usability issue.

Horizontal space is **_extremely_** precious/valuable in our app (DataTables galore with lots of columns), so isRail is a must, and for the same reason we can't just add arbitrary dead space between the rail and the table to act as accidental mouse-stray mitigation.

### The solution

Why not offer a prop like your Tooltip component has, **enterDelayMs**? That way we can prevent the nav from opening accidentally when the mouse strays into the rail zone for just a fraction of a second. People are **_aiming_** for the expansion chevron or checkbox, so the mouse cursor only strays into the rail for, say, a tenth of a second. If people linger for, I dunno...half a second to a full second, then sure...they obviously want to stay there and open the nav. We'll figure out the ideal delay ourselves with trials once the option is available.

Taylor Jones thought this was worthy of opening an enhancement for, and IMHO this would really improve SideNav usability / user happiness. Hopefully it won't be a lot of work, so a very nice ROI.

### Examples

Just view your own example here:

https://react.carbondesignsystem.com/?path=/story/components-ui-shell--side-nav-rail

This has a lot of dead buffer space between the body and nav to prevent accidental strays, mind you, but the idea is similar: do a quick mouse circle around the page body and get close to the rail...the instant you stray into the rail area, BOOM! the nav slams open. It's just way too eager. If it only just waited a small amount of time by default, even that would be a big improvement...but a user-specific delay would be best (so we can choose how long is long enough).

### Application/PAL

CDD Hub

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

Shankar-CodeJunkie:
Closes #13791

***Add enterDelayMs-type prop to Specify the duration in milliseconds to delay before displaying the sidenavigation***

#### Changelog

**Changed**

- Modified SideNav.tsx to include a new prop of type `enterDelayMs` to specify the duration in milliseconds to delay before displaying the sidenavigation
- Modified UIShell.stories.js to include a playground story, where developers can test with two input params from browser. Earlier, the playground part for UIShell wasn't available, and hence I created a new one that expects two props ( isRail & enterDelayMs

#### Testing / Reviewing

- Ran yarn test
- I have also ran story book testing and modified the playground by giving different duration for prop `enterDelayMs` and I see the delay in opening up the sidenavigation. 

image


Shankar-CodeJunkie:
I see one of the test is failing and I am investigating the reason for it .  I will push the changes, once I have a fix for it . Thanks

packages/react/src/components/UIShell/SideNav.tsx
```diff
@@ -101,7 +103,7 @@ function SideNavRenderFunction(
   //   : t('carbon.sidenav.state.closed');
 
   const className = cx(customClassName, {
-    [`${prefix}--side-nav`]: true,
+    [`${prefix}--side-nav`]: false,
```
tw15egan: I don't think we want to set this class to false, this is where all of the style regressions are coming from. If we need to modify the `isRail` styles, we should add a new class and unset / reset styles on an as-needed basis.

tw15egan (changes requested):
Visual regressions due to class name removal, playground story respects the enter delay but it just flashes open and closes immediately.

packages/react/src/components/UIShell/SideNav.tsx
Shankar-CodeJunkie: @tw15egan : Yeah correct !  I will make the change and push the code.

Shankar-CodeJunkie: @tw15egan : While working on this , I got one more thought, what if,   isRail and enterDelay is enabled to hide the expansion for a few seconds, but  user is intentionally clicking the button expecting to see it immediately. 

I just introduced a onClick event to override the timedelay set by the developer, if user wants to see it immediately ! .  Let me know if this is not needed

packages/styles/scss/components/ui-shell/side-nav/_side-nav.scss
```diff
@@ -68,7 +68,7 @@
     width: 0;
   }
 
-  .#{$prefix}--side-nav.#{$prefix}--side-nav--rail:not(
+  .#{$prefix}--side-nav.#{$prefix}--side-nav--rail.#{prefix}--cds--side-nav--expand-on-hover:not(
       .#{$prefix}--side-nav--fixed
     ):hover,
```
tw15egan:
```suggestion
```

It seems like we can just remove this block, and just leave `.#{$prefix}--side-nav--expanded`

packages/react/src/components/UIShell/SideNav.tsx
```diff
@@ -107,6 +109,8 @@ function SideNavRenderFunction(
     [`${prefix}--side-nav--rail`]: isRail,
     [`${prefix}--side-nav--ux`]: isChildOfHeader,
     [`${prefix}--side-nav--hidden`]: !isPersistent,
+    [`${prefix}--cds--side-nav--expand-on-hover`]:
+      expanded || expandedViaHoverState,
```
tw15egan:
```suggestion
```

I think we can delete this, as it is toggling based on the same props as `${prefix}--side-nav--expanded`

tw15egan (changes requested):
I left some comments, testing locally I could achieve the same functionality without the additional classes. Let me know if I missed anything though!

Shankar-CodeJunkie:
Thanks @tw15egan .  Yes, I removed that additional blocks in the scss file, and it worked fine without the need of additional  class.

packages/react/src/components/UIShell/UIShell.stories.js
```diff
@@ -1019,4 +1022,76 @@ export const SideNavWLargeSideNavItems = () => (
   </>
 );
 
+const UIShellPlayground = (props) => {
```
tw15egan: Instead of adding a playground story for `isRail`, can we just add the new `enterDelayMs` control to the existing `SideNav w/ Rail` stories? If not, can we populate it with content so that it looks like [this story](https://react.carbondesignsystem.com/?path=/story/components-ui-shell--side-nav-rail-w-header)?

tw15egan (approved):
Small documentation issue, but the functionality looks great. Thanks for working on this!

packages/react/src/components/UIShell/UIShell.stories.js
Shankar-CodeJunkie: @tw15egan : Thanks @tw15egan  for your comments .   I populated the playground story with the  complete contents like the other story you had referenced in your comment.

francinelucca (changes requested):
This is looking great! I agree with @tw15egan , @Shankar-CodeJunkie could we put the controls in the sidenav rail story https://deploy-preview-13850--carbon-components-react.netlify.app/?path=/story/components-ui-shell--side-nav-rail instead of adding a playground for it? I believe adding a playground to the UIShell would involve a bigger effort than just these two controls since it's an umbrella for a lot of subcomps and we might not want to tackle that in the scope of this pr

Shankar-CodeJunkie:
@tw15egan @francinelucca :  As recommended, I've removed the playground story and modified the existing 
SideNav w/ Rail stories by adding the props for enterDelayMs for 1 and 3 seconds.

francinelucca:
This is almost there, @Shankar-CodeJunkie we meant to add a customizable control for the `enterDelayMs` prop so that it's customizable in the story, here's an example with `ClickableTile` and the disabled prop: https://react.carbondesignsystem.com/?path=/story/components-tile--clickable

image

https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tile/Tile.stories.js#L79

image

Shankar-CodeJunkie:
@francinelucca !.    Oh yes ..  I see it now, what needs to  be changed ! ..  Thanks for your help !


When I make the change, I see an extra `-` appears on the description column, and I m not sure how to remove it .   Please let me know if there is a way to remove it.  I tried to check other stories, and I am not sure, where the description is being rendered on the page


image

tw15egan:
@Shankar-CodeJunkie It seems like it's having an issue pulling the description from the `SideNav.propTypes` definition inside `SideNav.tsx`. Not sure why though 🤔

packages/react/src/components/UIShell/UIShell.stories.js
```diff
@@ -865,9 +868,26 @@ export const SideNavRail = () => (
...
+  enterDelayMs: {
+    control: {
+      type: 'number',
+    },
+    defaultValue: 1000,
+    description:
+      'Specify the duration in milliseconds to delay before displaying the sidenav',
+  },
+};
+
```
francinelucca: Adding description, defaultValue and setting enterDelayMs to 100 instead of 1000
```suggestion
SideNavRail.argTypes = {
  isRail: {
    control: {
      type: 'boolean',
    },
    defaultValue: true,
    table: {
      defaultValue: { summary: true }
    },
    description: 'Optional prop to display the side nav rail.'
  },
  enterDelayMs: {
    control: {
      type: 'number',
    },
    table: {
      defaultValue: { summary: 100 },
    },
    defaultValue: 100,
    description:  'Specify the duration in milliseconds to delay before displaying the sidenav'
  },
};

```

packages/react/src/components/UIShell/UIShell.stories.js
```diff
@@ -973,6 +993,23 @@ export const SideNavRailWHeader = () => (
...
```
francinelucca:
```suggestion
SideNavRailWHeader.argTypes = {
  isRail: {
    control: {
      type: 'boolean',
    },
    defaultValue: true,
    table: {
      defaultValue: { summary: true }
    },
    description: 'Optional prop to display the side nav rail.'
  },
  enterDelayMs: {
    control: {
      type: 'number',
    },
    table: {
      defaultValue: { summary: 100 },
    },
    defaultValue: 100,
    description:  'Specify the duration in milliseconds to delay before displaying the sidenav'
  },
};

```

francinelucca (changes requested):
Not sure why it's adding the '-' in the bottom but I wouldn't worry about it, other than that LGTM

francinelucca (approved):

Shankar-CodeJunkie:
Thanks @francinelucca  and @tw15egan for your help !.

PR #13864, opened 2023-05-23
fix(filename): provide aria-describedby to errored filenames

### LINKED ISSUE #11202: [a11y]:  File uploader error message not announced or exposed on button focus
### Package

@carbon/react

### Browser

Chrome

### Operating System

_No response_

### Package version

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader

### React version

_No response_

n/a

### Assistive technology

JAWS

### Description

When a user uploads too large a file, a message "File size exceeds limit" is displayed, along with a warning symbol. If this implementation is going to be accessible the message:
1. needs to be exposed when it appears
2. needs to be announced when focus lands on the associated delete button.

image

Problems captured in [ issues with JAWS in video](https://ibm.box.com/s/9g8i7u4jyosqhdpjz2mhn2fwkrkxdxzy). 

I'll mention that I suspect this 'file' is not even equivalent to the regular file that appears after uploading (since the file has not uploaded, right). It is probably better to just expose this error message as a modal. It would never appear in the list because it wasn't uploaded. That would involve design, but is the recommended approach.

Barring that, putting both the error and  the message into an alert region and making it an aria-describedby of the delete button would accomplish it, I think.
>File size exceeds limit
500kb max file size. Select a new file and try again.
 

### WCAG 2.1 Violation

[WCAG 3.3.1 Error Identification](https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-identified.html), etc.

### CodeSandbox example

https://react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application

### Steps to reproduce

To replicate
1. Start JAWS
2. choose a file that is larger than 500kb for upload. The focus seems to move to the new file and it will be outlined in red. However, focus is not actually there, and the text will not be announced
3. Tab to the 'file' in error. The text will not be announced.

### Code of Conduct

tay1orjones:
Closes #11202 

#### Changelog

**Changed**

- provide `aria-describedby` to errored filenames, associating the existing error message with the button

#### Testing / Reviewing

- Open the fileuploader `drag-and-drop-upload-container-example-application` story
- Upload a file over 500kb to get the error message
- JAWS should read the error message when it pops (Andrea and I confirmed this with JAWS)


andreancardona (approved):
🔥

alisonjoseph (approved):

PR #13867, opened 2023-05-23
fix(Select): storybook defaults cleanup

### LINKED ISSUE #13008: [a11y]: Select component implementation issues
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-select--default#storybook-preview-wrapper

### React version

_No response_

n/a

### Assistive technology

_No response_

### Description

During a rapid review of Carbon components, we identified a number of accessibility considerations. This issue captures those concerns, along with some implementation challenges identified by our more detailed review of the Select component, as part of our update of the accessibility content. 
Findings that are not specifically implementation issues are captured in a [separate issue on the carbon website repo](https://github.com/carbon-design-system/carbon-website/issues/3361). For the curious, the [original recording of the quick review](https://ibm.webex.com/ibm/ldr.php?RCID=210905c96aad3c6f890cc94fdc4f9523) is reachable (Password: wHhv2TW9 ). Select begins at about 11:25 in the recording.

Recording link: 

Key findings:

1. The default text "Choose an option" is not an actual option in the list. Effectively Carbon is using placeholder text to provide instructions. This is non-optimal for a number of reasons, gone into in the usage.pdf comments. Suffice it to say, that typically, the pre-populated option in a select should be one of the options in the list. It is _okay_ to have a blank, and that would be the recommendation if you do not want to have a  pre-populated option selected. Having a blank also arguably makes it easier for a user to tell when a select has not been completed.
The third item in https://github.com/carbon-design-system/carbon/issues/12721 about dropdowns covers the same basic problem.
I am aware that it is _possible_ to put in a pre-selected value that the user can never navigate to, etc. I would just argue it is not optimal and can lead to unintended issues, like the following...
image

2. the optgroup is incorrectly indicated as being selected when the select is opened up. The optiongroup can obviously not be selected. It is simply a subcategory title. I believe what's happened here is that the sleight-of-hand being done to hide the pre-populated field value is causing the optgroup to somehow be selected by default.

image

3. No option is taking focus when the select list is opened. Such is the expected behaviour, and the failure is again IMO stemming from the decision to populate with a needless 'placeholder' instruction. I have made a small movie to demonstrate how this _should_ be happening

https://user-images.githubusercontent.com/14298245/214163612-4cdc2531-572d-4fe7-b819-e17102c98dbf.mov

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?globals=outline:true&id=components-select--default&viewMode=story

### Steps to reproduce

1. Got to the [storybook for Carbon](https://react.carbondesignsystem.com/iframe.html?globals=outline:true&id=components-select--default&viewMode=story)
2. Select Select from the left nav
3. Open canvas in its own window, by choosing icon in upper right corner 
4. Operate the control by keyboard alone.

### Code of Conduct

francinelucca:
Closes #13008 

#### Changelog

**New**

- Link select stories to mdx docs
- Add guidance in docs to default options

**Changed**

- Remove default "Choose an option" option
- Add empty default option
- Removed categories from stories

#### Testing / Reviewing

Check `Select` stories in deployment:
- Make sure all have an empty default option that shows up on selection when select is opened
- Make sure all of them no longer have categories
- Click on Docs tab an read through new guidance, make sure it's coherent and makes sense.


alisonjoseph (approved):

laurenmrice:
- This change would also apply to the Fluid Select which is in the Experimental section.

francinelucca:
> * This change would also apply to the Fluid Select which is in the Experimental section.

Done! @laurenmrice

laurenmrice (approved):
Looks good to me!

PR #13879, opened 2023-05-25
fix(ListBoxMenuItem): ensure border-subtle renders correct value

tw15egan:
@aagonzales mention on Slack that the Dropdown / Combobox / Multiselect `$border-subtle` values were off. It's because the menus themselves render one level higher on the `layer` scale (01) , but `border-subtle` was still calculating to its base level (00). This change scopes all `ListBoxMenuItem` styles to their appropriate value, based on the presence of `cds--layer-two` or `cds--layer-three`, and defaults to `border-subtle-01` if not. I searched for all `$border-subtle` instances and added a block for each of the two layers, and cleaned up some styles issues that were present with the change in color of these new borders. 

#### Changelog

**Changed**

- Adjusted instances of `$border-subtle` to reference the layer directly i.e `$border-subtle-01` 

#### Testing / Reviewing

Check out the `Dropdown`, `Combobox`, `Multiselect` and other `withLayer` stories that utilize `border-subtle` and ensure values remain correct. THis mainly affects the menu items that appear when the menu is open, and when hovering them. This should only affect components that use `ListBoxMenuItem`.

### DIFF
--- a/packages/react/src/components/Layer/index.js
+++ b/packages/react/src/components/Layer/index.js
@@ -11,7 +11,7 @@ import cx from 'classnames';
 import { usePrefix } from '../../internal/usePrefix';
 import { LayerContext } from './LayerContext';
 
-const levels = ['one', 'two', 'three'];
+const levels = ['one', 'two', 'three', 'four'];
 const MAX_LEVEL = levels.length - 1;
 
 /**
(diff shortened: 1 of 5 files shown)


aagonzales (approved):
Yes! That is rendering correctly now 🎉

guidari (approved):
LGTM!

tw15egan:
Looks like there is still an issue with `Dropdown` on `Modal`, just need to add another override for that use-case as well

tay1orjones (approved):
Looks good once the modal override is added 👍

Issue #13706, opened 2023-05-02
[a11y]: Tag with filter has aria-label violation


kglickman:
### Package

@carbon/react

### Package version

1.23.0

### React version

16.4

### Automated testing tool and ruleset

IBM Equal accessibility checker - latest deployment

### Description

This happens in the Carbon React story book in the playground when setting Filter to true.  On the accessibility tab, it's just a warning.  In the accessibility checker, it's an error.

![image](https://user-images.githubusercontent.com/30597411/235736111-6eda04d4-8bb6-467b-8158-63284b7d257f.png)

### Reproduction/example

Reproducible in story book

### Steps to reproduce

See storybook for Tag.  
Go to playground.  
Set Filter to true.  
Go to A11y tab
Also run Accessibility checker.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tw15egan:
It seems like we can resolve this issue by simply removing the `aria-label` on the `filter` variant. A quick removal via dev tools resolves the error and the voiceover utility seems to still read properly.

tay1orjones:
[Reported on slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1684761247699849), this is also present on v7.x

tay1orjones:
@guidari I'm going to re-open this since we need to backport the fix to v10

francinelucca:
@guidari does this issue need to be closed?

guidari:
> @guidari does this issue need to be closed?

Yes! I don't know why it didn't close after this PR merge: 
https://github.com/carbon-design-system/carbon/pull/13882

I'll close. Thank you for letting me know

guidari:
This issue was resolved in version 11 and 10 by these 2 PRs:

v11: https://github.com/carbon-design-system/carbon/pull/13852
v10: https://github.com/carbon-design-system/carbon/pull/13882

PR #13882, opened 2023-05-26
fix(Tag): fixed a11y test by removing arialabel

guidari:
Closes https://github.com/carbon-design-system/carbon/issues/13706

`carbon-v10 update`

By removing the aria-label from the div tag the violation goes away.
I tested with the voice over from MacOS and the readability still works as expected.

### Testing / Reviewing

To reproduce the error or check the fix, you can follow these steps:

- Open the Tag storybook;
- In Controls section set the filter prop to true;
- Open the component itself in a new tab;
- Then run the IBM Equal Access Accessibility Checker;
- The violation should be gone.

### DIFF
--- a/packages/react/src/components/Tag/Tag-test.js
+++ b/packages/react/src/components/Tag/Tag-test.js
@@ -28,16 +28,14 @@ describe('Tag', () => {
   });
 
   it('should have an appropriate aria-label when (filterable)', () => {
-    const children = 'tag content';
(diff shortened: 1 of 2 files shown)


francinelucca (approved):
confirmed no longer seeing violation in deployment

alisonjoseph (approved):

harismu-ibm:
@guidari can we have a new release with this fix for v10 ?

PR #13871, opened 2023-05-24
fix(UIShell): don't require aria-label in HeaderPanel

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13728

Remove requirement for `aria-label`, which was causing a11y violations when present. 

#### Changelog

**Added**
- `children` to `PropTypes` definition

**Removed**

- Removed `aria-label` from `HeaderPanel` in UI Shell story
- Removed requirement of `aria-label` or `aria-labelledby` and associated code

#### Testing / Reviewing

Run a11y checker on `UI Shell with Header and Right Panel` and ensure no violations show up

PR #13894, opened 2023-05-31
UI Shell Switcher loses focus but remains expanded

### LINKED ISSUE #10349: UI Shell switcher loses focus but remains expanded
### Package

carbon-components

### Browser

Chrome, Safari

### Operating System

MacOS

### Package version

version on carbondesignsystem.com

### React version

_No response_

n/a

### Assistive technology

keyboard

### Description

When a user opens the content switcher menu (aka Site)  the user must tab through the menu items (okay, but not optimal; should be arrow). However, since the Tab ring is not restricted to the menu, it is possible to leave the menu and navigate to the main content, leaving the menu obscuring content. (Shown in the below screen shot, where the focus is clearly on the Develop tile, which is obscured by the menu.) 

Fortunately the user can dismiss the menu with Esc at any point, or I would rate this higher priority

One of two things should occur:

1. tab is restricted to the menu (just wraps around to top item) until dismissed with Esc (or Enter on X)
1. the menu is dismissed automatically when the focus leaves the last menu item

As mentioned earlier, I would advocate that this menu have arrow key interaction, but that is not as critical as having this menu not persist after it loses focus.

Potentially linked with https://github.com/carbon-design-system/carbon/issues/9519

### WCAG 2.1 Violation

It will likely violate WCAG 2.2 Focus Appearance, since it is likely to obscure an item with focus

### CodeSandbox example

can be replicated on carbondesignsystem.com

### Steps to reproduce

1. tab to the content switcher
2. press Enter to open
3. press tab until the focus has exited the content switcher

### Code of Conduct

guidari:
Closes #10349

It was added two new functionalities to the Ui Shell Switcher.

Now the `HeaderPanel` component is dismissed when it loses focus, following the same behavior as the `SideNav` component. Also, @mbgower mentioned that it would be good to have keyboard navigation in the `Switcher`, so that feature was also implemented.

#### Changelog

**New**

- Added keyboard navigation in the `Switcher` component
- Now the `HeaderPanel` is dismissed when loses focus

#### Testing / Reviewing

Go to the Storybook "UI Shell Header Base w/ Actions and Switcher" and check if you can navigate throught the `Switcher` and also test if the `HeaderPanel` is dismissed


tw15egan:
@guidari The focus handling looks good, but I think we'll also want to toggle the `isActive` prop in the story (so it doesn't remain active when the panel closes). We can hook up some logic in the story to handle this:

```jsx
  const [isExpanded, setIsExpanded] = useState(true);

  const handleClick = () => {
    setIsExpanded(!isExpanded);
  };

return (
<Header>
....
    <HeaderGlobalAction
          aria-label="App Switcher"
          isActive={isExpanded}
          tooltipAlignment="end">
          <SwitcherIcon size={20} />
    </HeaderGlobalAction>
)
```

It seems like we'll also want to hook up some logic to control the panel opening/closing on click as well, as I was able to get to a state where I could only re-open the menu by trying to focus inside of the closed menu.

guidari:
> @guidari The focus handling looks good, but I think we'll also want to toggle the `isActive` prop in the story (so it doesn't remain active when the panel closes). We can hook up some logic in the story to handle this:
> 
> ```js
>   const [isExpanded, setIsExpanded] = useState(true);
> 
>   const handleClick = () => {
>     setIsExpanded(!isExpanded);
>   };
> 
> return (
> <Header>
> ....
>     <HeaderGlobalAction
>           aria-label="App Switcher"
>           isActive={isExpanded}
>           tooltipAlignment="end">
>           <SwitcherIcon size={20} />
>     </HeaderGlobalAction>
> )
> ```
> 
> It seems like we'll also want to hook up some logic to control the panel opening/closing on click as well, as I was able to get to a state where I could only re-open the menu by trying to focus inside of the closed menu.
> 
> 

To achieve that behavior I wrapped the `Header` inside a `HeaderContainer` and get the logic from it and also enhanced the props from `HeaderPanel`.

Now I'm just trying to find a way to initiate the `HeaderPanel` open

guidari:
To achieve that behavior I wrapped the `Header` inside a `HeaderContainer` and get the logic from it and also enhanced the props from `HeaderPanel`.

Now I'm just trying to find a way to initiate the `HeaderPanel` open

guidari:
I added a few more functionalities to the `HeaderPanel` to be more flexible when the user wants to manipulate the expanded state.

But I noticed that inside the `Switcher` component, there is a violation due to the `SwitcherDivider`.
If I remove the `SwitcherDivider`, the violation goes away. However, I didn't remove it because it works fine with VoiceOver, since I filter the `<ul>` before rendering.

The violation is as follows: `<ul> and <ol> must only directly contain <li>, <script>, or <template> elements`.

I also added `StoryContent` in the stories because I implemented the possibility for the user to skip to the main-content.

@tw15egan let me know if I that change makes sense.
Thank you!

packages/react/src/components/UIShell/SwitcherItem.js
```diff
@@ -37,9 +43,28 @@ const SwitcherItem = React.forwardRef(function SwitcherItem(props, ref) {
     [`${prefix}--switcher__item-link--selected`]: isSelected,
   });
 
+  function setTabFocus(evt) {
+    if (match(evt, keys.ArrowDown)) {
```
tw15egan, line 48: I think we'll want to add this to prevent the screen from scrolling when using the arrow keys to navigate the menu, 
same for `ArrowUp`

```suggestion
    if (match(evt, keys.ArrowDown)) {
       evt.preventDefault();
```

packages/react/src/components/UIShell/UIShell.stories.js
```diff
@@ -586,57 +586,79 @@ export const HeaderBaseWActionsAndRightPanel = () => (
...
+              <Notification size={20} />
+            </HeaderGlobalAction>
+            <HeaderGlobalAction
+              aria-label={
+                isSideNavExpanded ? 'Close switcher' : 'Open switcher'
+              }
+              isActive={isSideNavExpanded}
+              onClick={onClickSideNavExpand}
+              tooltipAlignment="end">
+              <SwitcherIcon size={20} />
```
tw15egan, line 619: Can we do something like this?

```suggestion
         {isSideNavExpanded ? (
                <Close size={20} />
              ) : (
                <SwitcherIcon size={20} />
              )}
```

tw15egan (commented):
Looking really great! Just a few comments. 

I think it would also be good for @mbgower to take a look and see if this looks good 👍🏻 

https://deploy-preview-13894--carbon-components-react.netlify.app/?path=/story/components-ui-shell--header-base-w-actions-and-switcher

packages/react/src/components/UIShell/Switcher.js
```diff
@@ -5,13 +5,15 @@
...
 
-import React from 'react';
+import React, { useRef } from 'react';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import { AriaLabelPropType } from '../../prop-types/AriaPropTypes';
 import { usePrefix } from '../../internal/usePrefix';
 
 const Switcher = React.forwardRef(function Switcher(props, ref) {
+  const switcherRef = useRef(ref);
```
francinelucca: should this use `useMergedRefs` with a separate ref and the forwardRef instead? https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComboButton/index.js#L50

packages/react/src/components/UIShell/SwitcherItem.js
```diff
@@ -68,6 +93,22 @@ SwitcherItem.propTypes = {
...
+  /**
+   * If given, switcher item will render as a link with the given href
+   */
+  index: PropTypes.number,
+
+  /**
+   * event handlers
+   */
+  onKeyDown: PropTypes.func,
+  onKeyUp: PropTypes.func,
```
francinelucca: is this prop being used?

packages/react/src/components/UIShell/SwitcherItem.js
```diff
@@ -37,9 +43,28 @@ const SwitcherItem = React.forwardRef(function SwitcherItem(props, ref) {
...
+        currentIndex: index,
+        direction: 1,
+      });
+    }
+    if (match(evt, keys.ArrowUp)) {
+      handleSwitcherItemFocus?.({
+        currentIndex: index,
+        direction: -1,
+      });
+    }
```
francinelucca: This navigation is working correctly for me. I'd confirm with @mbgower wether we'd want to also keep the tab navigation, I feel like we usually either have the arrow navigation or the tab navigation, not both

francinelucca (changes requested):
LGTM, some comments and +1 on the close icon when opened comment from TJ
image
image

packages/react/src/components/UIShell/SwitcherItem.js
guidari: Actually not! I'll remove

packages/react/src/components/UIShell/Switcher.js
guidari: We can add that, for sure! 
Idk why, but it wasn't working with `useMergedRefs` before, but now I added and it is working fine

packages/react/src/components/UIShell/UIShell.stories.js
guidari: Added those two suggestions!

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -7166,6 +7175,18 @@ Map {
...
+      },
+      "index": Object {
+        "type": "number",
+      },
+      "onKeyDown": Object {
+        "type": "func",
+      },
+      "onKeyUp": Object {
+        "type": "func",
+      },
```
francinelucca: snapshot needs to be regenerated so that the unsued func are removed here

packages/react/src/components/UIShell/SwitcherItem.js
```diff
@@ -37,9 +42,29 @@ const SwitcherItem = React.forwardRef(function SwitcherItem(props, ref) {
...
+  }
+
   return (
     <li className={className}>
       <Link
+        onKeyDown={(evt) => {
+          setTabFocus(evt);
+        }}
```
francinelucca: I think you should keep the `onKeyDown` prop here just in case someone is using that to continue the support for it. 
image

francinelucca (changes requested):

francinelucca:
If we want to go for Gower's option 1 this looks good to me. I'm wondering if we should go for option 2 though 🤔, it feels like a better user experience for me. And I think that also saves us from having to worry about the content overflow. Thoughts @mbgower @tw15egan ?

packages/react/src/components/UIShell/HeaderPanel.js
```diff
@@ -27,14 +31,36 @@ const HeaderPanel = React.forwardRef(function HeaderPanel(
...
+  const eventHandlers = {};
+
+  if (addFocusListeners) {
+    eventHandlers.onKeyDown = (event) => {
+      if (match(event, keys.Escape)) {
+        setExpandedState(false);
+        onHeaderPanelFocus();
+        if (href) {
+          window.location.href = href;
+        }
```
francinelucca, line 64: now that i'm thinking about this I don't think it makes sense for the HeaderPanel to redirect to the main content when the headerPanel closes. I think this is an interaction that makes sense for the side rail but not necessarily here 🤔

guidari: I agree. Now that the menu can't be open on focus, this prop is not necessary

mbgower:
I think this is what @jeanservaas  was most in favour of:
 > close when clicked outside the sidenav / switcher
close when focus is lost
do not open when focus moves in, menu should only open with explicit action (enter, click)

One other thing. Because the switcher is only opened on activation, I do not think you want to have the shift+tab functionality the way you had it -- where on focus it opens the side panel and ascends from the bottom item.
The behaviour now, regardless of how the user dismissed the side panel -- whether by Esc or tabbing out -- is the switcher button is collapsed in the tab order, and opening it is 'fresh' (no preservation of last item before it was escaped). Does that make sense?

tw15egan:
>I also think -- and this is again something Jeannie should decide -- that the switcher changing to an X is superfluous. No one needs the 'X' to dismiss the opened menu/sidebar. It could just stay a switcher icon, which the user could always toggle closed again (if they don't just click away). When there was a need for the keyboard user to navigate to it to dismiss this side bar, the X made a bit more sense. Since the keyboard user never needs to activate to dismiss now, it seems distracting/inaccurate.

This is simply built into the story, not the component, and is just to bring parity to what is seen on the website to reduce questions about why they are different. We can revisit this as a whole with Jeannie when we do the UI Shell overhaul.

tw15egan:
@guidari would this just be scoped to `Switcher`?  Not sure we'd want it to always auto focus to first item in cases like this example: https://react.carbondesignsystem.com/?path=/story/components-ui-shell--header-base-w-actions-and-right-panel 

We don't always know what will be in `HeaderPanel`

guidari:
> @guidari would this just be scoped to Switcher? Not sure we'd want it to always auto focus to first item in cases like this example: https://react.carbondesignsystem.com/?path=/story/components-ui-shell--header-base-w-actions-and-right-panel

Would be only in the `Switcher`, because it is a `<ul>` tag.
I agree, we have let it open to the user to do what it wants in cases like that.

tw15egan:
@guidari great, thanks for clarifying! 😄

packages/react/src/components/UIShell/HeaderPanel.js
```diff
@@ -5,24 +5,66 @@
...
-import React from 'react';
+import React, { useRef, useState } from 'react';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import { usePrefix } from '../../internal/usePrefix';
+import { keys, match } from '../../internal/keyboard';
 
 const HeaderPanel = React.forwardRef(function HeaderPanel(
-  { children, className: customClassName, expanded, ...other },
+  {
```
tw15egan, line 17: I believe these got accidentally added back in, they were removed as part of https://github.com/carbon-design-system/carbon/pull/13871

packages/react/src/components/UIShell/HeaderPanel.js
```diff
@@ -5,24 +5,66 @@
...
+    className: customClassName,
+    expanded,
+    addFocusListeners = true,
+    onHeaderPanelFocus,
+    ...other
+  },
   ref
 ) {
   const prefix = usePrefix();
+  const accessibilityLabel = {
```
tw15egan: Same as above

packages/react/src/components/UIShell/HeaderPanel.js
guidari: Yes! I'm already fixing this merge conflict, thanks

guidari:
I made a list of the changes that were made in the `Switcher` from the beginning, just to make it easy to see what happened.

- The X icon was removed and kept only the Switcher icon
- The user can dismiss the `HeaderPanel` by clicking outside of it only if the the children component is the `Switcher`
- Close when focus is lost
- The `HeaderPanel` is dismissed when the user tabs out
- The `HeaderPanel` can be dismissed by pressing ESC
- The user can navigate using arrow keys or tab. With tab navigation you can close the `HeaderPanel` if you left it, but with arrow key the navigation is wrapped.
- The `HeaderPanel` do not open when focus moves in, it only open with explicit action (enter, click)

mbgower:
> I made a list of the changes that were made in the Switcher from the beginning, just to make it easy to see what happened.

Thanks for that @guidari. Has design had a chance to comment/approve those interactions and any visual changes?

> The X icon was removed and kept only the Switcher icon

This is the thing I'm most interested in ensuring gets approval by design, and since I don't see a preview there are a few questions I had about it.
- Do you have a visual change for the switcher icon when pressed? (i.e., its colors invert or it gets some treatment other than what it has for focus, etc)
- Are you moving the focus into the list on activation or leaving it on the switcher button?
- what role and pattern are you following? I've listed two basic patterns I've seen for this below. Both are options (and neither is prescriptive). There's a trend towards the disclosure pattern, which I think is closer to what you're doing at the moment. 

https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-links/
https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/

tw15egan:
@mbgower The preview can be viewed here: https://deploy-preview-13894--carbon-components-react.netlify.app/?path=/story/components-ui-shell--header-base-w-actions-and-switcher


The `x` interaction was just added into the story as a test to match the parity of what is seen on [the website](https://carbondesignsystem.com/) but isn't actually baked into the component. This would have to be configured by the end user, so we don't need to make any adjustments here, it was just to align the UI Shell story and what is seen on the website

francinelucca (changes requested):
This looks pretty good to me, just wondering if we could put the focus on the switcher trigger button on Escape? focus seems to get lost after close on escape

https://github.com/carbon-design-system/carbon/assets/40550942/cca17450-4003-4b6c-8f83-86e87dc4cb03

guidari:
> Thanks for that @guidari. Has design had a chance to comment/approve those interactions and any visual changes?

It was assigned a design review for it

> Do you have a visual change for the switcher icon when pressed? (i.e., its colors invert or it gets some treatment other than what it has for focus, etc)

Yes! The background color changes to be equals to the `HeaderPanel`, which indicates that it is open

> Are you moving the focus into the list on activation or leaving it on the switcher button?

We leave the focus in the switcher button

> what role and pattern are you following? I've listed two basic patterns I've seen for this below. Both are options (and neither is prescriptive). There's a trend towards the disclosure pattern, which I think is closer to what you're doing at the moment.

I'll check those links and see if there is anything that is missing and should be added. Thanks.

guidari:
> This looks pretty good to me, just wondering if we could put the focus on the switcher trigger button on Escape? focus seems to get lost after close on escape

Added the href prop so the user can choose where the focus should go. In the Storybook example, the focus is being redirected to the switcher button.
@francinelucca @mbgower

francinelucca (approved):
LGTM pending approval from design!

packages/react/src/components/UIShell/SwitcherItem.js
```diff
@@ -68,6 +96,21 @@ SwitcherItem.propTypes = {
    */
   className: PropTypes.string,
 
+  /**
+   * event handlers
+   */
+  handleSwitcherItemFocus: PropTypes.func,
+
+  /**
+   * If given, switcher item will render as a link with the given href
```
andreancardona:
```suggestion
   * If provided, switcher item will render as a Link with the provided href
```

packages/react/src/components/UIShell/SwitcherItem.js
```diff
@@ -68,6 +96,21 @@ SwitcherItem.propTypes = {
...
 
+  /**
+   * event handlers
+   */
+  handleSwitcherItemFocus: PropTypes.func,
+
+  /**
+   * If provided, switcher item will render as a Link with the provided href
+   */
+  index: PropTypes.number,
```
andreancardona, line 107: @guidari I'm a little confused on this prop description - is this the correct description for `index` - otherwise looks good!

laurenmrice (approved):

tw15egan (approved):
LGTM, nice work Gui! 👍 ✅

packages/react/src/components/UIShell/HeaderPanel.js
```diff
@@ -5,30 +5,92 @@
...
+  },
   ref
 ) {
   const prefix = usePrefix();
+  const headerPanelReference = useRef(null);
+  const headerPanelRef = useMergedRefs([headerPanelReference, ref]);
+
+  const controlled = useRef(expanded !== undefined).current;
+  const [expandedState, setExpandedState] = useState(expanded);
+  const expandedProp = controlled ? expanded : expandedState;
```
andreancardona, line 34: testing

packages/react/src/components/UIShell/SwitcherItem.js
guidari: Good catch! I'll fix it!

andreancardona (approved):

guidari:
> Does https://carbondesignsystem.com/ implement/inherit this issue/pr?

Hey @philljenkins it does!

tw15egan:
@guidari just a heads up, the website won't receive this until the gatsby theme is updated. I think the storybook is the best place to test the additions you made 👍🏻

PR #13919, opened 2023-06-02
fix(Tabs): do not dismiss disabled tabs

### LINKED ISSUE #13897: [Bug]: Dismissable Tabs - Disabled state on close icon
### Package

@carbon/react

### Browser

Chrome

### Package version

1.30.0

### React version

18.2.0

### Description

For dismissable tabs, when disabled the hover state on the close icon should be a disabled cursor, it is currently a pointer.

The storybook behavior should also be modified so the disabled tab does not close when the close icon is clicked (we should continue to forward the `onClick` event for said tab to allow for others to define this behavior if desired).

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tabs--dismissable

### Steps to reproduce

- Hover over close icon on disabled tab and notice cursor is a pointer instead of disabled.
- Click on close icon on disabled tab and notice tab gets closed, expected tab to not close.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

francinelucca:
Closes #13897 

#### Changelog

**Changed**

- Check for disabled tab before dismissing in story and docs

**Removed**

- `cursor: pointer` rule for close icon on dismissable tabs (non-disabled tabs inherit this from the tab and disabled tabs don't need it)

#### Testing / Reviewing

https://deploy-preview-13919--carbon-components-react.netlify.app/?path=/story/components-tabs--dismissable
https://deploy-preview-13919--carbon-components-react.netlify.app/?path=/story/components-tabs--dismissable-with-icons
Test dismissable tab story and make sure hovering over disabled tab close icon shows a disabled cursor and that when on close icon click the disabled tabs do not get dismissed. Everything else should continue to work as before including dismissing active tabs


guidari (approved):
LGTM! 

I just have a comment about something that I've been noticing in a few components.
When entering components with arrow key navigation, wouldn't it be good for us to indicate that the component has that kind of navigation?

Because I was testing the `Tabs` with VoiceOver, and I'm not sure if this kind of navigation is obvious for all users with poor vision (it probably is, but it wasn't obvious to me, haha).

The example would be something like this:

`<TabList aria-label="List of tabs with arrow navigation">`

tw15egan (approved):
LGTM 👍 ✅

Issue #13923, opened 2023-06-05
Use contextual layout tokens for all remaining components


janhassel:
This issue is to track support of contextual layout tokens in `@carbon/react` and `@carbon/web-components` for components not directly addressed in https://github.com/carbon-design-system/carbon/issues/6202

PR #13924, opened 2023-06-05
feat(progress-bar): support contextual layout tokens (density)

janhassel:
Ref #13923 

Adds support for contextual layout tokens to `ProgressBar`.

#### Changelog

**Changed**

- Updated ProgressBar styles to support `density` group
- Changed a few spacings to use logical properties

#### Testing / Reviewing

Locally run storybook
- `type="inline"`
  - The spacing between the label and the bar or icon adapts to `layout.density`
- `type="indented"`
  - The inline padding of the label and helper text adapt to `layout.density`

**Open question**: the ProgressBar supports two sizes: "big" and "small" which deviate both in naming and in values from the typical t-shirt sizes used in other components. Should the ProgressBar still participate in layout sizing but use its own overrides similar to `Tag` or should it be ignored here? For reference, other components using a `small` keyword are `Loading`, `SideNavIcon` and some skeletons.


alisonjoseph (approved):
This LGTM! Will bring the naming of the size props to the team to discuss.

francinelucca (approved):

Issue #13351, opened 2023-03-15
Tabs: Aligning to the Grid design guidance


sstrubberg:
### Acceptance criteria

- Finalize design spec and guidance.
  - [x] Explore left UI shell implications (non-grid influencers and grid influencers)
  - [x] Define main design guidance rules (Responsive behavior / Truncation / etc.)

laurenmrice:
## Design specs and guidance

**Implications of UI shell**
- Decision to have the same design guidance for UI shell grid influencer and non-grid influencer situations.

**Breakpoints**
- Max, XLG, LG: Use 8 grid-aware tabs or less.
- MD, SM: Grid-aware tabs will be automatically changed to auto-width tabs.

**Responsive behavior**
- Tabs should not wrap to multiple lines and stack.
- Include truncation of tab labels with ellipsis (...).
- Horizontal scroll containers when needed.

-----

<img width="1181" alt="Screen Shot 2023-04-07 at 12 47 04 PM" src="https://user-images.githubusercontent.com/43969356/230646162-4a340464-21a6-448d-9f6e-0ff3789a2f23.png">



<img width="1121" alt="Screen Shot 2023-04-07 at 12 46 30 PM" src="https://user-images.githubusercontent.com/43969356/230646074-28856df5-278e-4b78-8ec5-5020233c7588.png">

PR #13927, opened 2023-06-05
feat(Tabs): Contained tabs on the grid

### LINKED ISSUE #13353: Tabs: Implement contained tabs on the css grid
Reference design issue #13351

**Acceptance criteria**
- [ ] Add tests
- [ ] Add documentation - This will only work with CSS grid, no flexbox

francinelucca:
Closes #13353 

Adds `fullWidth` prop to `TabList` so that Tabs (container) can be aligned to the Grid

#### Changelog

**New**

- `fullWidth` prop now available for contained `TabList`
- 'contained fullWidth' tab story and e2e test
- tests for `fullWidth` prop and added mock for `useMatchMedia` in Tab testing
- 'cds-tabs--full-width' class for `fullWidth` contained Tabs and accompanying style
- Grid guidance using `fullWidth` prop  for Tabs

#### Testing / Reviewing

1. Check that stories match spec: https://github.com/carbon-design-system/carbon/issues/13351#issuecomment-1500455606
     - https://deploy-preview-13927--carbon-components-react.netlify.app/?path=/story/components-tabs--contained-full-width 
     - https://deploy-preview-13927--carbon-components-react.netlify.app/iframe.html?args=&id=components-tabs--contained-grid-test&viewMode=story 
3. Check developer guidance is true and coherent: https://deploy-preview-13927--carbon-components-react.netlify.app/?path=/docs/components-tabs--contained-full-width#tabs-and-the-grid---fullwidth-prop
4. Should have no visual/behavioral regression to default tab styles and functionality outside of `fullWidth`

TODO:

- [x] Remove Test story


francinelucca:
We're still ironing some dev things out but this should be good for design review @kingtraceyj . You can check it out at this url: https://deploy-preview-13927--carbon-components-react.netlify.app/iframe.html?args=&id=components-tabs--contained-grid-test&viewMode=story

kingtraceyj (approved):
this is working as intended! 💥

alisonjoseph:
Question: are we not aligning individual tabs on the grid anymore? That was the original intent of the design spec, but maybe it changed? 

Screenshot 2023-06-15 at 10 26 54 AM

packages/react/src/components/Tabs/Tabs.mdx
```diff
@@ -201,6 +207,51 @@ loaded when the Tab is clicked. In v11, to do this, you can this by setting
 </Tabs>
 ```
 
+### Tabs and the Grid - fullWidth prop
+
+By default, a `Tab` component is only as wide as it's content. This posses difficulties when trying to align tabs to the grid.
+Alternatively, you may choose to use the `fullWidth` prop to allow Tab elements to grow as big wide their container allows.
+Note that this feature is *only available* for `contained` tabs with less than 9 tabs, otherwise the prop will be ignored.
+`fullWidth` paired up with a wrapping `Grid` component will allow for "grid-aware" tabs:
```
alisonjoseph, line 217: I think we could add some more docs here no how to get the tabs to align on the grid exactly using this. 

> In order to have the tabs align on grid columns you will want to be sure to make the number of columns match the number of tabs. For example if you have 5 tabs you would want to use. `<Column lg={5} md={5}>`

alisonjoseph (changes requested):
This is looking great! I think we can make the docs a little clearer by adding / showing an example on how to actually get the tabs on the grid by matching the column numbers. We might want to show that in the code sample on the docs also.

francinelucca:
@alisonjoseph yeah we confirmed with Tracey & Lauren only the Container needs to be aligned to the Grid and if the number of columns coincide with the number of Tabs then the individual tab items will also be aligned to the Grid. We just added some extra guidance around that in the Docs, let us know what you think :)

alisonjoseph (approved):
Awesome! LGTM, and a lot easier to understand now with added docs 🥳

tay1orjones (approved):

PR #13937, opened 2023-06-07
feat(tile): support contextual layout tokens (density)

janhassel:
Ref #13923 

Adds support for contextual layout tokens to `Tile`, `ClickableTile`, `ExpandableTile`, `SelectableTile`, `RadioTile`.

#### Changelog

**Changed**

- Updated tile styles to support `density` group
- Changed a few spacings to use logical properties

#### Testing / Reviewing

Locally run storybook
- inner padding on all tiles should adapt to `layout.density`
- position of checkmark on `SelectableTile` and `RadioTile` should adapt to `layout.density`
- position of arrow button on `ExpandableTile` (with and without interactive content) should adapt to `layout.density`
- position of icon in `ClickableTile` should adapt to `layout.density`
- All of the above should work for both the current stable tiles as well as the experimental tiles


packages/styles/scss/components/tile/_tile.scss
```diff
@@ -175,8 +177,8 @@
     right: 0;
     bottom: 0;
     display: flex;
-    width: $spacing-09;
-    height: $spacing-09;
+    width: calc(#{layout.density('padding-inline')} * 2 + 1rem);
+    height: calc(#{layout.density('padding-inline')} * 2 + 1rem);
```
andreancardona: was trying to think if there is a better way to share this formula - it's seems a bit confusing and we're using it a lot - but I couldn't think of a better way 🤷🏽‍♀️

andreancardona (approved):

packages/styles/scss/components/tile/_tile.scss
janhassel: Good point. I just pushed an update where I made a local variable (`$-icon-container-size`) out of this calculation and commented on the logic behind it. Do you think this makes sense?

alisonjoseph (approved):

PR #13951, opened 2023-06-07
feat(tabs): support contextual layout tokens (size & density)

janhassel:
Ref #13923 

Adds support for contextual layout tokens to `Tabs`.

#### Changelog

**Changed**

- Updated tabs styles to support `size` and `density` groups

#### Testing / Reviewing

Locally run storybook
- Line tabs
  - should support `layout.size` with sm, md, lg (default: md)
  - padding-inline should adapt to `layout.density`
- Icon only tabs
  - should support `layout.size` with sm, md, lg (default: md)
  - should be square
- Contained tabs
  - should support `layout.size` with sm, md, lg, xl (default: lg)
  - padding-inline should adapt to `layout.density`
  - spacing between label and icon should adapt to `layout.density`
- Tabs with two lines
  - should support `layout.size` with lg, xl (default: xl)
- Tab content
  - padding should adapt to `layout.density`
- `<Tab iconSize="lg">`
  - should result in the tabs being locked to `lg`
- Tab horizontal overflow buttons should share same size as tabs


janhassel:
@aagonzales I want to tag you here as well as there are some differences percy reported. It looks like the labels in line tabs are not 100% centered in main right now due to a difference in padding: 
image
The visual specs on carbondesignsystem.com show that equal padding is desired though: 
image

Would love it if you could have a look at the implementation of this PR and see whether this achieves the desired visuals.

aagonzales:
The Percey screenshots aren't showing but when I manually compare it to the spec this PR still seems slightly off (1px maybe). The foreground is a Figma asset, the background is a code asset. 

image


I also think the 8px top and bottom isn't true. Not sure where that number came from looking at the assets. They should just be vertically centered.

janhassel:
@aagonzales I pushed an update positioning the label 1px lower. Checking with the latest Figma library I see an exact overlay now. Let me know if this works!

aagonzales:
@janhassel yup that's it! Matches up perfectly now. 👍

francinelucca (approved):

tw15egan (approved):
Sorry, missed this review! LGTM once merge conflict is resolved 👍🏻 ✅

janhassel:
@tw15egan I just resolved the merge conflict. I think the percy approval is the last failing check now.

PR #13952, opened 2023-06-07
fix(ExpandableSearch): fix defaultValue issues, add expanded prop

### LINKED ISSUE #13782: [Bug]: ExpandableSearch doesn't honor defaultValue
### Package

@carbon/react

### Browser

_No response_

### Package version

1.23.0

### React version

16.4

### Description

I'm specifying the defaultValue attribute for ExpandableSearch but when I click on the magnifying glass to open it, the defaultValue is not set.

This is a serious problem because the component doesn't really work as a controlled component with value.  It is too slow to handle fast typing, even with a debounce.

### Reproduction/example

https://stackblitz.com/edit/github-a6ym8m?file=src%2FApp.jsx

### Steps to reproduce

Click on the magnifying glass in the Stackblitz.  The search expands but does not show the defaultValue.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13782

Fixes an issue that was causing the `x` close button on `Search` to appear when a `defaultValue` was present in a non-expanded `ExpandableSearch`.

This PR also adds in the `isExpanded` prop, which will allow users to control the state of the `ExpandableSearch` themselves. 

#### Changelog

**New**

- `isExpanded` prop is added to `Search` so that the `ExpandableSearch` expansion state can be controlled. 
- Tests for new props

**Changed**

- Fixed an issue that was causing the `x` to overlay the `ExpandableSearch` expansion icon, which was causing the Search to clear on open, seeming like no value was being set. 

#### Testing / Reviewing

Go To `Search --> Test` story and ensure all 4 variants render as expected, and there are no regressions


packages/react/src/components/Search/Search.tsx
```diff
@@ -55,6 +55,11 @@ export interface SearchProps extends InputPropsBase {
    */
   disabled?: boolean;
 
+  /**
+   * Specify whether or not ExpandableSearch should render expanded or not
+   */
+  isExpanded?: boolean;
+
```
francinelucca, line 62: I'd rename this prop to reflect more what it's actually doing within the scope of the `Search` component, like `HideCloseButton`, since `Search` doesn't handle expanded capabilities I don't think it should have any knowledge of it either. And then you can set `HideCloseButton={!isExpanded}` from the `ExpandableSearch` component

francinelucca (approved):
LGTM! just one refactor suggestion

packages/react/src/components/Search/Search.tsx
tw15egan: It's just because all of `ExpandableSearch` props are pulled from `Search:SearchProps`. If I change it, it errors in `ExpandableSearch` and that prop name doesn't make sense in that context.

tay1orjones (approved):

PR #13958, opened 2023-06-08
feat(toggle): fixes disabled issue and adds test

### LINKED ISSUE #13956: [Bug]: `Toggle` still toggles while `disabled`
### Package

@carbon/react

### Browser

Chrome

### Package version

latest

### React version

_No response_

### Description

When using Toggle inside the playground, you can still toggle the input while being disabled

https://github.com/carbon-design-system/carbon/assets/12755042/2c404208-2790-49e7-87f2-2358bc4454e1

https://react.carbondesignsystem.com/?path=/story/components-toggle--playground&args=disabled:true

### Reproduction/example

Click storybook link

### Steps to reproduce

1. open toggle [storybook](https://react.carbondesignsystem.com/?path=/story/components-toggle--playground&args=disabled:true)
2. Toggle away
3. Still toggles while disabled

### Suggested Severity

### Application/PAL

Instana

### Code of Conduct

AlexanderMelox:
### Closes #13956 

**Before**
https://github.com/carbon-design-system/carbon/assets/12755042/071a7579-fc21-4b48-b101-2ef6382667a6

**After**

https://github.com/carbon-design-system/carbon/assets/12755042/e4438e1e-5109-413d-bc64-2af52d3a0e21

This PR fixes the issue where you are able to still toggle when the Toggle is disabled.

#### Changelog

**Changed**

- packages/react/src/components/Toggle/Toggle.js
- packages/react/src/components/Toggle/Toggle-test.js

#### Testing / Reviewing

Checked on storybook if you are able to toggle while disabled. Added test to make sure you are not able to toggle when disabled.


packages/react/src/components/Toggle/Toggle.js
```diff
@@ -37,7 +38,7 @@ export function Toggle({
   });
 
   function handleClick(e) {
-    if (!readOnly) {
+    if (!readOnly && !disabled) {
```
tw15egan: We probably don't want to fire the `onClick` when disabled, either, since it wouldn't match keyboard parity (can't be focused).

Should we just add `!disabled` to `L87`?

```jsx
if (buttonElement.current && e.target !== buttonElement.current && !disabled) {
```

tw15egan:
I'm also seeing some focus styles being applied when clicking on the `disabled` `Toggle`.

I was able to remove those styles locally by changing the selector to this on `L77` in `_toggle.scss`

```scss
  .#{$prefix}--toggle__button:not(:disabled):focus
    + .#{$prefix}--toggle__label
    .#{$prefix}--toggle__switch,
  .#{$prefix}--toggle__button:not(:disabled):active
    + .#{$prefix}--toggle__label
    .#{$prefix}--toggle__switch,
  .#{$prefix}--toggle:not(.#{$prefix}--toggle--disabled):active
    .#{$prefix}--toggle__switch {
```

Seems like the `HCM` fix at the bottom of the file uses the same selector, so should update that block as well.

packages/react/src/components/Toggle/Toggle.js
AlexanderMelox: So I tried adding it there but it's still showing the styles, I wonder if it has to do with CSS :disabled state

tw15egan: Yeah I think we need to update the selector like in my other comment, this will just prevent firing the `onClick`

AlexanderMelox: Got it, I'm figuring out the selectors now

AlexanderMelox:
> I'm also seeing some focus styles being applied when clicking on the `disabled` `Toggle`.
> 
> I was able to remove those styles locally by changing the selector to this on `L77` in `_toggle.scss`
> 
> ```css-scss
>   .#{$prefix}--toggle__button:not(:disabled):focus
>     + .#{$prefix}--toggle__label
>     .#{$prefix}--toggle__switch,
>   .#{$prefix}--toggle__button:not(:disabled):active
>     + .#{$prefix}--toggle__label
>     .#{$prefix}--toggle__switch,
>   .#{$prefix}--toggle:not(.#{$prefix}--toggle--disabled):active
>     .#{$prefix}--toggle__switch {
> ```
> 
> Seems like the `HCM` fix at the bottom of the file uses the same selector, so should update that block as well.

I went ahead and added the changes. I removed the CSS that shows focus and active states when the toggle is disabled. In terms of keyboard navigation should the toggle still have focus even thought it's disabled?

AlexanderMelox:
@tw15egan It's just focusing the selector tab at the bottom! So the disabled Toggle is not in the Tab Order :) 

https://github.com/carbon-design-system/carbon/assets/12755042/1160c4a0-9dd4-4d6f-8de0-403adca9e3d7

francinelucca (changes requested):
LGTM but seems like 1 tab stop is getting lost, I have to tab twice from the top panel to get to the bottom and it's not clear where the focus is on that "ghost" tab stop 

https://github.com/carbon-design-system/carbon/assets/40550942/51fe0d30-a8d0-4b58-89e2-86b75d936e98

tw15egan:
Yeah I also tried to capture it, its getting lost somewhere

AlexanderMelox:
I'm trying to see where the focus can go, we did disabled this so idk if this is enough. 

Screenshot 2023-06-09 at 10 52 11 AM

AlexanderMelox:
Screenshot 2023-06-09 at 10 54 47 AM
daa3536">

So i did some debugging and it looks like focus is still being applied while disabled, we just don't see the styles

tw15egan:
@AlexanderMelox weird, even setting the `tabIndex` on the `div` (and/or `button`) it still stays in the tab order 🤔

tw15egan (approved):

francinelucca (approved):

AlexanderMelox:
The tests are oddly failing

PR #13961, opened 2023-06-09
fix(UIShell): fix hover issue, reasonable default for enter/exit delay

### LINKED ISSUE #13891: [Bug]: `HeaderGlobalAction` tooltip is hidden by `HeaderPanel` and other tooltips
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.30.0

### React version

v18.2.0

### Description

We have a component in our project that implements the [UIShell right panel pattern](https://carbondesignsystem.com/components/UI-shell-right-panel/usage) with multiple right panels and discovered some issues with the tooltip of the `HeaderGloablAction` component used for the panel triggers. 

1. The tooltip is completely hidden by any `HeaderPanel` that is opened due to a lower z-index of the tooltip. It can be worked around for instance by setting `.cds--popover * { z-index: 9000; }` to place the tooltip on top.
2. The tooltip is partially hidden when hovering the corresponding `HeaderGloablAction` button and then moving to the neighbouring one. This is due to the tooltip's leave delay being quite long by default. We can set it manually, but in this case it would imo make sense to set the default to e.g. `100ms`. This can be worked around by setting `leaveDelayMs={100}` on the `HeaderGlobalAction`.

### Reproduction/example

https://stackblitz.com/edit/carbon-v11-z8fomg?file=src%2FApp.jsx

### Steps to reproduce

Side panels covering the tooltip:
1. Click on help icon in header.
2. hover over icons in header.

Tooltips covering each other:
1. Hover over one of the icon buttons inside the header.
2. Move mouse over to the other icon button while first tooltip is still displayed.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13891

#### Changelog

**New**

-  Adds in a reasonable default for the `enter` and `exit` delay on `IconButton` so they do not overlap each other

**Changed**

- Adjusted `popover` value in `Header` to be placed above the `Header` itself, so it appears when right panel is active

#### Testing / Reviewing

Ensure the UI Shell action button tooltips do not overlay each other when hovering from one to the next. 
Ensure the UI Shell action button tooltips appear _over_ the right panel / switcher


guidari (approved):
Awesome! Now the user is able to see the tooltip when hovering with the `HeaderPanel` opened

https://deploy-preview-13961--carbon-components-react.netlify.app/?path=/story/components-ui-shell--header-base-w-actions-and-switcher

andreancardona (changes requested):
we have some tests failing :/ other than that code looks good!

tw15egan:
@andreancardona seems like its just a Percy timeout

andreancardona (approved):
looks good!

Issue #13562, opened 2023-04-17
Add TypeScript types to Loading, InlineLoading


tay1orjones:

alissa9090:
I'll take it

Issue #13229, opened 2023-02-24
Add TypeScript types to Button, ButtonSkeleton, DangerButton, ButtonSet


tw15egan:

tay1orjones:
@DavidChristiansen You’ll need to add a `// @ts-ignore` above the line where you import the component

lewandom:
@tw15egan @tay1orjones I'd like to take this one next on my plate.

lewandom:
Waiting on https://github.com/carbon-design-system/carbon/pull/13735 before pushing a PR for this one.

PR #13992, opened 2023-06-13
chore: Types for FileUploader, FileUploaderButton, FileUploaderDropContainer, FileUploaderItem, FileUploaderSkeleton, Filename, Upload and ButtonSkeleton

### LINKED ISSUE #12545: Add TypeScript types to FileUploader, FileUploaderButton, FileUploaderDropContainer, FileUploaderItem, FileUploaderSkeleton, Filename

imp-dance:
Closes #12545 
Part of https://github.com/carbon-design-system/carbon/issues/13229
Part of https://github.com/carbon-design-system/carbon/issues/13562

#### Changelog

**New**

- Added types for the following components, and renamed to `.tsx`:
   - FileUploader
   - FileUploaderButton
   - FileUploaderDropContainer
   - FileUploaderItem
   - FileUploaderSkeleton
   - Filename
   - Upload
   - ButtonSkeleton 

**Changed**
  - `Filename.tabIndex` changed from `string` -> `number` in type and PropType.

**Additional notes**
	* Also added types to `Loading` and `ButtonSkeleton` components. (outside issue scope, but required for issue)
	* We should probably share a `ButtonKind` type. Could be derived from the constant in `prop-types/types.js`.


tw15egan:
@imp-dance looks like we have some test failures

imp-dance:
@tw15egan Yeah I saw, seems related to this change:

> * `Filename.tabIndex` changed from `string` -> `number` in type and PropType.

What's the appropriate way to do/propose a breaking change like this?

tw15egan:
@imp-dance if possible, let's keep it a string so it has less of an effect downstream. We could also add `number` as a valid propType along with string, if you think that would help

imp-dance:
@tw15egan Sure thing, just pushed a commit that adds both as string & number as valid types, refactored the code to properly parse them and updated the snapshot.

packages/react/src/components/Button/Button.Skeleton.tsx
```diff
@@ -10,13 +10,37 @@ import React from 'react';
...
+   * Optionally specify an href for your Button to become an `<a>` element
+   */
+  href?: string;
+
+  /**
+   * Specify the size of the button, from a list of available sizes.
+   * For `default` buttons, this prop can remain unspecified or use `default`.
+   * In the next major release of Carbon, `default`, `field`, and `small` will be removed
+   */
+  size?: 'sm' | 'md' | 'field' | 'lg' | 'xl' | '2xl';
```
tay1orjones: `field` is no longer a valid value for this prop. We must've missed removing it's usage as part of the v11 release. The usage of it for classname can be removed which should allow it to no longer be an allowed value here.

packages/react/src/components/FileUploader/Filename.tsx
```diff
@@ -31,10 +67,15 @@ function Filename({
...
             className={`${prefix}--file-close`}
             type="button"
-            {...rest}>
+            {...rest}
+            tabIndex={
+              rest.tabIndex !== undefined
+                ? parseInt(rest.tabIndex as string, 10)
+                : undefined
+            }
+            aria-describedby={invalid ? ariaDescribedBy : undefined}>
```
tay1orjones, line 78: This needs to go above the `...rest` like it was before. Modifying this to be after the `...rest` makes it so that consumers can't modify `aria-describedby`

packages/react/src/components/FileUploader/FileUploaderButton.tsx
```diff
@@ -13,9 +13,98 @@ import { ButtonKinds } from '../../prop-types/types';
...
+  /**
+   * Specify the size of the FileUploaderButton, from a list of available
+   * sizes.
+   */
+  size?: 'sm' | 'small' | 'field' | 'md' | 'lg';
+
+  /**
+   * @deprecated The `tabIndex` prop for `FileUploaderButton` has been deprecated since it now renders a button element by default.
+   */
+  tabIndex?: number | string;
```
tay1orjones, line 103: Could you share more of why this was updated to accept a string? It should only be a number IIRC

packages/react/src/components/FileUploader/FileUploaderButton.tsx
```diff
@@ -101,9 +197,9 @@ function FileUploaderButton({
         id={inputId}
         disabled={disabled}
         type="file"
-        tabIndex="-1"
+        tabIndex={-1}
         multiple={multiple}
-        accept={accept}
+        accept={accept?.toString()}
```
tay1orjones, line 202: I think this change is okay, but I do want to call out this change because I wonder if it's possible this could break consumers. `accept` is intended to be a comma-separated string, which is how `toString()` outputs an array of strings.

tay1orjones (changes requested):
This looks pretty solid, thanks! I found just a couple minor items to address, let me know what you think

packages/react/src/components/FileUploader/FileUploaderButton.tsx
imp-dance: You can read about that here: https://github.com/carbon-design-system/carbon/pull/13992#issuecomment-1589923529 - but essentially it used to be a string in the proptypes (you can see that in the snapshot diff). I guess it has worked previously just because of how HTML stringifies certain props if given as different types.  

I agree that it should really only be a number. Typescript should yell at them a few places if they've used strings previously, and then it should be a pretty easy refactor to fix in that case. 

But that's not a decicion for me to make, so I guess it's up to you guys to decide how you want to handle this change @tay1orjones & @tw15egan. Let me know where you land and I'll update the PR to match it! 🚀

packages/react/src/components/FileUploader/FileUploaderButton.tsx
imp-dance: Yes my thoughts exactly. 

I was looking at the code initially thinking "_how could this have worked previously_" (passing an array to a prop that expects a string) - but then tried to do `["foo", "bar"].toString()` in my console and got `"foo, bar"` - and so I think it's just been stringified my HTML previously. This change really just makes this behaviour more explicit.

packages/react/src/components/FileUploader/Filename.tsx
imp-dance: Sure thing. I just want to note here incase you missed it - this prop used to be called `aria-describedBy` previously, but now is changed to match the actual HTML attribute `aria-describedby` (no caps). 

\- `aria-describedBy`
\+ `aria-describedby`

Moving this above `...rest` will not really do anything though, since `rest` does not contain `aria-describedby`. In either case, the user can't overwrite the conditional behaviour, it seems to me.

packages/react/src/components/FileUploader/FileUploaderButton.tsx
tay1orjones: Ah sorry I missed that, GitHub auto-collapsed that part of the conversation and I didn't look into it. I think it's fine as is for now

packages/react/src/components/FileUploader/Filename.tsx
tay1orjones: Ah! Yes I missed that. Sorry my mistake 🤦‍♂️

tay1orjones (approved):

tay1orjones:
@tw15egan This is ready for a re-review

@imp-dance There's one minor conflict in button skeleton

tw15egan (approved):
LGTM 👍 ✅ Thanks for contributing! 🎉 💪🏻

imp-dance:
@tay1orjones Conflict should be fixed now, merged with latest main 👍

PR #14015, opened 2023-06-15
style(components): fix styles when only a single component is imported

### LINKED ISSUE #14008: Ensure all component styles using `layer` or `field` tokens @use the layer module
As previously reported https://github.com/carbon-design-system/carbon/discussions/13968, when importing component styles on an individual basis, some required modules are not included resulting in build errors or invalid styles.

This was recently made worse with the contextual layout token PR which introduced a number of new tokens that need to be made available to components.

```[tasklist]
- [ ] Go through every component stylesheet that uses `layer` or `field` tokens and add an `@use` for the layer module
```

Temporary workaround for this is to use `@use @carbon/styles` instead of individual component style modules.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14008

This ensures all component styles have the necessary imports to render correctly if only that single component is imported. 

#### Changelog

**Changed**

- Added necessary imports so the component rendered correctly in isolation (`layer`)
- Ensured `component-reset.reset` was being applied properly so there were no padding issues when the global reset was not present
- Reordered some style `@use` declarations

#### Testing / Reviewing

Question: Should font declarations be included in each component import? Right now it does not set `IBM Plex Sans` anywhere. If so, we just need to add `@include font-family('sans');` right after the `component-reset` for each component

Test out a bunch of components by pulling this down locally, going to `packages/react/.storybook/styles.scss` and commenting everything out. Then, test individual components like so `@use '@carbon/styles/scss/components/button';` and ensure it renders as expected. Also, there should be no visual differences when importing _all_ styles


francinelucca (changes requested):
Tested the following components:
- Copy Button ✅
- Button ✅
- Datepicker ✅
- Dropdown ✅
- Loading ✅
- MenuButton ✅
- Search ✅
- Select ✅
- Tabs ✅
- Breadcrumb
   I see a mismatch on the padding of the `ol`:
image
image

- Popover
  I see some extra spacing:
image
image

- TimePicker
  Invalid icon color is off:
image
image

Also noticed the difference in font, I do feel those should carry over, or is that supposed to be set somewhere at the lop level of the application? 🤔

tw15egan:
@francine pushed up some fixes, namely adding the `font-family` declaration to the `type-style` declaration blocks in the type package

tay1orjones (approved):
Looks good to me, I couldn't find any defects perusing the storybook for a while. Snapshots do need updated though, looks like the changes from adding the font-family and such are the root of the snapshot changes.

francinelucca (changes requested):
Tested the following:
- Accordion ✅
- Breadcrumb ✅
- Multiselect ✅
- Popover ✅
- Slider ✅
- StructuredList ✅
- ContentSwitcher ✅ 
- Treeview ✅
- Modal
  I see a difference in spacing:
 
image
image


- TimePicker
   Warn icon is read instead of yellow
image
image


Overall:
I see a difference in the text color:

https://github.com/carbon-design-system/carbon/assets/40550942/fac22a23-41b0-4620-964b-1e30c80b7765

tw15egan:
@francinelucca The text change is coming from the `-webkit-font-smoothing: antialiased` that is applied to the body, and is inherited by all components. I'll see where it makes most sense to try and emit this to body somewhere

tw15egan:
Reverted all type package changes, if a user is expecting to use Carbon type/fonts they would need to also pull in ` '@carbon/styles/scss/type'`

tay1orjones:
>  if a user is expecting to use Carbon type/fonts they would need to also pull in `@carbon/styles/scss/type`

@tw15egan That's probably going to be the default expectation by and large from everyone, right? I think we need to heavily document this in the repo (@carbon/react readme, @carbon/styles readme, v11.md changelog, etc) and the website.

Is this isolated to only use-cases where component styles are brought in individually? Consumers won't need to do this if they just `@use @carbon/styles`?

tw15egan:
@tay1orjones correct, if you pull in `@carbon/styles`, you get it all. This is _only_ if you pull in individual component styles, you'd also need to make sure you are pulling in our type package.

tay1orjones:
@tw15egan That makes sense 👍 

If they `@include type.reset();` across multiple files, it'll still result in duplication in the final stylesheet, right? Do we need to create a new module that automatically emits the type reset via an `@include` so the sass module system would prevent that duplication and only emit it once?

tw15egan:
I hadn't thought about that, I would assume the `type.reset()` would only be used once on the main `scss` file, as it just sets global styles that are inherited down

francinelucca (approved):
Confirmed previously reported issues have been fixed
Confirmed can correctly pull in types using `@use '@carbon/styles/scss/type' as type;` and `@include type.reset();`

Approving pending Taylor's docs comments

tay1orjones (approved):

PR #14017, opened 2023-06-15
fix: dropdown contrast focus styles

### LINKED ISSUE #12720: [a11y]: Dropdown component contrast for item with focus/selection fails
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

### React version

n/a

### Assistive technology

n/a

### Description

When the dropdown is open, there _is_ a thin blue focus indicator around the entire option list. However, the actual item with 'focus' (i.e., the item I have moved to with my arrow keys) is only indicated by a light grey highlight that contrasts only 1.2:1 against the dropdown list background. This applies to dropdown, multiselect and combobox, but is most noticeable on multiselect, where multiple items can be selected.
image

This must contrast 3:1 with the unfocused options. And then the text in the option must also contrast 4.5:1 against any background. It is likely that an inversion may be necessary (white text on darker background) This is how a default select element is done (although note that the browser default fails text contrast, so don't emulate it!). 
image

You could also pursue something other than a fill to distinguish the item with focus. For instance, the select-only combobox in the [ARIA Authoring practices](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) just uses a focus rectangle to show which has focus. (Note that they also add a black outline to the options list so that it is distinguishable from other page content. If you replaced your thin blue focus indicator around the list with a black line and then put a focus on the option itself, this would resolve it.)
image

### WCAG 2.1 Violation

### Reproduction/example

### Steps to reproduce

Open Dropdown in storybook and arrow through options.

### Code of Conduct

```[tasklist]
### Tasks
- [ ] Investigate the following and assess with design:
- [ ] Remove the 1px blue focus border around the listbox
- [ ] Change the focus border on the input to be 1px when the listbox is opened
- [ ] Apply a 2px blue focus border to the currently focused item within the listbox
```

andreancardona:
Closes #12720 

Paired up with @francinelucca - got a good solution going 🎉

Screenshot 2023-06-15 at 17 10 07

#### Testing / Reviewing
- This must be tested with keyboard navigation 
- Make sure Dropdown, Combox & Multiselect have the above focus styling and the light-grey background has been removed - through all the different variants
- Hover styles should remain the same on mouse
- @kingtraceyj tagging you for a visual review :)


kingtraceyj (approved):
focus styles match spec. looks good to go! 💥

guidari (approved):
LGTM! 
I just added a `aria-owns` in the ComboBox.tsx because when we turn on the VO in FireFox it closes the ComboBox when the user tries to navigate. 
Let me know if that make sense

tay1orjones (approved):

andreancardona:
Trying to see the best way to fix the failing avt test

andreancardona:
> > Trying to see the best way to fix the failing avt test
> 
> Is this related to the `aria-owns` that was added @andreancardona ? Looks like the tests broke after that addition I'll take a look to help fix it

yep it is related - just trying to find a solution for both the test and the FF bug

PR #14019, opened 2023-06-16
fix(contained-list): explicitly set "list" role

### LINKED ISSUE #13938: [a11y]: Contained List [all variants] - semantics not announced by the VO in iPhone
### Package

@carbon/react

### Browser

iPhone / Safari

### Operating System

MacOS

### Package version

carbon/react@1.30.0

Manual QA

### Assistive technology

Voice Over

### Description

VO iPhone in Safari does not identify the semantics of the UL present in the page. These are announced only as text without any semantics. The same is working in JAWS and in android talkback Samsung phone with chrome. Note: this issue was verified in android to cross check the screen reader behavior.

### WCAG 2.1 Violation

1.3.1 Info and Relationships

### Reproduction/example

Record `Number 26` in the below table:

### Steps to reproduce

User Impact: when the semantic structure of the element is not conveyed programmatically, Blind screen reader user would not be able to perceive it

Generally VO is expected to announce the beginning and the end of the list when we navigate using Left/right swipe. Note: It seems to be working in Unordered list component. And in the following example link as well.
https://codepen.io/svinkle/pen/YxJEdY

Press ESCAPE to close this message. 

### Code of Conduct

janhassel:
Closes #13938

Ref: https://bugs.webkit.org/show_bug.cgi?id=170179#c1
Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/list-style#accessibility_concerns

#### Changelog

**Changed**

- Explicitly set `role="list"` to `ul`

#### Testing / Reviewing

Follow reproduction steps in #13938


tay1orjones (approved):

francinelucca (approved):
Confirmed can now hear the beginning and end of the list on iPhone VO 🚀

PR #14022, opened 2023-06-16
fix(sass): remove colons from classnames

### LINKED ISSUE #3599: Website update to Carbon v11.31.3 causes build error
Due to new changes introduced in carbon v11.31.3 (https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Layout/index.js#L80-L86) we're getting some errors in website when trying to update. This is not a problem with @carbon/react itself but likely with some extra configurations needed on gatsby-theme

image

tay1orjones:
Closes https://github.com/carbon-design-system/carbon-website/issues/3599

`css-loader` is a dependency of `gatsby`, and when it converts classnames to js exports, `css-loader` does not convert reserved and control filesystem characters (`<>:"/\|?*`) contained within the className. [link to docs](https://github.com/webpack-contrib/css-loader#localidentname:~:text=The%20%5Blocal%5D,to%20%2D.)

Since the layout constraint classes contained colons, this resulted in css-loader outputting invalid js variable names that also contained colons (as shown in the original issue), causing gatsby site compilation to fail. This would likely fail in any webpack-based app as well.

#### Changelog

**Changed**

- modify usage of `:` within classnames to be `__` instead
- update related documentation

### DIFF
--- a/packages/react/src/components/Layout/index.tsx
+++ b/packages/react/src/components/Layout/index.tsx
@@ -156,11 +156,11 @@ const LayoutConstraint = React.forwardRef<ReactNode, LayoutConstraintProps>(
         size,
         density,
       }).map(([group, constraints]) => ({
-        [`${prefix}--layout-constraint--${group}:default-${constraints?.default}`]:
+        [`${prefix}--layout-constraint--${group}---default-${constraints?.default}`]:
           constraints?.default,
-        [`${prefix}--layout-constraint--${group}:min-${constraints?.min}`]:
+        [`${prefix}--layout-constraint--${group}---min-${constraints?.min}`]:
           constraints?.min,
-        [`${prefix}--layout-constraint--${group}:max-${constraints?.max}`]:
+        [`${prefix}--layout-constraint--${group}---max-${constraints?.max}`]:
           constraints?.max,
       }))
     );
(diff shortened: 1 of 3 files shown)


andreancardona (approved):

francinelucca (approved):
🚀

janhassel:
@tay1orjones Do we know why the grid module isn't throwing the same error even though it's also using colons in the classnames?

https://github.com/carbon-design-system/carbon/blob/8c21b1998401096ff59d3ab0fc31355c1f7f27c2/packages/grid/scss/_css-grid.scss#L218-L300
https://github.com/carbon-design-system/carbon/blob/8c21b1998401096ff59d3ab0fc31355c1f7f27c2/packages/react/src/components/Grid/Column.tsx#L380-L397

tay1orjones:
@janhassel Yeah I think it's because the CSS Grid styles aren't actually imported in `gatsby-theme-carbon` (the whole thing is built using flex-grid). Same thing with the website. If these were imported they'd probably show the same error. 

Renaming those classes to not have colons should probably be a breaking change and isn't something I want to do unless we absolutely have to. So far we haven't had any reports about these styles failing to build in webpack-based apps.

tw15egan (approved):

PR #14027, opened 2023-06-16
docs: add sprint planning docs

francinelucca:


sstrubberg (commented):
This is incredible. So thorough. You've covered every point of contact in preparing for sprint. My one suggestion, would be to scrub the document for any links that point to a hardcoded quarter. That way you aren't having to update this document every time we enter a new quarter.

francinelucca:
> This is incredible. So thorough. You've covered every point of contact in preparing for sprint. My one suggestion, would be to scrub the document for any links that point to a hardcoded quarter. That way you aren't having to update this document every time we enter a new quarter.

@sstrubberg only one of the links in the doc is referencing a quarter directly, I didn't see a way around this one so I just added a comment to manually update the quarter:

image

The other links are pointing to project board views which show up the current quarter when you open them but that should be updated at the board-level as the quarter changes and the same links will continue to work.

sstrubberg (approved):

tay1orjones (approved):
This is great, thank you! 🙏

PR #14028, opened 2023-06-18
fix(Datepicker): fix calendar cannot close issue

### LINKED ISSUE #14025: [Bug]: Datepicker - Cannot close calendar on keyboard interaction
### Package

@carbon/react

### Browser

Chrome

### Package version

1.31.0

### React version

18.2.0

### Description

https://github.com/carbon-design-system/carbon/assets/40550942/7f695c5d-d551-4ac7-817c-3ffbc7d1a43d

### Reproduction/example

https://stackblitz.com/edit/github-3wiety?file=package.json

### Steps to reproduce

Use keyboard navigation, enter 2022-01-01 as the first date, tab to second date and enter 2022-01-07.
Try to close the calendar, both via keyboard or outside click does not work

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tongyy:
Closes #14025

This is caused by misusing js built-in new Date api to parse date string. Almost all main stream browsers consider Y-m-d format as ISO 8601 format which has zero hour offset. 

ex: you can test this in console
new Date("2022-05-05");
Wed May 04 2022 20:00:00 GMT-0400 (Eastern Daylight Time)
new Date("2022/05/05")
Thu May 05 2022 00:00:00 GMT-0400 (Eastern Daylight Time)

#### Changelog

**Changed**

- use parseDate from flatpicker instead of using builtin new Date with date string as initial arg.

#### Testing / Reviewing

Follow reproduction steps in [14025](https://github.com/carbon-design-system/carbon/issues/14025)


packages/react/src/components/DatePicker/plugins/fixEventsPlugin.js
```diff
@@ -43,6 +43,8 @@ export default (config) => (fp) => {
     }
   };
 
+  const parse = (dateStr) => fp.parseDate(dateStr, fp.config.dateFormat);
```
francinelucca: let's give this function a more intentful name
```suggestion
  const parseDateWithFormat = (dateStr) => fp.parseDate(dateStr, fp.config.dateFormat);
```

francinelucca (changes requested):
Thanks for putting this PR together @tongyy! 
Mostly LGTM just a function rename suggestion.
Also if this is your first contribution feel free to add yourself to the contributors table: https://github.com/carbon-design-system/carbon/blob/main/.github/CONTRIBUTING.md#5-add-yourself-to-the-contributor-list

Note: noticed calendar does not close on tab out when using keyboard navigation and opened up https://github.com/carbon-design-system/carbon/issues/14033

packages/react/src/components/DatePicker/plugins/fixEventsPlugin.js
tongyy: updated

andreancardona (approved):
Tested and looks good thanks @tongyy! 🎉

francinelucca (approved):

tay1orjones (approved):
Thank you!

PR #14044, opened 2023-06-20
Added new React 18 hook `useId` and tests

### LINKED ISSUE #11513: React 18 Enhancements: Update `useId` usage to use new hook available in React 18
Depends on #11308 

React 18 introduces a new hook called `useId`: https://reactjs.org/blog/2022/03/29/react-v18.html#useid. From the docs:

> `useId` is a new hook for generating unique IDs on both the client and server, while avoiding hydration mismatches. It is primarily useful for component libraries integrating with accessibility APIs that require unique IDs.

This pattern matches the problem that our internal `useId` hook is attempting to solve. This issue is for transitioning our internal `useId` hook over to the new `useId` hook. However, tt's important that we do this in a backward-compatible way by checking for the existence of this hook before using it.

**Sample**

_Note: this is just an example of checking for support and does not specify a solution_

```jsx
import React from 'react';

export const useId = React.useId ?? function useId(prefix) {
  // implementation;
};
```

**Checklist**

- [ ] Update internal `useId` to use `React.useId`
- [ ] Make sure fallback is still supported for React versions <18

guidari:
Closes #11513

It was added the new hook in React 18 called `useId`. This hook will take of hydration for us now. 
We are checking if the user is running a compatible version and calling the function `nativeReactUseId`, otherwise we will keep using our old logic.

#### Changelog

**New**

- Added `useId` from React 18
- Added tests for the `useFallbackId` hook

**Changed**

- Changed the `useId.server-test.js` test. With this new `useId` hook an id will be genarate in the server also.

#### Testing / Reviewing

To test the code you have to switch between React versions. 

I created a [Stackblitz](https://stackblitz.com/edit/github-b5r8er-rfgyun?file=src%2FApp.jsx) where you can test that by following these steps:

- Change the React version to an older version (17, 16...) in the `package.json`
- Restart the server by running `npm install` and `npm run dev` 

In the right panel you should see the changes.

___________

To check in the Storybook you can go to the Accordion component and check if it has an id populate with the `useId` hook

Screenshot 2023-06-20 at 17 10 04


tay1orjones (approved):
This looks great, thanks for updating the server test as well! 

I can't think of any ways that this could impact consumers already using React 18. The format of the id will change slightly, but tests etc. shouldn't be being written against hardcoded Ids generated from this `useId()`.

tw15egan (approved):
Awesome 
LGTM 👍 ✅

PR #14076, opened 2023-06-23
fix(Search): close expandable search on escape key

### LINKED ISSUE #13941: [a11y]: ContainedList [with expandable search] - search edit field incorrect focus behavior
### Package

@carbon/react

### Browser
FF

### Package version

carbon/react@1.30.0

Manual QA

### Assistive technology
JAWS

### Description
The search field is triggered on down arrow navigation instead of user pressing enter on search button

### WCAG 2.1 Violation
3.2.1 On Focus

### Reproduction/example
Record `Number 29` in the below table:

### Steps to reproduce

In JAWS with FF, The search edit field is triggered when the user presses down arrow from the "search" button. This behavior confuses the blind screen reader user. Because the user may think that the search field is always present without requiring to activate the search button.

### User Impact
When an element automatically changes after receiving focus, it disorients users who have low vision or cognitive disabilities.

### Recommended fix
The search field is triggered on down arrow navigation instead of user pressing enter on search button

### Code of Conduct

### LINKED ISSUE #13943: [a11y]: ContainedList [with expandable search] - keyboard action need to close expanded search
### Package
@carbon/react

### Browser
FF

### Package version
carbon/react@1.30.0

Manual QA

### Assistive technology

JAWS

### Description

Search edit field

Unable to close the search edit field using keyboard

### WCAG 2.1 Violation
n/a

### Reproduction/example

Record `Number 31` in the below table:

### Steps to reproduce

The search edit field closes When the user clicks any ware in the screen. However there is no equivalent action for the keyboard only users.

When the standard keyboard operation method is not supported, it impacts the user’s ability to interact with page controls. This affects screen reader and keyboard-only users. 

### Recommended fix: 
User should be able to close the search field using keyboard
We can use the `esc` key to close expanded search

### Code of Conduct

francinelucca:
Closes #13943 

Closes https://github.com/carbon-design-system/carbon/issues/13941

#### Changelog

**New**

- `ExpandableSearch` closses on Escape key keyDown (if the input is clear, otherwise it'll clear the input first)
- Focus expand button on `ExpandableSearch` closed with Escape key
- Focus styles for ExpandableSearch expand button
- `aria-controls` & `aria-expanded` properties for `ExpandableSearch`'s activation button

**Changed**

- Corrected signature for search's `onExpand` (was not taking event param into account) and added `KeyboardEvent` as an event param type
- `ExpandableSearch` now opens on button activation (enter/space) instead of focus
- updated snapshots

#### Testing / Reviewing

Test `ExpandableSearch` and `ContainedList with ExpandableSearch` stories and make sure everything behaves as expected but should now:
- Open search on magnifier button enter/click instead of input focus
- be able to clear the input when populated using the `Escape` key
- be able to close the search when expanded using the `Escape` key (when the input is clear)
- Focus should also go to the search button when closed via Escape
- Should be re-expandable by activating said button. `Search` story should behave exactly the same as it did before these changes.


francinelucca:
Tagged visual review to get design's perspective on focusing the button on close via escape? Seems to me like the focus needs to go somewhere after closing via escape but also focusing the magnifier button means the expandable search will reopen on tab out since the input is the next tabbable element in the dom and it expands on focus, so you'd have to tab twice to skip over it. Would also love to hear your thoughts on this @mbgower 

Expandable: https://deploy-preview-14076--carbon-components-react.netlify.app/?path=/story/components-search--expandable
Deployment: https://deploy-preview-14076--carbon-components-react.netlify.app/?path=/story/components-containedlist--with-expandable-search

mbgower:
Yeah, there's something a little odd about this interaction. It might help to put it in a simple page where it might appear, so we can see how the interaction is expected to take place.

My take on it, in isolation, is as follows:
- It looks like the search button is in the tab order AND that it expands on focus
- I'm not sure why the search button should expand on focus. Why wouldn't it open on activation? Is the idea this is just for saving space? Otherwise, why not just have the search field persist and not bother with the search button?
- It makes no sense that I escape out of the search field and go to the search button, only to find the next item in the tab order is the search field. Either it's not in the order until I activate it OR it is in the search field, in which case, it seems impossible for the search button to exist as an independent tab stop
-

francinelucca:
@mbgower in [production](https://react.carbondesignsystem.com/?path=/story/components-search--expandable) the search button is not tabbable/focusable at all, the input is focusable and expands on focus. 
In this PR, where we're introducing "close on escape" I'm programmatically setting focus on the button after "Escape" for lack of a better place to put it, I guess my main question is where should the focus go, or is it ok to leave nothing focused on the page when the search collapses? 🤷🏽‍♀️

The problem here is with the introduction of that tab stop in the button and the input continuing to be "Expandable on focus" , we get this weird behavior, I'm really not sure what a "fix" here would look like 😅

mbgower:
> the input is focusable and expands on focus.

That's a little odd. But setting that aside for a moment, I think if the functionality of the Esc is to clear any search value entered, then after pressing Esc, it would just leave me at the cursor in an empty input (just like would happen in any input). You cannot collapse the search input while it has focus, IMO.  That defies the model you have set up.

If you look at the behaviour of the search button at the top of the carbondesignsystem.com page, you'll see a search input Esc pattern that works because the search button takes focus and its activation triggers the search input. That is one option.  (But I've personally always found that suspect on a web page; it saves little space and makes more time on task.) The other is that the search button in the table never exists, and it's just the input. But if that's so, IMO there is no way to allow the input to 'collapse' except when the user physically leaves it. Esc cannot logically do that in the current configuration.

A bigger question for design is 'what do you actually gain by hiding the search input?' I'd really like to understand the thinking.

francinelucca:
For some context, this change came about as a suggestion from Ram that he'd expect there to be a way to close the expandableSearch other than tabbing out of it.

@carbon-design-system/design from what @mbgower is saying, sounds like the only two options from an a11y standpoint here are to:
a) Not have the search be input expand on focus, but be activated via the button instead. This way the input could be collapsed via escape and the focus could be returned to the activation button.
b) Keep this as is and have the "escape" key be a "clear" for the input, in which the focus would stay on the input and the only way to collapse would be to tab out of it.

Any thoughts on either/or?

andreancardona:
@francinelucca I can bring this up in a call on Tuesday, June 27th - with Ram and see which is preferred / get some insight. Let me know if that works!

andreancardona:
@francinelucca @mbgower After speaking with Ram on today's call we have clarity on the desired interaction: 
- Not have the search be input expand on focus, but be activated via the button instead. This way the input could be collapsed via escape and the focus could be returned to the activation button.
- Add an aria-expanded on the button trigger to announce its expanded and collapsed state

laurenmrice:
@andreancardona The above makes sense to me.

tay1orjones:
> is adding a KeyboardEvent to onExpand a breaking change?

This is a grey area. While this might break consumers' tests, I think this is a necessary change to fix a high severity issue for non-sighted users. It might be disruptive, but imo it's worth the cost to fix non-sighted/keyboard nav users' broken experience.

laurenmrice (approved):
Looks good to me! ✅

I will defer to @mbgower on the below comment:

> In website there's a "double escape" behavior in which the first escape clears the input and the second one collapses it, wondering if we should mimmick that? or not have the escape key clear the input at all?

mbgower:
I think you have that working pretty well on the main search, right? I think it's good to emulate. Once there is a string entered, Esc will clear. If no string, Esc collapses

> In website there's a "double escape" behavior in which the first escape clears the input and the second one collapses it, wondering if we should mimmick that? or not have the escape key clear the input at all?

alisonjoseph (approved):
This LGTM and appears to be working as intended after reading through the comments. 😅

tw15egan (commented):
Functionality is there, but I had one question for @laurenmrice and the focus state: should the outline be `2px` to match the focus outline when the search is expanded? I figured it would match the focus appearance of a ghost button, but right now it's only 1px and seems off

laurenmrice:
@tw15egan Oh yep, good point. The focus should be 2px on the search icon and on the full expanded search for consistency. 👍

tw15egan (approved):
LGTM 👍 ✅

andreancardona:
also closes #13941

PR #14089, opened 2023-06-26
fix(upgrade): add size prop codemods to upgrade package

### LINKED ISSUE #14001: [Bug]: Codemods cannot be ran from the upgrade CLI, `size-prop-updates` `small-to-size-props`
### Package

@carbon/upgrade

### Browser

Chrome

### Package version

latest

### Description

There are two codemods in the [transforms folder](https://github.com/carbon-design-system/carbon/tree/main/packages/upgrade/transforms) that are not able to be ran, `size-prop-updates` `small-to-size-props`. It looks like these were committed before we added the codemod runner to `@carbon/upgrade`.

The [available migrations](https://github.com/carbon-design-system/carbon/blob/36685a0be5e62db11a60d938c76b1c49a402c99a/packages/upgrade/src/upgrades.js#L147) needs to be updated so that these codemods can be ran via `npx @carbon/upgrade migrate {migrationName}`.

### Reproduction/example

n/a

### Steps to reproduce

Try to run these codemods against a sample repo like the tutorial or an example from the monorepo, it will fail

image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14001

Adds in the 2 size codemods to `@carbon/upgrade` package

#### Changelog

**New**

-  `size-prop-update` and `small-to-size-prop` to `@carbon/upgrade`

#### Testing / Reviewing

Ensure it all builds properly. Not sure if we can test usage of something like `npx @carbon/upgrade migrate small-to-size-prop` until this is merged and a new version is released.


francinelucca (approved):
LGTM

andreancardona (approved):

PR #14112, opened 2023-06-28
feat(Tag): add as prop to Tag to specify wrapping element (#13959)

### LINKED ISSUE #13959: [Feature Request]: Add property for Tag to specify html tag to be rendered
### The problem

We need to render `Tag` as an inline element in text of documentation pages created by Markdoc from .md files. The problem is that `Tag` renders into `<div>`,  and text paragraph rendered by Markdoc is always wrapped in `<p>` element, so react throws hydration error, because `<div>` is not allowed in `<p>`.

Example of code, that throws hydration error:
```
import { Tag } from '@carbon/react';
const ParagraphWithTag = () => {
  return <p>Paragraph with <Tag>tag content</Tag> inside.</p>;
}
```

### The solution

Add `as` property to `Tag` component, where html tag can be specified, to which component is then rendered.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

kapetr:
Closes #13959

Added `as` prop to `Tag` component, so the wrapping element can be changed from default `div` (or `button` in case of `onClick` prop is provided). If `as` is provided it overrides the logic of choosing between `div` or `button`, I consider that correct behaviour, but please revise.

#### Changelog

**New**

- add `as` prop to `Tag` component


tw15egan:
@tulivlk thanks for contributing! First off, to update snapshots, you'll need to run `yarn test -u` from the root directory. Secondly, we'll need you to sign the DCO. Instructions can be found [here](https://github.com/carbon-design-system/carbon/pull/14112#issuecomment-1611860988)

tulivlk:

tw15egan (approved):
Looks great! Thanks for contributing

alisonjoseph (approved):
LGTM! 🥳

Issue #13946, opened 2023-06-07
[a11y]: ContainedList [with icons] - screen reader does not identify icons


andreancardona:
### Package

@carbon/react

### Browser
Safari & FF & iPhone

### Package version

carbon/react@1.30.0

### Automated testing tool and ruleset
Manual QA

### Assistive technology
JAWS & VO

### Description
Icons present in the component

Screen reader does not identify the Icons present in the component

### WCAG 2.1 Violation
n/a

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?args=&id=components-containedlist--with-icons&viewMode=story

### Steps to reproduce
Record `Number 34` in the below table:
https://airtable.com/shrFsaeHhImOkuVNR

**_If_** these icons convey meaningful information, these should be made accessible to the screen reader. Screen reader should identify the image role with appropriate alt text that describes the purpose of each image. 

This needs to be refined by the team

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

mbgower:
There is no guidance on icon usage in lists in the Structured List topic.
I completely agree that where the icon adds meaning to the list, it should have an ALT. I am much less clear on any situation where the icon for a list would be more than decorative.
For instance, in the [storybook version](https://react.carbondesignsystem.com/iframe.html?args=&id=components-containedlist--with-icons&viewMode=story), I cannot see how any of the icons provided would not be decorative add-ons to list items that might be fruit, plants, berries or fish.

![image](https://github.com/carbon-design-system/carbon/assets/14298245/fd46bd98-dcee-44f7-9eaa-9a21c6aa5bee)



The only scenario I can think of is if there was a list of shopping purchases and the icons were added to designate the food type. But even then, unless there were two identically named list items and the icon provided actual information to distinguish them, it seems likely to be decorative.
Are there examples?

andreancardona:
After discussing this with the team - it is confirmed this is only a descriptive icon and does not provide meaningful information. This not a violation - therefore, closing this issue. Thank you for the help @mbgower!

PR #14124, opened 2023-06-29
fix(Docs): add aria-label to ContentSwitcher icons

### LINKED ISSUE #13979: [a11y]: ContentSwitcher [icon only & icon only with layer] - screen reader doesn't identify icons
### Package

@carbon/react

### Browser
Safari & FF & iPhone

### Package version

carbon/react@1.30.0

Manual QA

### Assistive technology
JAWS & VO

### Description

Screen reader does not identify the Icons present in the component

Screen reader does not identify the Icons present in the component. Hence the user would not know their presence.

### WCAG 2.1 Violation
n/a

### Reproduction/example
Record `Number 49` in the below table:

### Steps to reproduce

If these icons convey meaningful information, these should be made accessible to the screen reader. Screen reader should identify the image role with appropriate alt text that describes the purpose of each image. 

This needs to be confirmed with the team 

### Code of Conduct

francinelucca:
Closes #13979 

#### Changelog

**Changed**

- Added `aria-label` attributes to icons in icon-only content switcher stories

#### Testing / Reviewing

Use VO or any other assistive technologies to make sure icon is read when going over the content switcher tabs


francinelucca:
@mbgower @carbon-design-system/developers @carbon-design-system/design calling out an "is this even necessary?" here. 
The VoiceOver already reads the associated "text" to a given tab in the content switcher so for our example previous to this change it read: "Table of Contents, selected tab, 1 of 1" and now it'll read "Table of Contents, selected tab, group 1 of 1, Table of Contents Icon"

Compare production: https://react.carbondesignsystem.com/?path=/story/components-contentswitcher--icon-only
To this deployment: https://deploy-preview-14124--carbon-components-react.netlify.app/?path=/story/components-contentswitcher--icon-only

is this redundancy ok? or should we use an example here in which the "test" and the icon denote different information? but I couldn't think of one 🤷🏽‍♀️

guidari:
> @mbgower @carbon-design-system/developers @carbon-design-system/design calling out an "is this even necessary?" here. The VoiceOver already reads the associated "text" to a given tab in the content switcher so for our example previous to this change it read: "Table of Contents, selected tab, 1 of 1" and now it'll read "Table of Contents, selected tab, group 1 of 1, Table of Contents Icon"
> 
> Compare production: https://react.carbondesignsystem.com/?path=/story/components-contentswitcher--icon-only To this deployment: https://deploy-preview-14124--carbon-components-react.netlify.app/?path=/story/components-contentswitcher--icon-only
> 
> is this redundancy ok? or should we use an example here in which the "test" and the icon denote different information? but I couldn't think of one 🤷🏽‍♀️

@francinelucca I feel like the "as is" looks better when the VO is reading. I also can't think of an example where the icon denote a different information from the "text" prop. 
I wonder what other folks think about that also.

andreancardona:
Yeah I think this one is similar to https://github.com/carbon-design-system/carbon/issues/13946 - where it doesn't seem to provide meaningful information.

andreancardona:
Just chatted with @mbgower that confirmed on JAWS this would be redundant label and it is not needed - closing the issue.

PR #14125, opened 2023-06-29
fix: storybook sort function

francinelucca:
Refactors sort function so that it works both on V6 & V7 store.

#### Testing / Reviewing

CI VRT should continue to pass and deployment storybook stories should be in correct order


tay1orjones (approved):
Thanks so much for fixing this 🙏 Tricky situation having to deal with both v6 and v7 story stores

andreancardona (approved):
this is cool! 💯

PR #14128, opened 2023-06-30
feat(Toggle): adds typescript typings to Toggle

### LINKED ISSUE #12564: Add TypeScript types to Toggle

erifsx:
Closes #12564

Adds typescript typing to the Toggle component

#### Changelog

**New**
- Copies the typings for Tooltip from the definitely typed repo and updates to match current typings.

**Changed**
- the jsDoc for useControllableState had onChange as a required field and the jsDoc now types it as optional

#### Testing / Reviewing

- should build
- storybooks should work
- all existing fields should be accessible from typescript


erifsx:

erifsx:
recheck

packages/react/src/components/Toggle/Toggle.tsx
```diff
@@ -1,16 +1,96 @@
...
+  /**
+   * If true, the side labels (props.labelA and props.labelB) will be replaced by
+   * props.labelText (if passed), so that the toggle doesn't render a top label.
+   */
+  hideLabel?: boolean;
+
+  /**
+   * Provide an event listener that is called when the control is toggled
+   */
+  onClick: MouseEventHandler<HTMLDivElement> | undefined;
```
francinelucca: looks like this function also gets called when the button is activated via keyboard? in that case we need to adjust the type
```suggestion
  onClick: MouseEventHandler<HTMLDivElement> | MouseEventHandler<HTMLButtonElement> | undefined;
```

francinelucca (changes requested):
Thanks for putting this PR together @erifsx ! mostly LGTM, just one minor change request 🙏🏻.
TIA

packages/react/src/components/Toggle/Toggle.tsx
erifsx: Thanks, good catch! 

Added typings for both mouse and keyboard events.

tay1orjones (approved):

francinelucca (approved):

PR #14131, opened 2023-07-03
fix(peerDependencies): make sass optional for styles - Solves @carbon/charts example issue

nstuyvesant:
I am a contributor to `@carbon/charts`. This PR solves a problem with the Carbon Charts StackBlitz examples forcing users to install sass. This is because each example calls `import '@carbon/styles/css/styles.css` and `sass` is a non-optional peerDependency for `@carbon/styles`.

As `@carbon/styles` can be used without `sass` (as is the case shown above importing just the CSS file), `sass` should not be a mandatory peerDependency. This PR solves that.

#### Changelog

**Changed**

- Added peerDependenciesMeta property to packages/styles/package.json indicating `sass` is an optional peer dependency

#### Testing / Reviewing

Setup any kind of project and add `@carbon/styles` as a dependency and leave out `sass`. You will get a warning from npm or yarn.

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -35,7 +35,7 @@
     "sass": {
       "optional": true
     }
-	},
+  },
   "dependencies": {
     "@carbon/colors": "^11.16.0",
     "@carbon/feature-flags": "^0.15.0",
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -31,10 +31,10 @@
   "peerDependencies": {
     "sass": "^1.33.0"
   },
-  	"peerDependenciesMeta": {
-		"sass": {
-			"optional": true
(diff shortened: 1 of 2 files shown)


theiliad (approved):
LGTM

tay1orjones (approved):
Thanks for adding this. Originally `@carbon/styles` only shipped sass. The compiled css file was was added after the initial package release and I don't think we had considered that the `sass` peer dep should be optional.

theiliad:
Thanks for the quick turnaround 💯

nstuyvesant:
Indeed! [This PR](https://github.com/carbon-design-system/carbon-charts/pull/1612) includes the updated `@carbon/styles`.

PR #14144, opened 2023-07-06
fix(UIShell): various fixes to avoid null pointer assertions

tay1orjones:
Surfaced in office hours this morning, a few bugs relating to UIShell and TextArea

#### Changelog

**Changed**

- protect against null pointer exceptions with optional chaining
- only pass handleSwitcherItemFocus to components that accept that prop


francinelucca (commented):
is it expected for the switcher not to show any options here? looks like the stories changed a bit so not sure if that's intended 
image
I'm also not able to open the switcher on this one 
image

tay1orjones:
@francinelucca sorry I forgot the return statements, should be fixed now 👍

francinelucca (approved):
🚀

tw15egan (approved):
LGTM 👍 ✅

PR #14153, opened 2023-07-07
fix(useid): guard against useId not being exported

tay1orjones:
We've gotten reports that after updating to v11.33.0, consumers using React React <v18 get compiler errors.

image

This is because the import of useId from React 18 is not guarded for lower versions of react that do not export that function.

#### Changelog

**Changed**

- guard the usage of useId by not importing it directly


tw15egan (approved):
LGTM 👍 ✅

guidari (approved):

PR #14156, opened 2023-07-07
fix: fixed VO and JAWS problem in FF and Safari

### LINKED ISSUE #13888: [a11y]: ComposedModal [all variants] - identifies visually hidden controls
### Package

@carbon/react

### Browser

Safari & FF

### Package version

carbon/react@1.30.0

Manual QA

### Assistive technology

JAWS & VO

### Description

ComposedModal [all variants]
Screen reader identifies visually hidden controls at the beginning and the end of the modal

### WCAG 2.1 Violation

4.1.2 Name, Role, Value 

### Reproduction/example

Record `Number 24` in the below table: 

### Steps to reproduce

In JAWS with FF, After the dialog is displayed, When we navigate with the up/down arrow keys through the dialog content, Screen reader identifies A button with the label "Focus sentinel" at the beginning and end of the dialog. This is visually not present and identified only by the screen reader.

In VO iPhone Safari, The same issue exists. Along with that, When We navigate to the "Focus sentinel" button at the beginning of the dialog, VO focus wraps back to the "Add" button automatically. Similarly, when we navigate to the "Focus sentinel" button in the end, the VO focus wraps back to the "close" button at the top.

### User impact: 
When screen reader identifies the visually hidden elements, it can be misleading for the blind user

### Recommended Fix: 
The mentioned controls should be hidden from the screen reader as well

### Code of Conduct

guidari:
Closes #13888 

Added `tabIndex={-1}` to avoid that screen readers passes throught it.

_I tested only with VO, but it should work on JAWS also._

#### Testing / Reviewing

- Open the FireFox and Safari
- Open Modal component
- Tab/Arrow in the component

You should expect the same behaivor as Chrome in both browsers


tw15egan:
The issue with this is that it breaks the focus wrap behavior, looks like the tab order now goes outside of the modal. We need to be able to focus these elements for the `focusWrap` behavior to work unfortunately

guidari:
> The issue with this is that it breaks the focus wrap behavior, looks like the tab order now goes outside of the modal. We need to be able to focus these elements for the `focusWrap` behavior to work unfortunately

Ohh! I got it

francinelucca (dismissed):

guidari:
@tw15egan I paired up with Taylor on that so we could tested using JAWS.

Thanks @tay1orjones for pairing up!

tay1orjones (approved):
This also moves the focus sentinels outside of the dialog element and ensures they have `tabIndex={0}`

francinelucca (commented):
The VO seems to be reading out the focus sentinels to me on safari, is that correct? 

https://github.com/carbon-design-system/carbon/assets/40550942/f35fed04-0722-4a24-95f7-6baafaee4123

guidari:
I'm not able to replicate that @francinelucca 🤔 

I'm in Safari using the VO

https://deploy-preview-14156--carbon-components-react.netlify.app/iframe.html?id=components-composedmodal--with-state-manager&viewMode=story

tay1orjones (approved):
Confirmed JAWS 2023 in firefox does not read focus sentinels 👍

francinelucca (approved):
LGTM in composedModal, I do see the same issue in Modal though, could we port this fix over to Modal? or open another issue for Modal?

guidari:
> LGTM in composedModal, I do see the same issue in Modal though, could we port this fix over to Modal? or open another issue for Modal?

We basically copy a few things from `Modal` to fix `ComposedModal` 😅. 
I tried to port the differences between those to test if would fix `Modal`, but it didn't. It might be worth open an issue to spend more time in it. 

I can open it and take a look

tw15egan (approved):

PR #14169, opened 2023-07-10
fix(tableheader): ensure onClick is called when not sortable

### LINKED ISSUE #14016: [Bug]: TableHeader ignores onClick prop when not sortable
### Package

@carbon/react

### Browser

Firefox

### Package version

v1.30.0

### React version

v17.0.2

### Description

`TableHeader`'s `onClick` prop is used on an internal `<button>` when `isSortable={true}`, but the `onClick` prop is ignored when `isSortable={false}`.

### Reproduction/example

https://stackblitz.com/edit/github-kibieg?file=src%2FApp.jsx

### Steps to reproduce

1. Create a standard table with Carbon components.
2. Add an `onClick` prop to `<TableHeader>` with no `isSortable` prop (defaults `false`).
3. Click on the `<TableHeader>`.

A workaround can be accomplished by using `onMouseDown` instead of `onClick`.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #14016 

#### Changelog

**Changed**

- add `onClick` to `th` when `!isSortable`

#### Testing / Reviewing

- `onClick` should be called when the header is not sortable

### DIFF
--- a/packages/react/src/components/DataTable/TableHeader.tsx
+++ b/packages/react/src/components/DataTable/TableHeader.tsx
@@ -96,7 +96,7 @@ interface TableHeaderProps
   /**
    * Hook that is invoked when the header is clicked
    */
-  onClick?: MouseEventHandler<HTMLButtonElement>;
+  onClick?: MouseEventHandler<HTMLElement>;
 
   /**
    * Specify the scope of this table header. You can find more info about this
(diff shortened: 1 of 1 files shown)


francinelucca (commented):
Is there a way keyboard users would be able to get at this? 🤔

tay1orjones:
@francinelucca good point, this is an invalid approach here

PR #14185, opened 2023-07-11
AVT test for Progress Indicator

### LINKED ISSUE #14121: Add AVT coverage for `ProgressIndicator` states: Hover/Active/Visited state

guidari:
Closes #14121 

Added AVT testing for the Progress Indicator.

I added to the interactive stories. In the defatul there is no navigation to test.

#### Testing / Reviewing

Check if the navigation that is being tested make sense.


packages/react/examples/custom-css-prefix/yarn.lock
```diff
@@ -1,4178 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
-# yarn lockfile v1
-
-
```
francinelucca: is this file meant to be deleted?

e2e/components/ProgressIndicator/ProgressIndicator-test.avt.e2e.js
```diff
@@ -0,0 +1,81 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+const { expect, test } = require('@playwright/test');
+const { visitStory } = require('../../test-utils/storybook');
+
+test.describe('ProgressIndicator', () => {
```
francinelucca, line 13: I think this is still missing a ".ToHaveNoACViolations(...)" check for each of the states: hover, active, visited. So from my understanding you'd perform the keyboard/page operations to get the component to a "hover" or "active" state, and then you'd run an avt assertion against that. I could be misunderstanding though... 😅

francinelucca (changes requested):

e2e/components/ProgressIndicator/ProgressIndicator-test.avt.e2e.js
guidari: @francinelucca At first I tought the same. However I was not sure about it, because there is no keyboard navigation in the default stories, that's why I only tested the interactive story. 

But I think you are right! I'm gonna add those tests.

For example:

Screenshot 2023-07-11 at 17 11 29

```  
test('accessibility-checker on hover @avt', async ({ page }) => {
    await visitStory(page, {
      component: 'ProgressIndicator',
      id: 'components-progressindicator--default',
      globals: {
        theme: 'white',
      },
    });

    await expect(page.getByText('First step')).toBeVisible();

    page.getByText('First step').hover();

    await expect(page).toHaveNoACViolations('ProgressIndicator-onhover');
  });
  ```

packages/react/examples/custom-css-prefix/yarn.lock
guidari: Ohh, it should!

e2e/components/ProgressIndicator/ProgressIndicator-test.avt.e2e.js
francinelucca: yeah I think you can do the same with the interactive story if necessary

guidari: I added the new tests.

I'm checking if the prop `complete` (visited state) and `current` (active) are being added correctly, Because I don't see how to check using "ToHaveNoACViolations" for these two props. Only the hover made sense for me.

If you have an idea, let me know @francinelucca 😁

francinelucca (approved):

tw15egan:
@guidari any reason why a `yarn.lock` file was removed here?

guidari:
> @guidari any reason why a `yarn.lock` file was removed here?

Ohh! Actually I made a mistake while I was working in my other issue about `yarn build`. I should put it back

tw15egan (approved):

PR #14193, opened 2023-07-11
fix(button): add min-height and bottom padding for wrapping

### LINKED ISSUE #13644: Button: text wrap specs
Design exploration and specs needed for #13406 refactor(Button): align icon top for wrap alignment

Dev will most likely implement: 

> Idea 3 (static): find a static distance for top since icons have absolute positioning
> I tried matching top spacing of icon with button padding, but that did not align properly.
> using pixel values for now because padding had calculations (below)
> SIZE = padding-top = CLASS
> sm = calc(0.375rem - 3px) = .cds—btn—sm
> md = calc(0.675rem - 3px) = .cds—btn—md
> lg = calc(0.875rem - 3px) = .cds—btn
> xl = 1rem = .cds--btn--xl:not(.cds--btn--icon-only)
> 2xl = 1rem = .cds--btn--2xl:not(.cds--btn--icon-only)

icon and text alignment get slightly off though and we will need to visually adjust especially for the small button. 
image

**Open UX/Visual questions:**

- Do we need documentation about text wrapping?
- What happens when buttons are close to each other and one wraps and the other doesn't? Can we even control something like this?
- XL button padding is a little strange with more padding on top than bottom. usually, when there is different padding, there is more on the bottom than top.

alisonjoseph:
Closes #13644

#### Changelog

**Changed**

- Update buttons to use min-height instead of height to allow for wrapping
- Add bottom padding equal to top padding

#### Testing / Reviewing

Check that nothing has changed on current button stories, check that buttons in demo wrapping story look correct. 

I'll delete demo story after this is approved.
https://deploy-preview-14193--carbon-components-react.netlify.app/iframe.html?args=&id=components-button--a-size-test&viewMode=story


tw15egan:
@alisonjoseph looks like the changes had some effect on the code snippet. I think we just need to unset the `min-height` so it respects the `height` value on that one. The overflow menu looks identical when comparing, not sure what Percy is complaining about there.

alisonjoseph:
Should be fixed @tw15egan just waiting to make sure Percy passes.

tw15egan (approved):
Looks great 🎉 🔥

francinelucca (approved):
LGTM!

laurenmrice (approved):
This looks good to me, but I also will let @kingtraceyj take a look. I do notice in our design docs/figma we have a Large (expressive) button, which has the `Large` (48px) height, with `$body-compact-02` type token. It should look as follows with wrapping: 

Frame 2 copy

alisonjoseph:
@laurenmrice I need to take another look at the expressive button styles, looks like this fix throws them off slightly. 
Screenshot 2023-07-12 at 12 41 57 PM

alisonjoseph:
@tw15egan yea that bottom padding is calculated via a line height calc, playing around with that.

laurenmrice:
@tw15egan Large is the only size button we have for Expressive.

alisonjoseph:
Added styles for expressive button wrapping too. @laurenmrice @tw15egan

tw15egan (approved):
Looks great!

tay1orjones (approved):
Looks awesome! 🎉

kingtraceyj (approved):
looks good! 💥

PR #13750, opened 2023-05-05
fix(Popover): nested popover - css fix

guidari:
Closes #13684 

When an IconButton is placed within a Popover, the iconDescription is always visible. Popover's *--open classes are not properly scoped to only impact children Popovers one "level" deep.

Repro
https://stackblitz.com/edit/github-b5r8er?file=src/App.jsx

In order to apply the style to only the first child of the CSS class `.#{$prefix}--popover`, we are specifically checking for its direct child

#### Testing / Reviewing

You can find a new component called 'Test Issue' inside `Popover` in Storybook. Checking out this component will allow you to test its functionality. Additionally, going to the `Button` storybook and checking if everything is okay with the 'Icon Button' would also be a good idea.

Once its approved I'll remove the storybook, or the second approver can remove

PR #14200, opened 2023-07-12
fix(Popover): scope popover styles to only influence direct children

### LINKED ISSUE #14195: [Bug]: Popover with interactive trigger causes tooltip layout issue
### Package

@carbon/react

### Browser

Chrome

### Package version

Latest

### React version

Latest

### Description

Hello! :wave:

I am looking to build something very similar to the settings menu found on the [Carbon disclosure pattern guidance page](https://carbondesignsystem.com/patterns/disclosures-pattern/#settings-and-filter-menus). This is for the row settings menu that we have for the Datagrid in Carbon for IBM Products which can be [seen here](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--empty-state). Our current implementation uses the `ToggleTip`. We received some design feedback that it should actually use the Popover so that there is no caret and it is placed directly below the trigger/settings gear icon.

The examples for Popover in the react storybook do not show an interactive element for the Popover trigger, only an icon. After attempting to include an `IconButton` as the `Popover` trigger, I noticed that the placement of the tooltip for the `IconButton` is not quite right.

```jsx
<Popover>
  <IconButton {...props}>
  <PopoverContent>
    content here
  </PopoverContent>
</Popover>
```
image

I was wondering if there might be an example of Popover with an interactive trigger so that it can be opened via keyboard. If not, is there a way to utilize IconButton as the Popover trigger or potentially even the possibility of removing the caret via a prop for the ToggleTip.

Thanks so much!

### Reproduction/example

https://stackblitz.com/edit/github-ndeemc-6aqjyv

### Steps to reproduce

Include an `IconButton` as the Popover trigger element so that it can be opened via keyboard.

### Suggested Severity

### Application/PAL

Carbon for IBM Products

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14195
Refs https://github.com/carbon-design-system/carbon/discussions/13836
Refs https://github.com/carbon-design-system/carbon/pull/13750

This changes all positioning logic (and some tab-tip specific styles) to only target their direct child. This should prevent issues where `IconButton` or other components that use a `Tooltip` or `Popover` aren't affected by the top-level positioning styles. 

#### Changelog

**Changed**

- Instead of using general selectors ( `.#{$prefix}--popover--bottom .#{$prefix}--popover-content`), I changed it so that they all are scoped e.g. `.#{$prefix}--popover--bottom > .#{$prefix}--popover .#{$prefix}--popover-content`

#### Testing / Reviewing

Pull down locally, and try changing the `align` prop of the `IconButton` in the test story to each different alignment, and ensure they render correctly. Also ensure there are no regressions for the existing `Popover` components


guidari (approved):
LGTM! 

I just found that Safari has glitch when it comes to the `drop-shadow` prop. I tried to solve it, but it doesn't look simple, the work around that other devs are saying, would be using `box-shadow`, but I also couldn't make it work 😅

Anyhow, just FYI @tw15egan. The fixes for the PR looks good. 🚀

Screenshot 2023-07-14 at 10 01 02

tw15egan:
@guidari whoa thats super weird, will create a separate issue to track that one. Nice catch 👍🏻

tw15egan:
@tay1orjones when you get a chance this just needs one more review

tay1orjones (approved):
Sorry @tw15egan my bad

PR #14198, opened 2023-07-12
Docs - Added yarn build instructions and new README files

guidari:
Closes #14098

- Added new instructions to make sure that the user have run `yarn build` before trying to start a example project.
- Added README.md files to examples that didn't have it.

### Changelog

Folders with README.md that were edited
  
    - examples/class-prefix/README.md
    - examples/codesandbox/README.md
    - examples/codesandbox-with-sass-compilation
    - examples/id-prefix
    - examples/incremental-migration
    - examples/light-dark-mode
    - examples/nextjs
    - examples/v10-token-compat-in-v11
    - packages/icons-vue/examples/vue-cli
    - packages/react/examples/create-react-app
    - packages/react/examples/next
    - packages/react/examples/react-router

Folders without README.md that were added

    - examples/codesandbox-styles
    - examples/custom-theme
    - examples/vite
    - packages/colors/examples/preview
    - packages/colors/examples/sass-modules
    - packages/grid/examples/css-grid
    - packages/grid/examples/preview
    - packages/icons/examples/preview
    - packages/layout/examples/preview
    - packages/pictograms/examples/preview
    - packages/themes/examples/preview
    - packages/themes/examples/preview-v10
    - packages/themes/examples/sass-modules
    - packages/type/examples/preview

These four examples below I'm getting errors when I  try to run. So I'm trying to fix them to understand if there is something else that should be done before running. But I might open an issue if gets too much time to fix it

    - packages/react/examples/codesandbox - ERROR
    - packages/react/examples/custom-css-prefix - ERROR
    - packages/react/examples/custom-data-table-state-manager - ERROR
    - packages/react/examples/drag-and-drop-file-uploader - ERROR

#### Testing / Reviewing

Check spelling and build instructions

Issue #13637, opened 2023-04-20
Migrate examples off of Create React App


tay1orjones:
Create React App project is [moving in a new direction](https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741), and there are quite a few examples within the repo using it or `react-scripts`. I think it makes sense to move off of it to avoid future hassle with outdated dependencies and errant dependabot vulnerability errors.

#### What should we move to?

We should consider the recommended options outlined by [the new react docs](https://react.dev/learn/start-a-new-react-project). This is where the majority of new react-based projects will begin their journey with Carbon and we should have examples that follow suit. Two in particular stand out

1. Vite - great for isolated and small/lightweight examples showcasing component prop configuration or usage 
2. NextJS - suitable for more complicated examples that involve needs provided by a framework - routing, SSR, etc.

Quite a few of the examples in the repo are old and outdated. In some cases they're using react-scripts v3 and causing errant dependabot vulnerability alerts. This includes the following directories:

- https://github.com/carbon-design-system/carbon/tree/main/examples
- https://github.com/carbon-design-system/carbon/tree/main/packages/react/examples

### Goals

- Remove examples that no longer provide value
  - Could we gather install stats for these to determine ones that aren't used much? Install may not be the perfect metric, sometimes examples are just used as a reference
- Update remaining examples to either vite or nextjs
- Examples should be able to be run: 
  - locally
  - on stackblitz
  - preferably on codesandbox as well though this may not be feasible due to their lack of scss support

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/14599
- [ ] https://github.com/carbon-design-system/carbon/issues/14771
- [x] https://github.com/carbon-design-system/carbon/issues/14462
- [ ] https://github.com/carbon-design-system/carbon/issues/14768
- [ ] https://github.com/carbon-design-system/carbon/issues/14769
- [ ] https://github.com/carbon-design-system/carbon/issues/15201
- [ ] https://github.com/carbon-design-system/carbon/issues/15202
- [ ] https://github.com/carbon-design-system/carbon/issues/15203
```

falsepopsky:
[...]

PR #14219, opened 2023-07-13
chore: updated next and react version in example

guidari:
After working on this [PR](https://github.com/carbon-design-system/carbon/pull/14198) it was found a few errors when running an example using NextJS.

To fix that it was updated the next and react versions in the `package.json` for each example.

**Changed**

- next and react version in `package.json` examples

#### Testing / Reviewing

Pull the branch locally and run the examples


tay1orjones (changes requested):
Elements previews look good and the light-dark-mode example looks good.

The incremental-migration one is not working for me though
image

guidari:
> The incremental-migration one is not working for me though

Hey @tay1orjones I didn't find a way to fix that on NextJS. The example works outside of the monorepo, but not inside it, for some reason.

But I could fixed it by migrating the example to use Vite instead of NextJS. Check this [Draft PR](https://github.com/carbon-design-system/carbon/pull/14317)

My suggestion is to use that migration as an example to this [issue](https://github.com/carbon-design-system/carbon/issues/13637), about how we can approach this transition.

tay1orjones (approved):
Sounds good, thanks @guidari!

alisonjoseph (approved):

PR #14250, opened 2023-07-19
[AVT]: Added avt tests for OverflowMenu

### LINKED ISSUE #14212: Add AVT coverage for `OverflowMenu` states: Keyboard nav

guidari:
Closes #14212

Added avt tests to variants in the stories and keyboard nav testing 

#### Testing / Reviewing

- CI should pass
- Ensure the tests written make sense to the user keyboard navigation


e2e/components/OverflowMenu/OverflowMenu-test.avt.e2e.js
```diff
@@ -0,0 +1,80 @@
...
+    // Tab and open the toggle button for the OverflowMenu
+    await page.keyboard.press('Tab');
+    const toggleButton = page.getByRole('button');
+    await expect(toggleButton).toBeFocused();
+    await page.keyboard.press('Enter');
+    await expect(toggleButton).toHaveClass(/cds--overflow-menu--open/);
+    // Navigation inside the menu
+    await expect(page.getByTestId('stop-app')).toBeFocused();
+    await page.keyboard.press('ArrowDown');
+    await expect(page.getByTestId('restart-app')).toBeFocused();
```
tay1orjones: Could we get these instead via role like this: `page.getByRole('button', {name: 'Stop app'})`? This way we could remove the reliance on the testid in the story

tay1orjones (commented):
One minor suggestion

e2e/components/OverflowMenu/OverflowMenu-test.avt.e2e.js
guidari: Hey @tay1orjones 
It doesn't work. Probably because it has a `div` inside the button.

```
<button class="cds--overflow-menu-options__btn" role="menuitem" tabindex="-1" data-testid="stop-app">
    <div class="cds--overflow-menu-options__option-content">Stop app</div>
</button>
```
 
I also tried another ways to get this button for example: 
`await expect(page.getByText('Stop app')).toBeFocused();`

But the only way it works was with the data-testid. And I agree with you, it doesn't look good to have that `data-testid` in the story... 
I'm gonna try to spend a little bit more time on it, to see if I find a solution that doesn't rely on using `data`testid`

e2e/components/OverflowMenu/OverflowMenu-test.avt.e2e.js
```diff
@@ -0,0 +1,102 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+const { expect, test } = require('@playwright/test');
+const { visitStory } = require('../../test-utils/storybook');
+
+test.describe('OverflowMenu', () => {
+  test('accessibility-checker @avt', async ({ page }) => {
```
tay1orjones: Minor thing, but you can put the `@avt` flag in the describe block and then you don't have to add it to every test name. As this file grows could be helpful to ensure everything in the block will get picked up by the test runner.

```suggestion
test.describe('OverflowMenu @avt', () => {
  test('accessibility-checker', async ({ page }) => {
```

guidari: Cool! Fixed!

Also just found a way to filter down descendant elements
```
 await expect(
      page.locator('button').filter({ hasText: 'Stop app' })
    ).toBeFocused();
```

e2e/components/OverflowMenu/OverflowMenu-test.avt.e2e.js
```diff
@@ -0,0 +1,104 @@
...
+      },
+    });
+    await expect(page.getByRole('button')).toBeVisible();
+
+    // Tab and open the toggle button for the OverflowMenu
+    await page.keyboard.press('Tab');
+    const toggleButton = page.getByRole('button');
+    await expect(toggleButton).toBeFocused();
+    await page.keyboard.press('Enter');
+    await expect(toggleButton).toHaveClass(/cds--overflow-menu--open/);
```
andreancardona: @tay1orjones do we want this to be `toBeVisible` instead of `toHaveClass`?

guidari: I think you are right @andreancardona 
It might be better to check if the menu is visible instead of checking for the `className`, because the CSS can be edited. I'm going to change that!

andreancardona: Yep that's what I was chatting with Taylor about - thanks :)

alisonjoseph (approved):

andreancardona (approved):
looks good!

tay1orjones (approved):

PR #14251, opened 2023-07-19
[BUG]: tabbing hidden elements on SideNav

### LINKED ISSUE #14151: [Bug]: UI Shell tabbing hidden elements in the UI
### Package

@carbon/react

### Browser

_No response_

### Package version

1.32.0

### React version

18.2

### Description

In the `SideNav` component there are a few things to improve:

- Replicate the behaivor of the `Switcher`, which is **NOT** open the `SideNav` on focus, but keep closing when loses focus
- If the first fix is done, probably it won't be necessary the prop `onSideNavBlur`
- If the prop `onSideNavBlur` needs to be kept, it will be necessary to fix the conflict between it and the `isFixedNav `

Slack conversation: https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1688406261790019

### Reproduction/example

https://stackblitz.com/edit/github-mpwg6m?file=src%2FApp.jsx

### Steps to reproduce

Tab the page with the `SideNav` collapse when the prop `isFixedNav` is active and you will see that it is possible to tab the hidden items.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

guidari:
Closes #14151

Fixed the hidden tab navigation when the `SideNav` is collapsed and also removed the open on focus from the `SideNav`. 
The `SideNav` only open on focus when the `isRail` prop is actived to keep the navigation going (that wasn't changed).

When the `isFixedNav` prop is being used the `SideNav` won't be dismissible by losing focus.

#### Changelog

**New**

- Added the `inert` attribute to avoid the children elements to be tabbable.
- Added `tabIndex` to avoid the tab action when the `SideNav` is collapsed.
- Added `tabIndex` attribute as a prop in `SideNavLink` and `SideNavMenu` to avoid break changes in case the current user is using that attribute.

**Changed**

- Now we are propagating `isRail` and `isSideNavExpanded` even if the `SideNav` doesn't contain the `isRail` prop.

**Removed**

- SideNav opening on focus.

#### Testing / Reviewing

- Go to the SideNav Rail w/ header variant and make sure the `isFixedNav` works fine and `isRail` also (can't be used both at the same time).
- Validate other variants that uses `SideNav` component.


packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
```diff
@@ -437,7 +443,9 @@ export const HeaderWSideNav = () => (
             aria-label="Side navigation"
             expanded={isSideNavExpanded}
             onSideNavBlur={onClickSideNavExpand}
-            href="#main-content">
+            href="#main-content"
+            inert={isSideNavExpanded ? false : -1} // With the `inert` prop you can disable the navigation from all children elements
+          >
```
tay1orjones: If `isSideNavExpanded` is already given to SideNav via `expanded`, would it make more sense for `inert` to be conditionally applied _within_ SideNav? Otherwise we're leaving this up to the consumer to properly implement, which is unlikely

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -7028,6 +7028,9 @@ Map {
       "className": Object {
         "type": "string",
       },
+      "isRail": Object {
+        "type": "bool",
+      },
```
tay1orjones: With three more components needing `isRail`, is it worth considering using context to store the value of `isRail` and using that context to get the value within the components that need it? Otherwise we'll be relying on consumers to pass `isRail` to these components, right? I think it's unlikely that we'll see folks properly configuring this if that is the case.

tay1orjones (changes requested):

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
guidari: Good call! I've just moved the isRail prop from being spread through the children elements and added a Context to the `SideNav` @tay1orjones

packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
guidari: Defenitily! Idk why I didn't do that before 🤔

tay1orjones (approved):

tay1orjones:
Thanks for making those changes!

alisonjoseph (approved):

PR #14254, opened 2023-07-20
[BUG]: Added onClick in the StructuredListRow

### LINKED ISSUE #14037: [Bug]: onChange event handler not called on StructuredListInput when selecting an entry in a structured list
### Package

@carbon/react

### Browser

Chrome

### Package version

1.31.3

### React version

16.10.0

### Description

See the following example on StackBlitz: https://stackblitz.com/edit/github-xfrb3k?file=src%2FApp.jsx

When the single entry in the structured list is selected, the onChange event handler is not called.

### Reproduction/example

https://stackblitz.com/edit/github-xfrb3k?file=src%2FApp.jsx

### Steps to reproduce

Click the "Text for the first entry" list entry to select it.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

guidari:
Closes #14037

After discussing with @tay1orjones, we decided to keep the `onChange` behavior and add an `onClick` prop to `StructuredListRow` so that users don't have to wrap StructuredList to call an `onClick` event.

_After approved, I'm going to delete the function in the story._

**New**

- Added `onClick` to the StructuredListRow component. 

#### Testing / Reviewing

- Go to the Selection variant for the StructeredList;
- Click on the row and check the console.


packages/react/src/components/StructuredList/StructuredList.js
```diff
@@ -182,7 +182,13 @@ export function StructuredListRow(props) {
       {...other}
       role="row"
       className={classes}
-      onClick={() => setSelectedRow(id)}
+      onClick={(event) => {
+        setSelectedRow(id);
+        // Avoid that the onClick gets called when arrow down
+        if (!event.target.className.includes('cds--structured-list-input')) {
+          onClick && onClick();
+        }
```
tay1orjones: @guidari Could you describe more of what this is intending to avoid?

guidari: @tay1orjones  When the user presses the ArrowDown key, the `onClick` method is called because the input element is considered to be "clicked." This code allows us to split the `onClick` and `onKeyDown` events into different behaviors.

tay1orjones: Ah interesting, I forgot arrow navigation selected the input. Relying on the classname of a different component feels like we're not maintaining a clear separation of concerns between these two components. I know they're related, but it still feels funky.

Would it be okay if we just passed this all the way through to the consumer no matter if it's on the input or not and then they could do this check themselves instead if they need to?

guidari: @tay1orjones You mean that we should let to the consumer to do this check instead of us? Sure, we can do that.

I just feel worried about the `onClick` being triggred when the user arrow down, maybe the user would think the same. But if we remove the checker should we point it out in docs, or just let it be?

tay1orjones: @guidari Wow, sorry for such a delayed response! Yes I think it's okay to leave it as is and let folks filter out the events they don't need

guidari: Cool! I just removed @tay1orjones

tay1orjones (approved):
LGTM!  👍

andreancardona (approved):
LGTM!

PR #14275, opened 2023-07-20
fix(DataTable): fix stacking issues with table toolbar + batch actions

### LINKED ISSUE #14137: [Bug]: tooltips render under icons
### Package

@carbon/styles

### Browser

_No response_

### Package version

?

### React version

_No response_

### Description

Reported in slack https://ibm-studios.slack.com/archives/C0M053VPT/p1687360867441669

image

We’re having a weird layering issue where tooltips are appearing under icons that are in a data table.

Two issues in the example:
tooltip text can only displayed in one line.
icon in table header displayed over tooltip for action button

workaround is to add a z-index to `.cds--batch-action `

### Reproduction/example

https://codesandbox.io/p/sandbox/optimistic-river-sd3q2p?file=%2Fsrc%2Fapp%2Fapp.component.html%3A28%2C59

### Steps to reproduce

See codesandbox

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14137

Adds some `z-index` values so that the batch actions and toolbars render over table content. Also tweaked the `clip-path` a bit so that longer tooltips can be rendered. This very slightly adjusts how much the toolbar slides down when fading in/out, but seems like a good compromise. Also added `opacity` transition to make this less noticeable.

#### Changelog

**New**

- `z-index` added to batch action bar, table toolbar

**Changed**

- tweaked `clip-path` to allow longer tooltips in batch action bar

#### Testing / Reviewing

Go to the batch actions story, and add a much longer string for the tooltips that appear. Ensure the animations look correct and there are no regressions

### DIFF
--- a/packages/styles/scss/components/data-table/action/_data-table-action.scss
+++ b/packages/styles/scss/components/data-table/action/_data-table-action.scss
@@ -27,6 +27,7 @@
   .#{$prefix}--table-toolbar {
     // Need for batch actions
     position: relative;
+    z-index: 1;
     display: flex;
     width: 100%;
     min-height: $spacing-09;
@@ -388,6 +389,7 @@
     justify-content: space-between;
     background-color: $background-brand;
     clip-path: polygon(0 0, 100% 0, 100% 0, 0 0);
+    opacity: 0;
     pointer-events: none;
     transform: translate3d(0, 48px, 0);
     transition: transform $duration-fast-02 motion(standard, productive),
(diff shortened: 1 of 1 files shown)


alisonjoseph (approved):
LGTM! Tagging design for a review because the slight animation change, I compared it with the live storybook and it seems good.

thyhmdo (approved):
@tw15egan I couldn't tell from interacting with it, thanks for the videos! I like this fix

tay1orjones (approved):

PR #14276, opened 2023-07-20
chore(downshift): update downshift to v8.x

### LINKED ISSUE #13749: Update downshift
We're two major versions behind on our downshift dependency. We attempted to update to v6 [a few times before](https://github.com/carbon-design-system/carbon/pulls?q=is%3Apr+update+downshift+is%3Aclosed) and hit major roadblocks - https://github.com/carbon-design-system/carbon/pull/9839, most recently https://github.com/carbon-design-system/carbon/pull/11331.  

Primary drivers for this update include:
- Ensure we avoid any future CVEs related to outdated transient dependencies of v5.x 
- There are features and fixes included in the update that will resolve some issues we have

```[tasklist]
### Related issues blocked by downshift being outdated
- [ ] https://github.com/carbon-design-system/carbon/issues/10634
- [ ] https://github.com/carbon-design-system/carbon/issues/14102
```

Something else related to this is an idea [mentioned here](https://github.com/carbon-design-system/carbon/pull/8211#issuecomment-809459238), to expose our ListBox internals https://github.com/carbon-design-system/carbon/issues/6177 so consumers can directly use them with downshift. This could be part of the path to upgrading downshift and/or deprecating existing parts of our API. I'm not 100% sure, but it's something to consider and explore.

### LINKED ISSUE #14102: Update downshift: research sprint
Acceptance Criteria:

- Attempt to update to latest.
- Review all 4 components for defects resulting from the update. (may not be possible in one sprint, might need to defer this into another sprint)
- Collect and categorize the defects, generate a plan of action to address everything.
- Playback findings to the dev team if needed.
- Create new issues or tasklists for all the items, we probably won't be able to incrementally deliver this. It'll be a draft PR that lives for a few sprints.

tay1orjones:
Closes #13749 
Closes #14102 

#### Changelog

**Changed**

- Modify downshift usages to reflect the changes made across v6, v7, and v8 migrations
  - v6 https://github.com/downshift-js/downshift/releases/tag/v6.0.0
  - v7 https://github.com/downshift-js/downshift/blob/master/src/hooks/MIGRATION_V7.md
  - v8 https://github.com/downshift-js/downshift/blob/master/src/hooks/MIGRATION_V8.md
- Update `Dropdown` avt tests to reflect changes in how downshift does not apply `aria-selected` to items as you arrow through them

#### Testing / Reviewing
- Thoroughly review the following _stable_ components:
  - ComboBox
  - Dropdown
  - MultiSelect
  - FilterableMultiSelect
- Thoroughly review the following _unstable_ components:
  - FluidComboBox
  - FluidDropdown
  - FluidMultiSelect

For each of the above components:
- [ ] Ensure no visual/styling regressions
- [ ] Ensure no UX/behavior regressions
  - [ ] Click behavior
  - [ ] Keyboard navigation behavior
- [ ] Ensure no voiceover regressions
- [ ] Ensure no a11y violation regressions


tw15egan:
Going to mark this for ready for review to get some more eyes on this to help determine if there are any regressions @tay1orjones and I have missed. 

We need to verify the following:
- No voiceover regressions
- No visual regressions
- No a11y violation regressions

tay1orjones:
@tw15egan I've updated the PR body to reflect the current state and what needs tested

guidari (commented):
Hey @tay1orjones @tw15egan I tested using Chrome, Firefox, Safari and side by side with [currently Storybook](https://react.carbondesignsystem.com/?path=/docs/getting-started-welcome--welcome) and the [PR Preview](https://deploy-preview-14276--carbon-components-react.netlify.app/?path=/docs/getting-started-welcome--welcome)

I categorized the findings into two groups: things that _**used to work**_ and things that _**didn't work before and it still not working.**_ 

**Used to work:**
- _DropDown Default_
is not reading the elements in Safari with VO  

- _MultiSelect Default / With Initial Selected Items / With Layer_
Not working properly when we select an item using keyboard navigation 
When selecting an option it works, but the focus comes back to the top, reseting the keyboard navigation.

- _FluidDropDown Default_
When using the VO and arrow through the options, it doesn’t read correctly. (Only in Chrome)

- _FluidMultiSelected Default / Condensed_
Not working properly when we select an item using keyboard navigation
When selecting an option it works, but the focus comes back to the top, reseting the keyboard navigation.
________________

**It didn't work before and it still not working.** 

- _FluidComboBox Default_
in Safari does’t read the options with VO when we arrow through the options.
________________

  Extras things that I noticed, but not related to the DownShift changes. Maybe is an intended behavior, I’m just curious.   👀
  
  In the Multiselect stories, when the variant is Filterable we are not able to select with Space key because the user is able type on input. So we have to use the Enter key. But the other variants we can only select using the Space key and not Enter. Is that an intended behavior?

tw15egan:
Thanks for the in-depth review @guidari! 

>  In the Multiselect stories, when the variant is Filterable we are not able to select with Space key because the user is able type on input. So we have to use the Enter key. But the other variants we can only select using the Space key and not Enter. Is that an intended behavior?

Yeah, we noticed that as well. I'd like to get some consensus on which buttons should allow an option to be selected. To me, I think this makes the most sense:

<kbd>Space</kbd> and <kbd>Enter</kbd> 
- `Dropdown`
- `Multiselect`

<kbd>Enter</kbd> 
- `FilterableMultiselect`
- `Combobox`


I also think they should _all_ open with <kbd>Space</kbd>, <kbd>Enter</kbd>, and <kbd>ArrowDown</kbd>. Does this seem like a good pattern to adhere to for the 4 Downshift components?

guidari:
> I also think they should _all_ open with Space, Enter, and ArrowDown. Does this seem like a good pattern to adhere to for the 4 Downshift components?

I agree on that also! There were a few components that wouldn't open on `Enter` or could be open on `Enter` only in FireFox

laurenmrice (changes requested):
@tay1orjones The only visual changes I found were with how Focus appears. I am unsure if this is supposed to happen with the downshift changes or if its actually a bug. 😬

**Dropdown and Multiselect (default styles)**

When clicking the dropdown field to open the list of options, our current storybook has the focus go around the whole dropdown list with a 1px stroke, and additionally, you can focus on each option in the list with a 2px stroke. The preview in this PR only has the focus on the dropdown field and on each option in the list with a 2px stroke and does not have a 1px focus around the whole dropdown list.

Dropdown (default style)
Multiselect  (default style)

------

**Dropdown (fluid style)**
When clicking the dropdown field to open the list of options, our current storybook has the focus go around the whole dropdown list with a 1px stroke, and additionally, you can focus on each option in the list with a 2px stroke.  The preview in this PR only focuses on each option in the open list, with no focus appearing on the whole dropdown list or the field.

Dropdown (fluid style)

------

**Multiselect (fluid style)**
When clicking the multiselect field to open the list of options, our current storybook has the focus go around the whole dropdown list with a 1px stroke, and additionally, you can focus on each option in the list with a 2px stroke. The preview in this PR only has the focus on the multiselect field and on each option in the list with a 2px stroke and does not have a 1px focus around the whole dropdown list.

Multiselect (fluid style)

tw15egan:
@laurenmrice I believe this is part of the new [1.2 combo box guidance](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/), and came straight from the downshift team. Will defer to the a11y experts, perhaps a good talking point at the next CAG call. I will say I personally find the new focus styles less "in your face". The double focus borders on the item/menu itself have always looked a bit odd to me

laurenmrice:
@tw15egan okay cool, thanks for clarifying!

laurenmrice (approved):
Going to approve. I don't see any other visual changes other than the focus states mentioned above, which were intentional changes.

mbgower:
@laurenmrice in regard to your [comments](https://github.com/carbon-design-system/carbon/pull/14276#pullrequestreview-1548312699) on the new focus indicators...

In some ways, it is a bit cleaner NOT having the outline around the whole list (essentially creating 3 focus indicators). There are likely some slight losses from a low vision perspective as they 'lose' a clear focus border to the dropdown. But my gut tells me that the focus indicator around both the input and the child with focus should be sufficient for them. As well, there is now a grey stroke around the whole list, which wasn't there before, which will help.

So I think the newer implementation is fine.

mbgower:
> I also think they should all open with Space, Enter, and ArrowDown. Does this seem like a good pattern to adhere to for the 4 Downshift components?

I'm a little leery of saying Space should trigger that. The main reason is that in combobox or any filtering situation , pressing Space would add a space at the start of the input field (or potentially even wipe out the existing value).
Enter does not traditionally open a select, but I can't think of any reason it _shouldn't_.

tw15egan:
@mbgower  `Combobox` and `FilterableMultiselect` will open on any keypress that triggers an input value change, so it is not a special use case for `Space` in particular, but rather just starting to filter the items. This is the current behavior and has not been changed.

mbgower:
Thanks for the clarification, @tw15egan I should have actually looked at it instead of going by memory! 

I just had a look and there are a couple of considerations here...

Space  _does_ clear the value in the combobox, but the selected items remains selected in the dropdown, which at first I thought would be fine. But I do see a curiosity at least with the current implementation, which is that with Space the user for a moment seems like they've lost the prior value.

To replicate

1. Go to https://react.carbondesignsystem.com/?path=/story/components-combobox--default#storybook-preview-wrapper 
2. set the combobox to an option, (say "option 1")
3. Shift+Tab away from the dropdown so that your option appears in the combobox
4. Tab back to it. The whole selected value will be selected as the combo takes focus

image

5. Press Space to open the dropdown. The existing value is cleared in the input, but it is still selected in the dropdown

image

Now comes the problem...

6. Press Esc to cancel your changes to the combobox. It collapses the combobox, but it seems to have cleared the value, rather than setting it back to "option 1" -- despite the fact it  that was currently selected. 
image

7. Tab away from it, however, and the old value magically reappears! 

https://github.com/carbon-design-system/carbon/assets/14298245/dd2d99be-c9a7-4da4-9717-bd8b13eb7ea8

guidari (approved):

tay1orjones:
Once this is in, I'd like to immediately cut a new prerelease so we can test this with some teams to ensure compatibility.

pwolfert:
@tw15egan, I've been tracking Carbon's implementation of Downshift for the past week because I'm having trouble getting Downshift to work with Safari + VO in my own design system. Yours was working with VO in the previous version (using v5.2 of Downshift), and I'm guessing it works after this PR as well that upgrades to v8. We're on v7 and have had to fill in the missing VO support with hacks. Those hacks have bitten us. Right now my plan is switch to [Adobe react-aria's implementation](https://react-spectrum.adobe.com/react-aria/useSelect.html), because I've given up.

I saw this note above:

> *DropDown Default*
> is not reading the elements in Safari with VO  

Do you have a moment to explain how you got it to work?

pwolfert:
I should probably create an issue on Downshift's repo, since it doesn't work in their own examples with VoiceOver. But I imagine it wouldn't be fixed by the time I need to fix it in my own code.

PR #14287, opened 2023-07-23
fix(DataTableSkeleton): fix TypeScript interface and proptypes

dbrugger:
`DataTableSkeleton` is not usable in TypeScript. I think the `headers` PropType was also wrong.

#### Changelog

**Changed**

- Use props in the component
- Use values from default prop type as default values for parameters
- If headers are specified, require the `header` value which is displayed

**Removed**

- default prop types

#### Testing / Reviewing

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -2444,14 +2444,6 @@ Map {
     ],
   },
   "DataTableSkeleton" => Object {
-    "defaultProps": Object {
-      "columnCount": 5,
-      "compact": false,
-      "rowCount": 5,
-      "showHeader": true,
-      "showToolbar": true,
-      "zebra": false,
-    },
     "propTypes": Object {
       "className": Object {
         "type": "string",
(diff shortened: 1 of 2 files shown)


dbrugger:
@tay1orjones @tw15egan can you have a look at the snapshot:
```
            "headers": Object {
              "args": Array [
    -           Array [
                Object {
    -               "type": "array",
    -             },
    -             Object {
                  "args": Array [
                    Object {
    -                   "key": Object {
    +                 "header": Object {
    +                   "isRequired": true,
                        "type": "string",
                      },
                    },
                  ],
    +             "isRequired": true,
                  "type": "shape",
                },
              ],
    -         ],
    -         "type": "oneOfType",
    +         "type": "arrayOf",
```
is that okay? I was adjusting the PropType until it matched the ts interface.

tay1orjones:
@dbrugger could you update the snapshot via `yarn test -u` in the root and push the changes up to this PR? We'll need that in order to get status checks passing.

Also, typescript types are not bound by semver (yet). The proptypes are bound by semver though and need to be treated like the source of truth. Any changes to proptypes must be backwards compatible.

dbrugger:
thanks @tay1orjones ! pushed the snapshot update. I wonder if the proptypes were correct earlier. When I added the typescript interface, it was able to compare the interface, the code and the proptype and it didn't match. It also did not look correct to me, as it should be just an array of objects with a `header` property, not either an array or an object with a property `key`.
I compared it with the `DataTable` (https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/DataTable.tsx#L281) and it seems logical that it should be very similar. By doing that I also changed the `header` property to be a `ReactNode` instead of just a string.

tw15egan (approved):

tay1orjones (approved):
Thanks for taking a second look and pushing up the snapshot. Here's a vite that confirms the existing proptypes were wrong: https://stackblitz.com/edit/github-pmkg5i?file=src%2FApp.jsx

What you've got here looks great. Thanks again!

PR #14297, opened 2023-07-24
feat(slider): add range slider

### LINKED ISSUE #6337: [Slider] Two Handles Defining the Range 🏖️
### Summary

It would be of great if the slider component would support two handles to enable the user to define a range instead of just one value.

If the developer enabled it, the user has two handles instead of one to control two numbers, presumedly minimum and maximum (to define a range).

### Available extra resources

This is an example created by the WH team

image

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/14050
- [ ] https://github.com/carbon-design-system/carbon/issues/14567
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/725
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3730
- [ ] https://github.com/carbon-design-system/carbon/issues/14549
```

### LINKED ISSUE #14567: [Slider] Two Handles Defining the Range: Implementation

m4olivei:
Closes https://github.com/carbon-design-system/carbon/issues/14567

Enhancement to the current React Slider component, allowing for a range selection of two  values.

Requirements: https://ibm.ent.box.com/notes/1252927738733?s=4s8h59gazpuxv0fyxb1xq1pcxx6qvcwl
Designs: https://www.figma.com/file/Jk3pWNBG6ZWIRoldq2tM1x/Range---Slider-Enhancement?type=design&node-id=425-22406&mode=design&t=7ah75PWKFPtNMLUj-0

Mostly done and ready for review. Things on my radar still to do:

- [x] ~~Rewrite Slider component to hooks~~, Will happen in a follow up PR
- [x] Write tests against the two handle functionality
- [x] How to internationalize announcement text for when a value is auto-corrected
- [x] e2e tests
- [x] Design review

#### Changelog

**New**

- `ariaLabelInputUpper` optional prop for the Slider component to use as the `aria-label` attribute for the upper input
- `nameUpper` optional prop for the Slider component that serves as the `name` attribute for the upper input
- `valueUpper` optional prop for the Slider component that serves as the default value for the upper value of the range
- `twoHandles` prop for the SliderSkeleton component which enhances the SliderSkeleton to be a range slider skeleton

**Changed**

- When the new `valueUpper` prop is set, all of `value`, `name` and `ariaLabelInput` props are applied to the lower bound.
- The `onBlur` callback function will provide a `handlePosition: HandlePosition` key in it's `data` argument to indicate which input / handle was changed to the given `value`.
- The `onChange` callback function will provide `value` and `valueUpper` for when the Slider has two handles. `value` refers to the lower handle value when the Slider has two handles.
- The `onRelease` callback function will provide `value` and `valueUpper` for when the Slider has two handles. `value` refers to the lower handle value when the Slider has two handles.

#### Testing / Reviewing

Extensive changes were made to the Slider component to accomodate this feature. Thourough regression testing should be done for the Slider with a single handle.

- [ ] Using the [Slider > Default](https://deploy-preview-14297--v11-carbon-react.netlify.app/?path=/story/components-slider--default) story, regression test comparing to the [current version](https://react.carbondesignsystem.com/?path=/story/components-slider--default).
- [ ] Visit the [Slider > Two Handle Slider](https://deploy-preview-14297--v11-carbon-react.netlify.app/?path=/story/components-slider--two-handle-slider) story.
- [ ] Test dragging each handle. In particular:
  - [ ] The handle nearest to the users click/touch on the track should activate and be the handle manipulated
  - [ ] The corresponding input should react, updating the value as the user drags
  - [ ] The user should not be allowed to go beyond the opposite handle
- [ ] Manipulate the values of the inputs manually or via keyboard arrow keys. Values should update and corresponding handles should update appropriately
- [ ] Test keyboard navigation and focus states.
  - [ ] The user should be able to tab focus on each handle
  - [ ] When focused on a handle the arrow keys should work to manipulate the handle
  - [ ] Holding shift and using the arrow keys should apply a multiplier to move the handle by a greater amount each step
- [ ] Test with a screen reader 
  - [ ] The screen reader should read each handle as a slider and speak the current value as the handle is moved just like the current single handle Slider.

The [Slider > Playground story](https://deploy-preview-14297--v11-carbon-react.netlify.app/?path=/story/components-slider--playground) is the best place for testing various Slider configurations (single and two handle variations).

- [ ] Set a value for `valueUpper`. This effectively puts the slider in two handle mode.
- [ ] Turn on `readOnly`. Slider should hide the handles and show the inputs as Read Only. Compare the visual style of the inputs to the [Text Input component](https://deploy-preview-14297--v11-carbon-react.netlify.app/?path=/story/components-textinput--playground&args=readOnly:!true).

Finally, the Slider Skeleton component has also been updated to support a `twoHandles` boolean prop.

- [ ] Visit [Slider > Two Handle Skeleton](https://deploy-preview-14297--v11-carbon-react.netlify.app/?path=/story/components-slider--two-handle-skeleton)
- [ ] Should show the skeleton with the range handles


tay1orjones (commented):
This is awesome! Playing around with it some I see a few nits but they're things already on your list to address (bound events).

If you think it would be helpful, this component is due to be converted from a class-based component to a functional component. If you're interested, this could be a big improvement to readability of the component being able to encapsulate some of the logic into `useEffect` and `useLayoutEffect` blocks. No sweat if not, but thought I'd mention it.

m4olivei:
Thanks!

> If you think it would be helpful, this component is due to be converted from a class-based component to a functional component. If you're interested, this could be a big improvement to readability of the component being able to encapsulate some of the logic into useEffect and useLayoutEffect blocks. No sweat if not, but thought I'd mention it.

I'm glad you brought that up! I wondered if you all had any such initiative. I'll try to include that as well here as time allows.

packages/styles/scss/components/slider/_slider.scss
```diff
@@ -117,10 +117,125 @@
...
+
+      &::after {
+        background-color: $interactive;
+      }
+    }
+
+    &:focus {
+      width: 16px;
+      justify-content: space-around;
+      border: 1px solid $interactive;
```
jkaeser: Adding this border in causes the control/thumb to jump ever so slightly when grabbing it. It's not a huge deal, but replacing it with an `outline` seems like a free micro-improvement.

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -478,9 +628,29 @@ export default class Slider extends PureComponent<SliderProps> {
...
+      );
+      const distanceToUpper = this.calcDistanceToHandle(
+        HandlePosition.UPPER,
+        clientX
+      );
+      if (distanceToLower <= distanceToUpper) {
+        activeHandle = HandlePosition.LOWER;
+      } else {
+        activeHandle = HandlePosition.UPPER;
+      }
```
jkaeser, line 732: Could this be the culprit for the unexpectedly active thumb? Reset to whatever the current active handle is when `distanceToLower === distanceToUpper`?

jkaeser (commented):
Coming along superbly, @m4olivei!

I noticed an issue when you bring the upper thumb into contact with the lower thumb, then try to move the upper thumb again. Despite clicking on the upper and it maintaining the same focus state, it appears you actually get control over the lower. See video: https://github.com/carbon-design-system/carbon/assets/12532727/e0ead19b-4839-4e08-b578-d3f513518f4f

KevinCamelo:
Hey! Love what you did here @m4olivei. Great work! 

A couple of things that came to mind:

- I appreciate that you removed the option for error states and just have the slider bind to the nearest value or the minimum or maximum. I like that approach a lot more and begs the question when error states would even appear in this case, similar to the base slider component.
- I noticed that the keyboard commands don't work when clicking on the slider itself. When a user clicks `Shift + → ` increase by 10 and when they click `shift + ←`  decrease by 10. And when they focus in and press `→ or ←` it should increase or decrease by 1. These interactions are present in the base slider component and should just carry over. 
- In Gray 100 and Gray 90, the color of the slider does not change to `$focus` on selection or focus.

WONDERFUL JOB! Love seeing this implemented in code!

packages/react/src/components/Slider/Slider.tsx
m4olivei: Nice catch. Turns out, you were close, and this was really helpful to quickly resolve it. The issue was that when the handles are pointing at the same value, `this.state.lowerLeft === this.state.lowerRight`, which meant that the `calcDistanceToHandle` always returned exactly the same value for both handles. In https://github.com/carbon-design-system/carbon/pull/14297/commits/3b3a7a065f7ce19d9f0102c5433211d0f64929d7 I offset the lower handle by it's width which more accurately places it for the purposes of the calculation, and we get the result we want. At least is was working in my testing locally 🤞 .

packages/styles/scss/components/slider/_slider.scss
m4olivei: _I think_ the jump that you see is actually the handle being positioned to its new value. Note that when dragging, the handle anchors to the cursor on the inner edge, which is the edge that points to the value set. This is unique from the default slider where the value is pointed at by the center of the handle, thus the handle anchors to the center or your cursor on drag.

Unless I'm misunderstanding the thing your seeing, the behavior is by design.

m4olivei:
> Coming along superbly, @m4olivei!

Thanks @jkaeser !

> I noticed an issue when you bring the upper thumb into contact with the lower thumb, then try to move the upper thumb again. Despite clicking on the upper and it maintaining the same focus state, it appears you actually get control over the lower. See video: https://github.com/carbon-design-system/carbon/assets/12532727/e0ead19b-4839-4e08-b578-d3f513518f4f

This should be resolved now. Thanks for catching that!

packages/styles/scss/components/slider/_slider.scss
jkaeser: Ah, I see. Thanks for the explanation!

m4olivei:
> Hey! Love what you did here @m4olivei. Great work!

Thanks @KevinCamelo! The designs were very well thought out such that they are a natual extension of the Slider, which made my life so much easier! Great work to you as well.

> I appreciate that you removed the option for error states and just have the slider bind to the nearest value or the minimum or maximum. I like that approach a lot more and begs the question when error states would even appear in this case, similar to the base slider component.

Glad you like that. Going with the approach was mostly bourne out of there being a lot more error cases with the two handle slider, specifically, setting a value beyond the limits of the currently set opposite value. You wouldn't ever want one of the range handles to cross past it's neighbor. I had that at one point, it was very weird looking. I settled on this approach, where the user is given freedom to enter whatever they want. If the value is valid, the range slider adjust as the user types. If the valid is invalid, the slider adjust to the extreme allowable, then auto-corrects on `<Enter>` or blur.

The error state _can_ appear for non-numeric values. The input being type `number` makes is very rare, but you can enter a partial exponential notation, eg. `10e`, which will fail the behind the scenes conversion to a number. It can also occur programatically.

> I noticed that the keyboard commands don't work when clicking on the slider itself. When a user clicks Shift + →  increase by 10 and when they click shift + ← decrease by 10. And when they focus in and press → or ← it should increase or decrease by 1. These interactions are present in the base slider component and should just carry over.

Nice catch! I inadvertandly broke that for the two handle slider. Now fixed.

> In Gray 100 and Gray 90, the color of the slider does not change to $focus on selection or focus.

Hum, looks like this may be an issue with the Default Slider as well ([link to production Storybook](https://react.carbondesignsystem.com/?path=/story/components-slider--default&globals=theme:g100))? I've fixed for both. Let me know if that's looking as expected now? If not, can you send me an annotated screenshot of the discrepancy you're seeing?

m4olivei:
I started writing some tests against the two handle slider. There needs to be more. I'm new to Jest and testing-library eco-system, so do correct me if you see things off.

@tay1orjones given the already big size of the PR, I'm thinking a good end goal for this PR is to address all feedback, and have a solid bunch of tests. The React hooks rewrite methinks should be a follow up PR and not block this. Thoughts?

In any case, I'm away on vacation the next two weeks, so things will be silent from my end until I'm back. I should be able to address any feedback in earnest when I'm back though, so keep it coming. Thanks for all the careful thought and review so far, and looking forward to more.

KevinCamelo:
> 
> Thanks @KevinCamelo! The designs were very well thought out such that they are a natual extension of the Slider, which made my life so much easier! Great work to you as well.
> 

Thank you Matthew! 

> Going with the approach was mostly bourne out of there being a lot more error cases with the two handle slider, specifically, setting a value beyond the limits of the currently set opposite value. You wouldn't ever want one of the range handles to cross past it's neighbor. I had that at one point, it was very weird looking. I settled on this approach, where the user is given freedom to enter whatever they want. If the value is valid, the range slider adjust as the user types. If the valid is invalid, the slider adjust to the extreme allowable, then auto-corrects on `<Enter>` or blur.
> 

I see! Great rationale — I enjoy this approach. Michael Gower did mention that it would be nice for there to be a way to inform our users that the value was autocorrected. I'm currently touring with this proposed solution — will collect some thoughts and get back to you. 

| Potential solution |
| ---- |
|rangeslide|

Thanks for fixing the theming and keyboard commands! I'll be presenting this solution at DSAG on the 17th of August. Once your back, I'll report back with the documentation associated with the designed solution. Hopefully we can release in experimental close to that date! 

Once again, thank you so much for the work on this Matthew. Enjoy your vacation.

tay1orjones (commented):
This is really looking great! Two things I noticed:

1. When a handle is focused/selected via click, this causes the handles to move. I think this might also be present on the base (non-range) slider. I think ideally the handle wouldn't move in this case.

https://github.com/carbon-design-system/carbon/assets/3360588/63701744-03c1-48a6-802e-af94905bef97



2. Also do we want to include the interaction where clicking on the track moves the handles? It's also part of the base Slider. With two handles, it's sometimes difficult to know if clicking on a section of the track is going to move the start handle or the end handle.





https://github.com/carbon-design-system/carbon/assets/3360588/ab87a1fd-e6c3-4bf9-bbdf-327c66315605

tay1orjones:
> @tay1orjones given the already big size of the PR, I'm thinking a good end goal for this PR is to address all feedback, and have a solid bunch of tests. The React hooks rewrite methinks should be a follow up PR and not block this. Thoughts?

@m4olivei Yes, totally agree. That would be excellent as a follow up to keep the review here reasonable. Another side benefit is being able to compare that future work against this pr/deploy preview.

tay1orjones (commented):
A few notes on the API design from reviewing the code:

- These new props should be [marked experimental](https://github.com/carbon-design-system/carbon/blob/main/docs/experimental-code.md) until the refactor to functional components is complete. `unstable_valueUpper`, for instance.
- Could we consolidate the `lower*` props to utilize the existing props like `value` for the the lower value/label/name etc?
  - Use `value` as the lower value, and remove `valueLower`
  - Remove `twoHandles`, and use `!!valueUpper` instead. Internally this could be `const isRange = !!valueUpper` or something for readability.
  - Use `isValid` for validity of the lower value (`value`), and remove `isValidLower`
  - Use `ariaLabelInput` for the lower input, remove `ariaLabelLower`
  - Use `name` for the lower name, remove `nameLower`
  
Overall this would mean the prop interface would be modified less to only include new props for the `upper*` value/label/name, etc.

It also looks like the public api snapshot needs regenerated to show all these prop additions by running `yarn test -u`

m4olivei:
Thanks for all the feedback @tay1orjones ! I'll be gradually addressing this week and next.

First off:

> Also do we want to include the interaction where clicking on the track moves the handles? It's also part of the base Slider. With two handles, it's sometimes difficult to know if clicking on a section of the track is going to move the start handle or the end handle.

The way it's implemented was recommended by @mbgower, noted in this [Medium post](https://medium.com/carbondesign/not-dragging-our-feet-a03ea57150e), specifically the section titled ["Solution: a clickable slider bar"](https://medium.com/carbondesign/not-dragging-our-feet-a03ea57150e#a9cf). Per the article, when your click is roughly equidistant between the two handles, which one moves is moot, only that one of them moves. Given that prior study, is there more recent precendant/direction that supports a change away from this behavior?

m4olivei:
> It also looks like the public api snapshot needs regenerated to show all these prop additions by running `yarn test -u`

@tay1orjones This is an odd one. I don't get any changes to the snapshot when I run `yarn test -u`. Looks like this is an existing issue that has something to do with how `Slider` is exported from `src/components/Slider/index.js`. The [existing snapshot](https://github.com/carbon-design-system/carbon/blob/main/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap#L7257-L7259) does not specify any of its props, eg.

```
  "Slider" => Object {
    "displayName": "ClassWrapper(Slider)",
  },
```

There are a couple of other components that have this same behavior, where the only property in the snapshot is `displayName` and none of the props are listed.

Not sure how to remedy that, although it may be worth a follow up to also fix the other components affected in the same way (`TileGroup` and `ContentSwitcher`).

tay1orjones:
@m4olivei Ah yes, sorry my mistake. The class wrapper obfuscates the component API in the public snapshot. This will be another benefit of moving the component to be a functional component! 🙂

m4olivei:
A couple of updates:

> @m4olivei Hi Matthew, I'm currently seeing a bug on my side when setting up "valueLower" and "valueUpper". The values don't update until I hit refresh and toggle on "twoHandles". I tested it on Safari, Chrome, and Arc.

@thyhmdo This should now be addressed. Note that due to some of the other changes here, you'll want to look to adjust `value` and `unstable_valueUpper`. You no longer need to toggle `twoHandles`, as now a value set for `unstable_valueUpper` will assume you want two handles. The `unstable_` prefix will drop when we have a stable [release per the documentation](https://github.com/carbon-design-system/carbon/blob/main/docs/experimental-code.md).

> Could we consolidate the lower* props to utilize the existing props like value for the the lower value/label/name etc?

@tay1orjones yep! Now updated.

> These new props should be marked experimental until the refactor to functional components is complete. unstable_valueUpper, for instance.

This has also been updated.

m4olivei:
OK, I think I've addressed all the comments so far on the PR here. Other updates may come from other feedback from the designers.

> When a handle is focused/selected via click, this causes the handles to move. I think this might also be present on the base (non-range) slider. I think ideally the handle wouldn't move in this case.

@tay1orjones I thought about this some, and I think the behavior that we have is by design. Given that the handle is design such that the vertical line points to the value, and the fact that we want the ability for click-only value setting, it follows that the handle should adjust even slightly as it does to a click on the handle itself.

I've also added an auto-correct notification requested by @KevinCamelo and @mbgower . When we do an auto-correct, the notification will show up. It will remain until a further valid interaction with the slider will clear it. Open to feedback.

m4olivei:
Just realized while writing tests (yay tests). What do we want for when `invalid` or `warn` is set in a two handles situation? For [example here](https://deploy-preview-14297--carbon-components-react.netlify.app/?path=/story/components-slider--playground&args=invalid:true;value:10;unstable_valueUpper:90):

image

packages/react/src/components/Slider/Slider-test.js
```diff
@@ -101,8 +122,7 @@ describe('Slider', () => {
       );
     });
 
-    // ArrowRight is not triggering change in slider, possibly due to userEvent bug
-    it.skip('should change the value upon interacting with the slider', async () => {
+    it('should change the value upon interacting with the slider', async () => {
```
m4olivei, line 136: I worked out that `userEvent` didn't flesh out `KeyboardEvent.which`, which we were relying on in the component. That property is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#obsolete_properties). I updated the component to not rely on `KeyboardEvent.which`. See commit https://github.com/carbon-design-system/carbon/pull/14297/commits/20e5b9dadf44c7e5babde780e336fe42fd020301. 

We no longer have to skip these 🙌

andreancardona: @m4olivei will take a look at this today thanks for your patience!

tay1orjones:
@m4olivei Let me know if you'd like me to help implement some e2e tests covering the various states with visual regression tests (vrt) and accessibility verification tests (avt). I'm happy to pair up on it or just push to your PR once you feel things are pretty stable.

packages/react/src/components/Slider/Slider-test.js
m4olivei: Thanks @andreancardona . Looking forward to your feedback.

m4olivei:
> @m4olivei Let me know if you'd like me to help implement some e2e tests covering the various states with visual regression tests (vrt) and accessibility verification tests (avt). I'm happy to pair up on it or just push to your PR once you feel things are pretty stable.

Sure, could use some help with this. Honestly, e2e tests weren't on my radar. Can you point me to documentation and/or existing e2e tests for the Slider component? 

I'm hoping what we have here is stable. There is one more feature request outstanding (tooltips showing current value of the handle) that I'll be talking over with @KevinCamelo today. I'm hoping we can do that as a follow up.

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -907,6 +1487,17 @@ export default class Slider extends PureComponent<SliderProps> {
...
                 </div>
               )}
+              {correctedValue && (
+                <div
+                  role="alert"
+                  className={classNames(
+                    `${prefix}--slider__status-msg`,
+                    `${prefix}--form-requirement`
+                  )}>
+                  The inputted value &quot;{correctedValue}&quot; was corrected
```
m4olivei: @tay1orjones It just occurred to me today, what is the preferred method of translating such a string? I assume it can't be hardcoded as english?

m4olivei:
Hey @tay1orjones 👋 . Friendly bump on my last two notes here. Thanks.

packages/react/src/components/Slider/Slider.tsx
tay1orjones: Yeah the pattern we have established for this `translateWithId`. Here's some docs: 

https://github.com/carbon-design-system/carbon/blob/b8701c9578019c1f4fe3f92d4ef0e1a23139e9bd/docs/style.md#L430-L450

tay1orjones:
@m4olivei Hey sorry for missing these notifications! For e2e tests we have recently begun breaking out e2e avt tests into their own file, so Slider has two:

* `e2e/components/Slider/Slider-test.e2e.js` - vrt tests and any other general e2e tests live in this file
* `e2e/components/Slider/Slider-test.avt.e2e.js` - avt tests live in this file

m4olivei:
@tay1orjones Had a chance to power through the leftover items today:

* Resolved the merge conflict, which was brought on by recently improved support for RTL writing mode (https://github.com/carbon-design-system/carbon/pull/14635 and https://github.com/carbon-design-system/carbon/pull/14580). 
* Added e2e tests following existing examples for Slider. Note that the Percy job has new snapshots that need to be approved as a result.
* Added translation support for the auto-correct announcement string, per the docs you passed along.

Hoping we can wrap this up 🤞 . Let me know if you need anything.

tw15egan (approved):
This is looking really great! Thanks for all the work you've put into this. I think the only thing I might add that would help users implement a two-handle slider is adding a little blurb in `Slider.mdx` to let people know that you enable two-handle mode by adding the `unstable_valueUpper` prop. Otherwise, LGTM 👍 ✅

m4olivei:
Thanks @tw15egan! I've added that piece to the `Slider.mdx` file. Let me know if you have any feedback,

tay1orjones (approved):
Absolutely stellar work here, thank you so much for all the effort you put into this! 🎉 🏆  

I don't see any blockers to merge. The e2e tests for avt and vrt look great and I just approved Percy 👍

laurenmrice (changes requested):
Hey @m4olivei This looks amazing! Thanks so much for working on it. I just have a couple visual questions:

- **Triangle handle:** In Figma, Kevin is using the `skip--forward--filled` icon at 16px. I notice that your triangle has straight corners instead of rounded corners like the icon has, are you currently using the icon here or is this being drawn code?

- **Warning state:** It came up in the Figma design review with me, Kevin, and Juan, that there needs to be a warning state for this type of situation Kevin mentions [in a previous comment](https://github.com/carbon-design-system/carbon/pull/14297#issuecomment-1665747876). Stylistically we need to show the warning icon to indicate it is a warning state. I would assume depending on which inputted value was changed, that input would be the one that gets the warning icon warning icon? For example:
Frame 204

m4olivei:
Thanks @laurenmrice !

> Triangle handle: In Figma, Kevin is using the skip--forward--filled icon at 16px. I notice that your triangle has straight corners instead of rounded corners like the icon has, are you currently using the icon here or is this being drawn code?

Yes, its currently being drawn in code. It's difficult to use the SVG icons b/c of the focus state. It's probably do-able, but, I'd have to bring in both the `skip--forward--filled` icon as well as the `play--filled--alt` icon (for focus state so we don't have to deal with the vertical line in the skip icons), where I'd then have to rotate the latter by 180 degrees for use on the right side. The SVG's also have some space around the shapes themselves thats difficult to dial in. In either case I'm having to compose things in CSS. Getting the rounded corners for triangles drawn in CSS is hard. 

I think ideally, to get the best of all worlds, we need some custom SVG icons based on the standard skip / play icons. I would need all of the following icons:

* skip--forward--filled
* skip--forward--filled (with focus border)
* skip--back--filled
* skip--back--filled (rotate 180, with focus border)

All of the above should have the same size canvas (sized to the focus rectangle - 15x24 ??), with the appropriate spacing built in such that the focus icons, when superimposed over top of their non-focus sibling, should overlap exactly, if that makes sense. Is that something we could produce?

> Stylistically we need to show the warning icon to indicate it is a warning state. I would assume depending on which inputted value was changed, that input would be the one that gets the warning icon warning icon? For example:

Sure! I'll work on adding a warning state for this.

KevinCamelo:
Hey @m4olivei! I took a stab at making those requested SVGs. You can find them here: https://ibm.box.com/s/cd0y13hon8jwxxvggpbqqa4arrxfyrlz

I checked the spec in inspect and made them 16x24 as that is the space they occupy. 

**Note (suggestion from my non-dev eyes):** I think we could just use `skip--forward--filled` and `skip--back--filled` and apply a box-shadow and recolor to account for the focus state. That way we can avoid using 4 svgs and only use the 2 base ones for all states.

m4olivei:
> @m4olivei small nit, but is there any way to prevent the slider jumping ~5 when focus is added to the Slider handle?

Thanks for the feedback @tw15egan . We previously discussed that, and resolved that it's be design. See https://github.com/carbon-design-system/carbon/pull/14297#issuecomment-1690087619

tw15egan:
Awesome, thanks for clarifying! This is looking great 👍🏻

m4olivei:
@KevinCamelo Should we set a background color on the thumbs, or is it OK that the track shows through? Here are some screenshot zoomed to 300% to make it obvious:

**Transparent (this is the way I presently have it):**

image

image

With background:

image

image

KevinCamelo:
I think you're spot on with your suggestion of a background color @m4olivei, I think it might need a small tweak. 

With the addition of a background-color on the entire icon, it leads to a middle indicator that hides and appears arbitrarily. In addition, the background on both sides creates a second line-segment. 

| Hide and show middle indicator | Second line segment | 
| --- |--- |  
|issues-box|Screenshot 2023-10-04 at 12 43 25 PM|

Instead, I think we could have the background-color only appear on the rail directly inside the icon area on the right-hand side. That way, the middle indicator will always be visible, the rail isn't split into two line-segments, and makes it easier to read what is actually being selected. The same treatment would be provided to the focus state. The color used here would likely be `$layer-02` if I'm not mistaken. 

Here's what that proposal looks like:
vert

m4olivei:
Yep, makes sense @KevinCamelo . I've implemented that. Take a look. Be sure as well to jump between themes and check RTL to make sure I covered us in those use cases.

m4olivei:
Forgot to tag @laurenmrice . I think we've addressed the feedback from your last comment. Could you take a look as well? Let me know if you need any adjustments. If all is well Percy needs to be approved for the visual tweaks we've made with the last few commits.

mbgower:
Looks good! I entered in a letter in an input just to see what would happen and got a placeholder error. Wasn't sure whether that gets resolved elsewhere or we need to have all error messages composed as part of this.
image

mbgower:
Actually, one other thing I noticed. The slider label  is coded as a label, but it doesn't look like it's associated with the two inputs.
I guess "for" is not great for us here, since the attributed can't take two values.... We need the inputs to expose that label as well as their "lower bound" and "upper bound" info so, if this was a slider on price, you'd want to hear something like "price lower bound" and "price upper bound". 
I can think of a few of ways of doing this. I'm sure you can think of more. One would involve a group label (which may only announce on the first input, which I think is okay). The other is to concatenate the two strings. A third may be to use a description for part of the label, effectively getting something like "Upper value, price". And I guess a fourth would be to put a span around the label and reuse it that way.
BTW, the label here is "Slider Label", which at the least should be "Slider label" and better, a realistic text string, like "Price range" or whatever.

KevinCamelo:
> Yep, makes sense @KevinCamelo . I've implemented that. Take a look. Be sure as well to jump between themes and check RTL to make sure I covered us in those use cases.

Looking great, Matthew. Went across all themes— just noting that the selected state seems to not have the correct color across themes. The selected rail should use the `border-interactive` color token to work across themes. The icon as well does not align with the color in the original slider component.

G10/White 

Screenshot 2023-10-05 at 10 47 53 AM


G90/100 (incorrect color on rail and icon)
Screenshot 2023-10-05 at 10 49 55 AM

How it should appear:
Screenshot 2023-10-05 at 10 54 07 AM

To match the dark theme of the OG slider: 
Screenshot 2023-10-05 at 10 55 18 AM

For the rail: `border-interactive`
For the icon: `interactive` upon click.

**Additional info on the icon:**

From what I can tell, in the base slider component instead of using the`icon-interactive`, they use the `interactive` token to achieve the desired recolor for the icon across themes. If you use `icon-interactive` the interactive state is white, which would not align with the OG slider.

m4olivei:
> Looks good! 

Thanks @mbgower ! 

> I entered in a letter in an input just to see what would happen and got a placeholder error. Wasn't sure whether that gets resolved elsewhere or we need to have all error messages composed as part of this. I was looking at https://deploy-preview-14297--v11-carbon-react.netlify.app/?path=/story/components-slider--two-handle-slider#storybook-preview-wrapper

That gets handled by the user passing a prop, `invalidText`. The Storybook for the two handle slider passes "Invalid text goes here", it can be customized, but is expected to be general enough to cover any invalid state, of which there are few, given the auto-correction built in.

> Actually, one other thing I noticed. The slider label is coded as a label, but it doesn't look like it's associated with the two inputs. I guess "for" is not great for us here, since the attributed can't take two values.... We need the inputs to expose that label as well as their "lower bound" and "upper bound" info so, if this was a slider on price, you'd want to hear something like "price lower bound" and "price upper bound".

Yep. I struggled with this earlier in the development. What I settled on, was that since the single label can't pair to both inputs, I leave the label disconnected (I guess it could be swapped for a `<span>`). Then I use `aria-label` on each of the inputs, which is expected to be passed as props (`ariaLabelInput` and `unstable_ariaLabelInputUpper`), the Storybook example passes "Lower bound" and "Upper bound" respectively. Using VoiceOver + Chrome, focusing on the input announces "10, Lower bound, stepper", and on the upper input "90, Upper bound, stepper". To acheive your example, you'd simply pass "price upper bound" and "price lower bound".

> BTW, the label here is "Slider Label", which at the least should be "Slider label" and better, a realistic text string, like "Price range" or whatever.

I've corrected the casing on that. I kind of want to avoid adopting price for the example, b/c of internationalization, but open to update if you like.

m4olivei:
> Looking great, Matthew. Went across all themes— just noting that the selected state seems to not have the correct color across themes. The selected rail should use the border-interactive color token to work across themes. The icon as well does not align with the color in the original slider component.

Thanks @KevinCamelo , nice catch. I've gone through and corrected those tokens.

laurenmrice (approved):
Looks good, thank you @m4olivei ⭐️

Issue #13258, opened 2023-02-28
[Fluid inputs] Modal form is not rendering correctly.


aagonzales:
How fluid form modal is currently appearing in storybook

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/11670886/221955998-75a0f101-823c-4996-a002-8e864d1ce69e.png">

How it should look
- Fluid inputs use the background color `$field-01` here to match the modal background.
- Fluid inputs should be flush to the edge of the modal.

<img width="744" alt="image" src="https://user-images.githubusercontent.com/11670886/221957674-53ec2013-7894-4a08-a00a-a988fa3ab7ed.png">

PR #14303, opened 2023-07-25
fix(Modal): ensure dropdown menu layer is correct

### LINKED ISSUE #13698: [Bug]: Dropdown & Fluid inputs inside modal - wrong background-color token
### Package

@carbon/react

### Browser

Chrome

### Package version

v11.28.0

### React version

18.2.0

### Description

In the Carbon React Modal (https://react.carbondesignsystem.com/?path=/story/components-modal--default), the dropdown/combobox menu children have `$field-01` as `background-color` which does not match the design spec
image

The background color token for the dropdown children should be `$layer-02` as per he design spec
image

This probably needs to be updated here https://github.com/carbon-design-system/carbon/blob/de8d41b529044ea6d2f38f963459cefead37daa5/packages/styles/scss/components/modal/_modal.scss#L54-L69

Also, just below that is the declarations for fluid inputs too, which should also be using` $layer-02` there instead of `$field-01`

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-modal--default

### Steps to reproduce

1. Go to https://react.carbondesignsystem.com/?path=/story/components-modal--default
2. Open dropdown
3. Notice background color of children

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13698
Refs https://github.com/carbon-design-system/carbon/issues/13258
Refs https://github.com/carbon-design-system/carbon/pull/13879

Ensures all listbox menus render the correct background color

#### Changelog

**Changed**

- Aligned menu tokens with the appropriate layer token (scope inside modal)

#### Testing / Reviewing

Ensure `Dropdown` renders the correct menu background when open


tay1orjones (approved):

alisonjoseph (approved):

laurenmrice (commented):
Can you check to see what the other state color tokens are for the options in the dropdown inside of the modal? For the dark themes it looks like the hover and selected states are wrong.
Gray 100

tw15egan:
@laurenmrice just added some more overrides to enforce those specific styles 👍🏻

laurenmrice (approved):

Issue #14300, opened 2023-07-25
Clarify `danger` is not a valid type of `IconButton` or `<Button hasIconOnly/>`


tay1orjones:
This has been a point of confusion for some time (https://github.com/carbon-design-system/carbon/issues/11484, https://github.com/carbon-design-system/carbon/issues/14295) and we have not clearly stated anywhere on the website or storybook (that I can find) that we do not support `danger` for the icon only button(s). 

This includes both `IconButton`, and `Button` with `hasIconOnly`.

```[tasklist]
### Pieces to fix
- [x] Double check the types for IconButton do not allow `danger`
- [ ] https://github.com/carbon-design-system/carbon/pull/14307
- [ ] https://github.com/carbon-design-system/carbon/issues/17036
- [ ] https://github.com/carbon-design-system/carbon-website/issues/4176
```

albert-schilling:
I don't understand why Danger should not be allowed as a kind for the IconButton. We use it a lot in our app. For example for delete buttons or any kind of IconButton that does something destructive.

PR #14307, opened 2023-07-25
docs(button): specify danger is not supported for icon only buttons

tay1orjones:
Part of https://github.com/carbon-design-system/carbon/issues/14300

#### Changelog

**Changed**

- Clarify in Button and IconButton `.mdx` docs that danger is not a supported option for icon only buttons
- Reformat doctoc comments to be compatible with MDX2

#### Testing / Reviewing

- Proofread the docs changes

### DIFF
--- a/packages/react/src/components/Modal/Modal.mdx
+++ b/packages/react/src/components/Modal/Modal.mdx
@@ -97,19 +97,20 @@ const ModalStateManager = ({
 };
 ```
 
+{/* <!-- prettier-ignore-start --> */}
+
 <Unstyled>
   <InlineNotification
     title="Warning"
     kind="warning"
     className="sb-notification">
-    <CodeSnippet type="inline" hideCopyButton>
-      {'<Modal>'}
-    </CodeSnippet>
-    /<CodeSnippet type="inline" hideCopyButton>{`<ComposedModal>`}</CodeSnippet>{' '}
+    <CodeSnippet type="inline" hideCopyButton>{'<Modal>'}</CodeSnippet>
+    <CodeSnippet type="inline" hideCopyButton>{`<ComposedModal>`}</CodeSnippet>{' '}
     has to be put at the top level in a React tree. The easiest way to ensure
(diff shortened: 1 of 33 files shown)


packages/react/src/components/Modal/Modal.mdx
```diff
@@ -96,11 +96,18 @@ const ModalStateManager = ({
...
+  <InlineNotification
+    title="Warning"
+    kind="warning"
+    className="sb-notification">
+    <CodeSnippet type="inline" hideCopyButton>
+      {'<Modal>'}
+    </CodeSnippet>
+    /<CodeSnippet type="inline" hideCopyButton>{`<ComposedModal>`}</CodeSnippet>{' '}
+    has to be put at the top level in a React tree. The easiest way to ensure
+    that is using React portal as shown in above example.
```
tw15egan: These `CodeSnippet` blocks seem to be causing crashes here and in `ComposedModal`

tay1orjones: Thanks for pointing this out. It should be fixed now.

To find valid syntax for .mdx files w/ MDX2, I've been doing the following:
1. Take the erroring portions (like this snippet here) and put it into the [mdx playground](https://mdxjs.com/playground/)
2. Starting at the most-nested components, reduce these down to single-line entries
```diff
- <CodeSnippet type="inline" hideCopyButton>
-  {'<Modal>'}
- </CodeSnippet>
+ <CodeSnippet type="inline" hideCopyButton>{'<Modal>'}</CodeSnippet>
```
3. Continue simplifying the formatting until the compiler is successful and doesn't report errors
4. Put the valid code over into the .mdx file and surround it with prettier ignore comments. These look a bit different in mdx2 and are surrounded in curly braces: `{/* <!-- prettier-ignore-start --> */}`

tw15egan (approved):
LGTM 👍 ✅

alisonjoseph (approved):

Issue #13680, opened 2023-04-27
[a11y]: Modal: doesn't return focus properly on dismissal


finken2:
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

v1.25.0

### React version

v16.12.0

### Automated testing tool and ruleset

manual testing

### Description

When launching a Modal from a button, focus goes inside the Modal, as it should. When dismissing the Modal, focus should return to the button where it was launched from. 
This issue was mentioned in #12727, but i thought it was worth splitting out separately. 

https://www.w3.org/WAI/WCAG21/Understanding/focus-order

Optimally, we'd automatically save off focus on launch and reset it on dismissal. Alternatively, having a prop for a selector to return focus too would also be an option.

### WCAG 2.1 Violation

WCAG 2.1, focus order 

### Reproduction/example

any implementation of Modal from a button

### Steps to reproduce

Launch a modal from a button. Dismiss modal. Note focus does not return to previous location. Focus is on the main `body` node after Modal dismissal.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
Thanks for reporting this. I can reproduce it in [the storybook](https://react.carbondesignsystem.com/?path=/story/components-modal--with-state-manager) by focusing the "Launch modal" button, pressing <kbd>enter</kbd> to open the modal, and pressing <kbd>esc</kbd> to close the modal. Focus is not returned to the button

tay1orjones:
How this was originally outlined in parent #12727 

> Whatever launches a dialog, when it is dismissed, that item should regain focus. In the implementations in React storybook, the focus is not shown when the dialog closes, and the user must shift+tab back to put the button in focus.

tay1orjones:
When this is tackled, we should cover the solution in automated keyboard nav tests like we're doing in https://github.com/carbon-design-system/carbon/issues/13747 and https://github.com/carbon-design-system/carbon/issues/13748

tw15egan:
[...]

PR #14320, opened 2023-07-26
feat: Nest modal support

### LINKED ISSUE #14104: [Modal]: Support nested modals
### The problem

Currently it's not possible to nest a modal within a modal without causing breaking problems. While we can and have argued that modals shouldn't be nested within modals, that hasn't stopped users from doing it and then opening issues about it. This is especially true for [carbon for ibm products](https://github.com/carbon-design-system/ibm-products/tree/main/packages/ibm-products) which extends modal several times. The biggest point of contention being with the Tearsheet component where we have seen numerous complaints and bug reports regarding the inability to use a modal within tearsheet. And yes you could argue (and we have) that if you're using Tearsheet in a way that requires an additional modal that perhaps a tearsheet isn't the ideal solution for you or your product, that still hasn't stopped it from happening.

So, I'm here to advocate for a look at the current modal code to see if we can make adjustments so that nesting modals is possible. I'm also here to offer my services to fix this problem personally.

### The solution

After examining the modal code I believe the main problem can be addressed by making some adjustments to the structure of modal and how certain events like onClose are triggered.

### Examples

_No response_

### Application/PAL

https://github.com/carbon-design-system/ibm-products/tree/main

### Business priority

None

### Available extra resources

as a developer i'd be more than happy to try to tackle this

### Code of Conduct

davidmenendez:
Closes #14104 

this is a concept for nested modal support changes. the general idea with this PR was just to demonstrate that a few small changes could make nested modal support possible. i wanted to gauge reception before continuing to integrate all the necessary testing and functionality that needs to be modified.


packages/react/src/components/Modal/Modal.js
```diff
@@ -283,32 +284,38 @@ const Modal = React.forwardRef(function Modal(
...
-      {/* Non-translatable: Focus-wrap code makes this `<span>` not actually read by screen readers */}
-      <span
-        ref={endTrap}
-        tabIndex="0"
-        role="link"
-        className={`${prefix}--visually-hidden`}>
-        Focus sentinel
-      </span>
-    </div>
+    <Portal>
```
tay1orjones: I think that rendering the modal in a portal by default would be a breaking change. The majority of usage in the wild probably already renders the modal via a portal and would result in nested portals. Users could/would run into not being able to access events of the innermost portal, https://github.com/facebook/react/issues/14540

tay1orjones (commented):
Very cool to see this working! I'm not sure if we could make this change right now though with the way it's proposed here.

tay1orjones:
The lack of focus returning to the trigger element that opened the modal (https://github.com/carbon-design-system/carbon/issues/13680) is particularly glaring when keyboard navigating through the nested modals. 

Figuring out a fix that is compatible with both non-nested and nested modals would be really nice. I'm not sure if it's a full blocker to landing nested modals here, but it's close.

packages/react/src/components/Modal/Modal.js
tay1orjones: One way to evaluate backwards compatibility would be to leave the existing 'with state manager' story as-is and use it as a testing grounds for how most modals are used right now.

davidmenendez:
apologies. been busy! i'll be looking at this today

packages/react/src/components/Modal/Modal.js
davidmenendez: good point

davidmenendez:
@tay1orjones i'll continue to look at the keyboard issues. i just pushed up a change to revert the portal issue you pointed out. the story uses the original `ModalStateManager` but it's using the internal `Portal` component. When you get a chance please let me know what you think about this implementation.

davidmenendez:
@tay1orjones whenever you have a chance to re-review it would be greatly appreciated!

packages/styles/scss/components/modal/_modal.scss
```diff
@@ -25,15 +25,10 @@
   .#{$prefix}--modal {
     position: fixed;
     z-index: z('modal');
-    top: 0;
-    left: 0;
     display: flex;
-    width: 100vw;
-    height: 100vh;
```
tay1orjones: I worry that other implementations of the modal in angular, svelte, etc. would be adversely affected by this change. @tw15egan do you think this would be a breaking change? We usually don't for styles, but I'm not totally sure in this case.

packages/react/src/components/Modal/Modal.js
```diff
@@ -294,12 +294,16 @@ const Modal = React.forwardRef(function Modal(
...
+      ref={ref}
       role="presentation"
-      ref={ref}>
+      onKeyDown={handleKeyDown}>
+      <div
+        className={`${prefix}--modal-background`}
+        onMouseDown={handleMousedown}
+        onBlur={handleBlur}
+        aria-hidden
+      />
```
tay1orjones: Framework implementation authors (web components, svelte, angular, etc.) will need to make this change as well, correct? Without it, after taking this update their modal's are probably going to break, right?

tay1orjones (commented):
Hey sorry for the delay. I left some comments below, I'm mainly worried about how ubiquitous modals are and if this change steps into the realm of being a breaking change. 

Also I think it would be great if we could confirm support for the new `launcherButtonRef` functionality added in https://github.com/carbon-design-system/carbon/pull/14355 works in the nested situation.

packages/react/src/components/Modal/Modal.js
tay1orjones: @davidmenendez Could you outline why this new separate element is needed?

davidmenendez:
thanks for the feedback! will work through this and reply soon

andreancardona:
@davidmenendez Hey David, just thought I'd follow up here :)

davidmenendez:
apologies. i've gotten a little behind on this. still a priority. i'll have a follow up ASAP!

davidmenendez:
double apologies. i'm working on this today 😬

packages/react/src/components/Modal/Modal.js
davidmenendez: i took inspiration from some other modal implementations i had seen that used this pattern, but that's a valid point about breaking changes. i'll revert it as i think i can make it work without it

ferdelamad:
Hello 👋 , thanks @davidmenendez for adding this feature!
Can the `evt.stopPropagation();` also be added to the `ComposedModal` please?

Are there any plans to merge and release this feature soon?

Thanks for all the hard work on carbon 🙌

davidmenendez:
@ferdelamad seems reasonable. 

> Can the evt.stopPropagation(); also be added to the ComposedModal please?

@tay1orjones any thoughts?

tay1orjones:
@davidmenendez Sorry for the delay here, was out for the holidays. I so appreciate your continued effort on this! In addition to the focus/onBlur fix you mention, I have a couple other thoughts:

* `Portal` is an internal helper component and is not exported. Looking at the source, I don't think there's anything particularly special in there anyway. Could the storybook example be updated to invoke `ReactDOM.createPortal()` again?
* Eventually, just before merging, I'd like to remove the `nested` story. Due to the UX concerns, nested modals aren't something we want to "recommend" by having a story for it.
* That said, I think it's important to document in a new section in `Modal.mdx`: _"Nesting of modals isn't recommended due to various UX concerns, but there should be no blockers to implementing nested modals. Here is an example"_ - and link out to a stackblitz essentially containing what the `nested` story contains right now
* I don't see why we couldn't add the fix to ComposedModal as well

It's really interesting that this so far has boiled down to just the `evt.stopPropagation` and some example/configuration code in storybook. @ferdelamad I'm eager to get this merged in, with the above ironed out I think this would be in great shape to do so 👍

davidmenendez:
@tay1orjones happy holidays! haha no worries. we've all been on cruise control these past few weeks 😂 

thanks for the feedback. i totally agree with your stance on removing the story as to not promote this practice of nested modals and to include such verbiage in the documentation. that's the approach we also want to take in the C4P library. i will make the updates you suggested ASAP 👍

davidmenendez:
@tay1orjones changes made 👍

davidmenendez:
i see there's an issue with the E2E test. i'll investigate and fix ASAP 👍

davidmenendez:
E2E test pass but now percy is failing. i'm not familiar with this testing so i'll have to do some investigation. i would definitely welcome any help there haha

tay1orjones (approved):
Awesome, thanks again for the work on this! I think the two remaining ideas can be contained in a new/follow-up PR if you or someone else wants to.

> - That said, I think it's important to document in a new section in Modal.mdx: "Nesting of modals isn't recommended due to various UX concerns, but there should be no blockers to implementing nested modals. Here is an example" - and link out to a stackblitz essentially containing what the nested story contains right now
> - I don't see why we couldn't add the fix to ComposedModal as well

davidmenendez:
yeah i can totally follow back up 👍

tw15egan (approved):
Who knew this is all it took! Thanks for following through with this @davidmenendez 🤘🏻 🎉

PR #14336, opened 2023-07-27
fix(layout): create mixin

ariellalgilmore:
Closes #

In order for `@carbon/web-components` to use the `layout-tokens` we need a separate `mixin`, so that we can apply the tokens inside the encapsulated web component and not the `root`

#### Changelog

**Changed**

- created a `emit-layout-tokens` mixins and then added the mixin to the `root`

#### Testing / Reviewing

confirmed via storybook that layout tokens are still working

### DIFF
--- a/packages/styles/scss/_layout.scss
+++ b/packages/styles/scss/_layout.scss
@@ -32,7 +32,7 @@ $layout-tokens: (
   ),
 );
 
-:root {
+@mixin emit-layout-tokens {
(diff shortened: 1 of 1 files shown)


tay1orjones (approved):
Makes total sense, thanks for submitting this!

tw15egan (approved):
Thanks for doing this 👍🏻

PR #14342, opened 2023-07-28
fix(styles): add display property to contained liste item class

### LINKED ISSUE #14341: [Bug]: Need to update contained list item styles to work with custom HTML elements
### Package

@carbon/styles

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v1.34.0

### React version

_No response_

### Description

Similar to issue #11579 - We are looking to release `contained-list` in Angular. In v11, we have updated the component structures to not use `<li></li>` (list item elements) where possible because of the custom tags used in Angular resulting in `List element is improperly structured` accessibility recommendation. To resolve this we've applied roles and the list item classes directly to the custom element host.

E.g.

```
<cds-contained-list-item role="listitem" class="cds--contained-list-item"></cds-contained-list-item>
```

Unfortunately, `cds--contained-list-item` class for contained-list does not have a display style, so the browser sets the display property to `inline` by default. This is currently blocking our contained-list component release.

A possible workaround we have is to export custom styles with the carbon-components-angular package, which is NOT ideal.

### Reproduction/example

NA

### Steps to reproduce

We haven't released the component so cannot reproduce. 

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Akshat55:
Closes #14341

Similar to PR #12050

Specify display CSS property for specific wrapper classes.

This is to apply the style upstream instead of in Carbon Angular. List item elements already have a display property of `list-item` by default, but since we have to use custom HTML elements in angular, we have to explicitly specify the display property on the host. So this change aims to bring the style changes upstream.

#### Changelog

**New**

- Specified display property for wrapper classes of accordion item, table, & context menu components.

#### Testing / Reviewing

- [ ] Build the styles package
- [ ] Comment out or remove ALL styles in React package storybook styles.scss (packages/react/.storybook/styles.scss)
- [ ] Import the built styles in React package storybook styles file:
```scss
@import '../../styles/css/styles.css';
```
- [ ] Run storybook & verify if there are any visual changes to Accordion, Data Table, & Context Menu components.

I have reviewed React package code & verified via storybook to ensure these changes do not impact the package, but VRT can be performed against multiple browsers.

### DIFF
--- a/packages/styles/scss/components/contained-list/_contained-list.scss
+++ b/packages/styles/scss/components/contained-list/_contained-list.scss
@@ -134,6 +134,7 @@
 
   .#{$prefix}--contained-list-item {
     position: relative;
+    display: list-item;
     list-style: none;
   }
 


alisonjoseph (approved):
LGTM! Not seeing any style regressions

tay1orjones (approved):
All the contained list variants are covered in vrt snapshots with Percy. Percy didn't detect any changes and I don't see any either 👍

PR #14351, opened 2023-07-31
fix(HeaderPanel): Initialize `onHeaderPanelFocus` with default value

sjbeatle:
#### Changelog

**Changed**

- `HeaderPanel` has a prop, `onHeaderPanelFocus`, which is typed to be an optional function, but the component logic calls the function indiscriminately as if it were a required property. This is causing adopters to have errors when using `HeaderPanel` without that property defined.

Updating the component to initialize the optional prop with an inert function. I'm not sure if that's an acceptable fix, or not. Alternatives may be:
* Make `onHeaderPanelFocus` required
* Check for `onHeaderPanelFocus` being defined prior to calling it within the component's logic.

#### Testing / Reviewing

Specs have been added. Checkout this branch and run `yarn test -- headerpanel` with and without the `onHeaderPanelFocus` prop being initialized to see the fix.

```typescript
    onHeaderPanelFocus = () => {},
```
vs.
```typescript
    onHeaderPanelFocus,
```


packages/react/src/components/UIShell/HeaderPanel.js
```diff
@@ -19,7 +19,7 @@ const HeaderPanel = React.forwardRef(function HeaderPanel(
     className: customClassName,
     expanded,
     addFocusListeners = true,
-    onHeaderPanelFocus,
+    onHeaderPanelFocus = () => {},
```
tw15egan: We're eventually going to need to migrate this to a serially [stable default value](https://github.com/carbon-design-system/carbon/issues/10777), so could we just switch this line to something like

```js
const noopFn = () => {};


const HeaderPanel = React.forwardRef(function HeaderPanel(
  {
    onHeaderPanelFocus = noopFn,
...
```

Otherwise, the change makes sense. Thanks for fixing this 👍🏻

packages/react/src/components/UIShell/__tests__/HeaderPanel-test.js
```diff
@@ -63,5 +64,38 @@ describe('HeaderPanel', () => {
...
+        </HeaderPanel>
+      );
+
+      await waitFor(() =>
+        expect(async () => {
+          screen.getByTestId('button').focus();
+          await userEvent.keyboard('{Escape}');
+        }).not.toThrow()
+      );
+    });
```
tay1orjones: Some minor nitpicks, suggested via the change below

- rename `callback` to `onHeaderPanelFocus`
- refactor to remove waitFor (these shouldn't be necessary I don't think)
- use accessible queries instead of `getByTestId`

```suggestion
    it('should call `onHeaderPanelFocus` callback, when defined', async () => {
      const onHeaderPanelFocus = jest.fn();
      render(
        <HeaderPanel onHeaderPanelFocus={onHeaderPanelFocus}>
          <button type="button">Test</button>
        </HeaderPanel>
      );

      screen.getByRole('button', { name: 'Test' }).focus();
      await userEvent.keyboard('{Escape}');

      expect(onHeaderPanelFocus).toHaveBeenCalled();
    });

    it('should not error when `onHeaderPanelFocus` is not defined', async () => {
      render(
        <HeaderPanel>
          <button type="button">Test</button>
        </HeaderPanel>
      );

      await expect(async () => {
        screen.getByRole('button', { name: 'Test' }).focus();
        await userEvent.keyboard('{Escape}');
      }).not.toThrow();
    });
```

tay1orjones (approved):
Looks great, thanks for this! Just a couple minor suggestions

packages/react/src/components/UIShell/__tests__/HeaderPanel-test.js
tay1orjones: @sjbeatle I went ahead and pushed up this suggested change to this PR, hope that's okay, let me know if there's something I missed!

packages/react/src/components/UIShell/HeaderPanel.js
tay1orjones: Yeah this could cause unnecessary component re-renders since the reference to the function will change every render.

tay1orjones: > Check for `onHeaderPanelFocus` being defined prior to calling it within the component's logic.

That's definitely an option, it's pretty common in the codebase to do something like `onHeaderPanelFocus && onHeaderPanelFocus()`

sjbeatle: > > Check for `onHeaderPanelFocus` being defined prior to calling it within the component's logic.
> 
> That's definitely an option, it's pretty common in the codebase to do something like `onHeaderPanelFocus && onHeaderPanelFocus()`

Yeah, I was debating that... the reason I landed where I did was because it is a single source that propagates throughout the rest of the logic. That way any future changes within the component won't require you to have the cognitive overhead of remembering to do an "null check" 👍

tay1orjones:
I don't think there's a way to override the DCO check specifically but we can merge with that one failing here I think.

tw15egan (approved):

Issue #13911, opened 2023-06-02
[Question]: SideNav not closing when clicking links


jeanbu:
### Question for Carbon

**Package**
@carbon/react

**Browser**
Chrome

**Package version**
@carbon/react: 1.19.0

**React version**
18.2.0

Description
As pointed out in [#3666](https://github.com/carbon-design-system/carbon/issues/3666) the sideNav menu is not collapsing when clicking a link element. The issue to close the menu by clicking anywhere on the overlay was solved in [#8296](https://github.com/carbon-design-system/carbon/pull/8296) but can't find any way to achieve the same behaviour as in the https://carbondesignsystem.com/ and automatically close the side menu. Has this issue been solved?

`const MainHeader = () => {
  return (
    <HeaderContainer
      render={({ isSideNavExpanded, onClickSideNavExpand }) => (
        <Header aria-label="Header navigation">
          <SkipToContent />
          <HeaderMenuButton
            aria-label={isSideNavExpanded ? "Close menu" : "Open menu"}
            onClick={onClickSideNavExpand}
            isActive={isSideNavExpanded}
          />
        ...
          <SideNav
            aria-label="Side navigation"
            expanded={isSideNavExpanded}
            isPersistent={false}
            onOverlayClick={onClickSideNavExpand}
          >
            <SideNavItems>
              <HeaderSideNavItems>
                <HeaderMenuItem element={NavLink} to={`/link1}>
                  Link1
                </HeaderMenuItem>...`

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

alisonjoseph:
Hi, can you clarify what you are asking for? Can you post a reproduction in [stackblitz](https://stackblitz.com/github/carbon-design-system/carbon/tree/main/examples/vite?file=src%2Findex.scss) and/or with video screenshot?

jeanbu:
Sure! I refer to the same issue as this [question](https://github.com/carbon-design-system/carbon/discussions/12708#discussion-4589282).

[Screencast from 13-06-23 10:17:08.webm](https://github.com/carbon-design-system/carbon/assets/105250478/a39549e2-65e5-44dd-a190-ca079973236d)

[...]

PR #14354, opened 2023-07-31
fix(SideNav): don't change expanded state when it's controlled

lamualfa:
#### Related issues

- https://github.com/carbon-design-system/carbon/issues/3666
- https://github.com/carbon-design-system/carbon/issues/13911#issuecomment-1588798106

The issue is closed but the problem still happened.

#### Changelog

**New**

- Don't change the `expanded` state if the `expanded` prop is `true` or `false`.

**Removed**

- Change the `expanded` state when the `expanded` prop is falsy value (`undefined`, `false`, `null` or `0`). Make the `SideNave` close even if the `expanded` prop that passed is `true`.

#### Before

`expanded` prop already `false`. But the `SideNav` set its internal state to `true`. Make the `SideNav` open again. 

https://github.com/carbon-design-system/carbon/assets/39755201/0de6f0f8-75ef-45a0-9696-edc5e9034ba0

### DIFF
--- a/packages/react/src/components/UIShell/SideNav.tsx
+++ b/packages/react/src/components/UIShell/SideNav.tsx
@@ -91,7 +91,7 @@ function SideNavRenderFunction(
   const navRef = useMergedRefs([sideNavRef, ref]);
 
   const handleToggle: typeof onToggle = (event, value = !expanded) => {
-    if (!controlled) {
+    if (controlled === undefined) {
       setExpandedState(value, enterDelayMs);
     }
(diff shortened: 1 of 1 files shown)


tay1orjones:
@lamualfa is there a way to test this in the storybook?

tay1orjones:
@lamualfa feel free to put together a story just for temporary testing in the storybook in this PR. We can use it to test using the deploy preview and then it can be deleted before merging.

PR #14355, opened 2023-07-31
feat(Modal): add launcherButtonRef prop to handle focus on close

### LINKED ISSUE #13680: [a11y]: Modal: doesn't return focus properly on dismissal

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13680

Adds a new `launcherButtonRef` to `Modal` and `ComposedModal`. This allows a user to pass in a `ref` to the button that is used to launch the `Modal`

#### Changelog

**New**

- `launcherButtonRef` prop added to `Modal` / `ComposedModal`.

**Changed**

- Updated the state manager stories to use the new `launcherButtonRef` prop to handle focus returning to the button when the `Modal` / `ComposedModal` is closed

#### Testing / Reviewing

Ensure focus returns to the button when closing either `Modal` / `ComposedModal` via keyboard or mouse click

### DIFF
--- a/e2e/components/Modal/Modal-test.avt.e2e.js
+++ b/e2e/components/Modal/Modal-test.avt.e2e.js
@@ -31,12 +31,12 @@ test.describe('Modal @avt', () => {
       },
     });
 
+    const button = page.getByRole('button', { name: 'Launch modal' });
+
     // Open the modal via keyboard navigation
     await page.keyboard.press('Tab');
-    await expect(
-      page.getByRole('button', { name: 'Launch modal' })
-    ).toBeFocused();
-    page.getByRole('button', { name: 'Launch modal' }).press('Enter');
+    await expect(button).toBeFocused();
+    button.press('Enter');
 
     // The first interactive item in the modal should be focused once the modal is open
     await expect(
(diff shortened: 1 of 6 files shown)


andreancardona (approved):
Other than the AVT test failing - looks good. I wonder if we can just specify that test to get it to pass?

tw15egan:
@andreancardona updated the tests to account for the new behavior

tay1orjones (approved):
Thanks for updating both Modal and ComposedModal 👍

Issue #14261, opened 2023-07-20
[a11y]: DataTable [dynamic] - SR table Row announcement


andreancardona:
### Package

@carbon/react

### Browser

FF & Safari & iPhone

### Package version

@carbon/react@1.32.0

### Automated testing tool and ruleset

Manual QA

### Assistive technology

JAWS & VO

### Description
SR table Row announcement

SR announces the number of Rows incorrectly when the table is focused

When the table is focused, SR announces the number of columns in the table incorrectly. It appears that the SR counts the visually hidden rows as well.

### Recommended Fix
Screen reader is expected to identify the correct number of Rows in the table

If the table structure is not conveyed properly, it’s difficult for screen reader users to navigate through the table content and correlate data. 

### WCAG 2.1 Violation

1.3.1 Info and Relationships

### Reproduction/example
Record `Number 64` in the below table:
https://airtable.com/shrFsaeHhImOkuVNR

https://react.carbondesignsystem.com/?path=/story/components-datatable-dynamic--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

francinelucca:
Announcement of row numbers: press “t” to go to table. Reads out 8 columns and 13 rows when there are less than 13 rows. This is probably due to the expandable content that is hidden being recognized as a row

Issue #14262, opened 2023-07-20
[a11y]: DataTable [dynamic] - expand/collapse control present in each row


andreancardona:
### Package

@carbon/react

### Browser

FF & Safari & iPhone

### Package version

@carbon/react@1.32.0

### Automated testing tool and ruleset

Manual QA

### Assistive technology

JAWS & VO

### Description
Expand/collapse control present in each Row

The Show/Hide content is identified by SR even when it is collapsed

The Show/Hide content is identified by SR even when it is collapsed. Note: This issue is observed in "expansion" and Batch Expansion  variants as well.

### Recommended Fix
SR is not expected to identify the visually hidden content. Hide the content and make it available for SR once its expanded.

If the screen reader identifies the visually hidden content, it tends to mislead users who are blind. 

### WCAG 2.1 Violation

4.2.1 Roles, Names and Values

### Reproduction/example
Record `Number 66` in the below table:
https://airtable.com/shrFsaeHhImOkuVNR

https://react.carbondesignsystem.com/?path=/story/components-datatable-dynamic--default

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
I think fixing this may also fix https://github.com/carbon-design-system/carbon/issues/14261

PR #14357, opened 2023-07-31
test(datatable): add @avt tests

### LINKED ISSUE #14118: Add AVT coverage for `DataTable` states: Selection, Expanded/collapsed state

tay1orjones:
Closes #14118 

#### Changelog

**New**

- add new avt tests for every datatable storybook story

**Changed**

- updated a few stories to pass accessibility-checker

#### Testing / Reviewing

- Checks should pass, double check the tests make sense


tw15egan:
@tay1orjones should we include some basic keyboard navigation tests here as well? Can probably capture most of them in the `Dynamic` story

tay1orjones:
@tw15egan Good call, I just pushed up a big block of keyboard nav related tests. This PR also now updates to the latest version of playwright so that we can use `.all()` to iterate through multiple elements if more than one matches a query. If testing locally (and when developing after this PR is merged) you'll need to run `yarn install` and then `npx playwright install`.

e2e/components/DataTable/DataTable-test.avt.e2e.js
```diff
@@ -0,0 +1,373 @@
...
+      ).not.toBeChecked();
+
+      // Pressing space should check the select all checkbox
+      await page.keyboard.press('Space');
+      await expect(
+        page.getByRole('checkbox', { name: 'Select all rows' })
+      ).toBeChecked();
+
+      // Every checkbox should be checked
+      for (const checkbox of await page.getByRole('checkbox').all()) {
```
tw15egan, line 120: This is super handy 👏🏻

e2e/components/DataTable/DataTable-test.avt.e2e.js
```diff
@@ -0,0 +1,373 @@
...
+      // Navigate to the first expansion button
+      await page.keyboard.press('Tab');
+      await expect(
+        page.getByRole('button', { name: 'Expand current row' }).first()
+      ).toBeFocused();
+      // Expand the first row
+      await page.keyboard.press('Space');
+      await expect(
+        page.getByRole('heading', { name: 'Expandable row content' }).first()
+      ).toBeVisible();
```
tw15egan, line 143: This test seems to pass regardless if the expandable row is toggled or not since the expandable row is always in the DOM. I'm guessing this is so we can animate the expansion, but I'm wondering if we should be toggling `aria-expanded` on `data-child-row` elements so that we can test this a bit better. 

Related: https://github.com/carbon-design-system/carbon/issues/14261

e2e/components/DataTable/DataTable-test.avt.e2e.js
```diff
@@ -0,0 +1,373 @@
...
+
+      // Navigate to the first row selection checkbox and check it
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Space');
+      await expect(page.getByText('1 item selected')).toBeVisible();
+
+      // Navigate backwards up into the batch action bar
+      await page.keyboard.press('Shift+Tab');
+      await page.keyboard.press('Shift+Tab');
```
tw15egan, line 152: Seems like two of these tab stops are required because the Search and Settings are focusable even when the batch action toolbar is active. Would it be worth spinning this into a separate issue to toggle `aria-hidden` and a `tabIndex` change based on `batchActionProps.shouldShowBatchActions`?

I was able to remove 1 tab stop by passing in a custom `tabIndex` to `TableToolbarSearch` and `TableToolbarMenu` in the `Dynamic` story like so:

```jsx
<TableToolbarMenu
   aria-hidden={ batchActionProps.shouldShowBatchActions ? true : false }
    tabIndex={ batchActionProps.shouldShowBatchActions ? -1 : 0 } />
```

But it doesn't seem to be passing properly to `TableToolbarSearch`. Small issue, but just wanted to note that

tw15egan (approved):
Looks great! Just had a few comments that can be addressed separate to this PR

e2e/components/DataTable/DataTable-test.avt.e2e.js
tay1orjones: Yeah I noticed this as well. I've actually got https://github.com/carbon-design-system/carbon/issues/14262 in my queue for this sprint, which I think will fix #14261 as well.

tw15egan: Awesome 👍🏻

alisonjoseph (approved):

e2e/components/DataTable/DataTable-test.avt.e2e.js
tay1orjones: Going to take care of this one in https://github.com/carbon-design-system/carbon/pull/14386

Issue #14361, opened 2023-08-01
[Bug]: Data Table Skeleton Set ClassName Props as Optional


Husnain062:
### Package

@carbon/react

### Browser

Chrome

### Package version

11.17.1

### React version

18.2.0

### Description

Some of the props are already set as optional and fixed in the PR below but className is still required that needs to be optional.

https://github.com/carbon-design-system/carbon/pull/14287

### Reproduction/example

Use the DataTableSkeleton Component

### Steps to reproduce

Use the `DataTableSkeleton` Component and try to use it without passing className props. 

### Suggested Severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

PR #14372, opened 2023-08-01
fix(DataTableSkeleton): fix TypeScript interface 

Husnain062:
DataTableSkeleton has a field className which should be optional:

**Changed**

- Set the className field optinal.
https://github.com/carbon-design-system/carbon/issues/14361


Husnain062:

guidari (approved):

tay1orjones (approved):
Thank you!

Issue #14294, opened 2023-07-24
Implement a Github triage bot with the ability to comment based on label triggers


aubrey-oneal:
In issue [#3622](https://github.com/carbon-design-system/carbon-website/issues/3622), we created designs for a simple Github Triage that relies on labels to trigger commenting. This issue to build the first version of this bot. 

[Box note](https://ibm.ent.box.com/notes/1270512433618) with labels and triggers

Resources:
- https://docs.github.com/en/actions/using-workflows/about-workflows
- https://github.com/marketplace/actions/respond-to-issue-based-on-label#create-workflow
- https://github.com/marketplace/actions/label-actions
- https://github.com/dessant/label-actions

PR #14378, opened 2023-08-02
feat: testing enhancement github bot

andreancardona:
Part of #14294 

This is to test the github bot - this is just the first of a series of tests

Worked on this with @garrigoose 🎉


.github/label-actions.yml
```diff
@@ -0,0 +1,12 @@
+# Configuration for Label Actions - https://github.com/dessant/label-actions
+
+# The `enhancement` label is added to issues
+enhancement:
+  # Post a comment
+  comment: >
+    Thank you for submitting a feature request. Here’s what to expect next. 
+    Your proposal is open and is now being [triaged](https://github.com/orgs/carbon-design-system/projects/51/views/1) by the Carbon team.
```
tw15egan, line 8: Should this also add the `proposal: open` label?

andreancardona: Oo I like that - updating now!

andreancardona: hmm I actually don't see an immediate way to do that - but this PR is just a test PR - so that's something we can play with

tw15egan (approved):

PR #14397, opened 2023-08-04
fix(datatable): remove toolbar search and menu from tab order

tay1orjones:
This removes the table toolbar search and menu actions from the tab order when batch actions are open, as discussed in https://github.com/carbon-design-system/carbon/pull/14357#discussion_r1281907787

#### Changelog

**Changed**

- update Search to put `...rest` after other props
- update stories to remove TableToolbarSearch and TableToolbarMenu actions from tab order when batch action bar is open/visible

#### Testing / Reviewing

* focus should move through the actions when the batch action bar is closed
* focus should not move through the actions when the batch action bar is open

https://github.com/carbon-design-system/carbon/assets/3360588/b7593fe3-5e58-4628-8dba-6fd184dda7b6

### DIFF
--- a/packages/react/src/components/Search/Search.tsx
+++ b/packages/react/src/components/Search/Search.tsx
@@ -243,7 +243,6 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
         {labelText}
       </label>
       <input
-        {...rest}
         autoComplete={autoComplete}
         className={`${prefix}--search-input`}
         defaultValue={defaultValue}
@@ -257,6 +256,7 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
         type={type}
         value={value}
         tabIndex={onExpand && !isExpanded ? -1 : undefined}
+        {...rest}
       />
       <button
         aria-label={closeButtonLabelText}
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Search/Search.tsx
```diff
@@ -257,6 +256,7 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
         type={type}
         value={value}
         tabIndex={onExpand && !isExpanded ? -1 : undefined}
+        {...rest}
```
tay1orjones, line 259: I'm not sure why we had this placed before all the existing props. Props passed by consumers would always be overwritten by what we had here.

It's possible this could cause some unintended effects downstream, but I don't think it's necessarily a breaking change. Thoughts?

tw15egan: Is the only "unintended effect" that props users were passing and expecting to work will now actually work? 😅

tw15egan (approved):
LGTM 👍 ✅

tw15egan:
@tay1orjones just needs updated snaps

alisonjoseph (approved):
LGTM!

packages/react/src/components/Search/Search.tsx
alisonjoseph: I don't think this is breaking either, more like a bug fix for something that should have been working.

Issue #14415, opened 2023-08-10
[Bug]: New version of SASS package does not work with Carbon (compatibility issue)


thanusiak:
### Package

carbon-components

### Browser

Chrome

### Package version

10.58.8

### React version

v18,v20

### Description

New version of SASS:
https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1651

seems to break the rem() function defined by Carbon.

This results in compilation problems during 'rpm run build'

SassError: 2 arguments required, but only 1 was passed.
   ╷
42 │     min-height: rem(40px);
   │                 ^^^^^^^^^
   ╵
  node_modules/carbon-components/scss/components/accordion/_accordion.scss 42:17                                   accordion()
  node_modules/carbon-components/scss/components/accordion/_accordion.scss 254:3                                   @content
  node_modules/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/import-once/import-once.scss 23:5  exports()
  node_modules/carbon-components/scss/components/accordion/_accordion.scss 253:1                                   @import
  node_modules/carbon-components/scss/globals/scss/styles.scss 119:9                                               @import
  src/App.scss 12:9

### Reproduction/example

Problem is during compilation

### Steps to reproduce

Import sass (https://www.npmjs.com/package/sass) in your project - its a dependency of sass-loader (for example).
Make sure you get the latest version (1.65.1).

Import
@import "~carbon-components/scss/globals/scss/styles.scss";
in your scss file

Try executing npm run build

Temporary w/a:

set:
 "sass": "1.64.2"
 in your override

### Suggested Severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

huineng:
yes, facing same issue

tw15egan:
Thanks for pointing this out! Seems like the issue is being caused by a namespace collision when we are importing _our_ rem function in like `@use '../../utilities/convert' as *;`. I can resolve this by changing our imports to `@use '../../utilities/convert'` and then calling the function directly i.e. `convert.rem(24px)`.

I'll work on a PR to fix this today.

[...]

PR #14420, opened 2023-08-10
fix(@carbon/styles): change `rem` function to `to-rem`

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/14415
Refs https://github.com/sass/dart-sass/issues/2059

Namespaces all `rem` functions to `convert.to-rem` used across the repo. ~Also, bumps `sass` to the latest version~ We'll bump this once the v10 fix is made so we can update `incremental-migration` example.This version bump was causing issues due to the change with `rem` upstream expecting two parameters https://github.com/sass/dart-sass/releases/tag/1.65.0

#### Changelog

**Changed**
- `rem` function changed to `to-rem`
- All `@use './utilities/convert' as *;` to `@use './utilities/convert';`
- All instances of `rem(` have been changed to `convert.to-rem(`

#### Testing / Reviewing

Ensure there are no visual changes and all sass files compile successfully.


tay1orjones (approved):
Could we also add some sassdoc comments where the `rem` function is declared, and then regenerate the sassdocs for the styles package? It might also be helpful if that comment lists the error message directly so searching the error returns those comments talking about how we have to use a namespace on `rem()`

Just trying to think about how we could try to prevent folks from asking about this in support, etc.

alisonjoseph (approved):

nelsonchen90:
Thanks, I hope this will be backport to v10 as well since the first ref issue is created for v10

tw15egan:
@nelsonchen90 yeah, we'll definitely get it backported as well since this will affect all users

PR #14423, opened 2023-08-11
fix(rem): add rem function back

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/14420
Refs https://github.com/carbon-design-system/carbon/issues/14415

Adds back in the `rem` function so that downstream users who have _not_ upgraded to `sass@1.65.0` are not broken when the new package is released. When they update to the new version of sass, they will simply need to change their usage of `rem` to `to-rem`

#### Changelog

**New**

- Adds back in the old `rem` functions with a `@depracted` comment

#### Testing / Reviewing

Ensure our build still passes and there is no usage of the `rem` function in our repo


tay1orjones (approved):

andreancardona (approved):
🎉

PR #14460, opened 2023-08-18
fix: slider to range step values

### LINKED ISSUE #14455: [Bug]: Slider with fractional values has javascript math error
### Package

@carbon/react

### Browser

Chrome

### Package version

1.35.0

### React version

18

### Description

If the step value is 0.3 and step multiplier 1, range 0 - 100 invalid values can be achieved.

### Reproduction/example

N/A

### Steps to reproduce

Steps
1. Drag the handle to 0.
2. Press up (value now 0.3)
3. Press up (value now 0.6)
4. Press up (value now 0.89999999999)

The simplest way to avoid this is to set the value of the input directly in javascript, the immediately read it back to the internally maintained value. 

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

lee-chase:
Closes #14455 

Javascript floating point maths is a bit problematic.

0.3 + 0.3 + 0.3 evaluates to 0.89999999999'

The range control does not normally have a problem with this, ensuring the nearest fixed length value is used.

While this could be fixed by replicating the math inside the standard input control, the simplest solution is to set the value of a range control and read it back. This PR does just that.

#### Changelog

**Changed**

- packages/react/src/components/Slider/Slider.tsx

#### Testing / Reviewing

1. Run Storybook locally and selecting the `Slider` component.
2. Set `Step` to 0.3.
3. Drag the handle to the value 0.
4. Press the up arrow 3 times.

The result should be 0.9 but is 0.8999999' due to JS floating point math.

Applying this PR we can see the value is 0.9.


tay1orjones (approved):
Clever to use the browser's built-in rounding to help with this situation. I like it! 👍

alisonjoseph (approved):

PR #14463, opened 2023-08-18
fix(numberinput): use native methods for stepping the value up/down

### LINKED ISSUE #6684: NumberInput and floating point values

- [x] `carbon-components`

## Detailed description

I was trying to use the NumberInput component and I found out that using the up and down buttons instead of the arrow keys on my keyboard (thus triggering the custom function that changes the value) causes the classic approximation problem of the floating point binary representation.

The native `<input type="number">` isn't affected by this issue because browsers take into account the inaccuracy of the binary representation and implement ways to avoid the problem.

- Google Chrome
- Android WebView

It should be noted that Firefox doesn't seem to be affected by this problem, at least visually

- `carbon-components` v10.17.0
- `carbon-components-svelte` v0.9.6

1. Add a NumberInput component to a page
2. Set the step property to 0.01
3. Press the increment button a few times

CodeSandbox: https://codesandbox.io/s/flamboyant-hellman-2cdnf

## Additional information

A possible solution to fix the problem could be to temporary treat the operands as integers, as presented in the following code, in the function that updates the value of the input when the arrow buttons are clicked:
```
const stepDecimalPlaces = step % 1 === 0 ? 0 : step.toString().split('.')[1].length;
const nextValue = stepDecimalPlaces === 0
    ? value + step * direction // integer values are not affected
    : Math.round(value * Math.pow(10, stepDecimalPlaces) + step * Math.pow(10, stepDecimalPlaces) * direction) / Math.pow(10, stepDecimalPlaces);
```

### LINKED ISSUE #12897: [Bug]: NumberInput empty input results in "The specified value "NaN" cannot be parsed, or is out of range."
### Package

@carbon/react

### Browser

Chrome

### Package version

1.19.0

### React version

16.13.1

### Description

Currently, on the new version of @carbon/react for the `NumberInput`, if you delete the content of the input and then use the + or - buttons, you get the following warning:
`The specified value "NaN" cannot be parsed, or is out of range.`

I think this is happening because `parseInt` is called on an empty string:
https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/NumberInput/NumberInput.js#L220

The previous version (prior to Carbon v11) didn't have this issue. You could clear the input and click the + button and it will treat the empty input as having a value of 0 while rendering an empty input, meaning that it will just increment the value by 1.
Right now if you try to force the value to 0 on an empty string this solves the buttons issue but creates a bad user experience since it inserts a 0 on the input when you might want just an empty input.

### Suggested Severity

### Reproduction/example

https://codesandbox.io/s/numberinput-i2lvtg?file=/src/App.js

### Steps to reproduce

1.- Clean the number input

2.- Open the browser console

3.- Click on either the + or - buttons

### Code of Conduct

tay1orjones:
Closes #6684 
Closes https://github.com/carbon-design-system/carbon/issues/12897

#### Changelog

**Changed**

- Use native `stepUp()` and `stepDown()` methods to step the value up and down

#### Testing / Reviewing

- Go to the NumberInput playground story, and enter a step of `0.01`
- Increment the value with the `+` stepper many times
- Validate that the input only displays even values incrementing by `0.01`
- We should no longer see the invalid behavior of the input getting values with rounding errors of multiple decimal places, like `50.999999998`


alisonjoseph (approved):
🙌 nice!

guidari (approved):
LGTM!! 🚀

Looks like we just need to remove the `LegacyRef` prop, so the lint is good to go!

tw15egan:
@tay1orjones this is awesome 🙌🏻

tw15egan:
This will also close https://github.com/carbon-design-system/carbon/issues/12897

PR #14441, opened 2023-08-15
fix(tile): support rtl direction

janhassel:
Part of https://github.com/carbon-design-system/carbon/issues/13619

Adds support for right-to-left to Tile component.

**Example: RadioTile**

Before
image

After
image

#### Changelog

**Changed**

- Updated `top`, `right`, `bottom`, `left` to respective css logical properties

#### Testing / Reviewing

1. Add `dir="rtl"` to the root `html` tag in storybook
2. Test all Tile stories (incl. experimental ones)

PR #14464, opened 2023-08-18
docs(storybook): add text direction auto/ltr/rtl switcher

### LINKED ISSUE #13620: Provide Right-to-left (RTL) mode/toggle in storybook
### The problem

There's currently no way to easily change the direction of the storybook canvas to test or view components in `rtl` mode.

### The solution

Provide a way to toggle `rtl` mode within the storybook canvas. It shouldn't impact the rest of the page.

Before we can do this we either need to run the styles through RTLCSS and include them in storybook or wait until we've refactored to use CSS Logical Properties and Values https://github.com/carbon-design-system/carbon/issues/13619

### Examples

There's some addons out there, but nothing officially from storybook. We might consider using one of these or rolling our own if we find it doesn't work as we'd like. 

- https://storybook.js.org/addons/storybook-i18n/

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

tay1orjones:
Closes #13620 

This PR improves our storybook toolbar to have a `Text direction` toolbar item that influences the `dir` attribute on the `html` element that wraps the story.

image

#### Changelog

**New**

- Add a `Text direction` toolbar item to storybook

**Changed**

- Provide a `title` for our custom toolbar items in storybook to prevent confusion and improve discoverability
- add `dev ::` prefix for toolbar items that are only visible in dev

#### Testing / Reviewing

- Go to a tile story, select `rtl` or `ltr`. 
- The component should reflow to display properly when rtl vs ltr is selected thanks to recent work in https://github.com/carbon-design-system/carbon/pull/14441
- Other components that have been updated to use logical properties should also render appropriately as it switches.

### DIFF
--- a/packages/react/.storybook/preview.js
+++ b/packages/react/.storybook/preview.js
@@ -29,7 +29,7 @@ const devTools = {
     description: "Set the layout context's size",
     defaultValue: false,
     toolbar: {
-      title: 'unstable__Layout size',
+      title: 'dev :: unstable__Layout size',
       items: [
         {
           value: false,
@@ -48,7 +48,7 @@ const devTools = {
     description: "Set the layout context's density",
     defaultValue: false,
     toolbar: {
-      title: 'unstable__Layout density',
+      title: 'dev :: unstable__Layout density',
       items: [
         {
           value: false,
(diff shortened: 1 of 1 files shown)


andreancardona (approved):
Tested and looks good! So cool 😎

alisonjoseph (approved):
🔥 Thanks for adding this!

Issue #14235, opened 2023-07-18
[Developer Experience]: 2Gb transfer for repo clone is out of proportion


dbrugger:
### The problem

As a developer who uses the library I sometimes stumble on issues, that I typically workaround with `patch-package` but also like to submit PRs for it. Even if it's just a small fix, using the GitHub editor alone is not sufficient because I have to run the `format` script on it. So I have to clone the repo (fork), but this is a 2Gb transfer, which is kind of out of proportion.

I am on gigabit fiber myself, and am still annoyed by this. I can imagine that this is a turn off for developers with less bandwidth, and can impact attracting new potential contributors.

### The solution

Somehow get rid of the yarn cache.

### Business priority

None

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

tay1orjones:
Yeah I agree it would be nice for these types of use cases if the repo size could be smaller. Most contributors clone and only pay this price once, issuing further updates from their local clones. If we ever want to take advantage of codespaces or similar virtual coding environments this might be of more concern, I'm not sure.

We're not way out of bounds though. [Github recommends](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#repository-size-limits) under 1GB, we're at 2GB, and the max allowed is 5GB. 

> Somehow get rid of the yarn cache.

From our testing this is only about ~500mb and ensures dependency integrity locally and in CI. Beyond this it would be interesting to use [`github/git-sizer`](https://github.com/github/git-sizer) to try and see where the problems are contributing to the size. My guess is it's just raw history as the project has been around for quite some time.

sstrubberg:
@dbrugger Thanks for bringing this to our attention. Unfortunately, it isn't a high priority for us at the moment. We'd welcome any research or PR that might help though. Closing for now.

PR #14470, opened 2023-08-19
chore: remove .yarnrc leftover

dbrugger:
When looking into a solution for https://github.com/carbon-design-system/carbon/issues/14235, I saw carbon is on yarn v3 but still has a `.yarnrc` floating around. This file is now ignored according to https://yarnpkg.com/migration/guide#update-your-configuration-to-the-new-settings

#### Changelog

**Removed**

- `.yarnrc`

#### Testing / Reviewing

- test local development flow


tay1orjones (approved):
My previous understanding of this was that we still needed this file because Netlify only formally supported yarn v1/classic. By incrementing the version in the `.yarnrc` we were sort of tricking Netlify into allowing our deploy previews to build.

Looking into it again though, I see https://github.com/netlify/build-image/issues/779 has now been resolved but there are still a couple outstanding issues as it relates to yarn berry/v3

* https://github.com/netlify/build-image/issues/615
* https://github.com/netlify/build/issues/1535

The deploy previews did succeed on this PR, so that's a good sign.

@carbon-design-system/developers-system I think it's worth trying to merge this and if we have issues with Netlify deploy previews we may need additional config. Or we could roll back to put `yarnrc` back in.

dbrugger:
i also find yarn versioning confusing. I recently updated one of our projects to v3, and it took a while for me to understand that yarn consists of two parts:
* the yarn cli on the system, which always is v1
* the yarn code in .yarn/releases which in this case is v3
on my system when I run `yarn --version` in the carbon repo, it shows `3.6.1`. when I run it in `~`, it shows `1.22.19`.

tw15egan (approved):

PR #14482, opened 2023-08-21
fix(components): guard event.persist() calls

tay1orjones:
[Raised in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1692626599191779?thread_ts=1689560271.450069&cid=C2K6RFJ1G)

Even though `persist()` is available and just a noop in later versions of react that don't have event pooling anymore, we still need to guard calls to these. It was pointed out that if we begin to call `persist()`, it can break tests where event handlers are mocked. Example:

> On our app we mock onChange function passed to the textArea like:
> ```js
> let onChange = jest.fn();
>     let component = getComponent('hello', onChange);
>     component
>       .find('textarea')
>       .props()
>       .onChange({ target: { value: '   test123' }, persist: jest.fn() });
> ```
> I had to add persist: jest.fn() into this bcs of that change
> 

#### Changelog

**Changed**

- modify all `event.persist()` calls to be guarded via `event?.persist?.();`

#### References

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/react/src/components/ComboBox/ComboBox.tsx
@@ -411,7 +411,7 @@ const ComboBox = React.forwardRef((props: ComboBoxProps, ref) => {
 
     if (event.target === textInput.current && isOpen) {
       event.preventDownshiftDefault = true;
-      event.persist();
+      event?.persist?.();
     }
   };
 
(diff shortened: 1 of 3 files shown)


darq37 (approved):
LGTM

guidari (approved):
LGTM!

tw15egan (approved):

PR #14483, opened 2023-08-21
[Docs] - Added new docs and story to Link component

### LINKED ISSUE #13808: Improve link + icon accessibility
**Ref discussion:** [#3134](https://github.com/carbon-design-system/carbon-website/discussions/3134)

## Improve link + icon accessibility

- [ ] When using [link (paired with icon)](https://react.carbondesignsystem.com/?path=/story/components-link--paired-with-icon), we need to add a way for the icon (ex: an ALT) to be read and understood by assistive technologies.

- [ ] Provide guidance to the user of when an ALT is needed for the icon and how to implement it. (*Note: Is this any easier with a button icon? In the case of a button, an icon added beside a text label is always superfluous. The button's text is the accessible name.)

---

**Example**
- Where the link text is called "Download" with a "Download" icon, it's redundant. But imagine an example where the link text is called "My trip diary" with a "Download" icon beside it, that is a VERY different outcome (download my diary) than one without the icon (take me to a page that is my diary).

Frame 1

### Code of Conduct

guidari:
Closes #13808

Added a new story to exaplain how to insert an `arai-label` in Link paried with icon.

#### Changelog

- New story;
- New docs in mdx.

#### Testing / Reviewing

Ensure the VO reads the `aria-label` and the docs are well explained.


guidari:
Obs: I also have another suggestions.

Instead of having the `PairedWithIcon` and `WithAcessibleAriaLabel` stories, can we have only one called `PairedWithIcon` with an acessible example? Or should we keep both? 🤔

packages/react/src/components/Link/Link.stories.js
```diff
@@ -67,6 +67,12 @@ export const PairedWithIcon = () => (
   </Link>
 );
 
+export const WithAcessibleAriaLabel = () => (
```
tw15egan:
```suggestion
export const WithAccessibleAriaLabel = () => (
```

packages/react/src/components/Link/Link.mdx
```diff
@@ -27,6 +27,16 @@ links, which should be used sparingly.
   <Story of={LinkStories.Default} />
 </Canvas>
 
+## Acessible Link with Icon
```
tw15egan:
```suggestion
## Accessible Link with Icon
```

packages/react/src/components/Link/Link.mdx
```diff
@@ -27,6 +27,16 @@ links, which should be used sparingly.
...
+
+If the text of your link doesn't correspond to the action represented by the
+icon, consider incorporating an `aria-label` for the icon. This will ensure that
+visually impaired users can comprehend the outcome of clicking the link.
+
+<Canvas>
+  <Story of={LinkStories.WithAcessibleAriaLabel} />
```
tw15egan:
```suggestion
  <Story of={LinkStories.WithAccessibleAriaLabel} />
```

tw15egan (approved):
Looks great, just a small spelling correction

tay1orjones (approved):
Looks good, thanks for adding the docs for it in the `.mdx`!

> Instead of having the PairedWithIcon and WithAcessibleAriaLabel stories, can we have only one called PairedWithIcon with an acessible example? Or should we keep both? 🤔

The two stories are so similar, I think consolidating down to the one `PairedWithIcon` story makes sense. Particularly since the aria-label is only needed when an icon is paired with the link.

PR #14491, opened 2023-08-22
feat: test github action bot

andreancardona:
Part of https://github.com/carbon-design-system/carbon/issues/14294

This is to test the github bot - this is just the second of a series of tests  

Worked on this with @garrigoose 🎉

MERGING for testing purposes


.github/workflows/contribution-approved.yml
```diff
@@ -0,0 +1,28 @@
...
+name: Add comment on Enhancement Label
+on:
+  issues:
+    types:
+      - labeled
+jobs:
+  add-comment:
+    ## TODO: figure out how to escape colons for labels
+    # if: github.event.label.name == "proposal: accepted" && github.event.label.name == "community contribution"
+    if: "${{ contains(github.event.label, 'proposal: accepted') && contains(github.event.label, 'communit contribution') }}"
```
tay1orjones, line 18: I went down a big rabbit hole last month with finding valid yaml syntax that allows colons. What I discovered was we need to use `format()` on the label name. Like this:

```
contains(github.event.labels.*.name, format('proposal{0} accepted', ':'))
```

This replaces the first parameter's `{0}` with the value of the second parameter `:`

Here's how it looks in what I delivered: https://github.com/carbon-design-system/carbon/blob/f838aaab7668c38e21636e25569c6da0707a0ad7/.github/workflows/add-to-merge-queue.yml#L21-L25

.github/workflows/contribution-approved.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Add comment
+        uses: peter-evans/create-or-update-comment@5f728c3dae25f329afbe34ee4d08eef25569d79f
+        with:
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            The Carbon team has accepted this proposal! Our team doesn’t have the cycles to work on this now, so we are requesting community contributors. Please see the labels for roles that are needed. If you are willing to help out, comment below and we will get in touch!
```
tay1orjones: Syntax may be slightly off here, but we could use the `github-script` action built by GitHub

```suggestion
        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 #v6.4.1
        with:
          script: |
            github.rest.issues.createComment({
              issue_number: context.issue.number,
              owner: context.repo.owner,
              repo: context.repo.repo,
              body: 'The Carbon team has accepted this proposal! Our team doesn’t have the cycles to work on this now, so we are requesting community contributors. Please see the labels for roles that are needed. If you are willing to help out, comment below and we will get in touch!'
            })
```

---

> If you are willing to help out, comment below and we will get in touch!

We won't be notified in any way of a plain comment, and so we probably won't get in touch in that situation. We could rephrase this into a couple options:

1. "If you are willing to help out, comment on this issue stating your interest and put your ideas or solution up into a pull request. The team will work with you to iron out any details. If you run into any issues or have questions, reach out on `[slack](link-to-internal-slack-channel)` or `[discord](link-to-discord-invite-from-readme-badge)`"
2. "If you are willing to help out, comment below and mention `@carbon-design-system/developers-system`. The team will get in touch."

For `#2`, we'll need to enclose the team `@mention` in backticks so it's escaped and doesn't actually mention the team in the comment posted from the action

tay1orjones (commented):
Looks pretty solid! Just a couple minor thoughts, these apply to both workflows

.github/workflows/contribution-approved.yml
garrikats: I was down that same rabbit hole this morning while working on this with Andrea. Is this a better way than the escape syntax below? It was not throwing any errors in the ide.

`"${{ contains(github.event.label, 'proposal: accepted') && contains(github.event.label, 'community contribution') }}"`

tay1orjones: @garrikats Yeah I tried it in this commit: https://github.com/carbon-design-system/carbon/commit/66004ba5f83caba029df13971a8d01b1e3001ced, and it still didn't work 😕 

Using `format()` was the only thing I found that worked. It's a shame because I think it's terrible for readability.

.github/workflows/contribution-approved.yml
garrikats: Happy to help out. Looking through the github-script docs to find the syntax for an action on label creation. Is the reason to use github-script instead of create-or-update comment to use a plugin from github instead of elsewhere (create-or-update is MIT maintained).

garrikats:
testing

PR #14531, opened 2023-08-29
refactor(@carbon/styles): use logical properties for better RTL support

### LINKED ISSUE #13619: Use CSS Logical Properties and Values for Styling
### The problem

Component styling currently uses typical properties for positioning and sizing, things like margin-top, padding-left, border-right, etc. 

These properties are static and are not impacted by the writing mode. When a page is rendered in `rtl`, padding-left does not automatically become padding-right. It is static on the left.

Due to this, we currently recommend that applications using Carbon with a requirement to support `rtl` mode use a postprocessor like [RTLCSS](https://rtlcss.com/). RTLCSS converts these static properties into their logical inverse property for `rtl` mode( `padding-left` becomes `padding-right`). The output stylesheet can then be loaded when the document is in `rtl` mode and override the default styles provided by Carbon.

We do not have RTLCSS as part of the pipeline within the monorepo. This means that if we were to alter the writing mode in the storybook canvas, the components would not have appropriate styling to be rendered correctly.

### The solution

We should instead use Logical Properties and Values to specify styling that will be impacted by the writing mode.

From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties):

> **CSS Logical Properties and Values** is a module of [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) introducing logical properties and values that provide the ability to control layout through logical, rather than physical, direction and dimension mappings.
> ...
> Logical properties and values use the abstract terms block and inline to describe the direction in which they flow. The physical meaning of these terms depends on the [writing mode](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Writing_Modes).

In effect, this means that we can specify things like padding, margin, text-alignment, etc. to be dependent on the writing mode rather than static right/left/top/bottom values.

### Examples

Here's an example from [this excellent article](https://web.dev/learn/css/logical-properties/) on the topic. It does a much better job of explaining all the facets of this than I can do here, please read. 

```scss
/* Without logical properties and values */
.my-element {
  padding-top: 2em;
  padding-bottom: 2em;
  margin-left: 2em;
  position: relative;
  top: 0.2em;
}

/* With logical properties and values */
.my-element {
  padding-block-start: 2em;
  padding-block-end: 2em;
  margin-inline-start: 2em;
  position: relative;
  inset-block-start: 0.2em;
}
```

```[tasklist]
- [ ] Make a new WIP branch for everyone to work off of (or do work against main if we can keep the stylelint rule to warning instead of error)
- [ ] Add a stylelint rule to enforce logical properties https://github.com/csstools/stylelint-use-logical
- [ ] Phase 1 mob programming session (do as many as possible in a timeboxed meeting together in a single sprint) - make changes to the WIP branch
- [ ] Phase 2 mob session (if needed)
- [ ] Phase ... mob sessions (if needed)
- [ ] Merge WIP branch into main
```

With this mapping I think we could get most (all?) of the way there in a single sprint https://github.com/csstools/stylelint-use-logical#property-and-value-mapping

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13619

Updates the repo to use CSS Logical Properties. This will help users render the components in RTL mode without the need for post-processors like RTLCSS

#### Changelog

**New**

- `stylelint-use-logical` added to `stylelint-config-carbon`
- Enabled linting to ensure logical properties are used going forward

**Changed**

- Updated all layout styles to use logical properties rather than `left`, `right`, `top`, etc...

#### Testing / Reviewing

Ensure there are no visual regressions. Select `rtl` via the `Text direction` dropdown via storybook and ensure components render as expected.

I have encountered one issue with `Popover`, which we can address in a separate PR. Since we use `bottom-right`, `top-left` etc, these styles are broken when RTL mode is enabled


alisonjoseph:
This looks great! Had a couple questions around what components should look like in rtl, these may be unrelated to this PR

Should accordion title be right aligned?
Screenshot 2023-08-29 at 2 09 23 PM

- [ ] Breadcrumb with overflow 
Screenshot 2023-08-29 at 2 11 06 PM

- [ ] Button
Screenshot 2023-08-29 at 2 11 38 PM

- [ ] Button set is missing the border 
Screenshot 2023-08-29 at 2 12 06 PM

- [ ] Checkbox missing spacing
Screenshot 2023-08-29 at 2 17 15 PM

- [ ] Codesnippet
Screenshot 2023-08-29 at 2 17 41 PM

Screenshot 2023-08-29 at 2 18 09 PM

- [ ] Combobox
Screenshot 2023-08-29 at 2 18 56 PM

- [ ] Datatable missing spacing to the right of text
Screenshot 2023-08-29 at 2 21 34 PM
same issue with skeleton state

Screenshot 2023-08-29 at 2 23 13 PM

- [ ] Spacing seems off on Dropdown
Screenshot 2023-08-29 at 2 24 15 PM

- [ ] Inline dropdown
Screenshot 2023-08-29 at 2 24 40 PM

- [ ] Icon button
Screenshot 2023-08-29 at 2 25 30 PM

- [ ] should the ... be after for inline loading?
Screenshot 2023-08-29 at 2 25 54 PM

- [ ] Is menu correct?
Screenshot 2023-08-29 at 2 26 41 PM

- [ ] MenuButton
Screenshot 2023-08-29 at 2 27 26 PM


... will keep reviewing starting with Modal later

tw15egan:
@alisonjoseph anything that uses `Tooltip` or `Popover` will be broken. I'm more concerned with making sure the components aren't broken in LTR. I think it might be best to fix the RTL-specific issues in a separate PR, what do you think? I just don't want to make this PR any harder to review

alisonjoseph:
@tw15egan Agree this is already a huge PR. My only worry is that if teams are using RTL, and then we release this they will have broken components. I wonder if we could just keep this branch and open PR's to fix things into this branch, and then release all at once?

tw15egan:
Good point, I guess we should create a new branch and we can work off that until we can fix these issues

tw15egan:
@alisonjoseph this is now pointing to a new `logical-properties` branch, should we merge this and push new fixes for the RTL issues to that branch? We can create an issue for outstanding issues we come across

alisonjoseph:
@tw15egan sounds good, lets merge it 👍

Issue #14544, opened 2023-08-30
Fix outstanding issues in RTL mode after switching to CSS Logical Properties


tw15egan:
### Package

@carbon/styles

### Package version

Latest

### Description

Refs https://github.com/carbon-design-system/carbon/pull/14531

List of issues that need to be fixed on the `logical-properties` branch. These are regressions introduced when RTL mode is enabled after switching to CSS Logical Properties. More to be added

Should accordion title be right aligned?
<img width="1429" alt="Screenshot 2023-08-29 at 2 09 23 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/d0b836e1-309a-4c8a-9ca8-bd5ace52de68">

- [x] Breadcrumb with overflow 
<img width="578" alt="Screenshot 2023-08-29 at 2 11 06 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/e831cc7b-41d1-4075-86fc-8a40d4240167">

- [x] Button
<img width="296" alt="Screenshot 2023-08-29 at 2 11 38 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/b7b83310-fd06-4f10-b108-422026ff0d56">

- [x] Button set is missing the border 
<img width="512" alt="Screenshot 2023-08-29 at 2 12 06 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/8917e94e-92f0-4143-ae53-184712e8e159">

- [x] Checkbox missing spacing
<img width="224" alt="Screenshot 2023-08-29 at 2 17 15 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/c7423d37-7a19-4223-bc44-8b264a78bba1">

- [x] Codesnippet
<img width="348" alt="Screenshot 2023-08-29 at 2 17 41 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/91cb938a-045c-4a92-aa1a-50d226d29364">

<img width="372" alt="Screenshot 2023-08-29 at 2 18 09 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/47223d23-8978-4d8e-ad6e-16f602e7499e">

- [x] Combobox
<img width="397" alt="Screenshot 2023-08-29 at 2 18 56 PM" src="https://github.com/carbon-design-system/carbon/assets/2753488/610f97de-2350-46af-aacf-f5266efc51ee">

- [x] Datatable missing spacing to the right of text
[...]

PR #14546, opened 2023-08-30
[WIP] fix(components): fix RTL issues with components

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14544

Fixes to improve RTL support after switching to CSS Logical properties. WIP

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/packages/react/.storybook/preview.js
+++ b/packages/react/.storybook/preview.js
@@ -306,20 +306,23 @@ const parameters = {
 const decorators = [
   (Story, context) => {
     const { layoutDensity, layoutSize, locale, dir, theme } = context.globals;
+    const [randomKey, setRandomKey] = React.useState(1);
 
     React.useEffect(() => {
       document.documentElement.setAttribute('data-carbon-theme', theme);
     }, [theme]);
 
-    React.useEffect(() => {
+    React.useLayoutEffect(() => {
       document.documentElement.lang = locale;
       document.documentElement.dir = dir;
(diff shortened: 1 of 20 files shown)


tw15egan:
@alisonjoseph latest changes should fix most of the issues, except: `Menu`, `MenuButton`, and `TreeView`

alisonjoseph (commented):
These all look great! 🔥Popover autoalign is the only one I noticed that still looked off (other than Menu/MenuButton)

Screenshot 2023-08-31 at 12 20 51 PM

tw15egan:
Ugh, yeah it seems like that is also calculating which property to apply in js. Want to tackle that in a separate PR?

tw15egan:
Looks like we can do something like this in `Popover` to determine if the page is in RTL. Next, need to figure out what changes need to be made 😄 

```js
    if (typeof document !== 'undefined' && document.dir) {
      document.dir === 'rtl' ? setRtl(true) : setRtl(false);
    }
```

tw15egan:
@alisonjoseph I think the latest changes fixed it without needing to change anything inside `Popover`. We just needed to refresh the page when the modes were switched, as well as update the storybook styles to use logical properties as well

alisonjoseph (approved):
🎉

PR #14547, opened 2023-08-30
[FilterableMultiSelect] Fixed input when pressed delete

### LINKED ISSUE #14516: [Enhancement]: FilterableMultiSelect component clears the entire input filter text, when backspace/delete is pressed.
### The problem

The FilterableMultiSelect component clears the **entire input filter** text, when backspace/delete is pressed. This leads to an issue, when a user intends to just remove the last character or a few characters, to obtain more filter results to select from, but the whole input gets cleared. (Also leads to diminish the user experience)
The functionality can be observed in Carbon React Storybook.
[https://react.carbondesignsystem.com/?path=/story/components-multiselect--filterable](url)

Below is a preview, where "option" is typed, followed by pressing backspace. As noticed the entire text gets cleared, instead of clearing the last character "n" only (making the input "optio"), which would be more ideal.

https://github.com/carbon-design-system/carbon/assets/40992344/9598652c-8ddd-41b7-97c7-7956c951ef69

### The solution

The ideal functionality for the filter text in FilterableMultiSelect should be similar to that of a text input component, where upon pressing backspace/delete, only the last input character entered should be cleared/deleted.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

guidari:
Closes #14516

Previously, when the user tried to delete a character, the entire input would be cleared. Now, the user can delete only one character at a time after pressing the delete key.

#### Testing / Reviewing

Test the variant `FilterableMultiSelect` to validate if it works as expected.

### DIFF
--- a/packages/react/src/components/MultiSelect/FilterableMultiSelect.js
+++ b/packages/react/src/components/MultiSelect/FilterableMultiSelect.js
@@ -200,7 +200,10 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
   }
 
   function clearInputValue() {
-    setInputValue('');
+    textInput.current.value.length === 1
+      ? setInputValue('')
+      : setInputValue(textInput.current.value);
+
     if (textInput.current) {
       textInput.current.focus();
     }
--- a/packages/react/src/components/MultiSelect/FilterableMultiSelect.js
+++ b/packages/react/src/components/MultiSelect/FilterableMultiSelect.js
@@ -199,8 +199,8 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
     }
   }
 
-  function clearInputValue() {
-    textInput.current.value.length === 1
+  function clearInputValue(event) {
+    textInput.current.value.length === 1 || match(event, keys.Escape)
       ? setInputValue('')
       : setInputValue(textInput.current.value);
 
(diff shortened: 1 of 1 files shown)


packages/react/src/components/MultiSelect/FilterableMultiSelect.js
```diff
@@ -200,7 +200,10 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect(
   }
 
   function clearInputValue() {
-    setInputValue('');
+    textInput.current.value.length === 1
+      ? setInputValue('')
+      : setInputValue(textInput.current.value);
+
```
tw15egan, line 206: I think we need to check if the key pressed is <kbd>esc</kbd>. If so, it should clear the input (if the menu is closed). It does close the menu on the first press (expected), but I can't clear the input anymore.

guidari: Good catch! I'll fix it

guidari: @tw15egan let me know if this fix make sense! 
Now the user can press `Escape` and the input will cleared and the menu dismissed.

tw15egan (approved):
Working as expected 👍🏻

andreancardona (approved):

Issue #14244, opened 2023-07-19
Improve `yarn build` time with distributed caching


tay1orjones:
CI Build times are lengthy and even with sharding complex tasks like e2e testing, we're bottlenecked by `yarn build` taking roughly 10-15 minutes to complete. [Average time to merge is ~30 minutes](https://github.com/carbon-design-system/carbon/queue/main). The impact of this was lessened by turning on GitHub's new merge queue feature, but we could nearly eliminate this bottleneck altogether by having a distributed (shared) build cache.

It's very easy to opt in to the distributed cache provided by NxCloud through lerna:

* https://nx.app/
* https://nx.dev/core-features/share-your-cache
* https://lerna.js.org/docs/features/share-your-cache
* https://github.com/lerna/lerna/tree/main/packages/lerna/src/commands/add-caching#readme

This proposal is just for using the distributed cache - I'm not necessarily interested in their "Smart Distributed Task Execution" feature as we already have a decent setup and lots of GitHub Actions runners.

```[tasklist]
- [ ] https://github.com/carbon-design-system/carbon/issues/17986
- [ ] https://github.com/carbon-design-system/carbon/issues/18196
```

PR #14559, opened 2023-09-01
chore: yarn global cache

dbrugger:
Contributes to https://github.com/carbon-design-system/carbon/issues/14235

#### Changelog

**Changed**

- Use yarn global cache
- Use npm registry

**Removed**

- Unneeded repo local yarn cache

#### Testing / Reviewing

`yarn install`

Build time improvements:
- Checkout step in GitHub Action down from 26s to 5-7s.
- test:e2e down from 2m 25s to 2m 7s
- lint unchanged at 2m 25s
- avt down from 27m 34s to 21m 18s
- test up from 19m 27s to 20m 41s


.yarnrc.yml
```diff
@@ -1,5 +1,11 @@
+enableGlobalCache: true
```
tay1orjones, line 1: This defaults to `true`, so should already be on.

.yarnrc.yml
```diff
@@ -1,5 +1,11 @@
...
+
+httpTimeout: 300000
```
tay1orjones, line 3: What's the benefit of increasing the timeout from the default (60000) to 300000?

.yarnrc.yml
```diff
@@ -1,5 +1,11 @@
...
+
 nodeLinker: node-modules
 
+npmRegistryServer: 'https://registry.npmjs.org/'
```
tay1orjones, line 7: The default registry, `registry.yarnpkg.com` is just a CNAME of the npm registry. Is there any tangible benefit to not using the default?

https://yarnpkg.com/getting-started/qa#why-registryyarnpkgcom-does-facebook-track-us

.github/workflows/ci.yml
```diff
@@ -14,40 +14,28 @@ concurrency:
...
-  format:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: Use Node.js 18.x
-        uses: actions/setup-node@v3
-        with:
-          node-version: '18.x'
       - name: Install dependencies
-        run: yarn install --immutable --immutable-cache
```
tay1orjones, line 37: We want CI jobs to fail if a change to the cache would be made as a result from this install. This helps ensure integrity of dependencies between local dev environments and CI.

tay1orjones (changes requested):
Could you clarify how these changes result in a smaller repo size or faster CI times? I have a few additional comments below, let me know what you think.

.github/workflows/ci.yml
dbrugger: still does. yarn knows if it's running in a CI

.yarnrc.yml
dbrugger: would prefer the official hostname for the registry, one less variable for security.

.yarnrc.yml
dbrugger: not sure if it's needed, but saw that used in a couple of repos using yarn so might be some kind of best practice. fine to remove also.

.yarnrc.yml
dbrugger: default unless some other settings are configured. this removes the local (in repo) cache that creates the tar balls in Git. global means the cache is in the home dir which is what we need to clean that up.

tay1orjones:
Thanks again for taking some time to discuss this today! As we discussed I'll close this for now and we'll look at revisiting it after we land https://github.com/carbon-design-system/carbon/issues/14244.

PR #14560, opened 2023-09-01
ci: include dedupe, format in lint job

dbrugger:
dedupe, format and lint run as independent and parallel jobs. This seems inefficient.

#### Changelog

**Changed**

- Moved dedupe and format job steps into the lint one
- Add node_modules cache to lint job

#### Testing / Reviewing

The entire lint job including dedupe and format now takes 2m 54s, compared to 2m 57 for format alone before, saving the other 42s + 2m 25s completely. This saves 50% resources for this part of the CI, without compromising the feedback loop imo.

We could even consider adding e2e as well, as it's quick too. But then the naming of the new job would be more difficult :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,30 +14,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  dedupe:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: Use Node.js 18.x
-        uses: actions/setup-node@v3
-        with:
-          node-version: '18.x'
-      - name: Run yarn dedupe
-        run: yarn dedupe --check
-
-  format:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: Use Node.js 18.x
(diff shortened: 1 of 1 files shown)


tay1orjones (changes requested):
Reporting on these individually has a tangible benefit of not having to look at logs to determine why the status check failed. I know for certain that if a `dedupe` status check fails, I can immediately pull down the PR, run `yarn dedupe`, and push up the changes. Same thing for the `format` check. Adding an additional step to check the logs to determine which part failed is a small but not insignificant piece of the dx puzzle.

Also we're not resource constrained in our CI environment. We've got lots and lots of runners being on GHEC. With the difference here being negligible, I think it's beneficial to keep these as separate jobs. 

The real issue in regards to CI times is jobs relying on `yarn build`. It's the most lengthy part of the process. For this we intend to soon try out a shared/distributed build cache, #14244 

Just my 2c and I'm curious what others think

tay1orjones:
Thanks again for taking some time to discuss this today! As we discussed I'll close this for now and we'll look at revisiting it after we land https://github.com/carbon-design-system/carbon/issues/14244.

PR #14561, opened 2023-09-01
ci: consolidate and optimize jobs

dbrugger:
Extension of https://github.com/carbon-design-system/carbon/pull/14560 which also addresses build and e2e

#### Changelog

**Changed**

- consolidated dedupe, format, lint, build, e2e

**Removed**

- build from avt and vrt

#### Testing / Reviewing

lint-build-e2e: 3m 27s vs 42s + 2m 57s + 2m 25s + 2m 25s
overall build time: 31 --> 25m

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint:
+  lint-build-e2e:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


tay1orjones (changes requested):
This comment applies here as well https://github.com/carbon-design-system/carbon/pull/14560#pullrequestreview-1611787461

> Reporting on these individually has a tangible benefit of not having to look at logs to determine why the status check failed. I know for certain that if a `dedupe` status check fails, I can immediately pull down the PR, run `yarn dedupe`, and push up the changes. Same thing for the `format` check. Adding an additional step to check the logs to determine which part failed is a small but not insignificant piece of the dx puzzle.
> 
> Also we're not resource constrained in our CI environment. We've got lots and lots of runners being on GHEC. With the difference here being negligible, I think it's beneficial to keep these as separate jobs. 
> 
> The real issue in regards to CI times is jobs relying on `yarn build`. It's the most lengthy part of the process. For this we intend to soon try out a shared/distributed build cache, #14244 
>
> Just my 2c and I'm curious what others think

Rather than combining jobs I'd like to explore a shared/distributed build cache, #14244

dbrugger:
they still show as individual steps. no need to dig into logs to see which step failed:
image

Maybe no resource constraints are enforced, but am sure these improvements combined save ~30-50% compute. Carbon reducing Carbon emission :)

dbrugger:
note the build itself can be improved as neither avt nor vrt really needs the full build. that is also part of one of these PRs, so that only what is needed is built instead of all packages.

PR #14565, opened 2023-09-01
fix(menuButton): fix RTL issues with MenuButton

tw15egan:
Fixes some positioning logic inside `Menu` and `MenuButton`. 

#### Changelog

**Changed**

- Adjusted positioning logic based on `RTL` 

#### Testing / Reviewing

Check `Menu` and `MenuButton` stories in both LTR and RTL modes. Should render as expected.

One note: I'm seeing an issue where the state seems to be stale if you go from `MenuButton` to `Menu` with RTL enabled. It appears on the wrong side. However, once you refresh it is fixed. Any ideas? Can probably be fixed later

### DIFF
--- a/packages/react/src/components/Menu/MenuItem.js
+++ b/packages/react/src/components/Menu/MenuItem.js
@@ -9,7 +9,7 @@ import cx from 'classnames';
 import PropTypes from 'prop-types';
 import React, { useContext, useEffect, useRef, useState } from 'react';
 
-import { CaretRight, Checkmark } from '@carbon/react/icons';
+import { CaretRight, CaretLeft, Checkmark } from '@carbon/react/icons';
 import { keys, match } from '../../internal/keyboard';
 import { useControllableState } from '../../internal/useControllableState';
 import { useMergedRefs } from '../../internal/useMergedRefs';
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Menu/MenuItem.js
```diff
@@ -126,6 +134,14 @@ const MenuItem = React.forwardRef(function MenuItem(
...
   }, []);
 
+  useEffect(() => {
+    if (document?.dir === 'rtl') {
+      setRtl(true);
+    } else {
+      setRtl(false);
+    }
+  }, []);
+
```
tay1orjones: This is fine, but there is a possibility the `dir` won't be set on the document, but on a parent element via `LayoutDirection`. You could augment this to look at the `LayoutDirectionContext` in addition to the `document?.dir`

tay1orjones (approved):
Looks good, just one minor thought

packages/react/src/components/Menu/MenuItem.js
tw15egan: Ah right, was wondering how we could check for that! Updated to check the `LayoutDirection` as well as document direction, working as expected 👍🏻

tay1orjones (approved):

PR #14576, opened 2023-09-05
Improve typescript types for ComboBox component

### LINKED ISSUE #14572: [Feature Request]: Improve types for ComboBox component
### The problem

Current types for the `ComboBox` component are really incomplete. They are assuming many `any` and `unknown` types. There is no good type provided for the item used by this component.

### The solution

Improve types for this component (e.g. provide generic type for an item)

### Examples

Follow the solution already implemented in [`Dropdown.tsx`](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Dropdown/Dropdown.tsx)

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

misiekhardcore:
Closes #14572 

Adds missing typescript types and intoduces generic `ItemType`

#### Changelog

**New**

- Added generic `ItemType` type
- Added missing types for all internal parts of the component

**Changed**

- Improved props types
- Fixed formatting

#### Testing / Reviewing

- building
- viewing the ComboBox stories
- creating a component that uses the ComboBox and testing the typescript hints for updated props types


misiekhardcore:

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -123,13 +165,13 @@ export interface ComboBoxProps
...
    */
-  downshiftProps?: React.ComponentProps<typeof Downshift>;
+  downshiftProps?: ComponentProps<typeof Downshift<ItemType>>;
 
   /**
    * Provide helper text that is used alongside the control label for
    * additional help
    */
-  helperText?: string;
+  helperText?: ReactNode;
```
guidari, line 174: Shouldn't we change this to `string` as it was before?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -722,7 +802,7 @@ ComboBox.propTypes = {
    * Provide helper text that is used alongside the control label for
    * additional help
    */
-  helperText: PropTypes.string,
+  helperText: PropTypes.node,
```
guidari, line 794: Shouldn't we change this to `string` as it was before?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -150,26 +192,26 @@ export interface ComboBoxProps
   /**
    * Message which is displayed if the value is invalid.
    */
-  invalidText?: ReactNodeLike;
+  invalidText?: ReactNode;
```
guidari, line 195: Here we used to have `ReactNodeLike` but maybe we could change to `string`, since we are using only text.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -239,13 +288,13 @@ export interface ComboBoxProps
    * Provide text to be used in a `<label>` element that is tied to the
    * combobox via ARIA attributes.
    */
-  titleText?: ReactNodeLike;
+  titleText?: ReactNode;
```
guidari, line 281: Here we used to have `ReactNodeLike` but maybe we could change to `string`, since we are using only text.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -260,425 +309,456 @@ export interface ComboBoxProps
   /**
    * Provide the text that is displayed when the control is in warning state
    */
-  warnText?: ReactNodeLike;
+  warnText?: ReactNode;
```
guidari, line 302: Here we used to have `ReactNodeLike` but maybe we could change to `string`, since we are using only text.

guidari (commented):
Thank you for the contribution! 🚀
Just a few comments about the types.

packages/react/src/components/ComboBox/ComboBox.tsx
misiekhardcore: I was following the way the things are in `Dropdown` component and there the `helperText` is a `ReactNode`. I don't mind making it a string as it was but sometimes its usefull to have it as a node. WDYT?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -37,21 +58,36 @@ const {
...
+  }
+  if (
+    item !== null &&
+    typeof item === 'object' &&
+    'label' in item &&
+    typeof item['label'] === 'string'
+  ) {
+    return item['label'];
+  }
+  return '';
```
tay1orjones, line 76: It was previously never possible for this function to return an empty string with the nullish coalescing `&&`. It would return `item` if `item.label` was nullish.

```suggestion
  return item;
```

packages/react/src/components/ComboBox/ComboBox.tsx
tay1orjones: It's not just Dropdown, every other component with a `helperText` prop has it as a node. I think it's a good idea to move it to be a node.

packages/react/src/components/ComboBox/ComboBox.tsx
tay1orjones: It's a node down in the proptypes, so I think this is fine 👍

packages/react/src/components/ComboBox/ComboBox.tsx
tay1orjones: Same thing here, it's `node` down in the proptypes.

packages/react/src/components/ComboBox/ComboBox.tsx
tay1orjones: This is also a `node` down in the proptypes.

tay1orjones (changes requested):
Not your fault, but the diff is super hard to read 😅 

By and large I think this looks okay, but there's one spot I think we should change what we return in the defaultItemToString. Let me know what you think.

packages/react/src/components/ComboBox/ComboBox.tsx
tay1orjones: I think this is okay https://github.com/carbon-design-system/carbon/pull/14576#discussion_r1322003055

packages/react/src/components/ComboBox/ComboBox.tsx
misiekhardcore: but how did it assure that the returned value is a string? maybe we can return
 ```
`${item}`
```

misiekhardcore: but anyways, the code for this is identical as the one used in both `MultiSelect.tsx` and `Dropdown.tsx`

tay1orjones (approved):

tay1orjones:
@misiekhardcore looks good, just a merge conflict to address

misiekhardcore:
Sorry for a bit mess in rebasing but I forgot that I should rebase to upstream, but it should all be good now. thanks @tay1orjones for you review

tw15egan (approved):

PR #14580, opened 2023-09-05
feat(styles): convert styles to CSS logical properties, enhance RTL support

### LINKED ISSUE #13619: Use CSS Logical Properties and Values for Styling

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13619

Converts all applicable layout properties to [CSS Logical Properties ](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values). This will greatly enhance RTL support throughout the design system.

#### Changelog

**New**

- `stylelint-use-logical` added to `stylelint-config-carbon` 

**Changed**

- Replaced all physical layout properties with logical properties. 

#### Testing / Reviewing

Ensure there are no regressions in LTR (default) mode. Use the Layout switcher to test each component in RTL mode and ensure it renders as expected


tw15egan:
Seems like the only LTR change is due to the addition of padding on `TableHeaderContainer` on the right side, which is needed so that when RTL mode is enabled, the text is not flush against the container. Everything else seems untouched 👍🏻

alisonjoseph (approved):
This all looks good to me! 💪

guidari (approved):
Hey @tw15egan 

I went throught all the components to see if I could spot something that might be missing, not sure if all the points I made are valids, but maybe it's worth taking a look!
I'm happy to help on these if needed!

Btw, awesome change in the storybook with the RTL mode!! 🚀🔥

____________

**- Checkbox - Skeleton example**
When using the margin in this format: `margin: 1px 1px 1px 1px` the values don’t get translate to “RTL”. Maybe we won't be able to use this method to write `margin`, `padding`, etc...

Screenshot 2023-09-08 at 14 21 10
Screenshot 2023-09-08 at 14 21 42


**- CodeSnippet - Multiline, Singleline** 
Horizontal scrollbar getting a blur in the left and right side

**- Pagination**
Those arrow should change directions? I search a little bit and found these examples:  
https://github.com/mui/mui-x/issues/8665 
https://bootstrap.rtlcss.com/docs/4.0/components/pagination/#working-with-icons
From my understanding it shouldn’t, or am I wrong? Take PaginationNav as an example maybe. There the arrows kept the directions

**- ProgressBar**
The ProgressBar should switch the side where it starts or kept the same side (left)

**- Slider**
Seems like the Slider is getting the wrong value, maybe a logic would have to be written to switch the `max` and `min` values when in RTL mode (idk if we can track that in JS).

packages/react/.storybook/preview.js
```diff
@@ -306,20 +306,23 @@ const parameters = {
...
     React.useEffect(() => {
       document.documentElement.setAttribute('data-carbon-theme', theme);
     }, [theme]);
 
-    React.useEffect(() => {
+    React.useLayoutEffect(() => {
       document.documentElement.lang = locale;
       document.documentElement.dir = dir;
+      // Need to set random key to recalculate Popover coordinates
+      setRandomKey(Math.floor(Math.random() * 10));
```
tay1orjones, line 319: Oh wow great call, I hadn't thought about the need to re-render in some instances like this.

packages/react/src/components/Menu/Menu.js
```diff
@@ -82,8 +86,18 @@ const Menu = React.forwardRef(function Menu(
...
 
       const pos = calculatePosition();
-      menu.current.style.left = `${pos[0]}px`;
-      menu.current.style.top = `${pos[1]}px`;
+      if (
+        (document?.dir === 'rtl' || direction === 'rtl') &&
+        !rest?.id?.includes('MenuButton')
+      ) {
+        menu.current.style.insetInlineStart = `initial`;
+        menu.current.style.insetInlineEnd = `${pos[0]}px`;
```
tay1orjones, line 94: I'm so glad these work with CSSOM like this 🙌

tay1orjones (approved):
Overall looks great to me. I agree with Gui's comments above, though I think we could address those in separate PRs if you'd like. The diff on this one is pretty meaty as-is 😅

packages/react/.storybook/preview.js
tw15egan: Thanks to your suggestions as a way to rerender components I thought of trying it, and it worked perfectly!

packages/react/src/components/Menu/Menu.js
tw15egan: Yeah, I was super happy this worked, since this seemed to resolve _most_ of the RTL issues with `Menu`. I had only been looking at the stylesheets!

PR #14581, opened 2023-09-05
[AVT] - Added keyboard nav test `TextInput`

### LINKED ISSUE #14579: Add Keyboard Nav AVT test for `TextInput` P1

guidari:
Closes #14579

Added keyboard navigation to `TextInput` component.

#### Testing / Reviewing

- Ensure the test pass
- Ensure that all the necessary navigation is covered.


tw15egan (approved):

guidari:
> @guidari do you think we should add a test for toggling password visibility?

Hey @tw15egan I didn't write the test because we have a test in this path `packages/react/src/components/TextInput/__tests__/PasswordInput-test.js`

But maybe would be also good to write with the keyboard nav. I'll make sure to add that!

tw15egan:
@guidari oh nice, if the behavior is already covered then we're good. Just want to make sure we cover our bases

tay1orjones (approved):

PR #14592, opened 2023-09-06
fix(DataTable): add tabIndex if contents are horizontally scrollable

### LINKED ISSUE #14575: [a11y]: `<div class='bx--data-table-content'>` is scrollable but not tabbable
### Package

carbon-components-react

### Browser

Chrome, Edge

### Operating System

Windows

### Package version

7.59.13

### React version

16.14.0

https://www.ibm.com/able/toolkit/tools/

### Assistive technology

_No response_

### Description

The scrollable element `<div class='bx--data-table-content'>` is scrollable but not tabbable. This is part of `<DataTable>`:

https://github.com/carbon-design-system/carbon/blob/0d49a9c3120c0185184c0148d450b6a3ec391ea9/packages/react/src/components/DataTable/Table.js#L38

The CSS which makes this element scrollable is here:

https://github.com/carbon-design-system/carbon/blob/v10/packages/components/src/components/data-table/_data-table-core.scss#L27

### WCAG 2.1 Violation

WCAG technique G202: Ensuring keyboard control for all functionality

### Reproduction/example

https://stackblitz.com/edit/github-gbmmdk?file=src%2FApp.jsx

### Steps to reproduce

Inspect the element `document.querySelector('.bx--data-table-content')`. You'll find that it is horizontally scrollable. However, it is not possible to tab to this element.

> This is important content that the designer decided to make scrollable and not always display **all* the content at the same time when the area is restricted in a relatively small space. When scrollbars appear, the developer needs to ensure the element will be scrollable via the arrow keys.

> Ensure scrollable elements that contain visible non-tabbable elements are not excluded from the tab-order by adding a tabindex attribute value greater than or equal to ”0” to the container element.

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14575

Checks if the table width is larger than the table container, and if so, adds a `tabIndex` so that the container can be focused. The user can then scroll the table horizontally with the arrow keys.

#### Changelog

**New**

- Similar to the wrap text table alignment logic, this denounces on window resize to check if the table header is larger than the table container. If so, toggle a `tabIndex` on the table container

#### Testing / Reviewing

Reduce any of the table stories to a size that causes the table to scroll horizontally. Ensure that you can tab to the table, and use the arrow keys to scroll. When the table is not scrollable, there should be no tab index / able to be focused.


alisonjoseph (approved):
LGTM

andreancardona (approved):
Tested and looks good! 💯

PR #14609, opened 2023-09-08
fix(Notification): label notification with subtitle if title not present

### LINKED ISSUE #14605: [a11y]: ActionableNotification aria-label should have more flexibility
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.37.0

### React version

17

IBM accessibility checker

### Assistive technology

_No response_

### Description

We had an a11y issue opened:
“Interactive component with ARIA role 'alertdialog' does not have a programmatically associated name”

This boiled down to the fact that we have an ActionableNotification with no `title` (just a `subtitle`).  The ActionableNotification code ties the aria-labelledby to the `title`, so if you don't have a title, you end up with this issue. 
[code link](https://github.com/carbon-design-system/carbon/blob/8049fe372ad20b45d4a569a2955902f904b57680/packages/react/src/components/Notification/Notification.tsx#L978)

It would be nice to have a bit more flexibility, so you could have a notification without title and not have an accessibility violation. Basically, fall back to the subtitle if there's no title. We'd need to add an `id` to the subtitle div and point to that in this case. (there's other options too of course, but that seems logical. Could also have a `hideTitle` or a separate prop for another aria-label)

We went ahead and added a title to our ActionableNotification for now to resolve the issue, but we shouldn't have to do that. 

### WCAG 2.1 Violation

_No response_

### Reproduction/example

see description

### Steps to reproduce

see description

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14605

If `title` is not present, use `subtitle` instead

#### Changelog

**Changed**

- Autogenerate a `subtitleId` and use that if a `title` is not present to set the `aria-labelledby` on the `alertdialog`

#### Testing / Reviewing

Don't pass in a `title` and ensure no a11y violations are found

### DIFF
--- a/packages/react/src/components/Notification/Notification.tsx
+++ b/packages/react/src/components/Notification/Notification.tsx
@@ -938,6 +938,7 @@ export function ActionableNotification({
   const [isOpen, setIsOpen] = useState(true);
   const prefix = usePrefix();
   const id = useId('actionable-notification');
+  const subtitleId = useId('actionable-notification-subtitle');
   const containerClassName = cx(className, {
     [`${prefix}--actionable-notification`]: true,
     [`${prefix}--actionable-notification--toast`]: !inline,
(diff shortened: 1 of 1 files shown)


andreancardona (commented):
🎉

andreancardona (approved):

tay1orjones (approved):
Interesting issue. I wonder if we have any other areas in the system that rely on optional props for values in aria-labelledby.

PR #14626, opened 2023-09-11
fix(multiselect): calculate scroll width for title for truncation

### LINKED ISSUE #13511: [Bug]: MultiSelect - items are not using `title` attribute
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

1.26.0

### React version

18.2.0

### Description

`title` attribute is not being added to the items of `MultiSelect` component.
As this works correctly for dropdown which items has only one children - `div` with a text - MultiSelect uses more complex structure with checkboxes, pseudo-elements etc. This causes calculation in [ListBoxMenuItem:L14  ](https://github.com/carbon-design-system/carbon/blob/e3eff64f219b16993ef6594a4c4b72416003097a/packages/react/src/components/ListBox/ListBoxMenuItem.tsx#L14) to always return `false`. Because of this `title` attribute is never being added.

I suppose that fixing this is not urgent, but I couldn't find any workaround.

Example of `Dropdown` and `MultiSelect` with the same items:
multselect

### Reproduction/example

https://codesandbox.io/s/silly-snyder-r2p4c5?file=/index.js

### Steps to reproduce

Use link with example or:
1. Create a new `MultiSelect` component:
```
<MultiSelect
      ariaLabel="MultiSelect"
      id="carbon-dropdown-example"
      items={items}
      label="MultiSelect menu options"
      titleText="MultiSelect title"
    />
```
2. Define items, one of them should use label long enough to be truncated:
```
const items = [
  "Item 1",
  "Item 2",
  "Item with a very long label that should be truncated and `title` attribute should be added"
];
```

_____

With this setup defined open the MultiSelect and hover over item with truncated text. With `title` attribute there should be browser tooltip.
Currently - as attribute is not there - nothing happens which makes it impossible to read full item's label.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

alisonjoseph:
Closes #13511

#### Changelog

**Changed**

- calculates with width of the correct element to show the title when text is truncated

#### Testing / Reviewing

Check that title shows for truncated items across Dropdown, Multiselect and ComboBox, and that nothing is broken.


andreancardona (approved):
Tested and looks good to me!

tw15egan (approved):

PR #14642, opened 2023-09-12
feat(Grid): export GridSettings

### LINKED ISSUE #14583: [Feature Request]: Add subgrid prop to Grid
### The problem

Currently the grid components automatically add the subgrid classes to nested Grid components. There are some cases, when setting up grid template areas, where that may not be ideal. Looking for a way to disable the subgrid classnames on nested `<Grid>` components.

### The solution

Add an optional boolean prop to the Grid to render without subgrid classnames / styles.

```
 <Grid subgrid={false}>
```

### Examples

_No response_

### Application/PAL

W3DS

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

alisonjoseph:
Closes #14583

#### Testing / Reviewing

TODO:
- [ ] Remove test story


packages/react/src/components/Grid/CSSGrid.tsx
```diff
@@ -20,18 +20,20 @@ function CSSGrid<T extends React.ElementType>({
...
   let mode: SubgridMode = 'wide';
+
   if (narrow) {
     mode = 'narrow';
   } else if (condensed) {
     mode = 'condensed';
   }
 
-  if (subgrid) {
+  if (subgrid && subGrid) {
```
tay1orjones: `subGrid` with the camelCasing feels a bit odd. Could we rename the internal variable coming from `useGridSettings` so the public api prop could be all lowercase `subgrid`?

```
const { subgrid: _subgrid } = useGridSettings();

...


if (_subgrid && subgrid)
```

tay1orjones (commented):
Outside my one minor comment below, I'm curious, would it work if someone used the existing `GridSettings` to override the context? Modified example from the storybook test story to show what I mean:

```js
    <Grid>
      <Column lg={8}>
        <GridSettings subgrid={false}>
          <Grid>
            <Column lg={4} />
            <Column lg={4}>
              <Grid>
                <Column lg={2} />
                <Column lg={2} />
              </Grid>
            </Column>
          </Grid>
        </GridSettings>
      </Column>
      <Column lg={8}>
        <Grid>
          <Column lg={4} />
          <Column lg={4} />
        </Grid>
      </Column>
    </Grid>
```

If so, maybe this could eliminate the need for a prop and we could just document this (off the beaten path) usage in storybook? I'm not sure if GridSettings is exported right now.

packages/react/src/components/Grid/CSSGrid.tsx
alisonjoseph: Yea this felt a little weird as I was doing it, but we had other camel case props so I went with it, I can update though!

Will also take a look at `GridSettings` first and see if its even needed at all 👍

tay1orjones (approved):
No way, that's awesome that `GridSettings` worked out!

guidari (approved):

PR #14644, opened 2023-09-12
fix(DataTable): don't set unique name for each radio input

### LINKED ISSUE #14268: [a11y]: DataTable [with-radio-selection] - keyboard nav radio button bug
### Package

@carbon/react

### Browser

FF

### Package version

@carbon/react@1.32.0

Manual QA

### Assistive technology

JAWS

### Description
Radio buttons in the table

Unable to navigate through the Radio Buttons using Up/Down arrow keys

When the focus is on the currently selected Radio button, We are unable to navigate to the other radio buttons using up / down arrow keys.

### Recommended Fix
This issue might be happening due to the name attribute being different for each radio button. Use the same name attribute for all the radio buttons.

When the standard keyboard navigation method is not supported, it impacts the user’s ability to interact with page controls. This affects screen reader users. 

### WCAG 2.1 Violation
2.1.1 Keyboard navigation

### Reproduction/example
Record `Number 74` in the below table:

https://react.carbondesignsystem.com/?path=/story/components-datatable-selection--with-radio-selection

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14268

Fixes an issue that was preventing a user from navigating a Radio selection `DataTable` with the arrow keys. Since each radio button had a `name` appended with the row id, it assumed they were all in their own groups. Now, they all get the same name for each table. If a user has multiple radio tables, they would just need to pass in their own `name` prop to `TableSelectRow` after the `getSelectionProps`

#### Changelog

**Changed**

- Removed unique id per row that was being appended to the `name`

#### Testing / Reviewing

Go to `DataTable --> Selection --> With Radio` and ensure you can navigate the `DataTable` with the arrow keys

### DIFF
--- a/packages/react/src/components/DataTable/DataTable.tsx
+++ b/packages/react/src/components/DataTable/DataTable.tsx
@@ -632,7 +632,7 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
           onClick,
         ]),
         id: `${this.getTablePrefix()}__select-row-${row.id}`,
-        name: `select-row-${row.id}`,
+        name: `select-row`,
         ariaLabel: t(translationKey),
         disabled: row.disabled,
         radio: this.props.radio || null,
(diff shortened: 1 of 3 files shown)


tay1orjones (approved):
Arrow keys work great 👍

guidari (approved):
LGTM! 🚀

PR #14646, opened 2023-09-13
feat(button): allow tooltips on disabled buttons via `disabledFocusable`

### LINKED ISSUE #8509: Tooltip doesn't show when button is disabled
# When button is in disabled state, the tooltip doesn't show anymore

- [x] `carbon-components: 10.29.0`
- [x] `carbon-components-react: 7.29.0`

## Detailed description

Issue is visible in storybook: https://react.carbondesignsystem.com/?path=/story/components-button--playground

image

I would expect the tooltip to show even if button is disabled (this was working before). We often use that option to indicate why the button is disabled.

1. Create iconOnly button
2. set it to disabled state (`disable` props)
3. See that the tooltip with description doesn't show anymore

## Additional information

This is a regression, it was working before

tay1orjones:
Closes #8509
[Related internal slack thread](https://ibm-studios.slack.com/archives/C0M053VPT/p1694528780737269)

This adds a new `disabledFocusable` prop to `Button`. It gives disabled styling to the button while allowing it to remain focusable. 

This aims to solve a longstanding problem where users have no way to understand why disabled buttons (particularly `IconButton`s) are disabled.

I'm opening this as a draft for now to get feedback from the design team, IBMa and the community. I hope we can reach a consensus to merge this, but it's still possible this exploration could fail for a number of reasons.

#### Changelog

**New**

- add `disabledFocusable` to button 
- add associated styling via new classes
- add test coverage

#### Testing / Reviewing

- Review the design elements, ensure all button variants render appropriately when given `disabledFocusable`. 
  - Is the existing focus ring acceptable design?
- The onClick event should not fire when given `disabledFocusable`

#### Open questions

These are sourced from conversations in Slack, GitHub, and Carbon team discussions.

- Does the aria spec allow for this?
- Can `disabledFocusable` be used on a button that is a link (has an `href`)?
- If it can receive focus, do the disabled styles need to be updated to pass color contrast?
- Is it confusing to have in the design system some disabled items be focusable and some aren't?
- Would the recommendation be to make all disabled components use `aria-disabled` or something like `disabledFocusable`?
- Is there any precedent for this elsewhere?

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -288,6 +288,9 @@ Map {
       "disabled": Object {
         "type": "bool",
       },
+      "disabledFocusable": Object {
+        "type": "bool",
+      },
       "hasIconOnly": Object {
         "type": "bool",
       },
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -192,7 +192,7 @@ const Button = React.forwardRef(function Button<T extends React.ElementType>(
   const { 'aria-pressed': ariaPressed } = rest;
   let otherProps: Partial<ButtonBaseProps> = {
     disabled,
-    'aria-disabled': disabledFocusable,
+    'aria-disabled': disabledFocusable ? disabledFocusable : undefined,
     type,
     'aria-describedby': dangerButtonVariants.includes(kind)
       ? assistiveId
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -121,6 +121,7 @@ const Button = React.forwardRef(function Button<T extends React.ElementType>(
     className,
     dangerDescription = 'danger',
     disabled = false,
+    disabledFocusable = false,
     hasIconOnly = false,
     href,
     iconDescription,
@@ -147,7 +148,7 @@ const Button = React.forwardRef(function Button<T extends React.ElementType>(
 
   const handleClick = (evt: React.MouseEvent) => {
     // Prevent clicks on the tooltip from triggering the button click event
-    if (evt.target === tooltipRef.current) {
+    if (evt.target === tooltipRef.current || disabledFocusable) {
       evt.preventDefault();
       return;
     }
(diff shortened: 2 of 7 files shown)


tay1orjones:
Here's my thoughts on the open questions

### Does the aria spec allow for this?

[Yes](https://www.w3.org/TR/html-aria/#docconformance-attr) 

> Unless otherwise stated, authors MAY use `aria-*` attributes in place of their HTML equivalents on HTML elements where the `aria-*` semantics would be expected. For example, authors MAY specify `aria-disabled=true` on a `button` element, while also implementing the necessary scripting to functionally disable the `button`, rather than the use `disabled` attribute.


### Can `disabledFocusable` be used on a button that is a link (has an `href`)?

Nope, it's not allowed per the spec, and the implementation accounts for this.

### If it can receive focus, do the disabled styles need to be updated to pass color contrast?

I don't think so.

The disabled styling [definitely should be applied](https://w3c.github.io/aria/#aria-disabled): 

> In addition to setting the aria-disabled attribute, authors SHOULD change the appearance (grayed out, etc.) to indicate that the item has been disabled.

Beyond this, I think adding `aria-disabled` and disabling click events for the button would qualify the button as "inactive" despite being focusable. Having focus doesn't mean the control is operable, so I think there would be no requirement for that to meet color contrast. 

> User Interface Components that are not available for user interaction (e.g., a disabled control in HTML) are not required to meet contrast requirements. An inactive user interface component is visible but not currently operable. An example would be a submit button at the bottom of a form that is visible but cannot be activated until all the required fields in the form are completed.

### Is it confusing to have in the design system some disabled items be focusable and some aren't?

To me it feels less confusing than a disabled `IconButton` not having a way to convey why it's disabled.

### Would the recommendation be to make all disabled components use `aria-disabled` or something like `disabledFocusable`?

No, `disabled` still has it's place. `disabledFocusable` would only be cases where the button needs to remain in the focus order (for a tooltip to be displayed, for instance). 

There's some interesting points in [this article](https://css-tricks.com/making-disabled-buttons-more-inclusive/) though that suggest using `disabled` results in _less inclusive_ buttons overall.

### Is there any precedent for this elsewhere?

Yes, Fluent UI has options for both `disabled` and `disabledFocusable`: https://react.fluentui.dev/?path=/docs/components-button-button--default

I struggled for quite a while to find a good name for this prop (`visuallyDisabled`, `partiallyDisabled`, `ariaDisabled`, etc) and funnily enough landed on `disabledFocusable` before I looked around to see if other design systems had implemented something like this and saw FluentUI using the same name!

tay1orjones:
I'm looking into why the pagination previous buttons are no longer getting disabled styling...

image

tay1orjones:
After a lot of consideration and review, we're not going to land this for a variety of reasons.

I know this isn't the outcome many folks wanted, but our team is committed to evaluating the root issue and providing more robust guidance around disabled elements. This work will be aimed at extending the [disabled states pattern](https://carbondesignsystem.com/patterns/disabled-states/) already present on the website. This and more is outlined in the issue I opened today:

- https://github.com/carbon-design-system/carbon-website/issues/3758

PR #14647, opened 2023-09-13
chore(ci): change lint:styles to only target scss files

tw15egan:
Updates the `lint:styles` command to only target `scss` files. Should prevent hanging when running locally.


tay1orjones (approved):
Thank you! 🙏 

cc @dbrugger

guidari (approved):

PR #14654, opened 2023-09-14
feat(popover): use floating-ui for autoAlign

### LINKED ISSUE #14493: Refactor Popover to use FloatingUI

tay1orjones:
Closes #14493 

This PR enhances Popover's experimental `autoAlign` functionality to automatically position the floating element to be within view, even on scroll. 

This uses [`floating-ui`](https://floating-ui.com/) and relies solely on the `fixed` strategy, using `position: fixed` on the floating element. In the majority of cases, Popover floating elements will no longer be clipped by parents with `overflow: hidden`.

This implementation requires two important changes I'd like to highlight:
1. The caret must be a child of the popover content instead of a sibling, so there is a dom change here.
2. The `.#{$prefix}--popover-container` must not have `position: relative`

#### Changelog

**Changed**

- Add new alignment options to match the floating-ui api with backwards compatibility. These also more properly reflect the "logical property" naming convention of start/end instead of left/right, for instance.
- Use floating-ui for positioning the caret and floating element, apply styles via CSSOM to comply with CSP.
- Feed floating-ui the values of popover's custom properties (`offset`, caret size, etc) when possible.
- Update Popover styling so no positioning styling is used when `autoAlign={true}`
- Update ToggletipButton to use forwardRef
- Update Slug styling to be compatible with `autoAlign`

**Removed**

- There was an unused ref on Tooltip, I removed it.

#### Testing / Reviewing

- View the popover stories, there are a few autoalign experiment stories in there. Ensure it works as intended and there's no regressions.
- View the Tooltip autoalign story and ensure the autoAlign functionality works with no regressions
- View the ToggleTip autoalign story and ensure the autoAlign functionality works with no regressions
- View the slug stories and ensure ensure the autoAlign functionality works with no regressions


tay1orjones:
After approval but before merging I'll do a pass to remove some of the demo stories

alisonjoseph (dismissed):
Amazing work 🔥
Everything with the functionality of the components seem to be working as expected. 

I noticed under component API on the overview page Popover is displaying as undefined.
Screenshot 2024-02-27 at 12 51 25 PM

Also, did we want to add any docs around experimental auto align to this page?

tw15egan:
This is looking awesome! I noticed a tiny issue with the `Slug` callout; seems to only be an issue with the side and bottom alignments. Seems to be off by `1px`

tay1orjones:
Sorry for the delay here, I ended up having to update the `align` prop definition across the whole system for all components using Popover. Also fixed a bunch of TypeScript issues.

@tw15egan @alisonjoseph both of those issues you found should be fixed

@guidari could you double check the changes to ToggletipButton to ensure I didn't break anything you added in the [interactive tag PR](https://github.com/carbon-design-system/carbon/pull/15808/files#diff-aff3fc17c1be744759ed294411fed5beccce2582e3eb2c39565e7bdddfc3a30aR256-R281)?

guidari:
@tay1orjones All good with the changes in the `ToggletipButton`. There is no a11y errors 🚀

tw15egan (approved):
This looks awesome! I noticed a small issue with the `Slug` storybook alignments using the old values, but I will create a separate issue for that and tackle that with the upcoming `Slug` changes.

alisonjoseph (approved):
🔥🚀 Awesome, LGTM!! (other than a couple merge conflicts)

laurenmrice (dismissed):
Looks Awesome @tay1orjones ! :star2: Just two comments:

**Tab tip experimental auto align story**
The example of the popover with no caret should be sitting directly underneath the icon box with no space between them.

Example:
Frame 1

**Playground story**
Align prop: When the caret is set to to true, there is no caret present in the popover for `right-top` alignment. All other alignments with caret and without caret look great!

tay1orjones:
@laurenmrice Thank you for catching both of those! They should be fixed now.

tay1orjones:
@laurenmrice Thanks! I just pushed up a commit that fixes that issue.

I also removed the autoalign story from VRT for now because
1. It's experimental
2. It was larger than percy would snapshot, so it was just an empty snapshot file

I added the popover playground story to vrt instead.

PR #14657, opened 2023-09-14
fix(UIShell): Export prop types for components

ibacher:
This is a quick bug-fix that just exports some types for the UIShell sub-components that have been converted to Typescript. Currently, the types end up declared as the argument for the component, but not actually visible to consuming Typescript code.

#### Changelog

**Changed**

- Types for UIShell sub-component props are now exported

#### Testing / Reviewing

In a project using `@carbon/react` and Typescript you should be able to use `SideNavProps` as a type in a consuming component without a TS 4203 error.


ibacher:

tay1orjones (approved):
Thanks for this!

denniskigen:
Good to go, @tay1orjones?

tw15egan (approved):
LGTM 👍 ✅

Issue #6775, opened 2020-09-03
File uploader: Enhanced UX for single file upload


garythornton:
```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/14502
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3705
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/720
```

### Summary

_Please describe your request in one or two sentences._

The current file uploader UX appears to be optimized for uploading multiple files - we would like to improve the UX for uploading a **single file**.

Existing experience:
<img width="347" alt="image" src="https://user-images.githubusercontent.com/39956961/92133556-6deed200-ee00-11ea-94b5-12c43e8d5aa2.png">

...having the provided-file displayed as a separate element below the drop target is sub-optimal when the drop target is embedded in form with other elements (such as a URL field for importing from a remote system, name and description fields). 

A current workaround is to replace the drop target with the provided file element, however if the drop target changes into a different component when the file is added, it might feel like the file has been 'processed' in some way, when actually nothing happens to the file until the user clicks the primary button (e.g 'Import'). 

_Clarify if you are asking for design, development, or both design and
development._

Needs visual design to update the following to Carbon X, and development. 

### Justification

_Provide the business reasons for this request._

We would like to improve the UX for uploading a single file, and avoid unnecessary work-arounds.

### Desired UX and success metrics

_Describe the full user experience for this feature. Also define the metrics by
which we can measure success/failure for the user._

We think it would be good to consider an option where a single file appears in the existing boundary of the drop target, like we have today in IBM App Connect:

<img width="393" alt="image" src="https://user-images.githubusercontent.com/39956961/92131827-557db800-edfe-11ea-9a06-936ba1118b46.png">

Advantages:
- the provided file appears where the user expects (where they dropped it / clicked to provide it).
- no additional screen real estate is required to represent the provided file.
- other surrounding elements are not displaced by the provided file element.

### "Must have" functionality

_Highlight any "must have" needs and functionality for the request._

_This should not be a full list of functionality; the Carbon team will work with
you to define functionality based on the desired UX._

### Specific timeline issues / requests

_Do you want this work within a specific time period? Is it related to an
upcoming release?_

_NB: The Carbon team will try to work with your timeline, but it's not
guaranteed. The earlier you make a request in advance of a desired delivery
date, the better!_

### Available extra resources

_What resources do you have to assist this effort?_

[...]

PR #14660, opened 2023-09-15
New single drag and drop single `FileUploader`

### LINKED ISSUE #14502: Single File Uploader: Implementation
See specs in parent issue https://github.com/carbon-design-system/carbon/issues/6775

* Duplicate the existing drag and drop story, rename it to be "drag and drop single"
* Add a prop to conditionally render a FileUploaderButton inside the FileUploaderItem (with a `ghost` kind) to enable the "replace" behavior, add necessary styling to position the button correctly inside FileUploaderItem
* Truncation and tooltip detection will only be evaluated on initial render via a useLayoutEffect

guidari:
Closes #14502

The new story should upload only one file at a time and the component flow should follow these [steps](https://github.com/carbon-design-system/carbon/issues/6775#issuecomment-718784963) on the linked image in the issue. A new addition is the inclusion of a "Replace" button, which serves as a clear indicator to the user that only a single image is accepted.

#### Changelog

**New**

- New story for Drag and Drop Single File Uploader

#### Testing / Reviewing

- Go to the `FileUploader` in the `Drag And Drop Upload Container Example Application` variant and test if the functionality is working as expected.


andreancardona (changes requested):
I don't know how much we need to account for this - but maybe @laurenmrice has design thoughts on how we can make this mobile friendly. It currently breaks at 350px

Screenshot 2023-09-18 at 23 19 10

guidari:
> I don't know how much we need to account for this - but maybe @laurenmrice has design thoughts on how we can make this mobile friendly. It currently breaks at 350px

Ohh good catch!! I totally forgot to test on mobile screens!
I'll see what I can do to look better.
Thanks

guidari:
@laurenmrice There are a few pixels within the image in the issue that are sized at 24px instead of the required 16px. As the `FileUploader` component already exists, I didn't change those specific pixel dimensions.

alisonjoseph (changes requested):
this is what I'm seeing in firefox, I think it has to do with it needing a width or max-width for `text-overflow: ellipsis;` to work correctly in that browser.

Screenshot 2023-09-19 at 12 38 44 PM

laurenmrice (changes requested):
@guidari Looking good Gui! ⭐️ Just noticed a couple of things:

-------

**“Replace” link:**
- This should be using the link component, not a ghost button.

**Spacing:**
- There should be 16px padding to the right of the close icon. I think this is what you were referring to that it is currently 24px. It really should be 16px for both variants of drag/drop file uploader (the multiple file uploader one and the single file uploader one). This will also affect the positioning of the loading and success states, which should have 16px padding to the right.

Frame 4

**Overflow text in file name:**
- When a file uploads with a long string a text as its File name, it currently looks how Alison showed in her image example. There should be 16px padding to the left of the `Replace`link, and then we can ellipsis the file name. Here is an example:

Frame 3

**File height size:**
- The `sm` prop currently does not change the height size of the file uploaded to 32px.

**Mobile:**
- I like your idea for Mobile sizing, maybe it will work with following the same rules at the Overflow text I mentioned above^.

guidari:
Thanks for the feedback guys! 

> **“Replace” link:**
> * This should be using the link component, not a ghost button.

@laurenmrice in this case we can't use the `Link`  component because we are using the `FileUploaderButton` component to take advantage of all the logic behind uploading a file. 

But we can replicate the style of a `Link` component in the button. 

I see two ways of doing that: 

**1 - We can create a `buttonKind` called "link" (or something else)**
**2 - I can override the css without creating a new `buttonKind` just for the `FileUploaderItem`**

Both ways we have a button that looks like a `Link` component.

I'm just worried about the number 1 because I'm not sure if a `buttonKind` called "link` would make sense. The prop `size` wouldn't affect this type of button. 

Of course, we could also work on top of the Ghost kind as a third option.

Wondering what you guys think @laurenmrice @andreancardona @alisonjoseph 

For now I feel like option 2 sounds good.

alisonjoseph:
@laurenmrice are we sure we want to use a Link component here? 
https://carbondesignsystem.com/components/link/usage/#when-not-to-use

> Use a button instead of a link for actions that will change data or manipulate how it is displayed, change a state, or trigger an action. Buttons should never be used for navigational actions.

Can we make ghost button work? Or do we need a new Button variant? Or should this just be a one-off set of styles for a button like @guidari mentioned?

laurenmrice:
Following up on the convo above, after speaking with Michael and showing this in the design crit, we are making the design a little simpler without a replace button. I provided the new specs in the [original issue](https://github.com/carbon-design-system/carbon/issues/6775#issuecomment-718784963) and went over them with Gui.

laurenmrice (changes requested):
The functionality of this looks great! There are just a couple small visual things.

----

- I believe this is a Firefox bug: the text in the uploaded file is a little too high in the container. The text should be center within the container for all sizes.  Example spec below:
Frame 1

-  The close icon looks like it’s a couple pixels too far to the right. From the bounding box of the icon (which is 16px x 16px) there then should be 16px padding on the right. However, when the icon is focused, the focus border should still stay at 24px x 24px (this is currently correct in the preview).  Example spec below:

Frame 2

guidari:
@laurenmrice 

> * I believe this is a Firefox bug: the text in the uploaded file is a little too high in the container. The text should be center within the container for all sizes.    Example spec below:

For some reason on FF was 21.5px the height for the filename. I just fixed and now it is the same as chorme (18px)

> * The close icon looks like it’s a couple pixels too far to the right. From the bounding box of the icon (which is 16px x 16px) there then should be 16px padding on the right. However, when the icon is focused, the focus border should still stay at 24px x 24px (this is currently correct in the preview).  Example spec below:
> 

Fixed! There was a `gap` that was causing that!

alisonjoseph (approved):
LGTM!

laurenmrice (changes requested):
Hey @guidari ! I just reviewed your PR again and noticed the close icon is not 16px away from the right edge of the uploaded file. I was just wondering if your PR preview updated or not?

---

- When I upload a file with a long overflowing text name, in all sizes (lg, md, sm) the file name is still a little high, maybe by 1px. This does not happen with a short/normal file name , that one looks good!  Also related to this, when I switch the size to medium or small, the close icon comes outside of the file to the right. This is happening in Firefox/Chrome/Safari.

Screenshot 2023-09-28 at 11 30 36 AM

----

- There is one issue with the error state in the large size where the file name and icons sit too low in the top field area. All the other sizes look good. This is happening in Firefox/Chrome/Safari.
Screenshot 2023-09-28 at 11 33 36 AM

guidari:
@laurenmrice Fixed! 
It's just weird the file name being 1px high 🤔 I added a `margin-top` to fix that.

laurenmrice (approved):
Looks great, Gui! ⭐️

andreancardona (approved):
Looks good! 🎉

Issue #14477, opened 2023-08-21
Improved Right-To-Left (RTL) Support


tay1orjones:
This is an umbrella issue for tracking various improvements for right-to-left support.

```[tasklist]
# Tasks
- [ ] https://github.com/carbon-design-system/carbon/issues/13620
- [ ] https://github.com/carbon-design-system/carbon/issues/13619
- [ ] https://github.com/carbon-design-system/carbon/issues/14627
- [ ] https://github.com/carbon-design-system/carbon/issues/14478
- [ ] https://github.com/carbon-design-system/carbon/issues/7264
- [ ] https://github.com/carbon-design-system/carbon/pull/14658
```

PR #14679, opened 2023-09-18
feat(React): update more components to use Text, improve RTL support

tw15egan:
Refs https://github.com/carbon-design-system/carbon/issues/14477
Refs https://github.com/carbon-design-system/carbon/issues/7264

Adds in more support for RTL mode by utilizing the `Text` component under the hood wherever we render visible text. Also sets the `TextDirection` at the root level of the story. 

#### Changelog

**New**

- Sets the `TextDirection` at the root level based on the `Text Direction` dropdown in storybook

**Changed**

- Changed visible text nodes to use the `Text` component

#### Testing / Reviewing

Ensure there are no regressions when viewing the components in RTL or LTR mode


packages/react/src/components/Text/TextDirection.tsx
```diff
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 import React, { ReactNode, useEffect, useMemo, useRef } from 'react';
 import { TextDirectionContext } from './TextDirectionContext';
 
-export type TextDir = 'ltr' | 'rtl' | 'auto';
+export type TextDir = 'ltr' | 'rtl' | 'auto' | string;
```
tw15egan, line 12: Needed to add `string` because some components that were changed to `Text` were spreading `...other`. This was causing TypeScript errors because the types for `dir` are `string | undefined` on `ComponentProps` (`ListItem` was throwing this)

packages/react/src/components/Text/Text.tsx
```diff
@@ -22,7 +22,7 @@ export type TextProps<T extends React.ElementType> = PolymorphicProps<
 function Text<T extends React.ElementType>({
   as,
   children,
-  dir,
+  dir = 'auto',
```
tay1orjones, line 25: Are we sure this is what this should be defaulted to? For some reason I thought we specifically didn't add a default, but I can't remember the reasoning.

tay1orjones (commented):
Code looks good, though I still need to take a glance through the storybook on all these. Just one question below about the default.

packages/react/src/components/Text/Text.tsx
tw15egan: This is reverting to what the default was, I (mistakenly) removed it [here](https://github.com/carbon-design-system/carbon/pull/14580/files#diff-ffaea91692b65e84832ac2f9545028d2f6944c8fe3a8120d1ccd6fed473a9b7fL25-L25)

tay1orjones (approved):

andreancardona (approved):
looks good 🔥

PR #14709, opened 2023-09-25
fix(expandable-search): set tabindex to 0

janhassel:
Currently, when using an `ExpandableSearch` its trigger button (while collapsed) has its `tabindex` set to `1`. This makes it really challenging to account for in componentized UI development. Instead, a `tabindex` of `0` would ensure it follows the logical order of the DOM.

#### Changelog

**Changed**

- changed `tabindex` for collapsed, expandable search from `1` to `0`
- updated test

#### Testing / Reviewing

- Manual keyboard testing


tay1orjones (approved):
Keyboard nav looks good 👍 thanks for the improvement

andreancardona (approved):
Tested and looks good on my end!

PR #14712, opened 2023-09-25
fix(modal): update hasScrollingContent styles

### LINKED ISSUE #13714: [a11y]: Modal: The 'fade' effect on dialogs with overflow body text creates some odd behaviours

> The Carbon guidance on [Overflow](https://carbondesignsystem.com/components/modal/usage/#overflow-content) is that "The content should visibly fade at the end of the modal body area to indicate there is additional content out of view."
> 
> However, scroll bars are already indicating additional content. 
> 
> There are negative results of this 'fade area':
> 1) Especially with forms, the fade affect doesn't really work, so the bottom part of the dialog is effectively a bunch of padding
> image 
> 2) The fading area has been implemented in a way that makes the browser think it is not within the modal, so if a user clicks inside the dialog in this space, the dialog is dismissed, as if the user had clicked outside it (which, as per issue 3, should only be happening on the passive -- which should never have enough text to need scrolling IMO) **Sev 2**
> Here is a shot of a dialog with scrolling added to the body, so you can see the area in the dialog affected by this
> Screenshot 2022-12-12 at 6 29 08 AM
>

alisonjoseph:
Closes #13714 

Update the styling for the scroll fade so that it appears inside the modal-content and there isn't an odd space below. This also fixes the issue where if you click in the space below the content it closes the modal. 

#### Changelog

**New**
- Add **WithScrollingContent** stories to `Modal` and `ComposedModal`

**Changed**

- swap margin for padding on `modal-content ` so the fade will appear inside the modal 
- update styles for `modal-scroll-content` to accommodate the margin/padding swap

#### Testing / Reviewing

Check styles for Modal and ComposedModal, focus should wrap entire content area above button.

### DIFF
--- a/e2e/components/ComposedModal/ComposedModal-test.e2e.js
+++ b/e2e/components/ComposedModal/ComposedModal-test.e2e.js
@@ -37,10 +37,19 @@ test.describe('ComposedModal', () => {
           theme,
         });
       });
+
       test('full width modal @vrt', async ({ page }) => {
         await snapshotStory(page, {
           component: 'ComposedModal',
           id: 'components-composedmodal--full-width',
+          theme,
+        });
+      });
+
+      test('with scrolling content @vrt', async ({ page }) => {
+        await snapshotStory(page, {
+          component: 'ComposedModal',
+          id: 'components-composedmodal--with-scrolling-content',
           theme,
         });
       });
--- a/e2e/components/Modal/Modal-test.e2e.js
+++ b/e2e/components/Modal/Modal-test.e2e.js
@@ -37,6 +37,14 @@ test.describe('Modal', () => {
           theme,
         });
       });
+
+      test('with scrolling content modal @vrt', async ({ page }) => {
+        await snapshotStory(page, {
+          component: 'Modal',
+          id: 'components-modal--with-scrolling-content',
+          theme,
+        });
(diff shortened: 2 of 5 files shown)


tay1orjones (approved):
Looks good! Any idea what's up with the chevron alignment in the select? It looks fine in the select story and I can't pinpoint what style has it so misaligned 🤔 

image

alisonjoseph:
@tay1orjones ohh its the inline margin-bottom style I had added to the select, its adding it to the element and not the wrapping div. Not sure we can fix that in the component without it being a breaking change. I updated the story to add a wrapper div. 

Also unrelated thing, noticed the chevron spacing appears to be off by 4px on dropdown and multiselect. I'll check withy design/open an issue.

Screenshot 2023-09-27 at 11 52 05 AM

andreancardona (changes requested):
One last thing - top modal chevron looks a little misaligned with the others

Screenshot 2023-09-27 at 16 32 45

alisonjoseph:
@andreancardona I already have a PR open to fix that (unrelatedhttps://github.com/carbon-design-system/carbon/pull/14725 to this PR, but I noticed too 😅 )

andreancardona (approved):

PR #14714, opened 2023-09-26
refactor(icons-react): add TypeScript types to icons-react package

### LINKED ISSUE #11317: [Bug]: @carbon/react/icons module does not provide IDE import auto-complete

lewandom:
Closes #11317

Adds TypeScript types generation to the `@carbon/icons-react` package and to `@carbon/react/icons` submodule.

#### Changelog

**New**

- TypeScript bundler in `@carbon/cli`
- Dynamic generation of TypeScript type definitions for icons exported by `@carbon/icons-react`

**Changed**

- `Icon` component migrated to TypeScript
- `react` and `reactNext` builders in `@carbon/icon-build-helpers` switched to compiling TypeScript code
- `icon-helpers` package converted to TypeScript

#### Testing / Reviewing

Icons rendering in Storybook (e.g. in Accordion or FileUploader components) should work without any visual and behavioral changes. In a TypeScript environment, you should be able to import and reference icons using `@carbon/react/icons` module, with props auto-completion working.


lewandom:
@jdharvey-ibm @metonym @joshblack you may be interested in this PR given your involvement in #11317 and #12034. Since it's a major change, I'd like to have as many eyes as possible looking at this.

tay1orjones (changes requested):
Thanks for this! I just took a cursory glance at the storybook and the elements preview. Storybook looks fine, but the icons aren't being rendered in the elements preview.

image

`[object Undefined]` should be a rendered preview of each icon

lewandom:
@tay1orjones thanks for spotting! Looks like I was mistaken that the `toString` function is unused. I restored both `toString` and `toSVG`, should help fix the problem.

lewandom:
With one more change, the icons now show up in the carbon-elements preview.

packages/cli/src/commands/bundle/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+
+    if (await fs.pathExists(packageJsonPath)) {
+      break;
+    }
+  }
+
+  return packageFolder;
+}
+
+module.exports = bundle;
```
jharvey10: For these bottom three functions, what are your thoughts about exporting them from `javascript.js` and importing them here instead of duplicating them in this file.

packages/icon-helpers/tsconfig.json
jharvey10, line 1: Any thoughts on if this would be a good time to create a top-level `tsconfig.base.json` and extend from that for here and in `@carbon/react`?

packages/react/src/components/FileUploader/Filename.tsx
```diff
@@ -11,11 +11,12 @@ import React from 'react';
...
 import { usePrefix } from '../../internal/usePrefix';
 import { ReactAttr } from '../../types/common';
-
 export type FilenameStatus = 'edit' | 'complete' | 'uploading';
 
+type SVGAttr = React.SVGAttributes<React.ReactSVGElement>;
+
 export interface FilenameProps
-  extends Omit<ReactAttr<HTMLElement>, 'tabIndex'> {
+  extends Omit<ReactAttr & SVGAttr, 'tabIndex' | 'type'> {
```
jharvey10, line 19: I'm not super familiar with this code, but I'm curious as to why this type change was needed as a part of the Icons type work.

packages/react/tasks/build.js
```diff
@@ -107,9 +114,10 @@ const babelConfig = {
     '@babel/plugin-transform-react-constant-elements',
   ],
   babelHelpers: 'bundled',
+  extensions: ['.ts', '.tsx', '.js', '.jsx'],
```
jharvey10, line 121: Also not super familiar with this code, but just want to check if we need to consider any other extensions such as `.mjs`, `.cjs`, etc.

packages/react/tasks/build.js
```diff
@@ -132,8 +140,17 @@ function getRollupConfig(input, outDir, useTS) {
...
+        noEmitOnError: true,
+        noForceEmit: true,
+        outputToFilesystem: false,
+        compilerOptions: {
+          emitDeclarationOnly: true,
+          rootDir,
+          outDir,
+        },
+      }),
+      babel(babelConfig),
```
jharvey10: Some of these typescript config values seems to exist across a few different files (perhaps in slightly different forms). Is it possible to cue off of tsconfig files for these instead? Or is this maybe a limitation of the rollup tooling that we can't easily do that type of thing?

packages/cli/src/commands/bundle/typescript.js
lewandom: @jdharvey-ibm sounds like a good idea, I can do this.

packages/icon-helpers/tsconfig.json
lewandom: @jdharvey-ibm I was basing on top of this comment for distinct tsconfig.json files: https://github.com/carbon-design-system/carbon/pull/12034#discussion_r964106011. But I like the idea of using a top-level base config. I can certainly introduce it.

packages/react/src/components/FileUploader/Filename.tsx
lewandom: @jdharvey-ibm it's because this switch: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/FileUploader/Filename.tsx#L59 makes the Filename component return either `button` or a `CheckmarkFilled` icon component. All icons inherit SVG attributes.

packages/react/tasks/build.js
lewandom: @jdharvey-ibm I think it's not needed, since this task is specific to react package. I've copied the value provided in this line from the original https://github.com/carbon-design-system/carbon/blob/main/packages/react/tasks/build.js#L173 location. Also, related: https://github.com/babel/babel/issues/11577.

packages/icon-helpers/tsconfig.json
jharvey10: Ah! Not gonna lie I forgot about that comment, but I do still agree with it 😂 . The major one was definitely `@carbon/react` so with that in a good place, this might be a good time to treat as much of that config as makes sense as the top-level stuff, and have thinner configs for both carbon-react and now icons-react.

It's totally up to you if you want to handle that in this PR or if you'd prefer we do that in a separate one.

packages/react/src/components/FileUploader/Filename.tsx
jharvey10: Ahh okay awesome. Thanks for clarifying!

packages/react/tasks/build.js
lewandom: @jdharvey-ibm yeah, this is problematic... We have (at least) 2 different kinds of TS configs:
1. The `tsconfig.json` files are mainly used in development environment, e.g.: by IDE, eslint, etc.
2. The inline rollup configs, which emit to build directories.

In this PR, I didn't quite think about the difference between group 1 and group 2 - I just based on whatever was there for each group. I guess unifying into one base config and then extending with options that absolutely must be different (e.g. rootDir/outDir) is the right way to go. I'll look at this together with addressing your other comment (https://github.com/carbon-design-system/carbon/pull/14714#discussion_r1341477056).

packages/icon-helpers/tsconfig.json
lewandom: I think it's a good idea to handle it here, after all I did introduce these duplicating lines and agree it doesn't look pretty. I'll do this in coming days.

lewandom:
@jdharvey-ibm I think I addressed your comments where you asked for extracting common TypeScript configuration. For better or worse - cause now there's a separate subproject that holds that config + some utils. Let me know if that's an acceptable solution.

jharvey10 (commented):
I think these changes look pretty awesome! I like the introduction of the `typescript-config-carbon` package and how it gets used by the other packages.

One question: Can we safely remove the typescript dependency from all of the other packages now? My thinking being that we could use this new `typescript-config-carbon` as the definitive source of a typescript version across the entire monorepo.

@tay1orjones Putting this on your radar to look over the new package, since you're way more familiar with the impacts of adding this on the repo as a whole.

lewandom:
> One question: Can we safely remove the typescript dependency from all of the other packages now? My thinking being that we could use this new `typescript-config-carbon` as the definitive source of a typescript version across the entire monorepo.

@jdharvey-ibm I think we can, given I've forgotten to add it into `icons-react` and the build still works ;) I'll make a PR to remove explicit dependency on `typescript` in `@carbon/react`.

metonym (commented):
Awesome work!

Does the [`package.json`](https://github.com/carbon-design-system/carbon/blob/main/packages/icons-react/package.json) for `@carbon/icons-react` need to be updated to include the entry point for types?

I.e.,

```diff
"main": "lib/index.js",
"module": "es/index.js",
+ "types": "lib/index.d.ts",
```

lewandom:
@metonym thanks for reviewing!

> Does the [`package.json`](https://github.com/carbon-design-system/carbon/blob/main/packages/icons-react/package.json) for `@carbon/icons-react` need to be updated to include the entry point for types?

That's a great question, that I don't know the answer for... So far everything works without it:
- `@carbon/react` doesn't have `types` defined, yet it works fine when imported in TypeScript projects,
- `@carbon/icons-react` doesn't have `types` defined in this PR, yet importing it in `@carbon/react` works (e.g. in Accordion component) and recursive use via `@carbon/react/icons` works too in my dev env.

Also, the `types` field is very poorly documented - I only found one place: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html, and there's no information in there what actually uses the `types` field, how this mechanism works overall, and whether there are alternatives (cause things apparently work without it).

I'm willing to have additional discussions on this, if you have more information. For now, I'd rather keep this as is, since it works, and `@carbon/react` doesn't define it either.

metonym (approved):

tay1orjones (approved):
Hey thanks for this! I think this looks fine overall. Extracting the typescript config into it's own package is a great enhancement 👍

tw15egan (approved):
LGTM, small merge conflict to resolve then we should be good to go. Thanks for making this enhancement!

metonym:
These changes are published in [@carbon/icons-react/v/11.29.0-rc.0](https://www.npmjs.com/package/@carbon/icons-react/v/11.29.0-rc.0).

I'm testing the types, and they work great.

```tsx
✅ import { WatsonHealth3DCursor } from "@carbon/icons-react";
✅ import { Accessibility } from "@carbon/icons-react/es";
```

However, I'm getting a type error when attempting to import using the full path.

For example:

```tsx
❌  import Icon from "@carbon/icons-react/es/4K";
```

The type def looks like this:

```tsx
// node_modules/@carbon/icons-react/es/4K.d.ts

import type { CarbonIconType } from './CarbonIcon';
export const _4K: CarbonIconType;

```

While the generated JS code (abridged) has a default export:

```js
// node_modules/@carbon/icons-react/es/4K.js
// ...
export { _4K as default };
```

Can someone else confirm that they are able to repro? I hesitate to open a bug unless confirmed.

lewandom:
@metonym thanks for finding this, I'll try to reproduce and address.

metonym:
As of version 11.29.0, `@types/carbon__icons-react` has [been removed](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67176) from DefinitelyTyped.

PR #14723, opened 2023-09-27
fix(UIShell): use hr instead of li with role

### LINKED ISSUE #14479: [a11y]: SideNavDivider -  role 'separator' is not valid for the element <li>
### Package

carbon-components-react

### Browser

Chrome

### Operating System

Windows

### Package version

v7.59.10

### React version

v16.13.1

### Assistive technology

_No response_

### Description

Using the <SideNavDivider/> within the shell sidebar results in a accessibility violation "The ARIA role 'separator' is not valid for the element \<li>"

### WCAG 2.1 Violation

4.1.2 Name, Role, Value

### Reproduction/example

https://stackblitz.com/edit/github-cqny7t?file=src%2FApp.jsx

### Steps to reproduce

Use a SideNavDivider within SideNavItems and run accessibility checker

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14479

Refactors UI Shell Side Nav Separator to use an `hr` instead of a `li` with `role="separator" to fix an a11y issue

#### Changelog

**Changed**

- changed `li` to `hr`, added styles to match current

**Removed**

- unnecessary role 

#### Testing / Reviewing

Run a11y checker on a UI Shell story with a side nav separator and ensure there are no violations

### DIFF
--- a/packages/react/src/components/UIShell/SideNavDivider.js
+++ b/packages/react/src/components/UIShell/SideNavDivider.js
@@ -13,7 +13,7 @@ import { usePrefix } from '../../internal/usePrefix';
 function SideNavDivider({ className }) {
   const prefix = usePrefix();
   const classNames = cx(`${prefix}--side-nav__divider`, className);
-  return <li role="separator" className={classNames} />;
+  return <hr className={classNames} />;
 }
 
 SideNavDivider.propTypes = {
--- a/packages/react/src/components/UIShell/SideNavDivider.js
+++ b/packages/react/src/components/UIShell/SideNavDivider.js
@@ -13,7 +13,11 @@ import { usePrefix } from '../../internal/usePrefix';
 function SideNavDivider({ className }) {
   const prefix = usePrefix();
   const classNames = cx(`${prefix}--side-nav__divider`, className);
-  return <hr className={classNames} />;
+  return (
+    <li className={classNames}>
+      <hr />
+    </li>
+  );
 }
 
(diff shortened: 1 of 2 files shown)


alisonjoseph:
I don't believe it's valid html to have a `hr` as a child of `ul` 
https://html.spec.whatwg.org/multipage/grouping-content.html#the-hr-element

other than that it looks fine and doesn't have any a11y checker errors 🤔

Likely the correct html for this would require two separate lists with a hr in between (likely not a simple update)

tw15egan:
@alisonjoseph I don't think this is rendered inside a list, it seems like we just export the "divider" component that is used to break space between menus:

https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/UIShell/UIShell.SideNav.stories.js#L355

alisonjoseph:
Screenshot 2023-09-27 at 11 13 17 AM

tw15egan:
@alisonjoseph can we just wrap it in an`li`? We have `buttons` as nested elements in `li` for the nested menus, so I would imagine that's fine then?

alisonjoseph:
@tw15egan oh yea, good idea. I don't see why not 👍

tw15egan:
@alisonjoseph updated 👍

alisonjoseph (approved):
LGTM, no a11y errors

andreancardona (approved):

PR #14726, opened 2023-09-27
feat: added $layer token in the background-color

### LINKED ISSUE #5410: [structured list] add background option
Have an option to add a background to structured list.

Artboard

guidari:
Closes #5410

Added the $layer token to the background-color, so the user can override the background using tokens.

#### Changelog

**New**

- New stories created to cover @vrt testing

#### Testing / Reviewing

Make sure the change is right. You can inspect the code to see the layer in the background


guidari:
Hey @tay1orjones

I put this to review, but I'm not sure if the storybook looks good enough. I mean, everything works fine, the thing is that I could re write the token. I was trying to rewrite the `packages/styles/scss/components/structured-list/_structured-list.scss` file in a similiar way like this example, but I had no lucky: 

Looks like we can't rewrite since the module was already loaded, so it can't be configured using "with".

Not sure if the helper function you mentioned was something like that.

https://github.com/carbon-design-system/carbon/blob/acead26eeb6f15ec6d39970e4af08aa23ee48cdc/examples/custom-theme/src/index.scss#L6-L14

andreancardona (approved):

tay1orjones (approved):
Sorry I missed this @guidari, based on what you're saying I think we're actually okay here. To rewrite this background, you'd need to globally redefine the `$layer` token when `@use`ing the `@carbon/react/scss/theme` the first time.

I think once this is merged we could create a stackblitz example that redefines `$layer`, it's sort-of difficult to override the tokens for testing within storybook

packages/styles/scss/components/structured-list/_structured-list.scss
```diff
@@ -72,6 +72,10 @@
     }
   }
 
+  .#{$prefix}--with-background-layer {
+    background-color: $layer;
+  }
```
tay1orjones: Is this necessary since it's already using layer in the `.#{$prefix}--structured-list` selector above in this file?

https://github.com/carbon-design-system/carbon/pull/14726/files#diff-4e80dc6063d7cc3b405f2301c8664f49deaf6b752ccbb5b16f5f4154eebe81f1R35

thyhmdo (approved):

packages/styles/scss/components/structured-list/_structured-list.scss
guidari: @tay1orjones 
It is not. But the thing is that I wasn't able to redefine the `$layer` in storybook to use as an example, that's why I did this.
If I remove this style and try to use the `.#{$prefix}--structured-list`it will not work in the storybook, since we are not redefine the `$layer`.
From what I remember I didn't find a way to redifine the `$layer`on storybook.

guidari: I'll remove the storybook that I added. Looking again it doesn't feel like it is worth to have a storybook only for that.

tay1orjones (commented):
Nice! This looks correct, but the changes to typescript-config-carbon and the `d.ts` files shouldn't be in here. If these are from a `yarn build`, I'm not sure why they're not ignored and were included here 🤔

PR #14720, opened 2023-09-26
refactor(React): initialize default props with serially stable values

tw15egan:
Closes #10777 
Closes https://github.com/carbon-design-system/carbon/issues/13424

Removes all instances of `defaultProps` throughout the React codebase and reactors them to serially stable default function parameters. 

#### Changelog

**New**

- Added a `internal/noopFn` export that is used anywhere we were setting a default value on a change handler i.e `onChange: () => {}` becomes `onChange = noopFn`

**Changed**

- Removed default props and instead set default function parameters for these default values.
- Updated some tests and stories to account for required props that need to be explicitly provided

**Removed**

- I removed some `isRequired` props that were not required in other places, or would throw unnecessary `console` warnings if the prop was not explicitly provided (even though we set a default). 

#### Testing / Reviewing
- For each component modified, go to their story and ensure no console warnings appear. 
- Ensure removed `isRequired` props are okay
- Double check `DataTable` and `Listbox` components, as these were some of the more heavily modified components

PR #14733, opened 2023-09-28
refactor(Modal): add TypeScript types

### LINKED ISSUE #13563: Add TypeScript types to Modal

lewandom:
Closes #13563

Adds TypeScript types to Modal component.

#### Changelog

**Changed**

- Added TypeScript types to SkeletonPlaceholder component

#### Testing / Reviewing

The Modal component imports for the storybook should be valid and rendering. In a TypeScript environment, you should be able to import and reference the Modal component and its typings.


packages/react/src/components/Modal/Modal.tsx
```diff
@@ -495,7 +670,7 @@ Modal.propTypes = {
    * Specify CSS selectors that match DOM elements working as floating menus.
    * Focusing on those elements won't trigger "focus-wrap" behavior
    */
-  selectorsFloatingMenus: PropTypes.arrayOf(PropTypes.string),
+  selectorsFloatingMenus: PropTypes.arrayOf(PropTypes.string.isRequired),
```
tw15egan, line 679: Can we keep this as not required? In https://github.com/carbon-design-system/carbon/pull/14720 we're removing all default props and this will throw console warnings (even if we provide a default function parameter)

lewandom: @tw15egan the `isRequired` only pertains to the `string` part, not to the array / prop itself. It basically disallows nullish values inside the array, but the array itself can be `undefined`. This is to match the TS type constraint: `selectorsFloatingMenus?: string[]`. Otherwise, we'd need to type it as `selectorsFloatingMenus?: (string | null | undefined)[]` and handle nullish values in the code (right now we don't). I think it's not worth it.

tw15egan: Oh right, thanks for clarifying! This works great then.

tw15egan (approved):

andreancardona (approved):
looks good thank you for the contribution 🎉

PR #14739, opened 2023-09-29
refactor(types): added types for SideNavSwitcher

### LINKED ISSUE #13608: Add TypeScript types to SideNavSwitcher

SunnyJohal:
Closes #13608

Added Types to the SideNavSwitcher component

#### Changelog

**New**

- Renamed file extension to .tsx
- Added types for the component props.

**Changed**

- Memoized onChange handler to prevent unnecessary re-renders

**Removed**

- n/a


packages/react/src/components/UIShell/SideNavSwitcher.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const className = cx(`${prefix}--side-nav__switcher`, customClassName);
+
+    const handleOnChange = useCallback<ChangeEventHandler<HTMLSelectElement>>(
+      (event) => {
+        if (onChange) {
+          onChange(event);
+        }
+      },
+      [onChange]
+    );
```
tay1orjones: Unless I'm mistaken, if the reference to the `onChange` passed into this component isn't stable, it's still going to re-render because `onChange` is in the dependency array and will be a new reference. The memoization needs to happen outside of the component.

The renders here are super cheap, too. I'm not sure if the useCallback is needed. What do you think?

SunnyJohal: You're right, it will still change the dep, good catch. I'll back it out now

SunnyJohal: reverted

andreancardona (approved):
looks good to me - thank you for the contribution!

tay1orjones (approved):
Thanks for making that change!

PR #14773, opened 2023-10-03
refactor: convert-header-file-to-typescript-and-add-type-checks

### LINKED ISSUE #13585: Add TypeScript types to Header

Cris-Chan:
Closes #13585 

converted header to .tsx and removed prop types replacing them with typeSCript interface. mimicked enforcement of AriaLabelPropType as well (not sure how we want to go about that but made a first pass)
added contributer 

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{Header.js}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Ran in story book to make sure no errors were thrown! Place in stroybook page to make sure no errors are thrown!


Cris-Chan:

packages/react/src/components/UIShell/Header.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const prefix = usePrefix();
+  const className = cx(`${prefix}--header`, customClassName);
+
+  return (
+    <header {...rest} className={className}>
+      {children}
+    </header>
+  );
+};
+
```
guidari: We should keep the propTypes for documentation reasons in storybook and for JavaScript users also.

```suggestion
Header.propTypes = {
  /**
   * Required props for the accessibility label of the header
   */
  ...AriaLabelPropType,

  /**
   * Optionally provide a custom class name that is applied to the underlying <header>
   */
  className: PropTypes.string,
};
```

guidari (changes requested):
Hey @Cris-Chan thanks for the contribution!! 🚀

Just one minor change.

packages/react/src/components/UIShell/Header.tsx
Cris-Chan: oops! think i read that at totally forgot!
committing the proptypes now!

Cris-Chan (commented):
prop types added back in, needed for storybook and javascript users

guidari (approved):
LGTM! 🚀

andreancardona (approved):
Looks good to me :)

andreancardona:
@Cris-Chan Hey Cris - it looks like we have some failing tests here :/ do you mind fixing those and we can re-review after. Thanks!

PR #14775, opened 2023-10-04
refactor: add typescript types to TreeNode and TreeView 

### LINKED ISSUE #13581: Add TypeScript types to TreeView, TreeNode

cesardlinx:
Closes #13581

Typescript annotations added to TreeNode and TreeView 

#### Changelog

**Changed**

- TreeNode and TreeView components now use Typescript annotations
- Renamed TreeNode and TreeView file extension to .tsx
- Renamed index file extension to .ts

**Removed**

- n/a

#### Testing / Reviewing

- I run the tests for TreeNode and TreeView and it passes, also I checked that the components are still working in Storybook

image


cesardlinx:

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -5,14 +5,168 @@
...
+interface TreeNodeProps {
+  /**
+   * Node ID
+   */
+  id: string | number;
+
+  /**
+   * The value of the active node in the tree
+   */
+  active: string | number;
```
tw15egan: Can we keep the props `isRequired` values the same as the PropType definitions? Seems like this is enforcing a lot of optional props as required.

cesardlinx: Absolutely @tw15egan , I've already made the necessary changes to align the `isRequired` values with the PropType definitions. This should no longer enforce optional props as required. Thanks for catching that!

andreancardona:
@cesardlinx Hey David, do you mind fixing the merge conflicts - we'll re-review and get it merged!

cesardlinx:
Hey @andreancardona sure thing! I'll fix it today 🙂

cesardlinx:
Hey @andreancardona I have fixed the merge conflict, please let me know if you need anything else

packages/react/src/components/TreeView/index.ts
```diff
@@ -8,7 +8,5 @@
 import TreeView from './TreeView';
 import TreeNode from './TreeNode';
 
-TreeView.TreeNode = TreeNode;
-
```
tay1orjones: I think we'll need to keep this in to avoid causing a breaking change for anyone using `<TreeView.TreeNode>`

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -104,6 +179,7 @@ export default function TreeView({
         keys.ArrowDown,
         keys.Home,
         keys.End,
+        // @ts-ignore
```
tay1orjones, line 182: What error is being suppressed here?

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -119,8 +195,9 @@ export default function TreeView({
     if (match(event, keys.ArrowDown)) {
       nextFocusNode = treeWalker.current.nextNode();
     }
+    // @ts-ignore
```
tay1orjones, line 198: What error is being suppressed here?

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -203,13 +373,19 @@ const TreeNode = React.forwardRef(
...
+        aria-current={treeNodeProps['aria-current'] || undefined}
+        aria-selected={treeNodeProps['aria-selected'] || undefined}
+        id={treeNodeProps.id as string}>
         <div className={`${prefix}--tree-node__label`} ref={currentNodeLabel}>
           {/* https://github.com/carbon-design-system/carbon/pull/6008#issuecomment-675738670 */}
           {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
           <span
             className={`${prefix}--tree-parent-node__toggle`}
-            disabled={disabled}
+            // disabled={disabled}
```
tay1orjones: I think this needs to remain

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -9060,89 +9057,6 @@ Map {
     "render": [Function],
   },
   "TreeView" => Object {
-    "TreeNode": Object {
-      "$$typeof": Symbol(react.forward_ref),
-      "propTypes": Object {
-        "active": Object {
```
tay1orjones: Once `TreeView.TreeNode` is added back in, regenerate snapshots with `yarn test -u` and this change should be reverted/go away.

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -5,15 +5,163 @@
...
+   */
+  'aria-selected'?: boolean | null;
+
+  /**
+   * Indicates that the element is perceivable but disabled
+   */
+  'aria-disabled'?: boolean;
+
+  /**
+   * Name of a role in the ARIA specificatio
```
tay1orjones:
```suggestion
   * Name of a role in the ARIA specification
```

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -151,20 +314,22 @@ const TreeNode = React.forwardRef(
...
-          return depth + 1 + depth * 0.5;
-        }
-        // parent node without icon
-        if (children) {
-          return depth + 1;
-        }
-        // leaf node with icon
-        if (Icon) {
-          return depth + 2 + depth * 0.5;
+        if (depth != null) {
```
tay1orjones, line 317: I don't think this is necessary since you updated the default function parameter to always set it to `0`?

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -203,13 +373,19 @@ const TreeNode = React.forwardRef(
...
     return (
       // eslint-disable-next-line jsx-a11y/role-supports-aria-props
-      <li {...treeNodeProps} aria-expanded={!!expanded} ref={ref}>
+      <li
+        {...treeNodeProps}
+        aria-expanded={!!expanded}
+        ref={ref}
+        aria-current={treeNodeProps['aria-current'] || undefined}
+        aria-selected={treeNodeProps['aria-selected'] || undefined}
+        id={treeNodeProps.id as string}>
```
tay1orjones, line 383: Why is this necessary instead of just leaving these within `{...treeNodeProps}`?

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -288,7 +464,7 @@ TreeNode.propTypes = {
    * Optional prop to allow each node to have an associated icon.
    * Can be a React component class
    */
-  renderIcon: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
```
tay1orjones: Narrowing proptypes is considered a breaking change. We'll need to leave this as is for now.

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -148,6 +225,7 @@ export default function TreeView({
           }
         }
       }
+      // @ts-ignore
```
tay1orjones, line 228: What error is being suppressed here?

tay1orjones (changes requested):
Thanks for diving in to this! I have just a few observations below.

packages/react/src/components/TreeView/TreeView.tsx
cesardlinx: This is the error I'm suppressing here.
image

packages/react/src/components/TreeView/TreeView.tsx
cesardlinx: The error here is the same as in the previous comment
image

packages/react/src/components/TreeView/TreeNode.tsx
cesardlinx: If I add it back I get this error
image

Basically what it is trying to say is that span element doesn't have a `disabled` prop. And it makes sense because only html elements such as button, input, select, textarea have the `disabled` attribute

packages/react/src/components/TreeView/TreeNode.tsx
cesardlinx: Here I'm performing the extra check because typescript detects it as possibly 'undefined'
image

packages/react/src/components/TreeView/TreeNode.tsx
cesardlinx: I'm glad you asked, this is necessary in order to make the types match in this section, otherwise we will get this error:
image

packages/react/src/components/TreeView/TreeNode.tsx
cesardlinx: The problem was that by leaving it as how it was I get this error:
image

It appears to be a conflict between TypeScript and PropTypes in this section. The closest issue I found resembling this problem is documented here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/28015

packages/react/src/components/TreeView/TreeView.tsx
cesardlinx: Here I'm suppressing the same error as before
image

cesardlinx:
Thank you @tay1orjones for the valuable feedback. I've addressed some of the issues you highlighted, and also I've provided responses to your observations.

packages/react/src/components/TreeView/TreeNode.tsx
tay1orjones: Oh my bad, yeah let's just remove it then

packages/react/src/components/TreeView/TreeNode.tsx
tay1orjones: Ah weird, yeah I see on second glance it's actually not the default function parameter that you updated, it's just the `depth: (depth ?? 0)` in the `cloneElement` call. This looks good to me then 👍

packages/react/src/components/TreeView/TreeView.tsx
wkeese: It's too bad that match.js hasn't been converted to Typescript yet, but I think you can remove the @ts-ignore by just fixing match.js:

```js
/**
 * @typedef Key
 * @property key {Array<string>|string}
 * @property which {number}
 * @property keyCode {number}
 *  @property code {string}        <-- new line here
 */
```

tay1orjones:
Hey there, sorry for the delay in response here. From what I can tell there's two pieces we need to resolve here

1. Remove the `disabled` from TreeNode instead of commenting it out
2. Find a workaround for `renderIcon` in TreeNode because we can't narrow the Proptypes without a breaking change

cesardlinx:
Hey @tay1orjones. Sorry, work has kept me busy lately and I almost forgot about this PR. I'll check it out during the weekend and I'll keep you posted.

cesardlinx:
Hey @tay1orjones, I've resolved the issues you pointed out. Let me know if there's anything else I can help with.

Gururajj77:
Hi @cesardlinx,

Apologies for the delay. Could you please resolve the merge conflicts so that we can proceed?

Thank you.

tay1orjones:
This was put together before the addition of the TreeView controllable api, which after resolving the merge conflicts there's many tests failing. We're going to close this one in favor of opening a new one based on these changes but with the controllable api taken into consideration. Sorry for the delay here

PR #14776, opened 2023-10-04
fix: remove flicker in AccordionItem (#9818)

### LINKED ISSUE #9818: [Bug]: Flicker in AccordionItem children
### Package

carbon-components-react

### Browser

Chrome

### Package version

7.45.0

### Description

We have a requirement to persist the Accordion state while navigating through the application, thus kept the expand/collapsed state in the redux store. After upgrading from v7.26.0 where it worked with no issues it started blinking when being expanded or collapsed (see screen recording attached):

https://user-images.githubusercontent.com/609459/136445429-4d74e446-136c-4ec0-bdda-70a70cbba5d0.mov

While checking on the AccordionItem sources I noticed that changing the condition from
```
  if (open !== prevIsOpen) {
    setAnimation(isOpen ? 'collapsing' : 'expanding');
    setIsOpen(open);
    setPrevIsOpen(open);
  }
```
to 
```
if (open !== isOpen) {...
```` 
resolves the issue.

### CodeSandbox example

https://codesandbox.io/s/recursing-feather-4mglx?file=/src/index.js

### Steps to reproduce

Just pass the open property and update it whenever an item is expanded/collapsed.

### Code of Conduct

### PR DESCRIPTION
**Pull request by cesardlinx**
Removes the flickering effect when using AccordionItem as a controlled component

Closes #9818

#### Changelog

**Changed**

- I added a storybook example of the accordion as a controlled component

**Removed**

- I removed an extra condition and a state that was the causing the problem

#### Testing / Reviewing

- I run yarn test to verify that the Accordion related tests are still running

image

- You can check the Accordion Controlled example in Storybook and verify that now the flickering is gone

Here you can see how it was working before:

Recording 2023-10-04 at 09 19 05

And this is how it is working now:

Recording 2023-10-04 at 09 22 40

### CODE REVIEW

**Comment by tw15egan** [comment]
Still seeing a flicker when the Accordion is being closed - MacOS running Chrome latest

**Comment by cesardlinx** [comment]
Nice catch! You're absolutely right, I have updated the previous commit, this is how the tests are running

image

And this is how it's working now:

Recording 2023-10-04 at 11 58 28

**Comment by tw15egan** [comment]
@cesardlinx unfortunately, this change removed the entire collapsing animation. I believe the animation itself may need to be reworked, and not rely on `display: none` since that seems to be causing the issue

**Comment by cesardlinx** [comment]
@tw15egan Yep! that's right. Let me elaborate more on why. Basically the main issue is that the collapsing animation is not actually "collapsing", instead it's "shrinking" as you can see here:

Recording 2023-10-04 at 14 19 02

So, the hacky trick that somebody used is to mixed the shrinking with the "display: none" in order to create the "collapsing" effect like this:

Recording 2023-10-04 at 14 20 29

And that's why you experience it like a small flickering when it collapses, because it's the combination of "shrinking" and then using "display: none". So what I did was to use only "display: none" in order to avoid the flickering.

**Comment by tw15egan** [comment]
@cesardlinx the preferred solution would be to retain the animation and remove the flicker. The animation itself may need to be reworked.

**Comment by cesardlinx** [comment]
@tw15egan I can rework the animation, but I don't think I can access it am I right?

**Comment by tw15egan** [comment]
Great, if you are interested in trying to fix the animation, it is located in `packages/styles/scss/components/accordion.scss`

**Comment by cesardlinx** [comment]
Nice! so cool, I thought that you were using an external package or something, that's great!! I'll can fix it then :D

**Comment by cesardlinx** [comment]
@tw15egan  Animation reworked! This is how it works now:

Untitled

I'm getting some snapshot errors though as you can see here:

image

Should I update the snapshots? Please let me know

**Comment by cesardlinx** [comment]
Hey @tw15egan any updates on this? were you able to verify how the animation is working now?

File: packages/react/src/components/Accordion/AccordionItem.tsx [hunk 1]
```diff
@@ -82,12 +81,6 @@ interface AccordionItemProps {
    * The accordion title.
    */
   title?: ReactNode;
-
-  /**
-   * The callback function to run on the `onAnimationEnd`
-   * event for the list item.
-   */
-  handleAnimationEnd?: AnimationEventHandler<HTMLLIElement>;
```
tay1orjones: Removing a prop would constitute a breaking change, which we can't do right now. Is there a path to reinstating this prop to be called when the animation ends?

packages/react/src/components/Accordion/AccordionItem.tsx
```diff
@@ -186,8 +161,10 @@ function AccordionItem({
           {title}
         </Text>
       </Toggle>
-      <div id={id} className={`${prefix}--accordion__content`}>
-        {children}
+      <div className={`${prefix}--accordion__wrapper`}>
+        <div id={id} className={`${prefix}--accordion__content`}>
+          {children}
+        </div>
```
tay1orjones: Is there a way to fix this without modifying the markup? This can break tests in downstream applications that rely on this component's markup to be stable between minor versions.

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -154,11 +153,18 @@ $content-padding: 0 0 0 $spacing-05 !default;
     text-align: start;
   }
 
+  .#{$prefix}--accordion__wrapper {
+    display: grid;
+    // Properties for when the accordion closes
+    padding: 0;
+    grid-template-rows: 0fr;
+    opacity: 0;
+    transition: all $duration-moderate-02 ease-in-out;
```
tay1orjones: What is the reason for changing this from the animation values it was previously? `$duration-fast-02` and `motion(standard, productive)` instead of `ease-in-out`?

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -154,11 +153,18 @@ $content-padding: 0 0 0 $spacing-05 !default;
     text-align: start;
   }
 
+  .#{$prefix}--accordion__wrapper {
+    display: grid;
```
tay1orjones: I'm not confident that this won't break existing things people place in this component that inherently rely on the previous `display: block`. We could try, but I think it increases the risk of needing to roll this back if we publish it and get feedback from teams that this has unintended consequences.

tay1orjones (changes requested):
I don't see the flicker anymore but have a few concerns below.

packages/react/src/components/Accordion/AccordionItem.tsx
cesardlinx: Got it! I have resolved this issue by re-adding the event.

packages/styles/scss/components/accordion/_accordion.scss
cesardlinx: I have changed it back to `$duration-fast-02` and `motion(standard, productive)` as suggested

packages/styles/scss/components/accordion/_accordion.scss
cesardlinx: I have resolved this issue by using another approach instead of CSS grid

packages/react/src/components/Accordion/AccordionItem.tsx
cesardlinx: @tay1orjones Unfortunately, there isn't a way to fix this without modifying the markup. Let me explain why. The accordion content currently has padding, and if we animate only the content, the accordion won't completely close smoothly. You can see this in the GIF 

Recording 2023-10-11 at 21 28 52

You can appreciate that there is a small gap, and it is due to the content padding. While it might be tempting to remove the padding when closing the accordion, it results in a less smooth and seamless animation. Here's an example of that approach in action 

Recording 2023-10-11 at 21 43 30

As you can observe, the animation looks bad because the content shifts due to the padding appearing and disappearing, resulting in a somewhat hacky solution. The most effective approach to address this issue is to introduce the wrapper div. This way, we can avoid dealing with the padding within the content.

cesardlinx:
Hey @tay1orjones. I've addressed all the issues you mentioned earlier. I would greatly appreciate your feedback when you have the time. Please take a look at my comments on changing the markup. I explain why we should add a wrapper around the accordion content.

packages/react/src/components/Accordion/AccordionItem.tsx
```diff
@@ -131,24 +129,33 @@ function AccordionItem({
...
   // panel
   function onClick(event) {
+    const toggleElement = event.target;
+
+    let content = toggleElement.nextElementSibling;
+
+    // when toggle element is not the button
+    if (toggleElement.nodeName !== 'BUTTON') {
+      content = toggleElement.parentElement.nextElementSibling;
+    }
```
tay1orjones: Gaining access to the content node this way feels brittle. Instead, could a ref be placed on the `div` containing the `children`, and then that `contentRef.current.style` be used in the `maxBlockSize` logic below?

packages/react/src/components/Accordion/AccordionItem.tsx
```diff
@@ -131,24 +129,33 @@ function AccordionItem({
...
+
+    if (content.style.maxBlockSize) {
```
tay1orjones: Would it make more sense to use `isOpen` here?

tay1orjones (commented):
Thanks for making all those changes! I'm sorry for such a delay in reviewing. This is looking good, I have just two small comments after re-reviewing.

tay1orjones:
Also snapshots need to be updated, just run `yarn test -u` from the project root and push up the changes. Thanks!

tay1orjones (approved):
Thanks for addressing that feedback. I don't see the flicker in the storybook example. Removing the test story after a second approval and before merging would be good 👍

cesardlinx:
Awesome! Thank you @tay1orjones for all the feedback, I will remove the test story after second approval 👍

andreancardona (approved):
looks good! @cesardlinx

cesardlinx:
Thank you @andreancardona I have removed the test story as @tay1orjones suggested 🙂

PR #14784, opened 2023-10-04
GitHub workflow actions

andreancardona:
Part of https://github.com/carbon-design-system/carbon/issues/14294

This is to test the github bot

Worked on this with @garrigoose & @tay1orjones  🎉

MERGING for testing purposes

### DIFF
--- a/.github/workflows/contribution-proposal-not-pursuing.yml
+++ b/.github/workflows/contribution-proposal-not-pursuing.yml
@@ -27,5 +27,7 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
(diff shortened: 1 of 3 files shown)


.github/workflows/contribution-proposal-not-pursuing.yml
```diff
@@ -27,5 +27,7 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: 'Thank you for submitting your proposal. Unfortunately, it doesn't align with our roadmap or philosophy at this time. We value your contribution and encourage you to keep engaging with the community. For more information about our triage workflow and philosophy, please visit [here](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/support.md#type-enhancement-).'
```
tay1orjones, line 30: Just want to point out for posterity that this link seems to be the issue. When a comment is posted through github-script that contains a link like this, the action fails due to an "unexpected token". Removing the link fixes the issue. It might be related to https://github.com/actions/github-script/issues/256

PR #14793, opened 2023-10-05
feat(menu): support menu modes

### LINKED ISSUE #14770: [Menu and Menu buttons] Decorative icons: Implementation
####  Add decorative icon option to Menu
- We need to have an option to display decorative icons in front of Menu label text.
  - Jan Hassel will help contribute the code implementation of the decorative icons Menu

menu with icons

## The solution

**The “Menu” component supports two functionalities**
(Let’s call them "Full" and "Basic" for the sake of this example.)
- Full: The menu **can** support nesting + selectable items, but **can not** support decorative icons.
- Basic: The menu **can** supports decorative icons, but **can not** support  nesting + selectable items.

**The “parent” sets the mode**
- Menu (standalone, like when used with useContextMenu): Full
- MenuButton: Basic
- ComboButton: Basic
- OverflowMenu: Full

## Examples

menu with icons

## Application/PAL

_No response_

## Business priority

Medium to low priority. More of a nice to have

## Code of Conduct

janhassel:
Closes #14770

Adds the menu modes "basic" and "full" described in #14770 to support icons in MenuButton and ComboButton while restricting the use of nested and selectable items.

Opening as draft PR as the error throwing should be added to the unit tests, but first I wanted to clarify if throwing errors in this case is aligned to the Carbon ecosystem. A potential issue I see with this is that the errors are defined in `MenuItem` and therefore only throw once the MenuButton or ComboButton is opened.

#### Changelog

**New**

- Added `props.mode` to `Menu` and its context 
- Throw errors when invalid Menu combinations are used

#### Testing / Reviewing

- Manual testing for now (add `MenuItemSelectable` or `MenuItemRadioGroup` as children to MenuButton and ComboButton)
- Automated testing tbd


tay1orjones:
Thanks for submitting this @janhassel! Sorry for the delay in reviewing.

The biggest concern I have with throwing an error is that if this were to happen in production, React would completely unmount everything. Devs should be using `ErrorBoundary` to prevent this, but it's not guaranteed. 

Instead of throwing an error, I think it might make more sense to pop a `console.warn` and scope that error to only when in a development environment. This would follow the same approach used for deprecations via the `deprecate()` helper. This relies on the internal `warning` helper that could be used directly instead of throwing the error. It has the built in invariant check for `__DEV__` so these would only pop in non-production environments:

https://github.com/carbon-design-system/carbon/blob/36685a0be5e62db11a60d938c76b1c49a402c99a/packages/react/src/internal/warning.js#L10-L27

tay1orjones (approved):
Thanks, this looks good to me!

andreancardona (approved):
apologies for the delay on this @janhassel I was OOO for a few days. Looks good 😎

janhassel:
@andreancardona It looks like the auto-merge was cancelled. Could you take a look?

PR #14801, opened 2023-10-05
feat(SkeletonIcon): Add types for `SkeletonIcon`

### LINKED ISSUE #13574: Add TypeScript types to SkeletonIcon

amanlajpal:
Closes: #13574

Added Typescript types to SkeletonIcon Component

Changelog

New

Added types for the component props

Changed

Renamed file extension to .tsx
Added non-null assertion on className as it was absent

Removed

PropTypes and its import statement

Testing / Reviewing

building
creating a component that uses the SkeletonIcon and testing which props are valid/invalid

image

### DIFF
--- a/packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
+++ b/packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
@@ -25,8 +25,8 @@ const SkeletonIcon: React.FC<SkeletonIconProps> = ({ className, ...other }) => {
   const prefix = usePrefix();
 
   const skeletonIconClasses = classNames({
+    className,
     [`${prefix}--icon--skeleton`]: true,
-    [className!]: className,
   });
 
   return <div className={skeletonIconClasses} {...other} />;
--- a/packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
+++ b/packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
@@ -5,32 +5,31 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 import { usePrefix } from '../../internal/usePrefix';
 
-const SkeletonIcon = ({ className, ...other }) => {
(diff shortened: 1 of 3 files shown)


amanlajpal:
Thank you @tw15egan 😊, with your support I have finally created a fresh PR ✅

packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
```diff
@@ -5,32 +5,31 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import PropTypes from 'prop-types';
```
tw15egan, line 8: We need to keep the `propType` definitions in so that the props are rendered in Storybook. They can go at the end of the file.

tw15egan (dismissed):

packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
amanlajpal: Got it! I have added `propType` definition.

guidari (approved):
LGTM! 🚀

packages/react/src/components/SkeletonIcon/SkeletonIcon.tsx
```diff
@@ -4,18 +4,29 @@
...
+  style?: React.CSSProperties;
+}
+
+const SkeletonIcon: React.FC<SkeletonIconProps> = ({ className, ...other }) => {
   const prefix = usePrefix();
 
   const skeletonIconClasses = classNames({
     [`${prefix}--icon--skeleton`]: true,
-    [className]: className,
+    [className!]: className,
```
tw15egan: We can instead change this to something like this:

```jsx
  const skeletonIconClasses = classNames(className, {
    [`${prefix}--icon--skeleton`]: true,
  });
```

amanlajpal: Okay, thank you for reviewing 🙂

tay1orjones (approved):

tw15egan (approved):

PR #14815, opened 2023-10-09
Update examples.js

sujitmahapatra:
### Short Description

This PR updates the build script for generating example files in the Carbon Design System repository.

#### Changelog

**New**

- None.

**Changed**

- Updated the build script to improve efficiency and readability.
- Added error handling for better error reporting during the build process.

**Removed**

- Removed unused dependencies and unnecessary code segments.

#### Testing / Reviewing

To verify the changes in this PR:

- Clone the repository and navigate to the project root directory.
- Run the updated build script with the following command:

### DIFF
--- a/e2e/components/ComboButton/ComboButton-test.avt.e2e.js
+++ b/e2e/components/ComboButton/ComboButton-test.avt.e2e.js
@@ -7,92 +7,62 @@
 
 'use strict';
 
-import { expect, test } from '@playwright/test';
+import { test, expect } from '@playwright/test';
 import { visitStory } from '../../test-utils/storybook';
 
 test.describe('ComboButton @avt', () => {
-  test('@avt-default-state ComboButton ', async ({ page }) => {
-    await visitStory(page, {
-      component: 'ComboButton',
-      id: 'components-combobutton--default',
-      globals: {
-        theme: 'white',
-      },
-    });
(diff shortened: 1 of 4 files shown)


tay1orjones (commented):
Hey there! Thanks for submitting these ideas through a pull request.

Generally when we think about tests we tend to [avoid hasty abstractions](https://kentcdodds.com/blog/aha-programming) and [prefer duplication over the wrong abstraction](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing). 

It's not to say that the ideas you're proposing are wrong (they aren't), but it's more that the established convention within the codebase is to not use mutable variables within test files (where possible), for instance.

A quote from the second article I linked above rings true for some pieces of this changeset:

> I feel like the utilities like `testComboButton` and `takeVisualSnapshot` can be nice, but the tests are simple enough that duplicating that code instead could simplify our test code a bit. It's just that the abstraction of the function doesn't really give us a whole lot of benefit for this small set of tests, and we incur the cost for maintainers to have to look around the file for where those functions are defined.

Similarly, I don't see a benefit to removing comments from the examples.js file. Condensing the `relativePath` conditional down to a single condition is less verbose, but also less explicit to someone reading and maintaining the file.

For these reasons we won't accept this pull request. Thank you so much for your time and contribution though! I'm happy to continue the conversation here if you had another goal in mind that I might have missed.

PR #14826, opened 2023-10-09
chore(test-utils): remove keyboard, tabbable

tay1orjones:
As I was researching how to fix the CI checks for https://github.com/carbon-design-system/carbon/pull/14814, I noticed that we don't use these keyboard test helpers anywhere in the codebase. The package is also marked private and isn't published to npm anymore, though it once was right after v10 was released.

Overall I think these are safe to remove and also enables us to remove `tabbable` from the dependency tree.

#### Changelog

**Removed**

- remove keyboard test utils
- remove related `tabbable` depdendency

#### Testing / Reviewing

- Double check that these functions aren't used anywhere and that checks pass


alisonjoseph (approved):
LGTM! Agree this should be safe to remove 👍

tw15egan (approved):

PR #14828, opened 2023-10-09
fix(zone): update styles to use the prefix

### LINKED ISSUE #14827: [Bug]: config prefix hardcoded to 'cds' in zone.scss for background and color properties
### Package

@carbon/styles

### Browser

_No response_

### Package version

11.33.0

### React version

_No response_

### Description

Lines 39 and 40 set the background and color properties scoped to given zone class. However, they are hardcoded with the `cds` prefix rather than dynamically using the prefix from config. This break when a user of your system sets a different prefix.

```
    39: background: var(--cds-background);
    40: color: var(--cds-text-primary);
```

suggested:
```
background: var(--#{config.$prefix}-background);
color: var(--#{config.$prefix}-text-primary);
```

### Reproduction/example

https://github.com/carbon-design-system/carbon/blob/release/v11.33.0/packages/styles/scss/_zone.scss

### Steps to reproduce

Set the prefix to something other than `cds`. Check zoned styles to see if background and color are broken

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #14827

#### Changelog

**Changed**

- update tokens to use the configurable prefix instead of a hardcoded `cds`
  - These style declarations were first introduced in https://github.com/carbon-design-system/carbon/pull/11895

#### Testing / Reviewing

- zone styles should now respect any configured prefix

### DIFF
--- a/packages/styles/scss/_zone.scss
+++ b/packages/styles/scss/_zone.scss
@@ -36,8 +36,8 @@ $-components: (
 
 @each $name, $theme in $zones {
   .#{config.$prefix}--#{'' + $name} {
-    background: var(--cds-background);
-    color: var(--cds-text-primary);
+    background-color: custom-property.get-var('background');
+    color: custom-property.get-var('text-primary');
 
     @each $key, $value in $theme {
       @if type-of($value) == color {


tw15egan (approved):
Thanks for fixing that 😄

guidari (approved):

PR #14832, opened 2023-10-09
Fixed name convention for avt testing

guidari:
Fixed name convention for avt testing

Convention doc:  https://github.com/carbon-design-system/carbon/blob/main/docs/testing.md#tags


tay1orjones (approved):
Thanks for doing a pass on this! I know the naming convention was up in the air for a while and we changed some things. Appreciate you doing this second pass to get them aligned. Thanks!

tay1orjones (approved):
Thanks for doing a pass on this! I know the naming convention was up in the air for a while and we changed some things. Appreciate you doing this second pass to get them aligned. Thanks!

alisonjoseph (approved):

PR #14860, opened 2023-10-10
Add LTS notification - v10 storybook

### LINKED ISSUE #14765: Add LTS notification prominently to v10 react storybook
This issue is for exploring how we can add a notification to the v10 react storybook highlighting that v7(v10) is going away.

- ~Could we add a banner across all of storybook? Would this require us making an addon, or could we do it a different way?~
- Add it to the welcome page
- Add it to the docs pages
- **Probably the easiest thing to do is make a new global decorator and inject the notification into the top of the canvas for every single story**

There's more details and a list of open questions on the [parent issue](https://github.com/carbon-design-system/carbon/issues/14736)

guidari:
Closes #14765 

I've tested each default story, and they all appear to be working well. However, when it comes to the UIShell, there's an issue with the sticky/fixed header. I'm considering the best course of action, which might involve excluding the UIShell from the LTS notification, to prevent any additional problems when working with that particular story.

#### Changelog

**New**

- Added LTS notification

#### Testing / Reviewing

- Check if the design looks good
- Check if the components looks good

**_OBS: The message it is just an example, we are still selecting the best one._**

### DIFF
--- a/packages/react/.storybook/Container.js
+++ b/packages/react/.storybook/Container.js
@@ -35,19 +35,21 @@ function Container({ story, id }) {
         <div
           style={{
             backgroundColor: 'black',
+            fontSize: '1rem',
             width: 'auto',
             padding: '1rem',
             margin: '-1rem -1rem 0px -1rem',
             display: 'flex',
             justifyContent: 'space-between',
             alignItems: 'center',
-            fontSize: '1rem',
+            g1p: '2rem',
           }}>
           <p style={{ color: 'white' }}>
             carbon-components-react@7.x is in maintenance mode, support is
             scheduled to end on September 30, 2024.
           </p>
           <a
+            style={{ textDecoration: 'none' }}
             href="https://github.com/carbon-design-system/carbon/blob/main/docs/release-schedule.md"
             target="_blank">
             Release schedule
--- a/packages/react/.storybook/Container.js
+++ b/packages/react/.storybook/Container.js
@@ -44,8 +44,8 @@ function Container({ story, id }) {
             fontSize: '1rem',
           }}>
           <p style={{ color: 'white' }}>
-            carbon-components-react {`v${PackageInfo.version}`} is in
-            maintenance mode, support is scheduled to end on September 30, 2024.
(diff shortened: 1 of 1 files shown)


packages/react/.storybook/Container.js
```diff
@@ -25,8 +25,26 @@ function Container({ story, id }) {
...
+      {!url && (
+        <div
+          style={{
+            backgroundColor: 'black',
+            width: 'auto',
+            padding: '1rem',
+            margin: '-1rem -1rem 0px -1rem',
+          }}>
+          <p style={{ color: 'white' }}>
+            Carbon v10 is currently in maintenance mode, support is scheduled to
```
tay1orjones: Would it make more sense in the storybook to be more explicit on what package and version this applies to? I think the meaning of "v10" could easily be lost on folks.
```suggestion
            carbon-components-react@7.x is in maintenance mode, support is scheduled to
```

tay1orjones (approved):
Overall I think it's fine if this doesn't show up on the uishell stories for now, this will hit the vast majority of users being on every other component.

packages/react/.storybook/Container.js
guidari: @tay1orjones Yeah that make sense! 
I also added a link to the release schedule, let me know if looks good.

Screenshot 2023-10-11 at 09 15 57

sstrubberg (approved):

alisonjoseph (dismissed):
Overall looks good, the link color is wrong for a black background, so may want to get a designer to take a look. Could be easiest to match what we have on the [website banner](https://github.com/carbon-design-system/carbon-website/pull/3769) style-wise?

guidari:
> Overall looks good, the link color is wrong for a black background, so may want to get a designer to take a look. Could be easiest to match what we have on the [website banner](https://github.com/carbon-design-system/carbon-website/pull/3769) style-wise?

Could be something like this: 

Screenshot 2023-10-13 at 10 13 14

tw15egan (approved):

PR #14877, opened 2023-10-11
Added trap focus and focus redirection

### LINKED ISSUE #14691: [a11y]: ActionableNotification - Screen Reader focus is not trapped within the alertdialog
[Original report](https://ibm-studios.slack.com/archives/C054WK4ES5S/p1694709381258889)

> Issue:
> 1. When the dialog is displayed initial focus is not set with in a dialog.
> 2. In TAB and Shift TAB navigation Screen Reader focus is not trapped with in a dialog
> 
> Recommended fix: 
> 1. When a dialog is displayed the Screen Reader focus needs to be set with in adialog and also the focus needs to be trapped with in a dialog while TAB and Shift TAB navigation and arrow key navigation.
>
> User Impact:
> 
> 1. When the screen reader focus is not restricted within the modal dialog, it also identifies parent page content that disorients screen reader users.
When the focus is not set on a logical point in the page after a user action, it creates a great deal of confusion. The whole process of navigating back to the relevant element can be tedious. note this defect mainly impacts keyboard-only users and those with cognitive disabilities. 
>
> Relates to WCAG 2.1 - 2.4.3 Focus Order 

It's also unclear if the `hasFocus` prop is working and pulling focus to the ActionableNotification on initial render. This was tested in JAWS FF and Voiceover safari iPhone but is present in all browsers.

guidari:
Closes #14691

Enhanced the ActionableNotification component by implementing trap focus functionality, alongside the addition of a tabIndex attribute to the parent div. This enhancement ensures that when a user receives a notification with the hasFocus prop set to true, the focus is correctly directed to the component.

#### Changelog

**New**

- Added the trap focus functionality;
- Added `tabIndex` in the parent div.

#### Testing / Reviewing

- Check if the focus is trapped in the notification
- Check if the focus is working correctly in the `TestFocus` variant 
____________

- [ ] Delete the `TestFocus` variant


packages/react/src/components/Notification/Notification.tsx
```diff
@@ -1003,20 +1035,30 @@ export function ActionableNotification({
           </div>
         </div>
       </div>
+      <div ref={innerModal} tabIndex={-1}>
```
tay1orjones: Does this need the tabindex?

packages/react/src/components/Notification/Notification.tsx
```diff
@@ -976,7 +998,17 @@ export function ActionableNotification({
       ref={ref}
       role={role}
       className={containerClassName}
-      aria-labelledby={title ? id : subtitleId}>
+      aria-labelledby={title ? id : subtitleId}
+      onBlur={handleBlur}
+      tabIndex={hasFocus ? 0 : -1}>
```
tay1orjones: This is causing an accessibility error

> The tabbable element's role 'alertdialog' is not a widget role

I don't think this is needed? I think when the component is rendered, focus should be placed on the action/button, not on the container.

tay1orjones (commented):
Just a couple thoughts here relating to the a11y failure in the tests

packages/react/src/components/Notification/Notification.tsx
guidari: It fixed something when I wrote it, but now if I remove works fine! haha

packages/react/src/components/Notification/Notification.tsx
guidari: Now it is focusing on the action button

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):
Looks good but we need to remove the test story. I'll do it real quick

PR #14892, opened 2023-10-12
feat: added support for `Alt+Down Arrow` and `Alt+Up Arrow` in ComboBox

### LINKED ISSUE #7998: [Feature request]: Support `Alt+Down Arrow` and `Alt+Up Arrow` in ComboBox
Follow-up from https://github.com/carbon-design-system/carbon/pull/7777#issuecomment-788985847:
- If the combobox has focus and the user types <kbd>alt+down arrow</kbd>, then the listbox should open without any items selected, i.e. same behavior as clicking on the combobox with the mouse. (Currently, <kbd>alt+down arrow</kbd> selects the first item, but that should only happen if the user just types unmodified <kbd>down arrow</kbd>).

- If the user types <kbd>alt+up arrow</kbd>, then the listbox should close and focus should return to the input, as described here: https://w3c.github.io/aria-practices/#combobox-keyboard-interaction

https://carbon-components-react.netlify.app/iframe.html?id=components-combobox--combobox

k-rajat19:
Closes #7998


packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -600,6 +600,19 @@ const ComboBox = forwardRef(
                   event.target.value.length
                 );
               }
+
+              if (event.altKey && event.key == 'ArrowDown') {
```
k-rajat19, line 597:
```
if(match(event, keys.Alt) && match(event,keys.ArrowDown))
{
 ////
}

```
this doesn't seem to be working ??

tay1orjones: There is no definition for `keys.Alt` in `packages/react/src/internal/keyboard/keys.js`. Could you add it and retry using `match()` again?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -600,6 +600,19 @@ const ComboBox = forwardRef(
...
+                event.preventDownshiftDefault = true;
+                if (!isOpen) {
+                  toggleMenu();
+                }
+              }
+              if (event.altKey && event.key == 'ArrowUp') {
+                event.preventDownshiftDefault = true;
```
tay1orjones, line 604: What does this prevent exactly?

tay1orjones (commented):
Thank you for submitting this change! Could you additionally add tests covering this functionality to prevent future regressions?

packages/react/src/components/ComboBox/ComboBox.tsx
k-rajat19: > ```
> if(match(event, keys.Alt) && match(event,keys.ArrowDown))
> {
>  ////
> }
> ```
> 
> this doesn't seem to be working ??

I had tried this after adding definition for Alt key
However, if I put any one condition it works as expected but both simultaneously are not working :(

packages/react/src/components/ComboBox/ComboBox.tsx
k-rajat19: It prevents the [default behavior](https://github.com/downshift-js/downshift/blob/master/README.md?plain=1#L952C1-L957C72) of `ArrowUp` button, usually default behaviour of `ArrowUp` is not shown when both button are pressed carefully but  some time I have detected this behavior (bottom-most index is highlighted)

packages/react/src/components/ComboBox/ComboBox-test.js
```diff
@@ -234,5 +236,35 @@ describe('ComboBox', () => {
       expect(firstListBox()).toBeEmptyDOMElement();
       expect(secondListBox()).toBeEmptyDOMElement();
     });
+    it('should open menu without moving focus on pressing Alt+ DownArrow', async () => {
```
k-rajat19, line 239: @tay1orjones I have added some test cases but they are not working correctly, glad if you could help

k-rajat19: @tay1orjones ping !

k-rajat19:
@tay1orjones can you review these changes?

k-rajat19:
@tay1orjones I have added tests, this PR is ready for review

andreancardona (approved):
thanks for making the changes and adding tests!

guidari (approved):

PR #14919, opened 2023-10-16
fix(DatePicker): ensure icon is correct if label extends past input

### LINKED ISSUE #14907: [Bug]: Date Picker Component. The calendar icon for React, Vue and Angular components, is following the label width
### Package

carbon-components

### Browser

Chrome, Safari, Firefox

### Package version

v11.40.0

### Description

When setting a Label for the component Date Picker, if the Label is too long, the calendar icon is following the Label width. So the icon is being dragged out of the component.

### Screenshots

StoryBook. 

## Vue - Single using Date
image

## React - Single With Calendar
image

## Angular - Single
image

### Steps to reproduce

- Import a Date Picker (Single) component and add it to the project

- Set a label that is larger than the width of the component.

- The icon should be dragged out of the component. 

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14907

Adds a wrapper around the input and SVG so that the position of the icon is always computed based on the input rather than the container

#### Changelog

**New**

- `span` around `input` and `svg`

**Changed**

- moved `position: relative` to the outer wrapper to the new `span`

#### Testing / Reviewing

Add a really long label to `Single DatePicker` and ensure the icon stays in the correct spot

### DIFF
--- a/packages/react/src/components/DatePickerInput/DatePickerInput.tsx
+++ b/packages/react/src/components/DatePickerInput/DatePickerInput.tsx
@@ -216,13 +216,15 @@ const DatePickerInput = React.forwardRef(function DatePickerInput(
         </label>
       )}
       <div className={wrapperClasses}>
-        {input}
-        {isFluid && <DatePickerIcon datePickerType={datePickerType} />}
-        <DatePickerIcon
-          datePickerType={datePickerType}
-          invalid={invalid}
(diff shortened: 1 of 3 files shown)


guidari (approved):

tay1orjones (approved):
Nice!

PR #14927, opened 2023-10-17
Datepicker allow input props issue

### LINKED ISSUE #14111: [Bug]: DatePicker allowInput prop doesn't work
### Package

@carbon/react

### Browser

_No response_

### Package version

1.32.0

### React version

v18.2

### Description

The `DatePicker` component has a prop called `allowInput` which by the description it provides should avoid the user to be able to type in the Input. The prop should restricts all input when set to false.

Reference issue: https://github.com/carbon-design-system/carbon/issues/7007
Reported on [Slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1687958132849259)
FlatPickr example: https://flatpickr.js.org/examples/ (first item)

Maybe we should switch the prop to the `DatePickerInput` and take care of that feature there instead of `DatePicker`

### Reproduction/example

https://stackblitz.com/edit/github-qqiayn?file=src%2FApp.jsx

### Steps to reproduce

Set the prop to false and try to type something.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

omkarajagunde:
Closes #14111

We were passing the prop `allowInput` but weren’t listening for the changes in the prop, hence `useEffect` was added to listen to the changes in prop and accordingly set the boolean value in the Flatpicker instance, 

After adding useEffect then Flatpicker by default clears whole date on Delete/Backspace keyboard button press when allowInput is set to false, 

we don't want anything to happen if allowInput is false when we try to edit/modify date hence a if case was added in `fixEventsPlugin.js` where we stop the default behaviour

Demo after the fix - 

https://github.com/carbon-design-system/carbon/assets/50138744/8abb6b45-eabd-466b-b8c2-fb9eaff2181e

#### Changelog

**New**

- `useEffect` was added to listen to allowInput prop changes.
- `if` case was added to refrain the default Flatpicker behaviour of clearing everything.

**Changed**

- No existing code was changed

**Removed**

- No existing code was removed

#### Testing / Reviewing

- In DatePicker storybook playground, change the allowInput prop to true, now we should be able to edit the date manually, now set the allowInput to false, now we can select the date in calendar but we no longer can edit the date
- verify whether the above point holds true across all datePickerType


omkarajagunde:

omkarajagunde:

omkarajagunde:
recheck

omkarajagunde:

packages/react/src/components/DatePicker/DatePicker.tsx
```diff
@@ -817,7 +826,10 @@ const DatePicker = React.forwardRef(function DatePicker(
 DatePicker.propTypes = {
   /**
    * flatpickr prop passthrough. Allows the user to enter a date directly
-   * into the input field
+   * into the input field, Note: If we set this flag to false and a user picks a date, 
+   * they won't be able to remove it by just clicking on it. 
+   * Instead, you'll need to set the value to an empty string yourself if you want to clear the date. 
+   * This will make the input behave like a controlled form element.
```
guidari: I few like the explanation is too much to address for the user. Might be good to shorten a little bit.

guidari (approved):
Awesome! Thanks for the contribution!! 🚀

Just one note.

packages/react/src/components/DatePicker/DatePicker.tsx
omkarajagunde: Sure will update this one

omkarajagunde: @guidari comment updated, let me know if there are any other changes?

omkarajagunde:
@tw15egan can you add your review to this Pr and then maybe merge if looks good?

tw15egan (approved):

PR #14935, opened 2023-10-17
feat(Combobox): allow custom value

### LINKED ISSUE #14006: [Combobox unique value]: New value dev implementation
Ref link for implementation https://github.com/carbon-design-system/carbon/issues/13903

#### Acceptance Criteria

Allow users to type in a new value in the ComboBox and accept it as a new value

- [x] Add a new prop to enable the behavior
- [x] Typing and hitting enter should no longer clear the input as described in the spec: https://github.com/carbon-design-system/carbon/issues/13903#issuecomment-1577463748
- [x] Evaluate if we need to modify the selectedItem state, but users could probably get the value without this even using target.value in the onChange callback if we have a ref on the input

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14006

Adds a new prop, `allowCustomValue`, that will retain the value typed into the `Combobox` on blur, and allow the menu to be closed with `Enter` when the value does not match an item in the list. 

#### Changelog

**New**

- `allowCustomValue` prop added to `ComboBox` so that a user can have their value retained on `Enter` or `Blur` if it does not match any item in the list. 
- Tests to cover the new functionality
- Docs to explain the new prop

**Changed**

- Adjusted types for the `onChangeData` to account for this new use case 

#### Testing / Reviewing

Go to the test story, and type in an item that is _not_ in the item list. You should be able to accept this value with Enter, and it should stay in the `Combobox` when the focus is lost.


thyhmdo (approved):
This works great! Thank you @tw15egan

tay1orjones (approved):
Looks great!

packages/react/src/components/ComboBox/ComboBox.stories.js
```diff
@@ -79,6 +79,33 @@ export const Default = () => (
...
+          console.log(e);
+        }}
+        id="carbon-combobox"
+        items={['Apple', 'Orange', 'Banana', 'Pineapple', 'Raspberry', 'Lime']}
+        downshiftProps={{
+          onStateChange: () => {
+            console.log('the state has changed');
+          },
+        }}
+        // itemToString={(item) => (item ? item.text : '')}
```
andreancardona: do we want to keep this commented out here? @tw15egan 

Other than this - looks good!

tw15egan: Removed 👍🏻

alisonjoseph (approved):

PR #14942, opened 2023-10-18
fix(react): uishell headermenu detect active descendants

cordesmj:
#### Changelog

**Changed**

- Updated the React HeaderMenu in UIShell so that it detects active descendant links when more than two levels of navigation are in the page heirarchy

#### Testing / Reviewing

- Create HeaderNavigation with three levels of nesting.
- Set a third-level link to 'isActive'
- Notice that the parent link in the main navigation is marked as active when the dropdown is not opened.

Related to https://github.com/carbon-design-system/carbon/pull/15027


tay1orjones (commented):
@laurenmrice Do we want to encourage or allow multiple levels of nesting in the UIShell header submenus? [The current spec](https://carbondesignsystem.com/components/UI-shell-header/usage) doesn't really specify one way or another that I could see, though the examples do not include any examples of submenu nesting.

cordesmj:
> @laurenmrice Do we want to encourage or allow multiple levels of nesting in the UIShell header submenus? [The current spec](https://carbondesignsystem.com/components/UI-shell-header/usage) doesn't really specify one way or another that I could see, though the examples do not include any examples of submenu nesting.

For context, this is to support some extensions to Carbon to handle three levels of nesting for w3 Publisher, if that helps.

kingtraceyj:
@cordesmj hi! could we learn a little more about your need for the 3rd level? We have purposefully limited that ability from the header to limit cognitive load for our users. our .com team has also limited that ability as well in their header designs for the same reasons.

cordesmj:
> @cordesmj hi! could we learn a little more about your need for the 3rd level? We have purposefully limited that ability from the header to limit cognitive load for our users. our .com team has also limited that ability as well in their header designs for the same reasons.

Please see my response [here](https://github.com/carbon-design-system/carbon/pull/15027#issuecomment-1783477387)

kingtraceyj:
@cordesmj @tay1orjones after further discussion with the team, we don't want to be a blocker for your team to implement your requirements. for side navigation, teams have asked for the third level for sometime and we ourselves have done design explorations around this. our specs can be found in figma.

we are a bit more cautious about opening up a third level for header navigation—it can begin to complicate the information architecture structure pretty quickly and hide information. but, again, we don't want to block you from delivering your specific requirements so we can push this through as well.

thanks for contributing!

tw15egan (approved):

tay1orjones (approved):

PR #9721, opened 2021-09-22
chore(tile): refactor to functional component

sstrubberg:
REF #9712 

Refactored most of the Tile component to make functional components using `React.forwardRef` to ensure refs were being forwarded appropriately.

- [ ] Tile
- [ ] ClickableTile
- [ ] ~ExpandableTile~
- [ ] TileAboveTheFoldContent
- [ ] TileBelowTheFoldContent

#### Notes

- I outright gutted the `getDerivedStateFromProps` from `ClickableTile`. Everything seems still be working without it and the [guidance from React](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops) suggests that you rarely need it.

#### Testing / Reviewing

1. Pull down the PR locally.

2. Run `yarn start:v11` from `packages/react`

4. Ensure that the following stories for Tile are all working as expected: 
- [ ] `Default`
- [ ]  `ClickableTile`
- [ ]  `Multi Select`
- [ ]  `Radio`

Note: I have not refactored `ExpandableTile` in this PR.

PR #14945, opened 2023-10-19
fix(tile): `ClickableTile` should navigate on keypress

tay1orjones:
[Reported on slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1697719961531719)

I can't find a good reason as to why the keydown handler would need to call `preventDefault()`. I _think_ it was added by accident during the functional refactor, https://github.com/carbon-design-system/carbon/pull/9721

#### Changelog

**Changed**

- No longer call `evt.preventDefault()` in the keydown handler

#### Testing / Reviewing

- In the ClickableTile story: focus the tile, press enter or space, the clickabletile should navigate. It previously would not navigate.
- This behavior should match what happens on click of the clickabletile.

### DIFF
--- a/packages/react/src/components/Tile/Tile.tsx
+++ b/packages/react/src/components/Tile/Tile.tsx
@@ -163,7 +163,6 @@ export const ClickableTile = React.forwardRef<
   function handleOnKeyDown(evt: KeyboardEvent) {
     evt?.persist?.();
     if (matches(evt, [keys.Enter, keys.Space])) {
-      evt.preventDefault();
       setIsSelected(!isSelected);
       onKeyDown(evt);
     }


tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):
tested & looks good!

alewitt2:
I know this issue is already merged, but i just ran into a related issue and i think i may have insight as to why `preventDefault()` was defined and also a another bug in the code

```
  function handleOnKeyDown(evt: KeyboardEvent) {
    evt?.persist?.();
    if (matches(evt, [keys.Enter, keys.Space])) {
      setIsSelected(!isSelected);
      onKeyDown(evt);
    }
    onKeyDown(evt);
  }
```

If a user defines their own callback for `onKeyDown`, then there is a good chance they want to control what happens when any key is pressed (including enter and space). That said.. the way the code was and is written, doesnt really solve the that problem for the user. Something like the following would be required:

```
  function handleOnKeyDown(evt: KeyboardEvent) {
    evt?.persist?.();
    if (matches(evt, [keys.Enter, keys.Space])) {
      setIsSelected(!isSelected);
      if (user defined onKeyDown) { // would need some logic to understand when a user defined the callback or not
        evt.preventDefault();
        onKeyDown(evt);
      }
    } else { // this else is necessary to prevent the callback being called twice when enter/space keys are pressed.
      onKeyDown(evt);
    }
  }
```

PR #14987, opened 2023-10-21
Fix: rework `DatePickerInput` type to extend `HTMLInputElement` instead of `HTMLDivElement`

dumaron:
Hey Carbon Team! I think the `DateInputPicker` type could be improved by making it extend `HTMLInputElement` instead of `HTMLDivElement`. This would allow us to pass the `name` prop to the component since `name` is available in inputs but not in divs.

I have attached a screenshot of the issue I have on my local environment since it seems that codesandbox is not able to detect the type issue.

Screenshot 2023-10-21 alle 13 56 51

#### Changelog

**Changed**

- replaced `HTMLDivElement` with `HTMLInputElement` in `packages/react/src/components/DatePickerInput/DatePickerInput.tsx`

#### Testing / Reviewing

I tested the new type by changing it on my local dev environment and using `bun link` to link the local package inside the software where I use the `DatePickerInput` component.


dumaron:

tay1orjones (approved):
Thanks for this!

andreancardona (approved):
Thank you for contributing @dumaron !

PR #14998, opened 2023-10-23
feat(Slug): implement AI Slug

### LINKED ISSUE #14947: C4AI Slug: Implement the interactive Slug component in `@carbon/react`
I think we can run this in tandem with the spec work in https://github.com/carbon-design-system/carbon/issues/14948

At the very least we could begin to scaffold out the component with an initial take on the design with room for interpretation as the spec firms up.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14947

Adds in the `unstable__slug` component

#### Changelog

**New**

- `unstable__Slug` component with `default`, `hollow`, and `inline` variants
- added a few more "temporary" AI tokens to handle light v dark mode (we can change these later)

#### Testing / Reviewing

Go to `unstable__Slug` under the experimental section at the bottom and verify against the spec


tw15egan:
@aagonzales pushed this up to get some feedback, spacing is definitely off on inline variants, but should be easy fixes. Wanted to get it up ASAP so we can iterate

aagonzales (changes requested):
I'm still reviewing but want to go ahead and post things as I see them. 

Adjustments needed for the inline slug: 
- corner radius: 1px
- padding top/bottom: 4px
- padding left/right: 8px

(top to be, bottom as-is)
image

aagonzales:
A couple more tweaks for the inline with text: @tw15egan 
- the padding top/bottom is still too big, can you take it down to 2px
- padding between AI and Text should be 4px not 8px
image

packages/react/src/components/Slug/Slug.js
```diff
@@ -0,0 +1,136 @@
...
+    kind,
+    size = 'xs',
+    slugContent,
+  },
+  ref
+) {
+  const prefix = usePrefix();
+
+  const slugClasses = cx(className, {
+    [`${prefix}--ai-slug`]: true,
```
tay1orjones: Not a blocker, but something to consider as this is still experimental, we might not need/want `ai` in the class names. My thinking is similar reasoning as to why the component is named `Slug` and not `AISlug`, re: eventually will we see broad use vs right now we only see this slug used for one thing.

tay1orjones (approved):
Incredible work, looks great to me! 🏆

andreancardona (approved):
looks good on my end!

aagonzales (changes requested):
woo-hoo! TJ these look awesome. Only two small things and the second isn't even that important. 

1. Its inline slug again, the hollow dot on the two smaller sizes are misaligned by 1px. Can you move the hollow dot over 1px to the left.

image


2. Ok this one probably doesn't matter too much but if its easy would we adjust the text colors of the content inside the example callout (very low priority for Thursday release)

image


Other than that I its good for the first release tomorrow 🎉

tw15egan:
@aagonzales updated 👍🏻

aagonzales (approved):
Looks good for v1, ship it! 👍 ✅

packages/react/src/components/Slug/Slug.js
tw15egan: Removed the `ai-` prefix across the board 👍🏻

PR #14805, opened 2023-10-06
feat(component): add SideNavMenu.tsx component

Nirajsah:
Closes #13606

I am adding a new file `SideNavMenu.tsx` with its types. However, there is an issue with the `@carbon/icons-react` library. It's important to note that there are no other issues with the `SideNavMenu.tsx` file.

#### Changelog

**New**

- Added `SideNavMenu.tsx` file with its associated types.

**Changed**

- None.

**Removed**

- None.

#### Testing / Reviewing

Reviewers, please follow these steps to verify this PR:

1. Pay special attention to how `@carbon/icons-react` is being used and whether the issue is noticeable.
2. Confirm that there are no other issues introduced by this PR.
3. Please provide feedback on the changes and the issue with `@carbon/icons-react`.

Your help in addressing this issue is much appreciated. Thank you!

PR #15004, opened 2023-10-24
fix(sidenavmenu): remove title attribute, remove duplicate .js file

tay1orjones:
[Surfaced in slack](https://ibm-studios.slack.com/archives/C0M053VPT/p1698162754028629?thread_ts=1697914618.868889&cid=C0M053VPT), the `title` attribute on the SideNavMenu section items is duplicative and causes the screenreader to announce the title twice. It's not needed and can be removed.

This is also a continuation of https://github.com/carbon-design-system/carbon/pull/14805 to clean up the old .js file from the .tsx migration that happened in that PR.

#### Changelog

**Removed**

- Removed the extra `title` attribute
- duplicate outdated SideNavMenu.js file

#### Testing / Reviewing

- The title attribute should no longer be present or be read via screenreaders. Here's a screenshot of the previous/invalid behavior:

image


tw15egan (approved):

andreancardona (approved):
thanks for fixing!

PR #15026, opened 2023-10-25
fix(ClickableTile): duplicate onKeyDown calls

### LINKED ISSUE #15028: [Bug]: ClickableTile double onKeyDown event
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v11.40.0

### React version

16.14.0

### Description

when have defined an `onKeyDown` callback for clickableTile, and you click enter or space, it calls the callback 2 times.

i think its this line that is duplicating the call https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tile/Tile.tsx#L167

### Reproduction/example

https://stackblitz.com/edit/github-ubvput?file=src%2FApp.jsx

### Steps to reproduce

- in the Stackblitz, the @carbon/react template is a little broken, so you'll need to comment out the line in _sliders.scss thats preventing the project from running and then re-run the project with `yarn dev`

- open the dev tools console
- click on the webpage (not on the tile)
- tab focus on the tile
- press Enter or Space (this will bring you to a new webpage, just close the new tab)
- look at the dev tools console and see that the event logged twice.
- Assuming you are still focused on the tile, push any other key and see that it only fires 1 event.

### Suggested Severity

### Application/PAL

IBM Cloud Satellite, Cloud PAL

### Code of Conduct

alewitt2:
Shouldnt call onKeyDown callback twice when enter or space is pressed.

If a user defines their own callback for onKeyDown, then there is a good chance they want to control what happens when any key is pressed (including enter and space). So preventDefault should be called for that use case.

reference: https://github.com/carbon-design-system/carbon/pull/14945#issuecomment-1780034376

Closes: #15028

ClickableTile double onKeyDown event

#### Changelog

**Changed**

- Fixes bug around ClickableTile onKeyDown

#### Testing / Reviewing

- create a ClickableTile like the following 
```
<ClickableTile onKeyDown={()=>console.log('my-event')} href="www.ibm.com">
some text
</ClickableTile>
```
- tab to the tile and press enter. the tile should log "my-event" 1 time and not link to the href. previously it would link to the href and log "my-event" 2 times.

### DIFF
--- a/packages/react/src/components/Tile/Tile.tsx
+++ b/packages/react/src/components/Tile/Tile.tsx
@@ -137,7 +137,7 @@ export const ClickableTile = React.forwardRef<
     href,
     light,
     onClick = () => {},
-    onKeyDown = () => {},
+    onKeyDown,
     renderIcon: Icon,
     ...rest
   },
@@ -164,9 +164,13 @@ export const ClickableTile = React.forwardRef<
     evt?.persist?.();
     if (matches(evt, [keys.Enter, keys.Space])) {
       setIsSelected(!isSelected);
+      if (onKeyDown) {
+        evt.preventDefault();
+        onKeyDown(evt);
+      }
+    } else if (onKeyDown) {
       onKeyDown(evt);
     }
-    onKeyDown(evt);
   }
 
   const v12DefaultIcons = useFeatureFlag('enable-v12-tile-default-icons');


alewitt2:

alewitt2:
closing in favor of https://github.com/carbon-design-system/carbon/pull/15029

PR #15029, opened 2023-10-26
fix(Tile): `ClickableTile` should focus when no href prop

### LINKED ISSUE #15024: [a11y]: Focus not working in ClickableTile when no href passeed
### Package

@carbon/react

### Browser

Chrome

### Operating System

_No response_

### Package version

1.35.0

### React version

1.35.0

saw in sandbox as well

### Assistive technology

_No response_

### Description

I am using a group of ClickableTiles to display clickable information. 
I do not set the optional  "href"  property for them (a click is expected to open in the future a side panel).
In this setting I do not see the tiles getting "focus" when hitting tab. 

While investigating this we noticed that if we do specify a "href"  property value, we do get the expected "focus" behavior.

[We are using 
 version: "1.35.0" - but did a test in pure sand box for this test code also using 1.40.0 and see the same.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

<div className="container">       <ClickableTile         data-auto-id="clickable-tile-1"         id="clickable-tile-1"         value="1"         clicked       >         Tile 1       </ClickableTile>       <ClickableTile         data-auto-id="clickable-tile-2"         id="clickable-tile-2"         value="2"         onClick={clickAction}       >         Tile 2       </ClickableTile>       <ClickableTile         data-auto-id="clickable-tile-3"         id="clickable-tile-3"         value="3"         onClick={clickAction}       >         Tile 3       </ClickableTile>     </div>

### Steps to reproduce

Add few ClickableTile instances to a div
Don't specify a href property
Once running click tab -> they are not passed through

(if u do place an href property value - doing the above will pass through the tiles)

e.g.:
<ClickableTile
        data-auto-id="clickable-tile-1"
        id="clickable-tile-1"
        value="1"
        clicked
      >
        Tile 1
      </ClickableTile>
      <ClickableTile
        data-auto-id="clickable-tile-2"
        id="clickable-tile-2"
        value="2"
        onClick={clickAction}
      >

### Code of Conduct

### LINKED ISSUE #15028: [Bug]: ClickableTile double onKeyDown event

alewitt2:
If no href is defined, clickableTile should still be able to gain focus. Also removes the duplicate onKeyDown call which causes the callback to run twice, which is undesirable when a user is trying to do something with the enter/space key.

Closes #15024 
Closes #15028

fixes Issues around using keyboard to navigate clickableTile

#### Changelog

**Changed**

- Allows user to tab to clickableTile when no href is defined
- removes the duplicate onKeyDown callback call when user presses "enter" or "space".

#### Testing / Reviewing

- create a ClickableTile like the following 
```
<ClickableTile onKeyDown={()=>console.log('my-event')}>
hello world
</ClickableTile>
```
- tab to the tile and press enter. the tile should log "my-event" 1 time. previously it would not allow you to tab to it until you added `tabIndex={0}`, and then it would log "my-event" 2 times.


tay1orjones (approved):
Thanks for this!

tw15egan (approved):
Working as expected, thanks for the fix! 🎉 💪🏻

PR #15033, opened 2023-10-26
feat: Add Typescript typings to Breadcrumb, BreadcrumbItems, and BreadcrumbSkeleton

Tresau:
Closes #13546 

Add typings to Breadcrumb, BreadcrumbItems, and BreadcrumbSkeleton.  During work, discovered incorrect PropType in BreadcrumbItem for 'aria-current' and opened #15034 as a result.

#### Changelog

**New**

- Added Typescript types to Breadcrumb, BreadcrumbItems, BreadcrumbSkeleton

**Changed**

- File extensions from .js to .tsx

**Removed**

- n/a

#### Testing / Reviewing

Verify that the storybooks run properly.
Run unit tests

PR #15052, opened 2023-10-27
fix: BreadcrumbItem aria-current PropType fix

### LINKED ISSUE #15034: [Bug]: Proptype for aria-current in BreadcrumbItem is incorrect
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.40.0

### React version

_No response_

### Description

The PropType definition for 'aria-current' is to broad.  it is currently defined as `'aria-current': PropTypes.oneOfType([PropTypes.string, PropTypes.bool])` however it will need to be modified to match `AriaAttributes['aria-current']`.  This is causing issues with adding Typescript typings as well as the possibility of breaking if the user uses a specific type of <BreadcrumbItem> and sets a non standard 'aria-current' value.

### Reproduction/example

https://codesandbox.io/s/carbon-aria-current-f9lcrj?file=/src/App.js

### Steps to reproduce

This does not show a break, however you can see that eslint is complaining that non standard strings can be used.  There is a code path where the BreadcrumbItem can create a <Link> component and pass it the aria-current value.  <Link> is already typed and only accepts the standard 'aria-current' values as shown below. 
image

### Suggested Severity

### Application/PAL

Cloud

### Code of Conduct

Tresau:
Closes #15034 

BREAKING CHANGE as it makes a PropType more specific.

The PropType for 'aria-current' in BreadcrumbItem is to broad, this change makes the definition more specific to match AriaAttributes['aria-current'].  Depends on #15033 as this PR is branched from that issue.

#### Changelog

**New**

- n/a

**Changed**

- aria-current proptypes of BreadcrumItem from `string, bool` to `bool, 'false', 'true', 'page', 'step', 'location', 'date', 'time'`

**Removed**

- n/a

#### Testing / Reviewing

Verify that the storybook has updated PropTypes
Verify tests pass
image


Tresau-IBM:
This will require a snapshot update but I am not certain how best to alter it.

alisonjoseph:
@Tresau-IBM you should able to run `yarn test -u` to update the snapshot, there is also a merge conflict that popped up

Tresau-IBM:
> @Tresau-IBM you should able to run `yarn test -u` to update the snapshot, there is also a merge conflict that popped up

Thank you, will address merge conflict and update snapshot, thank you

packages/react/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+import cx from 'classnames';
+import Link from '../Link';
+import { OverflowMenuHorizontal } from '@carbon/icons-react';
+import { usePrefix } from '../../internal/usePrefix';
+import { ForwardRefReturn } from '../../types/common';
+import { Text } from '../Text';
+
+export interface BreadcrumbItemProps
+  extends React.HTMLAttributes<HTMLLIElement> {
+  'aria-current'?: AriaAttributes['aria-current'];
```
tay1orjones: The TypeScript types aren't bound by semver right now and are provided on an as-is basis. I think it's fine to specify this here. It won't be classified as a breaking change.

packages/react/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+    PropTypes.oneOf([
+      'false',
+      'true',
+      'page',
+      'step',
+      'location',
+      'date',
+      'time',
+    ] as const),
+  ]),
```
tay1orjones: I'm not sure this warrants a breaking change either. [The spec](https://www.w3.org/TR/wai-aria-1.2/#aria-current) is quite clear on what values are allowed, though there is one exception (I bolded below for emphasis)

> The [aria-current](https://www.w3.org/TR/wai-aria-1.2/#aria-current) [attribute](https://www.w3.org/TR/wai-aria-1.2/#dfn-attribute) is a token type. **Any value not included in the list of allowed values SHOULD be treated by [assistive technologies](https://www.w3.org/TR/wai-aria-1.2/#dfn-assistive-technology) as if the value true had been provided.** If the attribute is not present or its value is an empty string or undefined, the default value of false applies and the [aria-current](https://www.w3.org/TR/wai-aria-1.2/#aria-current) [state](https://www.w3.org/TR/wai-aria-1.2/#dfn-state) MUST NOT be exposed by user agents or assistive technologies.

So technically any string is allowable and coalesces to `true`. You could argue these proptypes (plus the typings from React) are incorrect, but I don't think this presents much risk to have specific strings instead of allowing any string. If we get user feedback that some use case is missed, we can roll back this change very easily.

tay1orjones (approved):

tw15egan (approved):
LGTM 👍 ✅

tw15egan (approved):

PR #15057, opened 2023-10-28
fix: Modal Close button a11y issues

### LINKED ISSUE #13232: Add TypeScript types to IconButton

### LINKED ISSUE #14323: [Bug]: Modal close button uses wrong title
### Package

@carbon/react, @carbon/elements

### Browser

Chrome

### Package version

"@carbon/elements": "11.26.0", "@carbon/react": "1.34.0",

### React version

v18.2.0

### Description

Modal close button uses the wrong title even though the closeButtonLabel is set:

Screenshot 2023-07-27 at 15 16 15

Issue can be reproduced in Storybook: https://react.carbondesignsystem.com/?path=/story/components-modal--playground

According to this code, title gets its value from ariaLabel. Shouldn't it get it from closeButtonLabel? https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Modal/Modal.js#L210

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-modal--playground

### Steps to reproduce

Hover on the close button in Modal component.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

### LINKED ISSUE #14988: Modal: Tooltip is missing from the Close button on focus
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

V10, V11

### React version

v. 1.40

n/a

### Assistive technology

_No response_

### Description

When navigating with keyboard on a modal, tooltip is missing from the 'Close' button on focus. Tooltip is required for non-text content. 
image

Here's Carbon's accessibility guidance on buttons: 
Icon-only buttons have their labels exposed automatically on hover and focus.
https://carbondesignsystem.com/components/button/accessibility#behavior
image

  
Also, when using a mouse to hover the 'Close' button, the tooltip is incorrect. (This is also covered in [issue #14323](https://github.com/carbon-design-system/carbon/issues/14323))

image

### WCAG 2.1 Violation

Success Criterion 1.1.1: Non-text Content

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-modal--default&globals=theme:white

### Steps to reproduce

The violation can be found in Carbon site or storybook.

### Code of Conduct

k-rajat19:
Closes #14988 
Closes #14323 
Closes #13232

- Refactored `Modal` to use `IconButton` so that it uses  correct tooltip
-  fixed the value of  `title` 

#### Testing / Reviewing

Tested component in storybook it's working as expected


packages/styles/scss/components/modal/_modal.scss
```diff
@@ -155,7 +155,6 @@
   .#{$prefix}--modal-container {
     position: fixed;
     display: grid;
-    overflow: hidden;
```
k-rajat19: removed this so that tooltip is visible outside the modal

tay1orjones: I think we'll need to keep that to ensure the visual design remains consistent. I think instead we could set the `align` property to `left` to ensure it's not rendered outside the bounds of the modal.

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -387,20 +390,22 @@ const Modal = React.forwardRef(function Modal(
...
-        className={`${modalCloseButtonClass}__icon`}
-      />
-    </button>
+    <div className={`${prefix}--modal-close-button`}>
+      <IconButton
+        className={modalCloseButtonClass}
+        label={closeButtonText}
+        onClick={onRequestClose}
+        title={closeButtonText}
+        aria-label={closeButtonText}
```
tay1orjones:
```suggestion
        aria-label={closeButtonText}
        align="left"
```

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -387,20 +390,22 @@ const Modal = React.forwardRef(function Modal(
...
-      <Close
-        size={20}
-        aria-hidden="true"
-        tabIndex="-1"
-        className={`${modalCloseButtonClass}__icon`}
-      />
-    </button>
+    <div className={`${prefix}--modal-close-button`}>
+      <IconButton
+        className={modalCloseButtonClass}
```
tay1orjones, line 440: Is the wrapper `div` necessary, or could we just combine these classes onto the one element?

tay1orjones (changes requested):
Thanks for submitting this PR! Just a couple things to address below, let me know what you think.

packages/react/src/components/Modal/Modal.tsx
k-rajat19: Earlier I had tried doing this with `modalCloseButtonClass` but find hard to position the button, I think we need this wrapper in order to position button correctly

packages/styles/scss/components/modal/_modal.scss
k-rajat19: Yeah sure 👍

tay1orjones:
I saw CI was getting a compiler error, so I went ahead and pushed an update here that includes new TypeScript types for IconButton

packages/react/src/components/IconButton/index.tsx
```diff
@@ -5,16 +5,96 @@
...
+  /**
+   * Specify whether the IconButton is currently selected
+   */
+
+  isSelected?: boolean;
+
+  /**
+   * Specify the type of button to be used as the base for the IconButton
+   */
+  kind?: 'primary' | 'secondary' | 'ghost' | 'tertiary';
```
tw15egan: Seems like there is a type mismatch here since we don't allow danger icon buttons, but `danger` is a valid `kind` for `Button`

tay1orjones: I just updated the types to use the formal Button `kind` types, and omit the danger variants that we don't allow for IconButtons

tay1orjones (approved):

packages/react/src/components/IconButton/index.tsx
```diff
@@ -5,16 +5,99 @@
...
 import cx from 'classnames';
 
-const IconButton = React.forwardRef(function IconButton(props, ref) {
-  const {
+interface IconButtonProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  /**
+   * Specify how the trigger should align with the tooltip
+   */
+  align?:
```
tw15egan, line 29: Seems like this `align` is missing `left-bottom`, `left-top`, `right-bottom`, `right-top`. Based on the underlying `Tooltip`, we should be able to pull these in from `Popover`


```jsx
import { PopoverAlignment } from '../Popover';
  /**
   * Specify how the trigger should align with the tooltip
   */
  align?: PopoverAlignment;
```

tw15egan (changes requested):

packages/react/src/components/IconButton/index.tsx
tay1orjones: @tw15egan Do you know if that is intentional or an actual oversight? The proptype below doesn't include these and is only: 
```
align: PropTypes.oneOf([
    'top',
    'top-left',
    'top-right',
    'bottom',
    'bottom-left',
    'bottom-right',
    'left',
    'right',
  ]),
```

@laurenmrice is the iconbutton limited to only certain alignment options for the tooltip?

tw15egan: @tay1orjones seems like an oversight, not sure why we wouldn't have those two extra alignments for `left` and `right`?

tay1orjones: @tw15egan I'm not sure why, but back in the original PR those alignments were explicitly not included. https://github.com/carbon-design-system/carbon/pull/10095#pullrequestreview-807656822

tw15egan: Oh wow, nice find 😄  Must be because the `IconButton` tooltip is only supposed to be a few words at most (not breaking to a new line) and the other variants are not needed? We can wait for @laurenmrice to confirm if we need to omit those `Popover` types from `IconButton` or add them in 👍🏻

tay1orjones: Just confirmed we likely didn't have all of them at first because IconButton tooltips are usually only a word or two and might not need all the extra alignments.

Lauren [mentioned](https://ibm-studios.slack.com/archives/GD8JG6JR3/p1699303094236079) we can add them all in though, just in case people need them. I'll update the PR.

packages/react/src/components/IconButton/index.tsx
tay1orjones: Hey I just took another look and this mismatch persists even when using Omit on the ButtonKind type. I'll look into it.

jdharvey-ibm:
@tay1orjones @tw15egan I pushed some updated type stuff here that'll hopefully resolve the issues, but I left some TODOs in the code that need expert input. LMK if you want to discuss any of them!

tw15egan:
@tay1orjones I just pushed up some changes that should fix the Percy issues. Could you take another look before we (finally) get this merged in?

tw15egan (approved):
LGTM 👍 ✅

tay1orjones:
@tw15egan Thank you! and ah I forgot we had those TODO's in here 🤦‍♂️ Thoughts on merging as is? We could follow up with more comprehensive proptypes to address the TODOs later?

tw15egan:
@tay1orjones I'm fine merging this as-is and addressing those items separately, as this PR seems to blocking a few others

tay1orjones (approved):

Issue #15062, opened 2023-10-30
[Bug]: sass loading warning about deprecated use of `/`


alewitt2:
### Package

@carbon/react

### Browser

Chrome

### Package version

11.41.1

### React version

16.14.0

### Description

![image](https://github.com/carbon-design-system/carbon/assets/48691328/0f142611-5f4f-46a6-bbd4-39dac5d2c51b)

offending line https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/slider/_slider.scss#L149

because the `/` is wrapped in a `#{}` it doesnt count as being inside the calc

### Reproduction/example

na

### Steps to reproduce

try to pull in the latest 11.41.1 into a repo and include `@use '@carbon/react';`

### Suggested Severity

Severity 3 = workaround

You can use the --quiet-deps and/or --quiet option to tell sass not to emit these warnings
https://sass-lang.com/documentation/cli/dart-sass/#quiet

### Application/PAL

Satellite-UI

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

urbanisierung:
I was experiencing the same `warning` - thanks for opening the bug and thanks for the quick fix! Upgrading to `1.41.2` solves the issue. I think the bug can be closed now.

tw15egan:
Thanks for confirming 🙏🏻

PR #15063, opened 2023-10-30
fix(bug): slider.scss uses deprecated `/`

alewitt2:
an accidental deprecated division `/` snuck in and is causing sass loader errors downstream.

Closes # https://github.com/carbon-design-system/carbon/issues/15062

cant load `@use carbon/react` in scss

#### Changelog

**Changed**

- Fixes usage of deprecated division `/`

#### Testing / Reviewing

make sure slider still works as expected, and no sass warnings

### DIFF
--- a/packages/styles/scss/components/slider/_slider.scss
+++ b/packages/styles/scss/components/slider/_slider.scss
@@ -146,7 +146,7 @@
       block-size: convert.to-rem(2px);
       content: '';
       inline-size: convert.to-rem(6px);
-      inset-block-start: calc(50% - #{calc(convert.to-rem(2px) / 2)});
+      inset-block-start: calc(50% - #{convert.to-rem(2px) * .5});
       inset-inline-end: 0;
     }
 
--- a/packages/styles/scss/components/slider/_slider.scss
+++ b/packages/styles/scss/components/slider/_slider.scss
@@ -146,7 +146,7 @@
       block-size: convert.to-rem(2px);
       content: '';
       inline-size: convert.to-rem(6px);
-      inset-block-start: calc(50% - #{convert.to-rem(2px) * .5});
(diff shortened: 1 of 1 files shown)


packages/styles/scss/components/slider/_slider.scss
```diff
@@ -146,7 +146,7 @@
       block-size: convert.to-rem(2px);
       content: '';
       inline-size: convert.to-rem(6px);
-      inset-block-start: calc(50% - #{convert.to-rem(2px) / 2});
+      inset-block-start: calc(50% - #{calc(convert.to-rem(2px) / 2)});
```
tay1orjones: I don't think this will properly interpolate the function call. It might be easier to just multiply by half:

```suggestion
      inset-block-start: calc(50% - #{convert.to-rem(2px) * .5});
```

tay1orjones (changes requested):
Thanks for putting this up so quick! Just one minor issue below

tw15egan (approved):

packages/styles/scss/components/slider/_slider.scss
alewitt2: great solution. thanks

packages/styles/scss/components/slider/_slider.scss
```diff
@@ -146,7 +146,7 @@
       block-size: convert.to-rem(2px);
       content: '';
       inline-size: convert.to-rem(6px);
-      inset-block-start: calc(50% - #{convert.to-rem(2px) / 2});
+      inset-block-start: calc(50% - #{convert.to-rem(2px) * .5});
```
alewitt2:
```suggestion
      inset-block-start: calc(50% - #{convert.to-rem(2px) * 0.5});
```

tay1orjones (approved):

PR #15068, opened 2023-10-30
fix(Notification): allow undefined props in TypeScript

lewandom:
Fixes TypeScript type definitions for the Notification component family: makes `kind`, `onCloseButtonClick` and `role` props optional, to match component behavior (defaults set) and PropTypes definition.

#### Changelog

**Changed**

- Notification component family no longer requires setting `kind`, `onCloseButtonClick` and `role` props when used in TypeScript project

#### Testing / Reviewing

In a TypeScript project, using Notification components properly resorts to using default values for `kind`, `onCloseButtonClick` and `role` props.


tay1orjones (approved):
Thanks for fixing this 🙏 sorry we missed it earlier.

alisonjoseph (approved):

PR #15081, opened 2023-11-01
feat(gradient): linear gradient mixin

### LINKED ISSUE #15074: C4AI Gradient: Implement gradient in element and style packages
**Blocked by**
- #15071 
- #15072 
- #15073 
- #14929 

**Notes**
- Sizing should happen after the design issues have been started or closed.
- Issue can be started prior to design issues being completed to identify development constraints.
- Likely be a gradient Sass mixin where you can pass in direction, scale, etc.

```[tasklist]
### Tasks
```

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15074

Adds in an `ai-gradient` mixin (what should we call this?) that can be used to add in the AI aura `linear-gradient` to components. This takes in a `direction` and a `width`, with defaults of `right` and `50%`.

To Do:
- Get proper color tokens for gradient tokens (currently using the `rgba` value found in Figma kit. Do these map to existing colors?

#### Changelog

**New**
- Demo story under `unstable__LinearGradient`
- `ai-gradient` mixin that takes in a `direction` and a `width`
- 3 tokens, which we can rename + add to if needed
    - `ai-gradient-start-01`
    - `ai-gradient-start-02`
    - `ai-gradient-end`

#### Testing / Reviewing

Go to `unstable__linearGradient` and check out the values.


aagonzales:
> Get proper color tokens for gradient tokens (currently using the rgba value found in Figma kit. Do these map to existing colors?

@tw15egan Yes, they do. Its `coolGray-10` and `blue-10`

image

aagonzales:
I also think the word `layer` in the token name is confusing with our actually layer tokens so can we just call those tokens:
- aiGradientStart01
- aiGradientStart02

aagonzales (approved):
Cool! I think this is good for now then. Refined designs of the gradient will come later.

packages/react/src/components/LinearGradient/LinearGradient.stories.js
```diff
@@ -0,0 +1,35 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { Tile } from '../Tile';
+import { TextInput } from '../TextInput';
+import './linear-gradient-story.scss';
+
+export default {
+  title: 'Experimental/unstable__LinearGradient',
```
tay1orjones: I'm not sure about having this in the storybook long term. Feels more like something we should find a way to document in the Themes elements example. Not a blocker for this PR, but something to consider for the future.

tay1orjones (approved):
Looks solid to me. I agree the `ai-gradient` naming seems a bit off but I'm not sure what I'd prefer more. This is experimental so it's fine for now I think. 

I definitely prefer having it specific like this for now instead of just calling it `gradient`. We can make future name changes more specific (or less specific) whereas if we named it `gradient` we'd be much more locked in.

packages/react/src/components/LinearGradient/LinearGradient.stories.js
tw15egan: Yeah I can remove this before we merge 👍🏻

aagonzales:
Yeah I'm struggling to think of a good name for the tokens. "aura" is another option, except that's also what the dynamic, animated element will be called to. Maybe we can call them "ai-aura-dynamic" and "ai-aura-static"?

alisonjoseph (approved):

packages/styles/scss/utilities/_ai-gradient.scss
```diff
@@ -0,0 +1,42 @@
...
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@use '../theme';
+
+/// Experimental - name and structure subject to change. Use at your own risk!
+/// Adds AI gradient styles to a component
+/// @access public
```
mattrosno, line 12: Should `@access private` be used for experimental? If it's `public`, then it's part of our public API as supported.

http://sassdoc.com/annotations/#access

packages/styles/scss/utilities/_ai-gradient.scss
```diff
@@ -0,0 +1,42 @@
...
+/// @param {String} $direction  - Direction of gradient from: `left`, `right`, `top`, and `bottom`
+/// @param {Number} $width - Percentage width of gradient with regards to parent component
+/// @example @include focus-outline('outline');
+/// @group utilities
+
+@mixin ai-gradient($direction: 'right', $width: 50%) {
```
mattrosno, line 18: I'm also questioning our use of `ai-` in token and mixin naming. Following the token naming convention shown [here](https://carbondesignsystem.com/guidelines/color/tokens), would we want to go with `gradient-01`?

This could be more future proof. What happens when everything is "AI". We don't want our code base littered with `ai` prefixes/namespaces. A few years from now, that will seem silly.

mattrosno (commented):
@tw15egan @aagonzales @tay1orjones I understand this has been approved and merged, but here's my 2-cents on naming convention and why it could be beneficial to not use "AI" in any of our token / mixin / component naming.

packages/styles/scss/utilities/_ai-gradient.scss
aagonzales: Yeah @tay1orjones has mentioned the same thing about the "ai" term. We haven't finalized the token name yet just launching something so TJ isn't blocked for the next implementation steps. I think we were going to try and avoid "gradient" as well since that's not scalable (could evolve to something other than a linear gradient). I was thinking it could just be a flavor of "aura" (like `aura-static`) or whatever we end up calling that in the AI world.

PR #15086, opened 2023-11-02
fix(react): publish type definitions for @carbon/react/icons

lewandom:
TypeScript types were added to `@carbon/react/icons` in #14714, but are not consumable, cause... `index.d.ts` is not being published... Oops.

#### Changelog

**New**

- Added `icons/index.d.ts` to the list of files published by `@carbon/react`.

#### Testing / Reviewing

Can only test after this change is released. After it's released, should be able to use TypeScript types for icons.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -18,6 +18,7 @@
     "index.scss",
     "icons/index.js",
     "icons/index.esm.js",
+    "icons/index.d.ts",
     "icons/package.json",
     ".playwright/INTERNAL_AVT_REPORT_DO_NOT_USE.json"
   ],


alisonjoseph (approved):

lewandom:
@alisonjoseph no, that seems to be yet another, different problem with the types I added for icons-react.

tay1orjones (approved):

alisonjoseph:
@lewandom do you know what the issue is, is it something you can easily open a PR for to fix? It's also breaking the design language website.

lewandom:
@alisonjoseph I think I know what the issue is, trying to come up with a fix just now.

PR #15094, opened 2023-11-02
fix(icons-react): use declare keyword in bucket TypeScript types

lewandom:
An attempt to fix TypeScript compilation problem reported in https://github.com/carbon-design-system/carbon-website/pull/3787#issuecomment-1789480156.

#### Changelog

**Changed**

- Use declare keyword in icons-react TypeScript definition files for icon buckets.

#### Testing / Reviewing

The TypeScript typed code using icons-react should compile without issues.


lewandom:
@andreancardona this should fix the problem you spotted. I'm not 100% sure it's the right or complete solution though, since I weren't able to reproduce this problem in my TS projects.

lewandom:
@alisonjoseph meant to tag you, sorry.

alisonjoseph (approved):

tay1orjones (approved):

PR #15115, opened 2023-11-06
feat(Slug): Add `slug` prop to `TextInput`, `TextArea`, `NumberInput`, and `DatePicker`, add revert functionality to `Slug`

### LINKED ISSUE #15089: Add AI enhancements to React `TextInput`, `TextArea`, `NumberInput`, and `DatePicker`
For this issue we'll explore patterns and approaches for extending existing components to be able to contain a Slug with proper placement.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15089

Adds in a `slug` prop for `TextInput`, `TextArea`, `DatePickerInput`, and `NumberInput`

General usage would look like the following:
```jsx
const slug = (
  <Slug>
    <SlugContent>
      <div>
        <p className="secondary">AI Explained</p>
        <h1>84%</h1>
        <p className="secondary bold">Confidence score</p>
        <p className="secondary">
          Lorem ipsum dolor sit amet, di os consectetur adipiscing elit, sed do
          eiusmod tempor incididunt ut fsil labore et dolore magna aliqua.
        </p>
        <hr />
        <p className="secondary">Model type</p>
        <p className="bold">Foundation model</p>
      </div>
      <SlugActions>
        <IconButton kind="ghost" label="View">
          <View />
        </IconButton>
        <IconButton kind="ghost" label="Open Folder">
          <FolderOpen />
        </IconButton>
        <IconButton kind="ghost" label="Folders">
          <Folders />
        </IconButton>
        <Button>View literature</Button>
      </SlugActions>
    </SlugContent>
  </Slug>
);

<TextInput {...TextInputProps} slug={slug} />
```

#### Changelog

**New**

- `slug` prop added so a user can pass down a `Slug` component to be rendered inside a form element
- `<Slug>` has been updated to include two new props, `revertActive` and `onRevertClick`
  - `revertActive` props is to toggle the state of the `Slug`
  - `onRevertClick` is a callback function so that a user can restore the state of the edited input to whatever the stored AI content is

Example usage:
```jsx
const textInputRef = useRef();
const aiValue = 'Generated AI content';
const [isEdited, setIsEdited] = useState(false);

const handleRevertClick = (evt) => {
 setIsEdited(false);
 textInputRef.current.focus();
 textInputRef.current.value = aiValue;
};

const handleChange = (evt) => {
 setIsTextInputEdited(true);
}

const slug = (
 <Slug revertActive={isTextInputEdited} onRevertClick={handleRevertClick}>
   {/* Slug content here */}
</Slug>
);


return (<TextInput ref={textInputRef slug={slug} defaultValue={aiValue} onChange={handleChange} />);
```

**Changed**

- Refactored `TextArea` `.tsx` file to match the general file structure found in other component files.
- Reduced default gradient to `33%`
- Added a `useEffect` to `NumberInput.tsx` to trigger an update when the revert button is clicked to set the value back to the `defaultValue` that was passed in. 

**Removed**
- Removed outdated `Slug.js` file (replaced with `index.js`)
- Removed outdated `redo.svg` and `undo.svg` at `16`, `20`, `24` that were being displayed rather than the correct asset, the `32` one.

#### Testing / Reviewing

Go to `Form` --> `AI Form` to see an example of the `Slug` rendered inside the form inputs. This seems like a clean approach that will not expand the surface area too much across the repo and can be achieved with minimal new styles per component. `Form` --> `Revert Test` is a sandbox to test the `onRevertClick` functionality.


packages/react/src/components/Slug/Slug.js
tw15egan, line 1: This file should have been deleted during the refactor PR

packages/styles/scss/components/fluid-number-input/_fluid-number-input.scss
tw15egan, line 1: Diff will be easier to view when https://github.com/carbon-design-system/carbon/pull/15128 is merged

packages/styles/scss/components/number-input/_number-input.scss
tw15egan, line 1: Diff will be easier to view when https://github.com/carbon-design-system/carbon/pull/15128 is merged

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -164,12 +170,26 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
   );
   const { current: textAreaInstanceId } = useRef(getInstanceId());
 
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
```
tw15egan, line 173: Refactored this file a bit since it was quite scattered. I've basically moved all refs to the top, followed by effects, then class variable declarations, then normalized JSX elements. The only net new code is [L222](https://github.com/carbon-design-system/carbon/pull/15115/commits/995f6cce992242ef539ecbb9670f1304dfd4e392#diff-af10a0e0381c771b14de261caf250073efd2136b83cbbc136c18fcfb0f80aaccR222), [L320](https://github.com/carbon-design-system/carbon/pull/15115/commits/995f6cce992242ef539ecbb9670f1304dfd4e392#diff-af10a0e0381c771b14de261caf250073efd2136b83cbbc136c18fcfb0f80aaccR320-R327), and [L345](https://github.com/carbon-design-system/carbon/pull/15115/commits/995f6cce992242ef539ecbb9670f1304dfd4e392#diff-af10a0e0381c771b14de261caf250073efd2136b83cbbc136c18fcfb0f80aaccR345)

tw15egan:
@tay1orjones this PR is in a good spot to take a look and see if we should keep going down this route

aagonzales:
@tw15egan I've noticed the formatting inside the callout is off. 
image

aagonzales:
I think this way would be the correct error states with both the slug and error icon. 
image

aagonzales:
The other aspect of having AI in these inputs is the user can edit the content away from the AI data. And if that happens, then there needs to be a state change that removes the AI indicators. I'm not sure how complex of a task this would be to add but wanted to point it out. 

image

aagonzales:
> Does the Slug stay active if the input is disabled?

Yes, if the AI disabled the input than the user needs the explainability in the slug to understand why AI disabled the option, ect.

tw15egan:
@aagonzales alright made some updates. 

- Added the callout formatting back in, however this is just storybook styles. Right now, the Callout is just an unstructured Popover that teams will have to add styling and spacing to, unless we want to control each element inside the callout much more. 
- Added toggles to show `invalid`, `warning`, and `disabled` states
- Added a new story that has an example of how to handle the `revert` functionality. The state will need to be handled by application teams, but I wanted to ensure the components themselves were set up to handle it. Already caught an issue with `NumberInput`, so will verify each component going forward.

aagonzales (changes requested):
The revert icon looks like there's an extra 1px of padding on top of it. Can that be removed? The icon seems a little off centered.

image


Also can we move the revert button 8px to the right so the AI icon and revert icon line up. This is how they align at the moment.

image

Everything else looks good though!

tw15egan:
@aagonzales fixed 👍🏻

aagonzales (approved):
Everything looks good to me now 👍 🎉

alisonjoseph (changes requested):
This looks great in Chrome, seeing z-index and positioning issues in firefox.

Screenshot 2023-11-15 at 9 55 02 AM
Screenshot 2023-11-15 at 9 58 19 AM

tw15egan:
Of course, [Firefox doesn't support](https://developer.mozilla.org/en-US/docs/web/css/:has#browser_compatibility) the `:has` selector 🙃  

`layout.css.has-selector.enabled`  set to `true` in `about:config` fixes it, but it seems like the `blur` value is different. 

No positioning was touched for `Slug`, it's just using `autoAlign` by default, but you can pass in whatever specific `align` you need to `slug` to override. Seems like that may be an issue with the `Popover` in general, I can see the same issues in Chrome at specific screensizes.  I can also turn off `autoAlign` by default if that is a smoother solution, but will require positioning to be set.

tw15egan:
@alisonjoseph doing a bit of digging, there isn't an alternative selector to `:has` that would make this easy to fix. However, it seems like support will be added in [Firefox v121](https://caniuse.com/?search=%3Ahas). Looking at their release notes, it seems like they release a new minor each month. Stable ([v119.0](https://www.mozilla.org/en-US/firefox/119.0/releasenotes/)) was released 10/24, [v118.0](https://www.mozilla.org/en-US/firefox/118.0/releasenotes/) was released on 9/26, and [v117.0](https://www.mozilla.org/en-US/firefox/117.0/releasenotes/) on 8/29. Given this seemingly monthly cadence, this means the `:has` support _should_ land in Firefox stable at the end of December. 

Do you think this is fine, given:
- the experimental nature of these components
- it _can_ work behind a feature flag
- Firefox usage is ~2% globally

[v121.0a1 release notes](https://www.mozilla.org/en-US/firefox/121.0a1/releasenotes/)

alisonjoseph (approved):
Yes, I think that's fine @tw15egan

tay1orjones (approved):

PR #15118, opened 2023-11-08
fix(icons-react): avoid special tokens in bucket .d.ts exports

lewandom:
Addresses carbon-website build problem reported in https://github.com/carbon-design-system/carbon-website/pull/3787#issuecomment-1799329706.

#### Changelog

**Changed**

- Generated bucket TypeScript type definition files in icons-react use a safer way to export definitions to avoid conflicts with reserved token names, such as Boolean, Error or Function.

#### Testing / Reviewing

Fragile icon types should import and compile without problems. There should be no changes to icons usage in both TS and non-TS projects.


tay1orjones (approved):
Thanks for this!

tw15egan (approved):

PR #15121, opened 2023-11-08
allow setting tabIndex of MenuButton

joyful-these:
For external issue, fix needed for PR https://github.com/carbon-design-system/ibm-products/pull/3781

70817a91-2628-413c-9dab-888f1c10e368

When the MenuButton is hidden in the DOM, this currently creates an accessibility issue: `Element "button" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'`

This fix allows setting the tabIndex of MenuButton, so it can be set to -1 when the button is hidden.


tay1orjones:
@joyful-these I'm not able to see that issue. Could you outline what bug this fixes?

tay1orjones (approved):

andreancardona (approved):

PR #15128, opened 2023-11-08
refactor(NumberInput): remove pseudo elements and use box-shadow

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/15115

While working on adding the aura gradient to `NumberInput`, I realized our use of `::before` and `::after` elements were causing issues with the underlying backgrounds. This PR removes the usage of these pseudo-elements and instead sets dynamic inset `box-shadow`'s to style the stepper buttons when the input is focused, invalid, or the steppers are hovered while in these states. Also fixes some issues with the current implementation, seen below:

#### Changelog

**Changed**

- Mimicked the `::before` and `::after` pseudo-element styles with `box-shadow` on the stepper buttons.

**Removed**

- All styles relating to the `::before` and `::after`  pseudo-elements
- Some `light` specific styles that are not doing anything 

#### Testing / Reviewing

Go to `NumberInput` and `Unstable__NumberInput` stories, and ensure the stepper styles are correct when the buttons are focused, hovered, hovered when input focused, focused when input invalid, and focused and hovered when input invalid.

### DIFF
--- a/packages/styles/scss/components/fluid-number-input/_fluid-number-input.scss
+++ b/packages/styles/scss/components/fluid-number-input/_fluid-number-input.scss
@@ -71,51 +71,64 @@
     transform: translate(0);
   }
 
-  .#{$prefix}--number-input--fluid .#{$prefix}--number__control-btn,
   .#{$prefix}--number-input--fluid
     input[type='number']:focus
     ~ .#{$prefix}--number__controls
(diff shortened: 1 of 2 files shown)


andreancardona (approved):
tested and looks good! 🔥

tay1orjones (approved):

PR #15129, opened 2023-11-08
feat(slider): add tooltips for single and two handle sliders

### LINKED ISSUE #14549: [Feature Request]: Range Slider - Handle Tooltip Hovers
### The problem

The range slider component currently relies on the user to review the left min or right max value to determine what number is being currently selected. Based on a conversation with the[Design Systems Adoption Guild](https://github.com/carbon-design-system/carbon/issues/6337#issuecomment-1682494447) and the Carbon team, tooltip indicators would be a welcome addition to the pattern. 

### The solution

Upon hover on the minimum or maximum value handle, show a tooltip that displays the number that corresponds to the active handle. 

rangehover

### Examples

| Example | Source | 
| --- | --- | 
| Material| https://m3.material.io/components/sliders/specs |
| Flutter| https://api.flutter.dev/flutter/material/RangeSlider-class.html|

**Audit link:** https://app.mural.co/t/cloudappservicesdesign3998/m/cloudappservicesdesign3998/1692204645335/c54284851ac0a172eef2d7b2e6bd918416cd738e?sender=ua50b53708bb0fd6ce5c12121

### Application/PAL

Carbon for Cloud

### Business priority

### Available extra resources

None, but work has been done on the base version of the Range Slider component via #6337. What needs to be added is the tooltip interaction.

### Code of Conduct

m4olivei:
Closes #14549

Adds tooltips to the Slider componet handles. Currently present by default for both single handle and double handle. Subject to design approval.

#### Changelog

**New**

- Single and two handle Slider's now have a tooltip that displays on focus / hover that will show the current slider value

#### Testing / Reviewing

- [ ] Open the Default Slider story. Should work the same with no regressions. We expect no tooltips when there are text inputs.
- [ ] Open the Slider > Two Handle Slider story. Should work the same with no regressions. We expect no tooltips when there are text inputs.
- [ ] Pay particular attention to regressions in thumb / handle positioning. Was necessary to adjust CSS transforms to accomodate when there are tooltips present.
- [ ] Open the Slider > Slider with Hidden Inputs story. Here we expect tooltips to show on hover / focus.
- [ ] Tooltips should lock open while dragging handles.
- [ ] Tooltips should be positioned per design guidance
- [ ] Pay particular attention to the Tooltips, should behave just like the Tooltip component in isolation in terms of hover / delay / focus behaviours and spacing.
- [ ] Regression testing for Tooltip stories, eg. IconButton > Default, Tooltip > Default and friends.


tay1orjones (commented):
Code side of things looks pretty solid. The interaction when dragging is a bit odd though. The tooltip seems to disappear randomly. Also when you drag and get near the other handle, the other handle's tooltip appears on hover, which probably shouldn't happen. 


https://github.com/carbon-design-system/carbon/assets/3360588/7df091b2-1999-48d1-8ba4-fd452a4df404



Would it make sense to "lock" the tooltip open (or closed) when dragging? @laurenmrice do you have any thoughts?

tay1orjones:
@m4olivei If you haven't already, could you take a look at how this impacts the way in which the component is read via screenreaders? I wonder if it might make sense to hide the tooltips from assistive tech?

tay1orjones:
I also see there's some recent comments on the original issue questioning the approach https://github.com/carbon-design-system/carbon/issues/14549#issuecomment-1804057008

laurenmrice:
See update in connected issue https://github.com/carbon-design-system/carbon/issues/14549#issuecomment-1804762148

m4olivei:
I've pushed some updates based on yesterdays call and subsequent updates to the issue (https://github.com/carbon-design-system/carbon/issues/14549#issuecomment-1804762148)

* Added two stories for when inputs are hidden, for easier testing of tooltip scenarios
* Tooltips only employed when inputs are hidden
* Restored the `$border-interactive` background for the selected track. This was an unintentional regression.
* Removed space on the left hand side for the two handle slider when inputs are hidden. This was a bug that slipped in with the original range slider PR.
* Adjustments to tooltip positioning based on design. @KevinCamelo / @laurenmrice note that there wasn't any guidance provided for the tooltip offset for the default slider. I ended up using `$spacing-03`. Its enough to accomodate the way the thumb scales on hover focus. Let me know if we need to adjust.

Still to do:

* Lock the tooltip open when we're dragging. This is to address the bug @tay1orjones pointed out. @tay1orjones do you know of a way to lock a `<Tooltip />` component open? I can't find one after a cursory glance over the API.

m4olivei:
Thanks @tay1orjones , super insightful.

I took your second suggestion and added some drag detection within the `<Tooltip />` component.

laurenmrice (approved):
Looks good to me from a design perspective, thank you @m4olivei !

One question for dev: I do see that the sliders with hidden inputs are their own distinct stories rather than knobs to turn on/off within the main stories. Just wanted to make sure that was okay with you.

m4olivei:
Thanks @laurenmrice !

> One question for dev: I do see that the sliders with hidden inputs are their own distinct stories rather than knobs to turn on/off within the main stories. Just wanted to make sure that was okay with you.

This was bourne out of [the following requirement](https://github.com/carbon-design-system/carbon/issues/14549#issuecomment-1804762148):

> The range slider component will now use tooltips in an either/or scenario. If text-inputs are present tooltips will not be displayed. If no text-inputs are present, tooltips will be displayed on hover, focus, and selection.

As such, while there isn't an explicit knob to turn tooltips on/off, if you were testing in the Playground story, toggling the `hideTextInput` knob would show/hide tooltips as well accordingly.

There was an initial idea to add a `tooltip` prop, but once we settled on tooltips only being necessary when text inputs were absent, the additional prop was redundant.

KevinCamelo:
@m4olivei Thanks for sharing this. Looking fantastic— here's the small things I observed:

| Issue | Image | 
| -- | --- | 
| Upon checking the base skeleton state on Chrome and Safari, the handle icon gets squashed and stretched. Not sure if this is related to the regressions that happened. | Screenshot 2023-11-21 at 6 44 15 AM | 
| The tooltip only appears on hover, but I believe it should appear upon being active and interacted with. If a user misses the target, they won't have a tooltip unless directly hovered. | Sliders|

Again, awesome work @m4olivei!

CC: @laurenmrice

laurenmrice:
Following on to Kevins comment above^

It may be the fact there are 2 different kinds of active states happening and one of them is a bug? 

1. One is blue and the tooltip appears.
2. One is black and the tooltip disappears. (I think this is the bug)

Nov-21-2023 10-04-12

m4olivei:
Nice catches!

> Upon checking the base skeleton state on Chrome and Safari, the handle icon gets squashed and stretched. 

This is a regression, I hadn't even checked the skeletons. Oops 😊 .

> The tooltip only appears on hover, but I believe it should appear upon being active and interacted with. If a user misses the target, they won't have a tooltip unless directly hovered.

This is a bug. I'm not accounting for and locking the tooltip open, when the drag start begins with a track click, and not a direct click on the handle.

Will work on these probably tomorrow, but maybe today if I can burn through some other tasks.

m4olivei:
@KevinCamelo @laurenmrice 

* Skeleton slider issues are fixed, including a regression that snuck in on the last PR where handles were lost.
* The track click and drag issue is now fixed as well
* I also found and addressed RTL issues for the Skeletons and for the tooltip use cases

Should be ready for another look.

KevinCamelo:
@m4olivei LGTM! 💯 Great work. Love that we were able to sneak in the tooltips. I'm going to let Carbon have final say on if this is good to go but I think it is. CC: @laurenmrice 😄

laurenmrice:
Looks good to me! Thanks @m4olivei !

alisonjoseph (approved):
LGTM! 🥳

tay1orjones (approved):
🥳 🎉

m4olivei:
Yay, approved! @tay1orjones noticed that we got booted from the merge queue. I'm unclear on why, do you have any insight?

tay1orjones:
@m4olivei sorry, not sure. Sometimes the merge queue checks time out without reason. We're still figuring out the best settings for the merge queue. We got this in though just now 👍

m4olivei:
Thanks @tay1orjones !

PR #15219, opened 2023-11-16
feat: paginationNav support high page counts

### LINKED ISSUE #15014: [Feature Request]: Add `disableOverflow` option to PaginationNav to support high page counts
### The problem

We would like to migrate from our custom pagination component to PaginationNav. We have the publications catalogue, that has over 1600 pages. Our custom component has no problems navigating it, but as I'm trying the PaginationNav playground, it looks like it tends to freeze the browser just for thousands of pages. I'm using Chrome 118.

- [Setting `totalItems: 2500` will freeze my browser for about 5 seconds after the pagination is displayed.](https://react.carbondesignsystem.com/?path=/story/components-paginationnav--playground&args=totalItems:2500)
- [Setting `totalItems: 10000` will freeze my browser for about 45 seconds, during which Chrome warns twice that the page is not responding](https://react.carbondesignsystem.com/?path=/story/components-paginationnav--playground&args=totalItems:10000)

### The solution

I'm not sure where exactly the problem is, so it's complicated to propose something. I suppose the problem might by in the select associated with "..." button. It lists thousands of items, which is a bit hard to navigate for the user. So, maybe allow to disable the select to prevent this? Or limit the select to some finite number of items? Maybe something like this:

```
<PaginationNav maxSelectItems={10} />
```

### Examples

https://research.ibm.com/publications

### Application/PAL

https://research.ibm.com/

### Business priority

### Available extra resources

_No response_

### Code of Conduct

kSee04:
This commit updates PaginationNav to support high page counts. I add a prop to PaginationNav which short circuits logic causing present issue

Closes https://github.com/carbon-design-system/carbon/issues/15014

#### Changelog

**New**

- Added new prop `disableOverflow` which fixes the problem elaborated on in the issue.
- Added storybook example named "Overflow Disabled"

**Changed**

- Storybook playground for PaginationNav has been changed to include the new `disableOverflow` prop

**Removed**

- Nothing removed


kSee04:

kSee04:
recheck

tay1orjones (commented):
I think ideally the `...` would no longer be interactive and removed from the focus order, rather than a select without any options. As is, I think it would read oddly with screenreaders.

Could we just put `disabled` on the select and also not render any items into the select? 

@mbgower do you have any thoughts on what an ideal approach could be here? I'm not sure if we'd want the ellipsis to be focusable (and read) or not, and if `disabled` would be enough.

kSee04:
@tay1orjones Thanks for quick review! I've made those requested changes.

alisonjoseph (approved):
This LGTM, I'm not sure we need to keep the new "Overflow Disabled" story you added, I think having the prop in the playground is probably enough. thoughts @tay1orjones

tay1orjones (approved):
Agree, I just updated the PR to remove the test story. Thanks again!

PR #15222, opened 2023-11-16
feat(Tile): add `slug` prop to `SelectableTile` and `ExpandableTile`

### LINKED ISSUE #15187: Add AI enhancements to React `Tile`
> [!WARNING]  
> **Blocked by**
>
> - #15181

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15187

Adds in the `slug` prop for `Tile`, `ClickableTile`, `SelectableTile`, and `ExpandableTile`

#### Changelog

**New**

- `slug` prop added to `Tile components listed above. 
- `slug` prop for `ClickableTile` is `boolean` only, as it will just render an SVG of the hollow slug since it cannot be interactive. @aagonzales I just pulled the SVG from the figma file and [hardcoded it in](https://github.com/carbon-design-system/carbon/pull/15222/files#diff-01b5c6425a2bf9e38da62dd9dd9b2d1ddca77c19566bbe2fd82330d2f81139f8R196). Do we have any plans to add this to the icon library, or is this fine for now?
- New [tokens](https://github.com/carbon-design-system/carbon/pull/15222/files#diff-39ba82f25c140283a9a5947f5a94093a96f511cb66a4f037977584be50f37df4R225-R234) to handle the transition between linear gradients when hovering or selecting a tile. Can probably be reduced when we revisit tokens / token naming. 

**Changed**

- Refactored `slug.scss` to use the new `@mixin callout-gradient`. [This generates](https://github.com/carbon-design-system/carbon/pull/15222/files#diff-558c3556caa7f8d742e0c4ede862525b4c61cb134de3758aa69883aaa65cce45R49) the background gradient for the `Tile`, as well as the `hover` and `selected` states. `@include callout-gradient('hover');`

#### Testing / Reviewing

Go to `Tille --> Slug Test` and review that the `Slug` renders in the correct place, the slug callout renders as expected, and that the hover / selected states of `ClickableTile` and `SelectableTile` are correct


tw15egan:
@aagonzales few observations: 

- I haven't done the static dot grid, but I can try and add what I have from the explorations if we want that. 
- `ClickableTile` currently does not support CTA inside as per [guidance](https://carbondesignsystem.com/components/tile/usage/#clickable). Not sure if we can do this from an a11y perspective, but can take a deeper look
- Should the basic `Tile` support the `slug`? 
- When the tile contains an interactive item, we take away the full tile hover in `ExpandableTile`. This will effect the hover styles in the spec. 


Let me know how it looks and what else you notice.

aagonzales:
@tw15egan For clickable tile let's do this.
- Use the hollow slug and disable the tooltip, it shouldn't need explainability as clickable tile is normally just a link-out type of action instead of an option. 
- Added a spec for the hover values, this will need new tokens I think. 

image

aagonzales:
@tw15egan Selectable tile will need the same hover states and then it also has a selected state change as well.

image

aagonzales:
I'm also wondering if it would be easy to add an optional border radius prop for the callout as tile designs. 

image

tw15egan:
@aagonzales just pushed a bunch of changes, including a bunch of new tokens to handle the gradient shift on hover / selected. WIth regards to the rounded edges, will this always be enabled if `slug` is present? Or can the slug be present and not have rounded edges? Can you have rounded edges and _not_ have a `slug`?

alisonjoseph:
This is looking great @tw15egan! Do we need to adjust the padding on the right of the title to account for the slug?
Screenshot 2023-11-21 at 10 39 39 AM

tw15egan:
@alisonjoseph Hmm the title in there is just for demo purposes... I can fix it here, but not too sure how we can go about systematically accounting for the slug there. This is where I wish you could do rounded corner padding in components!  It would really help with this and other components like `TextArea` and `CodeSnippet`. Any ideas?

alisonjoseph:
@tw15egan oh yea, duh I knew that. That's tricky, we may just have to leave that up the to the teams building out the details. I don't know that we'd want to pad the entire tile over that much, especially the SelectableTile which already has padding to account for the checkbox.

guidari (approved):
LGTM! I didn't caught any issue. 🚀

alisonjoseph (approved):

tw15egan:
@aagonzales updated 👍🏻

aagonzales (changes requested):
Looking good, just one thing. On hover with the rounded corner prop, the hover background still has squared corners. Also the same for the selected state
image

image

aagonzales (approved):
Ship it!

PR #15232, opened 2023-11-20
fix(react): popover caret is not hidden when caret prop is false

### LINKED ISSUE #15155: [Bug]: Popover caret is not hidden when caret prop is false
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

1.42

### React version

17.0.2

### Description

Popover caret is not hidden when caret prop is false

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-popover--playground&args=caret:!false&globals=theme:g100

### Steps to reproduce

Open storybook page for Popover component and set `caret` control to `False`

https://react.carbondesignsystem.com/?path=/story/components-popover--playground&args=caret:!false&globals=theme:g100

### Suggested Severity

### Application/PAL

IBM Z Hardware Management Console

### Code of Conduct

harukikuri:
Closes #15155

Popover caret is not hidden when caret prop is false.

#### Changelog

**New**

- Added a new `caret` prop to the PopoverContent component to make the caret HTML element visible/invisible.
- Added a new className `${prefix}--popover-caret-close` to the PopoverContent component only when `caret` is `true`.
- Added a new SCSS code for `${prefix}--popover-caret-close` class.

**Changed**

- Inserted a new prop type for `caret` in the PopoverContent component in `PublicAPI-test.js.snap` for testing.

**Additional**

- in .all-contributorsrc file at line 1301, there is a curly brace forgotten. so added it.

#### Testing / Reviewing

(Cause)
Because `caret` prop does not get passed down to PopoverContent which contains the actual caret HTML element.
The PopoverContent component needs to know whether `caret` prop is `true` or `false` to make the caret HTML element visible/invisible. 

PopoverContent is used by Popover, DefinitionTooltip, and Tooltip components.

(Solution)
By adding `caret` prop to the PopoverContent component, the PopoverContent has control over the actual HTML element to be hidden/shown.

popover-content-type


alisonjoseph (changes requested):
Thanks for the PR, however I think this can be accomplished by simply adding more specificity to the scss, without needing to update anything else. The caret is already set to display:none by default, it just needs the specificity to only display when the parent --caret class is visible, and the code is already correctly setting that classname. 

Screenshot 2023-11-29 at 9 11 00 AM

alisonjoseph (approved):
Looks good! Thanks for the contribution 🥳

andreancardona (approved):
🔥 thank you for the contribution @kuri-sun

PR #15233, opened 2023-11-21
Added "as" prop to `Link`

### LINKED ISSUE #14715: [Feature Request]: Extend `Link` component to accept `as` prop
### The problem

Users might want to integrate with an internal routing package (react-router-dom).

[Full Slack discussion](https://ibm-cio.slack.com/archives/C2K6RFJ1G/p1695312023737749)

### The solution

Extend the base `Link` component to accept an `as` property, similar to how it is currently done by the UI Shell's `Link` component; Or, expose the UI Shell `Link` for external use.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

guidari:
Closes #14715

Added a new prop to `Link` component. Now users can manipulate the this component with a similar behaivour in the `Link` at UIShell.

#### Changelog

**New**

- prop "as" in `Link`

#### Testing / Reviewing

- Check if the `Link` works.
- "as" prop added to show how is working on the storybook in the default stories
_______
- [ ] Delete "as" prop from default stories


packages/react/src/components/Link/Link.tsx
```diff
@@ -92,7 +95,6 @@ const Link = React.forwardRef<HTMLAnchorElement, PropsWithChildren<LinkProps>>(
     });
     const rel = target === '_blank' ? 'noopener' : undefined;
     const linkProps: AnchorHTMLAttributes<HTMLAnchorElement> = {
-      className,
```
tw15egan, line 95: Do we want to remove this?

guidari: @tw15egan  I previously had move down the `className` to the component itself. But it also work as it was if I do an ternary if, because we don't want to style the component if the user is using the prop "as".
I moved it back

tw15egan: Oh that makes sense, I'm fine either way, I think we just need to set it to `undefined` rather than `''` because I was seeing this in the markup: 

Screenshot 2023-11-21 at 12 50 28 PM

guidari: Fixed it!

tw15egan (approved):
LGTM, but can we remove the `as` from the default story? We can try adding it to the playground story, but not sure if it will work

guidari:
> LGTM, but can we remove the `as` from the default story? We can try adding it to the playground story, but not sure if it will work

We can! Although looks a bit trick, since we have to refresh the story after changing the `as` prop

tw15egan:
@guidari it's fine if it doesn't work, doesn't seem like `Button` has a playground for it either.

tay1orjones (approved):
Snapshots need updated for CI to pass, otherwise looks solid 👍

sjbeatle:
This says it closed https://github.com/carbon-design-system/carbon/issues/14715, but looks like the `as` prop is expecting a `string`... even though the JSDoc for the prop says "custom element or **_component_**"

Does this update truly allow us to use the `Link` component from react-router-dom, for example? Similar to how we are able to do so for the side and top navigation components in the UI Shell?

Thanks!

guidari:
> This says it closed #14715, but looks like the `as` prop is expecting a `string`... even though the JSDoc for the prop says "custom element or **_component_**"
> 
> Does this update truly allow us to use the `Link` component from react-router-dom, for example? Similar to how we are able to do so for the side and top navigation components in the UI Shell?
> 
> Thanks!

Sorry for the delay Steven. I might have missed the notification in this PR.
Like we talked in the Carbon Developer Office Hours, we will improve the `as` prop.

Thank you!

sjbeatle:
> Sorry for the delay Steven. I might have missed the notification in this PR. Like we talked in the Carbon Developer Office Hours, we will improve the `as` prop.
> 
> Thank you!

No worries at all @guidari! Thank you!!

tay1orjones:
@sjbeatle @guidari I opened an issue for this: https://github.com/carbon-design-system/carbon/issues/15338

PR #15236, opened 2023-11-21
fix(menu): update styles

### LINKED ISSUE #14676: Menu and Menu buttons: dev/style updates
Discussed with Jan and Lauren, we will need to update:

## [Menu](https://react.carbondesignsystem.com/?path=/story/components-menu--playground) and menu items
- [x] Update `$text-primary` to `$text-secondary`, `$icon-primary` to `$icon-secondary`, 
- [x] In our guidance, the menu has `a min width of 160px`. This probably needs an update as well. Jan confirmed: 

> It’s already doing that in code for the MenuItem component. That means it’s currently effective for:
Menu
ComboButton
MenuButton
OverflowMenuV12
:grimacing:
Min: 10rem/160px
Max: 18rem/288px

---

## MenuButton, Overflow, ComboButton
- [x] Make sure there's no spacing between the Menu and Menu buttons
- Currently: MenuButton has 4px spacing, ComboButton has 4px spacing, OverflowMenu seems to have no spacing but this needs to be confirmed

---
## Need to confirm with devs

- [ ] **Scrollbar**: Consider a maximum height. But it should actually, since otherwise it could become unusable when larger than the viewport. In that case, a scrollbar would be needed. (Jan Hassel). Note: Also a Mac problem anyways. DESIGNER: Need to provide specs and give a maximum height.
- [x] **Tooltip when truncate**: Is this something that the product teams will need to turn on or customize? I don't think it's working in the Storybook currently. Dev solution: Add a title to MenuItem.

alisonjoseph:
Closes #14676

#### Changelog

**New**

- add title to `MenuItem`
- add max-height value to `Menu`

**Changed**

- update spacing between menu and menu buttons to 0px for `MenuButton` and `ComboButton`
- update tokens for Menu `$text-primary` to `$text-secondary`, and explicityly set svgs inside to `$icon-secondary`,

#### Testing / Reviewing

Check Menu, MenuButton and ComboButton that they display correctly with updated tokens and that long text that overflows shows title on hover.

Review scrollbar on MenuButton default story


thyhmdo:
@alisonjoseph this looks good! The only one thing is we still maintain the `text-primary` on hover and focus + hover states

Here is the reference https://carbondesignsystem.com/components/menu/style#interactive-state-color

alisonjoseph:
@thyhmdo just updated the hover color to the primary token. Do you know what the max-height value should be?

andreancardona:
The only weird thing I see is here - seems the dropdown content doesn't stick to the menubutton - not sure if we care to address this or not 👀 @alisonjoseph 
https://github.com/carbon-design-system/carbon/assets/32720851/4230e564-825e-4781-a171-1b7533ca247f

alisonjoseph:
@andreancardona I see that too, unrelated to anything in this PR though

andreancardona (approved):

thyhmdo (approved):
Looks great!

tay1orjones (approved):

PR #15238, opened 2023-11-21
test(Slug): add VRT, AVT tests

tw15egan:
Adds in basic tests for `Slug` and the `Form` story housing the `Slug` inside form input components

#### Changelog

**New**

- Adds in some `VRT` tests so that we can capture any regressions to `Slug` or the form inputs with `Slug` embedded in them. 
- Adds in `AVT` tests to ensure the components pass a11y requirements

**Changed**

- Fixed an a11y violation of the visible label not matching the aria-label

#### Testing / Reviewing

Ensure all AVT tests pass


tay1orjones (approved):

andreancardona (approved):
all tests are passing :)

PR #15248, opened 2023-11-24
fix(react): column span prop with percentage unit (#15223)

### LINKED ISSUE #15223: [Bug]: Column span prop with percentage unit
### Package

@carbon/react

### Browser

Chrome

### Package version

11.38.0

### React version

1.38.0

### Description

Pass a percent to the span property on the `Column`.

```
<column span="25%" />

// expected output
<div class="cds--col-span-50">

// actual output
<div class="cds--col-span-50%">
```

The span property isn't stripping the **%** suffix resulting in no styles being attached. % unit works fine when passed to breakpoint props (_sm_, _md_, etc) but doesn't work when span prop is utilized. Moreover, the type on span has to be a percent according to `ColumnSpanPercent` type

### Reproduction/example

https://stackblitz.com/edit/github-5bdwvv?file=src%2FApp.jsx

### Steps to reproduce

Create a grid with at least 1 column and pass `span="25%"` to column as a prop.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

harukikuri:
Closes #15223

<Column> component's span prop with percentage unit, `class` isn't stripping the % suffix.

```
<Column span="25%" />

// expected output
<div class="cds--col-span-50">

// actual output
<div class="cds--col-span-50%">

```

#### Changelog

**Changed**

- Added a new condition to strip the % suffix in `getClassNameForSpan`.

#### Testing / Reviewing

(Cause of the bug)
We restored the `span` prop to the <Column> component in [this PR](https://github.com/carbon-design-system/carbon/pull/13074).
But we didn't take the suffix-stripping into consideration in `getClassNameForSpan`.

(Solution)
We have to implement the same process(stripping the percentage suffix) as we did in `getClassNameForBreakpoints`.(line 365 in Cloumn.tsx)

note


alisonjoseph (approved):
LGTM! I wonder if we want to update the docs to include info around using the `span` prop? Maybe in the overview section

alisonjoseph:
@kuri-sun I think it would be fine to add to this PR if you'd like, or you can open a new issue/PR. up to you! thanks!

kuri-sun:
PR updated!
I make sure that the document(overview) for the Column component does have a new `span` and its description!
Thank you for the review!

andreancardona (approved):
looks good to me!

PR #15249, opened 2023-11-25
fix(react): downshift replacing id prop in dropdown (#15241)

### LINKED ISSUE #15241: [Bug]: Downshift replacing `id` prop in Dropdown
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.42.0

### React version

18.02

### Description

The `id` from Dropdown is not being set as expected. In the storybook example we set an array of item with `ids`, but when you check in the DOM what's being rendered it's actually the `ids` from Downshift.

It should be this `id` to be rendered: https://github.com/carbon-design-system/carbon/blob/ab4ca8a2b6fbdc557ccd090567383cdeae838570/packages/react/src/components/Dropdown/Dropdown.stories.js#L56

`<li aria-disabled="false" aria-selected="true" id="downshift-:rb:-item-1" role="option" class="cds--list-box__menu-item cds--list-box__menu-item--active">...</li>`

Reported on slack: https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1700650953038419

### Reproduction/example

### Steps to reproduce

- Inspect the option inside the `Dropdown` component.
- Check the `id`

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

harukikuri:
Closes #15241

The id from the Dropdown component is not being set as expected. In the storybook example, we set an array of items with ids, but when you check in the DOM what's being rendered it's actually the ids from Downshift.

#### Changelog

**New**

- Added the process to convert the `id` coming from the `getItemProps` function([Downshift library](https://www.downshift-js.com/downshift/)) into `id` coming from `item` prop.  only if the `item` object has an `id` property.

#### Testing / Reviewing

(Cause of the bug)
It passed the `id` prop from the Downshift library into the Dropdown component, even if a user declares the `id` property inside of each Dropdown item object.

(Solution)
If each Dropdown item object has an `id` property, we are going to use the `id` as the id of each `<li>` element, instead of using the id from the Downshift library.
But if each Dropdown item does not include an `id` property, we are just going to use the id from the Downshift library.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -516,6 +516,13 @@ const Dropdown = React.forwardRef(
                   item,
                   index,
                 });
+                if (
+                  item !== null &&
+                  typeof item === 'object' &&
+                  Object.prototype.hasOwnProperty.call(item, 'id')
+                ) {
+                  itemProps.id = item['id'];
+                }
                 const title =
                   isObject && 'text' in item && itemToElement
                     ? item.text


guidari (approved):
Awesome! Now the `id` is being rendered! 🚀

tay1orjones (approved):

PR #15250, opened 2023-11-25
refactor(Layer, useLayer): add TypeScript types

### LINKED ISSUE #13560: Add TypeScript types to Layer, useLayer

alissa9090:
Closes #13560

Adds TypeScript types to Layer component and useLayer hook.

Changelog

Added TypeScript types to Layer component and useLayer hook.


packages/react/src/components/Layer/index.tsx
```diff
@@ -26,33 +25,80 @@ export function useLayer() {
...
-      </BaseComponent>
+      {React.createElement(
+        component,
+        {
+          ref,
+          ...rest,
+          className,
+        },
+        children
+      )}
```
tay1orjones: Instead of doing this, could we instead define `as?: keyof ReactHTML;` within `LayerBaseProps`? There's quite a few usages of the `BaseComponent` pattern within .tsx if you do a "find all" in the codebase. I'd just like to keep this consistent if possible.

tay1orjones (approved):
This looks solid, just one question

packages/react/src/components/Layer/index.tsx
alissa9090: Hey, thanks for the comment! I've changed React.createElement to BaseComponent approach. Although I left PolymorphicProps for `as` prop. According to PropTypes it used to be a function or string or elementType. `keyof ReactHTML` will narrow it down to HTML tags only. Please let me know if this is the behavior you would prefer.

alissa9090:

alissa9090:
recheck

alisonjoseph (approved):

Issue #6728, opened 2020-08-26
[DataTable] Overflow of DataTable does not work properly if there are tooltips inside


jakubfaliszewski:
## What package(s) are you using?

- [x] `carbon-components`
- [x] `carbon-components-react`

## Detailed description

> Describe in detail the issue you're having.

When bottom row of DataTable has tooltip inside (`Button` with tooltip in demo) whole table becomes scrollable in Y axis and there's no possibility to see tooltip until user scrolls down a little 
> Is this issue related to a specific component?

`DataTable`>`.bx--data-table-content`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Tooltip should be visible without scrolling - like when there was no `overflow-x: auto` for `.bx--data-table-content`

> What browser are you working in?

Chrome 84

## Steps to reproduce the issue

1. Create DataTable
2. Add some element inside last row that sticks out from the row from the bottom

> Please create a reduced test case in CodeSandbox

Demo with Button with tooltip: https://codesandbox.io/s/relaxed-monad-tfc1k

![demox](https://user-images.githubusercontent.com/25402419/91309218-9d725e80-e7b0-11ea-9023-5460f2a449ad.gif)

galilmori:
Hello
We have the same problem with the datatable and tooltips in Power-iaas team.
Is there any advancements on this issue? this Bug is sometimes vey blocking for our clients who misses the information not realizing the scrollbar next to it (not to mention its esthetically very annoying for the UX designers).

Thank you for the information.

tw15egan:
@galilmori @jakubfaliszewski as a temporary workaround, you can always set `tooltipDirection="top"` on the last item in your table. Here's an example: https://codesandbox.io/s/ancient-voice-kdnzu?file=/src/App.js

galilmori:
@tw15egan unfortunately this workaround would only fix half  of the problem form me since we have icons close to the table border where it creates scrolls on two dimensions 
![image](https://user-images.githubusercontent.com/50196414/100203511-fc0bbd80-2f02-11eb-94d7-122e065f644b.png)

tw15egan:
You should also be able to use the `tooltipAlignment` prop to set the carrot to the other side

alldayalone:
I have the same issue

justindm234:
Same issue here.

sstrubberg:
Is this still an issue / have you tried using popover to resolve this?

jakubfaliszewski:
[...]

PR #15276, opened 2023-11-27
feat(DataTable): add `Slug` to `DataTable` variants 

### LINKED ISSUE #15189: Add AI enhancements to React `DataTable`

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15189

Adds in the ability to add the `Slug` to either a table row or column

#### Changelog

**New**

- `TableSlugRow` component added. Inside `TableRow`, before mapping over the row cells, a user can place the `<TableSlugRow>` component. The `TableSlugRow` component has a `slug` prop to match the other implementations.
- Styles for slug inside a table row and column
- Added a `slug` prop to `TableHeader`

**Changed**

- Modified `TableExpandRow` to ensure the `slug` is placed before the expansion arrow
- Reworked background colors to be applied on the row itself, rather than on each table cell. This was needed to ensure the linear gradient was not hidden under the table cell backgrounds. Any background color applied to a `td` should have been removed from that style block, and applied to one level higher, [example here](https://github.com/carbon-design-system/carbon/pull/15276/files#diff-e683988da838e0ccef52fcf700591850afec7ad94ca732ee5427e4f06fc4e392R537-R549)
- Changed all instances of `background` to `background-color` so that it does not override the `background-image` that provides the linear gradient.  
- Aligned all transitions to be the same speed based on property
- Fixed padding inside expandable table rows if checkbox, checkbox + slug are present
- Updated snapshots

#### Testing / Reviewing

Go to `Experimental --> Slug --> DataTable` and ensure the slug stories render as expected. 

### Notes
- Question for @aagonzales - right now, when the table is unsorted, the sort icon is not visible (only shows on hover). Should the icon always be visible if it is adjacent to the Slug - since it is right aligned, rather than next to the table column label. This is how I have it now, but let me know if this should be changed. 
- The Slug inside the Table Header is running into the same issues as in https://github.com/carbon-design-system/carbon/issues/6728 . The tooltip is cut off unless the `overflow-x` on `table-content` is removed. Hopefully refactoring Popover to use `floating-ui` will resolve these issues


tw15egan:
@tay1orjones I may break this apart and tackle the `Column` styles separately, since that may make this PR grow quite a bit. This already has quite a few changes to `DataTable`, and I want to make sure we don't have any regressions. Thoughts?
Screenshot 2023-11-28 at 4 29 32 PM

aagonzales (commented):
Can we add the gradient just in the column headers like we talked about. So it looks like this: 

image

aagonzales:
>Should the icon always be visible if it is adjacent to the Slug - since it is right aligned, rather than next to the table column label. This is how I have it now, but let me know if this should be changed.

@tw15egan No, I don't think the sort icon should be shown even if it is unsorted, it is visually confusing and actually makes it look like it is sorted. It feels weird to show sort with AI columns but not with the non-ai ones. 

Is that going to cause issues with the slug placements?

tw15egan:
@aagonzales would the slug just stay next to the label when the sort icon is active?

aagonzales:
@tw15egan Yeah I was just thinking that too, something like this instead for sort. Would the unsorted ones still have to have it beside the label, like shown?

image

tw15egan:
Yeah, that way, it also aligns with the unsortable Table headers as well. I can make that change

aagonzales (approved):
Ok, everything looks good to me now! 👍 ✅

guidari (approved):
LGTM! 🚀

tay1orjones (approved):

andreancardona (approved):
💯

PR #15312, opened 2023-12-04
fix(column): remove percentage character only if present

### LINKED ISSUE #15311: [Bug]: Column span is not working as expected when passed as string instead of number
### Package

@carbon/react

### Browser

_No response_

### Package version

1.44.0

### React version

18.2.0

### Description

When one passes strings for the breakpoint props on `Column`, the layout will render incorrectly.

### Reproduction/example

https://stackblitz.com/edit/github-5xgfbp?file=src%2FApp.jsx

### Steps to reproduce

```jsx
<Grid>
  <Column sm="4" md="8" lg="16">
    {/* ... */}
  </Column>
</Grid>
```

The example will always span 1 column regardless of the breakpoint.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

janhassel:
Closes #15311

See issue for details. Currently, if the user passes `md="8"` it will generate the following css classname: `cds--md:col-span-`. Passing in `lg="16"` will result to `cds--lg:col-span-1`.

#### Changelog

**Changed**

- Instead of always removing the last character, only remove a percentage sign if one exists

#### Testing / Reviewing

- See linked issue's reproduction


tay1orjones (approved):
Thanks for the fix!

I'm personally still on the fence about these percentage classes. 😅 We only kind of "soft-support" them - I think there's only classes for 25%, 50%, 75%, 100%? It was a new idea added in the initial CSSGrid implementation that IIRC we didn't have previously with FlexGrid. Anyway, I'm glad someone is using them, and appreciate you submitting this fix!

janhassel:
@tay1orjones I am actually not using the percentage classes in any designs and did not know they exist. 😅 I came across this when I was prototyping without TypeScript for a change and did not remember that the type should be `number` instead of `string`.

guidari (approved):

PR #15314, opened 2023-12-04
style(Notification): fix issue with alignment when inline

tw15egan:
Reported on Slack: 

Screenshot 2023-12-04 at 10 16 44 AM

#### Changelog

**Changed**

- A [div](https://github.com/carbon-design-system/carbon/pull/14877/files#diff-463089737f576a8b2bc258f3180fd32a1f7cf211fe6e3e3c1aca36d41cc998f5R1039) was added in #14877 which caused some issues when the `ActionableNotification` was in `inline` mode. I added the flex property to its new container so it behaves as expected. 

#### Testing / Reviewing

Go to `Notifications --> ActionableNotification --> Playground` and turn on the `inline` option. Ensure the notification close button does not wrap to a new line

### DIFF
--- a/packages/react/src/components/Notification/Notification.tsx
+++ b/packages/react/src/components/Notification/Notification.tsx
@@ -1036,7 +1036,9 @@ export function ActionableNotification({
           </div>
         </div>
       </div>
-      <div ref={innerModal}>
+      <div
+        className={`${prefix}--actionable-notification__button-wrapper`}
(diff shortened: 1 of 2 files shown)


tay1orjones (approved):
Thanks for such a quick fix! 🏎️💨

andreancardona (approved):
looks good!

PR #14989, opened 2023-10-21
Add TypeScript types to Switcher, SwitcherDivider, SwitcherItem #13584

shreeup:
Closes #13584 

Add TypeScript types to Switcher, SwitcherDivider, SwitcherItem

#### Changelog

**New**

- Added 3 new typescript equivalent of Switcher, SwitcherItem, SwitcherDivider components by executing git mv command and making required changes

**Changed**

- Updated snapshot to fix test

**Removed**

- Deleted old components written in .js

#### Testing / Reviewing

executed lint, run test suite and ensured no new failures after the commit, Added these component in UIShell storybook and verified.

PR #15333, opened 2023-12-06
fix(#15332) use toArray or checks for mapping children

### LINKED ISSUE #15332: [Bug]: Switcher Throws with NULL children
### Package

@carbon/react

### Browser

_No response_

### Package version

1.44.0

### React version

18.2.0

### Description

A recent change caused `Switcher` component to use throw when cloning children due to the change from `toArray` to just `map`.

https://github.com/carbon-design-system/carbon/pull/14989/files#diff-8f807f021a44c44f041949f1839b14e4428b5f2e1f9a2b5946b554a85db0fd13R112

As you can see here this line is just `.map` which means `null` children will get passed in.  If you use `toArray` and then `map` you will skip null children. Note the caveats section here https://react.dev/reference/react/Children#children-toarray

I am putting SEV 2, although this is a BREAKING Change that could get into production apps. For us our unit tests caught it but other teams may not be as lucky if they are using any Switcher that is programmatically controlled.  Not sure if any other items have had this change...

### Reproduction/example

https://github.com/carbon-design-system/carbon/pull/14989/files#diff-8f807f021a44c44f041949f1839b14e4428b5f2e1f9a2b5946b554a85db0fd13R112

### Steps to reproduce

Open Switcher will null child.
Screenshot 2023-12-05 at 19 47 19

### Suggested Severity

### Application/PAL

Aspera

### Code of Conduct

dabrad26:
Closes #15332 

Using toArray to get rid of null or other non element children.  See ticket for more info on how this was changed recently and the benefit of toArray over direct map.

### DIFF
--- a/packages/react/src/components/ProgressIndicator/ProgressIndicator.js
+++ b/packages/react/src/components/ProgressIndicator/ProgressIndicator.js
@@ -56,6 +56,10 @@ function ProgressIndicator({
   return (
     <ul className={className} {...rest}>
       {React.Children.map(children, (child, index) => {
+        if (!React.isValidElement(child)) {
+          return null;
+        }
+
         // only setup click handlers if onChange event is passed
         const onClick = onChange ? () => onChange(index) : undefined;
         if (index === currentIndex) {
--- a/packages/react/src/components/ContentSwitcher/ContentSwitcher.tsx
+++ b/packages/react/src/components/ContentSwitcher/ContentSwitcher.tsx
@@ -228,12 +228,12 @@ export default class ContentSwitcher extends React.Component<
         role="tablist"
         onChange={undefined}>
         {children &&
-          React.Children.toArray(children).map((child: ReactElement, index) =>
-            React.cloneElement(child, {
+          React.Children.toArray(children).map((child, index) =>
+            React.cloneElement(child as ReactElement, {
               index,
               onClick: composeEventHandlers([
                 this.handleChildChange,
-                child.props.onClick,
+                (child as ReactElement).props.onClick,
               ]),
               onKeyDown: this.handleChildChange,
               selected: index === this.state.selectedIndex,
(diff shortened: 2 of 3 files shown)


dabrad26:
I went through and quickly checked all uses of Children.map (https://github.com/search?q=repo%3Acarbon-design-system%2Fcarbon%20React.Children.map&type=code)

Quickly looking over each most had checks to validate that children were either Nodes or they had no dangerous code (calling on NULL or cloning).  I did find two possible risks with ContentSwitcher and ProgressIndicator and have added them to this PR.

packages/react/src/components/ProgressIndicator/ProgressIndicator.js
```diff
@@ -56,6 +56,10 @@ function ProgressIndicator({
   return (
     <ul className={className} {...rest}>
       {React.Children.map(children, (child, index) => {
+        if (!React.isValidElement(child)) {
```
dabrad26, line 59: This could be written in a few different ways; but I wanted to make as minimal impact; since bottom of this loop returns null I am just adding this one check for valid Element here instead of on each of the index checks below.

guidari (approved):
LGTM! 🚀

tay1orjones (approved):
Thanks so much for putting up this fix! Sorry we didn't catch this in the initial PR https://github.com/carbon-design-system/carbon/pull/14989 😓  

I personally find it super difficult to review file changes from .js to .tsx when the git diff splits the file into two separate changes like it did in that PR. If it would've been a unified diff we might've spotted this.

Sorry again for any inconvenience you had from this. Can we cut a new release for this once it's in @guidari @alisonjoseph?

dabrad26:
Thanks so much everyone for the quick release. Yeah. Typescript migrations are tough. We have done several over the years. So happy to help if you ever want to assign a typescript migration PR to me. :)

tay1orjones:
> happy to help if you ever want to assign a typescript migration PR to me. :)

@dabrad26 That would be such a great help, thanks for offering!

PR #15336, opened 2023-12-06
feat(DataTable): expand button can be on the right side

### LINKED ISSUE #10993: [Feature Request]: Putting expand button on the right hand side of the row
### Summary

Right now the expansion panel toggle button shows on the left when the expandable is true.

Screenshot from 2022-03-12 10-38-08

An option can be added to toggle the position possibly something like expansionButtonPosition: 'left' | 'right'

### Justification

It may be easier to reach the button on the right hand side. The reason I opened this one is I got a mockup design that has the expansion on the right though.

Expand row button works the same but is the last column instead of the first.

### Required functionality

Expand row button works the same but is the last column instead of the first.

_No response_

### Available extra resources

_No response_

### Code of Conduct

berndfuhrmann:
Closes #10993

DataTable expand button configurable to be on the right side.

#### Changelog

**New**

- CSS style
- React component setting added

**Changed**

- nothing

**Removed**

- nothing

#### Testing / Reviewing
Check PlayBook for DataTable and expandable rows (added one example to demo functionality and show appearance)


alisonjoseph (changes requested):
Hi, thanks for opening this PR. I'd like our designers to take a look at this to be sure this is something we'd like to add to the system, the linked issue isn't clear. FYI: @carbon-design-system/design 

Looks like the alignment is off here, I think if the chevron is on the right we wouldn't want that extra padding, and we'd need to make sure the padding was added to the other side.

Screenshot 2023-12-07 at 10 03 37 AM


If we do want to add this option I'd suggest we change the prop name to `expandButtonAlign` or `expandAlign` and the values to `start` and `end` 

We also shouldn't need an entire story just for this prop, we'd just want to make sure it shows up in the playground story.

Thanks!

tay1orjones:
@berndfuhrmann you mentioned in the original issue that this would be useful for your project. Could you share mockups or concrete examples of why/when having this at the end of the horizontal dimension in horizontal writing modes would be used rather than the existing default (at the start)?

If we accept this change we're going to need to document it in the design specification for this component on the website and we'll need to specify rationale for when and why to use this feature. Especially since it makes it much more difficult to discover the row has the ability to expand since it's at the end of the row as you scan it.

tay1orjones:
It doesn't look like the prop is available in any of the [playground story](https://deploy-preview-15336--v11-carbon-react.netlify.app/?path=/story/components-datatable-expansion--playground) controls.

tay1orjones:
Thanks again for opening this PR and working on this issue. 

Based on https://github.com/carbon-design-system/carbon/issues/10993 this feature isn't something the team believes should be included in this component. 

My recommendation for those still looking to use this alignment would be to either flip the document direction with `LayoutDirection` or `TextDirection`, or implement your own custom TableExpandRow component in your application that houses the changes outlined in this PR.

PR #15350, opened 2023-12-07
feat(Modal): add slug to Modal, ComposedModal

### LINKED ISSUE #15192: Add AI enhancements to React `Modal`
> [!WARNING]  
> **Blocked by**
>
> - #15184

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15192

Adds in the `slug` prop to `Modal` and `ComposedModal`.

#### Changelog

**New**

- `slug` prop added to `Modal` and `ComposedModal`
- `slugOverlay` tokens added to provide a lighter overlay in light themes
- `Modal` and `ComposedModal` added to the Examples folder inside `Slug` in a storybook. One shows with full-width button (gradient should start at `64px`), and the other has 3 buttons (gradient should start at `0px`)

**Changed**

- Added an offset to the `ai-gradient` mixin so that the gradient can start at a different value than 0 when buttons are full width inside the modal
- Overlay changed to use new tokens
- Box shadow added to both `Modal` variants when `slug` is provided 
- Fixed an issue that was causing the slug callout text not to use the custom story styles

**Removed**

- `overflow: hidden` on `.cds--modal-container` was causing the `slug` to be cut off; Contents inside still seem to scroll properly 

#### Questions
- This is currently using the `ai-gradient`, but should it be using the `callout-gradient` that is used in `Tile` and the `Slug` callout?
- I pulled the `drop-shadow` values directly from `Figma` and mapped them to  `box-shadow`, are these correct? 

#### Testing / Reviewing

Go to `unstable__slug --> Examples --> Modal` as well as `ComposedModal` and ensure everything renders as expected


aagonzales (commented):
Yeah that's looking correct except there is a visual difference between the composed and default modal. I think the default modal still has transparency on it somewhere which is why its a little darker. 

image

aagonzales (approved):
Ignoring the problems in the gray 10 theme that will hopefully be resolved with v2 visuals, this is good to go.

tay1orjones (approved):
LGTM 👍

andreancardona (approved):
just tested, looks good!

PR #15359, opened 2023-12-09
Refactor: add ModalWrapper types

### LINKED ISSUE #13564: Add TypeScript types to ModalWrapper

mattborghi:
Closes #13564

Add types for `ModalWrapper`.

#### Changelog

**New**

- Add types for `ModalWrapper`.

**Changed**

- File extensions from `.js` to `.tsx`/`.ts`.

**Removed**

- N/A

#### Testing / Reviewing

* Storybook runs as before.
* Unit tests passing.


packages/react/src/components/ModalWrapper/ModalWrapper.tsx
```diff
@@ -6,16 +6,50 @@
...
+  preventCloseOnClickOutside?: boolean;
+  primaryButtonText?: string;
+  renderTriggerButtonIcon: React.ElementType;
+  secondaryButtonText?: string;
+  selectorPrimaryFocus?: string;
+  shouldCloseAfterSubmit?: boolean;
+  status?: string;
+  triggerButtonIconDescription?: string;
+  triggerButtonKind: (typeof ButtonKinds)[number];
+  withHeader?: boolean;
```
tay1orjones, line 40: Would you mind copying the comments from the proptypes up into the interface here so that things like intellisense provide the additional usage/info/context?

tay1orjones (approved):
One minor thing to change but otherwise looks good I think. Thanks for making this contribution!

packages/react/src/components/ModalWrapper/ModalWrapper.tsx
mattborghi: The proptype doesn't have comments that's why I didn't add them here.

guidari (approved):
LGTM! 🚀

packages/react/src/components/ModalWrapper/ModalWrapper.tsx
guidari: True! It doesn't have prop comments. Might be because the file is kinda old, 7 years. 🤔

Anyways, in this case we can merge without it and leave as it is.

PR #7405, opened 2020-12-04
chore(project): update yarn to berry

joshblack:
This is take two of updating our project to use the latest versions of Yarn. 

#### Changelog

**New**

- Yarn has a zero-install folder under cache

**Changed**

- Update yarn to berry (2.3)
- Update packages to work with new `.bin` rules from Yarn berry

**Removed**

- Offline mirror has been replaced by cache

#### Testing / Reviewing

- Verify project build tools and deploys work as expected

PR #15369, opened 2023-12-11
chore: remove node-gyp

tay1orjones:
This removes node-gyp as a dependency in the root workspace.

It appears there used to be and still are quite a few packages that implicitly rely on node-gyp without actually listing it as a dependency. I assume when this was added in https://github.com/carbon-design-system/carbon/pull/7405 this was the case. 

* https://github.com/yarnpkg/berry/issues/1016
* https://github.com/yarnpkg/berry/issues/5974
* https://github.com/yarnpkg/berry/issues/1252
* https://github.com/yarnpkg/berry/issues/5804

I can't reproduce any issues after removing this dependency. I ran the following without issues:

```
yarn && yarn build && yarn format && yarn lint && yarn test
```

#### Changelog

**Removed**

- remove node-gyp as a devDependency

#### Testing / Reviewing

- If ci begins to fail, local builds stop working, or there are errors in the release, we can revert this change

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "jest-junit": "^16.0.0",
     "lerna": "^8.0.0",
     "lint-staged": "^15.0.0",
-    "node-gyp": "^7.1.2",
     "prettier": "^2.8.8",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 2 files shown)


alisonjoseph (approved):
Seems fine to me 👍

andreancardona (approved):
nothing is failing on my end, LGTM

PR #15371, opened 2023-12-11
chore: update release docs and deploy-packages workflow

alisonjoseph:
This PR fixes the problem where the gatsby theme wasn't getting updated with latest pacakges as the workflow was triggered before the release was tagged as latest. 

#### Changelog

**Changed**

- deploy-packages workflow is now triggered on complete of promotion workflow
- All website and theme related updates are moved to AFTER a release to promoted to latest

### DIFF
--- a/.github/workflows/deploy-packages.yml
+++ b/.github/workflows/deploy-packages.yml
@@ -7,7 +7,7 @@ concurrency:
 on:
   workflow_dispatch:
   workflow_run:
-    workflows: [Release]
+    workflows: [Promote]
     types:
       - completed
 
--- a/docs/release.md
+++ b/docs/release.md
@@ -334,7 +334,7 @@ cases, follow these steps below to ensure a proper patch release:
 - [ ] for each package (replace <kbd>carbon-components-react</kbd> with the
       package name):
   ```bash
-  npm dist tag add carbon-components-react@vX.Y.Z latest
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy-packages.yml
```diff
@@ -7,7 +7,7 @@ concurrency:
 on:
   workflow_dispatch:
   workflow_run:
-    workflows: [Release]
+    workflows: [Promote]
```
tay1orjones, line 10: Ah yeah because it needs the packages to be on latest ... this makes total sense. Thanks for fixing this!

docs/release.md
```diff
@@ -334,7 +334,7 @@ cases, follow these steps below to ensure a proper patch release:
 - [ ] for each package (replace <kbd>carbon-components-react</kbd> with the
       package name):
   ```bash
-  npm dist tag add carbon-components-react@vX.Y.Z latest
+  npm dist-tag add carbon-components-react@vX.Y.Z latest
```
tay1orjones, line 337: I added this small change, hope that's okay. Something that's been bothering me for quite a while but have kept forgetting to put into a PR

tay1orjones (approved):
Thanks for updating this to keep current!

guidari (approved):

Issue #13550, opened 2023-04-17
Add TypeScript types to ContentSwitcher, Switch


tay1orjones:

anton-tsymuk-viacomcbs:
Hi. If you do not mind, I would try to help with the migration. I opened a PR for this issue

wkeese:
Hmm, this isn't working for me, I get

> TS2305: Module '"@carbon/react"' has no exported member 'ContentSwitcher'.
> 12 import { ContentSwitcher } from "@carbon/react";

ISTM that there are a couple problems:

* There's no generated components/ContentSwitcher/index.d.ts file at all.   Because components/ContentSwitcher/index.js needs to be renamed to index.js.
* Even after that though, createClassWrapper() is a black-hole to the type information.  It needs to be converted to something like below but I haven't quite gotten it working yet

```ts
export function createClassWrapper<P>(Component: ComponentClass<P>): FunctionComponent<P> {
  function ClassWrapper(props) {
    return <Component {...props} />;
  }

  const name = Component.displayName || Component.name;
  ClassWrapper.displayName = `ClassWrapper(${name})`;

  return ClassWrapper as FunctionComponent<P>;
}
```

PR #15382, opened 2023-12-12
fix: declarations of ContentSwitcher, Switch, Slider, SkeletonIcon

wkeese:
Addresses errors like:

> TS2305: Module '"@carbon/react"' has no exported member 'ContentSwitcher'.
>
> 12 import { ContentSwitcher } from "@carbon/react";

This contains two types of fixes:

1. Whenever the underlying component is written in Typescript, the index file needs to be named index.ts.  If it's named index.js, the build won't even generate an index.d.ts file.  (The index file can also be named index.tsx, but that doesn't make sense unless it contains React markup.)

2. I updated `createClassWrapper()` so that Typescript knows the type of the returned functional component:

```ts
export function createClassWrapper<Props>(
  Component: ComponentClass<Props>
): FunctionComponent<Props> {
  ...
}
```

Note that the generated index.d.ts are a bit weird, for example

```ts
/// <reference types="react" />
declare const Slider: import("react").FunctionComponent<import("./Slider").SliderProps>;
```

If I add casting to the index.ts files themselves, ex:

```ts
const Slider = createClassWrapper(SliderComponent) as FC<SliderProps>;
```

That generates more canonical index.d.ts files:

```ts
import { FC } from "react";
import { SliderProps } from './Slider';
declare const Slider: FC<SliderProps>;
```

(So I can do that if you like.)

Of course, the `createClassWrapper()` calls should be temporary, as all remaining class components should be converted to functional components.

Refs #13550, #12557, #13574.


tay1orjones (approved):
Thanks for this! I think the existing way is fine, canonical looks nicer but they're generated files and nobody's going to be actually reading or maintaining them anyway. I don't think there's any functional difference between the two?

wkeese:
Yeah I agree. I haven't finished converting our project to Carbon 11 yet but after this change, I don't get the error about
ContentSwitcher anymore.

alisonjoseph (approved):

PR #15385, opened 2023-12-13
fix(progressindicator): correct padding on secondaryLabel in firefox

### LINKED ISSUE #15383: [Bug]: ProgressIndicator - Firefox - secondaryLabel is vertically aligned in the middle of the available space
### Package

@carbon/react

### Browser

Firefox

### Package version

1.45.0

### React version

18.2.0

### Description

In Firefox 20.0.1 (64-bit) the ProgressIndicator component exposes a behaviour wrt/ to its layout where the `secondaryLabel` has its top and bottom `margin` set to `auto` and thus in Firefox the `secondaryLabel` is vertically aligned in the middle of the available space where in Chrome it is aligned directly underneath the label.

This may make users think that the way how the `secondaryLabel` is rendered looks of cause it may look like as if it is not really belonging to the primary `label` of a `ProgressStep`. In any case the behaviour differs between Chrome and FF.

image

image

### Reproduction/example

none

### Steps to reproduce

See screenshots.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #15383 

#### Changelog

**Changed**

- specify top margin of 0 on secondary labels

#### Testing / Reviewing

- For some reason, chrome was resolving the top margin of `auto` to `0px`, potentially because it's `position: static`? I'm not totally sure. So in effect, Firefox was rendering correctly based on the styles written and chrome was wrong. but what we _want_ is how chrome was rendering it. So this should be correct now. 
- View the ProgressIndicator Playground story in Firefox and Chrome. Select the vertical toggle. Observe there is no top margin in either browser.

### DIFF
--- a/packages/styles/scss/components/progress-indicator/_progress-indicator.scss
+++ b/packages/styles/scss/components/progress-indicator/_progress-indicator.scss
@@ -329,7 +329,7 @@ $progress-indicator-bar-width: 1px inset transparent !default;
 
   .#{$prefix}--progress--vertical .#{$prefix}--progress-optional {
     position: static;
-    margin: auto 0;
+    margin: 0 0 auto;
     inline-size: 100%;
   }
 


alisonjoseph (approved):
LGTM

andreancardona (approved):
😎

tw15egan (approved):

PR #15390, opened 2023-12-13
Chore: Onboard @carbon/react to IBM Telemetry 🚀

francinelucca:
- Uninstall @carbon/telemetry from @carbon/react
- Add `telemetry.yml` config file
- Update postinstall script to call new telemetry script, remove legacy script call
- Add IBM Telemetry collection disclosure to README


tay1orjones (approved):
🎉 🙌 🚀

tay1orjones:
I'd like to get this same change made to the v10 branch. We can do it, or you, just let me know.

packages/react/telemetry.yml
```diff
@@ -0,0 +1,815 @@
+# yaml-language-server: $schema=https://unpkg.com/@ibm/telemetry-config-schema@v0/dist/config.schema.json
```
mattrosno: This config was generated with a script by @francinelucca to get all prop and prop values that we want to collect. We're still finishing the script to use when instrumenting other packages. Once that script is done, _maybe_ we could add that to a CI/CD script to automate this config's upkeep as components, component props, and supported component prop values change.

mattrosno:
> I'd like to get this same change made to the v10 branch. We can do it, or you, just let me know.

@tay1orjones let's instrument v11 first, then we can look at the v10 branch. We'll need to generate the allowed telemetry prop config for the v10 branch.

alisonjoseph (approved):

PR #15391, opened 2023-12-13
Chore: Onboard @carbon/react v10 to IBM Telemetry 🚀

francinelucca:
- Uninstall @carbon/telemetry from @carbon/react
- Add `telemetry.yml` config file
- Update postinstall script to call new telemetry script, remove legacy script call
- Add IBM Telemetry collection disclosure to README


packages/react/package.json
```diff
@@ -14,7 +14,8 @@
   "files": [
     "lib/**/*",
     "es/**/*",
-    "umd/**/*"
+    "umd/**/*",
+    "telemetry.yml"
```
francinelucca, line 18: @carbbon-devs we added this to the config because we think we need it for the config file to ship with the deployed package, but please confirm that 😬

tay1orjones: Yep I think that should do it

tay1orjones (approved):

alisonjoseph (approved):

Issue #14655, opened 2023-09-14
[Bug]: TableToolbarSearch is calling onChangeProp with incorrect parameters when there is a default value


laramlewis:
### Package

carbon-components-react

### Package version

v7.59.8

### React version

v16.14.0

### Description

In TableToolbarSearch.js, it is calling onChangeProp with two strings instead of an event as expected when a default value is specified. 

```
  useEffect(
    () => {
      if (defaultValue) {
        onChangeProp?.('', defaultValue);
      }
    },
    //eslint-disable-next-line react-hooks/exhaustive-deps
    []
  );
```

As a result the onChange handler is passed '' instead of an event with a target value.

### Reproduction/example

N/A

### Steps to reproduce

- Render a TableToolbarSearch component with a defaultValue and an onChange handler. 
- The onChange handler will be called with an emptyString instead of an event.

### Suggested Severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Application/PAL

IBM Knowledge Catalog (formally Watson Knowledge Catalog)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
Thanks for reporting this! Here's the relevant source you're referring to: 

https://github.com/carbon-design-system/carbon/blob/0d75092bfc865937cdb77b78e8e4d978cfcd33f3/packages/react/src/components/DataTable/TableToolbarSearch.tsx#L179-L187

Since the dependency array is empty, this is only ran once, after the initial render. What event would you expect `onChange` to be called with in this case? I agree in principle that the format of what's passed through the callback should ideally be the same as the event populated from the actual input's `onChange`, but I'm not sure what we'd change it to be. What are your thoughts?

laramlewis:
@tay1orjones - Thank you for your quick response. I discussed this with my colleague and we have the following thoughts:

- Creating a custom event is doable: https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events

- The problem with creating a ‘fake’ event is that users can’t tell if it’s coming from the ‘mount’ or actual user changes.
unless you also give that event an attribute to identify it’s different from the change event.

[...]

PR #14471, opened 2023-08-20
refactor(typescript): type annotations on TableToolbarSearch

jpsorensen:
Closes #12536 

Add Typescript annotations to TableToolbarSearch.

#### Changelog

**Changed**

- TableToolbarSearch.js converted to tsx

#### Testing / Reviewing

- In the packages/react directory, run yarn build.
- Find usages of TableToolbarSearch in its unit tests. Add them to the bottom of TableToolbarSearch.tsx and confirm that they compile.

PR #5089, opened 2020-01-17
fix(TableToolbarSearch): allow default values to be set

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/3170

The `DefaultValue` prop was being ignored. This checks for the presence of a `DefaultValue`, and if it exists, it is used as the default state of `value`. On the initial load, it sets the expanded state to `true` and runs the `onChangeProp` to filter the table. Modified the `handleOnInputValueChange` in `DataTable.js` to account for non-event based sorting (table loads with a default value). 

Screen Shot 2020-01-17 at 2 10 40 PM

#### Changelog

**New**

- `DefaultValue` prop works as intended

**Changed**

- `handleOnInputValueChange` now has a `defaultValue` parameter

**Removed**

- `DefaultValue` being passed directly down to `Search`. If `DefaultValue` exists, it is immediately set to `value` which is then passed down. 

#### Testing / Reviewing

- Open `src/components/DataTable/stories/with-toolbar.js`
- Add `defaultValue="1"` to `TableToolbarSearch`
- Open the `With Toolbar` Storybook example, verify only 1 result shows up.
- Remove the value, verify all results show up.
- Add new value and verify the correct results show up.

PR #15395, opened 2023-12-15
fix: TableToolbarSearch type for onChange

wkeese:
Refs #14471, #14655.

This is the minimal fix for TableToolbarSearch's `onChange` type.
In #14471, there was obviously some confusion between functions and events themselves.

However, the current way that `onChange()` is called is very strange, and I'd like to change it.
There are two ways in which onChange() is called:

1. When the user makes a change, `onChange()` is called like

```js
onChange(evt);
```

2. If the app creates the `<TableToolbarSearch>` with a default value, then `onChange()` is immediately called like:

```js
onChange('', defaultValue);
```

This is from #5089, from @tw15egan's change.  The especially strange thing is why it's passing an empty string instead of `null`.

Because of the empty string, the call signature needs to be:

```js
(
    event: '' | ChangeEvent<HTMLInputElement>,
    value?: string
  ) => void;
```

And also, client code needs to do a lot of type coercion, because even if `event` is defined, it might be a string.

But if we changed the code to pass null or undefined (rather than an empty string), then the call signature simplifies to:

```js
(
    event: ChangeEvent<HTMLInputElement> | null,
    value?: string
  ) => void;
```

Well, maybe not much simpler, but at least more canonical, and I think it removes the need for type coercion in the client.  You can just say `evt?.target`.

Thoughts?

See also #14655, which suggests passing a custom event or (probably better) to stop calling onChange() except when the value actually changed.  But that update would need to wait until Carbon 12 (?), whereas I think the fix in this PR can be made now.

#### Changelog

**Changed**

- Fixed onChange() signature to match code.


guidari (approved):
LGTM! 🚀

wkeese:
@tw15egan - What do you think about changing the code to call `onChange(null, defaultValue)` instead of `onChange('', defaultValue)`?

andreancardona (approved):
thank you for the contribution!

PR #15397, opened 2023-12-15
feat(treeview): add experimental controllable API

janhassel:
Ref #14451 / Point 1

This PR introduces an experimental controllable API for the TreeView to address point 1 mentioned in above linked discussion.

A user must opt-in via the feature flag `enable-treeview-controllable`.

#### Changelog

**New**

- feature flag `enable-treeview-controllable`
- TreeView `props.onActivate` to pass-through node's `onToggle` events
- TreeNode `props.defaultIsExpanded` for uncontrolled mode

**Changed**

- With above mentioned feature flag, the states `selected` and `active` can be controlled or uncontrolled based on the user's choice

#### Testing / Reviewing

- Storybook
- Test suites

--- 

@tay1orjones I wanted to implement the feature flag approach you shared a while back for this change. Is there any naming guidance for these yet? I avoided `v12` directly and essentially followed this structure:
```
enable-[component]-[feature]
```
Let me know what you think!


tay1orjones (approved):
Hey this looks really solid! Sorry it took so long for a review 😓 

I think the feature flag name is good - so far the loose (undocumented) thinking I have on it is `v12-*` flags are committed to be turned _on by default_ in the next major. We could always rename this flag if/when we think that's the case, but in the meantime I think naming it as is feels appropriate.

alisonjoseph (approved):

alisonjoseph:
This LGTM! @emyarod do you have a chance to give this a review before we merge?

PR #15407, opened 2023-12-17
Add Typescript types to PaginationNav

### LINKED ISSUE #13567: Add TypeScript types to PaginationNav

mattborghi:
Closes #13567

Add Typescript types to PaginationNav

#### Changelog

**New**

- Add Typescript types to PaginationNav

**Changed**

- Renamed files to `ts` and `tsx`.

**Removed**

-  N/A.

#### Testing / Reviewing

- Storybook runs as before.
- Unit tests passing.


packages/react/src/components/PaginationNav/PaginationNav.tsx
```diff
@@ -0,0 +1,635 @@
...
+  const ref = useRef<number | null>(null);
+
+  useEffect(() => {
+    ref.current = value;
+  });
+
+  return ref.current;
+}
+
+function getCuts(
```
andreancardona: @guidari I'm trying to think if there is a better name for this function. Thoughts??

packages/react/src/components/PaginationNav/PaginationNav.tsx
```diff
@@ -0,0 +1,635 @@
...
+  splitPoint: number | null = null
+) {
+  if (itemsThatFit >= totalItems) {
+    return {
+      front: 0,
+      back: 0,
+    };
+  }
+
+  const split = splitPoint || Math.ceil(itemsThatFit / 2) - 1;
```
andreancardona:
```suggestion
  const split = splitPoint || Math.ceil(itemsThatFit / 2) - 1;
```
Same question for `items that fit` 👀

packages/react/src/components/PaginationNav/PaginationNav.tsx
guidari: Maybe `calculateCuts()` would be nice, because it seems that is doing a few calculations there. 
I used as insparation the Menu.tsx that has the `calculatePosition()`

andreancardona: yesss love it!

andreancardona: @mattborghi can we update the function name to `calculateCuts` - thank you!

packages/react/src/components/PaginationNav/PaginationNav.tsx
andreancardona: @mattborghi can we update `itemsThatFit` to `itemsDisplayedOnPage` something like this - I'm open to suggestions

mattborghi: Done!

packages/react/src/components/PaginationNav/PaginationNav.tsx
mattborghi: Done!

tw15egan (approved):

tay1orjones (approved):
Sorry for the delay, thanks for this!

PR #15412, opened 2023-12-18
chore: use type-only keyword

dbrugger:
Use the type-only keyword for imports that are only used for typing.

#### Changelog

**Changed**

- Add `type` to some type-only imports

#### Testing / Reviewing

If the build is fine it's fine.


guidari (approved):
LGTM! 🚀

andreancardona (approved):

PR #15415, opened 2023-12-18
fix(border-subtle): bump value in dark themes

### LINKED ISSUE #14597: [Bug]: $border-subtle is using previous step's value
### Package

@carbon/styles

### Browser

_No response_

### Package version

1.37.0

### React version

_No response_

### Description

It seems like the $border-subtle tokens are always one step behind compared to the rest of the layering tokens.

image

This means that a UI element with a $layer-02 background will be paired with a $border-subtle-01 border. In the dark themes, these two values are the same though, resulting in the border being invisible.

image
image

### Reproduction/example

https://stackblitz.com/edit/github-recj8l?file=src%2FApp.jsx

### Steps to reproduce

see above

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14597

@janhassel pointed out that our dark theme `$border-subtle` values were lagging behind the corresponding `$layer` values.  

#### Changelog

**Changed**

- Bumped up the values of `border-subtle-01` and `border-subtle-02` in `g90` and `g100` so that they are visible in `Layer Two` and `Layer Three`. It seems a (possible) cause of this is that the `Theme` component assumes (and sets) a `layer-one` class by default, and not the background. We possibly need to look into setting a `background` class by default instead and creating new `-00` layer tokens. See [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Theme/index.tsx#L80) and [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Theme/index.tsx#L91C18-L91C18). We also emit layer-01 at root, instead of a possible layer-00, but this is the easiest fix and doesn't open a can of worms. 

#### Testing / Reviewing

Go to `Layer --> Test Story` to see value changes. Check other components in `g90` and `g100` to ensure the subtle borders look correct


tw15egan:
@guidari and I bumped up values for the `border-subtle-01` and `border-subtle-02` tokens so that they were visible (see [here](https://stackblitz.com/edit/github-recj8l?file=src%2FApp.jsx) for example). These changes would only affect a component in `g90` or `g100` on `Layer Two` or `Layer Three`. Let me know if these visual changes are acceptable -- You can view any `With Layer` story to see/inspect the changes or view the Percy link

alisonjoseph (approved):
LGTM! 👍

laurenmrice (approved):

tay1orjones (approved):
Just need to remove the test/demo stories before we can merge.

PR #15421, opened 2023-12-18
fix: modified _number-input.scss:91

### LINKED ISSUE #15414: [Bug]: NumberInput with hideSteppers set as true adds unnecessary padding when in invalid state
### Package

@carbon/react

### Browser

Chrome

### Package version

1.45.0

### React version

16.13.1

### Description

When using a `NumberInput` with `hideSteppers` set as `true` and this input becomes invalid, it adds a padding that hides the input value if the width of the input is not large enough.

image

### Reproduction/example

https://stackblitz.com/edit/github-cldk9i-yzpm36?file=src%2FApp.jsx

### Steps to reproduce

Create a NumberInput
Set its width to a small value, like 100px
Set it as invalid
Input value should be visible but end up being hidden because of the padding

### Suggested Severity

### Application/PAL

IBM RPA

### Code of Conduct

dev-ghostly:
Closes https://github.com/carbon-design-system/carbon/issues/15414

Fixes an issue with padding when `hideSteppers` is set to true

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Full report : 

The input was like this before
image_2023-12-18_183825308

Right now it will be like this :
image_2023-12-18_183903376

Changes did on styles/scss/components/number-input/_number-input.scss:91

Please note that it seems that the bug appeared only with invalid property and container under 150px and that the natural min-width of the NumberInput is 150px, which mean that you will have bug under that container size.

Edit : Problem, I actually modified no JS code, but 2 failed tests are blocking me

### DIFF
--- a/packages/styles/scss/components/number-input/_number-input.scss
+++ b/packages/styles/scss/components/number-input/_number-input.scss
@@ -88,7 +88,7 @@
     }
 
     &[data-invalid] {
-      padding-inline-end: convert.to-rem(112px);
+      padding-inline-end: convert.to-rem(128px);
     }
   }
 
--- a/packages/styles/scss/components/number-input/_number-input.scss
+++ b/packages/styles/scss/components/number-input/_number-input.scss
@@ -88,7 +88,7 @@
     }
 
     &[data-invalid] {
-      padding-inline-end: convert.to-rem(128px);
+      padding-inline-end: convert.to-rem(112px);
     }
   }
 
--- a/packages/styles/scss/components/number-input/_number-input.scss
+++ b/packages/styles/scss/components/number-input/_number-input.scss
@@ -88,7 +88,7 @@
     }
 
     &[data-invalid] {
-      padding-inline-end: convert.to-rem(128px);
+      padding-inline-end: convert.to-rem(112px);
     }
   }
 
@@ -377,6 +377,13 @@
 
   // No steppers
   .#{$prefix}--number--nosteppers input[type='number'] {
+    padding-inline-end: 0;
+  }
+
+  .#{$prefix}--number--nosteppers input[type='number'][data-invalid],
+  .#{$prefix}--number--nosteppers
+    .#{$prefix}--number__input-wrapper--warning
+    input[type='number'] {
     padding-inline-end: convert.to-rem(48px);
   }
 
--- a/packages/react/src/components/DatePicker/DatePicker-test.js
+++ b/packages/react/src/components/DatePicker/DatePicker-test.js
@@ -211,6 +211,7 @@ describe('Simple date picker', () => {
           default: module.DatePickerInput,
         }))
       );
+
       render(
         <React.Suspense fallback="Loading">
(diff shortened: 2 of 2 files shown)


raphyluke:

packages/styles/scss/components/number-input/_number-input.scss
```diff
@@ -88,7 +88,7 @@
     }
 
     &[data-invalid] {
-      padding-inline-end: convert.to-rem(128px);
+      padding-inline-end: convert.to-rem(112px);
```
tw15egan: This needs to stay `128px`, otherwise the text will run right up against the invalid icon. The [design spec](https://carbondesignsystem.com/components/number-input/style#default-input) calls for `16px` padding between the icon/steppers and the input.
Screenshot 2023-12-19 at 11 20 48 AM


The initial issue reported that the padding when the container was set to `150px` was causing the number to be cut off, since it always assumes the `steppers` to be present:
Screenshot 2023-12-19 at 11 24 54 AM
Screenshot 2023-12-19 at 11 24 05 AM


What we need to do is adjust the padding in the invalid state (as well as the warning state for that matter) when the `steppers` are not present. We can also remove the right padding when the input is not in an invalid/warning state and the `hideSteppers` prop is present

```scss
  .#{$prefix}--number--nosteppers input[type='number'] {
    padding-inline-end: 0;
  }

  .#{$prefix}--number--nosteppers input[type='number'][data-invalid],
  .#{$prefix}--number--nosteppers
    .#{$prefix}--number__input-wrapper--warning
    input[type='number'] {
    padding-inline-end: convert.to-rem(48px);
  }
```

Which should give us something like this: 

Screenshot 2023-12-19 at 11 31 50 AM
Screenshot 2023-12-19 at 11 31 38 AM

tw15egan (changes requested):

packages/styles/scss/components/number-input/_number-input.scss
dev-ghostly: This is looking a lot better thank you, as a beginner in open source project I thought about changing the code less possible, thank you a lot for your help, it learned me something 👍🏻

tw15egan: No worries, thanks for taking the time to make the fix! The problem seems fixed when `hideSteppers` is enabled
Screenshot 2023-12-19 at 4 08 12 PM

But I'm still seeing the reduced padding when the steppers are there and the input is invalid (`L91`)
Screenshot 2023-12-19 at 4 08 02 PM

dev-ghostly: My bad ! Bad ctrl-z while doing push lmao !
Suppose to be good now 😄

tw15egan (approved):
LGTM, thanks for contributing! 👍 ✅

guidari (approved):

PR #15422, opened 2023-12-19
fix: TableToolbarSearch parameter types

wkeese:
Refs #14471.

Fix `TableToolbarSearch` parameter types, in particular missing `closeButtonLabelText`.

Also, refactor the code as done in @types/carbon-components-react: Make `TableToolbarSearchParameters` extend SearchParameters.

IMO the `TableToolbarSearch` parameters have a lot of unnecessary differences from the `Search` parameters, but I didn't change that in this commit.

One open question is why the initial `expandedState` is based on `defaultValue`. That seems like a mistake.  If it is, I can roll it back in this PR. cc @tw15egan 

#### Changelog

**Changed**

- Fix parameters to TableToolbarSearch, in particular missing closeButtonLabelText

See also #15395.


packages/react/src/components/DataTable/TableToolbarSearch.tsx
```diff
@@ -159,9 +142,12 @@ const TableToolbarSearch = ({
...
 }: TableToolbarSearchProps) => {
   const { current: controlled } = useRef(expandedProp !== undefined);
-  const [expandedState, setExpandedState] = useState<
-    string | boolean | undefined
-  >(defaultExpanded || defaultValue);
+
+  // TODO: Why does defaultValue affect the initial expandedState?  This is from #5089.
+  const [expandedState, setExpandedState] = useState<boolean>(
+    Boolean(defaultExpanded || defaultValue)
+  );
```
tay1orjones: @wkeese Would this be because if there is a `defaultValue` the input should be expanded by default? That's the primary reason I can think of.

wkeese: I guess you're right, actually I'm unclear what "expanded" means for a search toolbar...

In any case I removed the TODO comment.

tw15egan (approved):

tay1orjones (approved):

PR #15431, opened 2023-12-19
fix: Datepicker closeOnSelect prop

### LINKED ISSUE #15215: [Bug]: Datepicker `closeOnSelect` prop is not working
### Package

@carbon/react

### Browser

_No response_

### Package version

1.42

### React version

18.2

### Description

The prop `closeOnSelect` is not working as expected. It is suppose to avoid the close action when the user select a date in the calendar if the prop is set as `false`.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datepicker--playground&args=closeOnSelect:!false

### Steps to reproduce

- Go to the Datepicker playground and set the `closeOnSelect` to false;
- Open tge calendar and select a date;
- (Expected) Calendar should NOT close 

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

AdityaPimpalkar:
Closes #15215

Datepicker should avoid the close action when the user select a date in the calendar if the `closeOnSelect` prop is set as `false`.

#### Changelog

**New**

- Test case to check the intended behaviour when the prop is set to the desired condition

**Changed**

_None_

**Removed**

_None_

### DIFF
--- a/packages/react/src/components/DatePicker/DatePicker-test.js
+++ b/packages/react/src/components/DatePicker/DatePicker-test.js
@@ -315,6 +315,46 @@ describe('Single date picker', () => {
     await userEvent.click(screen.getByText('clear'));
     expect(screen.getByLabelText('Date Picker label')).toHaveValue('');
   });
+
+  it('should respect closeOnSelect prop', async () => {
+    const DatePickerExample = () => {
+      const [date, setDate] = useState();
+      return (
+        <DatePicker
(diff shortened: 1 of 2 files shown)


AdityaPimpalkar:

guidari (approved):
LGTM! 🚀
Thanks for adding the tests also!

tay1orjones (approved):
Thanks!

PR #15443, opened 2023-12-23
fix(react): tableSelectAll checkbox checked=true when it's not(#15386)

### LINKED ISSUE #15386: [Bug]: TableSelectAll checkbox checked=true when it's not.
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

1.42.1

### React version

18.2.0

### Description

├── @carbon/icons-react@11.30.1
├── @carbon/react@1.42.1
├── @types/carbon__icons-react@11.29.0
├── react@18.2.0

### Reproduction/example

none

### Steps to reproduce

Screenshot 2023-12-13 at 12 36 46 PM

Going from `semi-checked` to `unchecked` results on a e.target.checked of `true` even though it should be `false` because the checkbox is unchecked.

```javascript
<TableSelectAll
{...getSelectionProps()}
onSelect={(e) => {
  console.log(e.target.checked)
}}
``` >

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

harukikuri:
Closes #15386 

when the state is completely unchecked, it shows as `checked=true`.

#### Changelog

**New**
- Created a new `onClickCheckBoxInput` function that handles `checked` status updates.

#### Testing / Reviewing

(Cause)
We change the checked state of the HTML element but since that process depends on useEffect, we can not see the changed value inside the onClick function(client side: onSelect) event object. So, we need to migrate this updating process to the actual onClick function so that users can see the correct checked state value associated with the event.

https://github.com/carbon-design-system/carbon/blob/d177d2685c5767abf7e9798359ebc7b47b13e313/packages/react/src/components/InlineCheckbox/InlineCheckbox.js#L50-L58

(Solution)
Created a new function called onClickCheckBoxInput that takes care of this update. Accordingly updated the onClick prop in the inputProps.


guidari (approved):
LGTM! 🚀

alisonjoseph (approved):

PR #15444, opened 2023-12-25
fix(react): fluidDropdownSkeleton className in wrong element (#15442)

### LINKED ISSUE #15442: [Bug]: `FluidDropdownSkeleton` `className` is being added to wrong element
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.46.1

### React version

v18.2.0

### Description

I'm using `FluidDropdown` in a `FluidForm`, and it's in a `flex` layout where I'd like it to grow equally with its peers to fill the available width in the row. This is working well, because I can apply styles via a `className` to the `FluidDropdown` and it gets put on the top-level DOM element of the dropdown i.e. the one that also has `cds--dropdown__wrapper cds--list-box__wrapper cds--list-box__wrapper--fluid`.

While the data for the dropdown is loading I'm using a `FluidDropdownSkeleton` but this is failing to grow equally with its peers because the styles are not applied to the right place, but to a child element.

For demonstration here's the moment where one dropdown has loaded but not the other. Note that `big-input` is applied at the top level to the top one (`FluidDropdown`), but applied at the second level of the bottom one (`FluidDropdownSkeleton`):

Snapshot of DOM when one dropdown has loaded but not the other

### Reproduction/example

https://stackblitz.com/edit/github-tymoau?file=src%2FApp.jsx

### Steps to reproduce

1. Use a `FluidDropdownSkeleton`.
1. Apply styles to the `FluidDropdownSkeleton` via `className` to influence its position in its parent, such as `flex-grow` or `flex-basis`.
1. Notice that the `FluidDropdownSkeleton` does not lay out as expected because the styles are not applied at the top level of the `FluidDropdownSkeleton`.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

harukikuri:
Closes #15442

FluidDropdownSkeleton `className` is being added to the wrong element.
Screenshot 2023-12-24 at 4 42 15 PM

#### Changelog

**New**
- created a new variable `wrapperContainerClasses` for joining `className`(from the client side) and `${prefix}--list-box__wrapper--fluid`.

#### Testing / Reviewing

The `className` prop is added to the parent element instead of the child element.


guidari (approved):
LGTM! 🚀
Thank you!

kuri-sun:
> LGTM! 🚀 Thank you!

Thanks for the review!

alisonjoseph (approved):

PR #15451, opened 2023-12-29
fix(react): headerMenu requiring isActive prop #15351

### LINKED ISSUE #15351: HeaderMenu requiring isActive prop [Bug]:
### Package

@carbon/react, @carbon/icons

### Browser

Chrome

### Package version

carbon/react 1.44.0, icons-react 11.31.0

### React version

18

### Description

For some reason the `HeaderMenu` component is requiring the `isActive` prop to always be true, or it breaks with:
`Unhandled Runtime Error
TypeError: Cannot read properties of undefined (reading 'isActive')
Call Stack
eval
node_modules/@carbon/react/es/components/UIShell/HeaderMenu.js (151:70)`
In earlier versions, I have never needed to provide the `isActive` prop to `HeaderMenu`. 

Also, when following the carbon, react, next.js docs exactly (https://carbondesignsystem.com/developing/react-tutorial/step-1), I am getting the following warning in the console: 
`app-index.js:32 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`
This is due to the use of the `Link` component around carbon components, as described in the docs. 

The CodeSandbox below is currently broken due to another [issue](https://github.com/carbon-design-system/carbon/issues/15306)
It will not let you import UI Shell components due to the following error in the next icons package: 
`Failed to compile
./node_modules/@carbon/icons-react/es/generated/bucket-8.js
Module parse failed: Export 'Infinity' is not defined (2620:1241)
|     WatsonHealthLaunchStudy_1.propTypes = iconPropTypes;
| }
export { IbmCloudPakIntegration, IbmCloudPakMantaAutomatedDataLineage, IbmCloudPakMulticloudMgmt,...`
Described in the linked issue.

Once that issue is fixed, the CodeSandbox will show the errors described above. 

### Reproduction/example

https://codesandbox.io/p/devbox/headermenu-issue-88wg4c?file=%2Fapp%2Fnavbar.js%3A26%2C47

### Steps to reproduce

run in dev, after installing dependencies 

### Suggested Severity

### Application/PAL

IBM CIO apps

### Code of Conduct

harukikuri:
Closes #15351

HeaderMenu component throws an error when there is some nested plain text element inside.

#### Changelog

**Changed**
- Put a null check for `child.props` for the `hasActiveDescendant` function.

#### Testing / Reviewing

(Cause of the issue)
This happens because the `hasActiveDescendant` function does not consider the case when the HeaderMenu component has plain text as a child. The example that @dbryan17 provides us is using plain text as a child of the HeaderMenu component, so plain texts don't have any `child.props`, the function will throw an error.
https://github.com/carbon-design-system/carbon/blob/3f70e1deab402402dfb18dcc3ae4c7057818ab82/packages/react/src/components/UIShell/HeaderMenu.js#L215-L222

Screenshot 2023-12-29 at 11 01 55 AM
For example, If you put plain text inside the HeaderMenu component, you will see the error.

(Solution)
The solution for this issue is to make this component accept plain text elements and have the `hasActiveDescendant` function throw `false` when it is the case. So that I make sure that put a null check for `child.props`.

### DIFF
--- a/packages/react/src/components/UIShell/HeaderMenu.js
+++ b/packages/react/src/components/UIShell/HeaderMenu.js
@@ -215,10 +215,11 @@ class HeaderMenu extends React.Component {
     const hasActiveDescendant = (childrenArg) =>
       React.Children.toArray(childrenArg).some(
         (child) =>
-          child.props.isActive ||
-          child.props.isCurrentPage ||
-          (child.props.children instanceof Array &&
-            hasActiveDescendant(child.props.children))
+          child.props &&
+          (child.props.isActive ||
+            child.props.isCurrentPage ||
+            (child.props.children instanceof Array &&
+              hasActiveDescendant(child.props.children)))
       );
 
     const accessibilityLabel = {


techtolentino:
✋🏼 Was this approved and merged? We're running to the same issue in a current project

kuri-sun:
@techtolentino  Sorry, I accidentally deleted the branch! I will recreate the PR again! Thanks for letting me know!

PR #15504, opened 2024-01-05
Update telemetry.yml project Id

francinelucca:


jharvey10 (approved):
👍  We are changing this because we will get better insights if we do "1 project = 1 package"

alisonjoseph (approved):

PR #15507, opened 2024-01-08
VoiceOver counter in `TextArea`

### LINKED ISSUE #14818: [a11y]: TextArea [playground] - remaining character count not announced by SR
### Package

@carbon/react

### Browser

FF, Safari & iPhone

### Package version

@carbon/react@1.32.0

Manual QA

### Assistive technology

JAWS & VO

### Description
As we type the input in the edit area the remaining character count is changing Dynamically but it is not announced by Screen Reader Instantly

If the screen reader fails to identify a dynamic content change, it affects users who are blind as they may not know about the change in the page.

### Recommended Fix
For the benefit of Screen Reader user its good to announce the Dynamic Changes instantly. Please refer the following link https://codepen.io/rikschennink/pen/LmoJYY

### WCAG 2.1 Violation

4.1.3 Status Messages

guidari:
Closes #14818

Added a code from this library called [short-and-sweet](https://github.com/rikschennink/short-and-sweet/tree/master) to announce in the VO how many characters are left to type.

It only works in the `counterMode = "character"` because it has the `maxLength` attribute in the `textArea` tag. The `counterMode = "word"` does not contain the `maxLength`

#### Changelog

**New**

- Added a new attribute called `data-counter-label`
- Added code from this library [short-and-sweet](https://github.com/rikschennink/short-and-sweet/tree/master)

#### Testing / Reviewing

- Go the Playground `TextArea` and use the VO to test it


andreancardona:
@guidari something is going on with the tests :(

tay1orjones:
Hey, I'm curious why is the existing solution not working? and why is short-and-sweet necessary? I'd prefer to avoid adding third party code for this if we can.

guidari:
> Hey, I'm curious why is the existing solution not working? and why is short-and-sweet necessary? I'd prefer to avoid adding third party code for this if we can.

@tay1orjones the existing implementation doesn't read the remaning amount of characters left to type in the `textArea` or `inputs`. This library could improve the a11y related to inputs that has the attribute `maxLength` implemented.

@andreancardona the code was broken because I was using the `short-and-sweet.min.js` file. I changed for the version that it is on the `src` folder from the lib.

andreancardona (approved):

guidari:
@andreancardona @tay1orjones 
I'll close this PR for these reason:

- I just figured it out that the word count starts when there is 10 characters left to type (from 490 to 500 characters). The solution I inserted counts the characters from the beggining, like from 0 to 500 characters.
- We don't need to add third party code since the existing solution takes care of warning the user.

PR #15515, opened 2024-01-09
fix(Modal): add tabindex to content body when content is scrollable, storybook enhancements 

### LINKED ISSUE #13710: [a11y]: Modal: Most of the modals are not properly dismissable in storybook

> The modals are not launchable or dismissable in storybook, except the one "With State Manager". This means basic modal behaviour cannot be assessed. This just seems like an implementation bug, but since modals have very specific behaviour, it's important the demos are functional

### LINKED ISSUE #13711: [a11y]: Modal: A keyboard user cannot activate the scrollbars on a dialog

> On the [Full Width composed modal](https://react.carbondesignsystem.com/?path=/story/components-composedmodal--full-width), the dialog's body contains a large text-only table. Scrollbars appear to let this table be viewed; however, while a mouse user can easily click on the scrollbars, they cannot be activated by a keyboard user. In order to allow scrolling, the body of the dialog needs to be given a tabindex (`tabindex="-1"` is probably adequate) and have the initial focus set to it. This is also a problem where there are form inputs. The body gets the focus in this situation, but since the keyboard user is inside a form input, the up/down and space keys are eaten by the input; if users have dedicated pg up/down keys, they have a work around **sev 2**
> 
> One way to address this is to put a tabindex on the dialog body. Ideally this would only occur where scrollbars are detected (so as not to put another tabstop in all dialogs, and so as not to disrupt the usual item taking focus). There is a hasScrollingContent boolean in Carbon, that if set to true resolves this, but it must be set by the developer. If developers were dilligent in its use, this would solve a lot of use cases; however, where a user zooms up there content or increases their text spacing, they may have scrollbars even though a dialog did not contain them with default text size. So it would be ideal if the presence of scrollbars could be detected automatically and the tabindex only set then.  
> Finally, ideally this would not take focus on modal load, but leave the default item with focus (as described in the accessibilty's keyboard section); the body would be in the tab order, so could be tabbed to.
>
> 
> ### Steps to reproduce
> 
> 1. Open the full screen version of the State manager https://react.carbondesignsystem.com/iframe.html?args=&id=components-modal--with-state-manager&viewMode=story
> 2. Increase browser zoom to 400%
> 3. Launch modal
> 4. Attempt to navigate by Tab around input

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/13711
Closes https://github.com/carbon-design-system/carbon/issues/13710

Adds some checks inside `Modal` and `ComposedModal` to see if the modal content is larger than the content viewbox. If it is, it sets a `tabIndex` on the content so a user can scroll it with a keyboard. The function is denounced and is updated on any window resize event. 

This also adds a `Button` to each `ComposedModal` and `Modal` story so that they can be opened and closed

#### Changelog

**New**

- Added an [effect](https://github.com/carbon-design-system/carbon/compare/main...tw15egan:modal-a11y-enhancements?expand=1#diff-728e08310e02c71a4fcea7c83cd5b62d22d105dc2e3dfec51e894eef621e1d21R433) to check if the content is larger than the viewbox, and if so, adds a `tabIndex`. It will also automatically toggle the `hasScrollingContent` prop.
- Added launcher buttons to each `ComposedModal` and `Modal` story

#### Testing / Reviewing

Check each `ComposedModal` and `Modal` story and ensure they can all be toggled open / closed. Increase your browser zoom or open up dev tools to make the Modal scrollable. The `hasScrollingContent` indicator should show up, and the content body itself should be tabbable, and you should be able to use your arrow keys to scroll up and down. Remove the zoom / dev tools, and the tab index should be removed and there should be no overflow indicator.


guidari (approved):
LGTM! Working as expected

tw15egan:
@tay1orjones can I get a review on this one?

tay1orjones (approved):
Sorry for the delay!

PR #15516, opened 2024-01-10
chore: fix Dropdown's translateWithId() signature

wkeese:
Restore `ListBoxMenuIconTranslationKey` type that was available in Carbon 10. Relatedly, fix `ListBoxMenuIcon`'s `translateWithId()` signature so `messageId` is type `ListBoxMenuIconTranslationKey`, rather than just type `string`.

Finally, do the identical fix to `Dropdown`'s `translateWithId()` signature, and export a `DropdownTranslationKey` type.

The `translationIds` objects from #435 (cc @joshblack) didn't seem to be helping, especially since they aren't exported from ListBox/index.ts. I removed `ListBoxMenuIcon`'s `translationIds` object as part of adding this typing.

Note that neither `ListBoxMenuIconTranslationKey` nor even `ListBox` itself is exported from @carbon/react.  But `Dropdown` is.


tw15egan (approved):
LGTM 👍 ✅

alisonjoseph (approved):

PR #15519, opened 2024-01-10
fix(Modal): secondaryButtons prop type mismatch

### LINKED ISSUE #15392: [Question]: Type mismatch in secondaryButtons prop in Modal component
### Question for Carbon

Hi. I've encountered an issue with type mismatch using the `Modal` component. The `Modal` component has a `secondaryButtons` prop that accepts string based on its TypeScript annotation. However, under the hood, it treats the prop as a node. I've shared all annotations below.
The `secondaryButtons` prop: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Modal/Modal.tsx#L185C31-L185C31
The `ModalSecondaryButton` type: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Modal/Modal.tsx#L34
The `secondaryButtons` actual validation: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Modal/Modal.tsx#L721C8-L721C8

Therefore, I have a reasonable question it is an intentional change or just a mistake?
Thanks in advance.

### Code of Conduct

htndev:
Closes https://github.com/carbon-design-system/carbon/issues/15392

The issue has a type mismatch. The `Modal` component has a `secondaryButtons` that accepts the `string` type but also differs from the existing implementation.

#### Changelog

**Changed**

- Fixed type mismatch

#### Testing / Reviewing

Check whether it accepts not only strings.

### DIFF
--- a/packages/react/src/components/Modal/Modal.tsx
+++ b/packages/react/src/components/Modal/Modal.tsx
@@ -32,7 +32,7 @@ export const ModalSizes = ['xs', 'sm', 'md', 'lg'] as const;
 export type ModalSize = (typeof ModalSizes)[number];
 
 export interface ModalSecondaryButton {
-  buttonText?: string;
+  buttonText?: string | React.ReactNode;
 
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
 }


htndev:

tay1orjones (approved):
Thanks!

andreancardona (approved):
🔥 Thank you!

Issue #15319, opened 2023-12-04
[Bug]: Keyboard Navigation Stuck Within Actionable Notification Component 


ellamathewsonibm:
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.42.1

### React version

v18.2.0

### Description

When using keyboard navigation on Actionable notification, tab focus cannot escape the alert and instead cycles through the action button and close button indefinitely. When the notification only has either the action or close buttons, the tab focus does not move from the button. 

https://github.com/carbon-design-system/carbon/assets/152651072/49814e2d-3df0-419d-b096-085836e42be9

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-notifications-actionable--playground&args=kind:info;lowContrast:!true;statusIconDescription:in;inline:!true

### Steps to reproduce

Go to Notifications -> Actionable -> Playground. Press `Tab` on keyboard to navigate a couple times. See focus remains switching between the x close icon button and the action button, never leaving the component. 

### Suggested Severity

Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.

### Application/PAL

w3 Carbon

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

kuri-sun:
Hello! I love to work on this issue!

kuri-sun:
Hello! I love to work on this issue! May I?

ellamathewsonibm:
> Hello! I love to work on this issue! May I?

Yes please! Thank you

lluisrojass:
@kuri-sun the behavior seems unintentionally aggressive. When you render an actionable notification without a close button and turn `hasEscape` to false, focus is trapped indefinitely.

lluisrojass:
@kuri-sun 

from: https://carbondesignsystem.com/components/notification/usage/#actionable-notifications
> Actionable notifications allow for interactive elements within a notification styled like an inline or toast notification. Actionable notifications, since they require user interaction, take focus when triggered and can be highly disruptive to screen readers and keyboard users. With actionable notifications, only one action is allowed per notification. This action frequently takes users to a flow or page related to the message, where they can resolve the notification

[...]

PR #15532, opened 2024-01-12
feat(StaticNotification): initial component functionality

### LINKED ISSUE #15697: [Bug]: Click on Link button in ActionableNotification is moving the cursor focus to the Close button.
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.47.0

### React version

17.0.2

### Description

When using ActionableNotification, the focus always moves to Close Button when clicked on the link. It doesn't allow to click on any Tabs or other buttons. Only by closing the notification, I am able to switch tabs. 

@

### Reproduction/example

https://codesandbox.io/p/sandbox/tex-yqr2z9?file=%2Fsrc%2Findex.js%3A17%2C53

### Steps to reproduce

Use ActionableNotification with Action button. Click on Action button, click on the textbox, the focus moves to the close button. And only by clicking on the close button, you will be able to get the focus into the textbox.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/15697

Based on conversations with @finken2 in https://github.com/carbon-design-system/carbon/issues/15319 and a few meetings later about blockers faced by their team, and consulting with IBMa - this introduces a new notification component with specific guidance outlined in storybook. 

We'll test this with users before adding tests, marking stable, etc.

#### Changelog

**New**

- Added experimental `unstable__StaticNotification`

#### Testing / Reviewing

- Ensure the warning only pops when it should, and doesn't when it shouldn't


guidari:
@tay1orjones  Locally I can see the warning in the `pop-the-warning` variant:
Error: interactive child node(s) should have an `aria-describedby` property with a value matching the value of `titleId`.

But if you check the deploy both variants looks the same. The `experimental-unstable-staticnotification--with-interactive-elements-that-pop-the-warning` is not poping the warning like is happening locally.

guidari (approved):
LGTM! 🚀
We just have to update the snapshot to fix the test error. For some reason I can't push a commit here

tay1orjones:
@guidari Yes this is actually intentional, the warning is guarded behind `__DEV__` which is only true in non-production environments. So the warning will never pop in production, but will be present in development environments where it can be appropriately addressed by devs. 

I'll see if I can get that snapshot updated to pass the ci check

tw15egan (approved):

PR #15313, opened 2023-12-04
fix: fixed TypeError of undefined issue triage action

guidari:
This issue attempts to fix the triage action to add a label when new issues are created

The job that it is failing: https://github.com/carbon-design-system/carbon/actions/runs/7060031440

Screenshot 2023-12-04 at 11 54 27

#### Testing / Reviewing

I'll re-run the job that is failing to verify if it will fix it

PR #15533, opened 2024-01-12
Triage Bot: update octokit code to v20

guidari:
The docs updated and might be because of this update that the `addLabels` was getting undefined in this [PR](https://github.com/carbon-design-system/carbon/pull/15313)

Old doc: https://octokit.github.io/rest.js/v16#:~:text=.issues.-,addLabels,-(%7B%0A%20%20owner
New doc: https://octokit.github.io/rest.js/v20#:~:text=.issues.-,addLabels,-(%7B%0A%20%20%20%20%20%20%20%20owner

If this change works it will not only fix the triage bot, but also the triage response and review label actions.

#### Testing / Reviewing

The test will be made after the PR is merged

### DIFF
--- a/actions/add-review-labels/index.js
+++ b/actions/add-review-labels/index.js
@@ -96,7 +96,7 @@ async function run() {
       return label.name === readyForReviewLabel;
     });
     if (hasReadyLabel) {
-      await octokit.issues.removeLabel({
+      await octokit.rest.issues.removeLabel({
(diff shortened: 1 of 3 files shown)


tay1orjones (approved):
Thanks again for catching this 🙏

andreancardona (approved):
Thank you for working on this! 🔥

PR #15535, opened 2024-01-12
fix(react): tooltip doesn't work when trigger has tabindex=-1

### LINKED ISSUE #15214: [Bug]: Tooltip doesn't work when trigger has tabindex="-1"
### Package

@carbon/react

### Browser

Chrome, Safari

### Package version

1.42.0

### React version

18.2.0

### Description

When the trigger button of a `Tooltip` has `tabindex` set to `-1` the tooltip doesn't appear at all on hover and only for a flicker on click.

Context / use case: the button is within a widget that manages the focus (e.g. grid).

### Reproduction/example

https://stackblitz.com/edit/github-mje2hg?file=src%2FApp.jsx

### Steps to reproduce

```jsx
// ✅ works
<Tooltip>
  <button>Trigger</button>
</Tooltip>
```

```jsx
// ✅ works
<Tooltip>
  <button tabindex="0">Trigger</button>
</Tooltip>
```

```jsx
// ⛔️ doesn't work
<Tooltip>
  <button tabindex="-1">Trigger</button>
</Tooltip>
```

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

harukikuri:
Closes #15214

When the trigger button of a Tooltip has tabindex set to -1 the tooltip doesn't appear at all on hover and only for a flicker on click.

#### Changelog

**Removed**

-  the `useEffect` (line 203 in Tooltip.tsx).
https://github.com/carbon-design-system/carbon/blob/8c0130499a08892c6e89e7f6d198d10462b1e017/packages/react/src/components/Tooltip/Tooltip.tsx#L204-L213

#### Testing / Reviewing

(Cause of this issue)
I concluded that this happens because of the `useEffect` (line 203 in Tooltip.tsx).
https://github.com/carbon-design-system/carbon/blob/8c0130499a08892c6e89e7f6d198d10462b1e017/packages/react/src/components/Tooltip/Tooltip.tsx#L204-L213

When the user has a `tabIndex=-1` element, `interactiveContent` is always `null` since `getInteractiveContent` checks whether the elements are focusable or not by `if (element.tabIndex === undefined || element.tabIndex < 0` condition in the `isFocusable` function.
https://github.com/carbon-design-system/carbon/blob/8c0130499a08892c6e89e7f6d198d10462b1e017/packages/react/src/internal/useNoInteractiveChildren.js#L87-L90

We can just remove this `useEffect` since we don't need the focusable state detection that makes TooltipContent always hidden.

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react/src/components/Tooltip/Tooltip.tsx
@@ -12,10 +12,7 @@ import { Popover, PopoverAlignment, PopoverContent } from '../Popover';
 import { keys, match } from '../../internal/keyboard';
 import { useDelayedState } from '../../internal/useDelayedState';
 import { useId } from '../../internal/useId';
-import {
-  useNoInteractiveChildren,
-  getInteractiveContent,
-} from '../../internal/useNoInteractiveChildren';
+import { useNoInteractiveChildren } from '../../internal/useNoInteractiveChildren';
 import { usePrefix } from '../../internal/usePrefix';
 import { type PolymorphicProps } from '../../types/common';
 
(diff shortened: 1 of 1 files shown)


tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):
looks good thank you @kuri-sun !

PR #15551, opened 2024-01-17
fix: 15327-incorrect-padding-sizes

### LINKED ISSUE #15327: [Data table] Incorrect Expandable panel padding sizes 🐛
### Problem
Link: https://carbondesignsystem.com/components/data-table/style#enhancements
The guidance states that it the expandable panel needs 16px padding-top and 24px padding-bottom an this should be consistent in all sizes. 

image

However, the code and Figma kit (seems to follow the specs from the code) have inconsistent paddings 

----
### Solution
In the past, it seems like [the issue was fixed](https://github.com/carbon-design-system/carbon/issues/4651) in the older version but it was not carried over to the newer version or this current version (**v11 1.44.0** ) 

This needs to update so that all sizes have consistent 16px padding-top and 24px padding-bottom for all sizes 

---

```[tasklist]
### Possible tasks in the future
- [ ] Update the specs for data table expansion in the Storybook 
- [ ] Update the Figma kit so it's consistent with the code
```

riddhybansal:
Closes #15327 

There were inconsistent padding in data tables.

#### Changelog

**Changed**

-  Changed some styles 

#### Testing / Reviewing

The guidance from figma designs states that  the expandable panel needs 16px padding-top and 24px padding-bottom and this should be consistent in all sizes.


packages/styles/scss/components/data-table/expandable/_data-table-expandable.scss
```diff
@@ -101,6 +101,8 @@
     td
     .#{$prefix}--child-row-inner-container {
     max-block-size: 100%;
+    padding-block: 1rem;
+    padding-block-end: 1.5rem;
```
tw15egan, line 105: We can just combine these in the `padding-block` definition 👍🏻 

```suggestion
    padding-block: 1rem 1.5rem;
```

tw15egan (approved):
Small nit, but otherwise looks great!

guidari (approved):
LGTM! 🚀

packages/styles/scss/components/data-table/expandable/_data-table-expandable.scss
guidari: @riddhybansal You can commit the suggestion directly in the "Commit suggestion" or push a new commit to address the changes TJ posted it

PR #15571, opened 2024-01-19
fix: added hasFocus as deprecated prop

### LINKED ISSUE #15534: [Notification]: Investigate deprecating `hasFocus` prop
I think we actually need to deprecate hasFocus - [the spec for alertdialog](https://w3c.github.io/aria/#alertdialog) states that focus should always be given to a focusable element within the dialog. This [originally came up in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1705079151446859)

In https://github.com/carbon-design-system/carbon/pull/14877, focus wrap behavior was implemented for the ActionableNotification. I think it was a slight oversight to leave `hasFocus` in. Even with `hasFocus` off, I believe the `wrapFocus()` helper always focuses the `currentActiveElement` that is passed in.

Even if it doesn't, this prop should be deprecated because the spec states that focus should _always_ move to the alertdialog. 

- [ ] Deprecate the `hasFocus` prop
- [ ] Ensure focus always moves to the ActionableNotification

I think many of the use cases previously using `ActionableNotification hasFocus="false"` should actually use StaticNotification once it's available, #15532

guidari:
Closes #15534

Deprecating `hasFocus` because we are gonna have a new variant in the future called `StaticNotification` to avoid confusion around the usage of the `ActionableNotification`.

**Changed**

- `hasFocus` prop is deprecated
- Removed `hasFocus` prop from Default and Playground stories

#### Testing / Reviewing

- Ensure focus always moves to the `ActionableNotification`

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -109,9 +109,7 @@ Map {
       "closeOnEscape": Object {
         "type": "bool",
       },
-      "hasFocus": Object {
-        "type": "bool",
-      },
+      "hasFocus": [Function],
       "hideCloseButton": Object {
         "type": "bool",
       },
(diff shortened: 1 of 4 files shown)


alisonjoseph (approved):

andreancardona (approved):
🔥

PR #15591, opened 2024-01-23
fix(Slug): adjust callout width, fix revert story

tw15egan:
Fixes some issues that @aagonzales pointed out with the latest `AI` updates

Screenshot 2024-01-23 at 12 21 11 PM

#### Changelog

**Changed**

- Set the same `max-width` for all callouts (action vs without action)
- Ensure the ai border on `NumberInput` is removed when the ai styles are not active (revert test story)

#### Testing / Reviewing

Ensure padding is correct for the action bar variant of the callout. Ensure the `NumberInput` border is removed when a user changes the ai value (`Slug --> Form --> Revert Test`)

### DIFF
--- a/packages/styles/scss/components/number-input/_number-input.scss
+++ b/packages/styles/scss/components/number-input/_number-input.scss
@@ -453,7 +453,10 @@
     padding-inline-end: convert.to-rem(112px);
   }
 
-  .#{$prefix}--number__input-wrapper--slug .#{$prefix}--number__control-btn {
+  .#{$prefix}--number__input-wrapper--slug
+    input[type='number']:not(:has(~ .#{$prefix}--slug--revert))
+    ~ .#{$prefix}--number__controls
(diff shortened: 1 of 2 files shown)


andreancardona (approved):

andreancardona:
One small thing I'm seeing is that when the disabled prop is set to true - there is a conflicting red / blue border in the bottom right hand corner of the input container. Let me know what you think!
Screenshot 2024-01-23 at 09 45 13

andreancardona (commented):
comment above

andreancardona (changes requested):
Comment above not sure what's happening with my github 👀

tw15egan:
Oh, nice catch, @andreancardona! Fixed, it should be good now.

andreancardona (approved):

aagonzales (approved):
Looks good!

PR #15609, opened 2024-01-26
docs(global-theme): clarify that styles won't be applied automatically

janhassel:
Expands the documentation around `GlobalTheme` to avoid confusion over why no css custom properties are emitted when changing its `theme` prop.


alisonjoseph (approved):
Thank you for updating these docs! 🙌

tay1orjones (approved):
Thank you so much, this is a great addition. We've had questions about this before and didn't have anywhere to point folks.

PR #15631, opened 2024-01-29
feat(Slug): add Tooltip to Slug

### LINKED ISSUE #15616: Add tooltip to AI `Slug`
Add a tooltip on hover (pre-click) for the default slug and inline slug (excluding the hollow slugs)
Text: "Explain AI"

image

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15616

Adds a conditional `Tooltip` around `Slug` if a `TooltipDescription` is provided. 

#### Changelog

**New**

- `TooltipDescription` provides the text that will be rendered inside the `Tooltip` for `Slug` that is activated on hover/focus of the `Slug` itself
- `TooltipAlignment` specifies the alignment of the `Tooltip` for `Slug`

**Changed**

- Added a `ConditionalWrapper` that wraps the `Slug` contents and is only rendered if `TooltipDescription` is provided. This avoids us having to set a `TooltipDescription` by default, which would be needed since a small popover appears even if there is no `description` provided for the `Tooltip`
- Updated the stories to use this new prop

#### Testing / Reviewing

Go to the default `Slug` story, ensure the `Tooltip` appears on hover/focus, and close when the `Slug` is clicked. 

@aagonzales do we want to add this `Tooltip` to all `Slug` examples? Is this label mandatory? If so, I can always render the `Tooltip`, but not sure what we want to have as the default fallback text

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -10372,6 +10372,28 @@ Map {
       "slugLabel": Object {
         "type": "string",
       },
+      "tooltipAlignment": Object {
+        "args": Array [
+          Array [
+            "top",
+            "top-left",
+            "top-right",
+            "bottom",
+            "bottom-left",
+            "bottom-right",
+            "left",
+            "left-bottom",
+            "left-top",
+            "right",
+            "right-bottom",
+            "right-top",
+          ],
+        ],
+        "type": "oneOf",
+      },
+      "tooltipDescription": Object {
+        "type": "string",
+      },
     },
     "render": [Function],
   },
--- a/packages/react/src/components/Slug/Slug.stories.js
+++ b/packages/react/src/components/Slug/Slug.stories.js
@@ -46,13 +46,13 @@ const content = <span>AI was used to generate this content</span>;
 export const Default = () => (
   <>
     <div className="slug-container slug-container-example">
-      <Slug autoAlign size="mini">
+      <Slug tooltipDescription="Explain AI" autoAlign size="mini">
         <SlugContent>{aiContent}</SlugContent>
       </Slug>
-      <Slug autoAlign size="2xs">
+      <Slug tooltipDescription="Explain AI" autoAlign size="2xs">
         <SlugContent>{aiContent}</SlugContent>
       </Slug>
-      <Slug autoAlign size="xs">
+      <Slug tooltipDescription="Explain AI" autoAlign size="xs">
         <SlugContent>
           {aiContent}
           <SlugActions>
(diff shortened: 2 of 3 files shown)


aagonzales:
@tw15egan Yes, I think the tooltip should show up on all the examples as well. I don't think it mandatory per say though unless its easier/cleaner that way then I think we can call it mandatory.

aagonzales:
Follow up question: could we change the tooltip behavior of the `hollow` slugs to reveal on hover? So there's parity between the two types.

aagonzales (commented):
I'm also seeing some bugs with the caret tip.
image
image

tw15egan:
As @aagonzales pointed out, wrapping the `Slug` in a `Tooltip` has caused a change in the underlying HTML, which has introduced style regressions. The issue is that instead of components seeing the `cds--slug` class at the top level, components are receiving `cds--icon-tooltip`. I also noticed an issue with triggering the `Slug` when any of the `ListBox` components had focus / open menu. It would take three clicks to trigger the `Slug` callout. In [the latest deployment](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-slug-examples--multiselect), the menu is closed, and the callout is immediately shown when clicked.

Most of the positioning styles are scoped to the `cds--slug` class, so we would need to go in and scope all `Slug` styles to the new wrapper, `cds--icon-tooltip--slug`. While starting this change, I noticed that we have some complex selectors that need to be updated as well, for example:

```scss
.#{$prefix}--text-area__wrapper--slug
    .#{$prefix}--text-area:not(:has(~ .#{$prefix}--slug--revert)) { }

.#{$prefix}--text-area__wrapper--slug
    .#{$prefix}--slug.#{$prefix}--slug--revert 
```

would need to become 

```scss
.#{$prefix}--text-area__wrapper--slug
    .#{$prefix}--text-area:not(
      :has(~ .#{$prefix}--icon-tooltip--slug .#{$prefix}--slug--revert)
    ) { }

.#{$prefix}--text-area__wrapper--slug
    .#{$prefix}--icon-tooltip--slug:has(.#{$prefix}--slug--revert) {
``` 

This is just a small example from one component; each component would need to be updated to use the new selector and ensure the new selector matches the updated HTML structure. This could introduce regressions, especially when we get into the invalid / warning states paired with the `revertActive` icon.

We would also need a way not to show the `Tooltip` when the `revertActive` prop is active, as that has its own `Tooltip`.

We should also consider moving the `ref` out to the `Tooltip` wrapper rather than placing it on the `Slug` container (currently a `div`).



I want to ensure this is a valid a11y requirement before making these changes, as it is a much more significant change than I had initially anticipated. We do not currently show any tooltip when the `Toggletip` is hovered/focused, and the `Slug` is just a `Toggletip` under the hood.

PR #15727, opened 2024-02-09
feat(Modal): update Modal styles to phase 2 AI spec

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15709

Updates `Modal` to use the new phase 2 AI specs

#### Changelog

**New**
- `$ai-overlay` token added

**Changed**

- `Modal` styles have been updated to phase 2 spec

#### Testing / Reviewing

Ensure `Modal` and `ComposedModal` under `Slug --> Examples` matches the spec and there are no regressions

PR #15734, opened 2024-02-12
refactor(Modal): use mask-image for overflow indicator

tw15egan:
Refs https://github.com/carbon-design-system/carbon/pull/15727

While working on https://github.com/carbon-design-system/carbon/pull/15727 I used `mask-image` to fade out the overflow content inside `Modal` when `hasScrollingContent` is set to true. This PR refactors `Modal` and `ComposedModal` to use this property instead of pseudoelements to achieve the effect.

We could also use this to refactor `CodeSnippet`, but we would need to change the dimensions for RTL since the gradient directions do not use logical properties

#### Changelog

**New**

- Applied `mask-image` to `.cds--modal-scroll-content`. The extra dimensions prevent the mask from overlaying the focus outline + scroll bar.

**Removed**

- Removed `cds--modal-content--overflow-indicator` and all associated styles

#### Testing / Reviewing

Ensure there are no style regressions on the `Modal` and `ComposedModal` with scrolling content.

### DIFF
--- a/packages/react/src/components/ComposedModal/ComposedModal.tsx
+++ b/packages/react/src/components/ComposedModal/ComposedModal.tsx
@@ -62,18 +62,13 @@ export const ModalBody = React.forwardRef<HTMLDivElement, ModalBodyProps>(
       : {};
 
     return (
-      <>
-        <div
-          className={contentClass}
-          {...hasScrollingContentProps}
-          {...rest}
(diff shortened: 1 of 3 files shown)


alisonjoseph (approved):
Nice! LGTM

guidari (approved):

PR #15739, opened 2024-02-12
Update NumberInput.tsx to typecast to number during onChange event

### LINKED ISSUE #15735: [Bug]: NumberInput gives string as a value with an onChange event
### Package

@carbon/react

### Browser

Chrome

### Package version

^1.48.1

### React version

^18.2.0

### Description

When you use the steps buttons (`+`, `-`), the onchange gives a number as a type but when you type the number on the input, it converts it into a string and gives that back to the `onChange` handler. 

As a result, wherever this NumberInput is used, everyone has to typecast it on onChange. There has been bugs created due to this discrepancy. 

Expected:

- NumberInput always uses the Number type when onChange is called. 

### Reproduction/example

https://codesandbox.io/p/devbox/nervous-rgb-7k88h8?file=%2Fpackage.json%3A25%2C15-25%2C22

### Steps to reproduce

- Use the sandbox above
- Use the buttons to increment/decrement the number ( you should see the type to be number and in green)
- Use the keyboard to type the number ( you should see the type to be string and in red)

### Suggested Severity

### Application/PAL

CP4AIOps, AIOps Insights.

### Code of Conduct

dkaushik95:
Closes #15735

This change would typecast the value from the input to number similar to when the step function typecasts. This way, the consumers will always receive a value with type number.

#### Changelog

**New**

- {{new thing}}

**Changed**

- Return value for onChange for value is a type number

**Removed**

- {{removed thing}}

#### Testing / Reviewing


dkaushik95:

tw15egan (approved):

tay1orjones (approved):
Thanks for this!

PR #15740, opened 2024-02-12
fix(modal): removed extra label on modal close `x` hover

### LINKED ISSUE #15732: [Bug]: modal close `x` hover shows 2 tooltip labels
### Package

@carbon/react

### Browser

Chrome

### Package version

`1.50.0` (also in `1.49.0`)

### React version

`17.x.x` (and whatever version used in storybook)

### Description

When hovering over the close button on a modal we are now seeing 2 tooltips:
image

I believe the additional tooltip was introduced by [this line](https://github.com/carbon-design-system/carbon/commit/8b4ba0980f3ec44d262626380eb64eb1e660895a#diff-728e08310e02c71a4fcea7c83cd5b62d22d105dc2e3dfec51e894eef621e1d21R443) which causes some browsers to automatically render a label with the attribute contents.

per https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns - this seems to be a problematic attribute to set and should be removed

### Reproduction/example

This issue could be seen on the `1.50.0` storybook https://react.carbondesignsystem.com/?path=/story/components-modal--default

### Steps to reproduce

Render the modal with a label

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

joker23:
Closes https://github.com/carbon-design-system/carbon/issues/15732

This PR will remove the `title` attribute on the modal close (`x`) button to prevent double labels appearing when hovering over the element.

#### Changelog

**Changed**

- Modal component - no longer render 2 hover over label

#### Testing / Reviewing

- render any modal and hover over the `x` close button


packages/react/src/components/Modal/Modal.tsx
```diff
@@ -194,7 +194,7 @@ export interface ModalProps extends ReactAttr<HTMLDivElement> {
   /**
    * Specify CSS selectors that match DOM elements working as floating menus.
    * Focusing on those elements won't trigger "focus-wrap" behavior
-   */
+   */ f;
```
tw15egan:
```suggestion
```

alisonjoseph (approved):

tw15egan (approved):
LGTM 👍 ✅

PR #15767, opened 2024-02-15
docs(notification): clean up static notification stories

tay1orjones:
I noticed during demo today that these stories could use a clean up. 

#### Changelog

**Changed**

- Improves StaticNotification stories to show a proper use of an inline link, including the required aria-describedby, and opting to use a div with a manually applied subtitle class instead of using the `subtitle` prop (which only allows a string)
- Updated documentation to more clearly state the impact of rendering this component after the initial render of the page as it relates to screenreader user focus and the DOM.

**Removed**

- unecessary testing stories that were focused on the aria-describedby warnings

#### Testing / Reviewing

- Ensure StaticNotification stories work and the docs make sense


packages/react/src/components/Notification/Notification.mdx
```diff
@@ -40,13 +40,21 @@ don't grab focus. Use them to provide the user with an alert, status, or log.
...
-element with a value that matches the `titleId` prop.
+content on the initial render of the page or modal because it will not be
+announced to screenreader users like the other notification components.
+
+As such, this should not be used for real-time notifications or notifications
+responding to user input (unless the page is completely refreshing and bumping
+the users focus back to the first element in the dom/tab order). If a
+StaticNotification is rendered after the initial render, screenreader users'
+focus may have already passed this portion of the DOM and they will not know
+that the notifiation has been rendered until they circle back to the beginning
```
tay1orjones:
```suggestion
that the notification has been rendered until they circle back to the beginning
```

tw15egan (approved):

guidari (approved):

PR #15778, opened 2024-02-19
fix(ChatButton): adjust padding when icon is not present

tw15egan:
Adjusts padding to be `16px` when an icon is not present in `Button`

#### Changelog

**New**

-Added non-icon variants to the `ChatButton` stories

**Changed**

- Reduced padding when the `Button` is not rendering an icon. 

#### Testing / Reviewing

Go to `Experimental --> ChatButton` and ensure the non-icon button padding is correct. Ensure there are no regressions to the icon variant.


aagonzales (approved):
Oh good catch 👍

alisonjoseph (approved):

guidari (approved):

PR #15788, opened 2024-02-20
fix: 15752 invalid tooltip on hovering the select component

### LINKED ISSUE #15752: [Bug]: Select component on hover shows invalid title/tooltip
### Package

carbon-components-react

### Browser

_No response_

### Package version

v8.41.2

### React version

v18.2.0

### Description

https://react.carbondesignsystem.com/?path=/docs/components-select--overview
When we hover to dropdown first time on default selected value (without changing option) hover title is not showing.
After changing value it show value on hover.

Expected: text should show on hover as title. value can be number or any unique key

Example:
```
<Select
  id={`select-2`}
  labelText="Select an option"
  helperText="Optional helper text">
  <SelectItem value="option-1" text="Option 1" />
  <SelectItem value="option-2" text="Option 2" />
  <SelectItem value="option-3" text="Option 3" />
</Select>
```

On hover it should show Option 1, Option 2, Option 3 but it is showing option-1, option-2, option-3. (after changing option from dropdown)

### Reproduction/example

https://react.carbondesignsystem.com/?path=/docs/components-select--overview

### Steps to reproduce

* hover on default selected value. see we do not have any hover title
* Change select dropdown value and again over on selected
* It will show value title not text.

Screenshot 2024-02-12 at 5 23 08 PM

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

huzaifabhutta:
Closes #15752

Fixed the invalid title/no title in Select component on hovering

**New**

- `selectDefaultTitle` which gets select element by it's id. if `defaultValue` is passed, it gets the option with `defaultValue` from select, otherwise get option by using the `selectedIndex`

**Changed**
- To show the valid tooltip replaced select value with option's text by using `selectedIndex` in `select.options`
- To show default option's text in title executed the `selectDefaultTitle` in `useEffect`
- Added a new test case which is '_should show SelectItem text as title_' which renders `Select` with `defaultValue` 'Option 2', expected result should be the title 'Option 2'

#### Testing / Reviewing

- Hover on Select component, it should show the selected Option as tooltip
- If there's no option selected, tooltip should not show

### DIFF
--- a/packages/react/src/components/Select/Select.tsx
+++ b/packages/react/src/components/Select/Select.tsx
@@ -12,6 +12,7 @@ import React, {
   ForwardedRef,
   ReactNode,
   useContext,
+  useEffect,
   useRef,
   useState,
 } from 'react';
@@ -157,6 +158,7 @@ const Select = React.forwardRef(function Select(
     warnText,
     onChange,
     slug,
+    defaultValue,
     ...other
   }: SelectProps,
   ref: ForwardedRef<HTMLSelectElement>
@@ -222,12 +224,20 @@ const Select = React.forwardRef(function Select(
     ariaProps['aria-describedby'] = helper ? helperId : undefined;
   }
 
+  const selectDefaultTitle = (defaultValue) => {
+    const selectElement = document.getElementById(id);
+    const title = defaultValue
+      ? selectElement?.querySelector(`option[value=${defaultValue}]`)?.text
+      : selectElement?.options[selectElement?.selectedIndex]?.text;
+    setTitle(title);
+  };
   const handleFocus = (evt) => {
     setIsFocused(evt.type === 'focus' ? true : false);
   };
 
   const handleChange = (evt) => {
-    setTitle(evt?.target?.value);
+    const selectedIndex = evt?.target?.selectedIndex;
+    setTitle(evt?.target?.options[selectedIndex]?.text);
   };
 
   const readOnlyEventHandlers = {
@@ -255,6 +265,10 @@ const Select = React.forwardRef(function Select(
       size: 'mini',
     });
   }
+  useEffect(() => {
+    selectDefaultTitle(defaultValue);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   const input = (() => {
     return (
@@ -263,6 +277,7 @@ const Select = React.forwardRef(function Select(
           {...other}
           {...ariaProps}
           id={id}
+          defaultValue={defaultValue}
           className={inputClasses}
           disabled={disabled || undefined}
           aria-invalid={invalid || undefined}
--- a/packages/react/src/components/Select/Select.tsx
+++ b/packages/react/src/components/Select/Select.tsx
@@ -224,20 +224,6 @@ const Select = React.forwardRef(function Select(
     ariaProps['aria-describedby'] = helper ? helperId : undefined;
   }
 
-  const selectDefaultTitle = (defaultValue) => {
-    const selectElement = document.getElementById(
-      id
-    ) as HTMLSelectElement | null;
-    if (defaultValue) {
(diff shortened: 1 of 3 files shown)


mHuzefa:

mHuzefa:
recheck

mHuzefa:
@alisonjoseph and @guidari, PR is ready for review. Thanks 😊

guidari (approved):
LGTM! 🚀 Thanks for the contribution!
Could you also change the `Select.stories.js` for the `SelectItem` components values to match "option-2" instead of "Option 2"

andreancardona:
@mHuzefa Hii :) thank you for contributing, is there a way you can fix the merge conflicts? Thank you!

laurenmrice (approved):

packages/react/src/components/Select/Select.tsx
```diff
@@ -222,12 +224,27 @@ const Select = React.forwardRef(function Select(
     ariaProps['aria-describedby'] = helper ? helperId : undefined;
   }
 
+  const selectDefaultTitle = (defaultValue) => {
+    const selectElement = document.getElementById(
+      id
+    ) as HTMLSelectElement | null;
+    if (defaultValue) {
+      const defaultOption: HTMLOptionElement | null | undefined =
+        selectElement?.querySelector(`option[value=${defaultValue}]`);
```
tay1orjones: Is there a way to declaratively get the value from `children` instead of using this imperative selector?

packages/react/src/components/Select/Select.tsx
```diff
@@ -255,6 +272,10 @@ const Select = React.forwardRef(function Select(
       size: 'mini',
     });
   }
+  useEffect(() => {
+    selectDefaultTitle(defaultValue);
```
tay1orjones: This doesn't need to be in a function, just place the code directly in the useEffect. It'll be easier to maintain this way being all in once place and not having to scroll up to find the function definition.

tay1orjones (changes requested):
Thanks for taking a look at this - I have a couple suggestions

packages/react/src/components/Select/Select.tsx
huzaifabhutta: Hi @tay1orjones, removed the imperative selector as I realized it was not needed since we could get value through `selectedIndex`

packages/react/src/components/Select/Select.tsx
huzaifabhutta: Done

packages/react/src/components/Select/Select.tsx
```diff
@@ -255,6 +258,13 @@ const Select = React.forwardRef(function Select(
       size: 'mini',
     });
   }
+  useEffect(() => {
+    const selectElement = document.getElementById(
+      id
+    ) as HTMLSelectElement | null;
+    setTitle(selectElement?.options[selectElement?.selectedIndex]?.text || '');
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
tay1orjones, line 267: This needs to run on every render of the component? If so, I don't think it needs to be in a `useEffect`.  If the `select` has a `ref` on it, can we use the ref to get the selected index/element instead of using document.getElementById?

We have a `mergeRefs` helper function that can be used to combine the forwarded `ref` and a local ref inside the component for this use if it's needed.

tay1orjones:
Hey, thanks again for all your work here! It's been a couple months since there was movement here. In the interest of keeping the PR queue focused on actionable in progress items I'm going to close this PR for now.

If/when you're ready to pick this back up feel free to reopen this PR or open a new one.

ashishkrz:
@mHuzefa @tay1orjones Do we have any closure on same? Was waiting for this PR to get merged.

PR #15804, opened 2024-02-20
feat(StaticNotification): add actionButton props

finken2:
Following onto https://github.com/carbon-design-system/carbon/pull/15532 , this makes the new StaticNotification more visibly like the ActionableNotifcation.

#### Changelog

**New**

- Accept the `actionButtonLabel` on and `onActionButtonClicked` props that ActionableNotification has.
- Storybook entry for StaticNotification with an action, with longer text so users can see how it wraps and aligns.

**Changed**

- use the styles from actionableNotification instead of inlineNotification, so static is laid out like actionable inline variant.

#### Testing / Reviewing

- Discussed with @tay1orjones beforehand on Slack. 
- Implemented and tested locally in storybook.

image

Let me know if any changes are needed. I'm happy to iterate on it. We're currently blocked from picking up any carbon builds until we can get this whole flow of focus grabbing on Actionable resolved.


tay1orjones (approved):
Awesome, thank you!

alisonjoseph (approved):

guidari (approved):
LGTM!

PR #15792, opened 2024-02-20
test(Slug): add API, keyboard tests for `Slug`

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15273

Adds in tests for the `Slug` component, as well as adding tests for components with a `slug` prop

#### Changelog

**New**

- Keyboard interaction tests for `Slug`
- API tests for `Slug`
- API tests for components that have a `slug` prop

**Changed**

- `Toggletip` now passes `...rest` props into `Popover`
- changed instances of `cds` to `${prefix}` in component API tests

#### Testing / Reviewing

Ensure all tests pass and cover intended use cases

PR #15808, opened 2024-02-21
feat: added interactive tag

### LINKED ISSUE #13337: Interactive tag: Selectable Tag Implementation
## Acceptance criteria

- Add the Selectable tag variant to the Tag component.
- Make sure this works against Jan's Contextual Layout System.

-----

**New Tag variant**
- Add the Interactive tag umbrella under the existing Tag component
  - Add Selectable tag variant

**Renaming things in Storybook**
*Question: Will renaming these tags going to be a breaking change? 
- Rename "Default" tag to "Read-only" tag.
- Rename "Filterable" tag to "Dismissible" tag.

**Moving things around in Storybook**
- Can we have the file structure like this in the React storybook?
  - Overview
  - Read-only
  - Dismissible
  - Selectable
  - Operational
  - Skeleton
  - Playground

**Icons**
- Having the ability to add icons to tags in all variants (except for the small size).

**Sizes**
- Add the Large size (32px height) for the Read-only, Dismissible, Selectable, and Operational tag variants.
- Do not allow icons in the Small size (18px height) for the Read-only, Dismissible, Selectable, and Operational tag variants.
*Note: The AI slug in the small size is fine as is. It is just decorative icons we are not allowing in the small size.

**Colors**
- Selectable tags should only be available in the color of the design specs provided. It will not be allowed to be used in the other colors.

-----

### Selectable Tag Design Spec

Selectable tag

### LINKED ISSUE #15529: Interactive tag: Operational Tag Implementation
## Acceptance criteria

- Add the Operational tag variant to the Tag component.

----

**New Tag variants**
- Add the Interactive tag umbrella under the existing Tag component
  - Add Operational tag variant

**Colors**
- Operational tags should not be available in the `outline` or `high contrast` color options.

**New color token (tag component level)**
- Add a `$tag-border-[color]` token to be used for the Operational tag variant border colors.
- Note: This token will always be the 40 step color in the light themes, and 50 step color in the dark themes. Provided an example below.

Example for Gray tag:

`$tag-border-gray`
| White theme | Gray 10 theme | Gray 90 theme | Gray 100 theme |
|--------|--------|--------| --- |
| Gray 40 | Gray 40 | Gray 50 | Gray 50 |

----

guidari:
Closes #13337
Closes #15529

Added `InteractiveTag` under experimental.

#### Changelog

**New**

- Operational, Dismissible and Selectable tags 

**Changed**

- `Tag.tsx` is a primitive/root component for the interactive tags
- Changed `Default` story to `Read Only` in Tag stories

**Removed**

- `filter`, `onClose`, `title` props from `Tag` are marked as deprecated and will be removed in V12
- Removed @avt-keyboard-nav test for `Tag` with `filter` turned on

#### Testing / Reviewing

- Go to Experimental section in storybook and test the interaction with the tags


packages/react/src/__tests__/index-test.js
```diff
@@ -182,6 +183,7 @@ describe('Carbon Components React', () => {
         "Switcher",
         "SwitcherDivider",
         "SwitcherItem",
+        "TYPES",
```
tw15egan: Should these be showing up in the component export test?

packages/react/src/components/Tag/Tag.tsx
```diff
@@ -30,6 +31,12 @@ const TYPES = {
   outline: 'Outline',
 };
 
+export const SIZES = {
+  sm: 'sm',
+  md: 'md',
+  lg: 'lg',
+};
+
```
tw15egan, line 39: It seems like changing this to an `export` is causing the `TYPES` and `SIZES` to be exported. Do we need to just change `export * from './Tag';` in `index.js` to only export the `Tag` itself?

packages/react/src/components/Tag/Tag.tsx
```diff
@@ -99,7 +106,7 @@ const Tag = <T extends React.ElementType>({
   className,
   id,
   type,
-  filter,
+  filter, // remove filter in next major release
```
tw15egan:
```suggestion
  filter, // remove filter in next major release - V12
```

Might be good to tag it with `V12` so it is easier to find these comments when we start the V12 effort

packages/react/src/components/Tag/InteractiveTag.mdx
```diff
@@ -0,0 +1,86 @@
...
+[Usage guidelines](https://www.carbondesignsystem.com/components/tag/usage)
+&nbsp;|&nbsp;
+[Accessibility](https://www.carbondesignsystem.com/components/tag/accessibility)
+
+## Overview
+
+Multiple or single tags can be used to categorize items.
+
+Use short labels for easy scanning. Use two words only if necessary to describe
+the status and differentiate it from other tags.
```
tw15egan:
```suggestion
Tags can be used to categorize items. Use short labels for easy scanning. Use two words only if necessary to describe the status and differentiate it from other tags.
```

packages/react/src/components/Tag/InteractiveTag.mdx
```diff
@@ -0,0 +1,86 @@
...
+<DismissibleTag type="red" className="some-class" title="Clear Filter">
+  {'Tag content'}
+</DismissibleTag>
+```
+
+## Selectable
+
+Selectable tags are used to select or unselect single or multiple items in a
+form situation. Selectable tags can also be used in filtering scenarios by label
+functionality when they do not need a dimsdale functionality.
```
tw15egan:
```suggestion
Selectable tags are used to select or unselect single or multiple items. Selectable tags can also be used in filtering by label scenarios when they do not need to be dismissed.
```

packages/react/src/components/Tag/InteractiveTag.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+When to use:
+
+- Use selectable tags to toggle between a selected or unselected state.
+- Use when needing to filter without an explicit dismissable functionality. (We
```
tw15egan, line 50:
```suggestion
- Use when needing to filter without explicit dismissible functionality. (We
```

packages/react/src/components/Tag/InteractiveTag.mdx
```diff
@@ -0,0 +1,86 @@
...
+When to use:
+
+- Use to view a list of items with the same tag in a tagclip, popover, or
+  breadcrumb detail view.
+
+When not to use:
+
+- Do not use operational tags as a replacement for links that direct you to an
+  entirely different page or launch you out to another tab.
+- Do not use in combination with Dismissable tags. Instead, consider letting the
```
tw15egan, line 78:
```suggestion
- Do not use it in combination with `DismissibleTag`. Instead, consider letting the
```

packages/react/src/components/Tag/InteractiveTag.mdx
```diff
@@ -0,0 +1,86 @@
...
+Selectable tags are used to select or unselect single or multiple items in a
+form situation. Selectable tags can also be used in filtering scenarios by label
+functionality when they do not need a dimsdale functionality.
+
+When to use:
+
+- Use selectable tags to toggle between a selected or unselected state.
+- Use when needing to filter without an explicit dismissable functionality. (We
+  need to be a little more specific around filtering for selectable versus
+  dismissable tags).
```
tw15egan, line 52:
```suggestion
  dismissible tags).
```

packages/react/src/components/Tag/InteractiveTag.mdx
```diff
@@ -0,0 +1,86 @@
...
+```
+
+## Operational
+
+Operational tags enable the user to view a list of all items associated with a
+given tag in different ways.
+
+When to use:
+
+- Use to view a list of items with the same tag in a tagclip, popover, or
```
tw15egan: Is this supposed to be `Toggletip`?

packages/react/src/components/Tag/InteractiveTag.stories.js
```diff
@@ -0,0 +1,305 @@
...
+          justifyContent: 'flex-start',
+          marginTop: '0.5rem',
+        }}>
+        <Tooltip label="View more" align="bottom">
+          <OperationalTag className="some-class" title="Clear Filter">
+            {'Tag content'}
+          </OperationalTag>
+        </Tooltip>
+
+        <Toggletip>
```
tw15egan, line 177: The `Toggletip` wrapper seems to be adding an extra tab stop here

packages/react/src/components/Tag/OperationalTag.tsx
```diff
@@ -0,0 +1,166 @@
...
+  purple: 'Purple',
+  blue: 'Blue',
+  cyan: 'Cyan',
+  teal: 'Teal',
+  green: 'Green',
+  gray: 'Gray',
+  'cool-gray': 'Cool-Gray',
+  'warm-gray': 'Warm-Gray',
+};
+
```
tw15egan, line 30: Can these be imported from `Tag`?

packages/react/src/components/Tag/Tag-test.js
```diff
@@ -17,6 +20,27 @@ describe('Tag', () => {
...
+      );
+      await expect(container).toHaveNoAxeViolations();
+    });
+
+    it('should have no Axe violations SelectableTag', async () => {
+      const { container } = render(
+        <SelectableTag type="red">test-tag</SelectableTag>
+      );
+      await expect(container).toHaveNoAxeViolations();
+    });
```
tw15egan: Do we need these tests anymore since we now run the a11y-checker in the e2e AVT tests?

packages/react/src/components/Tag/InteractiveTag.stories.js
```diff
@@ -0,0 +1,305 @@
...
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+export const Selectable = () => {
+  return (
+    <>
+      <SelectableTag className="some-class" title="Clear Filter">
```
tw15egan: It seems like `SelectableTag` and `OperationalTag` have a `title` prop that doesn't seem to do anything, these can probably be removed right?

tw15egan (commented):
Looking great! Just left some comments

packages/react/src/components/Tag/InteractiveTag.stories.js
guidari: @tw15egan 
To fix that I would have to place a `tabIndex={-1}` 
```
<ToggletipButton label="Additional information" tabIndex={-1}>
```

But to do that I would have to add the `...rest` to `ToggletipButton`. Is that cool?

Screenshot 2024-02-23 at 13 48 34

tw15egan: I think that's fine, [I just had to add ](https://github.com/carbon-design-system/carbon/pull/15792/files#diff-aff3fc17c1be744759ed294411fed5beccce2582e3eb2c39565e7bdddfc3a30a)`...rest` to `Toggletip` as well to handle some of the `Slug` tests. Not sure why we didn't have that in the first place

packages/react/src/components/Tag/OperationalTag.tsx
guidari: If we import from Tag it will come the `outline` annd `high-contrast` also. That's why I created separate for OperationalTag

packages/react/src/components/Tag/InteractiveTag.stories.js
tw15egan: However, that brings up another issue in this scenario; you've got a button inside a button, which I believe is an a11y violation. Is there a way we could change what either the `ToggletipButton` or `OperationalTag` is rendered as using an `as` prop or something?

packages/react/src/components/Tag/OperationalTag.tsx
tw15egan: Makes sense, thanks for clarifying!

packages/react/src/components/Tag/InteractiveTag.stories.js
guidari: Humm, good observation. I'll check it out how to approach that.

guidari: I changed the `ToggleTipButton` code to use the `as` prop instead of the `OperationalTag` because I wanted to keep the styling for focus, hover, etc in the Tag.
The code would be like that to add the prop `as`. 

Screenshot 2024-02-23 at 14 31 47

I commited the changes. Let me know what you think

tw15egan: That makes sense, but now the `aria` roles seem invalid since it's no longer a button. This isn't necessarily a blocker for merging, but it might be worth creating a follow-up issue if using `Toggletip` to wrap the `OperationTag` is one of the intended use cases. The operational tag itself looks great 👍

guidari: Cool! I'll create a follow-up issue
https://github.com/carbon-design-system/carbon/issues/15820

laurenmrice (changes requested):
Looking good Gui! Just some things below:

### Tag (stable) review

**Read-only story**

- The Read-only tag variant does not get the `$tag-border-[color]` color token in the IBM color palette colors (red, magenta, purple, blue, cyan, teal, green, gray, cool gray, and warm gray. Only the Operational tag variant get this border token. 

**Overview story**

- The tag preview at the top of this story should reflect the changes mentioned above. ^

**Playground story**

- When turning the filter prop to `true`, the tag increases in height size. When turning the filter prop from false to true, it should stay the same height size. 

- The `Lg` size does not have the correct padding. Here are the specs below for a tag without icons, a tag with a decorative icon on the left, and a dismissible icon on the right. 

Specs


- The Gray, Warm gray, and Cool gray values for this token: $tag-color-[color] should be step 100 in the Light themes, and step 10 in the Dark themes. They are a little different than the other tag colors for contrast reasons.

-----

### Tag (experimental) review

**Dismissible tag story**

- The Dismissible tag does not get the `$tag-border-[color]` color token in the IBM color palette colors (red, magenta, purple, blue, cyan, teal, green, gray, cool gray, and warm gray. Only the Operational tags get this border token. 

- Once this is fixed above^ the focus state around the close x icon should be flush with the container of the tag instead of being inset. 

focus

**Operational tag story**

- The decorative icons on the left of the tag do not have the correct padding, and I suspect this might be happening in all tag variants. I can only see it here in the Md size right now. Can I see all of the height sizes of this tag with the decorative icon on the left? Should there be a playground story after the “Selectable” story so people can change the size and turn on and off icons? Here are what the specs should be for this below. 

Specs

- The Interactive example of tooltip/toggletip/popover/spacing is a little off. Here is an image for each with specs. 

tooltip

toggletip

popover

**Selectable tag story**
- The tag icon and text need to be color token `$text-primary` in the unselected state.

guidari:
Hey @laurenmrice 
I addressed all your comments on the last commit!
The stories are just for testing purpose, I would like to schedule a quick call so we can check the best way to insert the tags in the stories.

laurenmrice (changes requested):
Looking really good Gui! :star2: Just a couple more things:

### Tag (stable) review
- Looks great!

---

### Tag (experimental) review

**Overview story**
- The image/demo examples on this story per variant should be adjusted, once the other examples are updated.

**Dismissible tag story**
- The demo image for this story doesn’t seem to be updated yet.
- The large, medium, and small tags need a radius of exactly 1 rem (16px). This might help with the focus problem on the large dismissible tag that has an outline.
- The outline tag styles focus border should overlap the tag border instead of being inset.

focus

**Operational tag story**
- The demo image for this story doesn’t seem to be updated yet.
- The Interactive examples need to have 4px/.25rem space between the tag and the tip of the tooltip/toggletip/popover.

spacing

- For the popover Interactive example, the tags in the popover need to be the same height as the tag it was triggered from. So in this case, the tags in the popover need to be 24px (md) in height instead of the the 32px (lg).

**Selectable tag story**
- The demo image for this story doesn’t seem to be updated yet. Otherwise looks good!

alina-jacob (commented):
Looks good Lauren and Gui! 🚀

alina-jacob (approved):
Looks good Lauren and Gui! 🚀

laurenmrice (approved):
Looks great Gui! Thank you 🙏

e2e/components/InteractiveTag/InteractiveTag-test.avt.e2e.js
```diff
@@ -0,0 +1,108 @@
...
+      id: 'experimental-unstable-interactivetag--dismissible',
+      globals: {
+        theme: 'white',
+      },
+    });
+    await expect(page).toHaveNoACViolations('DismissibleTag');
+  });
+
+  // Testing being skipped because it is failing in the ToggleTip that operational it's using
+  test.skip('@avt-advanced-states OperationalTag', async ({ page }) => {
```
tw15egan: Do we still need to skip this test with the changes made to `ToggleTip`?

packages/styles/scss/components/popover/_popover.scss
```diff
@@ -125,6 +125,17 @@ $popover-caret-height: custom-property.get-var(
 
   // Popover content
   .#{$prefix}--popover-content {
+    @include layout.redefine-tokens(
```
tw15egan, line 129: Should we add a small note here that this is because we redefine the tokens in `Tag`, so we need to reset them to the defaults here?

e2e/components/InteractiveTag/InteractiveTag-test.avt.e2e.js
guidari: Good catch! I just removed @tw15egan

tw15egan (approved):
LGTM 👍 ✅

alisonjoseph (approved):
🚀🔥Looks great!

wkeese:
Hi everyone,

This PR deprecated `<Tag>`'s `filter` and `onClose` attributes, yet DismissibleTag isn't available through the top-level index.js file.  Nor are any of the other new components (like InteractiveTag).  Can you add them?

PR #15819, opened 2024-02-23
Added PolymorphicProps to `Link`

### LINKED ISSUE #15338: Fix `as` type in `Link`
https://github.com/carbon-design-system/carbon/pull/15233 included an initial type for the `as` prop, but it's incorrect. It should use the PolymorphicProps

https://github.com/carbon-design-system/carbon/blob/95d40f7272a71033f058a881327fee42b2ab6842/packages/react/src/types/common.ts#L13-L20

For reference, this is used for the `UIShell/Link`:
https://github.com/carbon-design-system/carbon/blob/6df76910a77deb55f424d125eca40ff43721e3d8/packages/react/src/components/UIShell/Link.tsx#L30-L33

This approach can be modeled to be added to the base `Link`.

In the meantime the only workaround is to `//ts-ignore` the error that results from using the `as` prop

guidari:
Closes #15338

Added PolymorphicProps props to `Link` component. It had to changed the types previously wrriten to match the PolymorphicProps needs.

#### Testing / Reviewing

Check if the change make sense and it won't broke any type.


tay1orjones (approved):

tay1orjones:
Not sure why the deploy preview comment wasn't made here, but I just updated this with latest from main and it [triggered a build](https://app.netlify.com/teams/carbon-design-system/builds/65dcaf37d516c800081c8c66).

tay1orjones (approved):
Pending the CI checks I think this looks okay

andreancardona (approved):

PR #15828, opened 2024-02-27
fix: 15726 Playwright tag convention

### LINKED ISSUE #15726: [a11y]: Playwright tag convention
Check every file with this path `*-test.avt.e2e.js` and match the rule written is this [PR](https://github.com/carbon-design-system/carbon/pull/15692) for the `describe` and `test` property from Playwright.

The `test` property was already corvered before, so you probably want to correct only the `describe`

preetibansalui:
Closes #15726 

Check every file with this path *-test.avt.e2e.js and match the rule written for the describe and test property from Playwright.

**Changed**

#### Testing / Reviewing

Check every file with this path *-test.avt.e2e.js and match the rule written for the describe and test property from Playwright.


guidari (approved):
LGTM! 🚀

e2e/components/Breadcrumb/Breadcrumb-test.avt.e2e.js
```diff
@@ -10,7 +10,7 @@
 import { expect, test } from '@playwright/test';
 import { visitStory } from '../../test-utils/storybook';
 
-test.describe('breadcrumb @avt', () => {
+test.describe('@avt breadcrumb', () => {
```
andreancardona:
```suggestion
test.describe('@avt Breadcrumb', () => {
```

andreancardona (changes requested):
super small thing, otherwise looks great!

e2e/components/Breadcrumb/Breadcrumb-test.avt.e2e.js
preetibansalui: Hey @andreancardona , I believe the first line where you are suggesting the change, its showing it as a removed line, can you please take a look again. Thanks!

andreancardona (approved):

PR #15843, opened 2024-02-28
feat: add fluid button set

lee-chase:
The documentation on buttons https://carbondesignsystem.com/components/button/usage/
mentions fluid buttons and how they're handled. However, this does not appear to be implemented. 

This PR adds the `fluid` property to the `ButtonSet` component. 

With the `fluid` property set to true the button set exhibits the following behavior
- Controls the inline-size of the buttons
- Auto stacks the buttons based on `$min-inline-button-size: convert.to-rem(176px)`
- Re-orders buttons based on guidance
- Adds `withDisplayBox` to allow control of container width in storybook
- Adds `SetOfButonsFluid` story

Further work
- Setting a default button set in the story is not functional. NOTE: A default not functional elsewhere e.g. MenuButton Playground MenuAlignment.
- Notify users (via event handler) of stacking change.
- Allow for gaps between buttons
- Control auto stack with a property.

#### Changelog

**New**

- packages/react/.storybook/templates/WithDisplayBox/WithDisplayBox.scss
- packages/react/.storybook/templates/WithDisplayBox/index.js
- packages/react/src/components/Button/__story__/fluid-button-set-args.js

**Changed**

- packages/react/src/components/Button/Button.stories.js
- packages/react/src/components/ButtonSet/ButtonSet.tsx
- packages/styles/scss/components/button/_button.scss

#### Testing / Reviewing

Checked functional in storybook


tw15egan:
This is great, @lee-chase. Thanks for taking the time to add this in. I had a question: Should the single button max out at `320px`? It seems to grow until `520px` and then becomes a non-fluid variant, but I'm just trying to determine the correct spec. Similarly, what should they grow to in a 2, 3, or 4-button group? The 4 button group has them all stack at a `700px` width

thyhmdo:
Hi, sorry for the late review. This looks great. I have a few questions and concerns:

- This experience is different from the [Set of Buttons](https://react.carbondesignsystem.com/?path=/story/components-button--set-of-buttons). I was confused for 5 sec not knowing what to do. It could populate an option upfront. 
- From interacting with the "Stacked" prop, I'm not sure I see differences, unless I'm missing something.  
- @lee-chase Can the visual change here if I provide a spec? I was a little bit confused about how this worked since we don't have this experience anywhere in the storybook.

lee-chase:
@thyhmdo apologies for the lack of clarity. 

- I did try top populate this up front, but had some difficulty with Storybook, which seems to behave differently to IBM Products. Perhaps one of the devs could advise, happy to update. The simplest thing to do would be to move the various fluid button options into a folder.
- Stacked - if I'm honest I do not recall why I left this in this story. I have removed it and can see no issues in doing so.
- Not sure what you are asking with your final query.

The wrapper around the button story is something done in IBM Products and could probably do with a UX tidy up. The intent is to show the component but provide a simple mechanism to allow the user to see how the component behaves with different container sizes.
image

thyhmdo:
> @thyhmdo apologies for the lack of clarity.
> 
> * I did try top populate this up front, but had some difficulty with Storybook, which seems to behave differently to IBM Products. Perhaps one of the devs could advise, happy to update. The simplest thing to do would be to move the various fluid button options into a folder.
> * Stacked - if I'm honest I do not recall why I left this in this story. I have removed it and can see no issues in doing so.
> * Not sure what you are asking with your final query.
> 
> The wrapper around the button story is something done in IBM Products and could probably do with a UX tidy up. The intent is to show the component but provide a simple mechanism to allow the user to see how the component behaves with different container sizes. image

1. @tw15egan @guidari Do you have suggestions for this? 
2. Nice! I checked and it seems like the problem is solved.
3. Similar to components with layers, Anna provided some specs to have a particular look for it. In this visual, the label is quite long, the width extension could be replaced with something else that is close to our design spec (see image-red line) 
 - The label could be simplified, maybe something like: "Max width (component container)
 - The other label could be "Width varies based on the max width adjustment" 
image

guidari:
Hey @lee-chase 
I did a commit on your PR to populate an option up front like @thyhmdo mentioned. I guess that what you guys meant, right?

thyhmdo:
> Hey @lee-chase I did a commit on your PR to populate an option up front like @thyhmdo mentioned.
> 
> https://deploy-preview-15843--v11-carbon-react.netlify.app/?path=/story/components-button--set-of-buttons-fluid

that looks good! thanks @guidari

guidari (dismissed):
Tested and that looks good to me. 🚀

lee-chase:
@andreancardona I do not believe this is waiting for response from me.

tay1orjones:
Hey, thanks again for your work here! Had a brief discussion with the team about any blockers we see preventing this from merging.

One thing that came up was the container slider. We've had users struggle to understand and use controls inside stories like this. We'd like to move away from them where possible. One idea is to instead have a new custom toolbar item in the storybook toolbar. It would contain a list of px values stepping in multiples of 160px. `160px`, `320`, `480`, ... up to `1600`. The graph bar can be used for the icon. Selecting one of these items would modify the styles on a decorator that wraps each story to set the width of that container to the selected value. Benefits here are that it standardizes controls within the storybook toolbar outside of the story frame, and it fully parameterizes these container sizes into the url. With this we could vrt components at one/many/all of these container widths. 

On the code level I worry about determining if the buttons are currently stacked by imperatively evaluating the styles applied via `getComputedStyle()`. Is there another way we could determine this? It's not testable functionality as is because getComputedStyle() is a noop.

https://github.com/carbon-design-system/carbon/blob/184d1a9d72b6941c70d4804166371d85b3e3896f/config/jest-config-carbon/setup/setup.js#L33-L35

Sorting the buttons feels like something we would typically leave up to the consumer. I see the intended benefit, but using Children methods immediately limits the structure of what consumers can provide there. I'm not sure the benefits outweigh that cost?

lee-chase:
@tay1orjones & team

Switching from the container slider to a toolbar item is a possibility, but would need visual input as to how that is represented in the story. The viewport control has a similar behaviour and clearly marks out the usable area.

Whether or not the buttons are stacked is currently used to reposition box shadows and sorting. Even if sorting is deferred to the user they would need to monitor when stacking had occurred. In an ideal world we would provide `onStack` as an event.  Could we not just mock window.getComputedStyle for some tests?

If there is no possibility of a tooltip (currently icon button only) then `--stacked` css modifier can be simplified and replaced with an `overflow: hidden` on the button set. This does not solve the sort/stacked detection.

lee-chase:
@guidari any further work you would like on this?

AlexanderMelox:
Going to work on a feature and would love to have this in for use 🙂

packages/react/src/components/Button/Button.stories.js
```diff
@@ -182,3 +188,44 @@ export const Skeleton = () => {
     </div>
   );
 };
+
+export const SetOfButtonsFluid = {
```
tay1orjones: Instead of this I think we should create a new stories file at `packages/react/src/components/ButtonSet/ButtonSet.stories.js`

We should be able to still have it within the `Button` section of the storybook left nav, but as a new folder, by specifying `title: 'Components/Button/Set of buttons',` in the default export. The stories in the folder/file could be named
* With fixed buttons
* With fluid buttons

The fixed/fluid naming is how it's referenced in the website [guidance](https://carbondesignsystem.com/components/button/usage/#:~:text=Fixed%20versus%20fluid%20buttons)

packages/react/src/components/ButtonSet/ButtonSet.tsx
```diff
@@ -1,36 +1,127 @@
...
+      const resizeObserver = new ResizeObserver(() => {
+        setIsStacked(checkStacking());
+      });
+      resizeObserver.observe(fluidInnerRef.current);
+
+      return () => resizeObserver.disconnect();
+    }, [isStacked, stacked]);
+
+    useEffect(() => {
+      const newSortedChildren = React.Children.toArray(children);
```
tay1orjones: Would it make sense to move this out of the effect and instead make it the return value of a `useMemo()` with `children` as a dep?

packages/react/src/components/ButtonSet/ButtonSet.tsx
```diff
@@ -1,36 +1,127 @@
...
+};
+
+const buttonOrder = (kind) =>
+  ({
+    ghost: 1,
+    'danger--ghost': 2,
+    tertiary: 3,
+    danger: 5,
+    primary: 6,
+  })[kind] ?? 4;
```
tay1orjones, line 35: Where does this order come from? @carbon-design-system/design where in the website docs is this specified?

packages/react/src/components/ButtonSet/ButtonSet.tsx
```diff
@@ -1,36 +1,127 @@
...
+
+    useEffect(() => {
+      const newSortedChildren = React.Children.toArray(children);
+      newSortedChildren.sort(
+        (a: any, b: any) =>
+          (buttonOrder(a.props.kind || 'primary') -
+            buttonOrder(b.props.kind || 'primary')) *
+          (isStacked ? -1 : 1)
+      );
+      setSortedChildren(newSortedChildren);
```
tay1orjones: When this changes we need to make sure focus is handled properly and isn't lost or moved around. Are there any other a11y considerations for reordering the dom like this?

packages/react/src/components/ButtonSet/ButtonSet.tsx
```diff
@@ -1,36 +1,127 @@
...
+      };
+
+      /* initial value not dependant on observer */
+      setIsStacked(checkStacking());
+
+      if (!fluidInnerRef.current) {
+        return;
+      }
+
+      const resizeObserver = new ResizeObserver(() => {
```
tay1orjones: Can we reuse something from within the new utilities package? It'd be nice if we could (re)use a single ResizeObserver even when stacked or isStacked changes, for instance.

packages/styles/scss/components/button/_button.scss
```diff
@@ -21,6 +22,98 @@
...
+    max-inline-size: none;
+  }
+
+  .#{$prefix}--btn--ghost {
+    padding-inline-start: $spacing-05;
+  }
+}
+
+@mixin button-set-fluid-layout() {
+  container-name: button-set;
```
tay1orjones: Is this one necessary? It's not used with a `@container`.

packages/styles/scss/components/button/_button.scss
```diff
@@ -21,6 +22,98 @@
...
+
+    display: flex;
+    flex-direction: var(--flex-direction);
+    align-items: stretch;
+    justify-content: flex-end;
+    inline-size: 100%;
+  }
+
+  /* NOTE: Container queries are based on this value */
+  $min-inline-button-size: convert.to-rem(176px);
```
tay1orjones: Is this defined in the spec, or is it derived from just the example buttons? Does it need to be configurable? Is there a non-magic number we could use instead?

packages/styles/scss/components/button/_button.scss
tay1orjones, line 1: These styles were pretty difficult for me to understand - took me a bit to wrap my head around everything. Do you think it would make sense to reduce duplication by specifying a `$max-buttons: 4;` to then use as the basis for a loop to emit the styles for 1 button, 2 buttons, 3, 4? This would also benefit in that we could in theory support more than 4 in the future by just incrementing that variable.

packages/styles/scss/components/button/_button.scss
```diff
@@ -21,6 +22,98 @@
...
+      }
+    }
+  }
+
+  @container (width <= #{3 * $min-inline-button-size}) {
+    .#{$prefix}--btn-set__fluid-inner--triple .#{$prefix}--btn {
+      flex: 1 1 math.div(100%, 3);
+    }
+
+    // Note that by 640px(40rem) sidepanel expects 1 row
```
tay1orjones: Could you expand on this wrt sidepanel?

packages/react/src/components/ButtonSet/ButtonSet.tsx
```diff
@@ -1,36 +1,127 @@
...
   stacked?: boolean;
 }
 
+const fluidMultiples = ['none', 'single', 'double', 'triple', 'quadruple'];
+const fluidMultiple = (buttonArray) => {
+  if (buttonArray.length >= fluidMultiples.length) {
+    // too many buttons
+    return 'many';
+  } else {
+    return fluidMultiples[buttonArray.length];
```
tay1orjones: Is there an opportunity to avoid this altogether by instead determining the children count [in the styles directly with `:has()`](https://www.matuzo.at/blog/2022/counting-children)?

packages/react/src/components/Button/Button.stories.js
```diff
@@ -182,3 +188,44 @@ export const Skeleton = () => {
...
+  },
+  render: ({ ...rest }) => {
+    const buttons = rest['Fluid Buttons'];
+
+    if (!buttons || buttons === 0) {
+      return <div>Select one or more buttons.</div>;
+    }
+
+    return (
+      <WithDisplayBox>
```
tay1orjones: I don't think we want to go this route unfortunately. It didn't pass a visual design review and it's not easily testable with VRT.

Maybe we talked about this before, but I think a decent alternative is having a `containerWidth` control that explicitly states in the description that it's just for testing in storybook. 

We can then set this control via args in VRT tests to ensure this remains consistent at different sizes with different button counts. The other controls that aren't part of the component api could also have this note in their description.

It could use the Annotation decorator to state the current width and would provide a visual box to indicate to the user how to use `containerWidth`

tay1orjones (commented):
I took a fresh look and have some thoughts. Let me know what you think.

This will also need:
- Unit tests
- VRT test(s)
- AVT test(s), maybe?
- Docs in Button.mdx
- Code comments throughout that help explain to someone the basic flow of logic - Why/what is `newIsStacked()` called, container scss calculations, where buttonOrder comes from, etc.

packages/react/src/components/ButtonSet/ButtonSet.tsx
lee-chase: This order is defined by Carbon for IBM Products action set which is used in the full page flows, SidePanel, Tearsheet, Flyout.

packages/react/src/components/ButtonSet/ButtonSet.tsx
lee-chase: Will need a test, but React seems to get this correct in storybook. Tested by changing the container width on a timer.

packages/react/src/components/ButtonSet/ButtonSet.tsx
lee-chase: Other places seem to currently declare their own.
image

packages/styles/scss/components/button/_button.scss
lee-chase: In IBM Products we get

Component | Size | Buttons
---|---|---
Side panel | xs (256px) | Always stacks
" | sm (320px) | Always stacks
" | md (480px) | Stacks at 3
" | lg (640px) | never stacks
Tearsheet | narrow(540px) | never stacks (up to 4 buttons)

So min size applicable to all is 540px/4 = 135px
Smallest size used by SidePanel 481px/3 = 160.33px

So we might need to be configurable, although I do not believe any patterns use 4 buttons and 540px/3 is 180px. Not seeing any tests in IBM Products for this either.

Mid breakpoint/4 is 168px.

I think 176px was initially picked as a ball park number used elsewhere by Carbon (dropdown, listbox, tag, tooltip)

If supporting 4 on Tearsheet then this needs to be configurable if stacking 3 on size md side panel.

packages/styles/scss/components/button/_button.scss
lee-chase: Removed this line in simplification. That said, the large side panel, which is 640px wide, shows 3 buttons without stacking them.

packages/styles/scss/components/button/_button.scss
lee-chase: Refactored and expandable.

packages/react/src/components/Button/Button.stories.js
lee-chase: Moved to a story control, also included a control to make the container visible.

lee-chase:
@tay1orjones got nudged regarding this a couple of days back by with a desire to use it in an AI Chat workflow. 

I think I addressed the points, only outstanding is the magic number 176px, with mine being the 6th time this number has appeared in Carbon styles. It probably should be configurable, or at least adjustable for the IBM Products Narrow Tearsheet with 4 buttons. This could be added in if Tearsheet decides to adopt.

tay1orjones (approved):
LGTM!

lee-chase:
@guidari made some updates following comments from @tay1orjones and he's now approved if you'd lie to re-approve.

2nikhiltom (approved):
Thanks for doing this @lee-chase  ! There are a few updates we can make, but we can handle those in a separate PR.

laurenmrice:
Thanks so much for contributing this @lee-chase. We are going to merge this through with a label of experimental or preview to close this PR out, which @tay1orjones is going to add shortly. And I will do a more comprehensive visual review then. Any visual bugs we find, we will then fix in another PR.

alisonjoseph (approved):

laurenmrice (approved):
Going to do a separate visual review for this in [this issue](https://github.com/carbon-design-system/carbon/issues/21155).

PR #15846, opened 2024-02-29
fix: active states and focus states of inline and standalone links

### LINKED ISSUE #15613: [Link] Adding an underline on the focus state and active state
**Ref issue**: https://github.com/carbon-design-system/carbon-website/issues/3537

**PROBLEM**
The focus state of the [Standalone Link](https://react.carbondesignsystem.com/?path=/story/components-link--playground)  needs to have an underline. There's only padding to differentiate it from the focus state of the Ghost button. 

### Standalone link (default and focus state)
image

image

### Ghost button
image

image

----

**INCONSISTENCY**
Active state:  When you hold the click over the [inline link](https://carbondesignsystem.com/guidelines/2x-grid/usage), there's no underline. But when you hold the click on the [Standalone link](https://react.carbondesignsystem.com/?path=/story/components-link--paired-with-icon), there's an underline

------

**SOLUTION**

- [x] Need to add design specs here - Need a review from @laurenmrice 

4_3_608x456px

riddhybansal:
Closes #15613 

Added an underline on the focus state and active state

**Changed**

- Changed some styles to incorporate underline at different states

#### Testing / Reviewing
- Test the default and inline link component in active and focus states it should have underline.

### DIFF
--- a/packages/styles/scss/components/link/_link.scss
+++ b/packages/styles/scss/components/link/_link.scss
@@ -46,7 +46,10 @@ $link-focus-text-color: custom-property.get-var(
     &:active,
     &:active:visited,
     &:active:visited:hover {
+      @include focus-outline;
+
       color: $text-primary;
+      outline-color: $link-focus-text-color;
       text-decoration: underline;
     }
 
--- a/packages/styles/scss/components/link/_link.scss
+++ b/packages/styles/scss/components/link/_link.scss
@@ -54,6 +54,7 @@ $link-focus-text-color: custom-property.get-var(
       @include focus-outline;
 
       outline-color: $link-focus-text-color;
+      text-decoration: underline;
     }
 
     &:visited {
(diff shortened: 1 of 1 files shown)


tw15egan (approved):
LGTM 👍 ✅

laurenmrice (commented):
Looks great Riddhi! Just one thing:

**Focus state**
In Firefox and Safari I can not get the Focus state on the actual Link in the example/demo on any of the stories. I could get it in Chrome, and it looks good! Just was wondering if that is a bug or if you have to operate things differently in those browsers to get the Focus on the link.

guidari:
Seems like it is a problem on MacOS and FF.
I used this solution here from [StackOverflow](https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox) and worked in FF, but not for Safari

guidari (approved):
I got it working on Safari.
Go to Safari Settings -> Advanced
Enable the option: _**Press Tab to highlight each item on a webpage page**_

Screenshot 2024-03-01 at 13 36 53

I'll approve, since it is something from the browsers

riddhybansal:
@laurenmrice Its working on Safari as well , some settings needs to be changed.

PR #15844, opened 2024-02-28
chore: onboard remaining packages to IBM Telemetry 🚀

mattrosno:
## Description

Adds the config file and dependency necessary to start tracking telemetry data via [IBM Telemetry](https://github.com/ibm-telemetry/telemetry-js) for the packages:

- `@carbon/cli`
- `@carbon/colors`
- `@carbon/elements`
- `@carbon/feature-flags`
- `@carbon/grid`
- `@carbon/icon-helpers`
- `@carbon/icons`
- `@carbon/icons-react`
- `@carbon/icons-vue`
- `@carbon/layout`
- `@carbon/motion`
- `@carbon/pictograms`
- `@carbon/pictograms-react`
- `@carbon/styles`
- `@carbon/themes`
- `@carbon/type`
- `@carbon/upgrade`

The `jsx` config for the two `-react` packages were generated by running the [JS config generator](https://www.npmjs.com/package/@ibm/telemetry-js-config-generator) in the icon build helper package.

**NOTE:** This separate PR needs reviewed on the v10 branch with a new release after the PR is merged: https://github.com/carbon-design-system/carbon/pull/15852

#### Changelog

For each package listed above:

**New**

- Adds @ibm/telemetry-js dependency
- Adds telemetry.yml config file
- Adds readme telemetry notice
- Adds postinstall script to use the telemetry command
- Adds telemetry.yml as an exported file if the package specifies `files`

**Changed**

- N/A

**Removed**

- `@carbon/telemetry` in `@carbon/icons-react` and `@carbon/pictograms-react`

#### Testing / Reviewing

Please look through package.json config files and ensure all necessary modifications have been made so that the "telemetry.yml" config file is included in the release version of each package.

**PLEASE NOTE:** In order for IBM Telemetry to start collecting data for this project, a new build must be published including these changes.

PR #15852, opened 2024-02-29
chore: onboard remaining v10 packages to IBM Telemetry 🚀

mattrosno:
This PR's description applies to this PR too: https://github.com/carbon-design-system/carbon/pull/15844

The only differences:

- `@carbon/import-once` is instrumented here in v10, but no longer exists in v11 on main branch
- The JSX config generated by [this package](https://www.npmjs.com/package/@ibm/telemetry-js-config-generator) resulted in a couple differences with the v11 package due to some props being added in v11

Besides that, everything else is the same with a package's ID remaining the same between its v10 and v11 version so we have data continuity across major versions.


francinelucca (approved):
🚀

alisonjoseph (approved):

mattrosno:
@alisonjoseph thanks for reviewing and merging! As this is the v10 branch, do I need to request that this gets manually published or when should we expect that to happen?

alisonjoseph:
FYI @tw15egan / @guidari we will need a v10 release this sprint. 

@mattrosno we've adjusted our [release schedule](https://github.com/carbon-design-system/carbon/wiki/Release-radar) slightly, next release is March 14th

PR #15856, opened 2024-03-01
fix(StaticNotification): set describedby on action button

finken2:
Follow-on to [a previous PR](https://github.com/carbon-design-system/carbon/pull/15804) that added the actionButton capability to StaticNotification. I didn't realize that the aria-describedby was being enforced on the button in addition to the body. My storybook example didn't catch that, b/c the current`useInteractiveChildrenNeedDescription` is only checking the **first** interactive child.  So when i went to use it with just the button, kaboom.

Fully tested it out locally in our microservice, recreating the issue and fixing it. 
I also changed the proptype of `subtitle` to Node to match actionableNotification, making the swap easier.

#### Changelog

**New**

- pass titleId to the NotificationButton as `aria-describedby`. No new prop for this.
- additional storybook story for the action button only

**Changed**

- updated `subtitle` to be PropTypes.Node to match ActionableNotification. 

#### Testing / Reviewing

render a StaticNotification within live code with action button and embedded links.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -7239,7 +7239,7 @@ Map {
         "type": "string",
       },
       "subtitle": Object {
-        "type": "string",
+        "type": "node",
       },
       "title": Object {
         "type": "string",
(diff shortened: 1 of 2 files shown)


finken2:
Not picking up my 2nd commit, so going to recreate this PR....

PR #15858, opened 2024-03-01
fix(StaticNotification): set describedby on action button

finken2:
Follow-on to [a previous PR](https://github.com/carbon-design-system/carbon/pull/15804) that added the actionButton capability to StaticNotification. I didn't realize that the aria-describedby was being enforced on the button in addition to the body. My storybook example didn't catch that, b/c the current `useInteractiveChildrenNeedDescription` is only checking the **first** interactive child.  (We should also discuss whether that hook should be updated, and if we need even more functionality when there's multiple actions)

Fully tested it out locally in our microservice, recreating the issue and fixing it. 
I also changed the proptype of `subtitle` to Node to match actionableNotification, making the swap easier.

#### Changelog

**New**

- pass titleId to the NotificationButton as `aria-describedby`. No new prop for this.
- additional storybook story for the action button only

**Changed**

- updated `subtitle` to be PropTypes.Node to match ActionableNotification. 

#### Testing / Reviewing

Render a StaticNotification with action button alone, and links & action together.


tay1orjones (approved):
Thanks for updating this!

tw15egan (approved):

PR #15868, opened 2024-03-01
fix(wrapFocus): add experimental option to wrap focus without sentinels

tay1orjones:
Closes issue https://github.com/carbon-design-system/carbon/issues/13888

This adds a new implementation of focus wrap behavior that does not use sentinel nodes. We placed it behind a new feature flag, `enable-experimental-focus-wrap-without-sentinels`.

A key difference in the new implementation is that it relies on being called from a handler that occurs before focus moves, like `onKeyDown`.

#### Changelog

**New**

- add new helper function `wrapFocusWithoutSentinels()`
- add `enable-experimental-focus-wrap-without-sentinels` flag, off by default

**Changed**

- update ComposedModal, Modal, ActionableNotification, and OverflowMenu (FloatingMenu) to use `wrapFocusWithoutSentinels` when the feature flag is on
- new documentation in storybook, feature flag stories for each impacted component

#### Testing / Reviewing

- Look at the feature flag stories and check to see that 1. focus sentinels are not rendered, and 2. focus wrap behaves as expected


alisonjoseph (approved):

tw15egan (approved):
Looks great!

PR #15879, opened 2024-03-04
fix(fileuploaderitem): allow configurable className

tay1orjones:
[Surfaced in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1709148698522979). For some reason `className` wasn't being applied in the `cx()` call, instead relying on the prop spread on the `span`. So if you passed in a `className`, all the built in classes wouldn't be applied.

Here's a repro of the bug: https://stackblitz.com/edit/github-svhol5?file=src%2FApp.jsx

Is this intentional? I can't find a related PR or anything, but maybe I missed something 🤔 

#### Changelog

**Changed**

- use `className` in addition to the internal classnames

#### Testing / Reviewing

- If you pass in a className, the default styles should still be applied.


alisonjoseph (approved):
Oh weird, definitely seems like a bug and not intentional. 👍

tw15egan (approved):

PR #15881, opened 2024-03-04
feat: unstable side panel building on modal (proposal / talking point)

lee-chase:
An experiment to see what a SidePanel Modal might look like as an extended version of ComposedModal.

Following on from the AI work to create Web Component versions of @carbon/ibm-products/SidePanel and Tearsheet it appeared that SidePanel is similar to a ComposedModal in many ways. 

This PR does not implement the full behavior of the IBM Products side-panel, but could be extended to do so. It is intended to spark a discussion about where the base SidePanel component should live and what it should look like.

Guidance on usage of SidePanel https://pages.github.ibm.com/cdai-design/pal/components/side-panel/usage

#### Changelog

**New**

- SidePanel components in React and Styles


tay1orjones:
Thanks for opening these up! It will be great to have these to ground conversations as we discuss if these will be supported in Carbon. I think leaving them as drafts for now makes sense.

Related: 
* https://github.com/carbon-design-system/carbon/pull/15881
* https://github.com/carbon-design-system/carbon/pull/15882
* https://github.com/carbon-design-system/carbon/pull/15891
* https://github.com/carbon-design-system/carbon/pull/15926

tay1orjones:

PR #15882, opened 2024-03-04
feat: side panel as modal kind (proposal / talking point) 

lee-chase:
An experiment to see what a SidePanel Modal might look like as a stylised version of the ComposedModal.

Following on from the AI work to create Web Component versions of @carbon/ibm-products/SidePanel and Tearsheet it appeared that SidePanel is similar to a ComposedModal in many ways. 

This PR adds a single property `kind` to the Composed modal and uses styling SCSS to achieve most of the implementation. It is intended to spark a discussion about where the base SidePanel component should live and what it should look like.

Guidance on usage of SidePanel https://pages.github.ibm.com/cdai-design/pal/components/side-panel/usage

#### Changelog

**Changed**

- ComposedModal properties and additional styling


tay1orjones:

PR #15886, opened 2024-03-04
fix: remove title from tag component

### LINKED ISSUE #15885: [a11y]: tag being interpreted as a group by voiceover
### Package

@carbon/react

### Browser

_No response_

### Operating System

_No response_

### Package version

11.52

### React version

_No response_

voiceover

### Assistive technology

voiceover

### Description

voiceover is describing tags as groups. this issue was originally brought up in the [carbon for ibm products repo](https://github.com/carbon-design-system/ibm-products/issues/4406).

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tag--default

### Steps to reproduce

view the [tag story in storybook](https://react.carbondesignsystem.com/?path=/story/components-tag--default), activate voiceover, and try to navigate from one tag to the next.

### Code of Conduct

davidmenendez:
Closes #15885 

this address an accessibility issue with voiceover where tag's where being described as "groups". the underlying problem appears to be usage of a title attribute in the tag `span`. double checking the [title documentation in MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) i don't believe a title is necessary here and removing it alleviates the voiceover problem.

i was able to verify this change by running voiceover locally and verifying that this change wasn't a new violation in the equal access checker.


tay1orjones (approved):
A long time ago this was added so that if a user has a tag with lots of words and they set text-overflow:ellipsis on it, the title would show on hover to reveal the full text.


The design guidance now states that tags should be one or two words at most, so I think this isn't so much a concern anymore. Additionally I think using a Tooltip in that scenario would be preferred anyway?

All this to say I think the historical reason why this was initially added no longer applies. Let me know if I'm wrong!

tw15egan (approved):
LGTM 👍 ✅

PR #15887, opened 2024-03-05
`Tooltip` & `DefinitionTooltip` changed storybook and functionality

### LINKED ISSUE #12921: [a11y]: Tooltip examples and behaviour need to be updated
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-tooltip--default

### React version

_No response_

n/a

### Assistive technology

_No response_

### Description

The interaction of various tip examples is inconsistent and contradictory between tooltip and toggletip..
~1) The Live demo example on the [usage page](https://carbondesignsystem.com/components/tooltip/usage#live-demo) makes no sense. When the focus lands on the i icon, nothing happens. Then I tab again, and focus remains on the 'i' and suddenly it has a tip. Focus should immediately trigger thte tip to appear.~
~2) In the same Live demo section, changing the Variant selector to be definition tooltip also shows an implementation problem. The definition never appears on focus/hover but instead appears exclusively on click/activate. It is implemented as a TOGGLETIP instead of as a tooltip.~
3) In storybook, the definition tooltip is under its own "D" heading instead of being under tooltip. It is less likely to be discovered there, and should be under tooltip, as is the case with variants of other components
4) Further the definition tooltip is again implemented as a toggletip instead of a tooltip. The difference in behaviour makes no  sense. As mentioned in https://github.com/carbon-design-system/carbon-website/issues/3332, the definition tooltip should always expose on hover/focus.
5) The [playground](https://react.carbondesignsystem.com/?path=/story/components-tooltip--playground) for tooltip has the tooltip appearing before it has focus. Same for the [playground for the definition tooltip](https://react.carbondesignsystem.com/?path=/story/components-definitiontooltip--playground). This should obviously not be happening
6) As mentioned in the website issue, all the uses of the information icon ('i') for tooltips should be expunged from the storybook and from the Figma kits
image
7) A challenge for Carbon is that while the tooltip component has a requirement for text, the icon-only button does not, nor does the ghost button, yet they both need a tooltip to provide context.  So guidance in both Carbon and storybook and in Figman's buttons should be revamped so that the label/name attribute is populated -- except in some very limited cases where there is a reason that the tip should _not_ be exposed on focus/hover.

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/carbon-website/issues/3712
- [ ] [Tooltip a11y]: Playground implementation
- [ ] https://github.com/carbon-design-system/carbon-design-kit/issues/721
```

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://carbondesignsystem.com/components/tooltip/usage#live-demo

### Steps to reproduce

Go to both the live demo and the storybook examples
- https://carbondesignsystem.com/components/tooltip/usage#live-demo
- https://react.carbondesignsystem.com/?path=/story/components-definitiontooltip--default
- https://react.carbondesignsystem.com/?path=/story/components-tooltip--default

Using a keyboard
1. attempt to have tooltips display on focus (tab to them without having to activate)
2. attempt to dismiss using Esc

### Code of Conduct

guidari:
Closes #12921

There is a couple changes made in storybook and also in the functionality. Now the `Tooltip` and `DefinitionTooltip` have similar behaviour, except for the onClick on `DefinitionTooltip`.

#### Changelog

- Changed defaultOpen to false on `Tolltip` Playground
- Add onFocus to `DefinitionTooltip`

#### Testing / Reviewing

- Test the `Tooltip` and `DefinitionTolltip`

### DIFF
--- a/packages/react/src/components/Tooltip/DefinitionTooltip.tsx
+++ b/packages/react/src/components/Tooltip/DefinitionTooltip.tsx
@@ -105,7 +105,7 @@ const DefinitionTooltip: React.FC<DefinitionTooltipProps> = ({
         openOnHover ? setOpen(true) : null;
       }}
       onFocus={() => {
-        openOnHover ? setOpen(true) : null;
+        setOpen(true);
       }}
       open={isOpen}>
       <button
--- a/packages/react/src/components/Tooltip/DefinitionTooltip.tsx
+++ b/packages/react/src/components/Tooltip/DefinitionTooltip.tsx
@@ -104,6 +104,9 @@ const DefinitionTooltip: React.FC<DefinitionTooltipProps> = ({
       onMouseEnter={() => {
         openOnHover ? setOpen(true) : null;
       }}
+      onFocus={() => {
+        openOnHover ? setOpen(true) : null;
+      }}
       open={isOpen}>
       <button
         {...rest}
@@ -112,9 +115,6 @@ const DefinitionTooltip: React.FC<DefinitionTooltipProps> = ({
         aria-expanded={isOpen}
         onBlur={() => {
           setOpen(false);
-        }}
-        onClick={() => {
-          setOpen(!isOpen);
         }}
         onKeyDown={onKeyDown}
         type="button">
--- a/packages/react/src/components/Tooltip/DefinitionTooltip.tsx
+++ b/packages/react/src/components/Tooltip/DefinitionTooltip.tsx
@@ -116,6 +116,9 @@ const DefinitionTooltip: React.FC<DefinitionTooltipProps> = ({
         onBlur={() => {
           setOpen(false);
         }}
+        onClick={() => {
+          setOpen(!isOpen);
+        }}
         onKeyDown={onKeyDown}
         type="button">
         {children}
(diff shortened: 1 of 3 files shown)


guidari:
Hey @aagonzales .
When you have the chance, can I get a design consultation in this issue. They are simple changes, but just to make sure there is nothing missing.

packages/react/src/components/Tooltip/DefinitionTooltip.tsx
```diff
@@ -104,6 +104,9 @@ const DefinitionTooltip: React.FC<DefinitionTooltipProps> = ({
       onMouseEnter={() => {
         openOnHover ? setOpen(true) : null;
       }}
+      onFocus={() => {
+        openOnHover ? setOpen(true) : null;
```
tw15egan: Should this open on focus regardless of the `openOnHover` prop? Reading over the issue it seems like no matter what, it should open on focus.

guidari: True!! My bad

tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):
😎

aagonzales (approved):
Looks good to me too!

PR #15895, opened 2024-03-05
Added hideLabel to Dropdown family

### LINKED ISSUE #15853: [Bug]: Dropdown skeleton state missing label
### Package

@carbon/react, @carbon/styles, carbon-components, carbon-components-react

### Browser

_No response_

### Package version

Latest

### React version

_No response_

### Description

It appears that skeleton state of Dropdown component is missing the label. Only the field itself get skeleton-ized, and label went missing. This caused a small page jump when loading. Other common form inputs has the label in skeleton state already.

Also, because storybook doesn't contain the skeleton states of these components, I couldn't check what they are: 
- Multiselect
- Combo box
- Time picker

image

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-dropdown--skeleton

### Steps to reproduce

Click through for skeleton state story and look for label

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

guidari:
Closes #15853

Added `hideLabel` prop to Dropdown Skeleton. This will avoid the page to jump when loading if the user does not have a label added to the Dropdown or any of it's variants.

#### Testing / Reviewing

- Check if the Dropdown Skeleton is working as expected.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -2990,6 +2990,9 @@ Map {
       "className": Object {
         "type": "string",
       },
+      "hideLabel": Object {
+        "type": "bool",
+      },
       "size": Object {
         "args": Array [
           Array [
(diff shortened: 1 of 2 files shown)


tw15egan:
@guidari I do not see any skeleton label above the `Dropdown`

Screenshot 2024-03-05 at 2 29 26 PM

guidari:
> @guidari I do not see any skeleton label above the `Dropdown`, which seems like we should show by default to match the `TextInput` implementation

I thought I had commited and pushed 😅
Now should be on the deploy preview.
Screenshot 2024-03-05 at 17 01 17

tw15egan (approved):
LGTM 👍 ✅

andreancardona (approved):
🔥

PR #15904, opened 2024-03-06
chore(combobox): refactor to use useCombobox

### LINKED ISSUE #14684: Refactor ComboBox to use `useComboBox` instead of `<Downshift>` component
Downshift plans to move forward with their React Hooks API and deprecate and remove `<Downshift>` from the API:

* https://github.com/downshift-js/downshift/issues/1233

We need to refactor our ComboBox to no longer use `<Downshift>` and instead use the `useComboBox` hook. As part of this we'll need to ensure downstream consumers aren't impacted at all and nothing of significance changes. If we can't contain this without a breaking change, we'll need to put the refactor behind a feature flag (using the same component name) and slate it for v12.

### LINKED ISSUE #15927: [Bug]: Combo box incorrectly resets to previously selected value (or no-value) when a click occurs INSIDE the text input
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.52.0

### React version

v18.2.0

### Description

When using the Combo Box, it is expected that if you type in text and click away, it should reset to a valid state which would be Empty, or the previously selected value.

However, this behavior should not occur when clicking INSIDE the search input itself. For example, lets say I am typing into the Combo Box, and realize I had a misspelling. I click to move the cursor, and the text resets to the last valid state. This feels like a bug and is not a great user experience. 

This can be reproduced in the story book: https://react.carbondesignsystem.com/?path=/docs/components-combobox--overview

### Reproduction/example

### Steps to reproduce

1. Go to: https://react.carbondesignsystem.com/?path=/story/components-combobox--default
2. Type "Optyon" into the box
3. Realize you spelled it wrong and place the cursor before the "y"
4. Notice the value is cleared out ): 

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #14684 
Closes https://github.com/carbon-design-system/carbon/issues/15927

🚧 WIP


tw15egan:
We may want to look into fixing https://github.com/carbon-design-system/carbon/issues/15927 in this PR as well

tw15egan:
@tay1orjones Most of it is there, but it should autofocus the first item in the `Combobox` when it is opened when no item is selected. When an item is selected, the correct behavior is observed

tay1orjones:
I've fixed a couple things:

* I added forwardRef to `ListBoxTrigger` which got rid of the error failing all the unit tests
* TJ's note: the first item is highlighted when the menu opens when nothing is selected

A much bigger issue I have yet to track down is that opening the menu and selecting an item with <kbd>enter</kbd> doesn't work at all. The `selectedItem` is never updated. The [state reducer example](https://www.downshift-js.com/use-combobox/#state-reducer) from the downshift docs works properly without having to do anything special. I'm not sure why ours isn't working. I've added some failing conditions to the e2e test that should pass once it's fixed.

tay1orjones:
Two additional items we need to look into::

* This impacts styling for dropdown - the borders specifically are flashing on the disabled item, also every item looks disabled
* on hover of a listboxitem, it's applying the focus ring when it shouldn't be

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -539,276 +528,267 @@ const ComboBox = forwardRef(
...
-                warnTextId={warnTextId}>
-                <div className={`${prefix}--list-box__field`}>
-                  <input
-                    role="combobox"
-                    disabled={disabled}
-                    className={inputClasses}
-                    type="text"
-                    tabIndex={0}
-                    aria-autocomplete="list"
-                    aria-expanded={rootProps['aria-expanded']}
```
guidari, line 715: By removing this `aria-expanded` attr, might affect this test file _[test-helpers.js](https://github.com/carbon-design-system/carbon/blob/7b9a05be22b9bee34f70299f7568f1aeb6e7c06a/packages/react/src/components/ListBox/test-helpers.js#L57)_, right?

```
export const findPopupNode = () => {
  return document.querySelector('[aria-haspopup="listbox"]');
};

expect(findPopupNode()).toHaveAttribute('aria-expanded', 'true');
```

tay1orjones:
Duplicate of https://github.com/carbon-design-system/carbon/pull/16283

PR #15910, opened 2024-03-07
feat(structuredList): update to use radio button icon, radios are left aligned under featureflag

### LINKED ISSUE #15479: [Selectable Components: Structured List] Dev Implementation
## Acceptance criteria

- [x] Update the Structure List with the below changes
- [x] By default, no option will be preselected. 

[Final design solution](https://github.com/carbon-design-system/carbon/issues/14766#issuecomment-1874852921)

----

`_Radio button items will not be preselected by default_`

### Specs
_Here are the updates:_
> Both icon placements, i.e., the proposed icon on the left and the current icon on the right, will coexist for now. The proposed placement (icon on the left) will be placed under a feature flag. [Figma file solution](https://www.figma.com/file/w4QA4bnNZAiySPskA5qZMW/Selectable-Components?type=design&node-id=548-4530&mode=design&t=K)
> Without Background_icon on right
> without background icon left

---

> Additionally, we need to support an option for structured list when it has a a background applied. This is how the states look when there is a background applied to the structured list. [Figma file solution](https://www.figma.com/file/w4QA4bnNZAiySPskA5qZMW/Selectable-Components?type=design&node-id=548-4530&mode=design&t=KpyfeZDgqaFCYOZu-4)
> with background icon right
> with background icon left

---

### Padding

image

---

2nikhiltom:
Closes #15479 & #15931

**New**

- Icons will always be visible at the start of every row by default 
- Adds new Saas flag : "enable-v12-structured-list-visible-icons"
- Adds focus ring around the row : on click 

**Changed**

- Unselected/Unchecked icons are always visible at the start of every row
- Checked/selected icon is visible only when a row is selected 
- Icons are changed : checkmarks to radio icons

#### Testing / Reviewing

- Verify Unselected/Unchecked icons are visible on every row
- Verify Selected/Checked icons is visible only when a row is selected


2nikhiltom:
Hey @Kritvi-bhatia17 : Can I get your design review on this .

Kritvi-bhatia17:
Yes! sure @2nikhiltom will do that

tay1orjones (changes requested):
Feature flags should be used to modify component source code. The feature flag stories for that component would then show the result of the flag being turned on and the modified behavior of the component source code.

Is the current checkmark functionality built into any of the components' source, or does it require the consumer to compose the checkmark into the proper location within StructredListCell?

If it's already on the consumer to compose this properly, we wouldn't need a feature flag. We could just update the existing stories to show the new icon being used on the right.

Kritvi-bhatia17 (approved):
Looks good to me from design perspective!!
Nice work @2nikhiltom 🔥

Kritvi-bhatia17 (changes requested):
Hii @2nikhiltom! after speaking with @laurenmrice , I realized that we need to add a prop for both with and without background options under the 'Structured Lists' section for both 'Default' and 'Selection'.

tay1orjones:
I just pushed an update here that adds the feature flag to modify the styling, and also added a new `selection` prop to `StructuredListRow` that will handle adding the radio button icon on the left.

tay1orjones:
@2nikhiltom I think I got the focus ring working actually, take a look and let me know what you think

2nikhiltom:
I've implemented focus ring operations for keyboard and mouse navigation, following has changed.
- The first item in the list will now receives focus when navigating via the Tab key.
- Items receives focus rings upon being clicked and
- Focus rings are removed when the user clicks away/outside from the focused item

tay1orjones:
The PublicAPI snapshot is failing, just needs a `yarn test -u` and push up the changes

packages/react/src/components/StructuredList/StructuredList.tsx
```diff
@@ -263,14 +271,34 @@ export function StructuredListRow(props: StructuredListRowProps) {
...
       [`${prefix}--structured-list-row--header-row`]: head,
-      [`${prefix}--structured-list-row--focused-within`]: hasFocusWithin,
+      [`${prefix}--structured-list-row--focused-within`]:
+        hasFocusWithin && (selectedRow === id || selectedRow === null),
       [`${prefix}--structured-list-row--selected`]: selectedRow === id,
     },
     className
   );
-
+  const useClickOutside = (handler) => {
```
tay1orjones: There's a `useOutsideClick` helper in the codebase that does essentially this same thing that can be used instead.

packages/react/src/components/StructuredList/StructuredList.tsx
```diff
@@ -263,14 +271,34 @@ export function StructuredListRow(props: StructuredListRowProps) {
     `${prefix}--structured-list-row`,
     {
       [`${prefix}--structured-list-row--header-row`]: head,
-      [`${prefix}--structured-list-row--focused-within`]: hasFocusWithin,
+      [`${prefix}--structured-list-row--focused-within`]:
+        hasFocusWithin && (selectedRow === id || selectedRow === null),
```
tay1orjones: Looking at the diff, it seems like the condition here may need to be something closer to

`(hasFocusWithin && !selection) || (hasFocusWithin && selection && selectedRow === id`)

So that the existing behavior is preserved when `!selection`, and that we only check the `selectedRow` if we know `selection` is turned on.

packages/styles/scss/components/structured-list/_structured-list.scss
```diff
@@ -200,6 +199,27 @@
     fill: $icon-primary;
   }
 
+  @if not(
+    enabled('enable-v12-structured-list-visible-icons') or
+      $enable-v12-structured-list-visible-icons
```
tay1orjones: Both of these need to be false I think
```suggestion
  @if not(
    enabled('enable-v12-structured-list-visible-icons') and
      $enable-v12-structured-list-visible-icons
```

packages/styles/scss/components/structured-list/_structured-list.scss
```diff
@@ -242,3 +262,11 @@
...
   }
 }
+
+.#{$prefix}--structured-list-row--selected .structured-checks {
+  display: block;
+}
+
+.#{$prefix}--structured-list-row--selected .structured-unchecks {
+  display: none;
+}
```
tay1orjones:
```suggestion
```

tay1orjones (changes requested):
This is looking great, just a couple things to change and I think this will be good to go.

packages/react/src/components/StructuredList/StructuredList.tsx
2nikhiltom: Hey Taylor,
```
[`${prefix}--structured-list-row--focused-within`]:
        (hasFocusWithin && !selection) ||
        (hasFocusWithin && selection && (selectedRow === id || selectedRow === null)),
```
   Added `(selectedRow === id || selectedRow === null)  `condition as well to ensure focus on the first item/row when list is navigating through Tab keys and no-row is selected (selectedRow === null)

packages/react/src/components/StructuredList/StructuredList.tsx
2nikhiltom: Great catch, thank you. 
Updated the code to use existing `useOutsideClick` helper

2nikhiltom:
Hey @tay1orjones , thanks for the review.
I have made suggested changes and few other changes to ensure rows receive focus only when `selection` is enabled, eliminating focus rings on `default` stories.

2nikhiltom:
> The PublicAPI snapshot is failing, just needs a `yarn test -u` and push up the changes

Hey @tay1orjones , @andreancardona : it seems like the problem is still there even after running "yarn test -u" and pushing changes. Any ideas on what might be missing?

tay1orjones:
@2nikhiltom based on [the logs](https://github.com/carbon-design-system/carbon/actions/runs/8369783506/job/22916086871#step:8:82), it was failing on a snapshot change unrelated to the changes you made here. Generally that signals that the PR needs to be updated with the latest from `main` and then the snapshots updated again.

I pressed the "update with latest" button on the PR interface here, pulled down the PR locally, ran `yarn`, `yarn build`, and `yarn test -u` which resulted in two snapshot changes. I pushed them up - I think this should resolve the CI failures.

tay1orjones (approved):

tay1orjones:
@Kritvi-bhatia17 @andreancardona this is ready for your review

andreancardona (approved):
looks good to me - nice job @2nikhiltom !

Kritvi-bhatia17 (approved):
Great work Nikhil!! 🔥

Issue #13828, opened 2023-05-18
DatePicker: Expose all flatpickr's option/props in DatePicker


francinelucca:
### The problem

There are some flatpickr config [options](https://flatpickr.js.org/options/) we are not exposing through our `DatePicker`: `altFormat`,`altInput` and others. 

### The solution

We should forward all available options through `DatePicker`.

~We should also consider upgrading to the [latest version](https://www.npmjs.com/package/flatpickr/v/4.6.13) of flatpickr to hopefully cleanup some of our bug overrides and work with a more stable base.~

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

tay1orjones:
Related is the v12 plans for this component: https://github.com/carbon-design-system/carbon/discussions/13355

We could update flatpickr or maybe we want to focus on a new implementation that leaves flatpickr behind

PR #15918, opened 2024-03-07
feat(datepicker): add `parseDate` prop, improve default date parsing

### LINKED ISSUE #15432: [Bug]: DatePicker input seems to convert date if not given in the correct date format
### Package

@carbon/react

### Browser

Chrome

### Package version

1.4.0

### React version

18.2.0

### Description

When using a single DatePicker with the calendar, if you enter in a date in a different format, the component ends up with a totally different date selected, almost like it's tried to correct the input to be in the correct date format.
If I have a DatePicker in mm/dd/yyyy format and I enter the following date, 19th December 2023, in dd/mm/yyyy format, the DatePicker updates it to be 7th December 2024 instead.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datepicker--single-with-calendar

### Steps to reproduce

- click on the date picker so that the calendar appears
- while the date picker is in `mm/dd/yyyy` format, type 19/12/2023
- the date seems to convert to 07/12/2023

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #15432 
Part of #13828

This improves DatePicker's date parsing functionality when using the default `dateFormat`. It also adds a new `parseDate` prop for consumers to supply their own date parsing function when using other date formats.

How this works:
* When using the default dateFormat, the DatePicker will attempt to parse the year, month and day into a Date object.
* If an invalid month or day is provided, `1` will be used instead. 
* If a totally invalid date is provided (`a1/0a/a999`), the input will be cleared.
* If a custom `dateFormat` is used and a `parseDate` is not provided, the default within flatpickr will be used, which can cause odd date correction based on how native date methods work, https://github.com/carbon-design-system/carbon/issues/15432#issuecomment-1967447677

#### Changelog

**New**

- Add `parseDate` prop to DatePicker
- Provide improved date parsing when using the default date format
- Add tests to cover the new default parsing, the passing of a `parseDate` prop function, and completely invalid dates.

#### Testing / Reviewing

- Go to the DatePicker stories and start entering various invalid dates. Month/day should correct to `01` and totally invalid dates should clear the input.
- Ensure it works on all DatePicker variants/stories (range, simple, etc)


tw15egan (approved):
Works great! 👍 ✅

andreancardona (approved):
tested and looks good!

Issue #13807, opened 2023-05-15
Add native `dialog` element support


tay1orjones:
## Context

An initial exploration and delivery of a `Dialog` component was done a while ago: 
- https://github.com/carbon-design-system/carbon/issues/8826
- https://github.com/carbon-design-system/carbon/pull/9496

This experimental work was never exported or shown in storybook but still lives in the codebase. Most of the spec of that work stream was focused on managing focus and handling interactions such as <kbd>escape</kbd> closing the dialog.

## The current state

Since then, there has been some progress on the native dialog element spec and much of this focus and interaction behavior is now provided natively by browsers. 

- https://www.scottohara.me/blog/2023/01/26/use-the-dialog-element.html
- https://www.youtube.com/watch?v=ywtkJkxJsdg
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
- https://www.makethingsaccessible.com/guides/modal-vs-non-modal-dialogs

## Proposal

This issue is to propose considering using the native dialog element in place of the existing implementation that uses the `FocusScope` primitive.

Depending on what we consider to be "must have" functionality of a `Dialog`, it might be that the bulk of the requirements might be met by the `dialog` element. It might be a simple wrapper around that element is all we need.

### Pieces solved in https://github.com/carbon-design-system/carbon/pull/15926
- [x] Evaluate the interaction provided via the native element
- [x] Compare the native implementation to the one we have in the codebase today
- [x] Determine what we may need to provide to consumers in addition to what is provided by default by the browser implementation
- [x] Potentially refactor the implementation to not manage focus via the FocusScope helpers (potentially remove these helpers if they're not being used)

### Outstanding items that still need to be made into sub-issues
- [ ] Add Dialog and associated refactors/usages/etc. to `@carbon/web-components`
- [ ] Move to stable , work through stable checklist (code, website docs, etc.)
- [ ] ...

mbgower:
Some considerations for the assessment:
- granularity for placing focus when dialog appears (on dialog element, on controls inside dialog)
- effect on scrolling inside dialog, depending on focus placement
- keyboard retention in modal dialog
- naming/title for the dialog

[...]

PR #15926, opened 2024-03-08
experiment: native dialog component

lee-chase:
This adds an `unstable_dialog` component.

Created while investigating the possibility of creating a dialog component based on native dialog.

I included a side panel and tearsheet kind as part of the experiment. 

Previously some questions have been raised here, perhaps it can be used to answer some of them.

https://github.com/carbon-design-system/carbon/issues/13807

Of note is that the `modal` form blocks the whole page. The modal dialogs can be nested.

Also quite a bit of the code added was to cope with the lack of support for
Unfortunately needed quite a bit of code to cope with the lack of support for some of the in/out styling

https://caniuse.com/mdn-css_at-rules_starting-style 
https://caniuse.com/mdn-css_properties_display_is_transitionable

This much simpler code pen achieves a similar effect in Chrome and Edge
https://codepen.io/lee-chase/pen/QWPbRLp/9c8c3f23b42a395dc519839e12f5d647

NOTE: Use of `:where([data-attri="xyz"])` done to see how it compares with BEM modifier overload. `:where` keeps the specificity low.


lee-chase:
The following branch contains nested dialogs. This appears to be problematic for various styling reasons:
1. Dialog sizes are based on their containers (easiest to see in non-modal)
2. Child dialogs sit obey normal stacking rules, so only appear over the top of content before them (see secondary button in image below).
3. The child dialogs cause layout issues in their parents (plain dialog with 3 open see image).

image

4. Several other styling issues not resolved try stacking tearsheet or side panel kinds. Although these are probably just a case of additional styling.
image
image
image

Modal versions work better.
image
image
image

https://github.com/lee-chase/carbon/tree/nativeDialog

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
+import { usePrefix } from '../../internal/usePrefix';
+import cx from 'classnames';
+import { useMatchMedia } from '../../internal/useMatchMedia';
+import { Close } from '@carbon/icons-react';
+import { IconButton } from '../IconButton';
+
+const Dialog = React.forwardRef(
+  ({ children, kind, modal, onClose, open = false, width, ...rest }, ref) => {
+    const backupRef = useRef(null);
+    const localRef = ref || backupRef;
```
tay1orjones: We have a `mergeRefs` helper function you could use here

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
+    /**
+     * open state management only exists here to permit animation
+     *
+     * When support for allow-discrete is ubiquitous https://caniuse.com/mdn-css_properties_display_is_transitionable
+     *
+     * then this can be removed and CSS simplified
+     *
+     * https://codepen.io/lee-chase/pen/QWPbRLp/9c8c3f23b42a395dc519839e12f5d647
+     *
+     */
```
tay1orjones: For the purposes of this experiment I'd say lets avoid the state management and go all in on allow-discrete.

Support recently landed in the firefox nightlies, https://bugzilla.mozilla.org/show_bug.cgi?id=1805727 and the Webkit folks have landed it into main: https://github.com/WebKit/WebKit/pull/21515. Just a matter of time until we see it released I think.

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
-  'aria-labelledby': PropTypes.string.isRequired,
+  children: PropTypes.node,
 
   /**
-   * Provide children to be rendered inside of the Dialog
+   * kind default (dialog/modal), side-right, side-left, tearsheet
+   *
    */
-  children: PropTypes.node,
+  kind: PropTypes.oneOf(['', 'side-start', 'side-end', 'tearsheet']),
```
tay1orjones: I think I'd prefer this to be as thin a wrapper as possible around the native `dialog` and not have variant props within here. These types of abstractions can easily be made atop the Dialog in new components.

For instance in theory a new `Tearsheet` component could just wrap `Dialog` and add an additional class for the appropriate tearsheet styling - same thing with side-start and side-end.

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
-      } else {
-        node.removeAttribute('aria-hidden');
-      }
-    }, 0);
-  });
-}
+  /**
+   * Sets the width of the side panel
+   */
+  width: PropTypes.oneOf(['xs', 'sm', 'md', 'lg', 'xl']),
```
tay1orjones: Same thing here with this being pulled out in favor of being housed in a component composed atop `Dialog` like a new `SidePanel` or something

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
+   * Modal specifies whether the dialog is modal
+   */
+  modal: PropTypes.bool,
 
   /**
    * Provide a handler that is called when the Dialog is requesting to be closed
    */
-  onDismiss: PropTypes.func.isRequired,
-};
+  onClose: PropTypes.func,
```
tay1orjones: Elsewhere in the system we call this `onRequestClose`, might be worth unifying against that convention

packages/react/src/components/FocusScope/index.js
tay1orjones, line 1: Super excited we no longer need this!

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
+   * kind default (dialog/modal), side-right, side-left, tearsheet
+   *
    */
-  children: PropTypes.node,
+  kind: PropTypes.oneOf(['', 'side-start', 'side-end', 'tearsheet']),
+
+  /**
+   * Modal specifies whether the dialog is modal
+   */
+  modal: PropTypes.bool,
```
tay1orjones: I'm not sure yet where we'd apply modal vs non-modal within the system, but agree it's definitely worth having a prop 👍

tay1orjones (commented):
Thanks so much for looking into this, I agree - it answers quite a few questions!

The tooltip on the close button appears on top of the modal as intended, so that's great news. Once we get the autoAlign stuff in I'll be eager to test how it works with this.

I have a few other thoughts below. I wouldn't be opposed to eventually merging this in, not exporting it for now, and hiding the stories from the storybook. I think it's valuable to keep this on deck and continue to experiment, and merging it in would avoid PRs sitting around getting stale.

packages/react/src/components/Dialog/index.js
```diff
@@ -1,153 +1,231 @@
...
+    return (
+      // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
+      <dialog
+        {...rest}
+        className={cx(`${prefix}--dialog`)}
+        data-closing={isClosing ? '' : null}
+        data-is-open={localOpen && isOpen ? '' : null}
+        data-width={width}
+        data-modal={modal ? '' : null}
+        data-kind={kind ? kind : null}
```
tay1orjones: I'm not sure about this. Use of `:where` to bring specificity of carbon styles to `0` is an interesting idea, but I'm not sure about the data attribute approach. It would fragment the css portion of our style guide. Refactoring to take this approach everywhere would be a significant breaking change.

lee-chase: I don't disagree, but was suffering a bit of modifier overload with BEM. I still like the B & E but think some class attributes can look very overloaded with BE BEM1 BEM2 BEM 3BEM4. There are a few places where breaks from BEM e.g. `is-visible` which looks much more readable BE M1 M2 M3 M4. 

Presumably you'd like me to change to `B__E--M`

packages/react/src/components/Dialog/index.js
lee-chase: I'm not sure of the use-case for a modeless dialog either, but was just mirroring the standard HTML.

packages/react/src/components/Dialog/index.js
lee-chase: May still require @starting-style, will experiment. https://caniuse.com/mdn-css_at-rules_starting-style

https://bugzilla.mozilla.org/show_bug.cgi?id=1834876

NOTE: Without starting style the codepen does not transition in.

packages/react/src/components/Dialog/index.js
lee-chase: That's fair, it would be nice to be able to keep the potential versions alive somewhere. Do. you have any suggestions @tay1orjones ?

Wrapper components in this or seperate PR? They could equally be left as unstable/experimental.

packages/react/src/components/Dialog/index.js
lee-chase: That comment is a bit misleading, apart from `xl` the current ComposedModal has sizes.

Although I suppose by thin wrapper you may be considering that ComposedModal, SidePanel and any other variants would wrap `Dialog`.

packages/react/src/components/Dialog/index.js
lee-chase: Not sure `mergeRefs` is trying to do the same thing. Here a user may or may not pass a ref for the Dialog component. Either way we need a ref to call `showModal` and `close`

packages/react/src/components/Dialog/Dialog-story.js
```diff
@@ -7,149 +7,75 @@
...
-import { Portal } from '../Portal';
+import React, { useEffect, useState } from 'react';
+import Dialog from './';
+import Button from '../Button';
+import { action } from '@storybook/addon-actions';
 
 export default {
   title: 'Experimental/unstable_Dialog',
   component: Dialog,
-  includeStories: [],
```
tay1orjones, line 19: I think adding this back in will ensure these stories don't show up in the storybook for now

packages/styles/scss/components/_index.scss
```diff
@@ -23,6 +23,7 @@
 @use 'data-table/skeleton';
 @use 'data-table/sort';
 @use 'date-picker';
+@use 'dialog';
```
tay1orjones:
```suggestion
```

Let's not export these for now

tay1orjones (commented):
I think this puts the Dialog component on a much better track than it was previously! I still think we should keep it internal-only at this point and not export it or include it in storybook. If those changes are made I'd personally love to see this merged.

From here we can continue to iterate on this and explore the extra areas like building a new Modal implementation atop it, etc.

lee-chase:
@tay1orjones accepted/made changes to remove from storybook.

tay1orjones (approved):
I know it's silly but I'm genuinely excited about this one 😄
Native elements and standards feel so good.

alisonjoseph (approved):
Thanks for adding this 🥳

PR #15934, opened 2024-03-11
fix(ComposedModal): align handleMouseDown between Modal, ComposedModal

### LINKED ISSUE #15857: [Bug]: ComposedModal does not use selectorsFloatingMenus on mouse click events like Modal does
### Package

@carbon/react

### Browser

_No response_

### Package version

Latest, the problem persists in the main github branch too

### React version

_No response_

### Description

When using the ComposedModal the selectorsFloatingMenus prop is not used on mouse down events, meaning the selectorsFloatingMenus prop does not do anything for click events and the only way to stop react portals from closing the modal is to set preventCloseOnClickOutside to true.

**Code from the modal source showing what is expected:** 
_https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Modal/Modal.tsx#L297_
```
  function handleMousedown(evt: React.MouseEvent<HTMLDivElement>) {
    const target = evt.target as Node;
    evt.stopPropagation();
    if (
      innerModal.current &&
      !innerModal.current.contains(target) &&
      !elementOrParentIsFloatingMenu(target, selectorsFloatingMenus) &&
      !preventCloseOnClickOutside
    ) {
      onRequestClose(evt);
    }
  }
```

**Alternative logic in ComposedModal that prevents selectorsFloatingMenus from working as expected:**
_https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComposedModal/ComposedModal.tsx#L269_
```
    function handleMousedown(evt: MouseEvent) {
      evt.stopPropagation();
      const isInside = innerModal.current?.contains(evt.target as Node);
      if (!isInside && !preventCloseOnClickOutside) {
        closeModal(evt);
      }
    }
```

Similar issues to this have ben reported in the past and the issues have been automatically closed when the only response was to "prevent closing inside the closeModal callback"... this is not an appropriate resolution to this issue. Both modals should implement the selectorsFloatingMenus prop handling in exactly the same way otherwise it is just confusing for developers (I'm actually a little surprised that Modal doesn't just use ComposedModal internally anyway, rather than maintaining two separate components that do the same thing).

Also, as a side-note; it would be more efficient to call "innerModal.current?.contains" conditionally, only if preventCloseOnClickOutside is false. The current implementation will scan every parent in the entire DOM tree on every single mouse click, regardless of what preventCloseOnClickOutside is set to, which is incredibly inefficient for complex DOM trees and should be avoided unless absolutely necessary.

Ideally, the following code should be used for both modal types, this ensures the same functionality in both implementations and also fixes the unnecessary DOM tree traversal inefficiency.
```
  function handleMousedown(evt: React.MouseEvent<HTMLDivElement>) {
    const target = evt.target as Node;
    evt.stopPropagation();
    if (
      !preventCloseOnClickOutside &&
      !elementOrParentIsFloatingMenu(target, selectorsFloatingMenus) &&
      innerModal.current &&
      !innerModal.current.contains(target)
    ) {
      onRequestClose(evt);
    }
  }
```

### Reproduction/example

See description

### Steps to reproduce

See description

### Suggested Severity

### Application/PAL

IBM Internal

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15857

Fixes an issue preventing `selectorsFloatingMenus` from being used to prevent modal closures. Now aligned with `Modal` usage, and only runs if `preventCloseOnClickOutside` is set to false.

#### Changelog

**Changed**

- Aligned `handleMouseDown` usage between `Modal` and `ComposedModal`

#### Testing / Reviewing

Ensure there are no regressions to `ComposedModal` and that portals do not close the `ComposedModal` if they are passed in to `selectorsFloatingMenus`


alisonjoseph (approved):
LGTM, looks like there is a minor merge conflict

guidari (approved):

PR #15939, opened 2024-03-11
feat: adds stacking to PR #15926

lee-chase:
Created this PR as an addendum to #15926

Adds the concept of stacking sibling dialogs, note it is expected they are declared in the order displayed.

Adds to https://github.com/carbon-design-system/carbon/pull/15926

image
image
image
image

image
image
image
image


tay1orjones:

tay1orjones:
Related

* https://github.com/carbon-design-system/carbon/pull/15891
* https://github.com/carbon-design-system/carbon/pull/15882
* https://github.com/carbon-design-system/carbon/pull/15881

PR #15945, opened 2024-03-12
fix: fixed pixel in dropdown family

guidari:
This will fix a wrong pixel in the drop family

https://github.com/carbon-design-system/carbon/assets/52183462/931772a3-325a-4c6a-bc2f-b2efe4bbfa2d


tw15egan (approved):
Thanks for fixing that! 👍 ✅

tay1orjones (approved):
Wow, good eye! 🔍

alina-jacob (approved):
Looks good Gui! Thanks for fixing it! 🚀 🥳

PR #15952, opened 2024-03-13
feat: added dividers accross multiple states for AI select

### LINKED ISSUE #15860: Introduce Divider across multiple states for AI Select
image

riddhybansal:
Closes #15860 

Added dividers accross multiple states for AI select

#### Changelog

**New**

- Added dividers accross multiple states for AI select

#### Testing / Reviewing

- Go to the storybook and test the divider by adding warn/invalid to AI select


packages/styles/scss/components/select/_select.scss
```diff
@@ -284,27 +284,49 @@
...
-  .#{$prefix}--select--slug.#{$prefix}--select--invalid .#{$prefix}--slug,
-  .#{$prefix}--select--slug.#{$prefix}--select--warning .#{$prefix}--slug {
-    inset-inline-end: $spacing-10;
+  .#{$prefix}--select-input__wrapper[data-invalid]
+    .#{$prefix}--select-input
+    ~ .#{$prefix}--select__invalid-icon,
+  .#{$prefix}--select-input__wrapper
+    .#{$prefix}--select-input
+    ~ .#{$prefix}--select__invalid-icon {
+    inset-inline-end: convert.to-rem(82px);
```
guidari: Maybe you just to have to add the slug `.#{$prefix}--select--slug` class here to make sure it won't change the select.

Something like that:
```suggestion
.#{$prefix}--select--slug
  .#{$prefix}--select-input__wrapper[data-invalid]
    .#{$prefix}--select-input
    ~ .#{$prefix}--select__invalid-icon,
    .#{$prefix}--select--slug
  .#{$prefix}--select-input__wrapper
    .#{$prefix}--select-input
    ~ .#{$prefix}--select__invalid-icon {
    inset-inline-end: convert.to-rem(82px);
```

guidari (changes requested):
Hey @riddhybansal LGTM 🚀

Just on thing in the Select component
Screenshot 2024-03-13 at 11 00 01

guidari (approved):

tay1orjones (changes requested):
The `padding-inline-end` on the select needs to be modified so text doesn't run behind the icon.

image

tay1orjones:
A few other questions I just thought of:

* For the `px` values, could we derive those from the spacing tokens in certain situations? 
* Does this need to be supported for the other `size` options? If so, could we add a control to be able to see how it works with the other sizes?

tay1orjones:
> * Does this need to be supported for the other `size` options? 

@alina-jacob could you help answer this? The spec only refers to horizontal spacing, so I think the existing solution would work for all sizes _if_ the values should be the same on every size.

alina-jacob:
> > * Does this need to be supported for the other `size` options?
> 
> @alina-jacob could you help answer this? The spec only refers to horizontal spacing, so I think the existing solution would work for all sizes _if_ the values should be the same on every size.

Hi @tay1orjones you're right, the spacing stays the same across large, medium and small sizes.

packages/styles/scss/components/select/_select.scss
```diff
@@ -284,27 +284,56 @@
...
+    position: absolute;
+    background-color: $border-subtle-01;
+    block-size: convert.to-rem(16px);
+    content: '';
+    inline-size: convert.to-rem(1px);
+  }
+
+  .#{$prefix}--select--slug .#{$prefix}--slug::before {
+    display: none;
+    inset-inline-start: convert.to-rem(-9px);
```
alisonjoseph:
```suggestion
    inset-inline-start: calc(-#{$spacing-03} - 1px);
```

In the spirit of @tay1orjones [PR around magic numbers](https://github.com/carbon-design-system/carbon/pull/15995), would we want to update these values to be the padding value plus the divider size?

tay1orjones (approved):

alina-jacob (approved):
looks good Riddhi! 🚀 
thanks!

alisonjoseph (approved):

PR #13371, opened 2023-03-17
fix: add area-describedby to components that use helperText

francinelucca:
Closes #12722 

Adds `aria-describedby` attribute to components that use helperText:

- [x] Multiselect
- [x] Dropdown
- [x] Combobox
- [x] DatePickerInput
- [x] NumberInput
- [x] ProgressBar
- [x] Select
- [x] TextArea
- [x] ControlledPasswordInput
- [x] PasswordInput
- [x] RadioButtonGroup -- todo: confirm with michael

#### Changelog

**Changed**

-  Added `aria-describedby` attribute in `Multiselect`, `Dropdown`, `Combobox`, `DatePickerInput`, `NumberInput`, `Select`, `ProgressBar`,`TextArea`, `ControlledPasswordInput`, `PasswordInput`, `RadioButtonGroup`

#### Testing / Reviewing

Components should behave and look exactly the same. Voiceover now reads helper text when focused on input

Issue #12722, opened 2022-11-23
[a11y]: Dropdown does not use aria-describedby to link help text with component


mbgower:
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

React storybook

### React version

https://react.carbondesignsystem.com/?path=/story/components-multiselect--default

### Automated testing tool and ruleset

n/a

### Assistive technology

JAWS

### Description

Any helper text should be automatically surfaced to assistive technologies through the use of aria-describedby.
This should happen on ALL Carbon components that can have helper text.
it is in place for Text input, so it only needs to be emulated for Dropdown/MultiSelect Dropdown and any other components that do not have it
![image](https://user-images.githubusercontent.com/14298245/203648304-bdba6ed0-b85c-42e1-ae6b-1ce4094fe90f.png)
Steps to resolve:
1. Given the div holding the helper text an appropriate ID
2. use this ID with aria-describedby on the div holding the input.

This means that the input will have its programmatic label announced by a Screen Reader, and then have any helper text read out after a pause.

It also means that any caution or warning text appearing in the helper text area will also be announced.

### WCAG 2.1 Violation

https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-multiselect--default

### Steps to reproduce

1. Inspect the Helper text and confirm there is no ID associated with it.
2. If ID exists, inspect input and confirm there is no use of aria-describedby

OR
Turn on JAWS and tab to Dropdown. JAWS will read the input and any option value, but will not announce the helper text.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

mbgower:
Note. You could also consider giving this helper text a role of "status", which would ensure that any changes to the text due to cautions or errors would be announced. (For instance, a user puts in a wrong value, causing the message to updated.) This may already be handled in a different way by Carbon, but it is worth validating that and pursuing the use of the status role if there is nothing in place.

mbgower:
[...]

PR #15954, opened 2024-03-13
fix(combobox): invalid and warn text should be read by screenreaders

### LINKED ISSUE #10050: [a11y]: ComboBox validation error is not read out by JAWS
### Package

### Browser

Firefox

### Operating System

Windows

### Package version

v10.47.0, v7.47.1

### Assistive technology

JAWS 2021.2103.174

### Description

JAWS screen reader tool is not able to read the validation error of a `<ComboBox>` component. Found during manual AVT Testing. The interesting thing is, that the validation error is read out correctly for a `<Select>` component. 

Looking at the generated markup I noticed that the working `<Select>` component creates an **aria-describedby** attribute which refers to the validation error message (see screenshot below). The `<ComboBox>` component does not create such an attribute, maybe this is why JAWS does not read the error.

image

Just added the **aria-describedby** attribute temporarily to the `<input>` of the ComboBox and JAWS reads the validation error successfully (see screenshot below).  So I think using **aria-describedby** attribute in the same way as in the `<Select>` component should fix this issue.

image

### WCAG 2.1 Violation

_No response_

### CodeSandbox example

https://codesandbox.io/s/bold-babbage-ckktp?file=/src/index.js

### Steps to reproduce

1. Open CodeSandbox.
2. Open JAWS and navigate to the "Select language" **ComboBox** in the sandbox.
3. JAWS does not read out the validation error "Invalid selection".
4. Navigate to the "English" **Select** component.
5. JAWS does read out the validation error.

### Code of Conduct

tay1orjones:
Closes #10050 

This is sort-of a continuation of https://github.com/carbon-design-system/carbon/pull/13371, where `aria-describedby` was added, but only for the `helperText`. This adds support for `warnText` and `invalidText` being read as noted in https://github.com/carbon-design-system/carbon/issues/12722

> ... any caution or warning text appearing in the helper text area will [should] also be announced

#### Changelog

**New**

- Added two new props to ListBox so we can supply and id for the warn text node and the invalid text node

**Changed**

- Updated the Combobox input to have aria-describedby have a value pointing to the helperText, warnText, or invalidText, when appropriate

#### Testing / Reviewing

- Take Combobox for a spin in JAWS - when Combobox is given `invalid` and `invalidText`, the screenreader should read the `invalidText` when the input is focused.
- Do the same thing with voiceover


guidari (approved):
Tested and LGTM! 🚀

andreancardona (approved):
Confirmed! 😎

PR #15959, opened 2024-03-14
Tag ellipses with Tooltip

### LINKED ISSUE #15893: [Feature Request]: change tag title overflow content to truncate and ellipsis
## Acceptance criteria

- When a tags title becomes too long for the tags container, on focus, we want the title to be able to truncate, ellipsis, and show a tooltip with the full title name. This could happen in all the different variants of tag (Read-only, Dismissible, Selectable, and Operational)
- If we need a max width set on the tag container, I would not go more than 208px.

### Specs
Read-only tag

Dismissible tag

Selectable tag

Operational tag

### Code of Conduct

guidari:
Closes #15893

This PR will add the ellipses to tags that have a long text. When the ellipses is active the Tooltip will added so the user can read the whole tag

#### Changelog

**New**

- `max-inline-size` to tags of 208px.
- Tooltip added when ellipses is active in a Tag.
- Focus is active in Read-only and Dismissible tag if the ellipses is active.
- Added prop `text` to Interactive tags. That way we can keep the ellipses logic in the primitive `Tag` only.

**Removed**

- Remove `children` prop from interactive tags.

#### Testing / Reviewing

- Go to all Tags variants to test the Ellipses and Tooltip functionality.
- Validate if the prop `text` make sense


laurenmrice (changes requested):
Looking great Gui! Just a couple comments:

### Tag (Stable)

**Playground story**
Can we show one tag here with a normal tag title length instead of the ellipsis tag? I think it’s good to show the ellipsis tag in the Read Only story like you are showing, but I think we should just show a normal tag with a desired tag title length in the Playground story. 

### Tag (Experimental)

**Dismissible story**
The dismissible tag in the small size has a misaligned close icon for most of the colors.

Screenshot 2024-04-03 at 2 19 01 PM

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -155,8 +155,14 @@ function Tooltip<T extends React.ElementType>({
   }
 
   function onMouseEnter() {
-    setIsPointerIntersecting(true);
-    setOpen(true, enterDelayMs);
+    // Interactive Tags should not support onMouseEnter
+    const isMouseOnEnterDisabled = customClassName?.includes(
+      `${prefix}--tooltip-disabled-onmouseenter`
+    );
```
tay1orjones, line 161: Is there a different way this could be accomplished? Using a classname to determine this feels super brittle and worries me.

More broadly, this feels like it erodes at the specced functionality of a Tooltip. imo all Tooltips should behave uniformly. What's the reason for this to not follow the established pattern of supporting onMouseEnter?

If there's a use case established here for disabling onMouseEnter, I wonder if we should document that reasoning and extend the option to consumers via a new prop?

guidari: There is a different way that would be using the `enterDelayMs` from `Tooltip`. We could set a high value to avoid the activation. Also we could add a prop on `Tooltip`, but I'm against that because of your second paragraph, we don't want users to remove a core functionality from a `Tooltip`.

That was something @mbgower and @laurenmrice disccused, and the reason for removing the `onMouseEnter` is that we don't want users to think that the `Tag` could be interactive by mouse. Correct me if I didn't understand correctly Lauren/Gower.

laurenmrice: The problem we are trying to solve is that if a Tag, in any variant (Read-only, Dismissible, Selectable, Operational), has a tag title too long for its container width, it should truncate in the tag, and the full title should be able to be revealed to the user.

We decided to expose the full name of the truncated tag title in a tooltip if the user focuses on the tag. The tricky part is that we have a Read-only tag variant, which is not supposed to be "interactive". We felt this way was acceptable enough to apply to the Read-only tag, as with the other tag variants, because it has limited interactive functionality but still solves the problem. I also feel that truncated tag titles probably does not happen often, we just don't want them to wrap to a second line in the tag. 

If there is a better approach to solving the above, we can change it! But this is the original solution we had come up with. We discussed this during a design critique and a CAG call in February, but would be happy to discuss it again.

guidari: One other thing to consider is the click action on a `Tag`. When a `Tag` is clicked it receives focus for a brief moment and the `Tooltip` is activated which can cause a weird flicker looking like a "bug". Having the `onMouseEnter` as default would avoid that.

mbgower:
I have mild concerns with making a read-only tag become interactive simply to expose the tip. How is a user to understand that is the intent, especially where the tag itself is taking focus?

Here are a couple of other possibilities...
1. You make the ellipsis itself clickable/hoverable.
2. You make the text of the tag the target, not the tag itself
3. For either of these, you use a [definition tooltip](https://carbondesignsystem.com/components/tooltip/usage/#definition-tooltip) (dotted underline)

Benefits:
- you already get the interaction for both keyboard and mouse as defined by the Definition Tooltip
- the user does not confuse the intent with that of activating the tag for some purpose (i.e., a toggle)

packages/react/src/components/Tooltip/Tooltip.tsx
tay1orjones: Thanks for the context! I see what you mean, makes sense.

@guidari what if we provided an undocumented way to arrest the default functionality inside `onMouseEnter`? Something vaguely like this:

```js
  function onMouseEnter() {
    if (!rest?.onMouseEnter?.()) {
      setIsPointerIntersecting(true);
      setOpen(true, enterDelayMs);
    }
  }
  
// Then from within the tag we can have the handler return false
<Tooltip
  ...
  onMouseEnter={() => false}
>
  
```

This is still a bit wonky, but at least provides direct control using the existing API without relying on a classname that can change or be removed, etc.

We're already not allowing `onMouseEnter` to be user-configurable (it's never pulled off `rest`), so I think this wouldn't be changing anything a consumer would be using.

mbgower:
That that the use of a tooltip on an operable variation of tags seems like it should work fine; it's just when the tag  is not operable that I see the main issue.

tay1orjones:
Going to mark this as a draft since we're waiting on a final revisited design spec which should be complete soon

laurenmrice:
Just wanted to note here that the design spec has been updated and reflected in [the issue](https://github.com/carbon-design-system/carbon/issues/15893) last week. 👍

guidari:
NEW PR https://github.com/carbon-design-system/carbon/pull/16437

PR #15964, opened 2024-03-14
docs(featureflags): define naming convention, unify existing flag names

### LINKED ISSUE #15728: Document feature flag naming convention
With the increasing emphasis on feature flags, we should document the proposed naming structure for feature flags so they remain consistent as new ones are added or existing ones changed.

tay1orjones:
Closes #15728

This PR proposes a naming convention for feature flags and outlines an approach for how to utilize them with breaking changes in major version releases.

#### Changelog

**New**

- Added `enable-experimental-treeview-controllable` as a replacement for `enable-treeview-controllable` (which didn't follow the naming convention)

**Changed**

- Revised experimental code documentation
- Revised feature flag documentation in storybook
- Update TreeView source to use either of the flags for the controllable API, for backwards compatibility

**Removed**

- `enable-v11-release` flag from storybook docs, it's not used anywhere in the codebase

#### Testing / Reviewing

- Treeview feature flag stories should still function as they did before
- Does this approach make sense to you? Is there anything else we should consider? What might I have forgot or left out? 🤔 

---

This spins off some new work that we'll need to tackle:

* https://github.com/carbon-design-system/carbon/issues/15966
* https://github.com/carbon-design-system/carbon/issues/15972


docs/experimental-code.md
```diff
@@ -31,39 +38,109 @@ const unstable_meta = {
...
+
+Flags are typically introduced with the `enable-experimental-*` prefix. This
+marks that the API is likely to change due to feedback or unforseen instability
+or incompatibility. It's also possible (but unlikely) that the flag is removed
+in the future. During this time your feedback is very important to ensure we've
+met all concerns relating to this new feature. If you use these, make sure to
+keep an eye on our release notes where we'll outline any changes to feature
+flags across our regularly scheduled minor version releases.
+
+#### Flags prefixed `enable-v12-*`
```
janhassel: Do you think this section should describe the behavior version-agnostic instead of v12 focused (e.g. `enable-v#-*`)? I am just wondering if it could get confusing once v12 is released and the text would likely have to change to "v13".
On the other hand that's still far away and we'll first want to evaluate how this helps the v12 upgrade process; so probably nothing to worry about for now.

docs/experimental-code.md
```diff
@@ -31,39 +38,109 @@ const unstable_meta = {
...
+### Feature flags are opt-in
+
+Feature flags are typically marked `false` or "off" by default to ensure
+backwards compatibility. These flags can be turned on individually within a
+consuming project to allow opt-in of the features incrementally.
+
+Feature flags follow a prefix naming convention that indicate status:
+
+1. `enable-experimental-*`
+2. `enable-v12-*`
```
janhassel: This might be overcomplicating it, but would there ever be a need for an in-between state? Like, the feature is stable and the API will likely not change anymore (but theoretically still could), however it's not committed to the next major version yet?

This could allow for shorter major release cycles as the amount of breaking changes aka v# feature flags is less.

I think my main concern is the psychological aspect of it. In a team with few resources, am I gonna take the *risk* to opt-in to a feature "[...] that the API is likely to change due to [...] instability or incompatibility" if it's not absolutely critical for my product? Then how can we get a decent number of early adopters to test it out as a requirement for promoting it to stable? If there'd be a state in between where we say the API is *un*likely to change and it has been evaluated by a few testers but it's not subject to semver yet, that might lower the barrier.

Just a thought though, I'm torn on this.

janhassel (commented):
Great write-up! Overall I think this makes a lot of sense and is a great way forward. Just left two thoughts.

Additionally, I think it might be good to first tackle and finalize #15966 before this to avoid confusion over that aspect with the tree view flag (consistent deprecation warning, extracting the logic you added to the `TreeView` component to the `useFeatureFlag` hook).

docs/experimental-code.md
tay1orjones: Yeah this is a good point, I'll update to reflect this and mention `enable-v12-*` as an example of the prefix in use.

tay1orjones:
> I think it might be good to first tackle and finalize https://github.com/carbon-design-system/carbon/issues/15966

Yeah on second thought I think I'm going to rip out the flag work from this PR and have it be part of #15966. `enable-treeview-controllable` will be perfect for testing the new alias work.

docs/experimental-code.md
tay1orjones: I've been thinking about this too, am also torn. My main concern is the more "layers" we have, the more likely it is that consumers don't care to understand the naming convention or the slight nuance between them.

> If there'd be a state in between where we say the API is _unlikely_ to change

What if we reframed feature flags to aim for this by default? Flags really shouldn't be highly unstable. The goal is to deliver something we've vetted and are quite confident to fulfill the need or fix an issue. These things are ready for people to explore _today_ with the unlikely possibility of API changes based on user feedback.

We could drop the `experimental` naming to solve part of the psychological aspect too, I think. Code behind flags isn't so much "experimental" changes, moreso just "breaking" changes. This would keep the convention down to just committed or uncommitted flags:

* `enable-*`
* `enable-v#-*`

alisonjoseph: > Like, the feature is stable and the API will likely not change anymore (but theoretically still could), however it's not committed to the next major version yet?

I agree that this should really be our default, and I would say has been for the most part with the flags we already have in place. We've just never outright defined that anywhere. 

I like the idea of dropping the `experimental` part of the flag name, agree that people will be more hesitant to use something with that in the name.

janhassel: I like that plan!

andreancardona (approved):
yayyy 😎  looks so good!

tay1orjones:
@alisonjoseph @janhassel this is updated and ready for a re-review

alisonjoseph (approved):

janhassel (approved):
Looks great!

tw15egan (approved):

PR #15979, opened 2024-03-15
fix(icons-react): export CarbonIconType and CarbonIconProps

### LINKED ISSUE #15974: [Bug]: `CarbonIconType` is not exported from `@carbon/react/icons`
### Package

@carbon/react, @carbon/icons-react

### Browser

_No response_

### Package version

1.52.0

### React version

_No response_

### Description

Since upgrading to `@carbon/react` ~~v1.52.0~~ v1.41.0 (or later), `CarbonIconType` cannot be imported from `@carbon/react-icons` (or directly from `@carbon/icons-react`).

image

I think the issue is that before ~~v1.52.0~~ v1.41.0 type definitions for `@carbon/icons-react` came from DefinitelyTyped which exported `CarbonIconType`. Now that `@carbon/icons-react` ships its own types<sup>1</sup>, DefinitelyTyped removed their support for that package<sup>2</sup>. 

Reference:
1. Types added to `@carbon/icons-react`: carbon-design-system/carbon#14714
2. Types removed from `@types__icons-react`: DefinitelyTyped/DefinitelyTyped#67176

### Reproduction/example

https://stackblitz.com/edit/github-qo3ucm?file=src%2FApp.tsx

### Steps to reproduce

1. Install `@carbon/react` v1.52.0.
2. Try to import `CarbonIconType` from `@carbon/react/icons`.
3. Hover over import and view error: `Module '@carbon/react/icons' has no exported member 'CarbonIconType'`.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #15974 

`CarbonIcon.d.ts` is already [included in the build](https://unpkg.com/browse/@carbon/icons-react@11.38.0/es/CarbonIcon.d.ts), but none of it is exported in the `index.d.ts`.

#### Changelog

**Changed**

- Modify `icon-build-helper` to include a line in the `index.d.ts` exporting everything from `CarbonIcon.d.ts`

#### Testing / Reviewing

- Pull down and build locally, the generated `index.d.ts` should include a line that properly resolves to the relative path of the `CarbonIcon.d.ts` file


tay1orjones:
@lewandom @metonym let me know what you think. Was there a reason we didn't want to export this from the entrypoint? _Should_ we be exporting this from the entrypoint for consumers to potentially rely on?

metonym (approved):
LGTM.

Purely from a stylistic point-of-view, I would prefer named exports over wildcard for clarity.

```ts
export { CarbonIconProps, CarbonIconType } from "./CarbonIcon";
```

lewandom (approved):
I have no issues with exporting these types,sounds like a good idea. I'd agree with @metonym that perhaps it's better to explicitly name exported types. Other than the syntax, it gives better control over what's exported.

alisonjoseph (approved):

tay1orjones:
Thanks for the reviews! I've updated to named exports instead of the wildcard

metonym (approved):

tw15egan (approved):

PR #15993, opened 2024-03-18
fix: flush state for accordion component

### LINKED ISSUE #15849: [Bug]: Accordion with flush alignment does not meet design spec
### Package

@carbon/react

### Browser

Safari

### Package version

1.51.1

### React version

_No response_

### Description

According to the design guidance, the Accordion titles are flush with the dividing lines when using the `isFlush` variant.

image

In the storybook, this not the case currently:

image

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-accordion--playground&args=isFlush:!true

### Steps to reproduce

Enable `isFlush` in the storybook's controls.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #15849 

Accordion with flush alignment does not meet design spec

**Changed**

- flush styles
- fix for skeleton state for accordion

#### Testing / Reviewing

Go to Playground for accordion component, enable flush property, border should come in shorter to match flush spec


andreancardona (approved):
tested and looks good to me!

alina-jacob:
Hi @riddhybansal great work! 💯 
Just one tiny observation 👀 

This is the intended specs for the Hover state of [Flush Accordion](https://carbondesignsystem.com/components/accordion/style#interactive-states)
image

The deploy preview link has the horizontal divider stand out in **hover state**, is there a way to blend it the way it works for the regular accordion? 
Also in the preview it looks like the chevron goes missing on **hover state**, not sure why, but could you take a look! 
The chevron up to chevron down interaction is also missing, maybe that's solved if the hover state issue is fixed.
image

tw15egan:
@alina-jacob it seems like the divider is still there on hover when not `isFlush`, but it is less noticeable since the divider and hover area have the same surface area.

 Screenshot 2024-03-21 at 9 58 27 AM Should this be removed in the normal, non `isFlush` variant as well? 

@riddhybansal it seems like we need to add a `transition` property to the pseudo element, as it is currently instantly appearing / disappearing and causing a flicker on hover

alina-jacob:
> @alina-jacob it seems like the divider is still there on hover when not `isFlush`, but it is less noticeable since the divider and hover area have the same surface area.
> 
> Screenshot 2024-03-21 at 9 58 27 AM Should this be removed in the normal, non `isFlush` variant as well?
> 
> @riddhybansal it seems like we need to add a `transition` property to the pseudo element, as it is currently instantly appearing / disappearing and causing a flicker on hover
> 
>  [  ](https://private-user-images.githubusercontent.com/11928039/315319449-87b42d07-c6c3-48ce-8aa9-77cee7483ab8.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTEzNzgyOTYsIm5iZiI6MTcxMTM3Nzk5NiwicGF0aCI6Ii8xMTkyODAzOS8zMTUzMTk0NDktODdiNDJkMDctYzZjMy00OGNlLThhYTktNzdjZWU3NDgzYWI4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMjUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzI1VDE0NDYzNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWYwYzAzNWUwYThhNTIzZjI0ODkzODVmOWFlODBhMTgwNzY2NjA5N2U1ZDcwZTkxMzAxMWM1ZjhhN2MyYmMzNjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.CXH7aYG-nkOsL3Oc8PIvt9MrfcN9UmqNAVlDukCE44Y) [ ](https://private-user-images.githubusercontent.com/11928039/315319449-87b42d07-c6c3-48ce-8aa9-77cee7483ab8.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTEzNzgyOTYsIm5iZiI6MTcxMTM3Nzk5NiwicGF0aCI6Ii8xMTkyODAzOS8zMTUzMTk0NDktODdiNDJkMDctYzZjMy00OGNlLThhYTktNzdjZWU3NDgzYWI4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMjUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzI1VDE0NDYzNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWYwYzAzNWUwYThhNTIzZjI0ODkzODVmOWFlODBhMTgwNzY2NjA5N2U1ZDcwZTkxMzAxMWM1ZjhhN2MyYmMzNjQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.CXH7aYG-nkOsL3Oc8PIvt9MrfcN9UmqNAVlDukCE44Y)

yes ✅  @tw15egan, @riddhybansal 
the horizontal divider should be removed/hidden during hover states in flush and default modes for the Accordion. 

here's a [Figma ref file](https://www.figma.com/file/SH2gjkLOP9WvwGFn319iBT/Flush-accordion-specs?type=design&node-id=0%3A1&mode=design&t=1uf2A4zvQOMK7kxq-1) that  @laurenmrice has created and here's [Accordion's style documentation](https://carbondesignsystem.com/components/accordion/style) from the Carbon website indicating the same.

image

riddhybansal:
@alina-jacob the issue regarding the hover states are fixed and the issue of divider for both the flush and non flush states are addressed in the same PR

alisonjoseph (changes requested):
Removing the border on hover is causing a layout shift. You may have to update the color of the border to match the background?


https://github.com/carbon-design-system/carbon/assets/2753488/84ae924d-fa03-4865-95a6-7b77ad8bb7f9

alisonjoseph:
Taking a closer look I'm still seeing a weird flash/flickr on hover, however I want to confirm with design that the way its hanging outside of the content area is correct. Or should the text align between the two variants with the divider lines moved in, vs. the hover/focus area pushed out. 

@alina-jacob or @carbon-design-system/design can you weigh in here. This is a screen shot with the padding we add for all our storybook stories removed. You can see that the isFlush variant hangs outside the normal content area. 
Screenshot 2024-03-26 at 12 29 09 PM




Compared with how we handle something similar with ContainedList
Screenshot 2024-03-26 at 12 34 00 PM

alina-jacob:
Hi @alisonjoseph, 
Based on the deploy preview, the flush execution seems okay.
Is it an issue if the hover/active/focus state hangs out of the content area from a dev perspective?

- Riddhi has pushed the text and icons to the left and right to meet the width of the divider.
- The contained list screenshot that you have shared, that has the line pushed in to meet the text and the icons.
(is one better than the other from a dev perspective, because design wise, the current way also works)

alisonjoseph:
My opinion is that we wouldn't want it to hang outside of the content area, and instead push in. @tw15egan @laurenmrice thoughts?

laurenmrice:
For context, this is the type of area in a UI where our users are using the flush accordion to achieve better vertical text alignment. The focus and hover states would hit the edges of the left panel container, and those states would extend outside of the divider borders between the rows. People should not mix the flush alignment with the default alignment within one accordion group. 

And yes, to Alison's point above, we do not want to remove the hang for the hover and focus states for this because it would create visual tension between the edges of the state containers and the text.

Screenshot 2024-04-01 at 10 24 56 AM

alisonjoseph:
@laurenmrice thank you so much for the example! That clears things up for me. 

@riddhybansal If you see the in-context example above from Lauren you'll see how it should display in a container. So you'll want to adjust the styles so that the appearance of the border shifts inward by 1rem on either side instead of the entire accordion shifting outward by 1rem. 

Currently the border is set on `.cds--accordion__item`, for this you'll have to remove that border and use psuedo-elements instead. You'll need to do something similar for the hover on `.cds--accordion__heading`. Something similar to below. Let me know if you get stuck/need to pair up. 
Screenshot 2024-04-01 at 11 30 41 AM

alisonjoseph:
@riddhybansal whats the status on this one? Let me know if you need to pair up.

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -63,6 +63,21 @@ $content-padding: 0 0 0 $spacing-05 !default;
     &:last-child {
       border-block-end: 1px solid $border-subtle;
     }
+
```
tw15egan, line 68: Can we add a transition to the `border-color` for the `accordion-item` so it matches the background change? 

L62:
```scss
transition: border-color motion(standard, productive) $duration-fast-02;
```

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -285,19 +302,75 @@ $content-padding: 0 0 0 $spacing-05 !default;
...
-  padding-inline-start: 0;
+.#{$prefix}--accordion--flush .#{$prefix}--accordion__item::after {
+  position: absolute;
+  display: block;
+  block-size: 100%;
+  border-block-start: 1px solid $border-subtle;
+  content: '';
+  inline-size: calc(100% - $spacing-05 - $spacing-05);
+  inset-block-start: -1px;
+  margin-inline-start: $spacing-05;
```
tw15egan: Same as above, can we add a transition so the hover styles match the background color change?

```suggestion
  margin-inline-start: $spacing-05;
  transition: border-color motion(standard, productive) $duration-fast-02;
```

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -285,19 +302,75 @@ $content-padding: 0 0 0 $spacing-05 !default;
...
+.#{$prefix}--accordion--flush .#{$prefix}--accordion__item:last-child {
+  border-color: transparent;
 }
 
-.#{$prefix}--accordion--flush .#{$prefix}--accordion__content {
-  padding-inline-start: 0;
+.#{$prefix}--accordion--flush .#{$prefix}--accordion__item::after {
+  position: absolute;
+  display: block;
+  block-size: 100%;
```
tw15egan: Does this need to be 100%, or can it just be 1px? Seems to just be acting as a pseudo border

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -285,19 +302,75 @@ $content-padding: 0 0 0 $spacing-05 !default;
...
+//   position: absolute;
+//   display: block;
+//   block-size: layout.size('height');
+//   border-block-start: none;
+//   content: '';
+//   inline-size: calc(100% - $spacing-05 - $spacing-05);
+//   margin-left: 1rem;
+//   top: 0;
+// }
+
```
tw15egan: Is this still needed?

packages/styles/scss/components/accordion/_accordion.scss
```diff
@@ -285,19 +302,75 @@ $content-padding: 0 0 0 $spacing-05 !default;
...
+// .#{$prefix}--accordion--flush .#{$prefix}--accordion__item {
+//   position: relative;
+//   border-color: transparent;
+
+//   &:last-child {
+//     border-color: transparent;
+//   }
+
+// }
```
tw15egan: Is this commented out code still relevant?

tw15egan (dismissed):
I noticed a few issues:

`isFlush` enabled, hovering over an expanded `AccordionItem` seems to make a border disappear




<br />
<br />
<br />

Issues when hovering over expanded `AccordionItem` when `isFlush` is not being used



Issues when hovering over the 2nd to last item when `isFlush` is not present

riddhybansal:
@tw15egan removing border for the hover state of flush or default states is one of the requirement mentioned above by @alina-jacob . Oh may be we can have dividers in the expanded state. Is that what you mean ?

tw15egan:
@riddhybansal yes, it seems like the divider is being removed when it is expanded and hovering over the content section

riddhybansal:
Hi @tw15egan I hope all the above issues have been resolved. I restricted blending of border when accordian is in expanded state. But I wonder if we could blend the top border even if its expanded and somehow restrict the blending when the content is hovered. If yes how can I restrict parent element to apply its hovering properties when child is being hovered, any idea ?

alisonjoseph (changes requested):
I'm seeing some changes to the original accordion with the border disappearing on hover.
Screenshot 2024-04-26 at 8 23 40 AM

and this is in this PR

Screenshot 2024-04-26 at 8 25 00 AM

Is this intentional?

I'm also seeing some alignment issues between flush and default when testing locally with them side by side. 
Screenshot 2024-04-26 at 8 26 05 AM

alisonjoseph:
@riddhybansal I pushed some updates to this PR, we can talk through in a webex if needed. 

I added a test story so we can view the flush variant side by side, I also updated/fixed skeleton states. (will need to remove test story before merging)

alisonjoseph (approved):

aagonzales (approved):
Oh wow lots of comments in this PR but it looks like everything has been addressed. Final demo is working as expected.

tay1orjones (approved):

PR #15995, opened 2024-03-18
docs(style): add guidance on avoiding magic numbers

tay1orjones:
#### Changelog

**Changed**

- updated style.md docs to include magic number guidance

#### Testing / Reviewing

- Am I missing anything?


docs/style.md
```diff
@@ -863,6 +863,31 @@ this project:
...
+If we were trying to apply a `padding-inline-end` to the input to ensure the
+input text does not flow behind the icon, we could add up the individual parts
+of this that use spacing tokens, contextual layout tokens, and potentially local
+tokens(variables) that will inherently explain what the number is composed of.
+
+```diff
+- padding-inline-end: to-rem(32px);
+
++ $icon-width: 16px;
++ padding-inline-end: calc(layout.density('padding-inline') + $icon-width)
```
jharvey10: I'd advocate for taking this even a step further. IMO the real strength of eliminating magic numbers and constants is the reusability aspect. Explanation and composition is helpful, but it's still prone to anti-patterns such as:
```scss
$my-obscure-thing: 24.5px;
// ...
some-attr: calc(layout.density('padding-inline') + $my-obscure-thing);
```
Wherein the number is given an identifier, but the identifier is meaningless outside of its immediate surroundings.

In the example from the style doc, `$icon-width` should likely be defined as a reusable constant in a more generic space. And the introduction of file-local primitive/scalar values should be discouraged, especially if they are never reused.

Ideally, the developer's mental model is:
1) Check the reusable/global constants for an appropriate one given what is trying to be accomplished.
2) If one exists, use it. If not, start a conversation with the team as to why no such value currently exists (perhaps challenge the way it was intended to be used in the first place).
3) Decide to either introduce a new constant to meet the need; or rework the code in question to use other constants (or perhaps none at all).

tay1orjones: Great points! I appreciate you pushing further on this. I wrote the icon-width thing blind as just a contrived example. You prompted me to actually look though, and sure enough we have a constant for it 🙃 

I agree that local constants should be avoided. Incorporating your mental model breakdown would be really helpful. Will make an update shortly.

jharvey10 (approved):
LGTM! 🙌🏼 🙌🏼

alisonjoseph (approved):
Thanks for putting this together!

PR #15996, opened 2024-03-18
fix(modal): place aria-hidden on sentinel nodes

### LINKED ISSUE #14203: [a11y]: Modal [all variants] - identifies visually hidden controls
### Package

@carbon/react

### Browser

Safari

### Package version

1.32.0

### React version

18.2

### Description

VO is reading "Focus sentinel" when arrow down after the "Add" button in all `Modal` variants

This [PR](https://github.com/carbon-design-system/carbon/pull/14156) fixed `ComposedModal`. It might be good taking a look.

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?args=&id=components-modal--with-state-manager&viewMode=story

### Steps to reproduce

Arrow down the `Modal` with the VO active

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Testing some feedback from the Carbon Accessibility Guild to place aria-hidden on the sentinel nodes to ensure they're not read by screenreaders when using virtual cursor mode.

~If this is successful,~ We've confirmed this does fix the issue, so we should port it to all other components that use sentinel nodes, which should fix https://github.com/carbon-design-system/carbon/issues/14203

- [x] Modal
- [x] ComposedModal
- [x] Notification
- [x] FloatingMenu (used by the legacy non-menu-based OverflowMenu)

You can test this change on the [modal story of the deploy preview](https://deploy-preview-15996--v11-carbon-react.netlify.app/?path=/story/components-modal--default).


alisonjoseph (approved):

tw15egan:
@tay1orjones looks like tests may need to be updated?

tay1orjones:
@tw15egan I think I'm remembering now why we didn't do this before ... with the `aria-hidden` on there the checker reports a violation:

`Element "button" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'`

While this approach might be allowable in our view, we can't ship something that's going to pop violations in consumer environments. Eventually the checker [may be updated](https://ibm-studios.slack.com/archives/G01GCBCGTPV/p1711130558677409?thread_ts=1710791245.030649&cid=G01GCBCGTPV) to handle this use case or provide an exception but for now I think we're blocked on this one.

Issue #13672, opened 2023-04-26
[a11y]: Inline CodeSnippet copy to clipboard notification is not read by JAWS


tay1orjones:
### Package

@carbon/react

### Browser

Firefox

### Operating System

Windows

### Package version

1.27.0

### React version

v18

### Automated testing tool and ruleset

n/a

### Assistive technology

JAWS

### Description

[Reported internally](https://ibm-studios.slack.com/archives/C054WK4ES5S/p1682510942552969), when testing the inline code snippet with JAWS,

> After activating the "Copy to clipboard" button, The content is copied. However there is no feedback provided by SR. Hence the user may wonder whether the content copied or not.
>
> SR is expected to notify the user about the success of the copy function instantaneously
>
> When there is no feedback of an action that is triggered, Screen reader users would not know whether the action took place or not.

### WCAG 2.1 Violation

4.1.3 Status Messages

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-codesnippet--inline

### Steps to reproduce

Load the story and use JAWS to see how the "copied to clipboard" notification is not read.

andreancardona:
This issue is fixed as seen in this recording https://github.com/carbon-design-system/carbon/pull/16022

PR #16022, opened 2024-03-21
fix: testing aria codesnippet

### LINKED ISSUE #13671: [a11y]: Inline CodeSnippet content is not read by JAWS
### Package

@carbon/react

### Browser

Firefox

### Operating System

Windows

### Package version

1.27.0

### React version

v18

n/a

### Assistive technology

JAWS

### Description

[Reported internally](https://ibm-studios.slack.com/archives/C054WK4ES5S/p1682510942552969), when testing the inline code snippet with JAWS, 

> the screen reader does not identify the code snippet that is to be copied. Hence the Blind user would not know what content is to be copied
>
> Note: in Vo iPhone when the focus reaches, , SR announces "copy to clip board" button. After activation, the label changes to "Node-v". this confuses the user.
>
> SR should identify the code snippet along with the "code" role so that the content is programmatically identified

One additional recommended potential course of action included separating the content from the button and providing an `aria-describedby`. We could also review this with the Carbon Accessibility Guild to determine some other ways we could potentially tackle this.

It's possible this is related to or was introduced with https://github.com/carbon-design-system/carbon/pull/13460

### WCAG 2.1 Violation

1.3.1 Info and Relationships 

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-codesnippet--inline

### Steps to reproduce

Load the story and use JAWS to see how the content is not read

andreancardona:
closes #13671 

you can hear that this #13672 is also fixed here 

Tested a couple of different aria combinations - this seems to work

## TESTING 

make sure there no regressions in VO

https://github.com/carbon-design-system/carbon/assets/32720851/c2ee030f-e6c0-4511-bba4-bb0dbedf08a6


packages/react/src/components/CodeSnippet/CodeSnippet.js
```diff
@@ -187,12 +187,12 @@ function CodeSnippet({
         {...rest}
         align={align}
         onClick={handleCopyClick}
-        aria-label={deprecatedAriaLabel || ariaLabel}
+        aria-label={ariaLabel}
```
alisonjoseph: Did you mean to remove this? I think we may still want it in case anyone was using the `ariaLabel` prop?

alisonjoseph (approved):

tay1orjones (approved):

Issue #15919, opened 2024-03-07
[a11y]: dismissible tab has focus handling issue after activation; also last two items should not be dismissible


mbgower:
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-tabs--dismissable-with-icons#storybook-preview-wrapper

### Automated testing tool and ruleset

n/a

### Description

When pressing Delete to dismiss a tab item from the Dismissible tab (with or without icons) https://react.carbondesignsystem.com/?path=/story/components-tabs--dismissable-with-icons#storybook-preview-wrapper,
the focus is lost from all items. 
Expectation is that focus should be place on the prior item in the list or somewhere else on the page, but there is nothing that takes focus.

### WCAG 2.1 Violation

Focs Visible

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tabs--dismissable-with-icons#storybook-preview-wrapper

### Steps to reproduce

1. Open https://react.carbondesignsystem.com/?path=/story/components-tabs--dismissable-with-icons#storybook-preview-wrapper or https://react.carbondesignsystem.com/?path=/story/components-tabs--dismissable
2. Navigate to the main area by keyboard (I suggest you tab once, Enter on Skip to canvas and tab a couple more times)
3. Once focus is on the tab list, use the arrow key to move to an item other than the first one
4. Press delete; the item disappears, and focus also disappears (the focus being placed elsewhere is expected behaviour).
5. Press Tab once. The Reset button now has focus

The focus handling has a few considerations: if I delete the first item in the list, where should the focus go? If I delete the final item? I believe the interactions decided on is that the focus will always be placed on the preceding item, and when there is not a preceding tab item, the focus would be placed on the next tab item.
A second issue in this implementation is that there should never be a case where the final item is deleted. IMO, the dismiss mechanism should cease to exist (or get disabled) once there are only 2 tab items left, as according to the [documentation](https://carbondesignsystem.com/components/tabs/usage/#:~:text=They%20always%20contain%20at%20least%20two%20items), tabs "always contain at least two items".

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
[...]

PR #16033, opened 2024-03-21
fix(tabs): refactor dismissable tab html

### LINKED ISSUE #14820: [a11y]: Tabs [dismissible] - unable to dismiss in iPhone
### Package

@carbon/react

### Browser

Safari & iPhone

### Package version

@carbon/react@1.38.0

Manual QA

### Assistive technology

JAWS & VO

### Description
"No way of interaction available for VO users to delete A tab in iPhone.

NOTE: In call its understood there is delete icon which helps the visual users to delete the TAB but its not read by Voiceover."

### Recommended Fix
Provide Delete Name and Button Role for Delete ICON and make it keyboard operable

When the standard interaction is not possible, user is unable to accomplish A task.

### WCAG 2.1 Violation

4.1.2 Name, Role, Value

alisonjoseph:
Closes #14820
Ref #https://github.com/carbon-design-system/carbon/issues/15919

Previously you were only able to dismiss a dismissable tab with the delete key or mouse. This leaves voiceover users on an iphone unaware the dismiss icon even existed. 

This is a complete refactor of the html for dismissable tabs, it changes the icon to a button, and moves it outside of the containing tab button and into the tab order. This allows both keyboard and voiceover users to tab to the icon and dismiss the tabs.

#### Changelog

**Changed**

- Convert the dismiss icon to a button, move outside of the tab button
- Style updates to account for updated html structure
- Update text for screen-readers to say "remove _tab name_ tab" instead of "close tab"
- Updated tests 
- Update focus state after tab is removed

#### Testing / Reviewing

Make sure there are no regressions with any of the tab variants. Be sure to check dismissable in both regular and contained. 

Check that the tab order is correct and that the dismiss icon is read on voice over.


tay1orjones:
Thanks for digging into this! I think it looks good from a code review standpoint.

The landscape around this is kind of a mess. [This stackoverflow post](https://stackoverflow.com/questions/78071123/aria-tablist-role-that-contains-elements-other-than-tabs) explains it pretty well. There's also a big thread about it that hasn't been resolved in https://github.com/w3c/aria/issues/1440

I think this technically doesn't follow WAI-ARIA 1.2 because [a `tablist` should contain a list `tab` elements](https://www.w3.org/TR/wai-aria/#tablist:~:text=tablist%20role-,A%20list%20of%20tab%20elements%2C%20which%20are%20references%20to%20tabpanel%20elements.,-To%20be%20keyboard). 

<details>
  <summary>It adds an additional "Incomplete" entry from `accessibility-checker` </summary>
  
  image

</details>



@tombrunet when choosing between fixing this or following the spec, I think we should side with the user and fix it regardless of what the spec says, but what do you think?

Also, this contains a change that could break tests or custom functionality in consuming applications because of the shift in the DOM. Despite that, I would support shipping this to all users in our next minor without a feature flag due to the severity of the bug.

mbgower:
So, what we're effectively doing is saying that there is an item in the tab order between the tablist and the tab panel, which is "Dismiss current tab". It happens to be shown inside the tab item, but it is effectively outside the tablist.

I wonder if one approach here is to treat this somewhat like the skiptomain function. Rather than put the focus to the X in the tab item, we make a "Delete tab item" that appears to the right of the tablist after a user tabs out of the tablist. Does that solve anything? It suggests to me it would let us solve the discoverability and actionability requirement AND would not violate ARIA. @tombrunet ?

tombrunet:
@mbgower The largest problem with that approach is that if you change reading modes on the close button, you are out of context. If I go to the close button and then ask to read the previous word, I expect to hear the name of the tab I'm on, not the name of the last tab in the list. 

Looking at some other apps, it does seem that handling of these tab widgets is changing. VS Code has a behavior similar to what Carbon is proposing here (though the close button is only visible for the active tab and on mouse hover). Chrome is actually making every tab widget and close button tabbable.

mbgower:
@alisonjoseph, I've disrupted your fix with my comments, so I wanted to put in an epilogue on this.

I have a caution on the fix which needs to be addressed, as well as some suggestions on future changes, but once that issue is addressed, this looks good to go from an accessibility problem. Details below.

1) ISSUE: It looks like you've implemented this not just on the dismissible tabs, but on ALL the version of Tabs in storybook (default, contained, etc). That shouldn't be happening; it should only exist on the dismissible ones. In the screenshot, you can see that Jaws is focused on a small region to the upper right of the Monitoring tab which has nothing there visually. It is announcing this as "close tab button"
image

2) The implementation on the actual dismissible tab works. In this screen shot, you can see that, after the user presses Tab,  Jaws has properly put focus on the X. Good.
image

3) As I documented in the original issue, ARIA is planning a new attribute aria-actions, which should allow this approach to be possible AND pass current ARIA guidance. @tombrunet  and I have both reviewed your refactor and think what you've proposed not only aligns with some vendor approaches, but also seems to align with this direction of aria. Tom's team will figure out what to do with the Checker rules in regard to this version of the dismissible tab component, since it will currently fail a rule since it does not align with the current aria spec. But it's important to emphasize that it seems to _work_ fine for keyboard and screen reader user. 

4) I would like to suggest that making the message specific to the tab would be useful (i.e., instead of saying "Close tab button" it would say "Close Dashboard tab").

5) I also wonder if "Close" is the correct verb to use in this tooltip/message. "Remove" might make more sense, or even "Delete"? This is question for the content designer, I guess. My slight concern is that since a tab does not normally close (or dismiss, for that matter), a user may be unfamiliar with the operation or notion.

alisonjoseph:
@mbgower thanks for the detailed feedback, I believe I've addressed all of your concerns if you could take another look. 

1. It should no longer be announcing the button all on tab variants.
3. Focus should now go to the next tab after removing a tab, and not to the next close button. 
4. Close buttons should now have a specific message and announce "Remove Dashboard tab" instead of "Close tab"

mbgower:
All these things have been accomplished, nice:

- It should no longer be announcing the close button all on tab variants.
- Focus should now go to the next tab after removing a tab, and not to the next close button.
- Close buttons should now have a specific message and announce "Remove Dashboard tab" instead of "Close tab"

One thing I noticed is that when the last _active_ tab is deleted (Activity tab), things get a little odd, with either the focus going to the disabled 4th Settings delete button, or focus going up the page. 
image
I'd check with @tay1orjones if he remembers the approach we previously discussed about where to place focus in such a situation. I think if it's the last active item and there is a prior item it makes sense to put it there, and that applies if there are no items left at all (i.e., focus would go to Reset). I think that is the basic pattern Josh, Taylor and I discussed many moons ago about what would happen in a table with deletable rows, and the same seems to make sense here as well.

alisonjoseph:
I think this is ready for review now.

@mbgower I've updated it so the previous tab should now take focus after removing a tab.

mbgower:
> @mbgower I've updated it so the previous tab should now take focus after removing a tab. Just wanted to double check that we wouldn't want the active tab to take focus vs the previous one?

I'll have a look, but wanted to clarify that I wasn't saying that each time a tablist item is deleted, the focus should go to the prior tablist item; I think 'next tab item, where available' is a good approach, and supports the APG previously quoted.

I was specifically talking about the situation where the last item in the tablist is deleted  (or the last _active_ item in a tablist). In both cases, I think it probably should go to the prior tablist item.

alisonjoseph:
@mbgower ohhh I completely misread that, let me update this.

alisonjoseph:
@mbgower ok, focus states are updated.

mbgower:
That's working well with JAWS. I also confirmed it is usable and understandable with VO on iOS

andreancardona (approved):
Also tested with Jaws and looks good to me! 🎉

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -849,6 +850,32 @@ const Tab = forwardRef<HTMLElement, TabProps>(function Tab(
   const handleClose = (evt) => {
     evt.stopPropagation();
     onTabCloseRequest?.(index);
+
+    // set focus to corret tab after removing tab
```
alisonjoseph:
```suggestion
    // set focus after removing tab
```

tay1orjones (approved):
LGTM!

PR #16043, opened 2024-03-25
fix(15510): Update Interface to Include href Prop in SwitcherItem Type

### LINKED ISSUE #15510: [Bug]: TS error: href prop does not exist in SwitcherItem type
### Package

@carbon/react

### Browser

Chrome

### Package version

1.47

### React version

18.2

### Description

See screenshot below. Existing code still works at runtime. Worked around by moving the `href` prop up one line so it could be ignored with `@ts-ignore`.

image

### Reproduction/example

Any use of SwitcherItem with the href prop will show this.

### Steps to reproduce

Any use of SwitcherItem with the href prop will show this.

### Suggested Severity

### Application/PAL

CDD Hub

### Code of Conduct

2nikhiltom:
Closes #15510

Added optional 'href' prop to `BaseSwitcherItemProps` for link URLs.

**New**
Added href prop to `BaseSwitcherItemProps`

#### Testing / Reviewing
```tsx
<Switcher aria-label="Switcher Container" expanded={true}>
  <SwitcherItem aria-label="Link 1" href="#">
    Link 1
  </SwitcherItem>
</Switcher>
```

Test the usage of `SwitcherItem` with the `href` prop in a TypeScript environment within a `Switcher` component. Verify that the  TypeScript compiler should not throw error and the component renders without errors and correctly applies the provided `href` attribute


andreancardona (approved):

tay1orjones (changes requested):
This also needs: 
- [x] To be added to the PropTypes with the same comment
- [x] Pulled out of `...rest` to instead be a named prop within the component, and placed onto `Link`

tay1orjones (approved):
Awesome!

PR #16067, opened 2024-03-26
fix(structured-list): remove aria-busy from row

janhassel:
Currently, each `StructuredListRow` has `aria-busy="true"`, announcing that the content within is being updated. Since that is not the case by default though, the attribute should not be present.

#### Changelog

**Removed**

- Removed `aria-busy="true"` from `StructuredListRow`

#### Testing / Reviewing

- Automated testing
- Inspect DOM in storybook


andreancardona (approved):
hmm wondering why we had that on there 👀

tay1orjones (approved):

PR #15275, opened 2023-11-27
docs(examples): migrate custom data table state manager to vite

alisonjoseph:
Closes #15201

Migrate custom-data-table-state-manager example to vite

#### Changelog

**New**

- `packages/react/examples/custom-data-table-state-manager-vite`

**Removed**

- `packages/react/examples/custom-data-table-state-manager`

#### Testing / Reviewing

Check that example is working

PR #16075, opened 2024-03-26
fix: correct imports for custom data table state manager example

tay1orjones:
Follow up to https://github.com/carbon-design-system/carbon/pull/15275 - these invalid imports prevent the example from running within Stackblitz.

#### Changelog

**Changed**

- changed imports from `carbon-components-react` to `@carbon/react`


tay1orjones:
Here's an example stackblitz with the corrected imports that runs: https://stackblitz.com/edit/github-bk8d6m?file=src%2Fcomponents%2FCustomDataTable.jsx

alisonjoseph (approved):

andreancardona (approved):
looks good!

PR #16076, opened 2024-03-26
fix(multiselect): set optional chaining for undefined multiselect

### LINKED ISSUE #16052: [Bug]: Crash in MultiSelect with undefined selectedItems
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.53.0

### React version

v18.2.0

### Description

This is admittedly user error, but it seems like it would be a good idea to protect against it. 
If `selectedItems` has an array with `undefined` in it, [this PR](https://github.com/carbon-design-system/carbon/pull/15633) will try to read `text` from undefined and crash with the stack below:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'text')
    at @carbon_react.js?v=72b18db9:100065:90
    at Array.map (<anonymous>)
    at @carbon_react.js?v=72b18db9:100065:71
```

I hit it b/c we had some bad test code, but figured it would be a good idea to just add some optional chaining in that MultiSelect code so it doesn't crash. 

### Reproduction/example

https://stackblitz.com/edit/github-caiwsw?file=src%2FApp.jsx

### Steps to reproduce

Create a multiselect
in `selectedItems`, have an item that is undefined or null.

### Suggested Severity

### Application/PAL

IBM Cloud Satellite

### Code of Conduct

jesnajoseijk:
Closes #16052

Added optional chaining for undefined multi select value

#### Changelog

**New**

- {{new thing}}

**Changed**

- multiselect.tsx
- multiselect-test.js

**Removed**

- {{removed thing}}

#### Testing / Reviewing

{{MultiSelect component}}


packages/react/src/components/MultiSelect/tools/sorting.js
```diff
@@ -16,7 +16,7 @@
  * @returns {number}
  */
 export const defaultCompareItems = (itemA, itemB, { locale }) =>
-  itemA.localeCompare(itemB, locale, {
+  itemA?.localeCompare(itemB, locale, {
```
tay1orjones: Could you explain why this is necessary? My understanding is that this code is called via `sortItems`, which always runs each item through an `itemToString` function (`defaultItemToString`) that should return `''` if the item is undefined

https://github.com/carbon-design-system/carbon/blob/2ebd2cbac7ffa53492a75f655b374a4f37c73115/packages/react/src/components/MultiSelect/tools/sorting.js#L43-L45

jesnajoseijk: @tay1orjones On the mentioned scenario in [16052](https://github.com/carbon-design-system/carbon/issues/16052),we are getting the below mentioned error as well.
Screenshot 2024-03-28 at 10 40 31 AM

tay1orjones: Okay, a few suggestions would be to:

* Handle `undefined` upstream from this - could we filter out undefined values from `initialSelectedItems` within MultiSelect?
* Write a test for MultiSelect that reproduces the error, to indicate that whatever fix we land on here actually works as intended

jesnajoseijk: @tay1orjones undefined items are filtered out from the items list  and the associated test case is added in the new PR

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -37,6 +37,29 @@ describe('MultiSelect', () => {
...
+    const { container } = render(
+      <MultiSelect
+        id="custom-id"
+        label={label}
+        items={items}
+        itemToString={(item) => (item ? item.text : '')}
+      />
+    );
+    // eslint-disable-next-line testing-library/prefer-screen-queries
+    const labelNode = getByText(container, label);
```
tay1orjones: This should be possible with the allowed 
```js
screen.getByRole('combobox', { name: label })
```

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -37,6 +37,29 @@ describe('MultiSelect', () => {
...
+
+    await userEvent.click(labelNode);
+
+    // eslint-disable-next-line testing-library/prefer-screen-queries
+    expect(getByText(container, 'joey')).toBeInTheDocument();
+    // eslint-disable-next-line testing-library/prefer-screen-queries
+    expect(getByText(container, 'johnny')).toBeInTheDocument();
+    // eslint-disable-next-line testing-library/prefer-screen-queries
+    expect(getByText(container, 'undefined')).not.toBeInTheDocument();
```
tay1orjones: The queries here should also be possible with the allowed 
```js
screen.getByRole('option', { name: 'joey' })
```

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -37,6 +37,29 @@ describe('MultiSelect', () => {
...
+    const items = [{ text: 'joey' }, { text: 'johnny' }, { text: undefined }];
+    const label = 'test-label';
+    const { container } = render(
+      <MultiSelect
+        id="custom-id"
+        label={label}
+        items={items}
+        itemToString={(item) => (item ? item.text : '')}
+      />
+    );
```
tay1orjones: With those queries sorted you should be able to remove the container

```suggestion
    render(
      <MultiSelect
        id="custom-id"
        label={label}
        items={items}
        itemToString={(item) => (item ? item.text : '')}
      />
    );
```

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -373,14 +373,29 @@ const MultiSelect = React.forwardRef(
       selectedItems: selected,
     });
 
+    // Filter out items with an object having undefined values
+    const filteredItems = items.filter((item) => {
+      if (typeof item === 'object' && item !== null) {
+        for (const key in item) {
+          if (
+            Object.prototype.hasOwnProperty.call(item, key) &&
```
tay1orjones: `hasOwn()` is a bit more intuitive and should be functionally the same I think

```suggestion
            Object.hasOwn(item, key) &&
```

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -373,14 +373,29 @@ const MultiSelect = React.forwardRef(
...
+            item[key] === undefined
+          ) {
+            return false; // Return false if any property has an undefined value
+          }
+        }
+      }
+      return true; // Return true if item is not an object with undefined values
+    });
+
```
tay1orjones: I think it would make sense for this to be in a `useMemo` with `items` in the dependency array so it doesn't filter on every render.

tay1orjones (changes requested):
Pretty close here! Just a couple other things I noticed

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
jesnajoseijk: @tay1orjones  shortened this to just screen.getByRole('combobox') becuase name is coming in as  empty string and giving error as  :
TestingLibraryElementError: Unable to find an accessible element with the role "combobox" and name "test-label"

    Here are the accessible roles:

      combobox:

      Name "":
      <button
        aria-activedescendant=""
        aria-controls="downshift-:r0:-menu"
        aria-expanded="false"
        aria-haspopup="listbox"
        aria-labelledby="downshift-:r0:-label"
        class="cds--list-box__field"
        id="downshift-:r0:-toggle-button"
        role="combobox"
        tabindex="0"
        type="button"
      />

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -37,6 +37,27 @@ describe('MultiSelect', () => {
...
+
+    const labelNode = screen.getByRole('combobox');
+    await userEvent.click(labelNode);
+
+    expect(screen.getByRole('option', { name: 'joey' })).toBeInTheDocument();
+    expect(screen.getByRole('option', { name: 'johnny' })).toBeInTheDocument();
+    expect(
+      screen.queryByRole('option', { name: 'undefined' })
+    ).not.toBeInTheDocument();
```
tay1orjones, line 59: We also have a global setting that fails tests if they call console.warn/error, so this test will definitely cover the old behavior you're fixing in this PR.

tay1orjones (approved):
Thanks for tackling this!

tw15egan (approved):
LGTM, thanks for contributing!

PR #15349, opened 2023-12-07
[Bug] - Sidenav menu collapse

guidari:
Closes #14902

The menu was not being opened because there was an if else that was avoid that had a conflict with the `onClick` in the button.
By removing the collappse started working.

`onClick={() => {setIsExpanded(!isExpanded)}}`

#### New

- Added tests to make sure the collapse functionality is covered.

#### Changed

- Added numbers to the links so we can get the `SideNavMenuItem` in the playwright.

#### Testing / Reviewing

Check if the stories that uses the `SideNavMenu` is working both in mobile and desktop view.

PR #16096, opened 2024-03-29
fix: collapse sidenav menus when side nav is closed

### LINKED ISSUE #15900: [Bug]: SideNavMenu remains expanded, when Rail SideNav collapses
### Package

@carbon/react

### Browser

Edge

### Package version

1.51.0

### React version

18.2.0

### Description

SideNavMenu remains expanded, when Rail SideNav collapses

image

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-ui-shell-sidenav--side-nav-rail-w-header

### Steps to reproduce

1. Hover cursor over SideNav for it to expand
2. In expanded SideNav expand first SideNavMenu
3. Move cursor away from SideNav for it to collapse into rail
4. Now you should see that there is additional empty space between first item and second item in SideNav

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #15900 

Submenus should close when side nav closes

#### Changelog

**Changed**

- Style changes

#### Testing / Reviewing

Go to storybook , Open Side Nav -> Open Submenu -> Close side nav ,  sub menu  should collapse autromatically.


tay1orjones (changes requested):
Generally `display: none` can't be used because it [removes the element(s) from the accessibility tree](https://developer.mozilla.org/en-US/docs/Web/CSS/display#accessibility_concerns) and so it won't be read by screenreaders.  

Instead of a style-related fix, would it make sense to modify the internal state of SideNavMenu to account for the change in the parent SideNav?

This bug was introduced https://github.com/carbon-design-system/carbon/pull/15349, my guess is the fix for this would be something similar to what was removed, but without reintroducing the bug it was fixing.

packages/react/src/components/UIShell/SideNavMenu.tsx
```diff
@@ -94,6 +95,18 @@ const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
       [customClassName as string]: !!customClassName,
     });
 
+    if (isSideNavExpanded === false && isExpanded === true && isRail === true) {
```
tay1orjones:
```suggestion
    if (!isSideNavExpanded && isExpanded && isRail) {
```

packages/react/src/components/UIShell/SideNavMenu.tsx
```diff
@@ -94,6 +95,18 @@ const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
...
+      setIsExpanded(false);
+      setPrevExpanded(true);
+    } else if (
+      isSideNavExpanded === true &&
+      prevExpanded === true &&
+      isRail === true
+    ) {
```
tay1orjones:
```suggestion
    } else if (isSideNavExpanded && prevExpanded && isRail) {
```

tay1orjones (changes requested):

tay1orjones (approved):

guidari (approved):
LGTM! 🚀

PR #16107, opened 2024-04-01
Fixed export for Interactive Tags

guidari:
It was reported here in the [PR](https://github.com/carbon-design-system/carbon/pull/15808#issuecomment-2024563220) that the Interactive tags wasn't exposed in `@carbon/react`, probably because it wasn't exported in the Tag `index.ts` and in the root `index.js`

#### Testing / Reviewing

- Wait deploy on npm to test on Stackblitz


tay1orjones (approved):

andreancardona (approved):
😎

PR #16112, opened 2024-04-02
SelectableTile change event returns current state, name

### LINKED ISSUE #11052: [Feature Request]: ⌨️ Selectable tile component should return name/value on click
### Summary

Selectable tile component does not return a meaningful event on click/change. 

Considering Selectable tile accepts name & value fields as props (similar to radio button), I expect to receive checked (state), value/name, along with the react event. I expect this component to have a similar return type to checkbox or radio tiles onChange functions.

Additionally, I think it would be great if Selectable Tile is offered as a standalone component (Current state) as well as a multi select (Similar to radio tile which is wrapped in tile group). This way, the `TileGroup` can manage the state of the child selectable tiles and return an array on change.

### Justification

This would add more consistency to the on change/click events. Also, the event target varies based on where the user clicks (Label or the checkbox SVG), making it difficult to retrieve the target.value/name. 

As for a multiselect wrapper component around the selectable component feature, it would be really nice to have. It would allow us to get a list of all checked tiles and their values/names.

_No response_

### Required functionality

_No response_

_No response_

### Available extra resources

_No response_

### Code of Conduct

Jawahars:
Closes #11052 

OnChange event of SelectableTile currently returns only the event target. The event target could be inner elements or svg. It is better to return the State of the SelectableTile, name and the event. This will be consistent with TileGroup as well.

#### Changelog

**New**

- SelectableTile's onchange returns the State of the SelectableTile, name and the event

**Changed**

- OnChange event of SelectableTile currently returns only the event target

**Removed**

- {{removed thing}}

#### Testing / Reviewing


Jawahars:

Jawahars:
recheck

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -479,7 +479,7 @@ export const SelectableTile = React.forwardRef<
 
   function handleChange(event) {
     setIsSelected(event.target.checked);
-    onChange(event);
+    onChange(isSelected, name, event);
```
tay1orjones: Sometimes changing the signatures of callbacks like this can break folks relying on the exact signature. One thing to help avoid this would be to keep `event` in the same ( first) parameter position.

```suggestion
    onChange(event, isSelected, name);
```

tay1orjones (changes requested):
Thanks for taking a look at this! Just one small change and then I think this is good to go

alisonjoseph (approved):

tay1orjones (approved):

tay1orjones (approved):

PR #16047, opened 2024-03-25
refactor(ContainedList): add TypeScript types

Gururajj77:
Closes #16019 

Added typescript types to ContainedList and ContainedListItem

#### Changelog

**New**

Added `labelText` prop to ContainedList story for handling typescript check

**Changed**

- {{change thing}}

**Removed**

- Renamed ContainedList and ContainedListItem to files to tsx, added interfaces for the props.

#### Testing / Reviewing

No additional comments on testing.

PR #16115, opened 2024-04-02
refactor(containedlist): added typescript types to containedlist

### LINKED ISSUE #16019: Add TypeScript types to ContainedList and ContainedListItem

Gururajj77:
Closes #16019 

Added typescript types to ContainedList component

#### Changelog

**New**

- added interfaces for props to ContainedList and ContainedListItem

**Changed**

- Renamed files from .js to .tsx

#### Testing / Reviewing

This new PR holds the review comments from #16047 

No new observations on testing.


packages/react/src/components/ContainedList/ContainedList.tsx
```diff
@@ -5,14 +5,51 @@
...
+
+  /**
+   * The kind of ContainedList you want to display
+   */
+  kind?: 'on-page' | 'disclosed';
+
+  /**
+   * A label describing the contained list.
+   */
+  label: ReactNode;
```
tay1orjones: This accepts a string too, right?

```suggestion
  label: string | ReactNode;
```

packages/react/src/components/ContainedList/ContainedList.tsx
```diff
@@ -5,14 +5,51 @@
...
-import React from 'react';
+import React, { ReactNode } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { LayoutConstraint } from '../Layout';
 import { useId } from '../../internal/useId';
 import { usePrefix } from '../../internal/usePrefix';
 
-const variants = ['on-page', 'disclosed'];
+const variants: ('on-page' | 'disclosed')[] = ['on-page', 'disclosed'];
```
tay1orjones: Is there a way to use this root definition for both the proptypes and the TS interface? Something like this maybe

```js
const variants = ['on-page', 'disclosed'] as const;

// I think this will be typed as 'on-page' | 'disclosed' because the typeof operator returns the inferred type?
type Variants = typeof variants[number]; 


// ...then below in the interface:

kind?: Variants;
```

packages/react/src/components/ContainedList/ContainedList.tsx
```diff
@@ -76,9 +113,14 @@ function ContainedList({
...
-    action?.type?.displayName
-  );
+  const isActionSearch =
+    React.isValidElement(action) &&
+    (typeof action.type === 'string'
+      ? ['Search', 'ExpandableSearch'].includes(action.type)
+      : typeof action.type === 'function' &&
+        'displayName' in action.type &&
+        (action.type.displayName === 'Search' ||
+          action.type.displayName === 'ExpandableSearch'));
```
tay1orjones: Is there any way to simplify this? It's really dense and hard to read. Is there another spot in the codebase we validate/check displayName/type?

tay1orjones (changes requested):
Just a couple small things I noticed

packages/react/src/components/ContainedList/ContainedList.tsx
Gururajj77: Yes, I overlooked it, will make the changes.

packages/react/src/components/ContainedList/ContainedList.tsx
Gururajj77: I will to simplify this and make it more readable.

alisonjoseph (approved):

tw15egan (approved):
LGTM 👍 ✅

tay1orjones (approved):

PR #16132, opened 2024-04-04
test(e2e): add `toContainAStory` assertion to `visitStory`

tay1orjones:
This adds a new assertion to `visitStory` that checks to make sure an element with the `.cds--layout` class is present. The storybook config has a global decorator that applies this class via `<Layout>`

https://github.com/carbon-design-system/carbon/blob/f504b2e9f1656e876e74c27a95be8f3f58d967bc/packages/react/.storybook/preview.js#L344

This should ensure we avoid a bug where if you give `visitStory` invalid parameters (a wrong `id`, `component`, etc) the url won't point to a story but storybook still resolves it and gives their error screen:

image

Huge thanks to @matthewgallo for bringing this up!

#### Changelog

**New**

- Add `toContainAStory` assertion to `visitStory`

#### Testing / Reviewing

- You can view the error this produces by pulling this down locally, changing the name of a story in any .stories.js file, and running `yarn playwright test`


tay1orjones:
This fix has actually found 68 errors! It's why ci is failing. I'll work on fixing them

tay1orjones:
Looked at this today with @Gururajj77 and noticed that when an `id` of a story is updated to be correct, there's a high likelihood that we then see accessibility violations that need to be fixed.

Instead of fixing those accessibility errors in this PR, we should skip those tests to focus on getting this PR merged with the `toContainAStory` assertion to protect all the existing passing tests.

So the steps for this PR are now going to be:

1. Correct the `id` for any test that fails the `toContainAStory` assertion
2. If correcting that `id` results in new accessibility violations on that story, mark that test as `test.skip` for now
3. Repeat steps 1 and 2, going through all the failures of the `toContainAStory` assertion 
4. Once all tests pass the `toContainAStory` assertion, make a new issue of all the tests that were `test.skipped` in this PR. We'll go fix those new violations one by one in follow up PRs.

Gururajj77:
opened new PR with changes, #16582

Issue #16009, opened 2024-03-19
Properly deprecate PopoverAlignment type values


tay1orjones:
In https://github.com/carbon-design-system/carbon/pull/14654 all Popover-based components were unified to use new values for the `align` prop.

The old values need to be properly deprecated so we can ensure consumers know to stop using them.

Pull the old values out into their own type, mark it as deprecated via jsdoc
Update PopoverAlignment to be a union of the new values and the new deprecated type
Write a custom validator function for proptypes that pops a warning when a deprecated value is used
Refactor all usages where  `PopoverAlignment` is used to normalize `tooltipPosition` with `tooltipAlignment` such as in Button

```[tasklist]
- [ ] https://github.com/carbon-design-system/carbon/issues/16326
```

2nikhiltom:
closed via https://github.com/carbon-design-system/carbon/pull/16346

PR #16149, opened 2024-04-08
fix: showPasswordLabel and hidePasswordLabel not working

### LINKED ISSUE #15986: [Bug]: [TextInput.PasswordInput] showPasswordLabel and hidePasswordLabel not working
### Package

@carbon/react

### Browser

Chrome

### Package version

1.53.0

### React version

18.2.0

### Description

In TextInput.PasswordInput component `showPasswordLabel` and `hidePasswordLabel` not working. 

### Reproduction/example

https://stackblitz.com/edit/github-pe88wq?file=src%2FApp.jsx

### Steps to reproduce

https://stackblitz.com/edit/github-pe88wq?file=src%2FApp.jsx

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

preetibansalui:
Closes #15986 

ShowPasswordLabel and HidePasswordLabel props were not working.

#### Testing / Reviewing

1. Create a PasswordInput component 
2. hover on eye icon on right side of password input field.
3. It should display text we assign to prop 'showPasswordLabel'
4. Now click on that eye icon.
5. on hover it should display text we assign to prop 'hidePasswordLabel'


packages/react/src/components/TextInput/PasswordInput.tsx
```diff
@@ -309,6 +311,24 @@ const PasswordInput = React.forwardRef(function PasswordInput(
...
+      align = `${tooltipPosition}-right`;
+    }
+    if (tooltipAlignment === 'start') {
+      align = `${tooltipPosition}-left`;
+    }
+  }
+
+  if (tooltipPosition === 'right' || tooltipPosition === 'left') {
+    align = tooltipPosition;
+  }
```
tay1orjones, line 329: Is this necessary? Tooltip now supports new alignment options like `top-end` and `top-start`.

This code pattern is something we'll eventually want to remove from the entire codebase, I added it as a line item within 
* https://github.com/carbon-design-system/carbon/issues/16009

packages/react/src/components/TextInput/PasswordInput.tsx
```diff
@@ -329,18 +349,18 @@ const PasswordInput = React.forwardRef(function PasswordInput(
...
-        )}
-        {passwordVisibilityIcon}
-      </button>
+      <Tooltip
+        align={align}
+        className={`${prefix}--toggle-password-tooltip`}
+        label={passwordIsVisible ? hidePasswordLabel : showPasswordLabel}>
+        <button
+          type="button"
+          className={passwordVisibilityToggleClasses}
```
tay1orjones, line 356: Does this variable need to change at all? Are the classes defined in it all still needed? If not, are there some associated styles we can remove in `_text-input.scss`?

packages/react/src/components/TextInput/PasswordInput.tsx
```diff
@@ -329,18 +349,18 @@ const PasswordInput = React.forwardRef(function PasswordInput(
...
+          disabled={disabled}
```
tay1orjones, line 357: I think `Tooltip` handles this, but did you confirm the tooltip doesn't show if the button is disabled? I can't configure the storybook to disable that button.

packages/styles/scss/components/text-input/_text-input.scss
```diff
@@ -131,6 +131,11 @@
     transition: outline $duration-fast-01 motion(standard, productive);
   }
 
+  .#{$prefix}--toggle-password-tooltip .#{$prefix}--popover {
+    position: absolute;
+    inset-inline-start: -(convert.to-rem(40px));
+  }
```
tay1orjones: Does this work with all alignment options? Looking at it in the dev tools I don't see `position: absolute` modifying it in any way - could possibly be removed?

tay1orjones (commented):
Functionally looks great! Just a few questions on the code

packages/react/src/components/TextInput/PasswordInput.tsx
preetibansalui: Yes, I checked the disabled functionality, it's not showing the tooltip if we pass disabled as true. I tried to remove this as well to check if tooltip is handling it or not, but then it was showing tooltip even disabled was true. So I think we have to pass it here.

packages/react/src/components/TextInput/PasswordInput.tsx
preetibansalui: These styles were already there, however I tried to check the classes and found that we are adding/removing few classes based on disabled and tooltip alignments, although few of them we can remove I think so, will remove accordingly.

packages/react/src/components/TextInput/PasswordInput.tsx
preetibansalui: Oh, the proptypes defined in the component were still using the old values. Do you want me to cover these changes as part of this PR or take it as a separate PR with #16009.

benjamin-kurien (changes requested):
Hi Preeti, the functionality looks good. 
From a UX Perspective -  to avoid the container from bleeding off the page and hiding important information, consider restricting the container of the tooltip to be only aligned to end of the input field. [Refer this variant here](https://carbondesignsystem.com/components/tooltip/usage/#alignment)

preetibansalui:
Hi Benjamin, Thanks for the feedback. As per the suggestion, I will apply the default functionality of 'tooltipPosition' as 'bottom' and for 'tooltipAlignment' as 'end'. 

But just to add here, user can still change it as we are exposing these 2 props to user and possible values are as below:

tooltipPosition: 'top' | 'right' | 'bottom' | 'left';
tooltipAlignment: 'start' | 'center' | 'end';

tw15egan (approved):

benjamin-kurien (approved):

PR #16151, opened 2024-04-08
Fixed `Toggle` aria-labelledby

### LINKED ISSUE #16055: [a11y]: Toggle - there's currently no way to set aria-label if not showing the upper label for a Toggle
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

v1.53.0

### React version

v18.2.0

IBM accessibility checker

### Assistive technology

_No response_

### Description

The way the Toggle is currently setup, `aria-labelledby` gets set to the `__label` div of the Toggle. That div doesn't get any text unless you're showing the `labelText`. In situations where you're not showing `labelText`, it would be nice to use `aria-label` to provide a label for the toggle.  Since `aria-labelledby` is automatically set though, it will override `aria-label`.

There's a property `hideLabel` that you'd think might rectify this problem, but it doesn't do what you'd think. It hides the side labels on the toggle, which isn't the desired behavior.

Suggested fix:
a) don't set `aria-labelledby` if `labelText` is empty
or 
b) add a property that hides the upper label

### WCAG 2.1 Violation

_No response_

### Reproduction/example

n/a

### Steps to reproduce

create a toggle without labelText, look at the DOM.

### Code of Conduct

guidari:
Closes #16055

`aria-labelledby` is being set empty when there is no label for reference.

#### Testing / Reviewing

- Test with VO or inspect the `Toggle` button to check the attributes.

### DIFF
--- a/packages/react/src/components/Toggle/Toggle.tsx
+++ b/packages/react/src/components/Toggle/Toggle.tsx
@@ -191,7 +191,7 @@ export function Toggle({
         role="switch"
         type="button"
         aria-checked={checked}
-        aria-labelledby={ariaLabelledby ?? (labelText ? labelId : '')}
+        aria-labelledby={ariaLabelledby ?? (labelText ? labelId : undefined)}
         disabled={disabled}
         onClick={handleClick}
       />
--- a/packages/react/src/components/Toggle/Toggle.tsx
+++ b/packages/react/src/components/Toggle/Toggle.tsx
@@ -191,7 +191,7 @@ export function Toggle({
         role="switch"
         type="button"
         aria-checked={checked}
-        aria-labelledby={ariaLabelledby ?? labelId}
+        aria-labelledby={ariaLabelledby ?? (labelText ? labelId : '')}
         disabled={disabled}
         onClick={handleClick}
       />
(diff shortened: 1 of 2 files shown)


alisonjoseph:
Looks like this aria-label is still empty?

Screenshot 2024-04-15 at 11 36 02 AM

guidari:
> Looks like the last example [here](https://deploy-preview-16151--v11-carbon-react.netlify.app/?path=/story/components-toggle--with-accessible-labels) the aria-labeledby is still empty?

Yeah. It is grabbing the label by the id. If you check the [current storybook](https://react.carbondesignsystem.com/?path=/story/components-toggle--with-accessible-labels), you will see that is not reading the label because the `aria-labeledby` was obscuring the label or aria-label. That is why now it be can empty, so the user can set the aria-label

packages/react/src/components/Toggle/Toggle.tsx
```diff
@@ -191,7 +191,7 @@ export function Toggle({
         role="switch"
         type="button"
         aria-checked={checked}
-        aria-labelledby={ariaLabelledby ?? labelId}
+        aria-labelledby={ariaLabelledby ?? (labelText ? labelId : '')}
```
tay1orjones: Would it make sense for it to be `false` or `undefined` instead of an empty string? That way the attribute won't be placed on the element at all

guidari: That makes more sense!

tay1orjones (approved):

alisonjoseph (approved):

PR #16155, opened 2024-04-10
fix(slider): set null check for 0 value in slider

### LINKED ISSUE #16141: [Bug]: Slider component does not handle "0" value correctly when the slider has a negative minimum value
### Package

@carbon/react

### Browser

_No response_

### Package version

1.54.0

### React version

_No response_

### Description

When the slider component has negative values it is impossible to slide it to 0 as a 0 value causes the slider to be on the very left hand side i.e. the lowest negative value.

In the storybook example given the slider should be in the middle, but it goes to the very left hand side whenever the value is 0, regardless of the min/max values set.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-slider--playground&args=min:-100;value:0

### Steps to reproduce

The storybook link above is enough to reproduce, I have also found the cause of the issue so a fix should be easy/quick for the carbon team to implement.

The following file (inside the calcLeftPercent function):
https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Slider/Slider.tsx#L1039

```
else if (value && range) {
  // Prevent NaN calculation if the range is 0.
  return range === 0 ? 0 : (value - this.props.min) / range;
}
```

The above code snippet shows the problem code, it uses a boolean check `value &&` to check for undefined/null, this will cause the value `0` to show as undefined, causing the slider to be on the left hand side, instead of the middle as it should be when the range is -100 to 100.

```
else if (((value ?? null) != null) && range) {
  // Prevent NaN calculation if the range is 0.
  return range === 0 ? 0 : (value - this.props.min) / range;
}
```

I have coped the code into a local component and changed the check to the one above i.e. coercing undefined/null to null with ?? then checking if it's null, rather than using a boolean check (this could also just check if its null/undefined directly, but that's a little messier). This causes the slider to work as expected when using negative values.

There are other parts of this file where `value &&` is used to check the value which probably need reviewing too as they will all fail when the value is 0, which is a valid value and should be treated as such.

image

The screenshot above shows my updated component on top with the correct behaviour, and the original broken carbon component underneath.

### Suggested Severity

### Application/PAL

IBM Internal

### Code of Conduct

aninaantony:
Closes #16141

Handling null check for 0 value in slider

#### Changelog

**New**

- {{new thing}}

**Changed**

Added null check for 0 value in slider

**Removed**

- {{removed thing}}

#### Testing / Reviewing

{{ Slider component - negative minimum value}}

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1470,6 +1470,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "aninaantony",
+      "name": "Anina Antony",
+      "avatar_url": "https://avatars.githubusercontent.com/u/164350784?v=4",
+      "profile": "https://github.com/aninaantony",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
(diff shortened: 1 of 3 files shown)


aninaantony:

aninaantony:
recheck

aninaantony:
recheck

aninaantony:

aninaantony:
Closing.

PR #16156, opened 2024-04-10
fix(tag): increase disabled selector specificity

janhassel:
Because of selector specificity, the `high-contrast` and `outline` variant of the `<Tag>` currently don't support a disabled state.

#### Changelog

**Changed**

- Increased selector specificity
- Reset `outline` property in addition to `box-shadow`

#### Testing / Reviewing

- Storybook → Tag → Playground → disabled: true; type: high-contrast
- Storybook → Tag → Playground → disabled: true; type: outline

### DIFF
--- a/packages/styles/scss/components/tag/_tag.scss
+++ b/packages/styles/scss/components/tag/_tag.scss
@@ -258,12 +258,13 @@
     outline-offset: -1px;
   }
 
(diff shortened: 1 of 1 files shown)


alisonjoseph (approved):
LGTM! thanks!

tay1orjones (approved):

PR #16169, opened 2024-04-10
fix(button): only use tooltipAlignment if button hasIconOnly

### LINKED ISSUE #16140: [Bug]: tooltipAlignment is being passed to DOM element incorrectly for Button component
### Package

@carbon/react

### Browser

Chrome

### Package version

1.54.0

### React version

18.2.0

### Description

When a Button component contains the `tooltipAlignment` property, the property is being passed incorrectly to the button DOM element causing a warning in the console.  The use case here is that I want to display an icon only on small resolutions, but the full button with icon on larger screens

>Warning: React does not recognize the `tooltipAlignment` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `tooltipalignment` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

### Reproduction/example

https://stackblitz.com/edit/github-ptctcq?file=src%2FApp.jsx

### Steps to reproduce

Load the Stackblitz and open the console to see the warning generated

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #16140 

#### Changelog

**New**

- Add a new test to cover this

**Changed**

- Remove `tooltipAlignment` prop from the props spread onto `ButtonBase`

#### Testing / Reviewing

- All tests should pass, storybook stories should be unaffected


tw15egan (approved):
LGTM 👍 ✅

alisonjoseph (approved):

PR #16170, opened 2024-04-10
fix(FluidListbox): fix issues with ai-enabled fluid listbox components

### LINKED ISSUE #16168: [Bug]: Visual bugs with fluid components with AI Slug
### Package

@carbon/react

### Browser

Chrome

### Package version

Latest

### React version

_No response_

### Description

AI border shows on `Textarea` when in an invalid state
Screenshot 2024-04-10 at 1 21 23 PM

The invalid icon is not aligned to the top right corner of the invalid text area for list box components 
Screenshot 2024-04-10 at 1 21 19 PM

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/experimental-unstable-slug-form--form-example&args=invalid:!true

### Steps to reproduce

Enable `invalid` prop

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/16168

Fixes discrepancies between the `Select` and the `Listbox` components. It also fixes an issue with the border on `Textarea` not being removed when invalid. 

#### Changelog

**Changed**

- Added some specificity so that fluid positioning is correct

**Removed**
- `::before` divider for `invalid` icon when in fluid because those invalid icons are not next to the slug

#### Testing / Reviewing

Ensure the `invalid` and `warning` icons are correctly positioned, and the spacing between the `Select` and other `Listbox` component dividers match.


tay1orjones:
Do you think it makes sense to VRT these stories to avoid regressions like this?

tay1orjones (approved):
One small thing I noticed is that the Dropdown components have a greater width for some reason? Kind of ddd because when `!invalid` they're all perfectly aligned

image

tw15egan:
@tay1orjones The main issues were only viewable when the inputs were invalid; I don't know if we are using VRT tests for invalid states anywhere, are we? Regarding the invalid dropdown, I believe that when building it, it had to use an outer outline for some reason, not an inset outline. It could be added to a future sprint to take another look and see if we can improve upon it.

guidari (approved):
LGTM! 🚀

PR #16183, opened 2024-04-11
feat(Tile): update Tile AI styles to phase 3 spec

### LINKED ISSUE #15946: [AI] Phase 3: update tile component to new visuals
[Figma files](https://www.figma.com/file/7eMfOAyRclD4LuDFwRAa3p/Carbon-for-AI-Sketches?type=design&node-id=2302%3A103910&mode=design&t=CmPCQCPF6fEpnGDA-1)
Tiles - White v3
Tiles - Gray 100 v3

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/15946

Updates AI Tile to phase 3 spec

#### Changelog

**Changed**

- Adjusted a few drop shadows, removed underline when clickable tile is focused / hovered 

#### Testing / Reviewing

Ensure `Tile` matches AI phase 3 spec

### DIFF
--- a/packages/styles/scss/components/tile/_tile.scss
+++ b/packages/styles/scss/components/tile/_tile.scss
@@ -85,6 +85,8 @@ $-icon-container-size: calc(#{layout.density('padding-inline')} * 2 + 1rem);
 
     &:focus {
       @include focus-outline('outline');
+
+      text-decoration: none;
     }
 
     &:hover,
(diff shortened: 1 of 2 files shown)


tay1orjones (approved):
LGTM

aagonzales (commented):
The gradient is bleeding to high in the tile, I can't tell where the problem is coming from though. 
image

tw15egan:
@aagonzales, we added the `15%` stop for the hover gradient mixin; perhaps we need to add it to the base one as well? Just be aware that this is the same gradient background as the callout (just with `$layer` as the background), so it would also affect that gradient.

aagonzales:
@tw15egan ok, can you add it and I'll check both to make sure it still looks ok.

tw15egan:
@aagonzales updated. If that doesn't work, we can always break the `Tile` into it's own gradient like we did with `DataTable`

aagonzales (approved):
I think that did it, it looks right to me now. Work fine for callout too.

alisonjoseph (approved):

PR #16222, opened 2024-04-17
feat: proposal for system color preferences

lee-chase:
This needs a bit of re-work as the `GlobalTheme` component had almost entirely slipped me by. Mainly because `<Theme>` did not allow me to follow system preferences I nearly always rolled my own CSS.  Thought I'd submit regardless for feedback to see if, with some additional changes it would be considered.

This PR adds a possible mechanism for use with `<Theme>` to defer to system color preferences.

The benefit of something like this would be to remove the need for Carbon consumers to implement there own theme classes and media queries to use system settings.

#### Changelog

**Changed**

- Modified GlobalTheme/Theme to include new options and a `compliment` property. The compliment property allows a system prefs based UI to still switch to an alternative. 
- Modified the CSS ouput by `zones.scss` to include media queries and new classes.
- Updated the Theme story.

### DIFF
--- a/packages/react/src/components/Theme/Theme-story.scss
+++ b/packages/react/src/components/Theme/Theme-story.scss
@@ -9,6 +9,8 @@
 @use '@carbon/styles/scss/theme';
 
 .theme-section {
+  display: flex;
+  gap: 1rem;
   padding: 1rem;
   background: theme.$background;
   color: theme.$text-primary;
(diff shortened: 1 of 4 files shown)


lee-chase:
Hey @tay1orjones submitted an alternative approach which adds no extra CSS.

It is missing `compliment` for now, but I think is a lot cleaner #16230

tay1orjones:
Closing in favor of https://github.com/carbon-design-system/carbon/pull/16230

PR #16223, opened 2024-04-17
fix(modal): spacing below scrolling content

### LINKED ISSUE #15890: [Bug]: Modal fade effect applied without setting `hasScrollingContent` prop
### Package

@carbon/react

### Browser

_No response_

### Package version

1.52.0

### React version

_No response_

### Description

The storybook docs on composed modals state that a visual fade effect can be applied to the modal content by setting the `hasScrollingContent` prop to `true` (see [storybook docs page](https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview#overflow-content)). 

In fact, the styling is applied without setting the prop, see [`ComposedModal` code](https://github.com/carbon-design-system/carbon/blob/2e8523541f0509dee7f30a1a08947a326ade14f2/packages/react/src/components/ComposedModal/ComposedModal.tsx#L61). This broke one of our modals, because content is hidden at the top and bottom of some elements.

The CSS property `mask-image` should only be applied, if the prop is specified imo. This behaviour right now rpevents our teams to upgrade to any Carbon version above `1.50.0`.

### Reproduction/example

https://stackblitz.com/edit/carbon-v11-prxvfh?file=src%2FApp.jsx

### Steps to reproduce

Look at the modal in the example link, the prop is not set, but the `mask-image` property is applied anyways

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

alisonjoseph:
Closes #15890

#### Changelog

**Changed**

- update padding to margin on last element of modal with scrolling content

#### Testing / Reviewing

Visually nothing should have changed with modal, you can add something like a button as the last element in storybook to test this update. Previously it would have applied padding inside the button vs. margin below.

### DIFF
--- a/packages/styles/scss/components/modal/_modal.scss
+++ b/packages/styles/scss/components/modal/_modal.scss
@@ -363,7 +363,7 @@
 
   // Required so overflow-indicator disappears at end of content
   .#{$prefix}--modal-scroll-content > *:last-child {
-    padding-block-end: $spacing-06;
+    margin-block-end: $spacing-06;
   }
 
   // -----------------------------


tay1orjones (approved):

tw15egan (approved):
Thanks for fixing this 🙏🏻

PR #16227, opened 2024-04-18
feat(RadioTile): update `RadioTile` to accept `Slug`, phase 3 updates

### LINKED ISSUE #16224: AI enhancements: Add AI variant for radio selection tile
We have a multi-select AI tile but we are missing the radio (single) select tile. We want to add the AI mix-in to both the stable and feature flag versions of tile. 

**AI versions:**
- [ ] Radio select stable
- [ ] Radio select feature flag

Tiles - radio - White v3

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/16224

Adds in AI styles for `RadioTile`

#### Changelog

**New**

- Adds in support for `slug` prop to `RadioTile`

**Changed**

- Updated the `unstable__slug --> Examples --> Tile` story to include 3 `RadioTile`, 3 `RadioTile` with feature flags enabled, and updated the `SelectableTile` to show 3.
- Fixed some `z-index` issues with `RadioTile` and `SelectableTIle` when the AI slug is clicked

#### Testing / Reviewing

Ensure the `unstable__slug --> Examples --> Tile` story looks correct and the styles are applied properly


aagonzales (changes requested):
Looking good but in the feature flag verison the unselected tiles are supposed to always have the unchecked radio button. Like so:

image

aagonzales (changes requested):
OK that looks good but I just noticed another thing. The focus on the radio tiles (both kinds) should be a 2px border. Its showing at 1px right now
image

packages/react/src/components/Slug/slug-story.scss
```diff
@@ -164,3 +165,24 @@
...
+
+.slug-selectable-tile-container .cds--tile-group > div > div {
+  width: 33%;
+}
+
+.slug-selectable-tile-container .cds--tile-group .slug-radio-tile,
+.slug-selectable-tile-container .cds--tile-group .slug-selectable-tile {
+  margin-right: 1rem;
+  height: 200px;
+}
```
guidari: Shouldn't we use logical properties in this part or it doesn't matter in this case?

guidari (approved):
LGTM! 🚀
Just one question

packages/react/src/components/Slug/slug-story.scss
```diff
@@ -164,3 +165,24 @@
...
+  display: flex;
+}
+
+.slug-selectable-tile-container .cds--tile-group > div > div {
+  width: 33%;
+}
+
+.slug-selectable-tile-container .cds--tile-group .slug-radio-tile,
+.slug-selectable-tile-container .cds--tile-group .slug-selectable-tile {
+  margin-right: 1rem;
```
alisonjoseph:
```suggestion
  margin-inline-end: 1rem;
```

packages/react/src/components/Slug/slug-story.scss
alisonjoseph: I updated the margin, but I don't think it matters too much for this example.

alisonjoseph (approved):

aagonzales (approved):
Everything looks good now! RIP TJ

tay1orjones (approved):

PR #16230, opened 2024-04-19
feat(theme): improve support for respecting system color preferences

lee-chase:
Closes #16222 

This PR adds a possible mechanism for use with `<Theme>` to defer to system color preferences.

This provides for a much simpler mechanism for the user to switch themes based on system settings. It emoves the need for Carbon consumers to implement there own theme classes and media queries to use system settings.

In addition the theme context reports `themeCompliment` and `isDark`. The first enables the new `<ThemeCompliment>` component and the second is extremely useful for those that need to switch colors or images based on a light or dark theme.

#### Changelog

**Changed**

- Modified GlobalTheme/Theme adding a `system` option to the theme property as well as `themeSystemDark` and `themeSystemLight` properties.
- Updated other stories to report the additional information.
- Updated the `useTheme` context to output theme, themeCompliment and isDark.
- Added a ThemeCompliment component and story.

NOTE: This is as far as I can tell non-breaking.
NOTE 2: If accepted the docs will need updating.


lee-chase:
@tay1orjones I think #16222 should be closed in favour of this PR and people encouraged to use the `<Theme>` component instead of rolling there own CSS.

tay1orjones:
Hey @lee-chase! I met with a couple folks on the team about this yesterday, a few things to note:

* Theme compliments/inverse should be fully configurable, we don't want to force anything from a design perspective. If someone wants to use white/g10 for light/dark, they can.
* The [theme docs](https://carbondesignsystem.com/elements/color/usage/#light-or-dark-mode) on the website outline a pretty robust framework for how this should work
  * It includes one [notable piece](https://carbondesignsystem.com/elements/color/usage#light-or-dark-mode:~:text=Inline%20themes%20with,when%20switching%20modes.) - you're allowed to mix inline theme contrast between the different modes (dark on dark for a low-contrast moment, dark on light for a high-contrast moment). So we don't want to force or limit anyone to using a specific inverse/compliment for light or dark.
* The default theme out of the box is `white`, while most projects use `g10`. If we provide a "pair-default", design would prefer the pair-default to be `g10`/`g100`. I'm not sure if that makes sense though, we wouldn't want to change the default theme. For now we might want to just have `white`/`g100` as the pair-default, if we bake a pair-default in at all.

---

What would you think about trying to drop the changes to Theme - `themeCompliment`, `themeSystemDark`, `themeSystemLight` to reduce the internal complexity? Since `themeSystemDark`, and `themeSystemLight` can't be limited to two options, I'm not sure that additional complexity justifies itself when `isDark` gets folks the majority of the way there:

```js
const isDark = useTheme();
...
<GlobalTheme theme={isDark ? "white":"g100"}/>
```

```js
// Similarly, for configuring a low contrast "dark on dark" local override 
const isDark = useTheme();
...
<Theme theme={isDark ? "g90":"g100"}/>
```

Am I missing something there? You're much more hands-on familiar with the use case here than I am - it's very possible this doesn't solve the main problem you're running into. 

Otherwise just a few follow-on concerns

* As you've shown, with this approach there's no need to modify the scss and we could instruct other framework authors to recommend or follow this approach.
* We'd need to document this in storybook and ideally on the [website themes code page](https://carbondesignsystem.com/elements/themes/code/).
* Also would be interesting to make a new example or perhaps modify the [light-dark-mode example](https://github.com/carbon-design-system/carbon/tree/main/examples/light-dark-mode) to showcase this with or against the approach there that [uses context](https://github.com/carbon-design-system/carbon/blob/b6f0cb3de55335f3a228af1eb860455f984b877c/examples/light-dark-mode/src/components/ThemePreference.js#L4-L7).

Also thanks again for putting this together, sorry it's taken such time to get a response back on this.

packages/react/src/components/Theme/index.tsx
```diff
@@ -129,3 +133,7 @@ Theme.propTypes = {
 export function useTheme() {
   return React.useContext(ThemeContext);
 }
+
+export function usePrefersDarkScheme() {
```
tay1orjones, line 137: This also needs to be added to the top level exports https://github.com/carbon-design-system/carbon/blob/7e364d8c1d35cda0f92d95dc1abf3eef2546309d/packages/react/src/index.js#L303

tay1orjones (approved):
You pulled together my disparate thoughts so well 😅 

I think this looks great! One minor thing I'll add and then LGTM

alisonjoseph (approved):
Nice, I like the approach you landed on!

packages/react/src/index.js
```diff
@@ -300,7 +300,7 @@ export {
   TextDirection as unstable_TextDirection,
 } from './components/Text';
 export { DefinitionTooltip } from './components/Tooltip/DefinitionTooltip';
-export { GlobalTheme, Theme, useTheme } from './components/Theme';
```
lee-chase, line 303: Was the export of useTheme @tay1orjones accidentally removed

darq37: I think this is a bug

PR #14641, opened 2023-09-12
style(Checkbox): align styles with vertical radio buttons, figma kit

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/14628

Aligns vertical `RadioButton` and `Checkbox` styles so that they fit the same dimensions. Also brings styles more in line with the Figma kit. The experiment can be found [here](https://stackblitz.com/edit/github-81xkqc-bkl4li?file=src%2FApp.jsx,src%2Findex.scss).

#### Changelog

**Changed**

- Adjusted styles so that `Checkbox` and `RadioButton` have similar bounding boxes / padding

#### Testing / Reviewing

Go to `Checkbox --> Test` and ensure there is no alignment changes when the component switches from `RadioButton` to `Checkbox` and vice versa

PR #16236, opened 2024-04-22
feat(multiselect): new All option for multiselect

### LINKED ISSUE #16117: MultiSelect (select all option): dev implementation
## Acceptance criteria
- [x] Add the Select all option enhancement to the Multiselect component. 

### Design spec

Image

Gururajj77:
Closes #16117

Select All feature for multiselect

#### Changelog

**New**

- All option created

**Changed**

- changes in the logic for multiselect and selection.js


packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -226,6 +227,8 @@ export interface MultiSelectProps<ItemType>
    */
   items: ItemType[];
 
+  itemsWithSelectAll: ItemType[];
```
guidari: Add to MultiSelect.propTypes and also would good to insert a description, so when the user reads the Overview page it has more context.

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -268,6 +271,11 @@ export interface MultiSelectProps<ItemType>
    */
   readOnly?: boolean;
 
+  /**
+   * to Show/Hide All option
+   */
+  selectAll?: boolean;
```
guidari: Add to MultiSelect.propTypes

packages/react/src/components/ListBox/ListBoxMenuItem.tsx
```diff
@@ -39,6 +39,11 @@ export interface ListBoxMenuItemProps extends ReactAttr<HTMLLIElement> {
    * Specify whether the item should be disabled
    */
   disabled?: boolean;
+
+  /**
+   * Specify whether the item is Select All option
+   */
+  isSelectAll?: boolean;
```
guidari: Add to ListBoxMenuItem.propTypes

packages/react/src/components/ListBox/ListBoxMenuItem.tsx
```diff
@@ -63,6 +68,7 @@ const ListBoxMenuItem = React.forwardRef<HTMLLIElement, ListBoxMenuItemProps>(
       children,
       isActive = false,
       isHighlighted = false,
+      isSelectAll = false,
       title,
```
guidari: I noticed we have `title` and `children` on the ListBoxMenuItem.propTypes but no in the interface ListBoxMenuItemProps. We could also add it there

guidari (changes requested):
Hey @Gururajj77 and @preetibansalui 
Awesome new functionality! it is working perfect 🚀
Just a couple things I noticed

guidari (approved):

packages/react/src/components/ListBox/ListBoxMenuItem.tsx
tay1orjones: Could be something to defer for #6177

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -226,6 +227,11 @@ export interface MultiSelectProps<ItemType>
    */
   items: ItemType[];
 
+  /**
+   * This is similar to the items prop, but it is used to render the items when selectAll prop is enabled
+   */
+  itemsWithSelectAll?: ItemType[];
+
```
tay1orjones: Is this necessary? It's not used in the stories at all. If it is necessary, lets make a story to showcase how this prop is used.

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -268,6 +274,11 @@ export interface MultiSelectProps<ItemType>
    */
   readOnly?: boolean;
 
+  /**
+   * to Show/Hide All option
+   */
+  selectAll?: boolean;
```
tay1orjones: Should this be prefixed with `has`?

```suggestion
  hasSelectAll?: boolean;
```

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -393,6 +400,23 @@ const MultiSelect = React.forwardRef(
       });
     }, [items]);
 
+    const itemsWithSelectAll = selectAll
+      ? [selectAllOption, ...filteredItems]
+      : filteredItems;
```
tay1orjones: How does this work when the component is controlled via `selectedItems`? Does the user get the selectAllOption in the `items` returned through `onChange`? _Should_ they get that back? How would someone using the controlled component check the select all, or mark it indeterminate? Do they need to or does it happen automatically? 

It might be worth adding a controlled multiselect select-all example to storybook to discover the answers to some of these things, if we need to make some workarounds, and/or what to document in storybook. 

If there are unavoidable prop conflicts, like `selectAll` just plain doesn't work with the component in a controlled state, then ideally we'd flag this with a warning in the dev console. 

> e.g.: "MultiSelect's `selectAll` prop does not work when in a controlled state using `selectItems`"

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -649,6 +673,9 @@ const MultiSelect = React.forwardRef(
       selectedItems.length > 0 &&
       selectedItems.map((item) => (item as selectedItemType)?.text);
 
+    const selectedItemsWithoutSelectAll = selectedItems.filter(
+      (item: any) => item.id !== 'select-all-option'
+    );
```
tay1orjones: Is there are more reliable way to determine this? Thinking about how some consumers might already be rolling their own version of this feature and using this same id. It might be unavoidable, but thought I'd ask what you think.

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -580,5 +557,34 @@ describe('MultiSelect', () => {
         `${prefix}--list-box__wrapper--slug`
       );
     });
+
+    it('should select all options when selectAll prop is true', async () => {
```
tay1orjones: Along the same lines as the questions about controlled state - adding a test for how this is used with fully controlled `selectedItems` would be awesome. Select all, select none, indeterminate, etc.

tay1orjones (commented):
Generally looks pretty good! I have some questions about fully-controlled usage and if or why `itemsWithSelectAll` is required. I'd think that `selectAll` would be enough to toggle the feature in the component without any additional props.

packages/react/src/components/MultiSelect/MultiSelect.tsx
Gururajj77: We do not need this as a prop, removed it.

packages/react/src/components/MultiSelect/MultiSelect.tsx
Gururajj77: Have renamed it to `hasSelectAll`

packages/react/src/components/MultiSelect/MultiSelect.tsx
Gururajj77: no we are using a new property in the `selectAll` object, called as `selectAllFlag` which just a boolean so that we don't mess with id being same as user's data.

packages/react/src/components/MultiSelect/MultiSelect.tsx
preetibansalui: Hi Taylor, Please find the response:

_Does the user get the selectAllOption in the items returned through onChange?_
Yes user will get selectAllOption in the items returned through onChange.

_Should they get that back?_
It's the part of Items now.

_How would someone using the controlled component check the select all, or mark it indeterminate? Do they need to or does it happen automatically?_
If they are using controlled component, they won't have access to selectAll or indeterminate state. To enable these things user has to enable 'hasSelectAll' as true.

_If there are unavoidable prop conflicts, like selectAll just plain doesn't work with the component in a controlled state, then ideally we'd flag this with a warning in the dev console._
If we are adding the prop 'hasSelectAll' prop in controlled state, it will work simply as the default one. Only issue is we can have conflict of logics, for eg - the current controlled component is selecting disabled items as well while we are not. 

While writing these responses, I found that we have to disable select All functionality in case of controlled state. User can either use selectedItems or hasSelectAll prop. We will add a console warning as suggested.

tay1orjones (commented):
Hey just wanted to comment what we chatted about on the side - I'm still a bit concerned around the lack of compatibility with the component being fully controlled through `selectedItems`. The prop warning is a decent stopgap measure but I think we should also mark it as experimental. This way, if/when we get feedback that hasSelectAll needs to work with `selectedItems` we can make the necessary changes without having to do a breaking change. 

Once marked experimental I think this is good to merge

laurenmrice:
This also needs a final design review right? I would add Benjamin to this as a reviewer if it needs a design review.

Gururajj77:
> This also needs a final design review right? I would add @benjamin-kurien to this as a reviewer if it needs a design review.

@benjamin-kurien, please review the new feature

benjamin-kurien (changes requested):
1. To check accessibility focus states
2. There has to be 2 different text parameters: prompt text( def value= Choose options), selected text (def value = Options selected)
3. Clicking on indeterminate state should unselect all options

preetibansalui:
> 1. To check accessibility focus states
> 2. There has to be 2 different text parameters: prompt text( def value= Choose options), selected text (def value = Options selected)
> 3. Clicking on indeterminate state should unselect all options

Hey @benjamin-kurien , I have made following changes, please review it.
1. Create a separate story as 'Select All'
2. Click on indeterminate state should deselect all selected items
3. All option should not display active state when clicked on it.
4. Added a new prop as 'labelforSelectAll'.
4. Changed options to match documentation as suggested by @tay1orjones.

tay1orjones (changes requested):
Overall these changes look great to me and the storybook example in particular is so much better with the item label changed to "All roles" 🙌 

I did notice that the checkbox and text of the first/select-all is not vertically centered within the item, it sits slightly too low. Might be off by just 1px or so.

CleanShot 2024-06-13 at 14 35 07

preetibansalui:
> Overall these changes look great to me and the storybook example in particular is so much better with the item label changed to "All roles" 🙌
> 
> I did notice that the checkbox and text of the first/select-all is not vertically centered within the item, it sits slightly too low. Might be off by just 1px or so.
> 
> CleanShot 2024-06-13 at 14 35 07

Hey @tay1orjones , Thanks for reviewing this, we are using same classes for the first option as others. It might be possible you are looking the first item as highlighted one so its making you feels like its not centred align. Pls check the screenshots for the sizing: 

image
image
For the first element we are just adding a bottom border which too takes place of the border in case of other options.

tay1orjones:
@preetibansalui Yeah you're right, it's broken in prod too - this is react.carbondesignsystem.com

CleanShot 2024-06-14 at 08 26 30

I'm not sure how it got past Percy VRT. Would you mind fixing it in this PR? I know it's a bit beyond the original scope of this issue, but would be great to get it fixed.

preetibansalui:
> @preetibansalui Yeah you're right, it's broken in prod too - this is react.carbondesignsystem.com
> 
> CleanShot 2024-06-14 at 08 26 30
> 
> I'm not sure how it got past Percy VRT. Would you mind fixing it in this PR? I know it's a bit beyond the original scope of this issue, but would be great to get it fixed.

Done. It was a small fix only.
image

laurenmrice (changes requested):
Hey all! Looking good! Just wanted to chime in here with a couple design related comments:

**Select All story**
- The “All roles” option should be at the top of the menu in the multiselect dropdown and not at the bottom, unless the design spec for this has changed.
- The first option that currently gets focus in the menu called “Editor” should be text color token `$text-secondary` in the focus and enabled states. The text color only turns to $text-primary on the hover and selected states.

**Playground story**
- When I set the `hasSelectAll` prop to `true`, the text for the select all option does not appear in the menu of the multiselect dropdown.
- The “All roles” option should be at the top of the menu and not at the bottom in the multiselect dropdown, unless the design spec for this has changed.
- The first option that currently gets focus in the menu called “An example option that is real…” should be text color token `$text-secondary` in the focus and enabled states. The text color only turns to $text-primary on the hover and selected states.

preetibansalui:
> Hey all! Looking good! Just wanted to chime in here with a couple design related comments:
> 
> **Select All story**
> 
> * The “All roles” option should be at the top of the menu in the multiselect dropdown and not at the bottom, unless the design spec for this has changed.
> * The first option that currently gets focus in the menu called “Editor” should be text color token `$text-secondary` in the focus and enabled states. The text color only turns to $text-primary on the hover and selected states.
> 
> **Playground story**
> 
> * When I set the `hasSelectAll` prop to `true`, the text for the select all option does not appear in the menu of the multiselect dropdown.
> * The “All roles” option should be at the top of the menu and not at the bottom in the multiselect dropdown, unless the design spec for this has changed.
> * The first option that currently gets focus in the menu called “An example option that is real…” should be text color token `$text-secondary` in the focus and enabled states. The text color only turns to $text-primary on the hover and selected states.

Hey @laurenmrice , Thankyou for reviewing this PR. But I am confused with the comments. I can see the All roles in the top only. Attaching the video of preview, please let me know if I misunderstood something.

https://github.com/carbon-design-system/carbon/assets/146315451/970c83fb-955c-40d5-93c0-2eff8d98f8bd

laurenmrice:
@preetibansalui Thanks. I was using the Deploy Preview (https://deploy-preview-16236--v11-carbon-react.netlify.app/) link in your PR to review it. This is how the designers usually review PRs, but I guess this link has not been updated with the latest changes. 
Screenshot 2024-06-20 at 9 13 54 AM

Would you be able to update this, or get help from a dev to update it? I don't know what the procedure is of updating that link. Thanks!

preetibansalui:
Had a call with @laurenmrice & @benjamin-kurien and found that in Chrome and Safari its fine but in firefox browser only 'All roles' is coming as a last option. . 
For another issue, option text should be of `$text-primary` for highlighted text as well but we were able to replicate it in Multiselect, Dropdown & combobox. So Lauren will be creating a separate issue for this.

packages/styles/scss/components/checkbox/_checkbox.scss
```diff
@@ -62,7 +62,7 @@
     display: flex;
     cursor: pointer;
     min-block-size: convert.to-rem(20px);
-    padding-block-start: convert.to-rem(2px);
+    padding-block-start: convert.to-rem(1px);
```
tay1orjones, line 65: It doesn't look like this fixed the vertical alignment issue. Playing around with it quickly seems like maybe the `label:before` needs to be modified? In the history I'm not sure if this maybe introduced this? https://github.com/carbon-design-system/carbon/pull/14641

benjamin-kurien (approved):
Looks good. Thanks for making the changes.

guidari (approved):
LGTM! 🚀

laurenmrice (changes requested):
Just a bug I am seeing here, if I select all of the options one by one inside of the dropdown menu and then close the dropdown menu  and re-open it, the "All" option moves lower in the list. Looks like this is happening in Firefox/Chrome/Safari.

test

Gururajj77:
Hi @laurenmrice , I have fixed the changes, where the the `select all` option will always be on top and won't be part of the sorting algorithm so that user can see the select all option at first whatever the label is.

laurenmrice (approved):
Looks good! Thank you ⭐️ @Gururajj77 @preetibansalui

tay1orjones (changes requested):
This is getting closer for sure! I hope it's okay with you, I made a couple changes:

* I removed `isSelectAll` from ListBoxMenuItem and refactored the class to be placed inside MultiSelect. Also updated the styles to reflect this change. https://github.com/carbon-design-system/carbon/pull/16236/commits/13ba4156cedc6d45c7914596a471433c8455668a
* Updated the prop names to be `selectAll` and `selectAllItemText`. I know we went back and forth on this 😅 I went in this direction because it ensures the props are next to one another in the alphabetical props list in the source but also in the storybook prop tables. https://github.com/carbon-design-system/carbon/pull/16236/commits/2f236f895c39248f34df901218e6edf3a4e40023
* Renamed `selectAllFlag` to `isSelectAll` so usages are a bit more readable - e.g. `item.isSelectAll` https://github.com/carbon-design-system/carbon/pull/16236/commits/624fc1e982355c9a58b15d034d657148289aab9c

Let me know what you think about these.

I found a couple things that are broken:
1. The alignment is still off. Do we have a separate issue for this? I think it'd be nice to fix here if possible.
2024-07-29 at 13 34 21-CleanShot-CleanShot
2. There's some odd focus management stuff going on. Two issues I noticed. One with where focus is placed after selection via click, and another with keyboard navigation inside the list. Adding tests to cover these usage patterns would be awesome.


https://github.com/user-attachments/assets/e26f2501-6422-44fe-a9fd-6cbc947c9e7c


https://github.com/user-attachments/assets/52544509-0c5c-41d9-a067-b75bac096f35


3. AVT tests are failing

packages/react/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -652,6 +685,10 @@ const MultiSelect = React.forwardRef(
       selectedItems.length > 0 &&
       selectedItems.map((item) => (item as selectedItemType)?.text);
 
+    const selectedItemsWithoutSelectAll = selectedItems.filter(
+      (item: any) => !item.isSelectAll
+    );
```
tay1orjones: Should this be guarded behind `selectAll`? If so, the usages will need to be updated.

packages/styles/scss/components/multiselect/_multiselect.scss
```diff
@@ -42,6 +42,26 @@
     min-inline-size: auto;
   }
 
+  // When select all is turned on, the first item in the listbox
+  // is always the "Select all" option and gets special border
+  // treatment
+  .#{$prefix}--multi-select.#{$prefix}--multi-select--selectall
+    .#{$prefix}--list-box__menu-item:first-child
+    .#{$prefix}--list-box__menu-item__option {
+    padding: 0.6875rem 1rem;
```
kennylam:
```suggestion
    padding: 0.6875rem $spacing-05;
```

Should use token here. Also, what is the `0.6875` magic number? Should it be documented in the comments?

packages/react/src/components/MultiSelect/MultiSelect.tsx
Gururajj77: I believe we do not need a guard for this as it's a simple filtering which will remove items with isSelectAll key-value, which will pass anyway if user would not pass an object with isSelectAll

laurenmrice:
@Gururajj77 Is this ready to re-reviewed?

Gururajj77:
> @Gururajj77 Is this ready to re-reviewed?

Yes, it's ready @laurenmrice

benjamin-kurien (changes requested):
Selected items in the dropdown, needs to move up to the top of the list when dropdown is reopened. This functionality seems to missing. Please check.

benjamin-kurien (commented):
This looks good from a functional perspective. From a formatting perspective, the following alignments could get minor fixes: 

1. Vertical alignment for the checkbox with the text
2. Vertical alignment for indeterminate icon within the checkbox 

Thank you for your work on this.

preetibansalui:
> This looks good from a functional perspective. From a formatting perspective, the following alignments could get minor fixes:
> 
> 1. Vertical alignment for the checkbox with the text
> 2. Vertical alignment for indeterminate icon within the checkbox
> 
> Thank you for your work on this.

Hi Benjamin, I have fixed the alignment issue, Can you please review the changes again. Thanks!

benjamin-kurien (approved):
This looks good. All alignment issues seems fixed. No functional issues either. Thank you.

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -166,14 +162,6 @@ describe('MultiSelect', () => {
...
       document.querySelector('[aria-selected="true"][role="option"]')
     ).toBeNull();
-
-    await userEvent.click(itemNode);
-
-    expect(itemNode).toHaveAttribute('data-contained-checkbox-state', 'true');
-
-    await userEvent.click(itemNode);
-
-    expect(itemNode).toHaveAttribute('data-contained-checkbox-state', 'false');
```
tay1orjones: This block

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -239,18 +227,6 @@ describe('MultiSelect', () => {
...
-    // eslint-disable-next-line testing-library/prefer-screen-queries
-    const itemNode = getByText(container, item.label);
-
-    expect(itemNode).toHaveAttribute('data-contained-checkbox-state', 'false');
-
-    await userEvent.keyboard('[Enter]');
-    await userEvent.keyboard('[ArrowDown]');
-    await userEvent.keyboard('[Enter]');
-
-    expect(itemNode).toHaveAttribute('data-contained-checkbox-state', 'true');
```
tay1orjones: This block

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -345,11 +321,6 @@ describe('MultiSelect', () => {
       const labelNode = getByText(container, label);
 
       await userEvent.click(labelNode);
-
-      expect(
-        // eslint-disable-next-line testing-library/no-node-access
-        document.querySelector('[data-contained-checkbox-state="true"]')
-      ).toBeInstanceOf(HTMLElement);
```
tay1orjones: and this block

tay1orjones (changes requested):
This looks really great! The removals in the tests create some gaps in the test coverage. We should be able to find equivalents for those to keep the same coverage while no longer relying on `data-contained-checkbox-state`

tay1orjones (approved):
I pushed a fix for the tests, LGTM!

akhandsi:
@Gururajj77 hi, can you please explain the intent of

```
let selectAll = filteredItems.some((item) => (item as any).isSelectAll);
    if ((selected ?? []).length > 0 && selectAll) {
      console.warn(
        'Warning: `selectAll` should not be used when `selectedItems` is provided. Please pass either `selectAll` or `selectedItems`, not both.'
      );
      selectAll = false;
    }
    ```
    
Seems this warning may not be needed?

I have a use-case where i do need to pass in selectedItems since these are driven by external component to remember user's selection. I.e lets say i have 3 objects 

```
 [
  {id: 'selectAll', text: 'Select All', isSelectAll: true},
   {id: 'item1', text: 'item 1'},
   {id: 'item2', text: 'item 2'},
   {id: 'item3', text: 'item 3'}
 ]
```

then i would want to be able to preload the multi-select component with the three items being selected on initial load.

Is that use-case not expected?

PR #16250, opened 2024-04-23
style(Slider): fix issues with hidden input variant misalignment

tw15egan:
Fixes an issue that was causing alignment issues when the `hiddenInput` prop was selected

Before:

After: 

#### Changelog

**Changed**

- Set position to `absolute` that was being overridden from recent changes to `popover`
- Changed `top-left` to `top` for two handle variants

**Removed**

- Some unnecessary position styles

#### Testing / Reviewing

Ensure both single handle and two handle variant of the `hiddenInput` `Slider` component looks correct


guidari (approved):

tay1orjones (approved):
LGTM!

PR #16259, opened 2024-04-24
fix(Multiselect): remove focus when MultiSelect is open

### LINKED ISSUE #16252: [Dropdown: Multi-select] Fixing the focus state
### Description

Removing the focus state from the parent item in the [opened multi-select](https://react.carbondesignsystem.com/?path=/story/components-multiselect--default) for both cases:

- [x] [Default](https://react.carbondesignsystem.com/?path=/story/experimental-fluid-components-unstable-fluidmultiselect--playground)
>
> image

---

- [x] [Fluid](https://react.carbondesignsystem.com/?path=/story/experimental-fluid-components-unstable-fluidmultiselect--playground)
> 
> image

---

### LINKED ISSUE #16257: [Filterable Multi-Select] Inconsistency in Accessing Default and Fluid Options
### The problem

Inconsistency in accessing the [default](https://arc.net/l/quote/vqrwreth) and [fluid](https://react.carbondesignsystem.com/?path=/story/experimental-fluid-components-unstable-fluidmultiselect--playground) filterable multi-select options

image

### The solution

Adding consistent example stories/ sub section to the Fluid version in storybook similar to the Default one.

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/16252
Closes https://github.com/carbon-design-system/carbon/issues/16257

Removes focus from `MultiSelect` when the dropdown menu is opened. The focus should return to the input when it is closed.

This also adds a distinct `FilterableMultiselect` story under `FluidMultiSelect` for consistency with the non-fluid `Multiselect` story

#### Changelog

**New**

- `Filterable` story under `Fluid components --> FluidMultiselect`

**Changed**

- Only adds focus to the `Multiselect` button when the input is closed

#### Testing / Reviewing

Open the `Multiselect`; there should be no focus border around the input, only the item. Ensure the `Filterable` variant of `FluidMultiselect` is discoverable


Kritvi-bhatia17 (changes requested):
DAMN, @tw15egan, that was quick! 🔥

1. Sorry, I might not have been clear about the [focus state explanation](https://github.com/carbon-design-system/carbon/issues/16252) (both for default and fluid multi-select).
>
> image

2. What was the initial thought behind ["With Initial Selected Item"](https://deploy-preview-16259--v11-carbon-react.netlify.app/?path=/story/components-multiselect--with-initial-selected-items) showing this upfront?

3. Should we also introduce `Filterable With Layer` under [fluid multi-select](https://arc.net/l/quote/steuehgt) as well?
> image

tw15egan:
@Kritvi-bhatia17 I don't know if I agree with #1, as we keep the focus for `FilterableMultiselect` and `Combobox` when an item is selected because a user can still type in the input. What affordance does the added input focus give a user when an item is selected vs not? 

#2 is just an example of using the `initialSelectedItems ` prop / ensures it is working properly

#3 Sure I can add that 👍🏻

Kritvi-bhatia17:
> @Kritvi-bhatia17 We keep the focus on `FilterableMultiselect` and `Combobox` when an item is selected because a user can still type in the input. What affordance does the added input focus give a user when an item is selected vs not? It gains "focus" but doesn't gain any new functionality
> 
> `#2` is just an example of using the `initialSelectedItems ` prop / ensures it is working properly
> 
> `#3` Sure I can add that 👍🏻

`#1` In the selected multi-select, as a user can clicks on the `2 X tag` (actionable), that's why a focus state is applied to the parent element.

tw15egan:
@Kritvi-bhatia17 Focus states are usually added as an affordance for keyboard users, not mouse users, but I can add this if you would like. We don't have any of these "double" focus states outside the `ListBox` components, so we are setting precedence here that may need to be expanded to other components in the repo. When the menu is closed, a keyboard user will clear the selection by pressing the escape key when the focus is on the input. Focus used to go to the `x` icon directly, but that was removed in https://github.com/carbon-design-system/carbon/pull/13173

tw15egan:
@Kritvi-bhatia17 updated 👍

Kritvi-bhatia17:
> @Kritvi-bhatia17 Focus states are usually added as an affordance for keyboard users, not mouse users, but I can add this if you would like. We don't have any of these "double" focus states outside the `ListBox` components, so we are setting precedence here that may need to be expanded to other components in the repo. When the menu is closed, a keyboard user will clear the selection by pressing the escape key when the focus is on the input. Focus used to go to the `x` icon directly, but that was removed in #13173

Hi @tw15egan! Thanks for sharing this with me. I'm not entirely sure about this, so I'll reach out to Lauren to get a little more context.

Kritvi-bhatia17 (approved):
LGTM! 
Great work @tw15egan!! 🔥

tay1orjones (approved):

laurenmrice (approved):

Kritvi-bhatia17:
> @Kritvi-bhatia17 Focus states are usually added as an affordance for keyboard users, not mouse users, but I can add this if you would like. We don't have any of these "double" focus states outside the `ListBox` components, so we are setting precedence here that may need to be expanded to other components in the repo. When the menu is closed, a keyboard user will clear the selection by pressing the escape key when the focus is on the input. Focus used to go to the `x` icon directly, but that was removed in #13173

Hi @tw15egan, I connected with Lauren to confirm about this and she said that:
While the point you're making is valid, but we we are adding this because of a11y reasons. We show focus on the first item in the menu for [Menu](https://react.carbondesignsystem.com/?path=/story/components-menu--playground), and same with [Combo button](https://react.carbondesignsystem.com/?path=/story/components-menu--playground) and [Overflow menu](https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default), and those were a11y decisions. So it seems like we are kind of doing this across multiple components in our system already.

alisonjoseph (approved):

PR #16283, opened 2024-04-30
chore(combobox): refactor to use useCombobox

### LINKED ISSUE #14684: Refactor ComboBox to use `useComboBox` instead of `<Downshift>` component

### LINKED ISSUE #15927: [Bug]: Combo box incorrectly resets to previously selected value (or no-value) when a click occurs INSIDE the text input

preetibansalui:
Closes #14684 
Closes https://github.com/carbon-design-system/carbon/issues/15927

Refactored the code to use usecombobox

#### Changelog

**Changed**

Code refactored

#### Testing / Reviewing

It should work in similar ways as existing combobox.


packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -855,8 +837,9 @@ ComboBox.propTypes = {
   /**
    * Additional props passed to Downshift
    */
-  // @ts-ignore
-  downshiftProps: PropTypes.shape(Downshift.propTypes),
+  downshiftProps: PropTypes.object as React.Validator<
+    UseComboboxProps<unknown>
+  >,
```
tay1orjones, line 845: I'm concerned this could be a breaking change

preetibansalui: Any suggestions please? 
Functionality wise it's working fine although with 'onStateChange' like what we were using this prop earlier.

tay1orjones: Yeah you're right this looks fine, my mistake

tay1orjones:
@tw15egan @guidari This one's ready for review

It looks like this solves https://github.com/carbon-design-system/carbon/issues/15927 too. 

I've closed the original PR, https://github.com/carbon-design-system/carbon/pull/15904

guidari (changes requested):
Hey Preeti! Sorry for the late review on this.

A few things I noticed
1 - The ComboBox only opens when the icon is clicked. it should open if is clicked in the anywhere inside the input.
2 - When opening the Combobox with keyboard and searching for "3" (which is the disabled option) and after that pressing Enter, the Combobox is closed. It should stay open. 
Also looks like is not announcing the disabled option in the VO when searching.

I'm happy to pair up to check any of those cases, if you want.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -421,65 +405,71 @@ const ComboBox = forwardRef(
...
-          }
-          break;
-        case clickButton:
-        case clickItem:
-        case unknown:
-          setHighlightedIndex(getHighlightedIndex(changes));
-          break;
-      }
-    };
+            break;
```
guidari: We can remove that.
```suggestion
```

tay1orjones:
> A few things I noticed
> 1 - The ComboBox only opens when the icon is clicked. it should open if is clicked in the anywhere inside the input.
> 2 - When opening the Combobox with keyboard and searching for "3" (which is the disabled option) and after that pressing Enter, the Combobox is closed. It should stay open.

It would be awesome if as these are fixed, new test cases could be written to cover this functionality and ensure it continues to work as expected

guidari (approved):
LGTM! 🚀
Let's open an issue to track the VO not reading the disabled item

tay1orjones (approved):
LGTM once the conflicts are fixed 🎉

PR #16284, opened 2024-04-30
feat: export props for accordion

### LINKED ISSUE #16154: [Feature Request]: Export props for TypeScript components
### The problem

In the `@carbon/react` project multiple components which have been converted into TypeScript in Carbon 11 are lacking either exports on their props,. Without access to the props, it is much more difficult to create wrappers for components, pass along props as part of parent components, and assigning types to objects that can help build out components neatly. This has affected my team's efforts to move over from Carbon 10 to 11.

For example, in my team's product we leverage the base component's props (aliased here as `CarbonAccordionProps`) to create the typings for a wrapper for the `Accordion` component with a simple overwrite. Without these props, we would have to redefine all of the parameters in our codebase by hand, which would open us up to issues down the road if the Carbon component's props change.

image

Included here is a list of all props that are missing exports that my team have found. Note that there may be additional components with this issue that we haven't identified yet.
```[tasklist]
- [ ] https://github.com/carbon-design-system/carbon/issues/16364
- [ ] https://github.com/carbon-design-system/carbon/issues/16365
- [ ] https://github.com/carbon-design-system/carbon/issues/16366
- [ ] https://github.com/carbon-design-system/carbon/issues/16367
- [ ] https://github.com/carbon-design-system/carbon/issues/16368
- [ ] https://github.com/carbon-design-system/carbon/issues/16369
- [ ] https://github.com/carbon-design-system/carbon/issues/16370
- [ ] https://github.com/carbon-design-system/carbon/issues/16371
- [ ] https://github.com/carbon-design-system/carbon/issues/16372
- [ ] https://github.com/carbon-design-system/carbon/issues/16373
- [ ] https://github.com/carbon-design-system/carbon/issues/16374
- [ ] https://github.com/carbon-design-system/carbon/issues/16375
- [ ] https://github.com/carbon-design-system/carbon/issues/16376
- [ ] https://github.com/carbon-design-system/carbon/issues/16377
- [ ] https://github.com/carbon-design-system/carbon/issues/16378
- [ ] https://github.com/carbon-design-system/carbon/issues/16379
- [ ] https://github.com/carbon-design-system/carbon/issues/16380
- [ ] https://github.com/carbon-design-system/carbon/issues/16381
- [ ] https://github.com/carbon-design-system/carbon/issues/16382
- [ ] https://github.com/carbon-design-system/carbon/issues/16383
- [ ] https://github.com/carbon-design-system/carbon/issues/16384
- [x] TileGroupProps (completed via #16384)
```
We've also noticed some components are missing an `index.ts` or `.d.ts` file - this has caused a similar issue when trying to import props belonging to that component. Below is a list of the components we've found with this issue:
```[tasklist]
- [x] MultiSelect (completed via #16147)
- [ ] https://github.com/carbon-design-system/carbon/issues/16386
- [ ] https://github.com/carbon-design-system/carbon/issues/16385
- [ ] https://github.com/carbon-design-system/carbon/issues/16444
- [ ] https://github.com/carbon-design-system/carbon/issues/16547
```
### The solution

To fix this, we need to export the props (see attached for example of an interface lacking an export).

image

There are also some instances where a prop is missing an `index.ts` or `.d.ts` file - in those cases, we should be able to fix this issue by refactoring an existing `index.js` to `index.ts`. See below for an example of a healthy `index.ts` file.

image

### Examples

_No response_

### Application/PAL

IBM Storage Fusion

### Business priority

### Available extra resources

_No response_

### Code of Conduct

Gururajj77:
Closes #16154

Created a Draft PR to get the export reviewed

#### Changelog

**Changed**

- Exporting AccordionProps

### DIFF
--- a/packages/react/src/components/Accordion/index.tsx
+++ b/packages/react/src/components/Accordion/index.tsx
@@ -5,9 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import Accordion from './Accordion';
+import Accordion, { AccordionProps } from './Accordion';
 
 export default Accordion;
 export { Accordion };
+export type { AccordionProps };
 export { default as AccordionItem } from './AccordionItem';
 export { default as AccordionSkeleton } from './Accordion.Skeleton';
--- a/packages/react/src/components/Accordion/Accordion.tsx
+++ b/packages/react/src/components/Accordion/Accordion.tsx
@@ -11,7 +11,7 @@ import PropTypes from 'prop-types';
 import React, { PropsWithChildren } from 'react';
 import { AccordionProvider } from './AccordionProvider';
 
-interface AccordionProps {
+export interface AccordionProps {
   /**
    * Specify the alignment of the accordion heading
    * title and chevron. Defaults to `end`.


packages/react/src/components/Accordion/index.tsx
```diff
@@ -5,9 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import Accordion from './Accordion';
+import Accordion, { AccordionProps } from './Accordion';
```
tay1orjones, line 8:
```suggestion
import Accordion, { type AccordionProps } from './Accordion';
```

packages/react/src/components/Accordion/index.tsx
```diff
@@ -5,9 +5,10 @@
...
 
 export default Accordion;
 export { Accordion };
+export type { AccordionProps };
```
tay1orjones, line 12:
```suggestion
export { Accordion, type AccordionProps };
```

tay1orjones (changes requested):
We probably should be as explicit as possible with both the imports and exports. 


--- 

Do you think we could batch all the other ones into one PR or would you prefer to do separate ones for each listed in https://github.com/carbon-design-system/carbon/issues/16154

Also see my note over there: https://github.com/carbon-design-system/carbon/issues/16154#issuecomment-2089046474 - I think the expectation here for now is that we'll only be able to make these available through fully qualified import paths, e.g. 

```
import { type AccordionProps } from "@carbon/react/es/components/Accordion";
```

Does that match what you've understood so far here working on this? Maybe I'm incorrect.

Gururajj77:
This was a demo PR, actual PRs for the export are raised: #16447 and #16448

PR #16303, opened 2024-05-02
Added Popover example to DataTable filtering story

### LINKED ISSUE #15339: [Feature Request]: Add `Popover` example to DataTable filtering story
The [`DataTable / Filtering / Default`](https://react.carbondesignsystem.com/?path=/story/components-datatable-filtering--default) story in the react storybook showcases a pattern of filtering via search.

https://github.com/carbon-design-system/carbon/assets/3360588/4246c5ca-2a20-4b16-8ec0-0460a2c3a344

It would be ideal if this story could be enhanced to include a Popover in the TableToolbarContent, that would display some filter options as defined in the ["Multiple filters with batch updates"](https://carbondesignsystem.com/patterns/filtering/#multiple-filters-with-batch-updates) section of the Filtering pattern on the carbon website.

image

This came up in office hours today, [recapped on slack here](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1701876307858629?thread_ts=1701809609.941289&cid=C2K6RFJ1G).

guidari:
Closes #15339

The idea wast to build a filter functionality off the Datatable so the users can grab the code as a template example to build their own filter.

#### Changelog

**New**

- New template code in Datatable Filtering stories

#### Testing / Reviewing

- Make sure the overview.mdx make it clear the step to implement the `TableToolbarFilter.tsx` 
- Test the Datatable Filtering stories


packages/react/src/components/DataTable/DataTable.mdx
```diff
@@ -278,6 +278,71 @@ the `onInputChange` function provided to you from `DataTable`'s render prop and
 pass it to the `onChange` prop of `TableToolbarSearch` in your `TableToolbar`
 component.
 
+### Filtering with checkboxes
```
tay1orjones:
```suggestion
### Multiple filters with batch updates
```

Just to match the section on the [website](https://carbondesignsystem.com/patterns/filtering/#multiple-filters-with-batch-updates)

packages/react/src/components/DataTable/DataTable.mdx
```diff
@@ -278,6 +278,71 @@ the `onInputChange` function provided to you from `DataTable`'s render prop and
...
+
+To provide you with more examples, we created a separate file as a template for
+you to implement a filter using checkboxes.
```
tay1orjones:
```suggestion
The filtering story shows an example of how to implement the [batch filtering](https://carbondesignsystem.com/patterns/filtering/#multiple-filters-with-batch-updates) pattern.
```

packages/react/src/components/DataTable/DataTable.mdx
```diff
@@ -278,6 +278,71 @@ the `onInputChange` function provided to you from `DataTable`'s render prop and
...
+
+To implement that in your code you first have to copy the
+[`TableToolbarFilter`](https://github.com/carbon-design-system/carbon/blob/15339-datatable/packages/react/src/components/DataTable/stories/examples/TableToolbarFilter.tsx)
+component into your code. Then you just have to import inside the
+`TableToolbarContent` like that:
```
tay1orjones:
```suggestion
To implement this in your own project you can copy the
[`TableToolbarFilter`](https://github.com/carbon-design-system/carbon/blob/15339-datatable/packages/react/src/components/DataTable/stories/examples/TableToolbarFilter.tsx)
example component into your code. You can then import and use it inside `TableToolbarContent`:
```

packages/react/src/components/DataTable/DataTable.mdx
```diff
@@ -278,6 +278,71 @@ the `onInputChange` function provided to you from `DataTable`'s render prop and
...
+<TableToolbarFilter
+  onApplyFilter={handleTableFilter}
+  onResetFilter={handleOnResetFilter}
+/>
+```
+
+Now, in the same file where you have your DataTable with your data, you just
+need to write the code to handle the filters. You can use the code below as a
+template and then customize it as needed. Make sure to replace "YOUR_DATA" with
+the initial data you want to pass into the DataTable.
```
tay1orjones:
```suggestion
You can write the code to handle the filters in the same file where you have your DataTable with your data. Use the code below as a
template and then customize it as needed. Make sure to replace "YOUR_DATA" with
the initial data you want to pass into the DataTable.
```

packages/react/src/components/DataTable/DataTable.mdx
```diff
@@ -278,6 +278,71 @@ the `onInputChange` function provided to you from `DataTable`'s render prop and
...
+    });
+  }
+};
+
+const handleOnResetFilter = () => {
+  setRenderedRows(rows);
+};
+```
+
+Finally pass in the Datatable props the value from the `useState`.
```
tay1orjones:
```suggestion
Finally, pass the array of rows from the `useState` into the DataTable.
```

tay1orjones (approved):
This looks so good! 🔥 

Just a couple minor edits on the docs, otherwise this looks ready to merge to me.

alisonjoseph (approved):
Looks good!

alisonjoseph (changes requested):
Sorry, just went to close my preview window and realized the popover colors are different for settings and filter. I think settings is correct, but we should check with design if not sure. Functionality and docs look great though 🔥
Screenshot 2024-05-07 at 7 47 11 AM
Screenshot 2024-05-07 at 7 47 33 AM

guidari:
> Sorry, just went to close my preview window and realized the popover colors are different for settings and filter. I think settings is correct, but we should check with design if not sure. Functionality and docs look great though 🔥

Ohh! Good catch. I'm gonna take a look and check with design if they want to change something!

thyhmdo:
Good catch @alisonjoseph yeah the color should be the same with the settings' popover, or one layer up from it. 
I also noticed another thing is that these two buttons can be opened at the same time? Is there a way to close one popover if the user chooses to open another one? 

https://github.com/carbon-design-system/carbon/assets/77697351/729d4e7f-d9da-43f2-b5dc-edc773336541

guidari:
> Good catch @alisonjoseph yeah the color should be the same with the settings' popover, or one layer up from it. I also noticed another thing is that these two buttons can be opened at the same time? Is there a way to close one popover if the user chooses to open another one?

Fixed @thyhmdo ! Looks like the Popover gets an error when clicks on an OverflowMenu, which results in the state being kept open instead of closing it.
I can open an issue to track that.

alisonjoseph (approved):
LGTM!
Screenshot 2024-05-08 at 8 00 35 AM

alisonjoseph (changes requested):
Percy caught an issues with different themes

alisonjoseph (approved):
🔥

PR #15399, opened 2023-12-15
feat: Type pagination

Tresau:
Closes #13566

adds TypeScript types to Pagination and PaginationSkeleton. Depends on #15057

#### Changelog

- TypeScript types for Pagination and PaginationSkeleton

**Changed**

- .js -> .tsx and .ts extentions

**Removed**

- N/A

#### Testing / Reviewing

Verify that unit test run, verify that storybook works as intended

Issue #9754, opened 2021-09-27
[Feature Request]: <Pagination> to handle very large numbers of items


s100:
### Summary

We would like some sort of mechanism to make it so that when there are, say, 40,000,000 items to page through, at 25 items per page `<Pagination>` does *not* attempt to generate a dropdown with 1,600,000 options in it.

### Justification

We recently had a customer hit exactly this issue. They had 40,000,000 items and the `<Pagination>` component was causing their browser to apparently stall or crash while attempting to generate a dropdown list of all the pages. Simply omitting the `<Pagination>` and showing an apologetic message is one option, but it would be nice to do a little better than that.

### Desired UX and success metrics

I don't have a full picture of how I would like this to look, beyond "it shouldn't crash the browser, no matter how many items there are"... capping the dropdown size at 1000 entries would be an acceptable start. This is what we are currently doing, manually.

A better approach would be to have some kind of exponential navigation where we are not provided with access to every page, but a representative sample which is sufficient for us to navigate. Say, pages 1-10, 20, 30, 40, ..., 100, 200, 300, ... 1,599,900, 1,599,910, ..., 1,599,980, 1,599,990-1,600,000. If the current page number is somewhere in the middle, provide exponential options around that point, too.

### Required functionality

The existing functionality of the `<Pagination>` component is acceptable. This I think is mainly an alteration to how it generates its page selections.

### Specific timeline issues / requests

No specific timeline... we are working around the problem right now.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

tay1orjones:
Hi @s100, this has come up a few times

- #6836
- #5129
- #6146
- #4365

The easiest way to prevent this is to cap the amount of data sent to the table via your backend/API. Alternatively you could prevent the `totalItems` prop from being set to a number that causes the browser to break, and limit the amount of pages put into the select.
I think something like this could work:

```
// if we had a prop called `maxPages` and it was set to 100 by 
// default, totalItems could be calculated based on the selected pageSize

totalItems={
  totalItems < maxPages * pageSize
    ? totalItems
    : maxPages * pageSize
}
```

[...]

Issue #5546, opened 2020-03-05
Design for pagination number input 


designertyler:
### Summary
A large number of total items in the pagination select causes problems described in https://github.com/carbon-design-system/carbon/issues/5129 and https://github.com/carbon-design-system/carbon/issues/4365

Without alternative designs, a proposed workaround is to remove the ability to select a page or limit the height of the select window https://github.com/carbon-design-system/carbon/pull/5485 

### Desired UX and success metrics
Create designs that allow the user to navigate to a specific page without incrementally clicking the forward and back pagination nav arrow.

### Considerations:
- [ ] Error message
- [ ] Always require a value
- [ ] The behavior when the number entered is greater than the number of pages (Reset to the highest page?)
- [ ] What other components have this type of input inside a toolbar or component? What needs to be established for this type of pattern? 

### Exploration
Number input specs that are similar to the mobile ipnut component.
![image](https://user-images.githubusercontent.com/19270502/75998210-18278480-5ec6-11ea-82cd-2654701a228f.png)

aagonzales:
**Next steps:**
- define dynamic solution
- determine at what number range should a number input be used vs select
- create spec 
   - is it just an input or does it have the number input controls.
- write usage guidelines

tusharcoepit:
Team, do we have any timeline by when number input support will be available ? Is using `unstable_Pagination` the only option till then ?

devadula-nandan:
pagination today supports

> `renderPageSelect`  Provide a custom render function for the page-selection control. When provided, the default `<Select>` is replaced with the returned node.

which basically inverts the control to the implementation as per their use case.
people now can render
[number inputs](https://react.carbondesignsystem.com/?path=/story/components-pagination--with-render-page-select)
and whatever the design decision is.

PR #16306, opened 2024-05-02
fix(Pagination): handle unknown totalItems again

### LINKED ISSUE #16298: [Bug]: Pagination behavior change for unknown total items
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.56.0

### React version

v16

### Description

The gist of the issue is the next page button gets disabled when passing `undefined` for total items. In a case where the user is doing server-side paging, they don't know the total items, so we were passing `undefined` to represent that. 

[This PR](https://github.com/carbon-design-system/carbon/pull/15399) added a default to `totalItems` in Pagination. Previously, we were passing `undefined` for totalItems in the case of server-side paging, which worked fine. Now, `totalItems` gets defaulted to 1 and the next page button gets disabled. 

It looks like typescript was complaining when the default wasn't there, but we can make other changes to make typescript happy. Was there another reason the default needed to be added? I wouldn't think so; doesn't seem logical to just default to 1 total item.

I can put out a PR to fix this too, but would like confirmation on the expectation.  There's also another oddity with how the dropdown works for pages in this scenario that we can talk more about on Slack.

### Reproduction/example

https://stackblitz.com/edit/github-uanwpm?file=src%2FApp.jsx

### Steps to reproduce

Set up pagination with pagesUnknown and don't pass totalItems.

### Suggested Severity

### Application/PAL

cloud-pal-community and consumers in IBM

### Code of Conduct

finken2:
Closes #16298 

[a previous PR ](https://github.com/carbon-design-system/carbon/pull/15399) broke the ability to use pagesUnknown and totalItems. This fixes that and these things:
- next page button no longer gets disabled
- don't render the Select for page selection when pagesUnknown. (it previously rendered and did nothing)
- Fix text for "page x" when pagesUnknown
- Adds some margin next to that text so it looks better
- Add a story for pagesUnknown and totalItems undefined
- Simplify some of the logic where there were unneeded ternaries and double checks

#### Changelog

**Changed**

- Pagination: revert previous change that defaulted totalItems to 1
- Pagination: don't render non-functional page dropdown when pagesUnknown

#### Testing / Reviewing

New story in storybook
unit test added
send totalItems undefined 

image


packages/styles/scss/components/pagination/_pagination.scss
```diff
@@ -174,6 +174,11 @@
     margin-inline-start: 1rem;
   }
 
+  .#{$prefix}--pagination__right
+    .#{$prefix}--pagination__text.#{$prefix}--pagination__page-text.#{$prefix}--pagination__unknown-pages-text {
+    margin-inline-end: 1rem;
```
tay1orjones:
```suggestion
    margin-inline-end: $spacing-05;
```

tay1orjones (approved):
This is a really solid improvement, thanks for tackling it! I'm not sure why we didn't do something like this earlier. Your solution here is much better than what I was previously recommending in https://github.com/carbon-design-system/carbon/issues/9754

I do think a better long-term alternative to not rendering the select would be a number input, https://github.com/carbon-design-system/carbon/issues/5546

alisonjoseph (approved):
Nice!

PR #16327, opened 2024-05-06
fix: slider component does not handle zero properly

### LINKED ISSUE #15916: [Bug]: Slider component does not handle zero properly when it's outside of min/max values
### Package

@carbon/react

### Browser

Chrome

### Package version

"@carbon/react": "^1.52.0"

### React version

v17

### Description

While using the slider component and setting a min/max that excludes 0 (zero), then using the input box to type in 0, then moving focus out of that input box (onBlur event).
The box is populated with the lowest/highest allowed value (expected) but the warning message `The inputted value "<any non zero number>" was corrected to the nearest allowed digit.` is not present. Instead it's just the number 0.

When blurring away with a non zero number
Screenshot 2024-03-07 at 12 49 56 PM

When blurring away using zero
Screenshot 2024-03-07 at 12 52 42 PM

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-slider--playground&args=min:-10;max:-1&globals=theme:white

### Steps to reproduce

Described above. Can be re-produced using the official storybook

### Suggested Severity

### Application/PAL

Cloud pak for AIOps

### Code of Conduct

preetibansalui:
Closes #15916 

Slider Component does not handle Zero properly

#### Testing / Reviewing

1. Go to slider and add non zero items as min & max values (eg: -10 to -1)
2. Now enter zero in input box, it should auto correct the value to nearest possible value.
3. It should display the auto correct message properly.


Gururajj77 (approved):
LGTM 🚀

2nikhiltom (approved):

guidari (approved):
Tested and LGTM!

PR #16346, opened 2024-05-08
fix(popover): Deprecate PopoverAlignment type values

### LINKED ISSUE #16009: Properly deprecate PopoverAlignment type values

### LINKED ISSUE #16326: Deprecate PopoverAlignment type values
```[tasklist]
### Tasks : part of <https://github.com/carbon-design-system/carbon/issues/16009>
- [x] Pull the old values out into their own type, mark it as deprecated via jsdoc
- [x] Update PopoverAlignment to be a union of the new values and the new deprecated type
- [x] Write a custom validator function for proptypes that pops a warning when a deprecated value is used
```

### LINKED ISSUE #16467: Refactor deprecateValuesWithin helper to support optional propMappingFunction as argument
Issue : https://github.com/carbon-design-system/carbon/issues/16326
Enhancement of [this PR](https://github.com/carbon-design-system/carbon/pull/16346)

Change the signature of `deprecateValuesWithin` to accept below as it's arguments
```
  propType,
  allowedValues
  propMappingFunction
```
Create an `propMappingFunction` function and optionally pass it to `deprecateValuesWithin` which can guide the user more clearly about the `depreciated` prop value and it's `new_alternative_value`

2nikhiltom:
Closes #16326 #16467

In https://github.com/carbon-design-system/carbon/pull/14654 all Popover-based components were unified to use new values for the align prop.

This PR does the below 
- Deprecates old values so we can ensure consumers know to stop using them.
- PopoverAlignment is now a union of both  : `PopoverAlignment = DeprecatedPopoverAlignment | NewPopoverAlignment;`

- Adds a custom validator function for proptypes that pops a warning when a deprecated value is used

#### Changelog

**New**

Adds `deprecateValuesWithin` function to warn when deprecated value is used for `align` prop
Pulls the old values out into their own type : `DeprecatedPopoverAlignment` and  marks it as deprecated

#### Testing / Reviewing

Pass any old value for align pop in Popover component and verify that console warns about deprecated usage

```
<div>
      <Popover open={true} align="top-left" autoAlign></Popover>;
 </div>
```


packages/react/src/components/Popover/index.tsx
```diff
@@ -402,34 +411,52 @@ if (__DEV__) {
...
+      'top-start',
+      'top-end',
+      'bottom-start',
+      'bottom-end',
+      'left-end',
+      'left-start',
+      'right-end',
+      'right-start',
+    ]),
+    "The alignment values for 'align' prop : 'top-left', 'top-right', 'bottom-left', 'bottom-right', 'left-bottom', 'left-top', 'right-bottom', and 'right-top' are deprecated. Please use 'top-start', 'top-end', 'bottom-start', 'bottom-end', 'left-start', 'left-end', 'right-start', or 'right-end' respectively",
```
preetibansalui: Please check the sequence of new suggested props, I believe its not correct for 'left-start', 'left-end', 'right-start', or 'right-end'.

packages/react/src/prop-types/deprecateValuesWithin.js
```diff
@@ -0,0 +1,43 @@
...
+      didWarnAboutDeprecation[componentName] = {
+        ...didWarnAboutDeprecation[componentName],
+        [propName]: true,
+      };
+
+      if (allowedValues && !allowedValues.includes(props[propName])) {
+        warning(
+          false,
+          message ||
+            `The value for \`${propName}\` prop has been deprecated, It will be removed in the next major release. Allowed values is/are: ${allowedValues.join(
```
preetibansalui: Can we add component name as well in warn message.

preetibansalui (changes requested):
Few minor changes only, rest is looking fine to me.

packages/react/src/prop-types/deprecateValuesWithin.js
2nikhiltom: Sure! Updated the same, also this string will only be displayed when `message` argument is not passed.

preetibansalui (approved):
LGTM :+1:

packages/react/src/prop-types/deprecateValuesWithin.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import { warning } from '../internal/warning';
+
+const didWarnAboutDeprecation = {};
+
+export default function deprecateValuesWithin(
+  propType,
+  message,
+  allowedValues
```
tay1orjones: I think this order of the params might be better? Also instead of a configurable message, what if we were able to pass in a mapping function that would be called (if available) to generate the message?

```suggestion
  propType,
  allowedValues
  propMappingFunction
```

packages/react/src/prop-types/deprecateValuesWithin.js
```diff
@@ -0,0 +1,43 @@
...
+      didWarnAboutDeprecation[componentName] = {
+        ...didWarnAboutDeprecation[componentName],
+        [propName]: true,
+      };
+
+      if (allowedValues && !allowedValues.includes(props[propName])) {
+        warning(
+          false,
+          message ||
+            `In \`${componentName}\` component value for \`${propName}\` prop has been deprecated, It will be removed in the next major release. Allowed values is/are: ${allowedValues.join(
```
tay1orjones: Here we can optionally call that mapper function so we can guide the user more clearly and cut down the message length. If we don't have the mapper, we can instead provide a list of all the valid options.


```suggestion
            `\`${props[propName]}\` is a deprecated value for the \`${propName}\` prop on the \`${componentName}\` component. \`${propMappingFunction ? \`Use ${propMappingFunction(props[propName])} instead\` : \`The allowed values for \`${propName}\` include: ${allowedValues.join(', ')}\`}\`. \`${props[propName]}\` will be removed in the next major release.\`
```

With the mapper:
> `top-left` is a deprecated value for the `align` prop on the `Popover` component. Use `top-start` instead. `top-left` will be removed in the next major release. 

Without the mapper:
> `top-left` is a deprecated value for the `align` prop on the `Popover` component. The allowed values for `align` include: 'top', 'top-start', '...'. `top-left` will be removed in the next major release.

packages/react/src/components/Popover/index.tsx
```diff
@@ -402,34 +411,52 @@ if (__DEV__) {
...
+      'bottom-start',
+      'bottom-end',
+      'left',
+      'left-start',
+      'left-end',
+      'right',
+      'right-start',
+      'right-end',
+    ]
+  ),
```
tay1orjones: Then reorder the params and pass in that mapper function

```suggestion
  align: deprecateValuesWithin(
    PropTypes.oneOf([
      'top',
      'top-left', // deprecated use top-start instead
      'top-right', // deprecated use top-end instead

      'bottom',
      'bottom-left', // deprecated use bottom-start instead
      'bottom-right', // deprecated use bottom-end instead

      'left',
      'left-bottom', // deprecated use left-end instead
      'left-top', // deprecated use left-start instead

      'right',
      'right-bottom', // deprecated use right-end instead
      'right-top', // deprecated use right-start instead

      // new values to match floating-ui
      'top-start',
      'top-end',
      'bottom-start',
      'bottom-end',
      'left-end',
      'left-start',
      'right-end',
      'right-start',
    ]),
    [
      'top',
      'top-start',
      'top-end',
      'bottom',
      'bottom-start',
      'bottom-end',
      'left',
      'left-start',
      'left-end',
      'right',
      'right-start',
      'right-end',
    ],
    mapPopoverAlignProp
  ),
```

tay1orjones (approved):
This looks really great! I have a couple suggestions on the signature of `deprecateValuesWithin`. Let me know what you think

packages/react/src/prop-types/deprecateValuesWithin.js
2nikhiltom: Agreed , this would look so much better and clear to users 🔥 .
I will take up this enhancement.

2nikhiltom:
Hey @tay1orjones !
Thanks for the enhancement suggestions 💯 
I have changed the signature of `deprecateValuesWithin` and is ready for the review

alisonjoseph (approved):

tay1orjones (approved):
Awesome stuff!

PR #16437, opened 2024-05-10
New Tag Overflow

### LINKED ISSUE #15893: [Feature Request]: change tag title overflow content to truncate and ellipsis

guidari:
Closes #15893
This reference the [old PR](https://github.com/carbon-design-system/carbon/pull/15959)

This PR will add the ellipses to tags that have a long text. When the ellipses is active the Tooltip will added so the user can read the whole tag.

The logic to use the `text` prop in the primitive component couldn't be applied here because of the difference between the way the `Tooltip` would be displayed in the Dismissable and ReadOnly Tags.

#### Changelog

**New**

- `max-inline-size` to tags of 208px.
- Tooltip added when ellipses is active in a Tag.
- Focus is active in Read-only and Dismissible tag if the ellipses is active.
- Added prop `text` to Interactive tags. It will avoid the user to add a `children` inside the InteractiveTag.
- ReadOnly tag is using `DefinitionTooltip` and the rest of the tags are using `Tooltip`

**Removed**

- Remove `children` prop from interactive tags.

#### Testing / Reviewing

- Go to all Tags variants to test the Ellipses and Tooltip functionality.
- Validate if the prop `text` make sense


laurenmrice (changes requested):
Looking really good @guidari ! Just a couple things below and then it should be good to go! Thank you for all of your patience and help with this work! ⭐️


---

###   Tag (stable)

**Read only story**
- The red tag that is truncated does not have the correct text color. The color should be `$tag-color-red`. The text is also sitting too high in the container, it should be vertically center. (This looks like a Firefox bug specifically.)
- The cursor should be the arrow and not the pointer. I am also not seeing the browser based tooltip on hover in Chrome and Firefox. It appears in Safari.

### Tag (unstable)

**Dismissible story**
- When hovering over the text in the tag, the cursor should be an arrow. It only should turn to a pointer cursor when it hovers over the close icon area.
- Disabled state: The truncated tags border gets removed, it should still be present in the disabled state like the other tags.
- Disabled state: Parts of the tag in the first interactive example is still interactive but should not be.

**Operational story**
- The browser tooltip text says “aaa” but it should say the full title of the tag. 

**Selectable story**
- Looks great!

laurenmrice (approved):
Looks good! Thank you @guidari ⭐️

alisonjoseph:
This is looking great, just seeing some weirdness when you click on Operational or Selectable Tags with the tooltip flashing. 

https://github.com/carbon-design-system/carbon/assets/2753488/bccf3907-6956-4e43-8130-fd96ef13f2be

alisonjoseph (approved):

tay1orjones (approved):

Kilian-Collender:
@alisonjoseph @tay1orjones this PR is causing lots of issues for our teams as there is a query selector being used against an id but without any css escaping
`const elementTagId = document.querySelector(`#${tagId}`);`

I'm not 100% sure why this even needs to be a querySelector, but it the id contains reserved characters in css it breaks

Kilian-Collender:
Also there is no validation of the element being present in,  fails when trying to access offsetWidth
```
const isEllipsisActive = (element: any) => {
    setIsEllipsisApplied(element.offsetWidth < element.scrollWidth);
    return element.offsetWidth < element.scrollWidth;
  };
```

guidari:
Hey @Kilian-Collender 
Thanks for reporting that!

wkeese:
This was a disturbing change because it broke backward compatibility between Carbon 11 and an earlier version of Carbon 11.

Specifically that you used to specify the tag text as children (ex: `<DismissibleTag>hello</DismissibleTag>`) but now you need to specify it via the `text` attribute: (ex: `<DismissibleTag text="hello"/>`).

PR #16466, opened 2024-05-14
fix(menu): include leave intent delay for sub menus

### LINKED ISSUE #16395: [Bug]: Menu component does not protect access to refs in focusItem
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.49.0

### React version

v18.2.0

### Description

In the findIndex function here https://github.com/carbon-design-system/carbon/blob/e70f08d867fc276f7a77e2d703a0462afae545ab/packages/react/src/components/Menu/Menu.tsx#L237

It tries to use `contains` against the current ref value without checking that the ref value is not null. 

During our automations this is throwing the following error `Cannot read properties of null (reading 'contains')`  as `item.ref.current` is null .

This requires just a simple check to ensure that the value is not `null` before calling `contains`

### Reproduction/example

NA

### Steps to reproduce

Hard to reproduce consistently manually as it only really happens when the interaction is very fast, but it has been possible to see this consistently in automations.

The key part to the issue is that the submenu items are being created in async manner, so the refs are not being maintained.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Kilian-Collender:
Closes https://github.com/carbon-design-system/carbon/issues/16395

[Menu component does not protect access to refs in focusItem](https://github.com/carbon-design-system/carbon/issues/16395) by adding protection to the accessing of the refs.

#### Changelog

**New**

Adds leave intent delay for sub menu as it can accidentally trigger a premature closure of the sub menu in the following cases:

1. the mouse moves across and down as the user selects a lower sub menu, causing the mouse to leave the menuItem.
2. the menu has a scroll bar, as the mouse crosses this to enter the sub menu it triggers the closure before the user can access the sub menu.

Including a delay in the closure allows for the users mouse to reenter the submenu and cancel the closure of the menu.

**Changed**

- The sub menu does not close immediately when the mouse leave the menu item or menu items children.
- Add optional menu item aria-label that uses prop or label value

#### Testing / Reviewing

Unit tests have been added to validate the opening and closing of the sub menu using fake timers.

Manual testing can be easily done by opening the submenu and quickly moving the mouse out and back into the menu item and the menu will close.

### DIFF
--- a/packages/react/src/components/Menu/MenuItem.tsx
+++ b/packages/react/src/components/Menu/MenuItem.tsx
@@ -80,6 +80,7 @@ export interface MenuItemProps extends LiHTMLAttributes<HTMLLIElement> {
 }
 
 const hoverIntentDelay = 150; // in ms
+const leaveIntentDelay = 300; // in ms
 
 export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
   function MenuItem(
@@ -110,6 +111,10 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
     const hasChildren = Boolean(children);
     const [submenuOpen, setSubmenuOpen] = useState(false);
     const hoverIntentTimeout = useRef<ReturnType<typeof setTimeout> | null>(
+      null
+    );
+
+    const leaveIntentTimeout = useRef<ReturnType<typeof setTimeout> | null>(
       null
     );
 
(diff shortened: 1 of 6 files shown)


tay1orjones (approved):
Excellent, thanks for contributing this! 🙏

alisonjoseph (approved):
LGTM!

tay1orjones:
@Kilian-Collender I can reproduce the avt error in storybook

image

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -241,6 +256,7 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
         aria-disabled={isDisabled ?? undefined}
         aria-haspopup={hasChildren ?? undefined}
         aria-expanded={hasChildren ? submenuOpen : undefined}
+        aria-label={ariaLabel || label}
```
tay1orjones: @Kilian-Collender I'm assuming this why the `avt` CI check is failing. What was the reasoning for adding this one?

Kilian-Collender: I can remove it again, but I had do it to make it easier to label and access the menuItem in the tests. I'm not 100% sure why it is complaining in the AVT as the visible label should match or include the text

PR #16476, opened 2024-05-15
Tooltip fixes on hover and ESC

### LINKED ISSUE #16152: [a11y]: tooltip's behaviour fails WCAG in two ways
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

@1.54.0

### React version

https://react.carbondesignsystem.com/?path=/story/components-tooltip--default

n/a

### Assistive technology

n/a

### Description

There are two issues with the current tooltip behaviour:
1. It should be dismissible with Esc, even if activated by hover
2. If I move the pointer over top of the exposed tooltip, it should not go away.

These are described in https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html as the Dismissible and Hoverable requirements.

### WCAG 2.1 Violation

Content on Hover or Focus (Level AA)

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tooltip--default

### Steps to reproduce

1. Got to https://react.carbondesignsystem.com/?path=/story/components-tooltip--default
2. Hover over the i icon
3. Press Esc. Nothing happens. Expected behaviour is the tooltip should disappear. 
4. With the tooltip still visible, move the pointer so it is over top of the tooltip. It disappears. Expected behaviour is the tooltip should remain visible so long at the pointer is over either the target of the tooltip (so that a low vision user can move their magnification to it and read the tip)

### Code of Conduct

guidari:
Closes #16152

Added an `useIsomorphicEffect` and `useCallBack` to handle the ESC key when the user is hovering the `Tooltip`.

#### Changelog

**New**

- Added `onMouseEnter` in `PopoverContent` to keep the state open when hovering the content.
- Added test for the Escape key while the tooltip trigger is on hover state.

#### Testing / Reviewing

- Go to `Tooltip` default stories
- Make sure there is no regressions.


packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -140,19 +141,39 @@ function Tooltip<T extends React.ElementType>({
     triggerProps['aria-describedby'] = id;
   }
 
-  function onKeyDown(event: React.KeyboardEvent) {
-    if (open && match(event, keys.Escape)) {
-      event.stopPropagation();
```
tay1orjones, line 149: Was this originally in here to avoid impacting other components on the page that may be listening to keydown on the escape key? Like if the tooltip is within a modal, maybe?

tay1orjones (commented):
Functionally looks good, though there's one thing to note, this may continue to act oddly in the storybook due to the story being in an iframe.

If focus is not within the iframe the hover interaction will work but the <kbd>escape</kbd> key does not. Clicking into the iframe and then trying again the <kbd>escape</kbd> key works and closes it. I don't think there's anything to fix here, but if the issue is reported again I'd like to have this comment to point to 😄 


https://github.com/carbon-design-system/carbon/assets/3360588/b2a3725c-41ba-4755-ad70-a05ba7e3c998

packages/react/src/components/Tooltip/Tooltip.tsx
guidari: Ohh good catch! I didn't notice I removed that

tay1orjones (approved):

alisonjoseph (approved):

PR #16478, opened 2024-05-16
fix: useTheme removed from API in error

lee-chase:
Addresses API change made in error in #16230

Removal of useTheme changes the API removing the ability to check the current theme.

NOTE: Perhaps tests of components and functions exposed in the public API should import directly from `packages/react/src/index`. 

#### Changelog

**Changed**

- packages/react/src/index

#### Testing / Reviewing

Ran tests.


tay1orjones (approved):
Wow, thanks for catching this. That's 100% my fault, no clue how I missed that

guidari (approved):

PR #16490, opened 2024-05-16
fix(menu): remove explicit color token from icons

janhassel:
I noticed that the icons in Menus (MenuButton, ComboButton) didn't render in the correct color when the item was disabled or danger (hover). This PR removes the explicit color definition and therefore relies on the default `currentColor` to sync the icons with the text.

| Scenario | Screenshot |
| :- | :- |
| Before – disabled | image |
| After – disabled | image |
| Before – danger | image |
| After – danger | image |

#### Changelog

**Removed**

- Removed explicit color definition from icons in menus

#### Testing / Reviewing

Storybook

@alisonjoseph I noticed that you explicitly mentioned adding colors to the svgs in #15236. I tested several scenarios in storybook and the color seems to be inherited correctly, but I wanted to make sure I'm not missing an edge case you found?

### DIFF
--- a/packages/styles/scss/components/menu/_menu.scss
+++ b/packages/styles/scss/components/menu/_menu.scss
@@ -77,19 +77,11 @@
     &:focus {
       @include focus-outline('outline');
     }
-
-    svg {
-      color: $icon-secondary;
-    }
   }
 
   .#{$prefix}--menu-item:hover {
     background-color: $layer-hover;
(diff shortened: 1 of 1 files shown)


alisonjoseph (approved):
This LGTM, all variations seem to work. I looked over my previous PR and I'm not sure why I added these in.

tay1orjones (approved):

PR #16492, opened 2024-05-17
feat: Add floating ui to dropdown

### LINKED ISSUE #15865: Add floating-ui to `@carbon/react` Dropdown
Similar to how we've updated Popover to use floating-ui for autoAlign positioning, the Dropdown needs to have this as well.

At a high level this includes:
- [ ] Adding an `autoAlign` prop to Dropdown that should turn off most/all listbox styling
- [ ] This might also require altering the ListBox and it's related components as well, adding `autoAlign` to those components to turn off styling.
- [ ] When `autoAlign` is used, only the `top` or `bottom` alignment options should be used
- [ ] Add an `autoAlign` story to the 
- [ ] Once complete, validate that Dropdown is no longer clipped by modals or any other parent container with overflow: hidden - document this in storybook for review (can be removed before merging)
- [ ] Ideally this could also include some new abstractions of common code shared between the floating-ui implementation of Popover and Dropdown, to be further used when adding floating-ui functionality to other components.

preetibansalui:
Closes #15865 
 
Implement Floating UI in Dropdown

**Changed**

Implement floating ui in Dropdown


AustinGitHub:
Hi @preetibansalui do you know what the status of this PR is and ETA of it being merged in? I am told that this will resolve an issue with the scrolling issue when we use a dropdown that would resolve an issue like this https://github.com/carbon-design-system/carbon/issues/7565 

Thanks!

riddhybansal (changes requested):
Do we really need an autoalign prop in this ? Autoalignment should not be conditional in case of dropdown. Floating UI should be added to it by default. @tay1orjones What are your views on this ?

preetibansalui:
> Do we really need an autoalign prop in this ? Autoalignment should not be conditional in case of dropdown. Floating UI should be added to it by default. @tay1orjones What are your views on this ?

As discussed on Slack, we are keeping autoAlign Prop as of now.

riddhybansal (approved):
This LGTM !!!

alisonjoseph (approved):
This LGTM! Do we want to set autoAlign to true in all of the Modal stories and not just the Playground?

preetibansalui:
> This LGTM! Do we want to set autoAlign to true in all of the Modal stories and not just the Playground?

I think it would be better if we remove it from here, As we already have a separate issue where we will be creating stackblitz examples for all related issue. I will remove it, Thanks for bringing this up.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -270,6 +278,46 @@ const Dropdown = React.forwardRef(
     }: DropdownProps<ItemType>,
     ref: ForwardedRef<HTMLButtonElement>
   ) => {
+    const { refs, floatingStyles } = useFloating(
+      autoAlign
+        ? {
+            // placement: direction,
```
tay1orjones: I think leaving this in will instruct floating-ui to prefer the specified direction where possible
```suggestion
            placement: direction,
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -270,6 +278,46 @@ const Dropdown = React.forwardRef(
...
+            // containing block (usually the viewport). It will in many cases also
+            // “break” the floating element out of a clipping ancestor.
+            // https://floating-ui.com/docs/misc#clipping
+            strategy: 'fixed',
+
+            // Middleware order matters, arrow should be last
+            middleware: [
+              flip({
+                fallbackAxisSideDirection: 'none',
+              }),
```
tay1orjones: `none` is the default, I think we can remove this

```suggestion
              flip(),
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -92,6 +94,11 @@ export interface DropdownProps<ItemType>
    */
   ariaLabel?: string;
 
+  /**
+   * Will auto-align the dropdown on first render if it is not visible. This prop is currently experimental and is subject to future changes.
```
tay1orjones: This comment is now outdated (no longer limited to the first render), here's my take at a better one we could use for every component with an `autoAlign` prop:

```suggestion
   * **Experimental**: Will attempt to automatically align the floating
   * element to avoid collisions with the viewport and being clipped by
   * ancestor elements.

File: packages/react/src/components/Dropdown/Dropdown.tsx [hunk 4]
```diff
@@ -270,6 +278,46 @@ const Dropdown = React.forwardRef(
...
+          if (refs.floating.current) {
+            refs.floating.current.style[style] = floatingStyles[style];
+
+            // refs.floating.current.style.width = parentWidth + 'px';
+          }
+        });
+
+        if (parentWidth && refs.floating.current) {
+          refs.floating.current.style.width = parentWidth + 'px';
+        }
```
tay1orjones: I think we can use `size()` instead https://floating-ui.com/docs/size#match-reference-width

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -592,6 +643,11 @@ Dropdown.propTypes = {
     'This prop syntax has been deprecated. Please use the new `aria-label`.'
   ),
 
+  /**
+   * Will auto-align the dropdown on first render if it is not visible. This prop is currently experimental and is subject to future changes.
```
tay1orjones:
```suggestion
   * **Experimental**: Will attempt to automatically align the floating
   * element to avoid collisions with the viewport and being clipped by
   * ancestor elements.
```

tay1orjones (changes requested):
This looks super good, just a couple small things

preetibansalui:
> This looks super good, just a couple small things

Thanks Taylor for the amazing suggestions, just pushed the changes as suggested. Please review the PR again.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -340,6 +390,7 @@ const Dropdown = React.forwardRef(
       [`${prefix}--dropdown--readonly`]: readOnly,
       [`${prefix}--dropdown--${size}`]: size,
       [`${prefix}--list-box--up`]: direction === 'top',
+      [`${prefix}--dropdown--autoalign`]: autoAlign,
```
guidari, line 393: Hey @preetibansalui 
Is this class doing something? I don't see it in the code

preetibansalui: Hey @guidari , we are adding this class only in case when we are using 'autoAlign' as true. And later we are using this class in modal popup to allow menu overflow.

tay1orjones (approved):

PR #16543, opened 2024-05-22
fix: added floating UI to menu and menuButton

### LINKED ISSUE #14296: [Bug]: `MenuButton` menu items remain fixed on page scroll
### Package

@carbon/react

### Browser

Chrome

### Package version

1.34.0

### React version

18

### Description

When a `MenuButton` has menu items visible and you begin to scroll, the button scrolls with the page but menu items remain fixed on the page.

### Reproduction/example

https://stackblitz.com/edit/github-sjnuql?file=src%2FApp.jsx

### Steps to reproduce

* Go to: https://stackblitz.com/edit/github-sjnuql?file=src%2FApp.jsx
* Click on the button to show the menu items
* While the menu is still visible, begin to scroll on the page
* error: the button scrolls, as expected, but the menu items remain fixed and in-view

https://github.com/carbon-design-system/carbon/assets/7853451/877e23d4-a671-4566-876c-39f4d154e785

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

### LINKED ISSUE #15521: Add floating-ui to `@carbon/react` Menu, MenuButton
### Package

@carbon/react

### Browser

_No response_

### Package version

v11.47.0

### React version

18.02

### Description

`ComboButton`, `MenuButton` and `OverflowMenu` use the `Menu` component to render a list of items. When the user open the menu it is calculate a position in the screen to render the `Menu` below the button trigger. That works fine until you have a scrollable screen, because the position is fixed based on the screen and the button iself.

### Reproduction/example

https://stackblitz.com/edit/github-uqfys7?file=src%2FApp.jsx

### Steps to reproduce

Open the menus in the StackBlitz

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #15521
Closes #14296  

This PR enhances MenuButton experimental autoAlign functionality to automatically position the floating element to be within view, even on scroll.

This uses [floating-ui](https://floating-ui.com/) and relies solely on the fixed strategy, using position: fixed on the floating element. In the majority of cases, MenuButton floating elements will no longer be clipped by parents with overflow: hidden.

Changed

- Use floating-ui for positioning  floating element.
- Feed floating-ui the values of menuButton custom properties (flip,size) when possible.
- Added a prop named legacyAutoalign to menu that specifies if the parent element is using floating Ui for positioning the -menu or it sending X,Y coordinates. On the basis of it calculation of positioning the menu is  made conditional.

Removed

- Sending X and Y coordinates for the floating element (Menu) from MenuButton. All the calculations like width of floating element , X and Y positions are calculated by floating UI itself.

Testing / Reviewing

- View the menuButton stories, there is autoalign experiment story in there. Ensure it works as intended and there's no regressions.
- Make sure that the menu does not detach itself from menuButton.


alisonjoseph:
@riddhybansal I think you need to run `yarn test -u` to update the snapshot to get the test to pass.

alisonjoseph:
This is looking great! I'm not sure we want to allow this design? I think probably just up and down?
Screenshot 2024-05-22 at 12 07 26 PM

riddhybansal:
@alisonjoseph Yeah !! The issue has been addressed!!

guidari:
@riddhybansal Nice! 
Looks like the prop `menuAlignment` got broken. You can check on the "Playground" or "With Menu Alignment" stories

Screenshot 2024-05-23 at 08 49 49

riddhybansal:
Hey @guidari , Got it fixed 😄

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -178,6 +200,8 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
 );
 
 MenuButton.propTypes = {
+  autoAlign: PropTypes.bool,
```
alisonjoseph: Missing code comments, something like
 `Will auto-align the MenuButton. This prop is currently experimental and is subject to future changes.`

alisonjoseph (dismissed):
This is looking good, I just have a couple questions around props.

Should `autoAlign` be true or false by default? Seems like its true, but that isn't documented anywhere, just wanted to be sure that was intended?

I'm also unsure what the `legacyAutoalign` prop is doing? Missing some code comments there, I think it also should be `legacyAutoAlign`

packages/react/src/components/MenuButton/index.tsx
riddhybansal: So there should not be any prop called autoalign in menubutton , need to remove this from menubutton as we are removing calculation of x and y coordinates for menu to be align. So size and positions are all handled by floating UI. If we keep autoalign off there will always be issues with menu not being intact with menubutton.  So there is no choice for user for autoalign it will always use floating UI for alignment


leagacyAutoAlign is a prop in menu for the components using floating UI . For example- In menuButton menu is intacted by floating UI , so sending legacyAutoalign as false, so that menu component does not calculate its position and by default its true for the components not with floating UI yet.

guidari (commented):
Nice! `menuAlignment` is working now!!

Just a note about what @alisonjoseph  said, I did a few changes locally to understand the `legacyAutoAlign` prop. I'm not sure if I'm missing something, but I guess we can remove that.

I did remove locally and the code is working as expected with floating UI in `MenuButton` without affecting other components. The calculation can be removed once we address Floating UI to all components using `Menu`, but the calculation doesn't seem to interfere with Floating UI logic.

tay1orjones:
@guidari one concern I have with abandoning the legacy auto align logic is backwards compatibility. With Menu being a primitive it's more likely consumers have built things on top of this. 

Removing the existing styling/calculations and replacing it with floating-ui it may look like an in-place swap but I'm not sure. That's why I thought preserving the old functionality through `legacyAutoAlign` was prudent. 

It might be overly cautious though. What do you think? Also @janhassel you've looked at this a ton, curious if you have any perspective here?

One other note to mention is that when @riddhybansal and I paired up on this we observed the old styling/calculations to be interfering with the floating-ui style and calculations. Leaving both in seems like it would be unnecessary and prone to bugs.

packages/react/src/components/MenuButton/index.tsx
tay1orjones: Yeah MenuButton has always had built in auto-positioning logic, it's just a bit more simple (and buggy) than what we'll be providing through this PR. I agree we can remove `autoAlign` from MenuButton. I don't think any API changes in MenuButton would be required, it'll be the same as it was before this PR.

guidari:
> Removing the existing styling/calculations and replacing it with floating-ui it may look like an in-place swap but I'm not sure. That's why I thought preserving the old functionality through `legacyAutoAlign` was prudent.

That makes sense. Better safe than sorry.

packages/styles/scss/components/menu/_menu.scss
```diff
@@ -44,6 +44,7 @@
   }
 
   .#{$prefix}--menu--open {
+    inline-size: fit-content;
```
tay1orjones: Is this necessary still? I thought since the `${prefix}--menu--shown` was now guarded by `legacyAutoAlign` that this change wouldn't be needed.

tay1orjones (approved):
Looks great! I think the `MenuAlignment` type is good as is.

alisonjoseph (approved):
🔥LGTM!

PR #16571, opened 2024-05-24
fix(tag): use refs to handle component access

### LINKED ISSUE #16570: [Bug]: Tag components use query selector with id value without protection
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.58.0

### React version

v18.2.0

### Description

The following PR https://github.com/carbon-design-system/carbon/pull/16437 introduces the use of a querySelector call in the tag components which uses the id directly. If the id value contains anything that isn't a valid query then it breaks. It should use a ref to access the instance of the Tag DOM node.

Also the helper function `isEllipsisActive` accesses a element prop directly without validation and this causes an error if the element isn't a valid DOM node. 

### Reproduction/example

https://stackblitz.com/edit/github-ejg4ae?file=src%2FApp.jsx

### Steps to reproduce

open browser console and see errors

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Kilian-Collender:
Closes https://github.com/carbon-design-system/carbon/issues/16570

When using the querySelector it is easily broken if the id has reserved characters also the isEllipsisActive helper had no protection for a non element. So now using refs to access the DOM element and the helper has been extracted for reuse and includes extra protection.

#### Changelog

**New**

- {{new thing}}

**Changed**

- Update how the tag components reference themselves

**Removed**

- {{removed thing}}

#### Testing / Reviewing

- Follow the example in the original bug to test that the original issue does not take place.

N.B. There was no testing in place around the recent changes that were added, it is strongly advised that we need tests here both unit and E2E.

### DIFF
--- a/packages/react/src/components/Tag/Tag.tsx
+++ b/packages/react/src/components/Tag/Tag.tsx
@@ -6,7 +6,7 @@
  */
 
 import PropTypes from 'prop-types';
-import React, { useLayoutEffect, useState, ReactNode } from 'react';
+import React, { useLayoutEffect, useState, ReactNode, useRef } from 'react';
 import classNames from 'classnames';
 import { Close } from '@carbon/icons-react';
 import setupGetInstanceId from '../../tools/setupGetInstanceId';
@@ -15,6 +15,7 @@ import { PolymorphicProps } from '../../types/common';
 import { Text } from '../Text';
 import deprecate from '../../prop-types/deprecate';
 import { DefinitionTooltip } from '../Tooltip';
+import { isEllipsisActive } from './isEllipsisActive';
 
 const getInstanceId = setupGetInstanceId();
 export const TYPES = {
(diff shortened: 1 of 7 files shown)


tay1orjones (approved):
Thanks again for the quick fix!

tay1orjones:
Hey real quick before we merge this I'm going to correct the deprecation notices

alisonjoseph (approved):
Thanks for the fix!

guidari:
Thanks for that! I added a `forwardRef` to Tag, so the ref can use in the variants. If the CI pass I think we are good to merge it!

packages/react/src/components/Tag/isEllipsisActive.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright IBM Corp. 2024
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export const isEllipsisActive = (element: any) => {
+  if (element) {
+    console.log('element offste', element?.offsetWidth < element?.scrollWidth);
```
Kilian-Collender: console.log here

packages/react/src/components/Tag/Tag.tsx
```diff
@@ -172,9 +179,9 @@ const Tag = <T extends React.ElementType>({
   }
 
   if (filter) {
-    const ComponentTag = BaseComponent ?? 'div';
+    const ComponentTag = (BaseComponent as React.ElementType) ?? 'div';
     return (
-      <ComponentTag className={tagClasses} id={tagId} {...other}>
+      <ComponentTag ref={tagRef} className={tagClasses} id={tagId} {...other}>
```
Kilian-Collender: should this not be ref too now?

guidari: Actually we can remove the ref from this one. The `filter` prop should be replace with the new variant `Dissmisible Tag` so we don't have to add the new functionality to the old `filter` tag.

Kilian-Collender: but isn't the `Dismissible Tag` experimental? so you still need it there until it is fully deprecated

guidari: I see your point. It is something I will take to the team to talk about it. Since we have implemented in the `Dissmisible Tag` it would be quick to implement on the old filter prop.
But for now the spec we have is only in the new Intereactive Tag.

guidari (approved):

alisonjoseph (approved):

tay1orjones (approved):

PR #16581, opened 2024-05-27
feat(textarea): override chrome autocomplete styles

### LINKED ISSUE #16341: [Bug]: Browser autocomplete changes background color of TextInput when using dark theme
### Package

carbon-components-react

### Browser

Chrome

### Package version

v7.59.22

### Description

While using the dark theme, whenever the browser suggests an autocomplete suggestion and the user selects it, the background color of the TextInput turns light blue. This clashes harshly with the dark theme. The same happens in light theme, but the light blue color works fine there.

### Screenshots

Screenshot 2024-05-07 at 11 22 52 AM

### Steps to reproduce

https://stackblitz.com/edit/github-mzvzu9?file=src%2FApp.jsx

1. Open settings in chrome
2. Choose "Autofill and passwords"
3. Choose "Addresses and more"
4. Enable "Save and fill addresses"
5. Add an address with your name
6. Open the stackblitz link above
7. Click the text input and then select the auto complete suggestion

This was the only way that I could force the browser to use the auto complete feature. However, we have seen this is lots of places on IBM Cloud for various text inputs unrelated to addresses & names.

### Code of Conduct

divya-281:
Closes #

closes #16341

#### Changelog

**Changed**

- Override chrome autocomplete styles


alisonjoseph (changes requested):
We avoid the use of `!important` in the codebase, can you provide more context on why this is needed for this fix? Or, better, see if there is a way to avoid using it?

divya-281:
> We avoid the use of `!important` in the codebase, can you provide more context on why this is needed for this fix? Or, better, see if there is a way to avoid using it?

Screenshot 2024-06-04 at 4 35 35 PM
Hi Alison, Here the styles of the user agent stylesheet was using important I sed important to override them

divya-281:
> We avoid the use of `!important` in the codebase, can you provide more context on why this is needed for this fix? Or, better, see if there is a way to avoid using it?

Screenshot 2024-06-04 at 4 35 35 PM

Hi @alisonjoseph, Here the styles of the user agent stylesheet was using important I used important to override them

tay1orjones:
@alisonjoseph I'm not sure that we can do this even with `!important`. Check [this](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill) out from MDN:

> Note: The user agent style sheets of many browsers use `!important` in their `:-webkit-autofill` style declarations, making them non-overridable by webpages without resorting to JavaScript hacks. 

@divya-281 could you look into what the user agent stylesheet for this includes?

---

Also:

> "For the best browser compatibility use both `:-webkit-autofill` and `:autofill`."

We should add `:autofill` selectors to this block alongside the `:-webkit-autofill` ones.

divya-281:
Screenshot 2024-07-17 at 3 41 57 PM

Hi @tay1orjones the background color and color are the styles that need to be overridden

tay1orjones:
@divya-281 so if we add that selector here, are you able to remove the `!important`?

tay1orjones:
Hey, thanks again for all your work here! It's been a couple months since there was movement here. In the interest of keeping the PR queue focused on actionable in progress items I'm going to close this PR for now.

If/when you're ready to pick this back up feel free to reopen this PR or open a new one.

PR #16585, opened 2024-05-28
feat: Added floating-ui hook usage for `ComboBox`

### LINKED ISSUE #16470: Add floating-ui to `@carbon/react` ComboBox
This issue is for implementing floating-ui within ComboBox. This will be very similar to the implementation of floating-ui in Dropdown, #15865 

The high level tasks are
* Add an experimental `autoAlign` prop
* Add the useFloating hook, only provide a config object when `autoAlign`, an empty config object when `!autoAlign`
* Set the refs.setReference on the trigger
* Set the refs.setFloating on the floating element (listbox)
* Apply `floatingStyles` to the `refs.floating.current` using CSSOM
* Modify styles as needed by adding a new auto align class and use `:not()`

Gururajj77:
Closes #16470 

Added floating-ui for combobox

#### Changelog

**New**

- Imported `useFloating` hook from `floating-ui` library.
- added new experimental prop, `autoAlign` which toggles the use of the `useFloating` hook.
- a new `useEffect` hook to apply styles to the `Combobox` using `CSSOM`

**Changed**

- added `ref.setReference` to `ListBox` and `ref.setFloating` to `ListBox.Menu`

#### Testing / Reviewing

Check the new Experimental Auto Align story from the Combobox component.


riddhybansal (changes requested):
Do we really need an auto align prop in this case ? Since its a comboBox,  auto align should not be an optional behavior like popover. Auto align should be a default behaviour and should not be changed in conbobox.So implementation of floating UI in this case should not be optional( i.e should not depend on auto align prop)

@tay1orjones What do you think on this ?

Gururajj77:
> Do we really need an auto align prop in this case ? Since its a comboBox, auto align should not be an optional behavior like popover. Auto align should be a default behaviour and should not be changed in combobox. So implementation of floating UI in this case should not be optional( i.e should not depend on auto align prop)
> 
> @tay1orjones What do you think on this ?

We can definitely do this if it's not considered a breaking change

guidari:
> Do we really need an auto align prop in this case ? Since its a comboBox, auto align should not be an optional behavior like popover. Auto align should be a default behaviour and should not be changed in conbobox.So implementation of floating UI in this case should not be optional( i.e should not depend on auto align prop)

Yeah, I agree with Riddhi on this! It is the same approach I'm having on `MultiSelect` and `FilterableMultiselect`, I'm just making the Floating UI alignment as default.

packages/react/src/components/ListBox/test-helpers.js
```diff
@@ -105,3 +106,5 @@ export const generateItems = (amount, generator) =>
     .map((_, i) => generator(i));
 
 export const customItemToString = ({ field }) => field;
+
+export const waitForPosition = () => act(async () => {}); // Flush microtasks. Position state is ready by this line.
```
tay1orjones, line 110: Just want to make a note that this comes from https://floating-ui.com/docs/react#testing and more context is here: https://github.com/floating-ui/react-popper/issues/368#issuecomment-1340413010

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -150,6 +151,11 @@ export interface ComboBoxProps<ItemType>
    */
   ariaLabel?: string;
 
+  /**
+   * Will auto-align the dropdown on first render if it is not visible. This prop is currently experimental and is subject to future changes.
```
tay1orjones: This comment is now outdated (no longer limited to the first render), here's my take at a better one we could use for every component with an `autoAlign` prop:

```suggestion
   * **Experimental**: Will attempt to automatically align the floating
   * element to avoid collisions with the viewport and being clipped by
   * ancestor elements.
```

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -821,6 +857,10 @@ ComboBox.propTypes = {
     PropTypes.string,
     'This prop syntax has been deprecated. Please use the new `aria-label`.'
   ),
+  /**
+   * Will auto-align the dropdown on first render if it is not visible. This prop is currently experimental and is subject to future changes.
```
tay1orjones:
```suggestion
   * **Experimental**: Will attempt to automatically align the floating
   * element to avoid collisions with the viewport and being clipped by
   * ancestor elements.
```

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -342,6 +349,33 @@ const ComboBox = forwardRef(
...
       ...rest
     } = props;
+    const { refs, floatingStyles } = useFloating(
+      autoAlign
+        ? {
+            strategy: 'fixed',
+            middleware: [
+              flip({
+                fallbackAxisSideDirection: 'none',
+              }),
```
tay1orjones: `none` is the default so we can remove it

```suggestion
              flip(),
```

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -342,6 +349,33 @@ const ComboBox = forwardRef(
       slug,
       ...rest
     } = props;
+    const { refs, floatingStyles } = useFloating(
+      autoAlign
+        ? {
+            strategy: 'fixed',
```
tay1orjones, line 358: I think if we include the `placement` config option with the `direction` consumers will still be able to specify it to open `top` by default

```suggestion
            placement: direction,
            strategy: 'fixed',
```

tay1orjones (changes requested):
Looks so good 🔥 just a couple minor things

riddhybansal:
> Do we really need an auto align prop in this case ? Since its a comboBox, auto align should not be an optional behavior like popover. Auto align should be a default behaviour and should not be changed in conbobox.So implementation of floating UI in this case should not be optional( i.e should not depend on auto align prop)
> 
> @tay1orjones What do you think on this ?

This is discussed on slack

riddhybansal:
LGTM !! Just need to run yarn format !!

tay1orjones (approved):
There were some weird unrelated changes highlighted in percy - might just be due to being out of date from latest with main. I just pushed the button to update, if status checks pass I think this LGTM 🎉

alisonjoseph (approved):
LGTM 🔥

guidari (approved):

riddhybansal (approved):
LGTM

Issue #7581, opened 2021-01-15
[Slider] Slider with Increment proposal


fangshih:
I am submitting a ...

- [x]  Feature request
- [ ]  Design defect
- [ ]  Source code defect
- [ ]  Demo/documentation defect
- [x]  Other (design proposal)

### Summary
An additional pattern for Slider

This pattern displays incremental marks in the slider for options available to the user. The display for the slider’s current value is not a text input and therefore cannot be edited by directly clicking on it.

### Overview: 
"Slider with Increment" is useful for users who need to select a specific value or when the values offered are not intuitive or predictable. The incremental marks help them identify at a glance how many options are available to them, while the current value display informs them of their current selection. The value display is not editable to prevent the user from typing in nonexistent values that they may infer are options.

### Screens
![Artboard Copy 3](https://user-images.githubusercontent.com/4862573/104769588-24f74480-5735-11eb-8d8f-cbac37c95aab.png)

1. Labels: Labels inform the user of the slider's purpose
2. Increments: All options available in the slider are shown visually as evenly spaced marks
3. Value: The selected value is displayed in text, along with units if appropriate

### When to use
Use this pattern when:
1. Sliders lack an intuitive range of values, such as values that do not progress linearly (ie incremental values of 4, 16, 32, and 64).
2. Multiple sliders appear next to each other, but offer a radically different amount of options or step values

### Designing with Slider with Increment
This slider largely appears and functions like the basic slider component, with the smallest value on the left and the largest on the right. Users can click and drag the handle to snap to available increments, as delineated in the marks. The mark of a value lower than the current selection assumes the black color of the handle. 

### Behaviors
Similar to the basic slider's disabled state, "Slider with Increment" can be disabled when the user filters it out or otherwise chooses not to view or interact with it.

### Best practices

**Do not -** 

- Do not use the Slider with Increments if there are too many increments. it is recommended sliders do not contain more than 25 increments.

- Do not use in combination with the basic slider on the same page. The inconsistencies when next to each other may cause confusion. Instead, determine which slider component best represents your goals and stick with one.

**Do**

- It is recommended a Slider with Increment is only disabled if the user turns them off or otherwise chooses not to view it. Disabled states are not required to meet color contrast standards, and are therefore not considered readable in the Slider with Increments when used.

- If the slider needs to be disabled, but it is important for the user to be able to know the value she selected, ensure there is an alternate way to view the selected value, such as in text in another place on the screen. 

### Visual guidance
[...]

PR #16616, opened 2024-05-29
feat(slider): expose formatLabel to value tooltip

dkaushik95:
This PR exposes formatLabel to the tooltip for devs to use

- expose formatLabel to thumbWrapper for twoHandles as well as single handle.
- add examples on storybooks on how it could be used

Doesn't close a PR but gets us closer to the comment here: https://github.com/carbon-design-system/carbon/issues/7581#issuecomment-1113437295

#### Changelog

**New**

- `formatLabel` now also works with the Tooltips for the thumbs which gives you more options on how to customize how the value looks.
- Two new storybooks added to describe the change

**Changed**

- Changed the definition of `formatLabel` to represent the above addition

#### Testing / Reviewing

- Ran unit test to make sure nothing is breaking
- You can test it by changing the storybooks:
    - [Slider With Hidden Inputs And Custom Format](http://localhost:3001/?path=/story/components-slider--slider-with-hidden-inputs-and-custom-format)
    - [Slider With Hidden Inputs And Format Label](http://localhost:3001/?path=/story/components-slider--slider-with-hidden-inputs-and-format-label)
    
#### Screenshots

Screenshot 2024-05-29 at 10 13 36 AM
Screenshot 2024-05-29 at 10 13 49 AM
Screenshot 2024-05-29 at 10 13 56 AM

### DIFF
--- a/e2e/components/Slider/Slider-test.avt.e2e.js
+++ b/e2e/components/Slider/Slider-test.avt.e2e.js
@@ -92,6 +92,20 @@ test.describe('@avt Slider', () => {
     await expect(page).toHaveNoACViolations('Slider-with-layer');
   });
 
+  test('@avt-advanced-states slider with custom format', async ({ page }) => {
+    await visitStory(page, {
+      component: 'Slider',
+      id: 'components-slider--slider-with-custom-value-label',
+      globals: {
+        theme: 'white',
+      },
+    });
+
+    await page.keyboard.press('Tab');
+    await expect(page.getByRole('slider')).toHaveValue('Medium');
+    await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
+  });
+
   // Prevent timeout
   test.slow('@avt-keyboard-nav', async ({ page }) => {
     await visitStory(page, {
--- a/e2e/components/Slider/Slider-test.avt.e2e.js
+++ b/e2e/components/Slider/Slider-test.avt.e2e.js
@@ -92,19 +92,45 @@ test.describe('@avt Slider', () => {
     await expect(page).toHaveNoACViolations('Slider-with-layer');
   });
 
-  test('@avt-advanced-states slider with custom format', async ({ page }) => {
-    await visitStory(page, {
-      component: 'Slider',
-      id: 'components-slider--slider-with-custom-value-label',
-      globals: {
-        theme: 'white',
-      },
-    });
+  test.slow(
+    '@avt-advanced-states slider with custom format',
+    async ({ page }) => {
+      await visitStory(page, {
+        component: 'Slider',
+        id: 'components-slider--slider-with-custom-value-label',
+        globals: {
+          theme: 'white',
+        },
+      });
 
-    await page.keyboard.press('Tab');
-    await expect(page.getByRole('slider')).toHaveValue('Medium');
-    await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
-  });
+      // Test for label changes
+      // Initial value
+      await page.keyboard.press('Tab');
+      await expect(page.getByRole('slider')).toBeVisible();
+      await page.keyboard.press('Tab');
+
+      await expect(page.getByRole('slider')).toHaveAttribute(
+        'aria-valuetext',
+        'Medium'
+      );
+      // Move to high
+      await page.keyboard.press('Shift+ArrowRight');
+      await expect(page.getByRole('slider')).toHaveAttribute(
+        'aria-valuetext',
+        'High'
+      );
+
+      // Move to Low
+      await page.keyboard.press('Shift+ArrowLeft');
+      await page.keyboard.press('Shift+ArrowLeft');
+      await expect(page.getByRole('slider')).toHaveAttribute(
+        'aria-valuetext',
+        'Low'
+      );
+
+      await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
+    }
+  );
 
   // Prevent timeout
   test.slow('@avt-keyboard-nav', async ({ page }) => {
--- a/e2e/components/Slider/Slider-test.avt.e2e.js
+++ b/e2e/components/Slider/Slider-test.avt.e2e.js
@@ -92,45 +92,44 @@ test.describe('@avt Slider', () => {
     await expect(page).toHaveNoACViolations('Slider-with-layer');
   });
 
-  test.slow(
-    '@avt-advanced-states slider with custom format',
-    async ({ page }) => {
-      await visitStory(page, {
-        component: 'Slider',
-        id: 'components-slider--slider-with-custom-value-label',
-        globals: {
-          theme: 'white',
-        },
-      });
+  test('@avt-advanced-states slider with custom format', async ({ page }) => {
+    await visitStory(page, {
+      component: 'Slider',
+      id: 'components-slider--slider-with-custom-value-label',
+      globals: {
+        theme: 'white',
+      },
+    });
 
-      // Test for label changes
-      // Initial value
-      await page.keyboard.press('Tab');
-      await expect(page.getByRole('slider')).toBeVisible();
-      await page.keyboard.press('Tab');
+    await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
+  });
 
-      await expect(page.getByRole('slider')).toHaveAttribute(
-        'aria-valuetext',
-        'Medium'
-      );
-      // Move to high
-      await page.keyboard.press('Shift+ArrowRight');
-      await expect(page.getByRole('slider')).toHaveAttribute(
-        'aria-valuetext',
-        'High'
-      );
+  test.slow('@avt-keyboard-nav slider with custom format', async ({ page }) => {
+    // Test for label changes
+    // Initial value
+    await page.keyboard.press('Tab');
+    await expect(page.getByRole('slider')).toBeVisible();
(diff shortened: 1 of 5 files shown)


dkaushik95:

guidari:
Hey @dkaushik95 sorry for the late reply.

This is something would be nice for us to talk with the design team, so we can delivery a consistent solution like it was mentioned here https://github.com/carbon-design-system/carbon/issues/7581#issuecomment-1113437295

I'll make sure to bring that discussion in further meetings with the team.

dkaushik95:
Thanks @guidari , please let me know if there's any other way we can achieve this without this change because we will be needing this functionality soon in our project.

bradelectro:
@guidari This change will enable us to do the migration we need in our product, Instana. Can we please merge?

tay1orjones:
This approach seems good to me. The design considerations on that issue are mostly related to the addition of increment "tics", which will still need a scalable solution for programatic/dynamic labels. I think this provides that pretty well. Seems like it would meet the needs of the increments feature if we do eventually take that on.

A few suggestions to get this closer to merge:

- [ ] Add unit tests covering the new behavior
- [ ] Condense both stories into one single story - I wonder if a name like "Slider With Custom Value Label" would make more sense to someone with no context on what the feature is or the props used?
- [ ] Add that story to AVT and VRT - each of these should first move focus to one of the slider handles (so a tooltip is visible) before running `toBeAccessible` or `snapshotStory`
- [ ] The [`aria-valuetext` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role#:~:text=The%20optional%20aria,aria%2Dvaluenow%20increases) needs to be placed on the input when using this label and the value should be updated just the same as the tooltip. Right now when using this with voiceover, only the numeric value is read while changing the value. You can get the "medium" to be read by pressing up or exiting the slider, but I think with `aria-valuetext` "medium" can be read instead or at least alongside/after the numeric value.


https://github.com/carbon-design-system/carbon/assets/3360588/99e7e96f-d40f-4de1-baa7-538327e4fb6b

bradelectro:
@tay1orjones Thank you for being open to our suggestions. What is AVT / VRT please?

guidari:
> @tay1orjones Thank you for being open to our suggestions. What is AVT / VRT please?

Hey @bradelectro 
- AVT testing is located in this path`e2e/components/Slider/Slider-test.avt.e2e.js`. This test helps us finding any a11y issues and also validate against keyboard navigation.
- VRT testing is located in this path `e2e/components/Slider/Slider-test.e2e.js`. This test can guard us against any visual changes in the story.

I'm happy to help with this PR to get all checks that Taylor had mentioned.

tay1orjones:
I had a conversation with @laurenmrice just now about the design-related concerns of this and we both agreed non-numeric labels are widely used enough (and supported in the a11y spec via `aria-valuetext`) that we should support it. 

If/when the increment/tics are implemented the intent is to allow the input to still be hidden, which means the dynamic tooltip value (the functionality in this PR) will be needed.

dkaushik95:
Thanks @tay1orjones  and @guidari. I’ll complete the tests and fix the a11y issue and let you know once I’m done.

dkaushik95:
Here's what I've done:
- Added `unit` test to cover the new behavior for both variants
- Added `VRT` test using `snapshotStory`
- Added `AVT` test by checking the label after pressing tab and using `toHaveNoACViolations`
- Added the `aria-valuetext` and tested this by using `VoiceOver` on Mac. I was able to get the formatted values on change. 

Screenshot 2024-06-21 at 3 14 44 PM
Screenshot 2024-06-21 at 3 14 23 PM

@guidari , @tay1orjones , can you please help me with verifying that the change works well? Thank you! :D

e2e/components/Slider/Slider-test.avt.e2e.js
```diff
@@ -92,6 +92,20 @@ test.describe('@avt Slider', () => {
...
+      component: 'Slider',
+      id: 'components-slider--slider-with-custom-value-label',
+      globals: {
+        theme: 'white',
+      },
+    });
+
+    await page.keyboard.press('Tab');
+    await expect(page.getByRole('slider')).toHaveValue('Medium');
+    await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
```
guidari: For @avt-advanced-states we just have to keep the line below.
```suggestion
    await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
```

To add keyboard navigation you can use the `@avt-keyboard-nav` and add the tests.

I would suggest the following tests for keyboard navigation:
- Tab to the Slider to verify focused element and it's value "Medium".
- After pressing Shift+ArrowRight the value should be "High"
- After pressing two times Shift+ArrowLeft the value should be "Low"

 `await page.keyboard.press('Shift+ArrowRight');`

packages/react/src/components/Slider/Slider.stories.js
```diff
@@ -53,6 +53,41 @@ export const SliderWithHiddenInputs = () => (
...
+  />
+);
+
+export const SliderWithCustomValueLabel = () => (
+  <Slider
+    labelText="Slider label with low/medium/high"
+    value={50}
+    min={0}
+    max={100}
+    stepMultiplier={10}
```
guidari: We can change here to 50 so when the user press Shift + ArrowKey it increases to the next value.
```suggestion
    stepMultiplier={50}
```

packages/react/src/components/Slider/Slider.stories.js
```diff
@@ -53,6 +53,41 @@ export const SliderWithHiddenInputs = () => (
...
+    step={1}
+    noValidate
+    hideTextInput
+    formatLabel={(val) => `${val}%`}
+  />
+);
```
guidari:
```suggestion
```
I'm not sure if we should keep this story. Maybe we can show in the `fomartLabel` [description here](https://github.com/carbon-design-system/carbon/blob/a4f9f5584b1ae1bd043d27bbf32a84ff73750631/packages/react/src/components/Slider/Slider.tsx#L344-L347) how to set the `formatLabel` like that: 
```js
formatLabel={(val) => `${val}%`} 
```
We already have the `SliderWithCustomValueLabel` as an example also.

guidari (changes requested):
Hey!! The changes are looking good! Just a couple notes

e2e/components/Slider/Slider-test.avt.e2e.js
dkaushik95: > Add that story to AVT and VRT - each of these should first move focus to one of the slider handles (so a tooltip is visible) before running toBeAccessible or snapshotStory

I assumed we have to move focus before we run tests. I'll change that. Thank you :)

e2e/components/Slider/Slider-test.avt.e2e.js
```diff
@@ -92,6 +92,46 @@ test.describe('@avt Slider', () => {
     await expect(page).toHaveNoACViolations('Slider-with-layer');
   });
 
+  test.slow(
+    '@avt-advanced-states slider with custom format',
```
guidari: Hey @dkaushik95 
The `@avt-advanced-states` and `@avt-keyboard-nav` should be separate tests.

In the  `@avt-advanced-states` for the new stories, just keep this line:
```js
      await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
```

For the `@avt-keyboard-nav` you can keep only he keyboard navigation you are testing.

Here it is an example of the difference between the two tests:

Screenshot 2024-06-27 at 14 08 28

guidari (changes requested):

e2e/components/Slider/Slider-test.avt.e2e.js
dkaushik95: Got it! Thanks for the info! I'm new to the structure :D I'll spin out a commit soon

e2e/components/Slider/Slider-test.avt.e2e.js
```diff
@@ -92,6 +92,45 @@ test.describe('@avt Slider', () => {
...
+      id: 'components-slider--slider-with-custom-value-label',
+      globals: {
+        theme: 'white',
+      },
+    });
+
+    await expect(page).toHaveNoACViolations('Slider-with-custom-value-label');
+  });
+
+  test.slow('@avt-keyboard-nav slider with custom format', async ({ page }) => {
```
guidari, line 107: It is missing the visitStory here
```suggestion
  test.slow('@avt-keyboard-nav slider with custom format', async ({ page }) => {
  await visitStory(page, {
      component: 'Slider',
      id: 'components-slider--slider-with-custom-value-label',
      globals: {
        theme: 'white',
      },
    });
```

guidari (changes requested):

e2e/components/Slider/Slider-test.avt.e2e.js
dkaushik95: Should be fixed now

guidari (approved):
LGTM! @dkaushik95 
Thanks for the contribution!!

alisonjoseph (approved):

kingtraceyj (approved):
looks good to go visually. i'll open an issue to address usage guidelines. i think we have some open questions about when to change labels, how many tick marks are allowed, breakpoints, etc.

tay1orjones (approved):
Thank you again for this 🎉

PR #16644, opened 2024-06-03
fix: interactive tile expand button

lee-chase:
Interactive tile expand button ignores spacebar press.

#### Changelog

InteractiveTile expand/close

#### Testing / Reviewing

Buttons should be operable via the spacebar. In the existing code the expand/collapse of the interactive tile has this disabled to cope with text box usage. 

This PR processes the keypress to check that it is targeting the expand button.


alisonjoseph (approved):
This LGTM, should we add a test around this?

guidari (approved):
LGTM! 🚀

To add some tests you can use this as an example. 
- Add a new `Tab`;
- Change the selector;
- Change from `Enter` to `Space`

https://github.com/carbon-design-system/carbon/blob/6bd58e650e51b2d3e1b3f2f1607c8a99bef10680/e2e/components/Tile/Tile-test.avt.e2e.js#L112-L127

tay1orjones:
@lee-chase would you be up for adding tests to this?

tay1orjones (approved):
I've added the necessary tests and corrected some that were failing

PR #16647, opened 2024-06-03
fix: fixed VO when typing a value

### LINKED ISSUE #14707: [ally]: Combobox Default & with Layer - VO doesn't announce keyboard input
## Environment

> Operating system

> Browser

## Detailed description

If the VO can provide the feedback to the user about the availability of the keyboard and instantaneously announce. the modification that happens upon providing the input, it will be helpful

When the screen reader doesn’t identify the page elements, it impacts the experience of users who are blind or have vision impairment. They can miss important information or fail to perform an action on the web page. 

NOTE: The same issue is observed even in the aria combobox example that is given below.
https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/

> What happened instead?
In voice over on iPhone, after double tapping the "edit combo" the keyboard is displayed visually. However, the same is not announced by screen reader. Hence the user may not realize that the input can also be typed.

Default: https://react.carbondesignsystem.com/?path=/story/components-combobox--default
With layer: https://react.carbondesignsystem.com/?path=/story/components-combobox--with-layer

## Additional information

- Screenshots or code
- Notes

guidari:
Closes #14707

The VO was not reading the input if the valud was not present in the list box.

By removing the `aria-owns` and  `aria-controls` only if the menu is open, the VO can read the input typed.
`aria-owns` is not being used in the [downshift example](https://www.downshift-js.com/use-combobox/) either, so I guess it is not a problem to remove.

#### Testing / Reviewing

- Open ComboBox component with VO
- Type something in the input and wait to read


Gururajj77 (approved):
Looks Good!

alisonjoseph (approved):

PR #16186, opened 2024-04-12
fix(16059): updates RadioButton and RadioButtonGroup proptypes

2nikhiltom:
Closes #16059

props `type` updated for `RadioButton` and `RadioButtonGroup`

#### Changelog
event type updated : `React.ChangeEvent<HTMLInputElement>`
 ```
onChange?: (
    selection: string | number,
    name: string,
    event: React.ChangeEvent<HTMLInputElement>
  ) => void;
```

**Changed**
`event` type changed from `any` to `React.ChangeEvent<HTMLInputElement>`
`selection` type changed from `unknown` to  `string | number`

#### Testing / Reviewing
Use `RadioButtonGroup`, `RadioButton` with `onChange` props in typescript environment  and verify that the TypeScript compiler should not throw error

   ```
 <RadioButtonGroup
      name="test16059"
      onChange={(value: string | number) => console.log(value)}
      valueSelected={1}
    >
      <RadioButton labelText="test1" value={22} />
      <RadioButton labelText="carbon" value={22} />
    </RadioButtonGroup>
```

PR #16648, opened 2024-06-03
feat: improve `RadioButton` and `RadioButtonGroup` types

cuppajoey:
In #16186, the `selection` argument of `RadioButtonGroup`'s `onChange` prop was changed to `ReactNode` instead of `string | number` as written in the PR description. In addition to fixing that I have added some missing types and improved others.

#### Changelog

- Export prop interfaces for `RadioButton` and `RadioButtonGroup`.
- Narrow `onChange` argument type for `RadioButtonGroupProps`.
- Reference `RadioButtonProps` for `RadioButtonGroup` types where we are expecting a `RadioButton` value.
- Add missing types for component `refs` and event handlers in `RadioButton` and `RadioButtonGroup`.
- Simplify `getRadioButtons()` with a type assertion on `children` argument and refactoring to use an early return statement.

#### Testing / Reviewing

- [ ] Check `RadioButton` and `RadioButtonGroup` for type errors
- [ ] Check for regressions with `RadioButtons` used inside of a group (e.g., radio buttons are still correctly displayed and the group value changes when radio selections change.)

### DIFF
--- a/packages/react/src/components/RadioButton/RadioButton.tsx
+++ b/packages/react/src/components/RadioButton/RadioButton.tsx
@@ -72,8 +72,8 @@ export interface RadioButtonProps
    * the underlying `<input>` changes
    */
   onChange?: (
-    value: string | number,
-    name: string | undefined,
+    value: RadioButtonProps['value'],
+    name: RadioButtonProps['name'],
     event: React.ChangeEvent<HTMLInputElement>
   ) => void;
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/RadioButton/RadioButton.tsx
```diff
@@ -98,80 +98,85 @@ export interface RadioButtonProps
   required?: boolean;
 }
 
-const RadioButton = React.forwardRef((props: RadioButtonProps, ref) => {
```
cuppajoey, line 101: `ref` is now correctly typed.

Most of the following diff below is due to prettier's formatting adjusting the indentation. I recommend viewing with ⚙️ whitespace changes hidden.

packages/react/src/components/RadioButton/RadioButton.tsx
```diff
@@ -72,8 +72,8 @@ export interface RadioButtonProps
    * the underlying `<input>` changes
    */
   onChange?: (
-    value: string | number,
-    name: string | undefined,
+    value: RadioButtonProps['value'],
+    name: RadioButtonProps['name'],
```
cuppajoey, line 76: We can directly reference other types within the same interface when the expected value is the same (e.g., `props.name` and `props.value` are what we're passing to the `onChange` handler). 

Referencing types this way provides more context and reduces the work later, should `RadioButton` types change in the future.

This is also how it was originally typed (See [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/carbon-components-react/lib/components/RadioButton/RadioButton.d.ts#L20))

packages/react/src/components/RadioButton/RadioButton.tsx
```diff
@@ -98,80 +98,85 @@ export interface RadioButtonProps
...
     }
-  );
 
-  const inputRef = useRef<HTMLInputElement>(null);
+    const innerLabelClasses = classNames(
+      `${prefix}--radio-button__label-text`,
+      {
+        [`${prefix}--visually-hidden`]: hideLabel,
+      }
+    );
```
cuppajoey, line 131: This was an automatic prettier adjustment.

packages/react/src/components/RadioButton/RadioButton.tsx
```diff
@@ -98,80 +98,85 @@ export interface RadioButtonProps
...
+      slug,
+      required,
+      ...rest
+    } = props;
+
+    const prefix = usePrefix();
+    const uid = useId('radio-button');
+    const uniqueId = id || uid;
+
+    function handleOnChange(event: React.ChangeEvent<HTMLInputElement>) {
```
cuppajoey, line 122: As a bonus, I added some missing types while working in these files.

packages/react/src/components/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -87,10 +88,11 @@ export interface RadioButtonGroupProps
    * the group changes
    */
   onChange?: (
-    selection: React.ReactNode,
```
cuppajoey, line 90: The `selection` will be the same as the `RadioButton` value. `ReactNode` is too broad.

packages/react/src/components/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -166,30 +169,38 @@ const RadioButtonGroup = React.forwardRef(
...
-          value: value,
-          onChange: handleOnChange,
-          checked: value === selected,
-          required: required,
-        };
-
-        if (!selected && (radioButton as ReactElement)?.props.checked) {
-          newProps.checked = true;
+      const mappedChildren = React.Children.map(
+        children as ReactElement<RadioButtonProps>,
```
cuppajoey, line 173: `children` is now typed as a `RadioButton` allowing us to inherit the types from `RadioButtonProps`. Previously, all prop values referenced in this function were typed as `any`.

This way, we no longer need to keep asserting the type of `radioButton`.

packages/react/src/components/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -166,30 +169,38 @@ const RadioButtonGroup = React.forwardRef(
...
+        (radioButton) => {
+          if (!radioButton) {
+            return;
+          }
```
cuppajoey, line 177: The null-check from 🔴 184 was moved up here as an early return.

cuppajoey (commented):
As a review aid, I left a few comments for additional context.

guidari (commented):
LGTM!  🚀

guidari (approved):

packages/react/src/components/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -230,7 +241,7 @@ const RadioButtonGroup = React.forwardRef(
     const divRef = useRef<HTMLDivElement>(null);
 
     // Slug is always size `mini`
-    let normalizedSlug;
+    let normalizedSlug: ReactElement | undefined;
```
2nikhiltom, line 244: This can be done across all components for Type Safety and Clarity

2nikhiltom (approved):
Looks good to me 🔥

alisonjoseph (approved):

PR #16653, opened 2024-06-04
fix: closing of calendar without selecting end date

### LINKED ISSUE #16473: [Bug]: DatePicker Range has unintended behavior when selecting dates
### Package

@carbon/react

### Browser

Chrome

### Package version

1.57.0

### React version

16.14.0

### Description

This bug is in 1.57.0, but not in any previous versions. (currently downgraded to >=1.56.0 and bug was not present)
When selecting first date in date range, calendar closes instead of staying open for user to pick second date in date range.
When selecting second date range, the first date range contains BOTH first and second dates in selection. 
We're using calendar as controlled component, with onChange functions and value properties. Again, this has worked until the recent update in 1.57.0. 

### Reproduction/example

https://stackblitz.com/edit/github-sf1rjr?file=src%2FApp.jsx,src%2Fmain.jsx

### Steps to reproduce

-click start date, select date (note that calendar closes, this is not correct calendar function)
-select end date
-user can see that start date contains string "05/13/2024 to 05/14/2024", which is not correct calendar function

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #16473 

Fixed closing of calendar without selecting end date

#### Changelog

**Removed**

- Removed fp.close() from the rangePlugin as it was not doing anything.

#### Testing / Reviewing

See if the calendar in range mode does not close without selecting end date.


Gururajj77 (approved):
Looks good!

guidari (approved):

PR #16139, opened 2024-04-05
fix(multiselect): add useEffect to selection hook to update externally

ychavoya:
Closes #16124

Update MultiSelect's `useSelection` hook to include a `useEffect` that updates the selected items when they are updated externally (controlled component), since the recently added `useState` is only executed on initial render.

#### Changelog

**Changed**

- Ability to update a controlled MultiSelect's selected items externally

#### Testing / Reviewing

Added a [Storybook](http://localhost:3000/?path=/story/components-multiselect--controlled) to test controlled MultiSelect components. The MultiSelect in this storybook should update its selected values when the included buttons are pressed, and a storybook action should be triggered.

Also added a unit test that tests this behavior.

PR #16689, opened 2024-06-04
Added Floating UI to `MultiSelect` and `FilterableMultiSelect`

### LINKED ISSUE #16507: Add floating-ui to `@carbon/react` MultiSelect and FilterableMultiSelect

guidari:
Closes #16507

Added Floating UI to `MultiSelect` and `FilterableMultiSelect`. To work the test we had to add the `act` to wait for the floating to position.

#### Testing / Reviewing

- Test the Default and Filterable stories
________

- [ ] Remove `div` from Default and Filterable stories.


riddhybansal (approved):
Awesome !!! Floating UI

packages/react/src/components/MultiSelect/MultiSelect.stories.js
```diff
@@ -226,16 +226,24 @@ Playground.argTypes = {
...
+        <MultiSelect
+          label="Multiselect Label"
+          id="carbon-multiselect-example"
+          titleText="Multiselect title"
+          helperText="This is helper text"
+          items={items}
+          itemToString={(item) => (item ? item.text : '')}
+          selectionFeedback="top-after-reopen"
+        />
+      </div>
```
tay1orjones: I'd be in favor of updating all the stories to follow this format if you think it's a good idea. Would be nice to be able to set `direction` on the playground and see how it changes the auto align behavior.

On the other hand I'm not sure if this would be too distracting to have on every story. What do you think?

Either way we should call `scrollIntoView` so the component is immediately visible on render. I think there's an example of it in the popover autoAlign stories

tay1orjones (commented):
The stories need updated but overall looks good! 🔥  

Do you think we need to put this behind an `autoAlign` prop like the others? Why or why not?

packages/react/src/components/MultiSelect/MultiSelect.stories.js
guidari: - Added `autoAlign` I forgot we are not exposing as default the FloatingUI for now.
- Added new stories called Experimnetal Auto Align
- Added changes to Playground

I don't think we have to copy the format to every story in this point. We can copy in every story once we have it as default.

tay1orjones (approved):
A couple status checks to fix but LGTM!

guidari:
I'm removing the test that it is failing.
The test was added in this [PR](https://github.com/carbon-design-system/carbon/pull/16139) but it was removed 3 days ago in this [PR](https://github.com/carbon-design-system/carbon/pull/16793).
I'm not sure why wasn't removed when I merged with main 🤔

PR #16701, opened 2024-06-05
feat(vertical-tabs): new tab variant

### LINKED ISSUE #16305: Vertical Tab variant
### The problem

PLG getting started pattern needs a stacked tab variant

### The solution

[Specs ](https://www.figma.com/file/qGVzyBMsr4nR7clX6rnZOe/Tabs?type=design&node-id=183%3A369443&mode=design&t=itD492w9y2QQFhxa-1)so far (about 90% done)

**Need to confirm:**
- Indicator size
- Overflow gradient (Should be similar to scrollable modal or scrollable tabs)
- Footer/Content area (should it be sticky and shouls we allow anything other than text and link?)

Vertical tab specs

### Examples

_No response_

### Application/PAL

PLG

### Business priority

### Available extra resources

_No response_

### Code of Conduct

ariellalgilmore:
#16305 

Adds a new vertical variant to tabs
Screenshot 2024-06-05 at 10 54 44 AM

#### Changelog

**New**

- TabListVertical component
- TabPanels new update to check if vertical variant and set the same max height to each TabPanel
- Tab new update to add ellipsis with Tooltip only on vertical variant

**Changed**

- added `overflow-x: hidden` to default Tabs ... fixed an issue where an unwanted horizontal scroll would appear if the length of the tabs is too long

-----
**Other notes:**

- using `float: inline-start` to align the vertical tabs. This avoid wrapping the component in `<Grid>`, which causes complexities when going between `sm` and other breakpoints, but I know `float` isn't really used any more
- for TabPanels, I have to unhide/hide each Panel to get the current height and then use `style.height = ...` for each TabPanel  to set what the height of the largest height for each. I am not noticing any flashing, but could be a concern
- keeping as a draft since waiting for confirmed overflow styles and if the `sm` breakpoint should be default tabs or contained tabs
- still need to update docs too

#### Testing / Reviewing

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.tsx
+++ b/packages/react/src/components/Tabs/Tabs.tsx
@@ -12,6 +12,7 @@ import debounce from 'lodash.debounce';
 import PropTypes from 'prop-types';
 import React, {
   useCallback,
+  useLayoutEffect,
   useState,
   useRef,
   useEffect,
@@ -76,6 +77,7 @@ const TabContext = React.createContext<{
 });
 
 const lgMediaQuery = `(min-width: ${breakpoints.lg.width})`;
+const smMediaQuery = `(max-width: ${breakpoints.md.width})`;
 
 // Used to keep track of position in a list of tab panels
 const TabPanelContext = React.createContext<number>(0);
@@ -213,6 +215,33 @@ function getNextIndex(
       return (index + 1) % total;
 
     case match(event, keys.ArrowLeft):
+      return (total + index - 1) % total;
+
+    case match(event, keys.Home):
+      return 0;
+
+    case match(event, keys.End):
+      return total - 1;
+
+    default:
+      return index;
+  }
+}
+
+/**
+ * Get the next index for a given keyboard event
+ * given a count of the total items and the current index
+ */
+function getNextIndexVertical(
+  event: SyntheticEvent,
+  total: number,
+  index: number
+): number {
+  switch (true) {
+    case match(event, keys.ArrowDown):
+      return (index + 1) % total;
+
+    case match(event, keys.ArrowUp):
       return (total + index - 1) % total;
 
     case match(event, keys.Home):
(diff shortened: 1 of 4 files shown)


alisonjoseph:
This is looking good! My one question, which may be more for design, is around the 25% width. I think this might work as a default, but I can see scenarios where someone might need to align/wrap this on the grid. Perhaps it needs a width prop where you can set a width, or leave it as 100% so the user can place it within their own container? 

We solved a somewhat similar problem for grid-aligned contained tabs with a combination of grid components and a fullwidth prop. https://react.carbondesignsystem.com/?path=/docs/components-tabs--overview#tabs-and-the-grid---fullwidth-prop

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -1246,12 +1551,58 @@ export interface TabPanelsProps {
...
+      // Should only apply same height to vertical Tab Panels
+      if (tabContainer?.classList.contains(`${prefix}--tabs--vertical`)) {
+        hiddenStates.current = refs.current.map((ref) => ref?.hidden || false);
+
+        // un-hide hidden Tab Panels to get heights
+        refs.current.forEach((ref) => {
+          if (ref) {
+            ref.hidden = false;
+          }
+        });
```
tay1orjones, line 1571: If I understand the intent here, I think you could avoid having to unhide them like this by using `useLayoutEffect`. It's [specifically geared towards](https://react.dev/reference/react/useLayoutEffect#measuring-layout-before-the-browser-repaints-the-screen) logic that needs to measure the dom. 

OR maybe this justifies having a wrapper component around the tabs and the tabpanels. That way we wouldn't have to measure anything and it could be done with CSS. Honestly I'd be up for that approach over measuring with js to match heights. Are there any downsides of a wrapper element that you can think of?

tay1orjones (commented):
Some `scrollIntoView` related observations/questions:

<details>
  <summary>It's possible to select a tab and not be able to see it. The focus ring is obscured by the gradient which might be an accessibility failure?</summary>
  
  https://github.com/carbon-design-system/carbon/assets/3360588/c2788d04-550b-4360-88c7-4b5bda0d87ed
</details>


<details>
  <summary>When a tab is selected and the browser size changes the tab isn't visible in some cases - both in the vertical config and the smaller horizontal config</summary>
  
  https://github.com/carbon-design-system/carbon/assets/3360588/ea700b2f-90c1-4074-b1a8-14627ab29cb9
</details>

packages/react/src/components/Tabs/Tabs.tsx
ariellalgilmore: I thought the `useLayoutEffect` would work here as well! 

When trying this:
```
useLayoutEffect(() => {
      refs.current.forEach((ref) => {
        console.log(ref?.offsetHeight);
      });
});
```
Screenshot 2024-06-07 at 9 53 47 AM
^ i would only get the following printed... the only visible height 48 and the rest 0


The only side effect of a wrapper element that I can see is possibly a bit of a structure change, but there would be a lot of benefits layout wise for vertical tabs. This could also solve the issue @alisonjoseph brought up where I am using percentages.

ariellalgilmore: just confirming the above that adding a wrapper with styles such as:

```
.#{$prefix}--tabs-wrapper {
    display: flex;
  }
```

then would remove the need for using javascript and css would automatically handle the heights. If both the default and vertical tabs have this wrapper though then there would need to be some javascript to handle when to apply `display: flex` or not to handle the different layouts for vertical and default

kingtraceyj:
@alisonjoseph the design intent is for the tablist and tab panel to be [grid-aligned](https://www.figma.com/design/qGVzyBMsr4nR7clX6rnZOe/Tabs?node-id=394-80822&t=5OMnNEEcQGDGD433-1)
-tablist 4 columns at max, x-large, and large breakpoints
-tablist 2 columns at Medium breakpoint
-tablist turns into contained tabs at small breakpoint

@tay1orjones i think that you are right in that is is (or will be) an accessibility violation if [focus is obscured](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum) so we must keep the focused/active tab in view.

ariellalgilmore:
closing this PR in favor for: #16738

PR #16742, opened 2024-06-12
fix(16740): End Date calendar in range variant remains open on Click away

2nikhiltom:
Closes #https://github.com/carbon-design-system/carbon/issues/16740

This PR fixes the End Date calendar in range variant remains open on Click away 

#### Changelog
```
 // overriding the flatpickr bug where the startDate gets deleted on blur
if (inputTo === target && !inputFrom.value && lastStartValue.current) {
      if (isValidDate(parseDateWithFormat(lastStartValue.current))) {
        inputFrom.value = lastStartValue.current;
        if (inputTo.value) {
          fp.setDate(
            [inputFrom.value, inputTo.value],
            false,
            fp.config.dateFormat
          );
          fp.updateValue();
        }
      }
    }
```

For [setDate](https://flatpickr.js.org/instance-methods-properties-elements/#setdatedate-triggerchange-datestrformat) function :  `false`  is passed as second argument to avoid forcing any onChange .

`updateValue` is invoked to maintain the selected range (blur color on selected range) when the calendar is opened again.

https://github.com/carbon-design-system/carbon/assets/63502271/e4ff095f-b203-47f6-a163-c80a45575924

#### Testing / Reviewing
1. Open datepicker with a range
2. Select start and end date
3. Open end date calendar, do not select any date and click outside of calendar and verify if calendar closes
4. Made sure the selected range is visible when the  calendar is opned again 
5. Verify existing functionality is intact

Issue #16740, opened 2024-06-12
[Bug] [DatePicker]: End Date Calendar in Range variant remains Open on click away


2nikhiltom:
### Package

@carbon/react

### Package version

v1.58.0

### React version

18.02

### Description

When using the `datepicker` with a `range` selection, the `end date` calendar does not close when the user clicks away, causing it to always stay open. 

In contrast, the `start date` calendar closes correctly when the user clicks away. 

https://github.com/carbon-design-system/carbon/assets/63502271/4d34a330-1ea1-4bec-9f9c-9c601243d14c

Surfaced while debugging : https://github.com/carbon-design-system/carbon/issues/16337__

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datepicker--range-with-calendar

### Steps to reproduce

1.`datepicker` with a `range`
2. Select start and end date
3. Open end date calendar, do not select any date and click outside of calendar (Notice that calendar does not close)  

### Suggested Severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

alisonjoseph:
This bug was introduced in 1.58

PR #16758, opened 2024-06-12
fix(16337): close DatePicker calendar on away click when inside Modal or ComposedModal

### LINKED ISSUE #16337: [Bug]: <DatePicker> hosted inside of <Modal> or <ComposedModal> does not close calendar when clicking outside
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

v1.56.0

### React version

v18

### Description

When a <DatePicker> is hosted inside of either <Modal> or <ComposedModal>, click on the <DateInput> and the calendar popover can not be dismissed by clicking outside of the calendar.

https://github.com/carbon-design-system/carbon/assets/38996291/6623e372-5dfe-441c-a2c4-028c886c1848

### Reproduction/example

https://stackblitz.com/edit/github-datbmn?file=src%2FApp.jsx

### Steps to reproduce

1. The example already has the modal opened by default. Click on the date input, and no matter where you click inside the `<ComposedModal>` and outside of the calendar, the calendar just does not close. It can only be closed by tab away or select the date.
1. Close the modal by clicking the close button.
2. There is another `<DatePicker` outside of `<ComposedModal>`. It has no problem closes its calendar popover by clicking elsewhere in the page.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/16337

The root bug with  DatePicker was : When using the datepicker with a range selection, the end date calendar does not close when the user clicks away, causing it to always stay open ([Issue](https://github.com/carbon-design-system/carbon/issues/16740) -> [PR](https://github.com/carbon-design-system/carbon/pull/16742))

Fix for https://github.com/carbon-design-system/carbon/issues/16337 is dependent on  [PR](https://github.com/carbon-design-system/carbon/pull/16742)
Also this issue was happing because `evt.stopPropagation()` was added on `Mousedown` event to add support for [Nest modal](https://github.com/carbon-design-system/carbon/pull/14320)

To fix https://github.com/carbon-design-system/carbon/issues/16337  : I thought to keep this action behind a prop 

- Introduced a new prop `isNested`, in modal.tsx.
- The event handler will now run `evt.stopPropagation()` only when the `isNested` prop is present.

image

- Additionally, we can update the documentation to include guidelines advising against the use of DatePickers within nested modals.

I am looking for reviews and suggestions on this.
If `isNested` prop works as a fix , I will be implementing this is in `ComposedModal` as well 

#### Testing / Reviewing

- Take pull in local
- Open Modal storybook->default 
- Select start and end date in the datepicker 
- Open end date calendar, do not select any date and click outside of calendar and verify if calendar closes 

Verify the functionlity when `isNested`  is passed 
- Pass `isNested` as a prop in `Modal`
- Select start and end date in the datepicker 
- 
image

-  Open end date calendar, do not select any date and click outside of calendar and verify that does not close

### DIFF
--- a/packages/react/src/components/DatePicker/plugins/fixEventsPlugin.js
+++ b/packages/react/src/components/DatePicker/plugins/fixEventsPlugin.js
@@ -109,9 +109,10 @@ export default (config) => (fp) => {
         if (inputTo.value) {
           fp.setDate(
             [inputFrom.value, inputTo.value],
-            true,
+            false,
             fp.config.dateFormat
           );
+          fp.updateValue();
         }
       }
     }
--- a/packages/react/src/components/Modal/Modal.stories.js
+++ b/packages/react/src/components/Modal/Modal.stories.js
@@ -16,6 +16,8 @@ import Dropdown from '../Dropdown';
 import SelectItem from '../SelectItem';
 import TextInput from '../TextInput';
 import mdx from './Modal.mdx';
+import DatePicker from '../DatePicker';
+import DatePickerInput from '../DatePickerInput';
 import {
   StructuredListWrapper,
   StructuredListHead,
@@ -45,9 +47,10 @@ export const Default = () => {
         modalHeading="Add a custom domain"
         modalLabel="Account resources"
         primaryButtonText="Add"
+        // isNested
         secondaryButtonText="Cancel">
         <p style={{ marginBottom: '1rem' }}>
-          Custom domains direct requests for your apps in this Cloud Foundry
+          nikhil domains direct requests for your apps in this Cloud Foundry
           organization to a URL that you own. A custom domain can be a shared
           domain, a shared subdomain, or a shared domain and host.
         </p>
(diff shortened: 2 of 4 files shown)


packages/react/src/components/DatePicker/plugins/fixEventsPlugin.js
2nikhiltom, line 1: The changes in fixEventsPlugin.js will be made available from https://github.com/carbon-design-system/carbon/pull/16742

packages/react/src/components/Modal/Modal.stories.js
```diff
@@ -45,6 +47,7 @@ export const Default = () => {
         modalHeading="Add a custom domain"
         modalLabel="Account resources"
         primaryButtonText="Add"
+        //isNested
```
riddhybansal, line 50: I think this comment can be removed!!

packages/react/src/components/DatePicker/plugins/fixEventsPlugin.js
```diff
@@ -109,9 +109,10 @@ export default (config) => (fp) => {
         if (inputTo.value) {
           fp.setDate(
             [inputFrom.value, inputTo.value],
-            true,
+            false,
```
riddhybansal, line 112: Does previously sending the second value as true , was forcefully firing onChange event ?

riddhybansal (commented):
This looks like a great hack @2nikhiltom . Great work I must say !!

PR #16786, opened 2024-06-14
feat: Add floating ui to new overflow menu

### LINKED ISSUE #16541: Add floating-ui to `@carbon/react` new OverflowMenu (behind the `enable-v12-overflowmenu` flag)

preetibansalui:
Closes #16541 

Add floating ui to new overflow menu(behind the enable-v12-overflowmenu flag)


packages/react/src/components/OverflowMenu/OverflowMenu.featureflag.stories.js
```diff
@@ -62,6 +62,66 @@ export const _OverflowMenu = () => {
...
+  const onClick = action('onClick (MenuItem)');
+
+  return (
+    <>
+      <div
+        style={{
+          marginTop: '200px',
+          display: 'flex',
+          justifyContent: 'space-between',
+        }}>
```
tay1orjones: Could this use the setup used in the popover autoalign stories instead? Right now I can't test to see what happens when scrolling side to side.

The popover autoalign stories center the element in a large box and uses `scrollIntoView` to ensure it's visible when the story is rendered.

tay1orjones (changes requested):
The alignment options need to reflect [the spec](https://carbondesignsystem.com/components/menu-buttons/usage#alignment) and be restricted so that the floating element doea not appear to the left or right of the trigger element, only above and below. The `start` and `end` options for `top` and `bottom` are valid too I believe. 

These two are invalid though, for instance:

| | |
|--------|--------|
|  |  |

preetibansalui:
> The alignment options need to reflect [the spec](https://carbondesignsystem.com/components/menu-buttons/usage#alignment) and be restricted so that the floating element doea not appear to the left or right of the trigger element, only above and below. The `start` and `end` options for `top` and `bottom` are valid too I believe.
> 
> These two are invalid though, for instance:
> 
> 	

@tay1orjones : Thanks for the suggestions, I have added the fallbackPlacements, please review.

tay1orjones (approved):
🎉

riddhybansal (approved):
This looks so good !!

PR #16788, opened 2024-06-14
feat(UIShell): pass-through props on <HeaderContainer />

### LINKED ISSUE #15975: [Feature Request]: HeaderContainer should either treat its render children as a function or expose a props pass-through
### The problem

The UI Shell `<HeaderContainer />` component expects `render` to be a React component. Using the render props pattern (shown in the official storybook example [here](https://react.carbondesignsystem.com/?path=/story/components-ui-shell-header--header-w-side-nav&globals=theme:white)) will trigger un/re-mounting, [which also remounts all children, dumping all state](https://legacy.reactjs.org/docs/reconciliation.html#elements-of-different-types). 

An effective solution might be to hoist state above `<HeaderContainer />`, however this promotes poor organization, and does not handle for all issues (e.g. child `img` sources are re-fetched causing visual jitter). A user could also provide a React component for `render`, with the limitation that it be severed from accessing any local parent state or depend on any external prop, which would only leave room for trivial use cases, or require convoluted workarounds.  
  

### The solution

We could treat `render` as a function and not a component, however it would represent a non-backwards compatible change which would require a major version upgrade according to semver. 

A simple, backwards compatible solution would be to expose a props pass-through object in `HeaderContainer` which is spread onto the child component. 

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

lluisrojass:
Closes #15975

#### Changelog

**New**

- Adds a regression test 
- Changes documentation on the `render` `propType`

**Changed**

- Allowing pass-through props on `<HeaderContainer />`'s  `render`

#### Testing / Reviewing

- The unit test verifies new functionality 
- TS Users can test by consuming `HeaderContainer` using a `render` component w/ new props & verifying that `<HeaderContainer />` type checks via prop inference.
- TS users can also regression test by verifying the `render` function approach type-checks properly


lluisrojass:

lluisrojass:
recheck

lluisrojass:
recheck

lluisrojass:
recheck

lluisrojass:
Unsure why ci checks are failing, re-opening this PR

PR #9816, opened 2021-10-06
fix(OverflowMenu): add stopPropagation call to handleClick method

rzhekova:
Closes #9744 

To fix the issue I added an [event.stopPropagation()](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation) call to the handleClick method in OverflowMenu which stops it from happening. The event method appears to prevent propagation and bubbling up of events which is probably what happens with the click in OverflowMenu. 

I have also added a unit test.

Thanks!

PR #16792, opened 2024-06-17
fix: popover closing issue when clicked on overflow menu

### LINKED ISSUE #16357: [Bug]: Popover doesn't close when OverflowMenu is Clciked
### Package

@carbon/react

### Browser

_No response_

### Package version

v11.56.0

### React version

18.02

### Description

Popover doesn't close when an OverflowMenu is clicked. 

The issue might be here:
https://github.com/carbon-design-system/carbon/blob/afce28df13ae3475eb32c7b65e0599eb86775458/packages/react/src/components/Popover/index.tsx#L169-L173

console.log the `event.target` to see the error

### Reproduction/example

https://stackblitz.com/edit/github-aaf3yh?file=src%2FApp.jsx

### Steps to reproduce

Open the Popover then click on the OverflowMenu.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #16357 

Popover doesn't close when OverflowMenu is Clicked. The issue is caused after this [PR](https://github.com/carbon-design-system/carbon/pull/9816) . event.stopPropagation was added to fix calling of handleOnClick twice. The issue doesnot persist now even after removing event.stopPropagation. To ensure the previous issue is not encountered after this PR, test cases has been added.

#### Changelog

**Removed**

- event.stopPropagation() from handleOnClick.

#### Testing / Reviewing

Go to storybook->popover->test popover should get closed after clicking on overflowMenu or anywhere else.
Test the functioning of overflow menu stories also , it should work as intended and click event should not be called twice.


packages/react/src/components/OverflowMenu/OverflowMenu-test.js
```diff
@@ -171,5 +171,29 @@ describe('OverflowMenu', () => {
...
+          <OverflowMenuItem className="test-child" itemText="one" />
+          <OverflowMenuItem className="test-child" itemText="two" />
+        </OverflowMenu>
+      );
+
+      // Find the OverflowMenu button
+      const button = screen.getByRole('button');
+
+      // Click the OverflowMenu button
+      fireEvent.click(button);
```
tay1orjones:
```suggestion
      userEvent.click(button);
```

packages/react/src/components/OverflowMenu/OverflowMenu-test.js
```diff
@@ -171,5 +171,29 @@ describe('OverflowMenu', () => {
         'true'
       );
     });
+
+    it('should call onClick handler only once per click', () => {
```
riddhybansal, line 175: it('should call onClick handler only once per click', async() => {

guidari (approved):

tay1orjones (approved):

yaoterry:
this fix actually broken our application using the OverflowMenu but there is application logic in onClick property for the OverflowMenu, and the onClick callback will also update the other property for the OverflowMenu, then the OverflowMenu open state will be messed up, the menu body can not be opened properly.

Add the event.stopPropagation() back or use a timeout in our onClick logic can work around this issue.

PR #16823, opened 2024-06-19
fix: invalid resolution of icon-helpers

tay1orjones:
The most recent v11.60.0 release [failed](https://github.com/carbon-design-system/carbon/actions/runs/9582636321/job/26422050126) because of an invalid resolution of `@carbon/icon-helpers` ([logs](https://github.com/carbon-design-system/carbon/actions/runs/9581789601/job/26419359700)). The `incremental-migration-vite` example relies on `carbon-components-react@7.x`, which relies on an older version of `@carbon/icons-react` than what we use in the rest of the monorepo. We couldn't resolve this via `dedupe` or `hoistingLimits` (the previous fix, #14481).

The solution we landed on here is to remove the `incremental-migration-vite` example from the repo to being a standalone stackblitz link. We've updated the readme to point to the link.

#### Changelog

**Changed**

- remove `incremental-migration-vite` project files
- modify `incremental-migration-vite` readme.md to link to the new stackblitz example

#### Testing / Reviewing

- There's not really a way to test this - we're just expecting CI checks to pass and we'll be cherry picking this into a new patch of `v11.60.1`


yarn.lock
```diff
@@ -1921,7 +1914,7 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@carbon/icon-helpers@npm:^10.28.0, @carbon/icon-helpers@npm:^10.49.0, @carbon/icon-helpers@workspace:packages/icon-helpers":
```
tay1orjones, line 1924: This is the source of the invalid resolution: `@carbon/icon-helpers@npm:^10.28.0`

guidari (dismissed):

yarn.lock
```diff
@@ -1921,7 +1914,7 @@ __metadata:
...
+"@carbon/icon-helpers@npm:^10.49.0, @carbon/icon-helpers@workspace:packages/icon-helpers":
```
tay1orjones, line 1917: Fixed here after removing the example from the workspaces

guidari (approved):

PR #16838, opened 2024-06-21
Tests: Added test suite to check the `selectedItems` prop from external state

### LINKED ISSUE #16828: [Test]: Add tests related to `selectedItems` for `MultiSelect`
### Description

As we had faced issues like #16716 and #16469, the crux of these issues seems to be like we have no tests covered the controlled state, the `selectedItems` prop. Will need some test suites to cover these, so that if any code commit happens on these there will be tests to catch them if they break.

Gururajj77:
Closes #16828

test to check the `selectedItems` from the external state

#### Changelog

**New**

- new `describe` test suite for controlled selected items

#### Testing / Reviewing

check the CI for this test suite or run `yarn test MultiSelect-test.js` by pulling down locally.


Gururajj77:
will be writing same tests for the `FilterableMultiSelect` too

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -528,4 +530,84 @@ describe('MultiSelect', () => {
...
+              kind="secondary"
+              onClick={() => setSelectedItems([])}>
+              Clear
+            </Button>
+          </ButtonSet>
+        </>
+      );
+    };
+
+    it('should initially render controlled multiselect with a given label', () => {
```
tay1orjones: Is this one just kind of a smoke test to ensure the component will even render with the given props? I'm not sure how valuable it is. I might be missing some context though, let me know what you think.

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
```diff
@@ -528,4 +530,84 @@ describe('MultiSelect', () => {
...
+      expect(isElementVisible(labelNode)).toBe(true);
+
+      expect(
+        // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+        container.querySelector(
+          '[aria-expanded="true"][aria-haspopup="listbox"]'
+        )
+      ).toBeNull();
+    });
+    it('should allow the items to be controlled from external state', async () => {
```
tay1orjones, line 654: The only negative I see with this approach is that the test is inherently impure. It primarily tests `ControlledComponent`, and only by proxy the `MultiSelect`'s `selectedItems` and `onChange`. If either `ControlledComponent` or `MultiSelect` were to change, there's a (small) possibility the "fuzziness" of this test would make it still pass when it shouldn't or vice versa. 

Check out [this article](https://moxio.com/blog/4-methods-for-testing-controlled-react-components/) - it does a better job explaining some approaches to this. The test you have now is most of "Route 4", just without the spy, which would be a great thing to add to this. 

Another option would be "Route 3" using `rerender` and an `onChange` mock, which also seems like a decent option. Maybe we have both! Here's a rough idea of "Route 3":



```js
it('should support controlled component functionality with selectedItems and onChange', async () => {
  const onChange = jest.fn();
  const { rerender } = render(<MultiSelect items={items} selectedItems={[items[0]]} onChange={onChange} />);

  // The selected items should match what's passed into selectedItems
  expect(screen.getAllByRole('option')[0]).toHaveAttribute('aria-selected', 'true');

  // onChange should fire for interactions
  await userEvent.click(screen.getAllByRole('option')[1])
  expect(onChange).toHaveBeenCalledTimes(1);
  expect(onChange).toHaveBeenCalledWith([items[1]]);

  // If the onChange event data is not used to update selectedItems, the selection should remain as it was before
  expect(screen.getAllByRole('option')[0]).toHaveAttribute('aria-selected', 'true');
  expect(screen.getAllByRole('option')[1]).toHaveAttribute('aria-selected', 'false');

  // Force a re-render with updated selectedItems matching the initial selection and the onChange event data
  render(<MultiSelect items={items} selectedItems={[items[0], items[1]]} onChange={onChange} />);

  // Now both should be selected
  expect(screen.getAllByRole('option')[0]).toHaveAttribute('aria-selected', 'true');
  expect(screen.getAllByRole('option')[1]).toHaveAttribute('aria-selected', 'true');
});
```

tay1orjones (commented):
Looks good, though I have a couple ideas for how we could make this even better. Let me know what you think!

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
Gururajj77: Had put this in the beginning to make sure that the `ControlledMultiSelect` actually renders if we have passed items through `selectedItems`. Just like we have one of this test in the beginning of `MultiSelect`, just to make sure if anyone made changes to how `selectedItems` work, then this could potentially catch the test.

packages/react/src/components/MultiSelect/__tests__/MultiSelect-test.js
Gururajj77: Added a test suite looking at this example, please let me know of any changes. I have kept the "route-4" type of test also as an example for us on how it could work from the test side, so that users can refer to it and write their tests for their implementation.

tay1orjones (approved):
Awesome!

2nikhiltom (approved):

riddhybansal (approved):

Issue #16145, opened 2024-04-08
[Bug]: FilterableMultiSelect focus clearing issue


crichtonibm:
### Package

@carbon/react

### Browser

Chrome, Firefox, Edge

### Package version

1.54.0

### React version

18.2.0

### Description

When using a FilterableMultiSelect inside of a Tab/TabPanel and you use the search bar to filter the list, if there is a scroll bar and you use the mouse to scroll the list, the search term clears out, just as if you were to click outside of the FilterableMultiSelect. Doing the same while not in a Tab/TabPanel does NOT clear the search term (as expected and desired).

Note, if you use the mouse wheel to scroll the list while nested in a Tab/TabPanel, the search term also does not clear (again, as expected and desired)

### Reproduction/example

https://stackblitz.com/edit/github-32jqsw

### Steps to reproduce

Select the filterable multi select inside the tab, search on aa and then try to scroll the list

### Suggested Severity

Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
This issue is still present. The specific interaction is grabbing the scrollbar with the mouse and scrolling it clears the input when it shouldn't. https://stackblitz.com/edit/github-32jqsw-2uli1s?file=package.json

https://github.com/user-attachments/assets/e974c21c-29dd-4567-9855-984c04565692

crichtonibm:
Bump to this one ... will this be considered soon? It is still an issue for us

PR #16840, opened 2024-06-21
fix: focus issue in filterable multiselect

### LINKED ISSUE #16830: [Bug]: Focus on FilterableMultiSelect always persists once click on the control
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

1.60.1

### React version

18.2.0

### Description

Hi Team, 

I have upgraded the carbon library in my application and replace the Multiselect drop down to FilterableMultiSelect. I am observing one issue while using this control.

After the page loads, If click on the control, the focus always persists,  if we move to another control by clicking through the mouse or moving one control to another control using tab button

Please look into this issue and provide a solution on this.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/experimental-unstable-slug-examples--filterable-multiselect

### Steps to reproduce

1. Open this URL: https://react.carbondesignsystem.com/?path=/story/experimental-unstable-slug-examples--filterable-multiselect
2. Click on the control
3. Next click outside of the control
4. The focus of the control remains there. it should be removed from the control
image

### Suggested Severity

### Application/PAL

https://react.carbondesignsystem.com/?path=/story/experimental-unstable-slug-examples--filterable-multiselect

### Code of Conduct

riddhybansal:
Closes #16830 

Focus issue in filterable multiselect

#### Changelog

**New**

- Added onBlur

#### Testing / Reviewing

Test filterable multiselect , it should work as intended. 
Using the Storybook Filterable, if you click in the select and type opt ... then try clicking on the (x) clear button, it should work , and it should lose focus after clicking outside.

Please verify this PR does not create another issue that has been closed already #16145


riddhybansal:
My only concerned over here is , is this the intended behaviour - when the filterable multiselect is opened clicking outside first closes the menu , and focus remains on input field to type something and then clicking outside again clears the focus.

Gururajj77:
@riddhybansal , clicking the (x) button isn't clearing the text box or clearing the selected options, is that intended behaviour?? Please correct me if I missed something

https://github.com/carbon-design-system/carbon/assets/89023023/3bd0a602-561d-4fba-a903-bfbb3db263e8

alisonjoseph:
I'm also seeing the same issue as @Gururajj77

guidari:
> My only concerned over here is , is this the intended behaviour - when the filterable multiselect is opened clicking outside first closes the menu , and focus remains on input field to type something and then clicking outside again clears the focus.

Also looks like the focus doesn't go away when we are tabbing out from the `ListBoxMenu`, maybe if we add an onBlur there it would fix your comment and the tab navigation...

> @riddhybansal , clicking the (x) button isn't clearing the text box or clearing the selected options, is that intended behaviour?? Please correct me if I've missed something

Good catch. In the [1.59.0](https://deploy-preview-16696--v11-carbon-react.netlify.app/?path=/story/components-multiselect--filterable) is working as expected the close icon.

We should also write tests for those two scenarios, to avoid any changes that might affect that in the future.

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -714,6 +712,11 @@ const FilterableMultiSelect = React.forwardRef(function FilterableMultiSelect<
         }
       },
       onFocus: () => setInputFocused(true),
+      onBlur: () => {
+        !isOpen && setInputFocused(false);
+        setInputValue('');
+        // setIsOpen(false);
```
tay1orjones:
```suggestion
```

tay1orjones (approved):

guidari (approved):

PR #16847, opened 2024-06-21
fix: switch to onClick from onMouseDown in modal components

### LINKED ISSUE #16846: [Bug]:Unintended ListBox onClick behavior component when inside Modal
### Package

@carbon/react

### Browser

Chrome

### Package version

1.60.2

### React version

18

### Description

This behavior was noticed when clicking on the label of a `Dropdown` component inside the `MultiAddSelect` component in the carbon for ibm products library. The associated issue can be found [here](https://github.com/carbon-design-system/ibm-products/issues/5450).

After some investigation I tracked the behavior down to the `Modal` component. You'll notice that `Dropdown` doesn't suffer this strange behavior when outside the context of a modal, which you can see in [storybook](https://react.carbondesignsystem.com/?path=/story/components-dropdown--inline). However, when you do the same to a dropdown inside of a modal you'll see a slightly different behavior. You can see this behavior directly in [storybook](https://react.carbondesignsystem.com/?path=/story/components-modal--with-scrolling-content) by clicking on the label of either the dropdown or multiselect and holding down.

This behavior was introduced when `stopPropagation()` was added to the `handleMousedown` handlers in `Modal` and `ComposedModal`. After testing and debugging locally the problem can be fixed by switching from `onMouseDown` to `onClick`. I'm not an expert on the differences between `onMouseDown` vs `onClick` but from what I understand because `onMouseDown` fires first the `stopPropagation` inclusion may have started to disrupt the event handling within `Modal`. I tested this by removing `stopPropagation` in the handler and leaving it as `onMouseDown`. When including `stopPropagation` it appears as though `onClick` is needed instead to avoid this behavior.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-modal--with-scrolling-content

### Steps to reproduce

Scroll down inside the modal until you reach either the `MultiSelect` or `Dropdown` components. Click the label once to open the content. Once open, click and hold down on the label. The component should be in a closed state until you release the cursor, which is not the intended / expected behavior.

### Suggested Severity

### Application/PAL

c4p

### Code of Conduct

davidmenendez:
Closes #16846

switches the outside click detector event in `Modal` and `ComposedModal` from `onMouseDown` to `onClick` in order to alleviate some strange behaviors noticed with components built on `ListBox` inside of modal components.

#### Changelog

**Changed**

- switched `onMouseDown` event to `onClick` in `Modal` and `ComposedModal`

#### Testing / Reviewing

- view the steps / description in the issue #16846 and follow the same steps against these changes


riddhybansal (approved):
This looks like a fix to me !!

Gururajj77 (approved):

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -421,7 +421,7 @@ const ComposedModal = React.forwardRef<HTMLDivElement, ComposedModalProps>(
         ref={ref}
         aria-hidden={!open}
         onBlur={!focusTrapWithoutSentinels ? handleBlur : () => {}}
-        onMouseDown={handleMousedown}
+        onClick={handleClick}
```
tay1orjones: Before this change a consumer could pass in `onClick`, but with this change they'll no longer be able to do that. The reverse is also true for `onMouseDown`, but that's going to be less of a concern. I doubt anyone has placed `onMouseDown` on a modal, because it would never have been called before this change.

To fix we can either conditionally call rest.onClick from within `handleClick`, or use the composeEventHandlers helper we have. I think we could use optional chaining to avoid errors for when they're not supplying their own `onClick`? 🤔 

```js
onClick={composeEventHandlers([rest.?onClick, handleClick])}
```

tay1orjones (changes requested):

packages/react/src/components/ComposedModal/ComposedModal.tsx
davidmenendez: great point! will fix 👍

davidmenendez:
@tay1orjones resolved 👍

tay1orjones:
@davidmenendez would you be up for writing a small test to ensure the consumer-provided onClick is called like we expect?

davidmenendez:
@tay1orjones tests included 👍

guidari (approved):
LGTM! 🚀

davidmenendez:
@tay1orjones when you have a chance 👍

alisonjoseph (approved):

davidmenendez:
@tay1orjones 👀 😬

tay1orjones (approved):
Sorry for the delay, was out on vacation last week 😅

PR #16866, opened 2024-06-25
fix(combobox): memoize floating ui ref to prevent max call depth

### LINKED ISSUE #16857: [Bug]: ComboBox crashes when typing fast with allowCustomValue=true
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.60.2

### React version

v18.2.0

### Description

ComboBox with `allowCustomValue` will crash if you type too fast.

This error occurs in the console
```
Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

### Reproduction/example

https://stackblitz.com/edit/github-ffn5vq?file=src%2FApp.jsx

### Steps to reproduce

Type quickly in the ComboBox.
Occurs in this storybook --> https://react.carbondesignsystem.com/?path=/story/components-combobox--allow-custom-value or in the linked [stackblitz](https://stackblitz.com/edit/github-ffn5vq?file=src%2FApp.jsx)

Video in storybook:

https://github.com/carbon-design-system/carbon/assets/52974257/7aac7e58-94c4-4967-849c-0bc73867d537

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #16857

This PR does to primary things:
1. Memoize the value of `getMenuProps` to avoid an infinite loop
2. Only apply floating ui refs when `autoAlign={true}`

The first fixes the problem outlined in the original issue, the second is an indirect improvement. If the refs had initially been guarded, this bug would've been isolated to users using `autoAlign`. Because the refs were not guarded, the bug was present for everyone regardless of if `autoAlign` was being used.

From what I can tell, there is not a reliable way to unit test this as the keyboard entry rate isn't configurable.

#### Changelog

**Changed**

- Wrap `getMenuProps` in `useMemo`
- Conditionally apply refs

#### Testing / Reviewing

- Go to the default combobox story. Type super fast - like just totally spam the keyboard. The component should not error
- Do the same on the `allowCustomInput` combobox story
- Do the same on the `with experimental autoalign` combobox story
- Ensure `autoAlign` works as expected on combobox by resizing the browser window, playing with scroll, and opening/closing the listbox


2nikhiltom (approved):

guidari (approved):
LGTM!

PR #16873, opened 2024-06-26
fix(search): a11y warning on search not expanded

### LINKED ISSUE #4766: Investigate datagrid `button` without tabbable element in datagrid toolbar
Reported by @kristastarr in the [Datagrid review](https://github.ibm.com/CDAI-design/pal/issues/1575) from the accessibility audit.

```
 Component with “button” role does not have a tabbable element
```

This one could be a Carbon issue or an example/Storybook-specific one. Let’s investigate and either address or reopen.

image

> ### What to do
> After deciding that the component should be keyboard-interactive, use the tabindex attribute to include the component in the tab sequence
> 
> Otherwise, change the role to a non-widget or non-interactive type component so there is no expectation that the component should be operable with the keyboard
> 
> Rule ID: widget_tabbable_exists
> Reason ID: fail_no_tabbable

ariellalgilmore:
Closes https://github.com/carbon-design-system/ibm-products/issues/4766

IBM Equal Access Accessibility Checker - warning on the search icon with `role="button"` since it's not tabbable

#### Changelog

**Changed**

- added a conditional on the `aria-label` and `role` for the search icon to only have values when `onExpand`

#### Testing / Reviewing

Ran the IBM Equal Access Accessibility checker on the component

Before:
Screenshot 2024-06-26 at 9 58 07 AM

After:
Screenshot 2024-06-26 at 9 57 24 AM

### DIFF
--- a/packages/react/src/components/Search/Search.tsx
+++ b/packages/react/src/components/Search/Search.tsx
@@ -228,9 +228,9 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
       however, it does not need a keyboard event bc the input element gets focus on keyboard nav and expands that way*/}
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
       <div
-        aria-label={onExpand ? 'button' : ''}
+        aria-label={onExpand ? 'button' : undefined}
         aria-labelledby={onExpand ? uniqueId : undefined}
-        role={onExpand ? 'button' : ''}
+        role={onExpand ? 'button' : undefined}
         className={`${prefix}--search-magnifier`}
         onClick={onExpand}
         onKeyDown={handleExpandButtonKeyDown}
--- a/packages/react/src/components/Search/Search.tsx
+++ b/packages/react/src/components/Search/Search.tsx
@@ -228,9 +228,9 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
       however, it does not need a keyboard event bc the input element gets focus on keyboard nav and expands that way*/}
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
       <div
-        aria-label="button"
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Search/Search.tsx
```diff
@@ -228,9 +228,9 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
       however, it does not need a keyboard event bc the input element gets focus on keyboard nav and expands that way*/}
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
       <div
-        aria-label="button"
+        aria-label={onExpand ? 'button' : ''}
```
guidari:
```suggestion
        aria-label={onExpand ? 'button' : undefined}
```

packages/react/src/components/Search/Search.tsx
```diff
@@ -228,9 +228,9 @@ const Search = React.forwardRef<HTMLInputElement, SearchProps>(function Search(
...
         aria-labelledby={onExpand ? uniqueId : undefined}
-        role="button"
+        role={onExpand ? 'button' : ''}
```
guidari:
```suggestion
        role={onExpand ? 'button' : undefined}
```

guidari (changes requested):
LGTM!
Just one note. Can we apply `undefined`, so we don't have an attribute without a value.

Screenshot 2024-06-26 at 15 22 49

packages/react/src/components/Search/Search.tsx
ariellalgilmore: yes good call!! thank you!

riddhybansal (approved):
LGTM 🚀

tay1orjones (approved):

alisonjoseph (approved):

guidari (approved):

PR #16898, opened 2024-06-30
Fix (16747)(A11y) : Proper Conveyance of Word/Character Count by screen readers 

### LINKED ISSUE #16747: [a11y]: [Text area][Playground] -  The word/character count not conveyed properly by JAWS
### Package

@carbon/react

### Browser

Firefox

### Operating System

_No response_

### Package version

v1.58.0

### React version

18.02

Manual QA

### Assistive technology

JAWS

### Description

The numbers` X/maxCount` are confusing and unclear in their purpose. For instance, it's not specified whether this represents a count of characters or words, which makes it difficult for screen reader users to understand.

When the maxCount word limit is set along with the enableCounter prop, the screen reader announcement for the word/character count used to only triggers once when 10 words/characters were left.

Also, no announcements are made when the maximum word/character count is reached. This inconsistency makes it difficult for screen reader users to understand the remaining count and when the limit is hit.

This functionality can be seen when  `enableCounter` prop os set to `true`

image

<br class="Apple-interchange-newline">

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-textarea--playground

### Suggested Severity

3

### Code of Conduct

2nikhiltom:
Closes #16747

When the `maxCount` word limit is set along with the `enableCounter` prop, the screen reader announcement for the word/character count used to only triggers once when `10 words/characters` were left. 

Also, no announcements are made when the `maximum` word/character count is reached. This inconsistency makes it difficult for screen reader users to understand the remaining count and when the limit is hit.

**This PR does two things** 

1. Incremental Announcements 
It will now make announcements for each of the last 10 characters/words, properly conveying the remaining count.

2. Max Count Announcement
An additional announcement will be made when the maximum word/character count is reached, ensuring users are informed when the limit is hit.

#### Changelog

The `useAnnouncer` function is updated to track when the maximum limit is reached and ensure announcements are made for each of the last 10 characters/words.

The test cases for `useAnnouncer` have been updated to reflect the new changes and ensure proper functionality.

Within the component, useAnnouncer is called inside a `useEffect` hook to manage the timing of announcements.
This is to ensures that same announcements are not repeated unnecessarily and are made after a delay, allowing both the counter and the announcement of the actual text being written to be audible

#### Testing / Reviewing

1. Vist textArea  playground in deployPreview
2. Enable enableCounter prop
3. Set counterMode to character/word (test for both)
4. Set maxCount to 12 
5. Type character/word: 
6. Run JAWS/VoiceOver and verify that the announcements are made for each of the last 10 characters/words and also verify that Max Count reached announcement is made

Also verify that readOnly status is announced by screenReader when readonly is true


packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -416,11 +416,33 @@ const TextArea = React.forwardRef((props: TextAreaProps, forwardRef) => {
...
+        // Set the new content after a small delay
+        setTimeout(
+          () => {
+            if (announcer) {
+              announcer.textContent = ariaAnnouncement;
+              setPrevAnnouncement(ariaAnnouncement);
+            }
+          },
+          counterMode === 'word' ? 2000 : 1000
+        );
```
tay1orjones: I think it might be best to be sure the timeout is cleared in the useEffect cleanup just in case the component unmounts before the timeout expires. In theory this should be garbage collected but it's a small add for this extra assurance.

packages/react/src/components/TextInput/TextInput.tsx
```diff
@@ -310,7 +316,25 @@ const TextInput = React.forwardRef(function TextInput(
...
+  const [prevAnnouncement, setPrevAnnouncement] = useState('');
   const ariaAnnouncement = useAnnouncer(textCount, maxCount);
+  useEffect(() => {
+    if (ariaAnnouncement && ariaAnnouncement !== prevAnnouncement) {
+      const announcer = announcerRef.current as HTMLSpanElement | null;
+      if (announcer) {
+        // Clear the content first
+        announcer.textContent = '';
+        // Set the new content after a small delay
+        setTimeout(() => {
```
tay1orjones: Same thing here with clearing this timeout just in case

tay1orjones (approved):
This looks good, but one small optimization we could make

riddhybansal (approved):

PR #16911, opened 2024-07-04
fix: changed types from `string` to `ReactNode` for `InlineNotification` and `ToastNotification`

### LINKED ISSUE #16909: [Bug]: The `subtitle` prop should be `node` type rather than `string` type
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.53.1

### React version

v17.0.2

### Description

The `subtitle` prop in `InlineNotification` in Carbon 11 is now a `string` type. In Carbon 10, the prop is a `node` type. This creates React warning all over the places when we upgrade from Carbon 10 to 11. Can you change the `subtitle` prop back to a `node` type?

### Reproduction/example

N/A

### Steps to reproduce

Here is an example of my code snippet:

```
                    <InlineNotification
                      kind="error"
                      role="alert"
                      lowContrast
                      hideCloseButton
                      title={t('Failed to fetch')}
                      subtitle={
                        <>
                          {t(
                            'An error occurred when fetching secrets. Try again later.'
                          )}
                          {errorGuid && (
                            <span className="cpc--secret-input__error-guid">{` (${errorGuid})`}</span>
                          )}
                        </>
                      }
                    />
```

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #16909 

fix: changed types from `string` to `ReactNode` for `InlineNotification` and `ToastNotification` for both typescript interfaces and prop-types

#### Changelog

**Changed**

- changed `string` to `ReactNode` for `InlineNotification`
- changed `PropTypes.string` to `PropTypes.node` for `InlineNotification`
- changed `string` to `ReactNode` for `ToastNotification`
- changed `PropTypes.string` to `PropTypes.node` for `ToastNotification`

#### Testing / Reviewing

Pull down to local, try to pass a component or any element which is not a string and is `ReactNode` and no warning should come up for typings.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -4325,7 +4325,7 @@ Map {
         "type": "string",
       },
       "subtitle": Object {
-        "type": "string",
+        "type": "node",
       },
       "title": Object {
         "type": "string",
@@ -9100,7 +9100,7 @@ Map {
         "type": "string",
       },
       "subtitle": Object {
-        "type": "string",
+        "type": "node",
       },
       "timeout": Object {
         "type": "number",
--- a/packages/react/src/components/Notification/Notification.tsx
+++ b/packages/react/src/components/Notification/Notification.tsx
@@ -378,7 +378,7 @@ export interface ToastNotificationProps extends HTMLAttributes<HTMLDivElement> {
   /**
    * Specify the subtitle
    */
-  subtitle?: string;
+  subtitle?: ReactNode;
 
   /**
    * Specify an optional duration the notification should be closed in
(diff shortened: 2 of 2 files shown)


2nikhiltom (approved):

tay1orjones (commented):
This was intentional. With the role of `alert`, `log`, or `status`, screenreaders generally do not convey any semantics surrounding the content. Formatting, emphasis, bold, etc. is ignored so for v11 we scoped this to being a string. Plus of course the limitation of those roles not being able to contain interactive elements either.

The escape hatch is `children`. It's still a node as usual. The caveat is that anything put in there beyond plain text will be inaccessible.

Recently [StaticNotification](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-staticnotification--with-action-button-and-links) was added based on feedback surrounding this and other limitations with notifications in v11. It's more open ended but does very little a11y-wise, which makes it very easy to create inaccessible situations. The overview docs in storybook cover this in more detail.

Was there a specific use case for opening this up to accept a node again?

Gururajj77:
Closed this PR as the change will be problematic for screen reader users.

PR #16918, opened 2024-07-05
style(datatable): isolate row hover styles from Toggletip links

cuppajoey:
#15965 attempted to prevent table row hover states from affecting links within nested `Toggletips`; however, the solution was incomplete and caused a style regression.

**❌ Before**

https://github.com/carbon-design-system/carbon/assets/14837881/84eb108c-f9c4-41b8-8f1d-b481c3024f8b

As demonstrated above:
- When a `Toggletip` is adjacent to a link, the intended hover styles are no longer applied to the link (i.e., the link color does not change when the row is hovered on). 
- When a `Toggletip` is in an adjacent cell, the row hover state works, but it affects the `Toggletip`'s link.

### 📝 Changelog

**Changed**

- Modify CSS selector to _only_ change a link's color when it is a child of a table cell _but_ not within a `Toggletip`.

**✅ After**

https://github.com/carbon-design-system/carbon/assets/14837881/9f66b883-f59c-43be-a56a-595260a2fb80

### 🧪 Testing / Reviewing

- [ ] Checkout and build branch
- [ ] Add this modified `DataTable` storybook example: [carbon-datatable-toggletip-story.js](https://gist.github.com/cuppajoey/b828d1f9bf20270c6cfd73a409875f0b#file-carbon-datatable-toggletip-story-js)
- [ ] Run storybook
- [ ] Open the modified test case ("Test with Tooltips" if you use copied my Gist)
- [ ] Hover over each table row and confirm the link styles are correct.


riddhybansal:
Hey can you create a test story, so that we can test in deploy that can be removed once we approve.

cuppajoey:
> Hey can you create a test story, so that we can test in deploy that can be removed once we approve.

@riddhybansal I've added a test story which you can find under [DataTable → Basic → Test with Tooltips](https://deploy-preview-16918--v11-carbon-react.netlify.app/?path=/story/components-datatable-basic--test-with-tooltips)

tay1orjones (approved):
Thanks for doing this! LGTM once the test story is removed.

cuppajoey:
> Thanks for doing this! LGTM once the test story is removed.

My pleasure! I'll wait until the final review is finished then I'll remove the test story.

alisonjoseph (approved):
LGTM! Will merge after you remove test story

cuppajoey:
> LGTM! Will merge after you remove test story

Test story is removed ✅

PR #16927, opened 2024-07-09
feat(icons): deprecate and remove AI lettering icon

conradennis:
Deprecating the `AI` lettering icon. The icon is causing confusion among product teams. The definitive AI icon should be the Carbon for AI slug.

Let me know if I should remove the icon from the `icons.yml` or `categories.yml` files as well.

Per the request of @mjabbink 
original issue for reference: https://github.ibm.com/brand/ui-icons/issues/335

#### Changelog

**Removed**

- 32/AI.svg


packages/icons/deprecated.yml
```diff
@@ -25,3 +25,7 @@ deprecated:
     reason:
       Icon represents the old avatar for the Watson brand, which is retired
       since the watsonx brand launch. Replaced by the icon watsonx
+  - name: AI
+    reason:
+      Icon is causing confusion and should be removed from the repo and library
+      in favor of product teams using the Carbon for AI slug
```
alisonjoseph:
```suggestion
      Icon should not be used. Teams should use the Carbon 
      AILabel component.
```

alisonjoseph (changes requested):
Small suggestion for the deprecation reason. The slug is in the process of being renamed to AILabel. We can't actually remove the icon as it would be a breaking change, however we can hide the icon from the icon library on the Carbon website to help reduce confusion similar to what we did for IBM Z https://github.com/carbon-design-system/carbon-website/pull/2781

packages/icons/deprecated.yml
conradennis: Works for me, I'll update the reason in the yml file.

alisonjoseph (approved):

2nikhiltom (approved):

riddhybansal (approved):

PR #16970, opened 2024-07-16
fix(overflowmenu): always use button kind=ghost

### LINKED ISSUE #16836: [Bug]: `OverflowMenu` styles overridden by `Button`
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.60.0

### React version

v18.2.0

### Description

When importing component-level stylesheets, rather than the full carbon stylesheet, `Button`'s styles will override `OverflowMenu`'s if the button stylesheet is imported later.

image

This is because `OverflowMenu` does not set the rendered `IconButton`'s `kind` prop which defaults to primary and the style selectors have an equal specificity (`.cds--btn--primary` vs `.cds--overflow-menu`).

#### 💡 Possible solution
The documentation for `OverflowMenu` states that it [_should_ only render a ghost button style](https://carbondesignsystem.com/components/menu-buttons/usage/#overflow-menu). If that's true, we should set the `kind="ghost"` which would prevent primary button styles from overriding it.

At the very least, we should set "ghost" as a default value and allow consumers to change it.

### Reproduction/example

https://stackblitz.com/edit/github-utkywz?file=src%2FApp.jsx,src%2Fbutton.scss,src%2Foverflow.scss

### Steps to reproduce

1. Render a `Button` and an `OverflowMenu` with default values.
2. Import `@carbon/styles/scss/components/overflow-menu`.
3. Later import `@carbon/styles/scss/components/button`.
4. Observe the `OverflowMenu` adopt `Button`'s primary styles.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/16836

This enforces the [design spec](https://carbondesignsystem.com/components/menu-buttons/usage#overflow-menu:~:text=Overflow%20menu%20has%20the%20ghost%20button%20style%20only) for OverflowMenu:

> Overflow menu has the ghost button style only

#### Changelog

**Changed**

- Add kind="ghost" to OverflowMenu
- Add kind="ghost" to next/OverflowMenu (behind feature flag)

#### Testing / Reviewing

- There should be no visible changes to OverflowMenu stories
- There should be no visible changes to the feature flagged OverflowMenu stories
- I confirmed this fix based on the [stackblitz example](https://stackblitz.com/edit/github-utkywz-udbg5t?file=src%2FApp.jsx,src%2Fbutton.scss,src%2Foverflow.scss) from the original issue. Adding `kind="ghost"` solves it.


alisonjoseph (approved):
LGTM! thanks for adding the test 😃 (I think it just needs a snapshot update?)

riddhybansal (approved):

guidari (approved):

PR #16988, opened 2024-07-17
fix(react): use `useId` instead of `getInstanceId`

tay1orjones:
[Brought up in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1721175434456939), there's quite a few components that use `setupGetInstanceId()` where they could/should be using `useId()` because:

1. `useId` in React 18+ is unified between server and client
2. `useId` has internal state to guard the id from changing based on rerenders of the parent component

The specific issue that highlighted this was that Modal's unique id changes every time it re-renders ([example](https://stackblitz.com/edit/github-n6nfpx-ecsmnm?file=package.json)). To avoid this you currently have to move state downwards ([example](https://stackblitz.com/edit/github-n6nfpx?file=package.json)), which isn't possible in all cases. Elsewhere in the codebase setupGetInstanceId was already guarded from changing on re-renders by using `useMemo` or `useRef`. 

There are still two class-based components that cannot take advantage of this, OverflowMenu (the old implementation, not the feature flagged one) and DataTable.

#### Changelog

**Changed**

- Modify functional components to utilize `useId` instead of `setupGetInstanceId`

#### Testing / Reviewing

For each component:

1. Go to the playground story of each component
2. Open the inspector, find the element with the class containing the unique id
3. Modify a prop in the controls pane, the id should not change

<details><summary>Example of a failure (in prod right now)</summary>
<p>

https://github.com/user-attachments/assets/cb0ad18d-8fa6-4515-8fc2-c7a3feb0d8d9

</p>
</details> 

<details><summary>Example with this PR</summary>
<p>

https://github.com/user-attachments/assets/5b7f7e99-89e8-456c-a246-8db609bdba45

</p>
</details>

### DIFF
--- a/packages/react/src/components/Checkbox/Checkbox.tsx
+++ b/packages/react/src/components/Checkbox/Checkbox.tsx
@@ -11,10 +11,8 @@ import classNames from 'classnames';
 import { Text } from '../Text';
 import { usePrefix } from '../../internal/usePrefix';
 import { WarningFilled, WarningAltFilled } from '@carbon/icons-react';
-import setupGetInstanceId from '../../tools/setupGetInstanceId';
+import { useId } from '../../internal/useId';
 import { noopFn } from '../../internal/noopFn';
-
-const getInstanceId = setupGetInstanceId();
 
 type ExcludedAttributes = 'id' | 'onChange' | 'onClick' | 'type';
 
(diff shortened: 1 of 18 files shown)


guidari (approved):
Interesting catch! LGTM! 🚀

riddhybansal (approved):
Clean !!

PR #16995, opened 2024-07-18
feat(popover): add `autoAlignBoundary` for configurable collision boundary

### LINKED ISSUE #16879: Floating UI: Add `detectOverflow` configurable collision boundary
In certain situations consumers may need to customize the boundary/element used to detect collisions. One example of this might be the modal - components like Popover may need to reflow based on the bounds of the modal content container, and not the viewport. 

To this we can implement the [detectOverflow](https://floating-ui.com/docs/detectoverflow) as a middleware 

Eventually we could also
* add props for consumers to pass in their own padding
* add this prop to other components using floating-ui autoAlign
* ensure this is configurable from anywhere in their react tree #17003

### LINKED ISSUE #16982: Floating UI Examples: Modal example updates (mask image fix)
We should be able to make this example once the https://github.com/carbon-design-system/carbon/pull/16947 fix is in

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/16879
Closes https://github.com/carbon-design-system/carbon/issues/16982
Relies on https://github.com/carbon-design-system/carbon/pull/17139

This adds a new prop that enables consumers to configure what element is used in the overflow detection for `autoAlign` via floating-ui. The new storybook example shows a custom boundary element (marked by a dashed border). The popover repositions based on this element, rather than the viewport.

#### Changelog

**New**

- Add `autoAlignBoundary` to Popover

#### Testing / Reviewing

- Look at the new storybook example and see if it works as you'd expect
- Ensure no regressions are present in popover


packages/react/src/components/Popover/Popover.stories.js
```diff
@@ -269,6 +269,75 @@ export const ExperimentalAutoAlign = () => {
...
+        placeItems: 'center',
+        overflow: 'scroll',
+        width: '800px',
+        height: '500px',
+        border: '1px',
+        borderStyle: 'dashed',
+        borderColor: 'black',
+        margin: '0 auto',
+      }}
+      ref={setBoundary}>
```
tay1orjones, line 295: TIL how this pattern works: https://www.youtube.com/watch?v=leEdHQh6nP8

ariellalgilmore (approved):
LGTM! the video was really helpful too!

alisonjoseph (approved):
nice! LGTM

PR #17002, opened 2024-07-19
fix(dropdown): memoize floating ui ref to prevent max call depth

### LINKED ISSUE #17001: [Bug]: Dropdown crashes with "Error: Maximum update depth exceeded..." coming from floating-ui
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.61.0

### React version

18.3.1

### Description

Dropdown component crashes when opening and trying to select an option after the update to use floating-ui.
This is only reproductible when running Cypress tests, but it produces an error similar to the one reported previously for Combobox component in https://github.com/carbon-design-system/carbon/pull/16866: 
```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. 
React limits the number of nested updates to prevent infinite loops.
```

### Reproduction/example

https://stackblitz.com/edit/github-rsnzd7?file=src%2FApp.cy.jsx

### Steps to reproduce

This can be only reproduced when running cypress tests.
The sandbox does not allow to run cypress tests, but you can download the project and run the test locally:
- Download the project and extract the zip
- Run `npm i` inside the project folder
- If cypress not available locally, run `npx cypress install`
- Execute cypress test with `npm run cypress:run`

The logs will contain the error
```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. 
React limits the number of nested updates to prevent infinite loops.
```

### Suggested Severity

### Application/PAL

IBM Event Automation

### Code of Conduct

EAlexRojas:
Closes https://github.com/carbon-design-system/carbon/issues/17001

Applies similar approach as for the Combobox component (in https://github.com/carbon-design-system/carbon/pull/16866) to memoize the value of `getMenuProps` to avoid an infinite loop.

#### Changelog

**Changed**

- Wrap `getMenuProps` in `useMemo`
- Conditionally apply refs

#### Testing / Reviewing

- Go to default dropdown story. Verify there is no regression for default behaviour.
- Ensure autoAlign works as expected on dropdown by resizing the browser window, playing with scroll, and opening/closing the listbox
- The original issue only appeared when running cypress tests, only way to validate this fixes the bug in that case would be to add a cypress test for the Dropdown component.

Note: The fix was validated in an existing app by applying a patch.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -14,6 +14,7 @@ import React, {
   MouseEvent,
   ReactNode,
   useEffect,
+  useMemo,
 } from 'react';
 import {
   useSelect,
@@ -497,8 +498,13 @@ const Dropdown = React.forwardRef(
           },
         };
 
-    const menuProps = getMenuProps();
-    const menuRef = mergeRefs(menuProps.ref, refs.setFloating);
+    const menuProps = useMemo(
+      () =>
+        getMenuProps({
+          ref: autoAlign ? refs.setFloating : null,
+        }),
+      [autoAlign]
+    );
 
     // Slug is always size `mini`
     let normalizedSlug;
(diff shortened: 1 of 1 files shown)


EAlexRojas:

tay1orjones (approved):
Thanks for catching this, submitting the fix, and testing it! Since it's so close to that other PR and you've said it fixes your test case, I think this looks good to merge

guidari (approved):
LGTM!

PR #17008, opened 2024-07-22
fix(16916): adds React fragment support for Switcher's direct child

### LINKED ISSUE #16916: [Bug]: Switcher element crashes application when a react fragment is passed as a child
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

`> v1.42.1`

### React version

v18.2.0

### Description

When you add a React Fragment as a direct child of the `<Switcher>` component, the application crashes with the following error in the console:

```
Uncaught TypeError: WeakMap key Symbol.for("react.fragment") must be an object
```

Screenshot 2024-07-04 at 10 51 21 a m

Example in v1.42.1 where it was working
https://stackblitz.com/edit/github-jhr7bv?file=src%2FApp.jsx

### Reproduction/example

https://stackblitz.com/edit/github-jhr7bv-rzbbuj?file=src%2FApp.jsx

### Steps to reproduce

- Add a `<Switcher>` component
- Add a React Fragment with other things inside `<></>`
- Check the console for errors

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/16916

~This PR fix an issue where the `Switcher` component crashes when a React Fragment is used as a direct child. 
It's fix involves modifying the child handling logic to properly handle the Fragment case~

> @2nikhiltom Nice, I see this outlined in the [package docs](https://www.npmjs.com/package/react-keyed-flatten-children#for-everyone-else). This makes the most sense to me. So instead of making these changes to Switcher, could this PR be updated to document using react-keyed-flatten-children? Maybe in a new document named /packages/react/docs/composition.md?

After discussions, we concluded to have this PR to update the usage doc  `UsingFragmentsWithSwitcher.md` 

#### Changelog

**New**

~Added flattening logic for children, including those within Fragments~
 ~Introduced unique index and key generation for Fragment children~
 
 Added a new .md file which suggest the usage of [react-keyed-flatten-children](https://www.npmjs.com/package/react-keyed-flatten-children#react-keyed-flatten-children)

**Changed**

#### Testing / Reviewing

~1. Verify that when `Switcher` component is passed with React Fragments as direct children, application does not crash~
~2. Verify focus navigation | keyboard navigation works like before~
~3. Verify that the existing functionality is intact~

Read and verify the .md files makes sence and that I did not miss anything


riddhybansal (approved):
Nice fix!!

Gururajj77 (approved):

preetibansalui (approved):
LGTM

tay1orjones (commented):
This is an interesting problem. I'm curious about a few things:

* Will this work with multiple levels of nesting?
* Does it preserve `key`s and props placed on nested and non-nested children and fragments?

More conceptually:

* Do you think our components should support this use case? The intent today is for `Switcher` to have direct children of `SwitcherItem` - that's it. No nested children, fragments, etc.
* Is there a way for a consumer to work around this on their own without having to modify the source of our components?

tay1orjones:
Did you look at https://github.com/grrowl/react-keyed-flatten-children? It only has a single dependency on `react-is`. We could use it in our components, but it also looks like consumers can use it on their own to use fragments if they'd like. Thoughts?

2nikhiltom:
Hey @tay1orjones , 
[react-keyed-flatten-children](https://www.npmjs.com/package/react-keyed-flatten-children#react-keyed-flatten-children) makes sure that it flatten arrays and React.Fragments into a regular, one-dimensional a**rray while ensuring element and fragment keys are preserved, unique, and stable between renders.**

The current implementation works fairly fine for single level of nesting,  preserving keys and props placed on nested and non-nested children and fragments becomes messy incase of multilevel nesting which is exactly solved by above utility.

The currently purpose of [Switcher](https://carbondesignsystem.com/components/UI-shell-header/usage/#anatomy) is to provides a way for the user to easily navigate between products and systems and expects a `SwitcherItem` as its direct child as SwitcherItem is expected to have a link/href to redirect somewhere when clicked
Instead of modifying the Switcher component to accommodate this, it makes more sense for consumers to use the react-keyed-flatten-children package to manage their use cases. This will keeps dependencies light and the component source clean.

tay1orjones:
> Instead of modifying the Switcher component to accommodate this, it makes more sense for consumers to use the react-keyed-flatten-children package to manage their use cases. This will keeps dependencies light and the component source clean.

@2nikhiltom  Nice, I see this outlined in the [package docs](https://www.npmjs.com/package/react-keyed-flatten-children#for-everyone-else). This makes the most sense to me. So instead of making these changes to Switcher, could this PR be updated to document using `react-keyed-flatten-children`? Maybe in a new document named `/packages/react/docs/composition.md`?

Gururajj77 (approved):

guidari (approved):

preetibansalui (approved):

PR #17022, opened 2024-07-23
fix: memoize getMenuProps for Multiselect and Filterable Multiselect

### LINKED ISSUE #16867: Only apply floating ui refs when `autoAlign={true}`
This is something I originally found in https://github.com/carbon-design-system/carbon/pull/16866, but we should make sure floating ui refs are conditionally applied only when `autoAlign`. This will prevent autoAlign/ref-related bugs from being present for consumers who aren't using autoAlign. We'll have to do this across all the various components using floating ui, but should be a small task for each.

```js
ref={autoAlign ? refs.setReference : null}
```

```js
ref={autoAlign ? refs.setFloating : null}
```

### LINKED ISSUE #17020: Floating UI: Memoize MultiSelect floating ui ref in to prevent max call depth
Issue for #17007

### LINKED ISSUE #17021: Floating UI: Memoize FilterableMultiSelect floating ui ref in to prevent max call depth
Issue for #17007

riddhybansal:
Closes #16867
Closes #17021 
Closes #17020

This PR does these  primary things to Multiselect and Filterable Multiselect
1. Memoize the value of `getMenuProps` to avoid an infinite loop
2. Only apply floating ui refs when `autoAlign={true}`

The first fixes the problem for max depth call that results in infinite loop, the second is an indirect improvement. If the refs had initially been guarded, this bug would've been isolated to users using `autoAlign`. Because the refs were not guarded, the bug was present for everyone regardless of if `autoAlign` was being used.

This PR also verifies if the refs of components - **popover**, **tooltip**, **dropdown**,  are guarded with autoalign prop.

#### Changelog

**Changed**

- Wrap `getMenuProps` in `useMemo` to **multiselect** and **filterable multiselect**
- Conditionally apply or verify refs in **multiselect** , **filterable multiselect** , **popover**, **tooltip**, **dropdown**

#### Testing / Reviewing

- Go to the default multiselect story. Type super fast - like just totally spam the keyboard. The component should not error
- Do the same on the `with experimental autoalign` multiselect story
- Ensure `autoAlign` works as expected on multiselect by resizing the browser window, playing with scroll, and opening/closing the listbox
- Do the same for filterable multiselect.

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/react/src/components/ComboBox/ComboBox.tsx
@@ -654,6 +654,7 @@ const ComboBox = forwardRef(
       (helperText && !isFluid && helperTextId) ||
       undefined;
 
+    // Memoize the value of getMenuProps to avoid an infinite loop
     const menuProps = useMemo(
       () =>
         getMenuProps({
--- a/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
+++ b/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
@@ -30,6 +30,7 @@ import React, {
   type KeyboardEvent,
   ReactElement,
   useLayoutEffect,
+  useMemo,
 } from 'react';
 import { defaultFilterItems } from '../ComboBox/tools/filter';
 import {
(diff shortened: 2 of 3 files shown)


2nikhiltom (approved):
Looks good to me 💯

guidari (approved):

preetibansalui (approved):
LGTM

Issue #11175, opened 2022-04-07
Plex “fi” ligatures aren’t working


mjabbink:
@alisonjoseph commented on [Mon Mar 28 2022](https://github.com/carbon-design-system/carbon-platform/issues/448)

Carbon Next
<img width="96" alt="Screen Shot 2022-03-28 at 1 01 38 PM" src="https://user-images.githubusercontent.com/2753488/160459216-4fb31f8b-9508-488a-b8c9-45fb3d6ea461.png">

Example of what should be happening from the Gatsby theme
<img width="98" alt="Screen Shot 2022-03-28 at 1 00 45 PM" src="https://user-images.githubusercontent.com/2753488/160459218-dce40498-149d-4a6d-8946-f4de62ec35b4.png">

_Note, this may be an issue coming from core Carbon_ 👀 

---

@mattrosno commented on [Mon Mar 28 2022](https://github.com/carbon-design-system/carbon-platform/issues/448#issuecomment-1080979886)

We didn't do anything custom for fonts; using Carbon v11 straight up. This is all happening for us with `@use '@carbon/react';` in `/styles/vendor/_carbon.scss`.

---

@mjabbink commented on [Wed Mar 30 2022](https://github.com/carbon-design-system/carbon-platform/issues/448#issuecomment-1083362680)

Probably core Carbon. I think the dotcom sites are not missing that as well.

---

@alisonjoseph commented on [Thu Mar 31 2022](https://github.com/carbon-design-system/carbon-platform/issues/448#issuecomment-1085355490)

Confirmed this is an issue coming from core Carbon

---

@mjabbink commented on [Thu Apr 07 2022](https://github.com/carbon-design-system/carbon-platform/issues/448#issuecomment-1092023697)

@alisonjoseph is there an issue made in core Carbon or should I make one?

tay1orjones:
We should add some example text to the [elements preview](https://carbon-elements.netlify.app/type/examples/preview/) that can test for common/all ligatures.

mjabbink:
FYI: The ligatures have been working for several years or more. Not sure what was done recently to remove the OpenType feature.

mjabbink:
FYI: The ligatures have been working for several years or more. Not sure what was done recently to remove the OpenType feature.

abbeyhrt:
[...]

PR #17025, opened 2024-07-23
feat(@ibm/plex): Update plex dependency to use new packages

tw15egan:
Updates `@carbon/styles`  to use individual `@ibm/plex-${font-style}` packages rather than importing all from `@ibm/plex`

#### Changelog

**New**

- 8 new packages for each of the `@ibm/plex` families 

Screenshot 2024-07-25 at 9 17 15 AM

- `font-feature-settings: 'liga' 1;` added to `_reset.scss` to ensure ligatures render even with `letter-spacing` set
- `$package-name` field added to font src resolver

**Changed**

- Updated the `url` field in font src resolver to use new package name format 

**Removed**

- Split fonts do not seem to ship `.woff`, so [they have been removed](https://github.com/carbon-design-system/carbon/pull/17025/files#diff-95fa1084e7c21be991e52f92b77cc405b040c0356534cd670dc5db7708739dafR16) from the individual font formats

#### Testing / Reviewing

Ensure the fonts are loaded correctly, either using the Akamai CDN or not. Ensure the `fi` ligature renders properly.


tay1orjones:
One thing to verify here is the ligature fixes from #11175 https://github.com/IBM/plex/issues/438 are coming through. The storybook doesn't turn on the `$use-akamai-cdn` flag, so the `fi` ligature is still broken. For example in the notification component:



This should be fixed by this PR where we consume the new packages that contain these updates to the font files. I know this is still in draft, but right now I don't see the ligatures in the deploy preview.

tw15egan:
@tay1orjones it seems like a component-level issue for some reason. I will take a deeper look. I see the ligatures when just putting text outside of the notification Screenshot 2024-07-24 at 2 38 04 PM

tay1orjones:
@tw15egan wow, yeah it's part of the [current css-fonts-3 spec](https://www.w3.org/TR/css-fonts-3/#:~:text=setting%20a%20non%2Ddefault%20value%20for%20the%20%E2%80%98letter%2Dspacing%E2%80%99%20property%20disables%20common%20ligatures.) and the working draft [css-fonts-4 spec](https://www.w3.org/TR/css-fonts-4/#:~:text=setting%20a%20non%2Ddefault%20value%20for%20the%20letter%2Dspacing%20property%20disables%20optional%20ligatures.)

> setting a non-default value for the ‘letter-spacing’ property disables common ligatures.

tay1orjones (approved):

alisonjoseph (approved):
🔥

PR #17031, opened 2024-07-24
fix: added readonly state variant to passwordInput

### LINKED ISSUE #16922: [Bug]: PasswordInput doesn't have read only styling
### Package

carbon-components

### Browser

Chrome, Safari, Firefox, Edge

### Package version

/

### React version

[@carbon/react@1.61.0](https://github.com/carbon-design-system/carbon/tree/main/packages/react)

### Description

PasswordInput doesn't match the readonly styling of the text input. Slack conversation implies PasswordInput doesn't have read only styling assigned. 

image

Slack thread: https://ibm-cloud.slack.com/archives/C0M053VPT/p1720442561210339

### Reproduction/example

N/A

### Steps to reproduce

N/A

### Suggested Severity

### Application/PAL

App Connect SaaS

### Code of Conduct

riddhybansal:
Closes #16922 

PasswordInput does not have readonly state

#### Changelog

**New**

- Added readonly state and test cases for the same

#### Testing / Reviewing

Check if readonly state of passwordInput is working as intented.


2nikhiltom (approved):
LGTM

packages/react/src/components/TextInput/PasswordInput.tsx
```diff
@@ -442,6 +444,8 @@ PasswordInput.propTypes = {
    */
   invalid: PropTypes.bool,
 
+  readOnly: PropTypes.bool,
```
alisonjoseph, line 456: missing comment here
```suggestion
   /**
   * Whether the PasswordInput should be read-only
   */
  readOnly: PropTypes.bool,
```

alisonjoseph (changes requested):
Looks like the show/hide icon is missing styling for readOnly, I believe it should use the $icon-disabled token, and not have any hover states/cursor change.

riddhybansal:
> Looks like the show/hide icon is missing styling for readOnly, I believe it should use the $icon-disabled token, and not have any hover states/cursor change.

I thought of a same but I could  see  hover things in disabled state as well , thought its what designed. So do we need to use the $icon-disabled token, and not have any hover states/cursor changes for both disable and readOnly state

riddhybansal:
> > Looks like the show/hide icon is missing styling for readOnly, I believe it should use the $icon-disabled token, and not have any hover states/cursor change.
> 
> I thought of a same but I could see hover things in disabled state as well , thought its what designed. So do we need to use the $icon-disabled token, and not have any hover states/cursor changes for both disable and readOnly state

Fixed for both disable and readonly state

alisonjoseph (commented):
I had one question about the cursor, I'm wondering if the show/hide cursor shouldn't change on hover when in readonly. @laurenmrice?

Kritvi-bhatia17 (commented):
> I had one question about the cursor, I'm wondering if the show/hide cursor shouldn't change on hover when in readonly. @laurenmrice?

Hi @alisonjoseph @riddhybansal,

I checked the readonly state for Combobox and NumberInput (which have interactive elements similar to PasswordInput) and noticed that the cursor does not change on hover. 
However, @laurenmrice can confirm this for further validation.

https://github.com/user-attachments/assets/c2ca410f-bd47-4c8c-acc8-a2871d453667

---

Also, a small doubt: are we planning to add the readonly state to the FluidPasswordInput as well?

laurenmrice (changes requested):
@Kritvi-bhatia17 @riddhybansal 

When hovering over the eye icon button in the read-only state, the cursor should be an `arrow` and not a pointer. We use the pointer cursor when you can actually interact with the icon button when its no longer in the read-only state.
Frame 1

Kritvi-bhatia17:
> Also, a small doubt: are we planning to add the readonly state to the FluidPasswordInput as well?

Hi @riddhybansal, could you please confirm this?

alisonjoseph (approved):
This LGTM. Did you want to add to the fluid version also, or create a new issue?

@Kritvi-bhatia17 do we already have issues for design to add to Figma? Looks like read-only states are missing there also. 👀

riddhybansal:
> This LGTM. Did you want to add to the fluid version also, or create a new issue?
> 
> @Kritvi-bhatia17 do we already have issues for design to add to Figma? Looks like read-only states are missing there also. 👀

I guess there are missing read only states for many fluid components we can tackle them together as a list of task maybe

laurenmrice (approved):
Yeah, lets add the fluid in at a future date in a separate PR. Looks good to me @riddhybansal ! ⭐️

PR #17038, opened 2024-07-24
chore: added types to TreeNode & TreeView

### LINKED ISSUE #13581: Add TypeScript types to TreeView, TreeNode

imp-dance:
Closes #13581
Relates to #12513

#### Changelog

**New**

- Typescript types for TreeNode & TreeView components

**Changed**

- Fallback on certain aria attributes should be `undefined`, not `null`

#### Testing / Reviewing

Did a tiny test, but it'd be useful if someone with intimate knowledge of how the component internals work could take a look. 

Some of the props (`depth`, `selected`)  seem to be meant for only internal use, so I've typed them as optional (and ensured that the JSDoc comment is there) - but made sure to cast them as non nullable inside the component - assuming the parent ensures that they are given.

I also had to use `// @ts-ignore` in a couple of spots, namely in the PropTypes - since the `CarbonIconType` does not seem to match `PropTypes.oneOfType([PropTypes.func, PropTypes.object])`, and `(string | number)[]` does not match `PropTypes.arrayOf(
    PropTypes.oneOfType([PropTypes.string, PropTypes.number])
  ),`. If you know of any way we could drop the use of ts ignore here and fix the types or proptypes, that would be great.


imp-dance:
Also looks like this is the last checkbox @ https://github.com/carbon-design-system/carbon/issues/12513 - which I guess means the issue should either be closed or be updated with new items 🚀

Gururajj77 (commented):
Hi @imp-dance! 👋

Great work on the PR, it's looking good overall. I have one small suggestion:

Could you please add comments to the remaining props in the interface? This helps improve the intellisense for the typings, making it easier for developers to understand and use the component.

Thanks for your attention to detail! 🙌

imp-dance:
> Hi @imp-dance! 👋
> 
> Great work on the PR, it's looking good overall. I have one small suggestion:
> 
> Could you please add comments to the remaining props in the interface? This helps improve the intellisense for the typings, making it easier for developers to understand and use the component.
> 
> Thanks for your attention to detail! 🙌

Done! 🚀

imp-dance:
Actually, hold on a bit - I'll update the index file as well, and make sure that `TreeView.TreeNode` is typed properly.

imp-dance:
@Gururajj77  Done 👍 I'd appreciate another quick look over the changes when you can 😎

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -202,17 +272,25 @@ export default function TreeView({
...
+                `${prefix}--tree-node--disabled`
+              )
+            ) {
+              return NodeFilter.FILTER_REJECT;
+            }
+            if ((node as HTMLElement).matches(`li.${prefix}--tree-node`)) {
+              return NodeFilter.FILTER_ACCEPT;
+            }
+            return NodeFilter.FILTER_SKIP;
+          },
```
imp-dance: Maybe it would be more clean to do an early return if the node is not an Element? 

```ts
if (!(node instanceof Element)) { return NodeFilter.FILTER_SKIP; }
```

We are hiding a potential `Cannot access "contains" on undefined` error here by not verifying that the node is an element. I just opted to keep the current implementation behaviors and only add types before checking in with you guys. Pretty sure there are multiple other instances of this in the codebase

Gururajj77: Thanks for the suggestion. @imp-dance, could you please make this change?

imp-dance: @Gururajj77 Done ✅

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -310,3 +388,6 @@ TreeView.propTypes = {
    */
   size: PropTypes.oneOf(['xs', 'sm']),
 };
+
+TreeView.TreeNode = TreeNode;
```
tay1orjones, line 392: Would you mind adding a test to ensure `<TreeView.TreeNode>` works?

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -136,29 +194,36 @@ export default function TreeView({
...
-    let nextFocusNode;
+    let nextFocusNode: Node | null = null;
 
     if (match(event, keys.ArrowUp)) {
       nextFocusNode = treeWalker.current.previousNode();
     }
     if (match(event, keys.ArrowDown)) {
       nextFocusNode = treeWalker.current.nextNode();
     }
+    // @ts-ignore - `matches` doesn't like the object syntax without missing properties
```
tay1orjones, line 217: I think this is fine since we haven't explicitly typed this helper yet.

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -5,23 +5,95 @@
...
+  onToggle?: (event: React.MouseEvent, node?: TreeNodeProps) => void;
+  /**
+   * Callback function for when any node in the tree is selected
+   */
+  onTreeSelect?: (event: React.MouseEvent, node?: TreeNodeProps) => void;
+  /**
+   * Optional prop to allow each node to have an associated icon.
+   * Can be a React component class
+   */
+  renderIcon?: CarbonIconType;
```
tay1orjones: This is too narrow, it should accept any class or functional component. Throughout the codebase we've been doing `ComponentType | FunctionComponent` or `React.ElementType`. Either should resolve the ts errors on the proptypes I believe.

tay1orjones (approved):
Thanks for taking this on! This looks good to me, just a couple notes

packages/react/src/components/TreeView/TreeNode.tsx
imp-dance: Fixed!

packages/react/src/components/TreeView/TreeView.tsx
imp-dance: Done, I just copy/pasted `should render children as expected` into `should render children as expected when using dot syntax` and refactored it to use `<TreeView.TreeNode>` instead of `<TreeNode>`.

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -32,12 +110,16 @@ const TreeNode = React.forwardRef(
...
-      selected,
+      selected: propSelected,
       value,
       ...rest
     },
     ref
   ) => {
+    // These are provided by the parent TreeView component
+    const depth = propDepth as number;
+    const selected = propSelected as (string | number)[];
```
imp-dance, line 121: I have another proposal.

Since the component internals expect that these should always be defined, should we do something like this to provide a better error for consumers that try to mount TreeNode outside of TreeView?

```typescript
if (depth === undefined || selected === undefined){
   throw new Error("It looks like you are trying to mount a TreeNode outside of a TreeView. To do this, ensure that the 'selected' and 'depth' props are provided.");
}
```

This would also mean that we wouldn't have to cast the types manually either.

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -26,18 +78,21 @@ export default function TreeView({
...
   const enableTreeviewControllable = useFeatureFlag(
     'enable-treeview-controllable'
   );
 
   const { current: treeId } = useRef(rest.id || uniqueId());
   const prefix = usePrefix();
   const treeClasses = classNames(className, `${prefix}--tree`, {
+    // @ts-ignore - will always be false according to prop types
     [`${prefix}--tree--${size}`]: size !== 'default',
   });
```
imp-dance, line 91: I also found this line of code a bit strange, considering `size` should only be `xs` or `sm` (according to types).

Gururajj77 (approved):
Looks good, thanks for the contribution

PR #17039, opened 2024-07-24
Removed "Test" stories from components.

guidari:
Removed "Test" stories from components.

Also, it is a good pratice to name our "Test" stories always as "Test", so it is easier to catch when we search in the [storybook](https://react.carbondesignsystem.com/?path=/story/elements-grid--sub-grid-test-story) if we forget to remove before merging a PR.


alisonjoseph (approved):
thanks for cleaning these up!

riddhybansal (approved):

PR #17047, opened 2024-07-26
fix: focus issue in menu

### LINKED ISSUE #16994: [Bug]: Inconsistent focus for menu if more than one menu item have nested menu items.
### Package

@carbon/react

### Browser

_No response_

### Package version

1.61.0

### React version

_No response_

### Description

In Menu component if we have more than one item with nested items in it, the focus(Highlights) become inconsistent. 

### Reproduction/example

Please check attached video

### Steps to reproduce

https://github.com/user-attachments/assets/0501fab7-5d8a-4c57-a29b-af789a81b2d9

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

preetibansalui:
Closes #16994 

Fix the inconsistency in menu where it has more than 2 menu items with submenus.

**Changed**

The issue was causing because of delay in [closing the submenu](https://github.com/carbon-design-system/carbon/pull/16466). 

Added floating ui - safePolygon so menu will not close when user hover on submenu.
Added floating ui - FloatingFocusManager to manage the focus.


guidari (approved):
LGTM! 🚀

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -97,26 +102,40 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
...
+      placement: rtl ? 'left-start' : 'right-start',
+      whileElementsMounted: autoUpdate,
+      middleware: [offset({ mainAxis: -6, crossAxis: -6 })],
+    });
+    const { getReferenceProps, getFloatingProps } = useInteractions([
+      useHover(floatingContext, {
+        enabled: true,
+        handleClose: safePolygon({
+          requireIntent: true,
+        }),
```
tay1orjones: Here I'm trying to get to `Italic`, but can't get there even though I'm in the triangle. The `Share with` item is being invoked when I expect it to be ignored. Maybe the [`blockPointerEvents` option](https://floating-ui.com/docs/usehover#blockpointerevents) could help?


https://github.com/user-attachments/assets/7aebcc5a-a9ce-41e6-b94d-f3c2058f0dd1

packages/react/src/components/OverflowMenu/next/OverflowMenu-test.js
```diff
@@ -99,7 +97,11 @@ describe('OverflowMenu (enable-v12-overflowmenu)', () => {
     );
     await userEvent.type(screen.getByRole('button'), 'enter');
     expect(screen.getByRole('button')).toHaveAttribute('aria-expanded', 'true');
-    await userEvent.click(document.body);
+
+    await act(async () => {
+      await userEvent.click(document.body);
+    });
```
tay1orjones, line 119: Is this for pulling focus out of the menu? I didn't think calls to `userEvent` needed to be wrapped in `act()`.

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -136,32 +155,19 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
...
-      setBoundaries({ x: -1, y: -1 });
     }
 
     function handleClick(
       e: KeyboardEvent<HTMLLIElement> | MouseEvent<HTMLLIElement>
     ) {
       if (!isDisabled) {
         if (hasChildren) {
+          setSubmenuOpen(true);
           openSubmenu();
```
tay1orjones: `setSubmenuOpen` is already called inside `openSubmenu`

```suggestion
          openSubmenu();
```

tay1orjones (changes requested):
This is such an excellent improvement! Just a couple things I noticed

packages/react/src/components/Menu/MenuItem.tsx
preetibansalui: [blockPointerEvents](https://floating-ui.com/docs/usehover#blockpointerevents) was blocking the entire menu from any event.  its happening only when any other pointer comes under the safePolygon. If you try it with` Share with ` It works fine. So I added a `delay` of 100ms to avoid the issue.

tay1orjones (approved):

PR #17048, opened 2024-07-26
fix(16865): fixes step operations in NumberInput

### LINKED ISSUE #16865: [Bug]: Unexpected Increment Behavior in NumberInput Component with Step Prop
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.60.2

### React version

v20.15.0

### Description

When using the NumberInput component with a step prop, the increment behavior is not as expected. Specifically, when the default value is set to "1000" and the step prop is set to 1000, pressing the "+" button should increment the value to "2000". However, the value first increases to "1001" instead. Upon pressing the "+" button again, the value increments to "2001".

Also, the storybook crashes when trying to replicate the same in storybook

**Expected Behavior:**

- Default value: 1000
- step prop value: 1000
- First press of the "+" button should set the value to 2000.
- Subsequent presses should continue to increment the value by 1000.

**Actual Behavior:**

- Default value: 1000
- step prop value: 1000
- First press of the "+" button sets the value to 1001.
- Second press sets the value to 2001.

### Reproduction/example

https://stackblitz.com/edit/github-mrrrn4-kvfdxe?file=src%2FApp.jsx

### Steps to reproduce

1. Set up a NumberInput component with a default value of "1000".
2. Add a step prop with a value of 1000.
3. Press the "+" button.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/16865

This PR fixes the Inconsistent Increment/Decrement behavior in NumberInput Component when value and step are  large 
The usage of native `stepUp()` and `stepDown()` methods is removed 
Added test case to verify the behaviour when value and step are large 

#### Changelog

`stepUp()` and `stepDown()` methods are removed and  Increment/Decrement operations are handled inside our own function

I suspect that native `stepUp()` and `stepDown()` methods can be unpredictable with large step values with large values
When the current value (e.g., 1000) matches the step size (e.g., 1000), it seems like browsers treat it as a step boundary. and only increment by just 1 instead of the full step

#### Testing / Reviewing

Test NumberInput component with below props and any other combination:  verify it works as expected 
```
// large values
 <NumberInput
        id="carbon-number"
        min={1}
        max={9900}
        value={1000}
        step={1000}
      />
```
      
```
//allowEmpty and step=0
  <NumberInput
      id="carbon-number"
      min={-100}
      max={100}
      value={''}
      step={0}
      allowEmpty
      
    />
```
```
//allowEmpty and step=10
<NumberInput
      id="carbon-number"
      min={-100}
      max={100}
      value={''}
      step={10}
      label="NumberInput label"
      helperText="Optional helper text."
      invalidText="Number is not valid"
    />
```


guidari (commented):
Is it just on my end or it is having performance issues? The screen freezes sometimes and then crash.
Screenshot 2024-08-13 at 09 44 19

2nikhiltom:
Hey @guidari 
This is interesting and happening on my system too. It can be seen on [Production](https://react.carbondesignsystem.com/?path=/story/components-numberinput--playground) as well.
Note: Happens on Playground story only,  when `Steppers` (+/-)are clicked rapidly. Should we open a separate bug for this?

guidari (approved):
LGTM! 
Let's open a new issue for the the performance then!

alisonjoseph (approved):

PR #17055, opened 2024-07-29
Fix types for translateWithId() methods.

### LINKED ISSUE #17054: translateWithId(id) methods list wrong type for id
### The problem

Many of the `translateWithId()` methods are listed as taking a `string` as the first parameter.  But those functions actually take something more like an enum.   So with the current Carbon, this app code fails:

```tsx
const comboboxTranslation = (
  id: "close.menu" | "open.menu" | "clear.all" | "clear.selection"
) => { ... }

...

<ComboBox ... translateWithId={comboboxTranslation}
```

### The solution

Declare translateWithId() with proper type for id aka key.

### Examples

_No response_

### Application/PAL

IKC

### Business priority

None

### Available extra resources

I'll make the PR

### Code of Conduct

wkeese:
Closes #17054

Add proper types for parameter to translateWithId() methods.
Each component's translateWithId() method can only take a certain
list of keys, not any string.  Update the Typescript to notate this.

Also, update components to extend InternationalProps rather than defining translateWithId() directly.

For historical reasons most of the existing translation keys are declared in a convoluted way.  For example:
    
```ts
export const translationIds = {
      'increment.number': 'increment.number',
      'decrement.number': 'decrement.number',
};
 
type TranslationKey = keyof typeof translationIds;
```
  
The simpler way is just:

```ts    
type TranslationKey = 'increment.number' |  'decrement.number’;
```
    
I didn’t update that in this PR, but it’s something to consider for the future.
    
#### Changelog

**Changed**

- Declarations of translateWithId() methods

#### Testing / Reviewing

I tested the changes against my app and it now compiles without any casts needed.

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/react/src/components/ComboBox/ComboBox.tsx
@@ -45,6 +45,7 @@ import deprecate from '../../prop-types/deprecate';
 import { usePrefix } from '../../internal/usePrefix';
 import { FormContext } from '../FluidForm';
 import { useFloating, flip, autoUpdate } from '@floating-ui/react';
+import { InternationalProps } from '../../types/common';
 
 const {
   InputBlur,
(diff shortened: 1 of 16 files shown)


tay1orjones (approved):
Thanks for doing this! Great improvement. 

The `InternationalProps` name has bothered me since it was added. I don't think we're going to have other props to put inside there to warrant a broad name like that. I just pushed an update to rename it to `TranslateWithId`. I think it'll help with readability too - you see exactly whats going on at a glance, without even using intellisense.

> For historical reasons most of the existing translation keys are declared in a convoluted way

40% of our consumers aren't using TypeScript. I think we'd probably need to maintain their access to the id/key values until the next major. I'll admit though that I'm not sure how many are actually using the exported `translationIds`.

guidari (approved):

PR #17056, opened 2024-07-29
fix(17041): use react-fast-compare to compare props update in DataTable

### LINKED ISSUE #17041: [Bug]: firefox is slow bug by the lodash.isEqual in DataTable with react component children in rows.
### Package

@carbon/react

### Browser

Firefox

### Package version

v1.60.3

### React version

v17.0.2

### Description

We create a the <DataTable> with various cell data with the react component, and we found in firefox, sometimes the firefox will be slow because DataTable is calling lodash.isEqual like:  https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/DataTable.tsx#L426, and after we added some dump console.log statement, it will tell us:
after isEqual(nextProps.rows, this.props.rows), it is elapsed for 150652 milli-seconds

it's not always there, but it looks firefox will protect itself if firefox javascript engine call stack size exceed some threshold count. Our component has some redux dispatch calls and it will notify the react to re-render the component, and mostly it happens intermittent and we do not find this issue in chrome.

From https://github.com/facebook/react/issues/19811, we should not use Lodash.isEqual to compare the object might contain react component, there are couple of replacement like fast-deep-equal or react-fast-compare package is doing better than Loash.isEqual.

### Reproduction/example

it's a little bit hard to reproduce in a simple app

### Steps to reproduce

Create a component with <DataTable> with rows containing react component, and move away from the react route for the component and move back to this component, this component will be mount/unmount for a couple of times.

then the firefox will be freezing and prompt to stop or debug script..

Add some debug code around https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/DataTable.tsx#L426 like:
 const beforeCompare = Date.now();
      if (!isEqual(nextProps.rows, this.props.rows)) {
        const elapsed = Date.now() - beforeCompare;
        console.log(`after isEqual(nextProps.rows, this.props.rows), it is elapsed for ${elapsed} milli-seconds`);

then it will print in debug console:
after isEqual(nextProps.rows, this.props.rows), it is elapsed for 150652 milli-seconds

it's not always there, and only happens in firefox.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

yaoterry:
Closes #17041 

This is use react-fast-compare to compare the updated properties  in DataTable's shouldComponentUpdate method, use lodash.isEqual will hurt performance issue from: https://github.com/facebook/react/issues/19811

#### Changelog

**New**

- {{new thing}}

**Changed**

- replace the lodash.isequal with react-fast-compare to improve performance, otherwise it will break the DataTable re-render unsafely under some circumstance.

**Removed**

- {{removed thing}}

#### Testing / Reviewing

this is just a 3rd party package replacement, we just need to make sure the DataTable can be rendered properly.


riddhybansal:
@tay1orjones since you were there in the conversation of this issue, What do you think this optimise the performance? If yes, should we try this with other components as well, i.e, using isEqual for element comparing in components like multiselect, filterable multiselect. Because there are issues we can generally see with firefox in many components.

tay1orjones (approved):

tay1orjones:
@riddhybansal Yeah this looks great, I think we can use it anywhere we use lodash's isEqual.

riddhybansal (approved):
Hey thanks for contributing !! This is awesome 🚀

PR #17059, opened 2024-07-29
fix: added rel prop in switcher item

### LINKED ISSUE #17052: [Bug]: TS error: rel prop does not exist in SwitcherItem type
### Package

@carbon/react

### Browser

Chrome

### Package version

1.62.1

### React version

18.3.1

### Description

Two nearly identical errors were fixed (#15510 and #16294) for this component for two other props, and once again another anchor prop, `rel`, is now flagging as a TypeScript compile-time error (it wasn't in Carbon 1.56). 

image

### Reproduction/example

Any .tsx file that uses the rel attr with SwitcherItem will show this, no working example needed (compile-time issue, not run-time)

### Steps to reproduce

Any .tsx file that uses the rel attr with SwitcherItem will show this, no working example needed (compile-time issue, not run-time)

### Suggested Severity

### Application/PAL

CDD Hub

### Code of Conduct

riddhybansal:
Closes #17052 

Added optional 'rel' prop to `BaseSwitcherItemProps`.

**New**
Added rel prop to `BaseSwitcherItemProps`

#### Testing / Reviewing
```tsx
<Switcher aria-label="Switcher Container" expanded={true}>
  <SwitcherItem aria-label="Link 1" href="#" rel="norelerror">
    Link 1
  </SwitcherItem>
</Switcher>
```

Test the usage of `SwitcherItem` with the `rel` prop in a TypeScript environment within a `Switcher` component. Verify that the  TypeScript compiler should not throw error and the component renders without errors and correctly applies the provided `rel` attribute

### DIFF
--- a/packages/react/src/components/UIShell/SwitcherItem.tsx
+++ b/packages/react/src/components/UIShell/SwitcherItem.tsx
@@ -58,6 +58,10 @@ interface BaseSwitcherItemProps {
    * Specify where to open the link.
    */
   target?: HTMLAttributeAnchorTarget;
+  /**
+   * The rel property for the link.
+   */
+  rel?: string;
 }
 
 interface SwitcherItemWithAriaLabel extends BaseSwitcherItemProps {
@@ -89,6 +93,7 @@ const SwitcherItem = forwardRef<ElementType, SwitcherItemProps>(
       onKeyDown = () => {},
       href,
       target,
+      rel,
       ...rest
     } = props;
 
@@ -132,6 +137,7 @@ const SwitcherItem = forwardRef<ElementType, SwitcherItemProps>(
           }}
           href={href}
           target={target}
+          rel={rel}
           ref={forwardRef}
           {...rest}
           className={linkClassName}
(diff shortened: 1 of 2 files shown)


packages/react/src/components/UIShell/SwitcherItem.tsx
```diff
@@ -58,6 +58,10 @@ interface BaseSwitcherItemProps {
    * Specify where to open the link.
    */
   target?: HTMLAttributeAnchorTarget;
+  /**
+   * The rel property for the link.
+   */
+  rel?: string;
```
tay1orjones, line 64: Is there a specific type off of HTMLAttributes or DOMAttributes or something that we could use here? It's not any string - it's one of a specific set of allowed/expected strings, right?

[MDN's list](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) is "some of the most important existing keywords" 🤔 very possible this doesn't exist but figured I'd ask

tay1orjones (approved):
Looks good! Just one question

alisonjoseph (approved):

packages/react/src/components/UIShell/SwitcherItem.tsx
riddhybansal: Yes we cannot add any string there are some specific keywords that can be used . What do you think should I create an interface and made rel at that type ??

PR #17072, opened 2024-07-31
fix(16740): DatePicker calendar close issue on clickAway, DatePicker not running onChange events, DatePicker setting and clearing date in range issues 

### LINKED ISSUE #16710: [Bug]: DatePicker is not running onChange after typing a value and clicking away
### Package

@carbon/react

### Browser

Firefox

### Package version

1.59.0

### React version

18.2.0

### Description

Something has changed with the DatePicker control since an earlier release (1.49.0 I believe, or thereabouts). It is evident when the `allowInput` property is set to true, thus allowing the user to type a date, as well as being able to choose it from the dropdown calendar.

The `onChange` callback for the DatePicker used to be called either when a new date was picked from the calendar, or when the date input lost focus. In the latest Carbon versions, it now appears that `onChange` is not called when the input loses focus, unless the loss of focus is due to the user tabbing away using the keyboard; if the control loses focus by the user clicking away from it with their mouse, `onChange` does not get called.

This means that an up to date value of date can be missed, or invalid values (caused when the user types something illegal) are not recognised and cannot be coped with.

We cannot use `onChange`/`onBlur` from the DateInputField as a workaround. Although we can use this to detect a valid typed date, we cannot use it to detect an invalid date. If an invalid date is typed in (say, "xxxx"), in the onBlur all we will know is that the user has typed "xxxx"; however, flatpickr will have automatically cleared this date out on leaving the field, due to it being invalid, and hence the input value we have in state ("xxxx") will not match the actual text being shown in the component (empty). 

Ultimately, `onChange` for the whole DatePicker should be the "go-to" handler for handling all new values, be they chosen from the dropdown, or typed by the user into the input field, and this is not working.

I would class this as a Sev 2 defect, because this basically stops us being able to accept manual date input from the user: the only reliable behaviour of the component is that experienced when dates are chosen from the dropdown calendar.

### Reproduction/example

https://stackblitz.com/edit/github-sf1rjr-jgpsi5?file=src%2FApp.jsx

### Steps to reproduce

If you choose a date from the DatePicker dropdown you will notice that the "Date picker date" correctly shows the chosen date.

If you then enter the date input field and type in a new date, and then click into the white area of the page in order to shift focus away from the DatePicker, you will notice that the "Date input date" shows the text that was entered, but the "Date picker date" is still showing the previously chosen date". This is the defect: an up to date value of date, as typed in by the user, is missed.

By comparison though, if you enter the date input field and type in a new date but then press the TAB key to leave the control, the "Date picker date" _is_ updated to show the new date. 

It seems that `onChange` for the DatePicker is not being called when a free text date is entered and focus is lost due to mouse action rather than keyboard action.

### Suggested Severity

### Application/PAL

IBM CloudPak for AIOps

### Code of Conduct

### LINKED ISSUE #16740: [Bug] [DatePicker]: End Date Calendar in Range variant remains Open on click away

### LINKED ISSUE #16991: [Bug]: Datepicker issue with Keyboard Navigation
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v1.62.0

### React version

v18.3.1

### Description

With keyboard navigation able to select dates but the focus is not going to the next elements , the focus is always trapping inside date picker end date

### Reproduction/example

https://stackblitz.com/edit/github-pjo2fh-rm6wjm?file=src%2FApp.jsx

### Steps to reproduce

Select the dates using keyboard after selection of end date press tab doesn't work (focus not going to the next element).

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

### LINKED ISSUE #17071: [Bug]: DatePicker setting and clearing date in range issues
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.62.1

### React version

v17.0.2

### Description

There is a problem with DatePicker component.
- after selecting date range if user clicks outside date component calendar is opened and can't be closed (issue can be reproduced on carbon's storybook too),
- when user clears date range start date input is not cleared,

https://github.com/user-attachments/assets/7ba96c66-3e29-4159-815d-0f82a14f8820

### Reproduction/example

https://stackblitz.com/edit/vitejs-vite-bngjrz

### Steps to reproduce

- create DatePicker with range and calendar
- select range
- click somewhere outside component

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes #16740 
Closes https://github.com/carbon-design-system/carbon/issues/16710 
Closes https://github.com/carbon-design-system/carbon/issues/17071
Closes https://github.com/carbon-design-system/carbon/issues/16991

 
#### Changelog

(a) Sets up a `mousedown` event listener on the document.
(b) Addes a  `handleMouseDown` function which checks if any click is made outside the `calendar`(icon) and `InputFields`.
  - If click is made outside `closeCalendar` function is called immediately
  - (which basically closes the flat picker instance `calendarRef.current.close()`  and removes the focus from inputField)
  - Adds event type `clickOutside` while calling `onCalendarClose` (**to capture it and return without processing** )
 
(c) Deprecates the use of setTimeout 

- Moved the logic from inside of `setTimeout` to  `handleCalendarClose` function.
-  a new `useEffect` is add to call  `handleCalendarClose` whenever calendar closes 

  ```
 useEffect(() => {
    if (calendarCloseEvent) {
      const { selectedDates, dateStr, instance } = calendarCloseEvent;
     //handleCalendarClose handles the core logic of setTImeout 
      handleCalendarClose(selectedDates, dateStr, instance);
      setCalendarCloseEvent(null);
    }
  }, [calendarCloseEvent, handleCalendarClose]);
```

Would love some review on usage of this interface 
```
 interface CalendarCloseEvent {
    selectedDates: Date[];
    dateStr: string;
    instance: object; //This is `Intance` of flatpicker
  }
```

#### Testing / Reviewing

Open deploy preview and visit all the storybook examples 

Each has demonstration on existing bugs fix (make sure calendar closes on clickaway, dates are rested on clear option, onchange events happens as expected  )
Storybook examples needs to be restored or older version before merging

### CODE REVIEW

File: packages/react/src/components/DatePicker/DatePicker.tsx [hunk 1]
```diff
@@ -444,31 +444,38 @@ const DatePicker = React.forwardRef(function DatePicker(
...
-          calendarRef.current.config.dateFormat
-        );
-      }
-      if (onClose) {
-        onClose(
-          calendarRef.current.selectedDates,
-          dateStr,
-          calendarRef.current
-        );
+  const onCalendarClose = useCallback(
```
tay1orjones: `savedOnClose` is already wrapping `onCalendarClose` in a useCallback. Could the new useCallback be removed?

packages/react/src/components/DatePicker/DatePicker.tsx
```diff
@@ -653,7 +661,7 @@ const DatePicker = React.forwardRef(function DatePicker(
           savedOnChange(...args);
         }
       },
-      onClose: savedOnClose,
+
```
tay1orjones, line 671: This is the only spot `savedOnClose` is used directly, but it's also used in the dependency array of the huge useEffect this is all wrapped in.

tay1orjones (changes requested):
The storybook interaction looks good, but I'm concerned this change as is could have some unintended effects due to how the code is modifying how `savedOnClose` is used.

guidari:
Hey Nikhil, looks like the overview page is not loading
Screenshot 2024-08-02 at 10 04 27

guidari (approved):
I pulled down locally to test and looks good!
You can also close this issue https://github.com/carbon-design-system/carbon/issues/16991, It's working fine now!
🚀

emyarod (approved):
looks good to me! I think we can update the stories to remove the console.logs now

riddhybansal (approved):
Great Work !! 🚀

tay1orjones (approved):

PR #17085, opened 2024-08-01
16879 detectoverflow configurable collision boundary

### LINKED ISSUE #16879: Floating UI: Add `detectOverflow` configurable collision boundary

preetibansalui:
Part of #16879 

Added `hide` from Floating-ui to Dropdown component so it will hide the floating dropdown if the reference (Dropdown Trigger point) is not visible.

#### Testing / Reviewing

1. Check the ExperimentalAutoAlignBoundaryTest Story in Dropdown. 
2. Open the dropdown and scroll it, the moment 'Option 1' hide from the screen, it should hide the floating dropdown options as well.


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -299,13 +300,31 @@ const Dropdown = React.forwardRef(
...
+                ? 'hidden'
+                : 'visible',
+            });
+        }
+      });
+    };
+
     useEffect(() => {
       if (autoAlign) {
         Object.keys(floatingStyles).forEach((style) => {
```
tay1orjones: If we get the middleware working with the hook, the visibility style could be applied by adding it to `floatingStyles` here before we iterate over it.

```suggestion
      if (autoAlign) {
        Object.keys({...floatingStyles, visibility: middlewareData.hide?.referenceHidden ? 'hidden' : 'visible'}).forEach((style) => {
```

Or maybe on the preceding line so it's less dense

tay1orjones (commented):
Did we want to try and see if we can use the `hide()` middleware directly with the useFloating hook? https://floating-ui.com/docs/hide

preetibansalui:
> Did we want to try and see if we can use the `hide()` middleware directly with the useFloating hook? https://floating-ui.com/docs/hide

Hey @tay1orjones, Thanks for the suggestion. It works fine in useFloating hook as well so kept it in hook to avoid extra code.

guidari (approved):
Awesome! LGTM! 🚀

tay1orjones (approved):
LGTM once the test story is removed

PR #17088, opened 2024-08-01
fix(useId): make react 18 import impossible to statically analyze

### LINKED ISSUE #17084: [Bug]: useId giving warning as it's not supported by react <18
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.63.0

### React version

16.14.0

### Description

A recent PR - https://github.com/carbon-design-system/carbon/pull/16988, which was part of the `v1.63.0` release of `@carbon/react` has moved to using `useId` in more places. There is handling for the correct version here https://github.com/carbon-design-system/carbon/blob/f22da5a696bd891a58aa0fc4c81f0fbe0030ba5d/packages/react/src/internal/useId.js#L26 but this seems to not be working since that PR. 

When running a storybook-smoke-test (`npm run storybook-smoke-test`) we are seeing:
```
WARN export 'useId' (imported as 'useId') was not found in 'react' (possible exports: Children, Component, Fragment, Profiler, PureComponent, StrictMode, Suspense, __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED, cloneElement, createContext, createElement, createFactory, createRef, forwardRef, isValidElement, lazy, memo, useCallback, useContext, useDebugValue, useEffect, useImperativeHandle, useLayoutEffect, useMemo, useReducer, useRef, useState, version)
```

I believe the initial PR to fix this was https://github.com/carbon-design-system/carbon/pull/14153, and I can't see that file being changed since so it's not immediately obvious why it's no longer working. 

### Reproduction/example

react@16.14.0, @carbon/react@1.63.0

### Steps to reproduce

- Create an environment with react `16.14.0`, `@carbon/react@v.1.63.0` and a storybook set up.
- Run `npm run storybook-smoke-test` and observe warning about `useId`

### Suggested Severity

### Application/PAL

Cloud Pak For Integration (Platform UI)

### Code of Conduct

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/17084

#### Changelog

**Changed**

- Further obfuscated the `React` import based on the result of explorations in https://github.com/mui/material-ui/issues/41190#issuecomment-2040873379
- Reorganized the code and variable names a bit so it's a bit more understandable

#### Testing / Reviewing

- Snapshots should not need updated, meaning id's should be generated the exact same as they were before


guidari (approved):
I just tested locally and it is working fine! 🚀

emyarod (approved):
looks good to me!

PR #17089, opened 2024-08-01
fix(dropdown): spread `dropdownProps` last

### LINKED ISSUE #17027: [Bug]: ComboBox downshiftProps missing reference to Downshift component where state change can be done
### Package

@carbon/react

### Browser

Chrome

### Package version

1.62.1

### React version

17.0.2

### Description

ComboBox has downshiftProps where we we used to utilize onStateChange function to make any change directly to Downshift Component by changing the state. Previously onStateChange used to receive Downshift component as 2nd argument, but now there is no way to receive the component, it is always undefined.

### Reproduction/example

https://stackblitz.com/edit/github-v56j87?file=src%2FApp.jsx

### Steps to reproduce

1. Select one Item from ComboBox
2. Now try to clear the selection using close icon, it is not getting cleared.
3. Now if you try to open ComboBox and now try doing clear, it will get cleared.

So basically the issue is, when item is selected and ComboBox is closed (not expanded), then item not getting cleared when you click close icon.

### Suggested Severity

### Application/PAL

Power Virtual Server

### Code of Conduct

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/17027

#### Changelog

**Changed**

- Update downshift-based components to spread `...dropdownProps` last so that consumers can fully configure downshift's behavior
- Update docs and prop/types comments to more clearly state this and the associated pitfalls

#### Testing / Reviewing

- Review downshift-based components in storybook, they should be unchanged

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.mdx
+++ b/packages/react/src/components/ComboBox/ComboBox.mdx
@@ -54,6 +54,11 @@ next element.
 Our `Combobox` component utilizes [Downshift](https://www.downshift-js.com/)
 under the hood to help provide complex yet accessible custom dropdown
 components. We provide access to the built in Downshift features with this prop.
+
+Use with caution: anything you define here overrides the components' internal
+handling of that prop. Downshift internals are subject to change, and in some
+cases they can not be shimmed to shield you from potentially breaking changes.
+
 For more information, checkout the Downshift prop
 [documentation](https://www.downshift-js.com/downshift#props-used-in-examples)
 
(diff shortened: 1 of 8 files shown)


guidari (approved):

emyarod (approved):
looks good to me!

ashishkrz:
[FIX NOT WORKING] @tay1orjones Just to confirm what is changed now? I see that the position of spreaded downshiftProps got changed but that will not fix this issue, I believe. Still the 2nd argument to onStateChange function remains undefined.

Could you please help over stackblitz https://stackblitz.com/edit/github-v56j87?file=src%2FApp.jsx making it work with the fix done?

Issue #17095, opened 2024-08-02
[Bug]: @carbon/react 1.63.0 breaks React 17 compatibility


wkeese:
### Package

@carbon/react

### Browser

Chrome

### Package version

1.63.0

### React version

17

### Description

PR #17015 breaks React 17 compatibility by directly referencing Carbon's used() in TableSelectRow.tsx

TableSelectRow.tsx should be importing the polyfill like this instead:

```js
import { useId } from '../../internal/useId.js';
```

Note that Carbon 11 claims to support React 17, see https://github.com/carbon-design-system/carbon/blob/main/packages/react/package.json#L44.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datatable-selection--default

### Steps to reproduce

Build against React 17 and run app

### Suggested Severity

None

### Application/PAL

WXD

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

gi:
#17098 should have fixed the `useId` issue

wkeese:
Thanks, looks good, I see the fix in 1.63.2.

PR #17098, opened 2024-08-04
react: fix usage of useId

gi:
This updates a couple of react package components and their usage of the `useId` function.

#### Changelog

**Changed**

- `TableSelectRow`: import `useId` from internal polyfill package instead of directly from React
- `TableToolbarSearch`: pass prefix to `useId` to avoid null value

#### Testing / Reviewing

Existing unit tests

### DIFF
--- a/packages/react/src/components/DataTable/TableSelectRow.tsx
+++ b/packages/react/src/components/DataTable/TableSelectRow.tsx
@@ -6,10 +6,11 @@
  */
 
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


preetibansalui (approved):

preetibansalui:
Fix for #[17095](https://github.com/carbon-design-system/carbon/issues/17095)

tay1orjones (approved):
Thanks, this looks great!

PR #17103, opened 2024-08-05
fix(PasswordInput): tooltip issue in show password label or hide password label

### LINKED ISSUE #16924: [Password-input] Tooltip issue on hide password icon hover
### Package

carbon-components

### Browser

_No response_

### Package version

1.61.0

### React version

18.2.0

### Description

In the password-input component, when the [show password functionality](https://react.carbondesignsystem.com/?path=/story/components-passwordinput--playground) is turned on and the hide password icon is hovered over, the tooltip does not display correctly.

Screenshot 2024-07-08 at 9 39 10 PM

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-passwordinput--playground

### Steps to reproduce

One of the developers on the team mentioned: "They noticed a condition where adding a blank value in **hidePasswordLabel** causes it to display like this. It might be because there have been recent changes to the alignment options in the tooltip.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

preetibansalui:
Closes #16924 

fixed broken tooltip if label is blank and added a console warning if label is blank.

**New**

 added a console warning if showPasswordLabel or hidePasswordLabel is blank.

### DIFF
--- a/packages/react/src/components/TextInput/PasswordInput.stories.js
+++ b/packages/react/src/components/TextInput/PasswordInput.stories.js
@@ -37,6 +37,8 @@ export const Playground = (args) => (
       labelText="Text input label"
       helperText="Optional help text"
       autoComplete="true"
+      hidePasswordLabel="sssfdfddcff dfdfdf"
+      showPasswordLabel=""
     />
   </div>
 );
--- a/packages/react/src/components/TextInput/PasswordInput.stories.js
+++ b/packages/react/src/components/TextInput/PasswordInput.stories.js
@@ -37,8 +37,6 @@ export const Playground = (args) => (
       labelText="Text input label"
       helperText="Optional help text"
       autoComplete="true"
-      hidePasswordLabel="sssfdfddcff dfdfdf"
-      showPasswordLabel=""
     />
   </div>
 );
--- a/packages/react/src/components/TextInput/PasswordInput.tsx
+++ b/packages/react/src/components/TextInput/PasswordInput.tsx
@@ -329,6 +329,11 @@ const PasswordInput = React.forwardRef(function PasswordInput(
   if (tooltipPosition === 'right' || tooltipPosition === 'left') {
     align = tooltipPosition;
   }
+  if (!hidePasswordLabel || hidePasswordLabel.trim() === '') {
+    console.warn('Warning: The "hidePasswordLabel" should not be blank.');
+  } else if (!showPasswordLabel || showPasswordLabel.trim() === '') {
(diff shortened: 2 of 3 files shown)


2nikhiltom:
Hey @preetibansalui 

Just a question : Should the tooltip be rendered on UI if the string is empty ? I see warning is console but the tooltip renders with no content.

Im curious of useCase where consumers would want to pass an empty string for hidePasswordLabel/ ShowPasswordLabel.
and Incase they do want an empty string, should we completely remove the tooltip itself ? 

image

Kritvi-bhatia17 (approved):
Looks good to me from design perspective! 🎉

preetibansalui:
> Hey @preetibansalui
> 
> Just a question : Should the tooltip be rendered on UI if the string is empty ? I see warning is console but the tooltip renders with no content.
> 
> Im curious of useCase where consumers would want to pass an empty string for hidePasswordLabel/ ShowPasswordLabel. and Incase they do want an empty string, should we completely remove the tooltip itself ?
> 
> image

Hi @2nikhiltom, you've raised a valid point.
1. While there's no purpose in providing an empty string for the tooltip, users are still able to enter any value in the label field. Therefore, we need to handle for the case of a blank string as well.
2. At this moment, we cannot make this field mandatory, as this would be a breaking change. However, we plan to designate it as a required field in version 12.

preetibansalui:
> > Hey @preetibansalui
> > Just a question : Should the tooltip be rendered on UI if the string is empty ? I see warning is console but the tooltip renders with no content.
> > Im curious of useCase where consumers would want to pass an empty string for hidePasswordLabel/ ShowPasswordLabel. and Incase they do want an empty string, should we completely remove the tooltip itself ?
> > image
> 
> Hi @2nikhiltom, you've raised a valid point.
> 
> 1. While there's no purpose in providing an empty string for the tooltip, users are still able to enter any value in the label field. Therefore, we need to handle for the case of a blank string as well.
> 2. At this moment, we cannot make this field mandatory, as this would be a breaking change. However, we plan to designate it as a required field in version 12.

Also, as per the a11y point of view, all clickable icons will have a tooltip. If no tooltip text is provided, a blank tooltip will be displayed.

2nikhiltom:
It makes sense from a11y point of view, thanks Preeti

2nikhiltom (approved):

emyarod (approved):
looks good to me!

riddhybansal (approved):

PR #17118, opened 2024-08-07
feat(combobox): expose downshift actions through downshiftActions ref prop

### LINKED ISSUE #17027: [Bug]: ComboBox downshiftProps missing reference to Downshift component where state change can be done

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/17027

Follow up to https://github.com/carbon-design-system/carbon/pull/17089, this ensures that downshift action functions can be called from within parent components using Combobox.

<details><summary>Click to expand notes on compatibility</summary>
<p>

The functionality behind both `downshiftProps` and `downshiftActions` isn't guaranteed and is subject to change. This PR patches a functionality mismatch internal to downshift. It does not cover all previous action:

| Actions previously available | Actions made available in this PR |
| ---------------------------- | --------------------------------- |
| clearSelection               |                                   |
| clearItems                   |                                   |
| closeMenu                    | closeMenu                         |
| openMenu                     | openMenu                          |
| selectHighlightedItem        |                                   |
| selectItem                   | selectItem                        |
| selectItemAtIndex            |                                   |
| setHighlightedIndex          | setHighlightedIndex               |
| toggleMenu                   | toggleMenu                        |
| reset                        | reset                             |
| setItemCount                 |                                   |
| unsetItemCount               |                                   |
| setState                     |                                   |
|                              | setInputValue                     |

If a consumer was using an action that is no longer available, such as `setState`, a functional equivalent can usually be found. This will require a code change in consuming projects.

```diff
- stateAndHelpers.setState({ inputValue: 'Item 1' });
+ downshiftActions.current.setInputValue('Item 1');
```

</p>
</details> 

#### Changelog

**New**

- Add `downshiftActions` prop to Combobox

**Changed**

- Update Combobox documentation for `downshiftActions` and `downshiftProps`
- Improve disclaimer warnings on both these props across all relevant downshift components
- Fix a bug in Combobox where the forwarded ref was not properly being applied through incorrect `mergeRefs` usage

#### Testing / Reviewing

The new test story in storybook implements the same logic referenced in the original issue report's [stackblitz reproduction](https://stackblitz.com/edit/github-v56j87?file=src%2FApp.jsx).

Go to the test story and validate base functionality. There should be no other changes to Combobox stories.

- [ ] ⚠️ Need to remove test story before merging

### DIFF
--- a/achecker.js
+++ b/achecker.js
@@ -10,7 +10,7 @@
 const path = require('path');
 
 module.exports = {
-  ruleArchive: 'latest',
+  ruleArchive: '17June2024',
   policies: ['Custom_Ruleset'],
   failLevels: ['violation'],
   reportLevels: [
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -1278,6 +1278,16 @@ Map {
       "disabled": Object {
         "type": "bool",
       },
+      "downshiftActions": Object {
+        "args": Array [
+          Object {
+            "current": Object {
+              "type": "any",
+            },
+          },
+        ],
+        "type": "exact",
+      },
       "downshiftProps": Object {
         "type": "object",
       },
--- a/packages/react/src/components/ComboBox/ComboBox.mdx
+++ b/packages/react/src/components/ComboBox/ComboBox.mdx
@@ -49,18 +49,64 @@ next element.
 <Combobox disabled />
 ```
 
-## Combobox `downshiftProps`
+## Combobox uses Downshift
 
 Our `Combobox` component utilizes [Downshift](https://www.downshift-js.com/)
 under the hood to help provide complex yet accessible custom dropdown
-components. We provide access to the built in Downshift features with this prop.
+components.
 
-Use with caution: anything you define here overrides the components' internal
-handling of that prop. Downshift internals are subject to change, and in some
-cases they can not be shimmed to shield you from potentially breaking changes.
+### `downshiftProps`
 
-For more information, checkout the Downshift prop
-[documentation](https://www.downshift-js.com/downshift#props-used-in-examples)
+`downshiftProps` is made available as a passthrough to the underlying downshift
+`useCombobox` hook.
+
+**Use with caution:** anything you define here overrides the components'
+internal handling of that prop. Downshift internals are subject to change, and
+in some cases they can not be shimmed to shield you from potentially breaking
+changes.
+
+For more information, checkout the Downshift `useCombobox` props
+[documentation](https://github.com/downshift-js/downshift/tree/v9.0.7/src/hooks/useCombobox#basic-props)
+
+### Combobox `downshiftActions`
+
+The downshift action methods are made available through the `downshiftActions`
+prop. While not recommended, this prop allows you to modify downshift's internal
+state without having to fully control the component.
+
+**Use with caution:** calling these actions modifies the internal state of
+downshift. It may conflict with or override the state management used within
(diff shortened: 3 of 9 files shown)


emyarod (approved):
the test story crashes when I interact with it

```sh
ReferenceError: component is not defined
    at Object.onStateChange (https://deploy-preview-17118--v11-carbon-react.netlify.app/ComboBox-stories.5248b449.iframe.bundle.js:1:29192)
    at callOnChangeProps (https://deploy-preview-17118--v11-carbon-react.netlify.app/1714.6bf0042b.iframe.bundle.js:1:31833)
    at https://deploy-preview-17118--v11-carbon-react.netlify.app/1714.6bf0042b.iframe.bundle.js:1:33956
    at Qj (https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:591606)
    at Hk (https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:611538)
    at Wk (https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:610306)
    at Pk (https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:610376)
    at Ek (https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:604186)
    at jg (https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:545214)
    at https://deploy-preview-17118--v11-carbon-react.netlify.app/7319.005706cc.iframe.bundle.js:528:601606
```

but changing `component` to `this` seems to fix the story and it looks like feature is working as expected

image

tay1orjones:
@emyarod ah my bad - I copied code from the stackblitz reproduction and forgot to update a couple of the usages from `component` to `changes` 🤦 It's fixed now

guidari (approved):
Cool! The stackblitz from this [issue](https://github.com/carbon-design-system/carbon/issues/17027) can console.log now
LGTM! 🚀

tay1orjones:
For posterity, here's the three approaches that I explored: 

1. A ref prop. The consumer passes in a ref that we mutate to contain the actions. It's what's in this PR.

2. A callback function prop. I don't think this would provide any benefit over using a ref. In both instances it's possible for the actions to be initially undefined, and they are only updated if the action functions change (since either approach is contained within a useEffect). Consumers would also need to store the values returned in the callback in state.

3. Use `useImperativeHandle` on forwardRef to expose the action functions. This would change the signature of the forwarded `ref` and would break anyone relying on `ref.current` to be a dom node. We would still expose the node in addition to the action functions, but it would require anyone using the ref to update their code. `ref.current` => `ref.current.inputNode`. 

Option 1 patches the bug, avoids consumers having to store anything in state, and doesn't introduce problems for projects using the forwardRef.

PR #17120, opened 2024-08-08
fix(16781): fixes Tab focus issue on accordion

### LINKED ISSUE #16781: [a11y]: Tab focus in accordion items incorrect, focusing on hidden items
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

@carbon/themes": "11.36.0

### React version

@carbon/react": "1.58.1

Manual testing

### Assistive technology

_No response_

### Description

As the user tabs through a Carbon accordion component, the focus moves towards the items inside the accordion, even if the accordion is not expanded. 

If there are many focusable items embedded within an expanded accordion, this could result in the focus being lost for quite awhile until the user finds the next focusable item.

### WCAG 2.1 Violation

2.4.3 Focus Order

### Reproduction/example

https://stackblitz.com/edit/github-rtptpp-wsfdvm?file=src%2FApp.jsx

### Steps to reproduce

Tab through the accordion and notice how the focus gets unfocused as the user tabs through. As you tab, the focus moves to the "link" inside the accordion, even though the accordion is closed. 

**Video demo:**
Recording

tabs

### Suggested Severity

### Code of Conduct

2nikhiltom:
Closes #16781

This PR prevents focus on reach onto collapsed according item, this usually happened when accordionItem was intractable element 

#### Changelog

**New**

Added `hidden={!isOpen} `attribute to wrapper div

**Changed**

Snapshot updated

#### Testing / Reviewing
Verify that when navigating accordion through Keyboard Tab navigation the focus does not reach on collapsed according item


packages/react/src/components/Accordion/AccordionItem.tsx
```diff
@@ -196,6 +196,7 @@ function AccordionItem({
       </Toggle>
       <div
         ref={content}
+        hidden={!isOpen}
```
kennylam, line 199: Would it be more appropriate to use `aria-hidden` here? Since `hidden` prevents the contents from being rendered at all, searches on collapsed content would not return any results. Although searching through collapsed, `aria-hidden="true"` content will not expand it, it will at least show a result in the browser.

kennylam: Actually, disregard this. I just tested and `aria-hidden` still allows tab focus even when `true`.

kennylam (approved):
LGTM!

tay1orjones (approved):
LGTM!

PR #17125, opened 2024-08-08
fix: 16882 Menu's target prop fix

### LINKED ISSUE #16882: [Bug]: MenuButton and OverflowMenu cannot configure the Menu's `target` prop
### Package

@carbon/react

### Browser

Chrome

### Package version

11.60.3

### React version

18

### Description

[Surfaced in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1719431119510709) 
MenuButton and OverflowMenu use Menu. Menu has a `target` prop for controlling what DOM element the `ul` is portaled into. This prop is not exposed through these components leaving consumers with a lack of control of where the menu is being rendered.

This needs to be exposed. I'd like to avoid prop drilling but in this case I don't think there's a viable alternative. It might be beneficial to call it `menuTarget` on these components.

### Reproduction/example

react.carbondesignsystem.com

### Steps to reproduce

Just take a look at the source, there's no `target` prop and `...rest` is spread onto the containing div.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

NabeelAyubee:
Closes #16882

Renamed the target prop to menuTarget for the Menu component to improve clarity.
Added the menuTarget prop to MenuButton and OverflowMenu components, allowing consumers to specify the DOM element into which the menu is portaled.

#### Changelog

**Changed**

- The target prop has been renamed to menuTarget for improved clarity.
- The menuTarget prop has been added to the MenuButton components to provide better control over menu rendering.
- Added support for backward compatibility
- Hide menuTarget prop from actions in storybook
- Added deprecation warning for target prop 

#### Testing / Reviewing

- Run yarn test to ensure that there are no code errors.
- Verify that the menuTarget prop correctly specifies the HTML element into which the menu is portaled. Ensure that the menu items are rendered in the target element as expected.
- Check the Storybook integration to ensure the new prop name is reflected and functioning correctly.


packages/react/src/components/Menu/Menu.tsx
```diff
@@ -88,7 +88,7 @@ interface MenuProps extends React.HTMLAttributes<HTMLUListElement> {
   /**
    * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
    */
-  target?: Element;
+  menuTarget?: Element;
```
tay1orjones, line 91: I think it's appropriate for this to just be called `target` still. Renames are breaking changes, so we'd have to leave the old one in and map it to the new one, put a deprecation warning on the old one, and write a codemod so folks can easily migrate.

packages/react/src/components/Menu/Menu.stories.js
```diff
@@ -45,10 +45,15 @@ export const Playground = (args) => {
   const selectableOnChange = action('onChange (MenuItemSelectable)');
   const radioOnChange = action('onChange (MenuItemRadioGroup)');
 
-  const target = document.getElementById('storybook-root');
+  const menuTarget =
+    document.getElementById(args.menuTarget) ||
+    document.getElementById('storybook-root');
```
tay1orjones: This can be removed, we don't typically expose element props like this within the actions pane.

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -80,6 +80,11 @@ export interface MenuButtonProps extends ComponentProps<'div'> {
    * Specify the tabIndex of the button.
    */
   tabIndex?: number;
+
+  /**
+   * Specify the menuTarget props of the menu.
+   */
+  menuTarget?: Element; // Add this line
```
tay1orjones:
```suggestion
  /**
   * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
   */
  menuTarget?: Element;
```

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -250,6 +257,11 @@ MenuButton.propTypes = {
    */
   // @ts-ignore-next-line -- avoid spurious (?) TS2322 error
   tabIndex: PropTypes.number,
+
+  /**
+   * Specify the menuTarget prop of the menu.
```
tay1orjones:
```suggestion
   * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
```

tay1orjones (changes requested):
Thanks for taking this on! I've left a few suggestions below. Two additional thoughts:

* `menuTarget` should also be added to the new/"next" OverflowMenu that is currently behind a feature flag as it uses Menu under the hood
* Menu's `target`, MenuButton's `menuTarget` and OverflowMenu's `menuTarget` should all be disabled in the actions pane in storybook, like how `children` is:

https://github.com/carbon-design-system/carbon/blob/6ef74406eda447bc80998ff2260beefe8a559048/packages/react/src/components/MenuButton/MenuButton.stories.js#L154-L157

packages/react/src/components/Menu/Menu.tsx
NabeelAyubee: I have added the target prop for backward compatibility issues. Added deprecation warning for target props.
Also added codemod for migration

packages/react/src/components/Menu/Menu.stories.js
NabeelAyubee: removed it also disabled menuTarget in actions as per our prior review

packages/react/src/components/MenuButton/index.tsx
NabeelAyubee: Added the changes

packages/react/src/components/MenuButton/index.tsx
NabeelAyubee: Added the changes

NabeelAyubee:
Hi, I have added the review changes. 

- Added target prop and showing deprecated warning for target props
- Added codemod for migrate
- Added props for overflow feature flag 
- disable menuTarget prop in action in storybook
- Review changes

NabeelAyubee:

nabeelmd-eGov:

NabeelAyubee:
recheck

PR #17133, opened 2024-08-09
Fixed renovate bot

guidari:
The `schedule` should go outside of the `extends`.

Inside the `extends` works just a few strings.

Screenshot 2024-08-09 at 11 27 26

If we want to move to weekday we should have set outside of `extends` and set under `schedule`

Screenshot 2024-08-09 at 11 28 04

A tricky detail to spot. That is why we haven't seen renovate PRs recently. Here it is the docs:
https://docs.renovatebot.com/presets-schedule/#scheduledaily:~:text=Schedule%20for%20typical%20non%2Doffice%20hours%20(night%20time%20and%20weekends).

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -11,10 +11,12 @@
 
     // https://docs.renovatebot.com/presets-npm/#npmunpublishsafe
     'npm:unpublishSafe',
-
-    // https://docs.renovatebot.com/presets-schedule/#scheduledaily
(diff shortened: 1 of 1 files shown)


emyarod (approved):
looks good to me!

tay1orjones (approved):
ooo great catch, I wondered about this when updating the accessibility-checker dep. Thanks!

PR #17135, opened 2024-08-09
fix(14831): update component and getter function types to match

### LINKED ISSUE #14831: [Bug]: Invalid TypeScript definition for multiple components (DataTable, Modal)
### Package

@carbon/react

### Browser

_No response_

### Package version

1.38.0

### React version

18.2.0

### Description

The typescript definition for the getSelectionProps DataTable does not match the JS implementation, which completely breaks strict builds when using the "select all" feature without manually disabling type-checking using comments or casts, which we shouldn't be encouraging.

In TypeScript the getSelectionProps is defined as:
```
getSelectionProps: (getSelectionPropsArgs: {
        onClick?: (e: MouseEvent) => void;
        row: DataTableRow<ColTypes>;
        [key: string]: unknown;
    }) => {
        ariaLabel: string;
        checked: boolean | undefined;
        disabled?: boolean | undefined;
        id: string;
        indeterminate?: boolean;
        name: string;
        onSelect: (e: MouseEvent) => void;
        radio?: boolean | null;
        [key: string]: unknown;
    };
```

Which causes type checking to error out when getSelectionPropsArgs is undefined, as is expected when using this in TableSelectAll as the examples do.

### Reproduction/example

N/A

### Steps to reproduce

Try to use TS in strict mode for the DataTable example code found in the storybook for selectable rows and it won't compile without errors.

But to fix it is super easy! Just need to add the missing optional question mark after getSelectionPropsArgs:

```
getSelectionProps: (getSelectionPropsArgs?: {
```

### Suggested Severity

### Application/PAL

IBM Internal

### Code of Conduct

## 2024 Update:
This is a deeper issue with the TypeScript support currently implemented, adding more details here to try and clean up these issues as I accidentally reported multiple related issues.

*See update here:* https://github.com/carbon-design-system/carbon/issues/13091#issuecomment-1989729035

According to TypeScript the props returned by the getHeaderProps function on data tables has a different onClick event handler type to the one expected by the header component.

Using the example from the storybook gives the following typescript error:
```
error TS2322: Type '{ children: ReactNode; isSortable: boolean | undefined; isSortHeader: boolean; key: string; onClick: (e: MouseEvent) => void; sortDirection: DataTableSortState; }' is not assignable to type 'TableHeaderProps'.
  Types of property 'onClick' are incompatible.
    Type '(e: MouseEvent) => void' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.

LINE: <TableHeader {...getHeaderProps({ header, isSortable: true })}>
```

**EDIT:**

This also appears to affect other parts of the DataTable components, I am now getting errors for the ToolBarSearch component onChange event handler having the wrong type when passing in the onInputChange prop from the render function directly.

```
 error TS2322: Type '(e: ChangeEvent<HTMLInputElement>, defaultValue?: string | undefined) => void' is not assignable to type '(event: "" | ChangeEvent<HTMLInputElement>, value?: string | undefined) => void'.
  Types of parameters 'e' and 'event' are incompatible.
    Type '"" | ChangeEvent<HTMLInputElement>' is not assignable to type 'ChangeEvent<HTMLInputElement>'.
      Type 'string' is not assignable to type 'ChangeEvent<HTMLInputElement>'.

LINE: <TableToolbarSearch onChange={onInputChange} />
```
 
Weirdly, the search bar error does not appear on stackblitz, it knows the signatures are different, but doesn't show any errors. But it also thinks that "'DataTable' cannot be used as a JSX component.", so I'm not sure if my example is working correctly on there, but "tsc" locally gives the above error.

EDIT2:

This also affects the TableExpandRow component too, the props don't match.
```
 error TS2322: Type '{ children: (Element | Element[])[]; ariaLabel: string; "aria-label": string; disabled: boolean | undefined; isExpanded?: boolean | undefined; isSelected?: boolean | undefined; key: string; onExpand: (e: MouseEvent) => void; }' is not assignable to type 'TableExpandRowProps'.
  Types of property 'isExpanded' are incompatible.
    Type 'boolean | undefined' is not assignable to type 'boolean'.
      Type 'undefined' is not assignable to type 'boolean'.

LINE: <TableExpandRow {...getRowProps({ row })}>
```

### Reproduction/example

https://stackblitz.com/edit/react-ts-dxtfrs?file=App.tsx

nevantan:
Update `<TableExpandRow />`, `<TableSelectAll />`, `<TableSelectRow />` props to make checked and isSelected optional since they already expect undefined or default to a boolean value anyway. This allows the values returned by getSelectionProps and getRowProps to be more consistent. This should resolve typescript errors that occur when using official examples.

Closes #14831

Making the argument for `getSelectionProps` optional allows the official examples to work without breaking anything with the change; the function already defaults the argument to an empty object. There is perhaps something to be said for refactoring the `getSelectionProps` function to better differentiate between the contexts in which it is used (`<TableSelectRow />` vs `<TableSelectAll />`), but the presence of an optional argument (as it is now) seems reasonable.

Switching the `onClick` handlers to use React's `MouseEvent` interface makes them compatible with `MouseEventHandler<HTMLButtonElement>`, which resolves a few of the errors on those components.

The `checked` prop was overly restrictive in the `<TableSelectRow />` and `<TableSelectAll />` prop definitions; the only places the prop is used within those components expects it to be optional. Relaxing the type on those two components resolves the typescript errors and seems to have been the original intent based on how it is typed in `DataTableRow`. This shouldn't be a breaking change to the component API since all currently allowed values are still valid after the change. The `isExpanded` prop on `<TableExpandRow />` is the same issue of being overly restrictive.

The `radio` key being returned from `getSelectionProps` unnecessarily included a `null` option - both the actual value and the type. Removing both references to `null` makes it compatible with `<TableSelectRow />` props.

#### Changelog

**New**

**Changed**

- DataTable `onClick` handlers now use React's `MouseEvent` interface instead of the native `MouseEvent`
- DataTable `getSelectionProps` arguments are now optional
- DataTable `getSelectionProps` is now accurately typed to return `checked` as an optional key
- DataTable `getSelectionprops` `radio` key is now `boolean | undefined` (changed from `boolean | undefined | null`)
- `<TableSelectRow />` and `<TableSelectAll />` `checked` prop is now optional
- `<TableExpandRow />` `isExpanded` prop is now optional

**Removed**

#### Testing / Reviewing

Run the code examples for `DataTable / Selection / Default` and `DataTable / Expansion / Default` in a TypeScript environment; the errors should be resolved.


nevantan:

emyarod (approved):
looks good to me!

tay1orjones (approved):
Hey sorry I missed this one in my notifications. This looks good to me

tay1orjones (approved):

tay1orjones:
There must've been some overlap with another PR. When I updated this with latest from `main` the files changed went down to just a handful of changes.

PR #17141, opened 2024-08-12
refactor: added &{} rules for styles below nested rules to follow `sass@1.77.7` ruleset.

### LINKED ISSUE #16962: [Bug]: `sass@1.77.7` Mixed Declarations deprecation warning in `@carbon/react`
Sass v1.77.7 introduced a new [deprecation warning for mixed declarations](https://sass-lang.com/documentation/breaking-changes/mixed-decls/). This causes noisy console output when working on this repo, as well as consumer environments, including the stackblitz [sandboxes](https://github.com/carbon-design-system/sandboxes/). 

> [!IMPORTANT]  
> This is just a warning and will not fail a build unless specifically configured to fail on warning messages. 

# How to fix

Do the following steps for each error:

1. Move the offending styles above the mixin
```diff
 .#{$prefix}--example__component {
+  padding-inline-end: layout.density('padding-inline');
+  transition: background-color motion(standard, productive) $duration-fast-02;

   @include button-reset.reset;
   
-  padding-inline-end: layout.density('padding-inline');
-  transition: background-color motion(standard, productive) $duration-fast-02;
 }
```
2. Run the storybook and look through components that utilize the styles for visual defects
3. If no defects, open a PR for the changes.

## If there are defects from local testing or failed percy snapshots on the PR
1. Move the styles back down below the mixin and enclose in a `& { }` block.
2. If needed, add a stylelint ignore to silence the "redundant nesting" and "duplicate selector" errors that are likely to pop up
```diff
 .#{$prefix}--example__component {
-  padding-inline-end: layout.density('padding-inline');
-  transition: background-color motion(standard, productive) $duration-fast-02;

   @include button-reset.reset;

+  /* stylelint-disable-next-line scss/selector-no-redundant-nesting-selector, no-duplicate-selectors */
+  & {
+    padding-inline-end: layout.density('padding-inline');
+    transition: background-color motion(standard, productive) $duration-fast-02;
+  }
 }
```

For more info on how we landed on this approach, see below https://github.com/carbon-design-system/carbon/issues/16962#issuecomment-2361175511 and the initial PR https://github.com/carbon-design-system/carbon/pull/17141

---

<details><summary>Click to expand original issue body text</summary>
<p>

### Package

@carbon/react

### Browser

Chrome

### Package version

v1.61.0

### React version

v18.2.0

### Description

Sass 1.77.7 recently introduced a deprecation which causes @carbon/react to fail on Vite with React18.2.0

https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1777
> **1.77.7**
Declarations that appear after nested rules are deprecated, because the semantics Sass has historically used are different from the semantics specified by CSS. In the future, Sass will adopt the standard CSS semantics.

Also: https://sass-lang.com/documentation/breaking-changes/mixed-decls/
> **Breaking Change: Mixed Declarations**
CSS is changing the way it handles declarations mixed with nested rules, and we want to make sure Sass matches its behavior.
(...)

### Reproduction/example

https://stackblitz.com/edit/github-2vsntp?file=package.json

### Steps to reproduce

simply install `sass@1.77.7` to use with @carbon/react

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

</p>
</details>

Gururajj77:
Part of #16962 

added the css rules that are written after the `@include` mixins in the scss, inside &{}, which is the recommended way to write styles after breaking change introduced by sass in the version 1.77.7

#### Changelog

**New**

- updated the sass to use the package version above 1.77.7

**Changed**

- added &{} rules for styles below nested rules


tay1orjones (commented):
I'm not sure what the best option is relating to the stylelint rule failures with `scss/selector-no-redundant-nesting-selector`, and `no-duplicate-selectors`. Some related discussion in these issues:

* https://github.com/stylelint-scss/stylelint-scss/issues/1020
* https://github.com/stylelint/stylelint/issues/7893

I wonder if as part of this we should validate if these styles need to come after the mixin to override values coming from the mixin. That way we'd only have the `& { }` ampersand nesting where we absolutely need it. The downside with doing that is it'll be remarkably slower than applying a blanket approach. I suppose we could infer that any styles that come after a mixin should use ampersand nesting. If there's no mixin involved we could avoid ampersand nesting. What do you think after testing the waters with the accordion styles?

The other failures look formatting-related which prettier should handle for you. I have the prettier vscode extension set to format on save so I don't have to run the command manually.

Also to recap from our convo earlier today - we don't need to update any sass version ranges in peer deps. Only the hard deps and devDeps need updated.

packages/styles/scss/__tests__/__snapshots__/type-test.js.snap
tay1orjones, line 1: This diff looks okay to me, it appears it's just changing the output of calc from multi-line to single line, and removes uneeded parens - e.g.

```diff
- "value": "calc(1.5rem +
-       0.25 *
-       ((100vw - 20rem) / 46)
-     )",
+ "value": "calc(1.5rem + 0.25 * (100vw - 20rem) / 46)",
```

So this removes 3 lines, which over the course of the entire file changes what line each selector/delcaration/property was on.

yarn.lock
```diff
@@ -24510,16 +24510,16 @@ __metadata:
   languageName: node
   linkType: hard
 
-"sass@npm:^1.51.0":
-  version: 1.62.1
```
tay1orjones, line 24562: Something between these version is why the type test snapshot is changing, though I couldn't pin down exactly what in the changelog https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1621

tay1orjones (commented):
Recapping our pairing session today, I left some comments about why the snapshot was changing.

guidari (approved):
LGTM! 🚀

tay1orjones (approved):

PR #17145, opened 2024-08-12
fix(#16882) Menu's target prop fix

### LINKED ISSUE #16861: [Bug]: MenuItem is not correctly focused when expanded within ComposedModal
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.60.0

### React version

v18.2

### Description

We are trying to use MenuButton within ComposedModal and found a strange bug. When you click the MenuButton to expand the menu, it should focus on the first menu item. However, the other button is being focused which is triggered by ComposedModal.

See the call stack below. (handleBlur and wrapFocus are from ComposedModal.)
image

### Reproduction/example

https://stackblitz.com/edit/github-yzyrus?file=src%2FApp.jsx

### Steps to reproduce

Click the MenuButton (Click me and you will see above button being focused!) and you will see the above button is focused. The console prints out the call stack too.

### Suggested Severity

### Application/PAL

IBM SOAR

### Code of Conduct

### LINKED ISSUE #16882: [Bug]: MenuButton and OverflowMenu cannot configure the Menu's `target` prop

NabeelAyubee:
Closes #16882

Renamed the target prop to menuTarget for the Menu component to improve clarity.
Added the menuTarget prop to MenuButton and OverflowMenu components, allowing consumers to specify the DOM element into which the menu is portaled.

#### Changelog

**Changed**

- The target prop has been renamed to menuTarget for improved clarity.
- The menuTarget prop has been added to the MenuButton components to provide better control over menu rendering.
- Added support for backward compatibility
- Hide menuTarget prop from actions in storybook
- Added deprecation warning for target prop

#### Testing / Reviewing

Testing / Reviewing
- Run yarn test to ensure that there are no code errors.
- Verify that the menuTarget prop correctly specifies the HTML element into which the menu is portaled. Ensure that the menu items are rendered in the target element as expected.
- Check the Storybook integration to ensure the new prop name is reflected and functioning correctly.


NabeelAyubee:

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -88,6 +88,12 @@ interface MenuProps extends React.HTMLAttributes<HTMLUListElement> {
   /**
    * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
    */
+  menuTarget?: Element;
```
tay1orjones: Since this prop is already within the Menu component it feels redundant to call it `menuTarget` instead of `target`

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -80,6 +80,17 @@ export interface MenuButtonProps extends ComponentProps<'div'> {
...
+  /**
+   * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
+   */
+  menuTarget?: Element;
+
+  /**
+   * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
+   * (Deprecated props)
+   */
+  target?: Element;
```
tay1orjones: Here I think it makes sense to only have one of these. Since it's referring to the target of the Menu, the name `menuTarget` feels appropriate.

```suggestion

  /**
   * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
   */
  menuTarget?: Element;
```

packages/upgrade/transforms/update-menu-target-prop.js
tay1orjones, line 1: I love that you wrote this, thank you! I think needing this codemod is another reason why we shouldn't rename this prop. If we leave it as-is this codemod can be removed and the user doesn't have to update any existing usages.

tay1orjones (changes requested):
Thanks again for taking this on! I don't think we should rename the `target` prop if possible. It makes the internal logic simpler, we can avoid the codemod, and consumers don't have to change any existing usages.

packages/upgrade/transforms/update-menu-target-prop.js
NabeelAyubee: > I love that you wrote this, thank you! I think needing this codemod is another reason why we shouldn't rename this prop. If we leave it as-is this codemod can be removed and the user doesn't have to update any existing usages.

Thanks. I'll remove the **menuTarget** prop and it's related changes, keeping only **target** prop.
Should I also rename the **menuTarget** prop in the **MenuButton** and **Overflow-featureFlag** components to maintain consistency?

packages/react/src/components/Menu/Menu.tsx
NabeelAyubee: Thank you for the suggestion. I have changed **menuTarget** prop back to **target** for **Menu** component.

packages/react/src/components/MenuButton/index.tsx
NabeelAyubee: I have made the suggested change.

packages/upgrade/transforms/update-menu-target-prop.js
NabeelAyubee: > I love that you wrote this, thank you! I think needing this codemod is another reason why we shouldn't rename this prop. If we leave it as-is this codemod can be removed and the user doesn't have to update any existing usages.

Thank you for the suggestions. I have reverted the **menuTarget** prop back to **target** for the **Menu** component. Additionally, the **OverflowMenu** and **MenuButton** components now use the **menuTarget** prop, as it refers to the **target** of the **Menu**. Also, the codemod has been removed since it is no longer needed.
Could you please re-review the changes?

tay1orjones (approved):
LGTM, thanks!

NabeelAyubee:
Hi @alisonjoseph & @emyarod , Can you please review the PR ? I've got one approval.

emyarod (approved):
looks good to me!

tay1orjones:
@NabeelAyubee looks like the build failed. Do you see the same locally when running `yarn test` from the root?

tay1orjones (approved):
Thanks for the assist @2nikhiltom! 🏆

PR #17191, opened 2024-08-16
fix(data-table): add single selection row for multiple tables support

### LINKED ISSUE #17190: [Bug]: carbon data table with radio button selection can not support multiple table in the same page
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.63.2

### React version

v17.0.2

### Description

this is a change from https://github.com/carbon-design-system/carbon/pull/14644, it's required the consumer that one page having multiple tables with radio selection should pass the TableSelectRow name. 
This is easy to be avoid from the library level, but not from the consumer application level, and the Table instance id can be supplied for that TableSelectRow name.

### Reproduction/example

N/A

### Steps to reproduce

our page has multiple tables, and it's failing for the radio button selection because they share a same name in different tables.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

yaoterry:
Closes #17190

{{short description}}

#### Changelog

**New**

- N/A

**Changed**

- Update the DataTable getSelectionProps to return the name with instanceId, so that it can support multiple tables with the selection in the same page.

**Removed**

- N/A

#### Testing / Reviewing

Make sure the Data table with Radio selection can be working well.


tay1orjones (approved):
Nice - thanks!

2nikhiltom (approved):

Issue #17173, opened 2024-08-15
[Feature Request]: Facilitate non-SCSS usage of Carbon tokens and mixins


lee-chase:
### The problem

Carbon uses SCSS to generate it's CSS. Use of CSS in JS, Less or other technique to apply Carbon is not always possible. This is a block to Carbonisation in some cases.

Currently some Carbon tokens can be accessed as Custom CSS properties, but this does not include all values e.g. motion easings. 

The custom properties are not well documented including the mechanisms for instantiating them. For example type mixins make  use of custom css properties, but it is not clear how these are instantiated in CSS.

![image](https://github.com/user-attachments/assets/9413e1a0-724e-4cda-bd35-067a51ba70c6)

Mixins can be discovered through inspection of the Javascript imports, and recreated. These mixins could be available as CSS classes e.g. `.productive-heading-04`.

### The solution

1. Document how each set of CSS custom properties are instantiated.
2. Publish a table of CSS tokens/mixins to Custom Properties or utility classes.
3. Provide Javascript utilities to parse/utilise the custom properties in JS.

### Business priority

Medium Priority = upcoming release but is not pressing

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

lee-chase:
Hey @tay1orjones assuming all values are output, then suitable documentation ticks most boxes for either JS in CSS and direct access to custom css properties ticks most boxes for non-SASS users. That does need to include, how to instantiate custom css properties like `--cds-productive-01-font-size`.

Not sold on utility classes myself, too easy to accidentally override, but do provide an non-sass way of collecting multiple CSS settings.

PR #17198, opened 2024-08-16
feat: custom type property output proposal

lee-chase:
Contributes to #17173

Adds the ability to output custom properties for type.

#### Changelog

**New**

- packages/type/custom-properties.scss

**Changed**

- package.json list of files

#### Testing / Reviewing

Added a simple test.


tay1orjones (commented):
What if they were emitted on `:root` behind a new sass config variable? Similar set up to what was done for the layout tokens, https://github.com/carbon-design-system/carbon/pull/14336, it would live in `@carbon/styles` but could put the `@include` behind the config variable. That way they'd conditionally be emitted inline in @carbon/styles and could use the existing mixin helpers like `declaration`.

lee-chase:
@tay1orjones I can add a feature flag, but making use of `enabled` from the type packages proved tricky.

With `enabled` not defined this `enabled('a-flag')` equates to true. 

I'm not sure how to access `enabled` without triggering a build error

```
@carbon/styles: sass.Exception [Error]: This module was already loaded, so it can't be configured using "with".
@carbon/styles:    ┌──> scss/_feature-flags.scss
@carbon/styles: 8  │ ┌ @forward '@carbon/feature-flags' with (
@carbon/styles: 9  │ │   $feature-flags: (
@carbon/styles: 10 │ │     'enable-css-custom-properties': true,
@carbon/styles: 11 │ │     'enable-use-controlled-state-with-value': true,
@carbon/styles: 12 │ │     'enable-css-grid': true,
@carbon/styles: 13 │ │     'enable-v11-release': true,
@carbon/styles: 14 │ │     'enable-experimental-tile-contrast': false,
@carbon/styles: 15 │ │     'enable-v12-tile-radio-icons': false,
@carbon/styles: 16 │ │     'enable-v12-structured-list-visible-icons': false,
@carbon/styles: 17 │ │     'enable-type-custom-properties': false,
@carbon/styles: 18 │ │   )
@carbon/styles: 19 │ │   !default
@carbon/styles: 20 │ │ );
@carbon/styles:    │ └─^ new load
@carbon/styles:    ╵
@carbon/styles:    ┌──> ../../node_modules/@carbon/type/scss/_styles.scss
@carbon/styles: 18 │   @use '@carbon/feature-flags' as *;
@carbon/styles:    │   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ original load
```

lee-chase:
This was my attempt with a feature-flag. https://github.com/lee-chase/carbon/tree/instantiateTypeTokensWithFeatureFlags

lee-chase:
Sorry, I just spotted that you said it would live in Carbon styles. I'll adjust and update the PR.

lee-chase:
@tay1orjones now has a feature flag.

tay1orjones (commented):
Nice! We could probably avoid the extra feature flag overhead by instead using just a sass config variable. A good example is `$css-font-face`

https://github.com/carbon-design-system/carbon/blob/539fc8f1b33a88083260b6b687ca510c09134700/packages/styles/scss/_config.scss#L15-L19

https://github.com/carbon-design-system/carbon/blob/36685a0be5e62db11a60d938c76b1c49a402c99a/packages/styles/scss/fonts/_index.scss#L43-L54

The config variables like this are the sort-of "default" convention for conditionally emitting styles where there isn't a js interaction that requires a js feature flag

tay1orjones (approved):
Snaps need updated, but otherwise LGTM!

There's a small chance that `true` by default could cause some issues, but we can change it to `false` in a subsequent PR if we run into any issues.

kennylam (approved):

packages/styles/__tests__/styles-test.js
```diff
@@ -80,3 +81,22 @@ describe('@carbon/styles', () => {
...
+      `);
+    const { stylesheet } = css.parse(result.css.toString());
+    const rules = stylesheet.rules.filter(
+      (rule) => rule.selectors[0] === '.selector'
+    );
+    const declarations = rules[0].declarations;
+
+    expect(declarations.length).toBe(191);
+  });
+});
```
tay1orjones: I just realized... I think the proper location for this to live would be in `packages/styles/scss/__tests__/type-test.js`

packages/styles/scss/__tests__/config-test.js
```diff
@@ -33,12 +33,14 @@ describe('@carbon/styles/scss/config', () => {
...
         $css--reset: false,
         $css--default-type: false,
+        $css--emit-type-custom-props: false
       );
 
       $_: get('config', (
         css--default-type: config.$css--default-type,
         css--font-face: config.$css--font-face,
         css--reset: config.$css--reset,
+        css--emit-type-custom-props,
```
tay1orjones:
```suggestion
        css--emit-type-custom-props: config.$css--emit-type-custom-props,
```

tay1orjones (commented):
I see two failures in CI, one is addressed below that I believe should fix it.

The other, in `packages/styles/scss/__tests__/type-test.js` is failing I assume because the output has changed due to the new flag being on by default. I think the assertion just needs updated?

PR #17225, opened 2024-08-20
fix(popover): prevent toggletip ref error

tay1orjones:
[Reported in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1724088462099059), this fixes Popover so that it won't incorrectly clone a ToggletipContent child. 

The condition was a bit strange and took me a bit to wrap my head around, so I also broke the logic out into something more verbose. I think it helps with readability and being able to more easily understand what's going on there, let me know what you think!

#### Changelog

**New**

- add test to cover the previous failure

**Changed**

- update popover child clone logic to not clone ToggletipContent children

#### Testing / Reviewing

- Pull this down locally so you can validate there are no console errors in dev.
- Look at all the toggletip stories and other component stories that use Toggletip, there should be no errors in the console
- The new test should pass in CI


2nikhiltom (approved):
No warning in console anymore, and the logic is more readable now 💯

Gururajj77 (approved):

alisonjoseph (approved):

annawen1 (approved):
LGTM!

PR #17239, opened 2024-08-22
fix: add new props to feature flag props

### LINKED ISSUE #17238: Add new boolean props for all existing feature flags
Add new boolean props for all existing feature flags

Update FeatureFlag internals to map the boolean props back into an object form to be used in the scope of FeatureFlagContext

riddhybansal:
Closes #17238

Add new boolean props for all existing feature flags

Update FeatureFlag internals to map the boolean props back into an object form to be used in the scope of FeatureFlagContext

#### Changelog

**New**

- Add new boolean props for all existing feature flags

-Update FeatureFlag internals to map the boolean props back into an object form to be used in the scope of FeatureFlagContext

- Added test cases for the props added

#### Testing / Reviewing

- Feature flag should render with boolean prop and flag object also. Added test case for the same.

### DIFF
--- a/packages/react/src/components/FeatureFlags/index.js
+++ b/packages/react/src/components/FeatureFlags/index.js
@@ -29,17 +29,37 @@ const FeatureFlagContext = createContext(GlobalFeatureFlags);
  * along with the current `FeatureFlagContext` to provide consumers to check if
  * a feature flag is enabled or disabled in a given React tree
  */
-function FeatureFlags({ children, flags = {} }) {
+function FeatureFlags({
+  children,
+  flags = {},
+  enableUseControlledStateWithValue,
+  enableV12TileDefaultIcons,
+  enableV12TileRadioIcons,
+  enableV12Overflowmenu,
+  enableTreeviewControllable,
+  enableExperimentalFocusWrapWithoutSentinels,
+}) {
   const parentScope = useContext(FeatureFlagContext);
   const [prevParentScope, setPrevParentScope] = useState(parentScope);
+
+  const combinedFlags = {
+    'enable-use-controlled-state-with-value': enableUseControlledStateWithValue,
+    'enable-v12-tile-default-icons': enableV12TileDefaultIcons,
+    'enable-v12-tile-radio-icons': enableV12TileRadioIcons,
+    'enable-v12-overflowmenu': enableV12Overflowmenu,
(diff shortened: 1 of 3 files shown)


packages/react/src/components/FeatureFlags/index.js
```diff
@@ -29,17 +29,37 @@ const FeatureFlagContext = createContext(GlobalFeatureFlags);
...
+  enableV12TileRadioIcons,
+  enableV12Overflowmenu,
+  enableTreeviewControllable,
+  enableExperimentalFocusWrapWithoutSentinels,
+}) {
   const parentScope = useContext(FeatureFlagContext);
   const [prevParentScope, setPrevParentScope] = useState(parentScope);
+
+  const combinedFlags = {
+    'enable-use-controlled-state-with-value': enableUseControlledStateWithValue,
```
tay1orjones, line 46: It doesn't look like this is actually used anywhere anymore, is that what you see as well? The useControlledStateWithValue in here doesn't seem to be used anywhere either https://github.com/carbon-design-system/carbon/blob/36685a0be5e62db11a60d938c76b1c49a402c99a/packages/react/src/internal/FeatureFlags.js#L49-L51

If so we should remove this one

packages/react/src/components/FeatureFlags/index.js
```diff
@@ -29,17 +29,37 @@ const FeatureFlagContext = createContext(GlobalFeatureFlags);
...
-function FeatureFlags({ children, flags = {} }) {
+function FeatureFlags({
+  children,
+  flags = {},
+  enableUseControlledStateWithValue,
+  enableV12TileDefaultIcons,
+  enableV12TileRadioIcons,
+  enableV12Overflowmenu,
+  enableTreeviewControllable,
+  enableExperimentalFocusWrapWithoutSentinels,
```
tay1orjones, line 40: Do these need the same defaults as what's defined in the root flag definition?

https://github.com/carbon-design-system/carbon/blob/b9abee7893060affe080b215d9feba8618f9cc10/packages/feature-flags/feature-flags.yml#L25-L52

tay1orjones (commented):
Looks great, thanks for adding tests! Just a couple questions

riddhybansal:
Hey there there is no need to merge this PR , commit of this PR is directly added here #17266

tay1orjones:
Closing in favor of #17266, I moved my review comments to there as well

PR #17241, opened 2024-08-22
fix(tag): add type="button" to interactive tag

### LINKED ISSUE #17240: [Bug]: Interactive tags don't have "type" attribute
### Package

@carbon/react

### Browser

_No response_

### Package version

1.61.0

### React version

_No response_

### Description

Currently, there are two ways to create interactive tags:
- Using the new `OperationalTag`
- Passing an `onClick` to a regular `Tag` (I assume this behavior will be deprecated at one point in favor for the first option; but that's not related to this issue).

In either case, the base element will be set to `<button>`. Unfortunately the `type` attribute is not set and cannot be overwritten by the developer as `type` is the prop to define the Tag's color.

This causes a problem when an interactive tag is used inside a `<form>` as all buttons inside forms implicitly have `type="submit"`. This means, clicking on an interactive tag would submit the form (initiating a GET request to the current page if not specified otherwise).

The real-world scenario of this came up by using C4IP's `TagSet` component inside C4IP's `CreateTearsheet` because every `CreateTearsheetStep` is a `form`. This breaks our user flow and we currently see no way to fix this on our side.

### Reproduction/example

https://stackblitz.com/edit/github-p7arbq?file=src%2FApp.jsx

### Steps to reproduce

```jsx
<form>
  <Popover>
    <Tag onClick={() => {}}>Trigger</Tag>
    <PopoverContent>Content</PopoverContent>
  </Popover>
</form>
```

1. Click on the "Trigger" tag
2. The page reloads

### Suggested Severity

### Application/PAL

watsonx

### Code of Conduct

janhassel:
Closes #17240

#### Changelog

**New**

- Add `type="button"` to tags that use `button` as the base element

#### Testing / Reviewing

- In the local storybook, visit the "unstable_InteractiveTag/Operational" story
- Inspect the tags and ensure the `<button>` has `type="button"`
- Recreate a demo inside a `<form>` with the reproduction code of #17240

### DIFF
--- a/packages/react/src/components/Tag/Tag.tsx
+++ b/packages/react/src/components/Tag/Tag.tsx
@@ -224,6 +224,7 @@ const Tag = React.forwardRef(function Tag<T extends React.ElementType>(
       disabled={disabled}
       className={tagClasses}
       id={tagId}
+      type={ComponentTag === 'button' ? 'button' : undefined}
       {...other}>
       {CustomIconElement && size !== 'sm' ? (
         <div className={`${prefix}--tag__custom-icon`}>


2nikhiltom (approved):
LGTM

guidari (approved):
LGTM!

PR #17258, opened 2024-08-23
fix(tilegroup): guard usage of potentially null children properties

tay1orjones:
[Surfaced in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1724390687794329)

When conditionally rendering `RadioTile`s within a `TileGroup` a race condition existed where children properties (`child.type`, `child.props.children`) could be null and would throw an uncaught exception.

Original repro: https://stackblitz.com/edit/github-mgvclx?file=src%2FExample%2FExample.jsx

#### Changelog

**New**

- Add a test to cover this race condition

**Changed**

- Use optional chaining on child property assertions

#### Testing / Reviewing

- You can recreate the repro locally to validate it'll work, but the new test covers it as well and should pass.


guidari (approved):
LGTM!! 🚀

2nikhiltom (approved):

PR #17266, opened 2024-08-25
feat(upgrade): adds featureflag-deprecate-flags-prop codemod and deprecates flags prop

### LINKED ISSUE #17243: Deprecate the existing `flags` prop, add codemod migration
- [x] Deprecate the existing `flags` prop
- [x] Write a codemod that converts the old `flags` format to the new boolean props
- [x] Add instructions for running the codemod to the `flags` deprecation warning message
- [x] Document the codemod in a new section on the feature flags .mdx page in storybook
- [x] Have codemod to remove old flags that are no longer needed, like [enable-v11-release](url)

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/17243

New props to feature flag props were added here https://github.com/carbon-design-system/carbon/pull/17239 

This introduces a new codemod to Convert FeatureFlags' `flags` object prop to individual `boolean` props
and  `deprecates` the existing `flags` prop
#### Changelog

**New**

- Added a new `featureflag-deprecate-flags-prop` codemod 
- Added `test fixtures` for all existing featureflags and tests to validate 
- `flags` prop for FeatureFlag is marked as deprecated 

**Changed**

- Updated `FeatureFlags/overview.stories.mdx` to document the codemod 
- Added instructions for running the `codemod` to the flags deprecation warning message

#### Testing / Reviewing

- Ensure that the test fixtures make sense - input is what someones code should look like before running the codemod, and the output is what we expect the codemod to generate/change it to.
- Make sure the docs read well and I didn't miss anything
- I used below command to run the codemode locally to verify it works  ( --parser=tsx used to process typescript files)
- `jscodeshift -t featureflag-deprecate-flags-prop.js --parser=tsx   _your_file_path_`

curious to hear how you feel about this


riddhybansal:
Hey @tay1orjones one doubt , what do you think on test fixture ? can we just make it generic for all feature flags , May be just this 

The input.js 

`<FeatureFlags 
  flags={{ 
    'enable-v12-tile-default-icons': true, 
    'enable-v12-overflowmenu': true, 
  }}> 
  <Tile /> 
</FeatureFlags>;
`


The output.js 
`<FeatureFlags 
  enableV12TileDefaultIcons 
  enableV12Overflowmenu

> 
  <Tile /> 
</FeatureFlags>;`

packages/react/src/components/FeatureFlags/index.js
```diff
@@ -65,9 +85,19 @@ FeatureFlags.propTypes = {
...
 
   /**
-   * Provide the feature flags to enabled or disabled in the current React tree
+   * Provide the feature flags to enabled or disabled in the current Rea,ct tree
    */
-  flags: PropTypes.objectOf(PropTypes.bool),
+  flags: deprecate(
+    PropTypes.objectOf(PropTypes.bool),
+    'The `flags` prop for `FeatureFlag` has ' +
+      'been deprecated. Please run the `featureflag-deprecate-flags-prop` codemod to migrate to individual boolean props.'
```
tay1orjones: Once added to the packages/upgrade/src/upgrades.js, update this to have the full command they can copy and paste to run

```
npx @carbon/upgrade migrate featureflag-deprecate-flags-prop --write
```

packages/react/src/components/FeatureFlags/index.js
```diff
@@ -29,17 +29,37 @@ const FeatureFlagContext = createContext(GlobalFeatureFlags);
  * along with the current `FeatureFlagContext` to provide consumers to check if
  * a feature flag is enabled or disabled in a given React tree
  */
-function FeatureFlags({ children, flags = {} }) {
+function FeatureFlags({
+  children,
+  flags = {},
+  enableUseControlledStateWithValue = false,
```
tay1orjones: I don't think this flag is actually used anywhere and can be removed. https://github.com/carbon-design-system/carbon/pull/17239#discussion_r1729308202

tay1orjones (commented):
This looks great! Just a few things from our pair reviewing session today:

* Add this codemod to the available migrations via @carbon/upgrade migrate, in packages/upgrade/src/upgrades.js
* Update the codemod to remove old flags that are no longer needed, like `enable-v11-release`. Mention this in the description of the migration in packages/upgrade/src/upgrades.js
* Add some new test fixtures mixing old flags and new to validate that the right one are being removed

packages/upgrade/src/upgrades.js
```diff
@@ -312,4 +312,44 @@ export const upgrades = [
...
+            });
+
+      await run({
+        dry: !options.write,
+        transform,
+        paths,
+        verbose: options.verbose,
+      });
+    },
+  },
```
tay1orjones: This should be moved up into the `migrations` array after `update-carbon-icons-react-import-to-carbon-react`

packages/react/src/components/FeatureFlags/__tests__/FeatureFlags-test.js
```diff
@@ -39,35 +39,39 @@ describe('FeatureFlags', () => {
...
-        b,
+        enableV12Overflowmenu,
+        enableTreeviewControllable,
       });
 
       return null;
     }
 
     render(
-      <FeatureFlags flags={{ a: true, b: false }}>
```
tay1orjones: Even though this usage is deprecated, I think we should still have a test covering its usage. `console.warn` will need spied/mocked like this test from ComboButton

https://github.com/carbon-design-system/carbon/blob/afce28df13ae3475eb32c7b65e0599eb86775458/packages/react/src/components/ComboButton/ComboButton-test.js#L191-L202

I'd just copy the old test into a new block above the new `should provide access to the feature flags for a scope` and call it `should provide access to the feature flags for a scope through deprecated flags prop` or something

packages/react/src/components/FeatureFlags/index.js
tay1orjones: This still applies and can be removed

tay1orjones (changes requested):
This looks so good! 🔥 just three minor things I saw

packages/react/src/components/FeatureFlags/__tests__/FeatureFlags-test.js
2nikhiltom: Great catch to cover all cases 🔥

tay1orjones (approved):
There's one more spot where `enable-use-controlled-state-with-value` is used that could probably be deleted, it's not used anywhere else in the styles

https://github.com/carbon-design-system/carbon/blob/b9abee7893060affe080b215d9feba8618f9cc10/packages/styles/scss/_feature-flags.scss#L11

kennylam (approved):

PR #17268, opened 2024-08-26
feat: combobox's new feature, autocomplete with typeahead

### LINKED ISSUE #14007: [ComboBox] Autocomplete implementation
Ref link for implementation https://github.com/carbon-design-system/carbon/issues/13903

Example: https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/ComboBox/ComboBox.tsx
Repo for ComboBox https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/ComboBox/ComboBox.tsx

This looks like it's not an easy update and going to take some time

Gururajj77:
Closes https://github.com/carbon-design-system/carbon/issues/14007

Added autocomplete and typeahead feature for combobox with custom filter to support the typeahead.

#### Changelog

**New**

- new `autocomplete` prop
- new `autocompleteCustomFunction` which is 1-to-1 mapping filter function which filters matching both character and position.
- new test to make sure `autocomplete` and `autocompleteCustomFunction` are working.

**Changed**

- added `getMenuProps` and `refs.setFloationg` to `useMemo` hook
- modified import for `act` in the test helpers

#### Testing / Reviewing

- navigate to new  "Autocomplete with Typeahead" story
- check that the suggestions are coming in the input box or not.


packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -643,8 +701,19 @@ const ComboBox = forwardRef(
         return itemToString(item);
       },
       onInputValueChange({ inputValue }) {
-        setInputValue(inputValue || '');
-        setHighlightedIndex(indexToHighlight(inputValue));
+        // setInputValue(inputValue || '');
+        // setHighlightedIndex(indexToHighlight(inputValue));
```
kennylam: Can remove.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -643,8 +701,19 @@ const ComboBox = forwardRef(
...
+        if (inputValue === null) {
+          setInputValue('');
+          setCursorPosition(0);
+        } else {
+          const newInputValue = (inputValue ?? '').endsWith(typeaheadText)
+            ? inputValue
+            : inputValue + typeaheadText;
+          console.log(newInputValue);
```
kennylam: Can remove.

kennylam (commented):
@Gururajj77 Are there design specs for the functionality of this new feature? I think in matching 1:1 with the filtering behavior, autocomplete is too aggressive and unpredictable. Unless users know exactly when to stop typing, extra characters get inserted into the matching string. If users try to delete the text, the autocomplete can insert other matching strings (try deleting `apple`), or even prevent them from deleting at all (try deleting `banana`). Using the clear (`x`) button also doesn't always clear the text. The only reliable way is to select all and delete.

https://github.com/user-attachments/assets/712bc5b2-9e8a-4574-9031-057c9316ce3b

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -424,6 +449,44 @@ const ComboBox = forwardRef(
...
+
+        if (filteredItems.length > 0) {
+          const suggestion = itemToString(filteredItems[0]);
+          return suggestion.slice(input.length);
+        }
+        return '';
+      };
+
+      if (autocomplete) {
+        if (autocomplete && inputValue.length >= prevInputLengthRef.current) {
```
tay1orjones:
```suggestion
        if (inputValue.length >= prevInputLengthRef.current) {
```

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -424,6 +449,44 @@ const ComboBox = forwardRef(
...
+          const suggestion = getTypeaheadSuggestion(inputValue);
```
tay1orjones: It's not totally necessary to have this in a function, could move the contents of the function inline here if we're not expecting any reuse of the function

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -991,6 +1094,11 @@ ComboBox.propTypes = {
    */
   autoAlign: PropTypes.bool,
 
+  /**
+   * **Experimental**: will enable autcomplete and typeahead for the input field
+   */
+  autocomplete: PropTypes.bool,
```
tay1orjones: Before this change a user could pass in `autocomplete`, which would be spread onto the `input` via `...rest` and it could contain [any string value allowed](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values) by that html attribute. After this change their usage would present a prop/type error since it's locked to boolean.

The easiest way to avoid that is probably to just name this something different. `typeahead` maybe?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -449,7 +512,9 @@ const ComboBox = forwardRef(
       inputValue: string | null
     ) =>
       items.filter((item) =>
-        shouldFilterItem
+        autocomplete
+          ? autocompleteCustomFilter({ item: itemToString(item), inputValue })
+          : shouldFilterItem
```
tay1orjones: Do we need to update the prop/type comment for `shouldFilterItem` to specify that the prop will be ignored when `autocomplete`/`typeahead` is on?

tay1orjones (commented):
Looks great! Just a few things we discussed in our review pairing session today

packages/react/src/components/ComboBox/ComboBox-test.js
```diff
@@ -446,4 +446,140 @@ describe('ComboBox', () => {
...
+      await waitForPosition();
+      const inputNode = findInputNode();
+      expect(inputNode).toHaveAttribute('autocomplete');
+    });
+    it('should use autocompleteCustomFilter when autocomplete prop is true', async () => {
+      render(<ComboBox {...mockProps} autocomplete />);
+
+      // Open the dropdown
+      const input = screen.getByRole('combobox');
+      fireEvent.click(input);
```
tay1orjones:
```suggestion
      await userEvent.click(input);
```

packages/react/src/components/ComboBox/ComboBox-test.js
```diff
@@ -446,4 +446,140 @@ describe('ComboBox', () => {
...
+    };
+
+    it('should respect autocomplete prop', async () => {
+      render(<ComboBox {...mockProps} autocomplete />);
+      await waitForPosition();
+      const inputNode = findInputNode();
+      expect(inputNode).toHaveAttribute('autocomplete');
+    });
+    it('should use autocompleteCustomFilter when autocomplete prop is true', async () => {
+      render(<ComboBox {...mockProps} autocomplete />);
```
tay1orjones: and I think technically we're supposed to call setup before render now so that userEvent methods will all use the same user "instance"

```suggestion
      const user = userEvent.setup();
      render(<ComboBox {...mockProps} autocomplete />);
```

and then the usages below change too
```diff
- await userEvent.click(...
+ await user.click(...
```

tay1orjones: https://testing-library.com/docs/user-event/setup#starting-a-session-per-setup

tay1orjones (commented):
A few extra things on the tests

tay1orjones:
I also tried to use this with `allowCustomValue` but with the existing bugs on `allowCustomValue` it's hard to know if there's any interaction. Is it worth adding some tests to cover autocomplete with allowCustomValue on.

thyhmdo (approved):
Looks good to me.

tay1orjones (approved):

PR #17276, opened 2024-08-27
feat: rename selectionMode to shouldSwitchOnKeyboardFocus

### LINKED ISSUE #16511: [Feature Request]: Improve ContentSwitcher documentation
### The problem

I think that both the name and the description of the `selectionMode` attribute of the `ContentSwitcher` component can be improved. Currently, it is:

```typescript
/**
    * Choose whether or not to automatically change selection on focus
*/
selectionMode?: 'automatic' | 'manual';
```

This is not totally true because the selection won't change on mouse focus. This attribute, based on a quick look at the [component's code](https://github.com/carbon-design-system/carbon/blob/23f19271f964a10a9fbc4616a43b05f5cc14e7d1/packages/react/src/components/ContentSwitcher/ContentSwitcher.tsx#L152), seems to be intended to determine whether the arrow left/right shortcuts should changethe selection index or just focus the left/right `Switch` child.

### The solution

Add a more descriptive attribute name and description, like: 

```typescript
/**
    * Choose whether or not to automatically change selection on left/right arrow pressed. Default: 'automatic'
*/
arrowShortcutSelectionMode?: 'automatic' | 'manual';
```

### Examples

_No response_

### Application/PAL

LLM as a Judge

### Business priority

### Available extra resources

_No response_

### Code of Conduct

ihabib1983:
Closes #16511

Rename `selectionMode` to `shouldSwitchOnKeyboardFocus` I used boolean here since the allowed values are only 2 (enabled or disabled) and it's enabled by default

#### Changelog

**New**

- New variable `shouldSwitchOnKeyboardFocus`

**Removed**

-  Deprecate old variable `selectionMode`


ihabib1983:

guidari (approved):
I tested on Playgroung and LGTM 🚀

alisonjoseph (dismissed):
I'm unsure on the prop naming you chose, wanted to get your input @tay1orjones as you commented on the original issue suggesting `keyboardSelectionMode` as the updated prop name. I'm also not sure we'd want to switch it to a boolean just for the sake of switching, will make migration slightly more complicated in the future.

tay1orjones (commented):
Yeah I agree this shouldn't be a boolean. Leaving it as an enum allows for API expansion without breaking changes. I think it would be cleaner to just update the comment to more explicitly describe it being related to keyboard navigation and avoid the rename altogether.

Plus we're aiming for changes like what you have now to be accompanied by an automated migration (codemod) to assist users in migrating their existing usages to the new API(s). In this case don't think the prop rename is worth that extra effort of having to write a codemod.

tay1orjones (approved):

tay1orjones:
I pushed an update that reduces the changes to only be an improvement to the prop description/comment.

alisonjoseph (approved):

Issue #16243, opened 2024-04-22
[Proposal]: Export internal utilities


tay1orjones:
## Proposal

There are a series of helper functions, hooks, and other utilities within the monorepo codebase that multiple consumers have requested be exported so that they can use them within their projects.

So far this idea hasn't been acted on because these have been considered internal-only helpers intended to not be bound by semver. By publishing these and binding them to semver we run the risk of over-burdening the core system maintainers. 

That said, this is a big opportunity for code reuse across the Carbon ecosystem. Having a dedicated package to surface helper utilities would reduce code duplication and encourage a unified approach to the purpose of each utility.

## Open questions
- [ ] How do we define a utility, and how do we determine if a utility should be exported for use across other projects?
- [x] Can some of these be framework agnostic so reuse could happen across the _entire_ ecosystem?
  - A: Yes
- [x] Do react-specific utilities need their own package? `@carbon/hooks`?
  - A: Yes, see plan of action below
- [x] Is there still a place for some truly internal-only helpers?
  - A: Yes, not all internal helpers will be made available. We'll make them public on a per-request basis
- [x] Do framework-specific utilities need their own package? `@carbon/web-component-utilities`?
  - A: Yes, see plan of action below
- [x] What utilities should live within the monorepo? Framework agnostic and react helpers? All helpers regardless of framework? Do framework authors prefer helpers to live in the monorepo or their repo?
  - A: Everything can be in the monorepo for now
- [x] Is this package(s) bound by semver?
  - A: Yes
- [x] Who is this package(s) intended to be used by? Only framework authors? Framework authors and PAL libraries? Any project using Carbon, including projects building products with Carbon?
  - A: Anybody, though framework authors and PALs will be the primary users
- [ ] How could we document these helpers?
- [ ] Should we establish a working group to help answer these questions?
- [ ] ...

## List of proposed helpers to export

* ... need to generate this list with interested parties, please comment below if you're interested

## Plan of action as of 7/23/2024

From a recent discussion with @janhassel:

> 1. Framework-agnostic utilities will live in `carbon-design-system/carbon/packages/utilities` and be published as `@carbon/utilities`
> 2. Framework-specific utilities will live in `carbon-design-system/carbon/packages/utilities-<framework>` and be published as `@carbon/utilities-<framework>`
> 3. All framework-agnostic utilities are forwarded by the framework-specific package. That is, `@carbon/utilities-react` will export all members from `@carbon/utilities`
[...]

PR #17280, opened 2024-08-27
chore: setup utilities package

janhassel:
Ref #16243

#### Changelog

**New**

- Set up new package: `@carbon/utilities`
  - Add `dateTimeFormat` module to it
- Set up new package: `@carbon/utilities-react`
  - Copy `useNoInteractiveChildren` from `@carbon/react` as a starter


packages/utilities-react/telemetry.yml
```diff
@@ -0,0 +1,10 @@
+# yaml-language-server: $schema=https://unpkg.com/@ibm/telemetry-config-schema@v1/dist/config.schema.json
+version: 1
+projectId: 00000000-0000-0000-0000-000000000000
```
mattrosno:
```suggestion
projectId: c2026f4e-7033-4076-9006-5e5c8cc81e58
```

packages/utilities/telemetry.yml
```diff
@@ -0,0 +1,10 @@
+# yaml-language-server: $schema=https://unpkg.com/@ibm/telemetry-config-schema@v1/dist/config.schema.json
+version: 1
+projectId: 00000000-0000-0000-0000-000000000000
```
mattrosno:
```suggestion
projectId: 854c4516-4c8f-43cf-b2e0-08e4190795e0
```

packages/utilities-react/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "peerDependencies": {
+    "react": "^16.8.6 || ^17.0.1 || ^18.2.0"
+  },
+  "devDependencies": {
+    "esbuild": "^0.23.0",
+    "rimraf": "^6.0.0"
+  },
+  "dependencies": {
+    "@carbon/utilities": "^0.0.1",
+    "@ibm/telemetry-js": "^1.5.0"
```
mattrosno: Can you install a newer version of `@ibm/telemetry-js`? The latest version is 1.6.1. I understand the `^` syntax will pick up the latest compatible v1 here. It's just that our version 1.5.0 had issues that we immediately fixed, so we want to prevent 1.5.0 from being used if possible.

packages/utilities/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "clean": "rimraf es lib",
+    "test": "echo yay && exit 0",
+    "postinstall": "ibmtelemetry --config=telemetry.yml"
+  },
+  "devDependencies": {
+    "esbuild": "^0.23.0",
+    "rimraf": "^6.0.0"
+  },
+  "dependencies": {
+    "@ibm/telemetry-js": "^1.5.0"
```
mattrosno: Here too. Can you specify something newer than 1.5.0? Latest is 1.6.1.

mattrosno (commented):
I added the project IDs here and some comments about the Telemetry package versions. Everything else looks good.

Reviewing this, https://github.com/ibm-telemetry/telemetry-config-schema?tab=readme-ov-file#js-scope, it's unclear to me if we want to set `allowedArgumentStringValues` in the JS functions configuration. I guess we can wait to see if we're capturing function arguments and if so, we can allowlist values like "long", "short", "narrow", etc. These are only for string arguments thought, and it looks like you're passing in locale and style in an object parameter, so given current Telemetry functionality, I don't think it's possible to track parameters here.

Also, I was going to suggest adding the `jsx` config to the React package, but it doesn't have components, just hooks, and the `js` scope with `functions` will capture React hooks so we're good there.

janhassel:
@mattrosno 

>  I guess we can wait to see if we're capturing function arguments and if so, we can allowlist values like "long", "short", "narrow", etc.

Will this actually work since these string values are passed as part of an object?

mattrosno:
> Will this actually work since these string values are passed as part of an object?

@janhassel unfortunately no, as we only capture booleans, numbers, and allowed strings. We didn't have a secure way to capture object values.

emyarod (approved):
looks good to me!

tay1orjones (approved):
LGTM, thanks again for doing this! 🎉

PR #17288, opened 2024-08-28
fix: added props in skeleton state

### LINKED ISSUE #16717: [Bug]: `NumberInputSkeleton` doesn't support helper text or sizes
Compare `NumberInputSkeleton` to Figma.

- Add size props
- Add show/hide helpertext

Question for @carbon-design-system/design do we want to add the the +/- buttons to the coded skeleton state?

riddhybansal:
Closes #16717 

Added size props in number input skeleton state

#### Changelog

**New**

-Added size prop

#### Testing / Reviewing

Check size prop is added for the skeleton state or not


riddhybansal:
Hey there @alisonjoseph , I have questions regarding the issue for which this PR is raised 
-  The +/- button or any icon for skeleton state for any of the component even in fluid components in storybook is not present however it is present in figma for numberInput @alina-jacob suggest to have a design audit , may be we can remove it from figma.
- For the helper text thing the show/hide is present in figma but its not there in both default and skeleton state. What should we do for this case should we add show hide for both default and skeleton. In this issue its mentioned we need for skeleton, but I think if needed we should add in both. Please suggest.

kennylam (approved):

alisonjoseph (changes requested):
I think you still need to add the corresponding styles for each size prop, I tested locally and they all appear the same size.

I think its fine to ignore the +/- for now since design is going to be doing an audit. 

For the regular component if no helperText is added then it doesn't show up, so a hideHelperText prop isn't needed, but I think we should still add to the skeleton state.

alisonjoseph:
@riddhybansal following up on this one

riddhybansal:
There is some error while commiting changes to this branch so created another PR for the same #17825

PR #17295, opened 2024-08-28
feat(ai-label): add ts types

### LINKED ISSUE #17232: Add TypeScript types to AI Label (formerly Slug)

ariellalgilmore:
Closes #17232

Add typescript types to AI-label

#### Changelog

**Changed**

- Converted AILabel file from JS to TSX.
- updated align to copy `Popover` syntax with deprecated values

#### Testing / Reviewing

functionality stay the same
checks still passing
storybook looks good


ariellalgilmore:
When checking the blame history in github you can still see the history even though it shows it as a new file

Screenshot 2024-08-28 at 8 21 25 AM

emyarod (approved):
looks good to me!

2nikhiltom (approved):
LGTM 💯

annawen1 (approved):

packages/react/src/components/AILabel/index.tsx
```diff
@@ -0,0 +1,365 @@
...
+    'top-right': 'top-end',
+    'bottom-left': 'bottom-start',
+    'bottom-right': 'bottom-end',
+    'left-bottom': 'left-end',
+    'left-top': 'left-start',
+    'right-bottom': 'right-end',
+    'right-top': 'right-start',
+  };
+  return mapping[deprecatedValue];
+};
```
tay1orjones: Not a blocker to merge at all, but at some point we should probably make this a reusable helper - we're going to be using this all over the codebase 😅

tay1orjones (approved):

packages/react/src/components/AILabel/index.tsx
tay1orjones: I totally forgot we have one already: 

https://github.com/carbon-design-system/carbon/blob/afce28df13ae3475eb32c7b65e0599eb86775458/packages/react/src/tools/createPropAdapter.js#L44-L67

ariellalgilmore: ooh thanks! I updated AI label!

PR #17301, opened 2024-08-29
fix(codesnippet): inline button should be width of content

tay1orjones:
[Reported in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1724954498462949), the inline code snippet button was incorrectly limited to 320px wide causing a display issue:

image

#### Changelog

**Changed**

- Update code snippet styles to unset the max-inline-width

**Removed**

- Remove the `declaration-property-value-disallowed-list` rule that disallowed `unset` due to lack of support in IE11. IE11 isn't in our support matrix anymore, so I think this can be safely removed from the config

#### Testing / Reviewing

- Go the code snippet inline story, edit the text to be a really long string. It should display properly. 

https://github.com/user-attachments/assets/3855998f-756c-4522-a87d-c50628dbe9e6

### DIFF
--- a/config/stylelint-config-carbon/rules/limit-language-features.js
+++ b/config/stylelint-config-carbon/rules/limit-language-features.js
@@ -81,12 +81,6 @@ module.exports = {
     'declaration-property-unit-allowed-list': {
       '/^animation/': ['ms'],
     },
-    // Specify a blacklist of disallowed property and value pairs within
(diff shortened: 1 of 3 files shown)


alisonjoseph (approved):
so fast! LGTM

guidari (approved):

PR #17304, opened 2024-08-30
Adding Codecov

### LINKED ISSUE #17124: Implement code coverage tooling
Based on the research, https://github.com/carbon-design-system/carbon/issues/17123, this issue tracks the implementation of a code coverage tool to track the coverage improvements in #16319

guidari:
Closes #17124

This PR is adding [Codecov](https://about.codecov.io/) to make sure we are aware of testing when updating or creating a component. Codecov only has access to carbon repository.

We will not see any CI or comments in this PR. We first need to merge this PR so Codecov will sync everything.
- "Once merged to your default branch, subsequent pull requests will have Codecov checks and comments."

#### Changelog

**New**

- Added new SECRET called `CODECOV_TOKEN` to settings inside `carbon` repository.
- Added Codecov app inside `Settings > GitHub Apps`
- Added `codecov.yml` to configure the codecov.

**Changed**

- Now the test in CI runs `--collectCoverage` to push the report to Codecov

#### Testing / Reviewing

- CI should pass
- I have a local Codecov configured where I tested everything, you guys can see the reports and I can show how it works if needed.
PR example: https://github.com/guidari/carbon/pull/6
Branch that simulates main branch: https://github.com/guidari/carbon/tree/main-codecov


alisonjoseph (approved):

tay1orjones (changes requested):
Hey, this is a great start! I didn't realize `components` is a new feature codecov [recommends](https://docs.codecov.com/docs/components#should-i-use-flags-or-components) instead of `flags` so you don't have to upload multiple coverage reports (one for each flag). One caveat I saw there:

> if you require per-flag coverage trend information and UI support immediately, then it is recommend to use Flags for the time being

I don't think trend information is necessary right now, but worth noting. We should explore the UI once this is merged to see if we can get the information we want. 

---

We might want to consider a few other pieces before merging:

- [x] Move the config file to `.github/codecov.yml`. I don't see documented support for per-package configs [in the docs](https://docs.codecov.com/docs/codecov-yaml#can-i-name-the-file-codecovyml). Did you find docs for multiple config files in a single repo?
- [x] Update the config file to configure a `component` entry for each package that has tests ran on `yarn test` - react, colors, upgrade, styles, etc. 
  - Right now the config uses what appears to be the [simpler way](https://docs.codecov.com/docs/commit-status#splitting-up-projects-example) to configure `components`. We might need to use the more advanced [`component_management` config option](https://docs.codecov.com/docs/components#getting-started-with-components) so we can configure thresholds for each package.
- [x] Make the status checks [non-blocking](https://docs.codecov.com/docs/common-recipe-list#set-non-blocking-status-checks) for now

What do you think about these?

tay1orjones (approved):
Looks great to me! I confirmed the SHA matches the tagged release in the repo for that action.

PR #17306, opened 2024-08-30
chore(actions): update image path in dev office hours slack block

kennylam:
The image being used in the Developer Office Hours Slack notification is no longer being used in the Carbon Website, meaning the image URI now 404s – which causes the [notification workflow](https://github.com/carbon-design-system/carbon/actions/runs/10579675670) to fail.

image

This updates the image to use GH's CDN, just like all other image references.

#### Changelog

**Changed**

- update image URI in Slack block


tay1orjones (approved):
Great catch! Thanks for figuring this out

alisonjoseph (approved):

PR #16910, opened 2024-07-03
Storybook 8

guidari:
Closes #16468

Storybook 8 is coming 🚀

#### Changelog

**New**

- Upgrade Storybook from `7.1.0` to `8.1.10`
- Added `@storybook/addon-webpack5-compiler-babel`

**Changed**

- `ArgsTable` Doc block replace by `ArgTypes` on `.mdx` files. | [docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#argstable-doc-block-removed)
- `<Story of={componentStory} />` replace by `<Canvas of={componentStory} />` | [docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#canvas-block)
- Switch `@storybook/addons` for `@storybook/manager-api` | [docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#canvas-block:~:text=%40storybook/addons%20%2D%20this%20package%20has%20been%20split%20into%202%20packages%3A%20%40storybook/preview%2Dapi%20and%20%40storybook/manager%2Dapi%2C%20see%20more%20here%3A%20New%20Addons%20API.)
- Fixed imports in `packages/react/src/index.js`

**Removed**

- Removed the `babelLoader.exclude` since there is no need anymore for that.

References:
- https://github.com/vitejs/vite/issues/5269
- https://github.com/evanw/esbuild/issues/2363#issuecomment-1171992206

#### Testing / Reviewing

Pull down locally

`yarn install`
`yarn build`
`yarn storybook`

- Please test all things in storybook 🙏🏻, docs, playground controls, skeleton, default stories, feature flags, actions... 
- Make sure we still have all the functionalities we used to have on v7.
Should all be running smoothly and let me know of any issues.
- CI should pass

PR #17308, opened 2024-08-30
fix(storybook): use plex in manager ui even when not installed

tay1orjones:
Something I noticed in #16910 - on machines that do not have Plex installed, the storybook uses a system default font. 

#### Changelog

**Changed**

- Update storybook's `manager-head.html` to load plex from the CDN into the [Manager](https://storybook.js.org/docs/addons#storybook-basics) portion of storybook (sidebar nav, controls pane, etc.)

#### Testing / Reviewing

- Open the deploy preview on a machine with plex installed
  - inspect in the firefox dev tools 'fonts' pane, the locally installed version of plex should be on the page in addition to the one(s) loaded from the deploy preview itself.
- Open the deploy preview on a machine without plex installed
  - inspect in the firefox dev tools ... you should see plex being loaded from `s81c.com/...`
  - The storybook Manager ui should be rendered with plex

| Windows w/out local Plex (old) | Windows w/out local Plex (this PR) |  Mac w/ local Plex (prod) |
|--------|--------|--------|
|  |  |  |


emyarod (approved):
looks good to me!

image

image

Gururajj77 (approved):

PR #17317, opened 2024-09-03
fix: fixed focus flow issue on right arrow press

### LINKED ISSUE #17281: [Bug]: TreeNode error in keypress handler
### Package

@carbon/react

### Browser

_No response_

### Package version

1.64.1

### React version

_No response_

### Description

Uncaught TypeError: Cannot read properties of null (reading 'lastChild').

Error occurs [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/TreeView/TreeNode.tsx#L227) due to expression in brackets evaluating to undefined. 

Apparently `currentNode` ref isn't being passed [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/TreeView/TreeNode.tsx#L345) but should be. 

### Reproduction/example

https://react.carbondesignsystem.com/?path=/docs/components-treeview--overview

### Steps to reproduce

- click on a TreeNode
- press right arrow on the keyboard multiple times
- observe error in console

### Suggested Severity

### Application/PAL

wdp-gov/lineage-ui

### Code of Conduct

Gururajj77:
Closes #17281

The bug mentioned in the above issue, shows that the expected behaviour of right arrow press on a focussed and a opened node should bring down to the first child of the opened node is missing. This has been fixed in this PR.

The reference to the behaviour is on our website's `TreeView` component [page](https://carbondesignsystem.com/components/tree-view/usage/#behaviors:~:text=the%20node%20container.-,Keyboard%20controls,-Key)

#### Changelog

**New**

- renamed `ref` to `forwardRef` to distinguish between our local ref
- new `setRefs` function, which checks for `forwardRef` to say its a parent node or not
- passed the `setRefs` to the `li` in the jsx

**Changed**

- added some error handling to make sure that we don't have any console errors

#### Testing / Reviewing

- open the deploy preview of the `TreeView`
- in the default story, focus on the first item
- navigate to a closed node which has more leaf nodes in it
- press right arrow once, to see that the node opens
- press right arrow once again, and the focus should move to the first child of the opened parent node.


packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -196,34 +208,32 @@ const TreeNode = React.forwardRef<HTMLLIElement, TreeNodeProps>(
...
         if (children && expanded) {
           if (!enableTreeviewControllable) {
             onToggle?.(event, { id, isExpanded: false, label, value });
           }
           setExpanded(false);
         } else {
-          /**
-           * When focus is on a leaf node or a closed parent node, move focus to
-           * its parent node (unless its depth is level 1)
-           */
```
emyarod: I think this comment is still relevant right?

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -196,34 +208,32 @@ const TreeNode = React.forwardRef<HTMLLIElement, TreeNodeProps>(
...
+            parentNode.focus();
+          }
         }
       }
       if (children && match(event, keys.ArrowRight)) {
         if (expanded) {
-          /**
-           * When focus is on an expanded parent node, move focus to the first
-           * child node
-           */
```
emyarod: same as above, I think this comment is still relevant right?

emyarod (approved):
looks good to me! just a question about the comment removals

packages/react/src/components/TreeView/TreeNode.tsx
Gururajj77: The comments are relevant, got missed in the cleanup 😅

packages/react/src/components/TreeView/TreeNode.tsx
Gururajj77: The comments are relevant, got missed in the cleanup 😅

tay1orjones (approved):
Thanks for adding those tests!

PR #17385, opened 2024-09-07
Add typescript types to fluid date picker and its skeleton state and fluid datepicker input

### LINKED ISSUE #17365: Add typescript types to FluidDatePickerInput

### LINKED ISSUE #17366: Add typescript types to FluidDatePicker and its skeleton state

riddhybansal:
Closes #17366
Closes #17365

Added typescript types to fluid date picker and its skeleton state and fluid datepicker input

#### Changelog

**Changed**

- Added typescript types to fluid date picker and its skeleton state and fluid datepicker input and converted index.js to index.tsx for both the components

#### Testing / Reviewing

FluidDatePickerInput, FluidDatePicker and its skeleton should work as intended.


riddhybansal:
Hey there @tay1orjones @Gururajj77 there is not any props defined for FluidDatePickerInput so what should be the type for this since no interface!

Gururajj77:
> Hey there @tay1orjones @Gururajj77 there is not any props defined for FluidDatePickerInput so what should be the type for this since no interface!

We can convert the file to `.tsx`. If we have a `propTypes` object, then we need to write an interface for the component. Looking at the `FluidDatePickerInput`, we don't have any `propTypes` object, we don't need an interface here.

packages/react/src/components/FluidDatePicker/FluidDatePicker.tsx
```diff
@@ -12,7 +12,41 @@ import DatePicker from '../DatePicker';
 import { usePrefix } from '../../internal/usePrefix';
 import { FormContext } from '../FluidForm/FormContext';
 
-const FluidDatePicker = React.forwardRef(function FluidDatePicker(
+export interface FluidDatePickerProps {
+  /**
+   * The child node(s)
+   */
+  children?: React.ReactNode;
+  /**
```
Gururajj77, line 20: Importing `ReactNode` in the imports section and then using just `ReactNode` instead of `React.ReactNode` will take advantage of React's tree-shakable API.

packages/react/src/components/FluidDatePicker/FluidDatePicker.tsx
```diff
@@ -6,13 +6,47 @@
...
+   */
+  className?: string;
+  /**
+   * Specify whether or not the control is invalid
+   */
+  invalid?: boolean;
+  /**
+   * Provide the text that is displayed when the control is in error state
+   */
+  invalidText?: React.ReactNode;
```
Gururajj77, line 31:
```suggestion
  invalidText?: ReactNode;
```

packages/react/src/components/FluidDatePickerInput/FluidDatePickerInput.tsx
```diff
@@ -11,11 +11,11 @@ import { FormContext } from '../FluidForm/FormContext';
 
 const FluidDatePickerInput = React.forwardRef(function FluidDatePickerInput(
   { ...other },
-  ref
+  ref: React.Ref<HTMLDivElement>
 ) {
   return (
     <FormContext.Provider value={{ isFluid: true }}>
-      <DatePickerInput ref={ref} {...other} />
+      <DatePickerInput id={''} labelText={undefined} ref={ref} {...other} />
```
tay1orjones: Could the `id` and `labelText` be removed here so users can still configure them?

tay1orjones (changes requested):

packages/react/src/components/FluidDatePickerInput/FluidDatePickerInput.tsx
riddhybansal: id and labelText are required props in DatepickerInput so need to pass something. What can we do other than this?

tay1orjones: @riddhybansal If they're required in `DatePickerInput`, they should be required here in `FluidDatePickerInput` too and passed through `...other`.

If typescript isn't automatically inferring that relationship then I think we need to explicitly specify that `FluidDatePickerInput` inherits/extends `DatePickerInputProps`.

tay1orjones (approved):

Gururajj77 (approved):

PR #17418, opened 2024-09-11
fix(accordion): add transition styles back

### LINKED ISSUE #17197: [Bug]: Accordion items do not animate
### Package

@carbon/react

### Browser

Chrome

### Package version

11.64.0

### React version

18

### Description

Regression caused by https://github.com/carbon-design-system/carbon/pull/16933, the accordion does not animate when opening and closing items.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-accordion--default

### Steps to reproduce

Go to the story and click to open an item. It does not smoothly transition as it did before in 11.63.0: https://deploy-preview-17079--v11-carbon-react.netlify.app/?path=/story/components-accordion--default

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

alisonjoseph:
Closes #17197

This fixes a regression that happened when https://github.com/carbon-design-system/carbon/pull/17120 was merged in to remove items from the dom when the accordion was closed. The issue with that was you can't animate/transition between display none and block. This PR uses [@starting-style](https://caniuse.com/mdn-css_at-rules_starting-style) along with [allow-discrete](https://caniuse.com/?search=allow-discrete) to be able to transition between display block. It currently has support in most major browsers, with partial support in Firefox (it will transition to display none but not from display none), this doesn't break anything and is an improvement from the current state of no transition at all. 

#### Changelog

- Remove hidden from AccordionItem
- Remove old/unused mixins from accordion scss 
- Add @starting-style and allow-discrete styles for transition animation
- Update test to check for active classname, remove style import from test(this was the only test file we were doing this)

#### Testing / Reviewing

Check accordion story and make sure it transitions as expected in Chrome, check with keyboard and ensure that the buttons inside the test accordion item aren't accessible when the accordion is closed.


ariellalgilmore (approved):
LGTM!

kennylam (approved):

PR #17435, opened 2024-09-12
fix(17398): fix Datepicker issue with Keyboard Navigation in Firefox

### LINKED ISSUE #17398: [Bug]: Datepicker issue with Keyboard Navigation in Firefox
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v1.65.0

### React version

v18.3.1

### Description

1. With keyboard navigation able to select dates but the focus is not going to the next elements , the focus is always trapping inside date picker end date (Issue in only firefox browser other browsers working as expected) .
2. Selecting the start date and pressing the Tab key, the focus moves to the close tearsheet button/ outside of tearsheet and then returns to the start and end date fields ( Issue in all browsers ) .

### Reproduction/example

https://stackblitz.com/edit/github-pjo2fh-rm6wjm?file=package.json

### Steps to reproduce

1. Select the dates using keyboard and on press of tab key focus not going to the next element. (Issue in firefox)
2. Select the start date press Tab key , focus moves out then again on Tab key press it comes to start date then goes to end date .

### Suggested Severity

### Application/PAL

Accessible Workplace Connection

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/17398

This fixes the keyboard (Tab) navigation order in FireFox 

#### Changelog

added  event.preventDefault(); 

#### Testing / Reviewing
open deploy preview in FireFox
Open DatePicker -  "range with Calendar"  storybook example 
Select the dates using keyboard after selection of end date press tab and verify that the focus moves to the next focusable element in the DOM
Also verify in other browser

### DIFF
--- a/packages/react/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react/src/components/DatePicker/DatePicker.tsx
@@ -911,6 +911,7 @@ const DatePicker = React.forwardRef(function DatePicker(
         document.activeElement === endInputField.current &&
         calendarRef.current.isOpen
       ) {
+        event.preventDefault();
         calendarRef.current.close();
         // Remove focus from endDate calendar input
         document.activeElement instanceof HTMLElement && // this is to fix the TS warning


preetibansalui (approved):

emyarod (approved):
looks good to me on all browsers!

guidari:
Hey @2nikhiltom looks like the focus still blocked in the last element on Firefox

https://deploy-preview-17435--v11-carbon-react.netlify.app/?path=/story/components-datepicker--range-with-calendar

2nikhiltom:
Hi @guidari ! The issues are similar, we verified that the fix on this PR worked before merging. Seems like something has recently changed in firefox/flatpicker . could we track this via - [issue](https://github.com/carbon-design-system/carbon/issues/17663)

PR #17436, opened 2024-09-13
[Code Connect] Added `tagTitle` prop to DismissibleTag

guidari:
Added `tagTitle` to code connect.


2nikhiltom (approved):

alisonjoseph (changes requested):
I don't believe `"Tag title"` is a Figma prop, if we need the component to always have a that prop as an example you would want to add it directly. 

`example: ({ ...props }) => <DismissibleTag tagTitle="optional title" {...props} />,`

Screenshot 2024-09-13 at 7 57 35 AM

guidari:
> I don't believe `"Tag title"` is a Figma prop, if you just want the component to always have a that prop you would just add it in the example. ` example: ({ ...props }) => <DismissibleTag tagTitle="title" {...props} />,`

Ohh NVM then. I thought I had to add the prop to the Code Connect also.

PR #17444, opened 2024-09-14
feat(modal): add new prop to disable close button from modal (#17410)

### LINKED ISSUE #17410: [Feature Request]: Modal - Disable close button
### The problem

Allow users to disable/hide the close button on the modal. This will be useful when the modal's open status is managed by the state, like when the users cannot close or skip the modal without completing the functions (validations, forms, etc.) within it.

### The solution

new prop `disableClose`

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

CaiqueSobral:
Closes #17410

Added a new boolean prop to Modal to not display the close button.
#### Changelog

**New**

- Added a new boolean prop to Modal to not display the close button.

#### Testing / Reviewing

The close button from a modal should not be displayed if "closeButtonDisabled" prop is set in the modal root


CaiqueSobral:

tay1orjones (changes requested):
Hey thanks for suggesting this! Curious, why does `onRequestClose` not work for your use case? 

I'm not sure if disabling the close button is something we want to encourage from a ux or a11y perspective. I think the intent would be to use onRequestClose to stop the modal from closing when necessary and additionally adding an inline notification to inform the user why the modal can not be closed.

@laurenmrice what do you think?

If we land on this being a good idea we'll additionally need to override `preventCloseOnClickOutside` to be `true`.

CaiqueSobral:
@tay1orjones. It is not actually a case of mine, it is better explained in this issue https://github.com/carbon-design-system/carbon/issues/17410

tay1orjones:
@CaiqueSobral Ah, my mistake, I thought you were the original issue author. For now this isn't something we're going to add to Modal without strong reasoning due to the accessibility implications.

Thanks again for your work here, I'm sorry to close this on you. Issues labelled https://github.com/carbon-design-system/carbon/labels/proposal%3A%20open and/or https://github.com/carbon-design-system/carbon/labels/status%3A%20needs%20triage%20%F0%9F%95%B5%EF%B8%8F%E2%80%8D%E2%99%80%EF%B8%8F typically aren't ready to be worked yet. The https://github.com/carbon-design-system/carbon/labels/needs%3A%20community%20contribution list is a better one to pull from if you're looking to pick something up.

PR #17455, opened 2024-09-17
17450 popover with auto align is breaking

### LINKED ISSUE #17450: [Bug]: Popover with autoAlign true is breaking
### Package

@carbon/react

### Browser

_No response_

### Package version

1.66.0

### React version

_No response_

### Description

Popover is not working if autoAlign is true.
image

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-popover--experimental-auto-align

### Steps to reproduce

[image
](https://react.carbondesignsystem.com/?path=/story/components-popover--experimental-auto-align)

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

preetibansalui:
Closes #17450 

Popover with autoAlign true is breaking itself

#### Changelog

**Changed**

added forwardref in ToggletipContent

#### Testing / Reviewing

1. Go to Storybook > Popover > Experimental Auto Align
2. It should work as expected.


2nikhiltom (approved):
LGTM

guidari (approved):

packages/react/src/components/Popover/index.tsx
```diff
@@ -370,7 +370,6 @@ export const Popover: PopoverComponent = React.forwardRef(
         autoAlign && displayName && ['ToggletipButton'].includes(displayName);
       const isAllowedTriggerComponent =
         autoAlign &&
-        displayName &&
         !['ToggletipContent', 'PopoverContent'].includes(displayName);
```
tay1orjones, line 373: If `displayName` is undefined won't this throw a reference error?

packages/react/src/components/Toggletip/index.tsx
```diff
@@ -331,19 +331,18 @@ interface ToggletipContentProps {
...
  * be rendered inside of the popover for this component.
  */
-export function ToggletipContent({
-  children,
-  className: customClassName,
-}: ToggletipContentProps) {
+export const ToggletipContent = React.forwardRef<
+  HTMLDivElement,
+  ToggletipContentProps
+>(function ToggletipContent({ children, className: customClassName }, ref) {
```
tay1orjones, line 337: The newly added `ref` isn't used anywhere in this component, so it will [always be `null`](https://react.dev/reference/react/forwardRef#my-component-is-wrapped-in-forwardref-but-the-ref-to-it-is-always-null)

tay1orjones (changes requested):
Could you share more on the problem and how these changes fix it? A couple pieces don't immediately make sense to me but It's very possible I might be missing something! 😄

packages/react/src/components/Popover/index.tsx
preetibansalui: I test it, it's not giving any error. Also, find this on internet: 
_Reference errors occur only when you try to use a variable that hasn't been declared. Since displayName is declared, there’s no error._
It does not give a reference error because displayName is defined as a constant with the value undefined.

packages/react/src/components/Toggletip/index.tsx
preetibansalui: Thanks Taylor for the suggestion, Now passing the ref to the component and added a condition in parent element while passing ref to this one.

packages/react/src/components/Popover/index.tsx
```diff
@@ -370,11 +370,12 @@ export const Popover: PopoverComponent = React.forwardRef(
...
+      const renderComponentName = item?.type?.render?.name;
```
tay1orjones: While the `render` property corresponds to the original function component that forwardRef wraps, I don't think it's safe to use. Minifiers and bundlers can change the function names and this could/should break in production.

packages/react/src/components/Toggletip/index.tsx
```diff
@@ -331,19 +331,21 @@ interface ToggletipContentProps {
  * `children` passed in as a prop inside of `PopoverContent` so that they will
  * be rendered inside of the popover for this component.
  */
-export function ToggletipContent({
-  children,
-  className: customClassName,
-}: ToggletipContentProps) {
+export const ToggletipContent = React.forwardRef<
```
tay1orjones: I think we might be able to get `displayName` to be what we need/expect by applying it to the forwardRef directly.

```suggestion
const ToggletipContent = React.forwardRef<
```

This way the displayName can be explicitly set before the export lower in the file after the component proptypes

```
ToggletipContent.displayName = "ToggletipContent";
export { ToggletipContent }
```

Then the conditions can in theory be reverted to what they originally were, checking the `displayName`

tay1orjones (changes requested):
A couple thoughts below, let me know what you think

packages/react/src/components/Toggletip/index.tsx
preetibansalui: oh, that's better idea, Thanks for the suggestions Taylor, I will do the changes.

tay1orjones (approved):

PR #17457, opened 2024-09-17
fix: ComboBox readonly state fixed

### LINKED ISSUE #17382: [Bug]: Combobox Read-only state incorrectly allows interaction
### Package

@carbon/react

### Browser

Chrome

### Package version

1.65.0

### React version

18.2.0

### Description

Setting Combobox to readOnly state still allows selection

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-combobox--playground&args=readOnly:!true

### Steps to reproduce

Click on the selected value on the combobox.
Notice that the options are displayed
CLick on other options and notice that other option is getting selected

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #17382 

Fixed the Combobox's Read Only state, which previously would allow the user to interact with the list box.

This modification has been done after referring the read only states of Dropdown and Multiselect.
Dropdown's readonly story for reference:
https://react.carbondesignsystem.com/?path=/story/components-dropdown--playground&args=readOnly:!true

#### Changelog

**New**

- Added read only prop check to stop the list box opening.
- Added onFocus and onClick event handlers for readOnlyEventHandlers.
- Added `event.preventDefault();` for Tab interaction for readOnly prop

#### Testing / Reviewing

- Navigate to the playground story of Combobox component
- Turn on the readOnly prop
- check the interaction and make sure it doesn't open or allows text focus, only the box focus is present.


alina-jacob (approved):
Looks good Guru! 🚀

preetibansalui (changes requested):
Hi Gururaj, The readonly prop is working fine now. Only one thing I noticed in deploy preview is when we press Tab Key, The focus move to other elements which is fine but combobox is still in highlight state, which I believe should not be. Please check attached screen recording.

After that I am not able to remove highlight from combobox even by clicking outside as well.

https://github.com/user-attachments/assets/e460caa3-2bb0-45c4-a3a2-d4cb9fcca800

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -749,6 +755,15 @@ const ComboBox = forwardRef(
...
           },
+          onClick: (evt: MouseEvent<HTMLInputElement>) => {
+            // Prevent the default behavior which would open the list
+            evt.preventDefault();
+            // Focus on the element as per readonly input behavior
+            evt.currentTarget.focus();
+          },
+          // onFocus: (evt: FocusEvent<HTMLInputElement>) => {
+          //   evt.target.blur(); // Immediately remove focus when received
+          // },
```
kennylam: Looks good! Can remove this now

preetibansalui (approved):
Working fine now 👍

kennylam (approved):

PR #17467, opened 2024-09-18
fix: makes aria-label prop optional in TabelExpandHeader 

### LINKED ISSUE #17309: [Bug]: aria-label is required for TableExpandHeader instead of optional
### Package

@carbon/react

### Browser

Chrome

### Package version

1.65.0

### React version

17.0.2

### Description

aria-label is required for TableExpandHeader even though its only used when enableToggle is set to true. Will give an error if aria-label is removed
image

https://github.com/carbon-design-system/carbon/blob/009351acad1227291c713ebc18073995da20aeba/packages/react/src/components/DataTable/TableExpandHeader.tsx#L34
image

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datatable-expansion--default

### Steps to reproduce

https://react.carbondesignsystem.com/?path=/story/components-datatable-expansion--default

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/17309

Makes aria-label prop optional 

#### Changelog

aria-label prop optional 
updated test cases to use `aria-label` instead if deprecated  `ariaLabel` 

#### Testing / Reviewing

Verify of functionality is intact and changes makes sense

### DIFF
--- a/packages/react/src/components/DataTable/TableExpandHeader.tsx
+++ b/packages/react/src/components/DataTable/TableExpandHeader.tsx
@@ -31,7 +31,7 @@ type TableExpandHeaderPropsBase = {
    * Specify the string read by a voice reader when the expand trigger is
    * focused
    */
-  ['aria-label']: string;
+  ['aria-label']?: string;
 
   /**
    * @deprecated The enableExpando prop is being replaced by `enableToggle`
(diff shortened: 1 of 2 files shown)


emyarod (approved):
looks good to me!

Gururajj77 (approved):

riddhybansal (approved):

PR #17520, opened 2024-09-18
Interactive Tags with AILabel

guidari:
It was noticed that we don't have an example with AILabel in Interactive Tag like we do here in the [website](https://carbondesignsystem.com/components/tag/usage/#ai-presence).

**New**
- Added the possibility to added an AILabel inside the Interactive Tags;
- Added a new stories called With AI Label to show the examples.

#### Testing / Reviewing

- Make sure the changes makes sense


guidari:
I’m a little bit concerned about this addition because of the "nested button issue." I tested the screen reader, and everything seems fine. We might be able to overcome this by using non-interactive elements (`<div>` or `<span>`) styled and given `role="button"` attribute to simulate the appearance and behavior of a button.

PR #17578, opened 2024-09-25
fix(toast-notification): restore bottom margin on details container

### LINKED ISSUE #17523: [Bug]: Toast notifications without a subtitle are missing bottom margin
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.66.0

### React version

_No response_

### Description

Toast notification messages that have a long title, but not a subtitle, appear too short due to their lack of bottom margin. This is because the title only has a top margin. The bottom margin only appears when the notification has a subtitle.

| without a subtitle | with a subtitle |
| --- | --- |
| image | image |

We need to fix the spacing to work in both cases.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-notifications-toast--playground&args=caption:;title:This+is+a+really+long+notification+title;subtitle:subtitle

### Steps to reproduce

This issue can be observed in Storybook. Visit the above link or:
1. Visit the `ToastNotification` Storybook.
2. Remove the default caption and subtitle.
3. Add a long enough title so that it breaks into multiple lines.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

emyarod:
Closes #17523

This PR restores the bottom margin on the toast notification to the details div instead of the caption or subtitle. It also adds some visual regression tests

#### Changelog

**New**

- component test snapshots

**Changed**

- toast notification bottom margin placement

#### Testing / Reviewing

Confirm that the bottom margin of the toast notification does not collapse with different combinations of titles, subtitles, and captions of different length

### DIFF
--- a/packages/react/src/components/Notification/__tests__/Notification-test.js
+++ b/packages/react/src/components/Notification/__tests__/Notification-test.js
@@ -12,7 +12,7 @@ import {
   ToastNotification,
   InlineNotification,
   ActionableNotification,
-} from './Notification';
+} from '../Notification';
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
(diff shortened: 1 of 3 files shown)


alisonjoseph (approved):

ariellalgilmore (approved):
LGTM!

riddhybansal (approved):

PR #17609, opened 2024-09-30
Fix types of `SideNavMenuItem`

### LINKED ISSUE #17608: [Bug]: Incorrect types in `SideNavMenuItem`
### Package

@carbon/react

### Browser

Firefox

### Package version

1.67.0

### React version

18.2.0

### Description

Setting the `as` prop in `SideNavMenuItem` causes a type error, despite it being a valid prop.

### Reproduction/example

https://stackblitz.com/edit/github-zdptur?file=src%2FApp.tsx

### Steps to reproduce

Go to the repro and mouseover `as` to see the type error.

Screenshot 2024-09-30 at 15 38 46

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

frankharkins:
Closes #17608

Since `SideNavMenuItem` forwards its props to `Link`, I think the type should be `ComponentProps<typeof Link>`, plus the extra prop that `SideNavMenuItem` supports (`isActive`). While technically unnecessary for the type system, I've left the extra props in the declaration to keep their descriptions in the docs. 

#### Changelog

**Changed**

- Type of `SideNavMenuItemProps` correctly describes the props `SideNavMenuItem` supports. This change includes adding the "as" prop, which was supported by `SideNavMenuItem` but showed a type error when used.

#### Testing / Reviewing

I'm not sure how best to test. There's a repro in #17608. If I should do something else to make reviewing easier, please let me know.


emyarod (approved):
looks good to me!

guidari (approved):
LGTM!

PR #17614, opened 2024-09-30
test: increase test coverage to 100% for Tag

### LINKED ISSUE #17503: Increase `Tag` unit test coverage

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/17503

Increase test coverage to 100% for Tag component 

#### Changelog

Added new test cases 

#### Testing / Reviewing

Test coverage should increase
Verify that test makes sense


emyarod (approved):
looks good to me!

packages/react/src/components/Tag/Tag-test.js
```diff
@@ -6,11 +6,12 @@
  */
 
 import { Add } from '@carbon/icons-react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
```
kennylam:
```suggestion
import { render, screen } from '@testing-library/react';
```

Doesn't seem like this is being used? If needed consider [userEvent](https://testing-library.com/docs/user-event/intro/) instead.

2nikhiltom: `fireEvent` was not used anywhere ,Good catch ! Thanks !! 🙌

emyarod (approved):
looks good to me!

alisonjoseph (approved):

PR #17630, opened 2024-10-02
fix(TreeNode): preserve nested node expand state

### LINKED ISSUE #17221: [Bug]: Treeview - opened node do not remain open even when parent is collapsed and then opened again
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

1.63.2

### React version

18.2.0

### Description

Have my opened nodes remain open even if their parent is collapsed and then opened again. The nodes should remain in the same open or closed state.

### Reproduction/example

https://stackblitz.com/edit/github-3xhbda-38shez?file=src%2FApp.jsx,package.json

### Steps to reproduce

1. Expand node "A"
2. Then collapse it's parent node "One"
3. Expand node "One" and node "A" would be in collapsed state. 

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

emyarod:
Closes #17221

This PR updates the rendering logic of parent tree nodes. The nodes are no longer removed from the DOM when upstream nodes are collapsed so the expansion state of nested parent nodes is preserved

#### Changelog

**Changed**

- parent node rendering logic

#### Testing / Reviewing

confirm that expanded state of nested parent nodes remains even when its own parent nodes are collapsed/expanded

### DIFF
--- a/packages/react/src/components/TreeView/TreeNode.tsx
+++ b/packages/react/src/components/TreeView/TreeNode.tsx
@@ -356,11 +356,13 @@ const TreeNode = React.forwardRef<HTMLLIElement, TreeNodeProps>(
             {label}
           </span>
         </div>
-        {expanded && (
-          <ul role="group" className={`${prefix}--tree-node__children`}>
-            {nodesWithProps}
-          </ul>
-        )}
+        <ul
+          role="group"
+          className={classNames(`${prefix}--tree-node__children`, {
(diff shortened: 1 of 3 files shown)


tay1orjones (approved):
Nice!


https://github.com/user-attachments/assets/13021971-8525-4282-9590-088ffb81ef83

kennylam (approved):

PR #17641, opened 2024-10-03
feat: codemod to deprecate light prop

### LINKED ISSUE #13257: Replace `light` prop with `<Layer>`
In v11, a new [color layering model](https://carbondesignsystem.com/guidelines/color/overview#layering-model) was introduced. This replaces usage of the `light` prop with a new `Layer` component.

This issue tracks all steps needed to fully replace the `light` prop with `Layer`.

```[tasklist]
### Steps to fully replace the `light` prop with `Layer`
- [ ] https://github.com/carbon-design-system/carbon/issues/9185
- [ ] https://github.com/carbon-design-system/carbon/issues/10987
- [ ] https://github.com/carbon-design-system/carbon/issues/12776
- [ ] https://github.com/carbon-design-system/carbon/issues/17115
```

### LINKED ISSUE #17115: Write a `@carbon/react` codemod to assist consumers in removing the `light` prop from components and instead wrapping those components in a `<Layer>`

riddhybansal:
Closes #17115 
Closes #13257

This introduces a new codemod to remove the `light` prop from components and embed it within the `<Layer>` component. 

Changelog

**New**
* Added a new `deprecate-light-prop` codemod
* Added test fixtures for components using the `light` prop and tests to validate the transformation

Testing / Reviewing
* Ensure that the test fixtures accurately represent the before and after states of the codemod transformation
* Review the codemod logic to ensure it handles all edge cases and component variations
* Test the codemod on a variety of real-world code samples to confirm its effectiveness

To run the codemod locally and verify its functionality, use the following command:

```bash
npx @carbon/upgrade migrate deprecate-light-prop --write

### CODE REVIEW

File: packages/react/src/components/ComboBox/ComboBox.tsx [hunk 1]
```diff
@@ -796,184 +797,188 @@ const ComboBox = forwardRef(
...
-        <ListBox
-          onFocus={handleFocus}
-          onBlur={handleFocus}
-          className={className}
-          disabled={disabled}
-          invalid={invalid}
-          invalidText={invalidText}
-          invalidTextId={invalidTextId}
-          isOpen={isOpen}
-          light={light}
```
tay1orjones, line 808: The codemod shouldn't be ran against our own code - we need to retain the light prop in the source of components so it will continue to work until we can remove it in v12.

packages/upgrade/src/upgrades.js
```diff
@@ -331,6 +331,40 @@ export const upgrades = [
           });
         },
       },
+      {
+        name: 'deprecate-light-prop',
```
tay1orjones, line 335: What if we renamed this to match the action being performed? The codemod doesn't deprecate the light prop in the users code, it more-so "removes" it or "refactor"s it. Maybe `refactor-light-to-layer`?

packages/upgrade/src/upgrades.js
```diff
@@ -331,6 +331,40 @@ export const upgrades = [
...
+        description: `
+         Deprecate the light prop and wrap components with Layer
```
tay1orjones, line 337:
```suggestion
         Refactor 'light' prop usage to instead wrap components with Layer
```

packages/upgrade/transforms/__testfixtures__/deprecate-light-prop.output.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Button, Layer } from '@carbon/react';
+
+function TestComponent() {
+  <div>
+    <Layer><Button>Click me</Button></Layer>
+    <Button>Another button</Button>
+    <Layer>
+      <Button>Nested light button</Button>
+    </Layer>
```
tay1orjones, line 10: The input for this piece in the other file is

```
    <Layer>
      <Button light>Nested light button</Button>
    </Layer>
```

If they're still using `light` inside a `Layer` would we want to nest it inside another layer instead?

```suggestion
    <Layer>
      <Layer>
        <Button>Nested light button</Button>
      </Layer>
    </Layer>
```

but I might be wrong here, not sure if you already tested this. Let me know what you find.

tay1orjones (changes requested):
This is awesome! 🔥 

I think we'll need to revert all the changes to the component source files, and I have a couple questions/comments.

riddhybansal:
Hey I have another PR for the same , there were issues with the conflict . Check Here :#18070

PR #17659, opened 2024-10-07
fix: ailabel autoalign was not working

### LINKED ISSUE #17571: AILabel `autoAlign` does not work
https://github.com/carbon-design-system/carbon/issues/17017#issuecomment-2369364831

preetibansalui:
Closes #17571 

AILable auto align was not working

#### Changelog

**New**

Added `shift` in middleware of popover component. It prevents the floating element from overflowing along its axis of alignment, thereby preserving the side it’s placed on.

#### Testing / Reviewing

1. Go to storybook > AILable > Default Story
2. Click on AI Label, AILabelContent should become fully visible and avoid overflowing along its axis..


packages/react/src/components/Popover/index.tsx
```diff
@@ -558,7 +560,7 @@ function PopoverContentRenderFunction(
   const prefix = usePrefix();
   const { setFloating, caretRef, autoAlign } = React.useContext(PopoverContext);
   const ref = useMergedRefs([setFloating, forwardRef]);
-
+  // console.log(refs)
```
alisonjoseph:
```suggestion
```

alisonjoseph (approved):
LGTM, just a small code comment to cleanup

tay1orjones:
Technically this doesn't follow the design spec for Popover. The caret placement is dynamic and isn't using one of the allowed/specified alignments for Popover.



This is from the original issue:



Do you know why does it not reflow to use an alignment option where the caret is on the left top like this from the spec?
2024-10-07 at 14 37 59-Popover – Carbon Design System-Google Chrome@2x

preetibansalui:
> Technically this doesn't follow the design spec for Popover. The caret placement is dynamic and isn't using one of the allowed/specified alignments for Popover.
> 
> 
> This is from the original issue:
> 
> 
> Do you know why does it not reflow to use an alignment option where the caret is on the left top like this from the spec? 2024-10-07 at 14 37 59-Popover – Carbon Design System-Google Chrome@2x

Hi Taylor, Thanks It didn't comes to my mind about design spec guidelines. Thanks for pointing this out.

> Do you know why does it not reflow to use an alignment option where the caret is on the left top like this from the spec? !

Its a very good question and it actually blew my mind. I think about this a lot and found if it will be right align then also it would hide the AI Label content from top part. pls check the attached picture. Also, its same for popover and toogletip components. The problem doesn't get noticed for them bcs their content window is small in size as compared to AILabel so it get adjusted in either direction. 
image

Please suggest what is the correct way to tackle the situation when available window size would hide the content in atleast one direction.

Note: by default it try to take left-top direction as we are providing `fallbackAxisSideDirection: 'start'` to flip in middleware.

tay1orjones (approved):

PR #17669, opened 2024-10-07
fix(date-picker): clearing input and selecting a date

### LINKED ISSUE #14529: [Bug]: If the DatePicker calendar is opened, and clear the currently selected date, then try selecting another date requires two clicks instead of one.
### Package

### Browser

Chrome, Safari, Firefox, Edge

### Package version

10.56.0

### React version

>=16.14.0

### Description

If the DatePicker calendar is opened, and clear the currently selected date, then we are experiencing an issue where selecting another date requires two clicks instead of one. In other words, the new date selection does not register with just one click; we need to click twice for the new date to be selected. If we don’t clear the current selected date then issue is not coming.

https://github.com/carbon-design-system/carbon/assets/136786799/d0f537fb-cbca-4384-9e8b-f6c6134f881b

### Reproduction/example

https://v7-react.carbondesignsystem.com/?path=/story/components-datepicker--single

### Steps to reproduce

https://v7-react.carbondesignsystem.com/?path=/story/components-datepicker--single
Select date picker it will open the calendar. Clear the input field and try to select the another date. It need 2 clicks to select another date.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

ariellalgilmore:
Closes #14529

When opening the date-picker, then clearing the input in date-picker and then clicking on a value the date-picker would still show empty. 

https://github.com/user-attachments/assets/11c35e5d-81b1-4178-b050-29791537fc10

https://github.com/user-attachments/assets/084f4c4f-7f58-4e17-82be-0f1371a27786

#### Changelog

**Changed**

- updated handleOnChange and added a keyPress event listener

#### Testing / Reviewing

make sure tests still pass. make sure works similiarly to flatpickr examples

Open storybook
- go to date-picker stories
- open the date picker select a date
- open again and clear the current date
- while it is still open click on a date
- date-picker should hold that current value in the input


2nikhiltom (approved):
Neat fix 🔥

tay1orjones (approved):
LGTM!

Issue #12494, opened 2022-11-03
Deprecate end-of-support v10 assets


tay1orjones:
This issue and all child issues within are part of the plan to deprecate `carbon-components` and `carbon-components-react` as originally proposed in the GitHub discussion:

- https://github.com/carbon-design-system/carbon/discussions/12179

This also includes v10.x versions of all packages in the monorepo, @carbon/colors, etc.

```[tasklist]
### Tasks
- [x] https://github.com/carbon-design-system/carbon/issues/12495
- [ ] https://github.com/carbon-design-system/carbon/issues/12496
- [x] Deprecate v10.x versions of elements packages on npm
- [ ] https://github.com/carbon-design-system/carbon/issues/12497
- [ ] https://github.com/carbon-design-system/carbon-website/issues/4112
- [ ] https://github.com/carbon-design-system/carbon/issues/17374
- [ ] https://github.com/carbon-design-system/carbon/issues/17633
- [x] Deprecate any straggler versions published after the initial deprecation for all impacted packages
```

GraceeFlower:
Hi team, could you please provide the rough DDL of the v12 release which will totally deprecate `carbon-components-react`? Thank you!

PR #17672, opened 2024-10-07
chore(project): ensure end-of-support packages are not published anymore

### LINKED ISSUE #17633: Remove deprecated/end-of-support packages from lerna version/publish flow
`carbon-components` and `carbon-components-react` have been marked as deprecated on npm and will not receive any more updates. 

This needs to be enforced in our tooling though. I believe the best/easiest way to do this without removing the files from the repo is to mark them as `private: true` so they won't be published via lerna anymore.

tay1orjones:
Closes https://github.com/carbon-design-system/carbon/issues/17633
Part of #12494 

#### Changelog

**Changed**

- Mark `carbon-components` and `carbon-components-react` as `private` so they will not be published to npm

#### Testing / Reviewing

- You can try a local dry run of publishing to npm if you want to, but we'll be able to do a true test of this live when we release this week.

### DIFF
--- a/packages/carbon-components-react/package.json
+++ b/packages/carbon-components-react/package.json
@@ -1,5 +1,6 @@
 {
   "name": "carbon-components-react",
+  "private": true,
   "description": "The Carbon Design System is IBM’s open-source design system for products and experiences. This package reached end of support on September 30, 2024 and will not receive any more updates.",
   "version": "8.68.0-rc.0",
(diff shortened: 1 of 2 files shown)


alisonjoseph (approved):
👍

ariellalgilmore (approved):

riddhybansal (approved):

PR #17673, opened 2024-10-07
fix(staticnotification): remove staticnotification from top of sidebar

tay1orjones:
Fixes the sidebar so StaticNotification docs page sits in it's proper spot in the left nav instead of being at the top 

image

Related https://github.com/carbon-design-system/carbon/pull/17443
#### Changelog

**Changed**

- Mark the docs .mdx page as a template so storybook handles it properly

#### Testing / Reviewing

- Deploy preview should show that it sits in the correct spot and is not like the above screenshot anymore


alisonjoseph (approved):
LGTM, thanks!

Gururajj77 (approved):

PR #17622, opened 2024-10-01
Fix DefinitionTooltip not opening on first click

frankharkins:
Closes #17070

The problem was `onFocus` opened the tooltip, then the `onClick` from the same event immediately closed it again.

This PR proposes a fix by changing `onClick` to `onMouseDown`, which occurs before `onFocus`.

#### Changelog

**Changed**

- Fixed bug whereby `DefinitionTooltip` with `openOnHover=false` immediately closed the first time it was clicked.

#### Testing / Reviewing

Open storybook, go to `DefinitionTooltip`, set `openOnHover` to `false`, then click the component.

https://github.com/user-attachments/assets/f1df60a3-9099-4f7e-b1d5-e11ae10596e7

PR #17679, opened 2024-10-08
test(DefinitionTooltip): fix unit tests and add coverage  for DefinitionTooltip

neuescodes:
**Changed**

- add and fix unit tests around tooltip opening functionality (`toBeVisible` was not working as expected in this scenario, I am instead checking for the `aria-expanded` attribute).


ariellalgilmore:
hi @Neues, thanks for the contribution! I believe this was already fixed with https://github.com/carbon-design-system/carbon/pull/17622

packages/react/src/components/Tooltip/__tests__/DefinitionTooltip-test.js
```diff
@@ -11,11 +11,31 @@ import React from 'react';
...
+    expect(screen.getByRole('button')).toHaveAttribute(
+      'aria-expanded',
+      'false'
+    );
     await userEvent.click(screen.getByText('URL'));
-    expect(screen.getByText(definition)).toBeVisible();
+    expect(screen.getByRole('button')).toHaveAttribute('aria-expanded', 'true');
+  });
+
+  it('should control tooltip display when using keyboard event', async () => {
```
ariellalgilmore: I believe this test is already being handled here: https://github.com/carbon-design-system/carbon/pull/17679/files#diff-97685f64e2290de562987881b78ac2d0c7c60611353b92ae00de186126cef815R52-R70

packages/react/src/components/Tooltip/__tests__/DefinitionTooltip-test.js
```diff
@@ -11,11 +11,31 @@ import React from 'react';
...
+  it('should display on click a definition provided via prop', async () => {
     const definition = 'Uniform Resource Locator';
     render(<DefinitionTooltip definition={definition}>URL</DefinitionTooltip>);
+    expect(screen.getByRole('button')).toHaveAttribute(
+      'aria-expanded',
+      'false'
+    );
     await userEvent.click(screen.getByText('URL'));
-    expect(screen.getByText(definition)).toBeVisible();
+    expect(screen.getByRole('button')).toHaveAttribute('aria-expanded', 'true');
```
ariellalgilmore, line 22: wondering if we can keep the check `toBeVisible` along with checking if `aria-expanded` is true

ariellalgilmore (approved):
looking good! just some small comments

packages/react/src/components/Tooltip/__tests__/DefinitionTooltip-test.js
neuescodes: Oops, yeah it is! I've removed mine.

packages/react/src/components/Tooltip/__tests__/DefinitionTooltip-test.js
neuescodes: I was debating leaving it but ultimately decided to remove it like I removed the others for the reason that it does not actually work as expected in this case; toBeVisible() returns true even when the tooltip is not expanded. I replaced the other instances where it was being used, and then elected to remove this one so that other devs don't get confused.

Neues:
> hi @Neues, thanks for the contribution! I believe this was already fixed with #17622

Sorry about this! I should've caught that, will try and be more mindful of duplicate issues open like this in the future.

packages/react/src/components/Tooltip/__tests__/DefinitionTooltip-test.js
tay1orjones: Yeah I've seen this before too, the `aria-expanded` does seem like the best workaround 👍

tay1orjones (approved):

ariellalgilmore (approved):

guidari (approved):

PR #17683, opened 2024-10-08
fix: use plex as per family

### LINKED ISSUE #11661: Plex is not up to date, package is too large

riddhybansal:
Closes #11661 

Updates `@carbon/styles`  to have flexibility of using both individual `@ibm/plex-${font-style}` packages and `@ibm/plex` by using a variable `use-per-family-plex` which is false by default.

#### Changelog

**New**

- 8 new packages for each of the `@ibm/plex` families 

Screenshot 2024-07-25 at 9 17 15 AM

- `font-feature-settings: 'liga' 1;` added to `_reset.scss` to ensure ligatures render even with `letter-spacing` set
- `$package-name` field added to font src resolver

**Changed**

- Updated the `url` field in font src resolver to use new package name format when `use-per-family-plex` is true

**Removed**

- Split fonts do not seem to ship `.woff`, so [they have been removed](https://github.com/carbon-design-system/carbon/pull/17025/files#diff-95fa1084e7c21be991e52f92b77cc405b040c0356534cd670dc5db7708739dafR16) from the individual font formats

#### Testing / Reviewing

Ensure the fonts are loaded correctly, either using the Akamai CDN or not. Ensure the `fi` ligature renders properly.


packages/styles/scss/_config.scss
```diff
@@ -43,6 +43,13 @@ $font-display: 'swap' !default;
 /// @group config
 $font-path: '~@ibm/plex' !default;
 
+/// The value used to prefix selectors and CSS Custom Properties across the
+/// codebase
```
tay1orjones:
```suggestion
/// Specify if the new per-family plex packages (e.g. @ibm/plex-sans) should be used
```

tay1orjones (changes requested):
This looks really good!🥳 

It would be nice if the existing documentation could be updated to reflect this change and surface the new config option. I'm specifically thinking about the .md we have in the repo that covers how to self-host plex and write your own font resolver, etc. 

Probably need some new docs as well covering stuff like:


* What happens if both `$use-akamai-cdn` and `$use-per-family-plex` are set to true?
* How would someone migrate from `$use-akamai-cdn`?
* How would someone migrate from basic/existing `$font-path` usage?
* What will be the default in the next major

tay1orjones:
If you'd like to put the documentation work in a different issue/PR, I'd be fine approving and merging this 👍

ariellalgilmore (approved):

docs/guides/ibm-plex.md
```diff
@@ -39,9 +44,19 @@ document.
 
 ### `@ibm/plex`
 
-IBM Plex is distributed through a package on NPM called
-[`@ibm/plex`](https://www.npmjs.com/package/@ibm/plex). The source for this
-package is available on [GitHub](https://github.com/ibm/plex#readme).
+IBM Plex is distributed through a package on NPM:
+
+- [`@ibm/plex`](https://www.npmjs.com/package/@ibm/plex): The complete package
+  containing all fonts
```
tay1orjones:
```suggestion
- [`@ibm/plex`](https://www.npmjs.com/package/@ibm/plex): The legacy package
  containing all fonts. This package is no longer updated and uses an outdated version of plex.
```

docs/guides/ibm-plex.md
```diff
@@ -220,6 +235,110 @@ To quickly load IBM Plex Sans, you could use the following `@font-face` block:
...
+### Migration Guide
+
+#### Migrating from Akamai CDN
+
+1. Install the required Plex family packages:
+
+```bash
+# Install only the font families you need
+npm install @ibm/plex-sans @ibm/plex-mono # Add other families as needed
+```
```
tay1orjones: This isn't necessary because the per-family packages are listed as hard dependencies in `@carbon/styles`
```suggestion
```

docs/guides/ibm-plex.md
```diff
@@ -220,6 +235,110 @@ To quickly load IBM Plex Sans, you could use the following `@font-face` block:
...
+
+```scss
+// Before (if you had specific imports)
+@import '@carbon/type/scss/font-face/mono';
+@import '@carbon/type/scss/font-face/sans';
+
+// After (using per-family packages)
+@import '@ibm/plex-mono/scss/index';
+@import '@ibm/plex-sans/scss/index';
+```
```
tay1orjones: Same thing here, this isn't necessary. To move off akamai they just need to remove `$use-akamai-cdn` and add `$use-per-family-plex: true;`

docs/guides/ibm-plex.md
```diff
@@ -220,6 +235,110 @@ To quickly load IBM Plex Sans, you could use the following `@font-face` block:
...
+
+In the next major version:
+
+- `$use-per-family-plex: true` will be the default configuration
+- The monolithic `@ibm/plex` package will be deprecated in favor of individual
+  family packages
+- Backward compatibility will be maintained:
+  - Akamai CDN option (`$use-akamai-cdn`) will continue to work
+  - Custom font path (`$font-path`) will remain supported
+  - Migration will not be required but recommended
```
tay1orjones:
```suggestion
```

tay1orjones (changes requested):
A couple things I saw - I'll work on a small refactor and push the changes to this PR

tay1orjones:
I pushed an update here, let me know what you think!

riddhybansal:
> I pushed an update here, let me know what you think!

Oh I see what you have done here !! These details will be helpful.

tay1orjones (approved):

PR #17706, opened 2024-10-09
Moving Interactive Tags to stable

guidari:
Closes #17694 

- Move Interactive Tags under Tag in storybook.
- Removed `slug` from Interactive Tags.
- Removed Playground story and added controls to Read Only story.
- Changed Overview page.

#### Testing / Reviewing

- All stories should work

PR #17711, opened 2024-10-10
[Unit Test] - Selectable Tag 

### LINKED ISSUE #17437: Add test coverage for SelectableTag

guidari:
Closes #17437 

Added test to Selectable Tag.
We can't cover 100% because of this topic: https://ibm-studios.slack.com/archives/C03430J3DBP/p1726857596307079
And we also are gonna remove slug in this [PR](https://github.com/carbon-design-system/carbon/pull/17706), so we don't have to cover that
#### Testing / Reviewing

- Test should be at 80.95%


packages/react/src/components/Tag/Tag-test.js
```diff
@@ -67,6 +68,30 @@ describe('Tag', () => {
     ).toBeInTheDocument();
   });
 
+  describe('Selectable Tag', () => {
+    it('should render a selectable tag', () => {
+      const { container } = render(<SelectableTag text="Tag content" />);
+
+      const selectableTag = container.querySelector(
+        `.${prefix}--tag--selectable`
+      );
```
tay1orjones: I commented this in another PR too, but would it make sense to use `container.firstChild` instead?

tay1orjones (approved):

annawen1 (approved):

PR #17727, opened 2024-10-11
feat(floating-ui): add feature flag for dynamic floating-ui styles

### LINKED ISSUE #17018: Floating UI: Allow floating-ui-based positioning without `autoAlign` collision detection
Right now to take advantage of the `position: fixed` strategy used by `floating-ui`, users must use `autoAlign`, which includes collision detection. 

In theory these two could be separated, allowing consumers to take advantage of the benefits of fixed positioning, most notably avoiding [clipping of ancestor elements](https://floating-ui.com/docs/misc#clipping), without having to use `autoAlign` collision detection.

Open questions
* How will consumers opt-in to this behavior? Adding another prop is everyone's first thought, but is there a way to do this in a more streamlined way?
* What does this mean for the popover/tooltip etc alignment styles? All approaches should be heavily VRT to prevent regressions across the three options.

Examples
* [Initial request](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1721424171905469?thread_ts=1710783915.774909&cid=C2K6RFJ1G)

annawen1:
Closes #17018 

This PR introduces a new feature flag (`enable-v12-dynamic-floating-styles`) to separate the `@floating-ui` logic. By turning on the flag, components will have dynamically positioned floating styles without the collision detection currently in place. 

To have the collision detect logic, adopter will set the `autoAlign` prop to true which will automatically include the dynamic styling - so the feature flag is not necessary in this case.

#### Changelog

**New**

- `enable-v12-dynamic-floating-styles` feature flag

**Changed**

- add checks in current components utilitizing `@floating-ui` to separate out the style and middleware logic

#### Testing / Reviewing

- go to deploy preview, `Feature flags` section
- go to `Dynamic floating styles` folder and inspect the components and see that the styles are set dynamically


emyarod (approved):
looks good to me!

ariellalgilmore (approved):

PR #17738, opened 2024-10-14
feat(icons): new AI label icon

conradennis:
Submitting a new AI label icon per the request of @aagonzales all updates and additions approved by BXD

Note: we may want to restrict the download option for this icon. Anna can follow up on that conversation.

1 new icon

updated categories.yml with addition
updated icons.yml with addition

#### Changelog

**New**

- 32/ai-label.svg


laurenmrice (approved):

alisonjoseph (approved):
LGTM - unsure what you mean about restricting the download of the icon, once we add it in anyone will be able to access it.

aagonzales:
@alisonjoseph Adam Cutler wants to be able to see the AI Label icon along-side the other AI icons in the icon library but ideally we only want people using the [AI Label](https://react.carbondesignsystem.com/?path=/docs/components-ailabel--overview) component and not this icon svg. 

Mike's original proposal for circumnavigating this would be to replace the "Download" actions with a "launch" action to view the component in Storybook instead.

aagonzales:
Figuring out the website issue doesn't need to block adding the icon to the pack though. So I think we're good to merge this in.

alisonjoseph:
@aagonzales oh ok, we can do something like that if you want to create an issue. Also looking at your screen shot I swear those hover states for download icons used to be square.

aagonzales:
OK I've opened an issue: https://github.com/carbon-design-system/carbon-website/issues/4323

PR #17740, opened 2024-10-14
feat(grid): add align start and align end prop

### LINKED ISSUE #17245: [Feature Request]: Property for Grid to change the alignment
### The problem

Currently the grid is centered by default. [Guidelines for product pages](https://carbondesignsystem.com/elements/2x-grid/usage#style-models) require them to be left aligned. In order to comply with the requirement devs need to override the CSS for the grid from `margin-inline: auto;` to `margin-inline: 0;`.

### The solution

Add a property for the grid which will allow to change the alignment in React. `centered="false"` for example.

### Code of Conduct

alisonjoseph:
Closes #17245

Add new align prop to Grid, update tests for Grid and FlexGrid

#### Changelog

**New**

- `align` prop
- Add tests for Grid component

#### Testing / Reviewing
View default Grid story on wide monitor to see align start and align end examples. (revert story demo before merge)


emyarod (approved):
looks good to me!

tay1orjones (approved):

packages/react/src/components/Grid/__tests__/Grid-test.js
```diff
@@ -7,9 +7,71 @@
...
 describe('Grid', () => {
+  let Grid;
+  let cleanup;
+  let render;
+  let screen;
+
+  beforeEach(() => {
+    jest.resetModules();
+    const FeatureFlags = require('@carbon/feature-flags');
+    FeatureFlags.enable('enable-css-grid');
```
tay1orjones, line 63: IIRC This is true by default

packages/react/src/components/Grid/__tests__/Grid-test.js
```diff
@@ -7,9 +7,71 @@
...
+
+    cleanup = require('@testing-library/react/pure').cleanup;
+    render = require('@testing-library/react/pure').render;
+    screen = require('@testing-library/react/pure').screen;
+    Grid = require('../Grid').Grid;
```
tay1orjones, line 68: Hey I'm curious, is this a new pattern we need to follow in certain situations? I know there's been some changes like this in best practices for `@testing-library` but haven't seen this before.

packages/react/src/components/Grid/__tests__/Grid-test.js
alisonjoseph: It should be, without adding this however it was testing the flex grid version of the component. I realized after seeing the tests for the new prop fail and console.logging the output. I wonder if it could be avoided here if I import and test CSSGrid instead? 

https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Grid/Grid.tsx#L16-L20

packages/react/src/components/Grid/__tests__/Grid-test.js
alisonjoseph: I actually pulled this code from the Column-test file when looking for an example of how we were enabling the feature flag within tests.

PR #17751, opened 2024-10-15
fix(Dropdown): remove extra space at the end

### LINKED ISSUE #17750: [Bug]: Dropdown extra space at the end causes mis-alignments
### Package

@carbon/react

### Browser

Chrome

### Package version

1.68.0

### React version

_No response_

### Description

When using Dropdown with settings (Inline/ titleText = "") there is extra space at the end caused by `grid-gap` making it hard to align properly in real use cases

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-dropdown--playground&args=titleText:;type:inline

### Steps to reproduce

1- Make sure titleText = "" and type="inline"
2- Inspect element:

Image

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tekno0ryder:
Closes #17750

#### Changelog

**Changed**

- Changed from `grid` to `flex` so that it only add `gap` if there is more than 1 child element to the wrapper.
  - are there any drawbacks to this change?

**Removed**

- Removed `.#{$prefix}--form-requirement` it seems not used since there is no reference from `Dropdown.tsx`

#### Testing / Reviewing

No other inline elements:
Screenshot 2024-10-15 at 7 29 47 PM

When there are other inline elements:
Screenshot 2024-10-15 at 7 07 10 PM

### DIFF
--- a/packages/styles/scss/components/dropdown/_dropdown.scss
+++ b/packages/styles/scss/components/dropdown/_dropdown.scss
@@ -27,7 +27,7 @@
     display: inline-grid;
     align-items: center;
     grid-gap: 0 convert.to-rem(24px);
-    grid-template: auto auto / auto min-content;
+    grid-template: auto / auto min-content;
 
     .#{$prefix}--label {
       @include type-style('body-compact-01');


tekno0ryder:

alisonjoseph (changes requested):
This breaks the error and warning states for the component. You can go to the playground story and set inline to true. Form requirement classes are used to display the error and warning messages, so those shouldn't be removed. Will probably need a different approach to fix this bug.

Also you'll probably want to use the `hideLabel` prop vs. setting the title to empty for a11y purposes. (this doesn't fix the grid gap issue however)

tekno0ryder:
@alisonjoseph Thanks for the review, have another I look I found relying on grid system to add more rows when needed works in all cases.

Screenshot 2024-10-24 at 10 42 12 PM

Screenshot 2024-10-24 at 10 41 50 PM

Screenshot 2024-10-24 at 10 23 30 PM

alisonjoseph (commented):
LGTM! Thanks

alisonjoseph (approved):

ariellalgilmore (approved):

PR #17752, opened 2024-10-15
fix(combobox): odd cursor behavior

### LINKED ISSUE #17222: [Bug]: Odd Cursor Behaviour While Using ComboBox Component
### Package

@carbon/react

### Browser

Chrome, Firefox

### Package version

v1.63.2

### React version

v18.3.1

### Reproduction/example

### Suggested Severity

### Application/PAL

IBM Data Replication

### Code of Conduct

ariellalgilmore:
Closes #17222 

Odd cursor behavior in combobox when typing "abc" then moving cursor between "a" and "b" the text was appear as "a1bc23" instead of "a123bc".

https://github.com/user-attachments/assets/c5cde70e-b0a3-4ea7-8b06-f1683df4d49d

resources: https://github.com/downshift-js/downshift/blob/master/src/hooks/useCombobox/README.md?plain=1#L581-L611

#### Changelog

**Changed**

- removed cursor positioning and added setInputValue(newValue) on the change: 

#### Testing / Reviewing

Check combobox storybook


Gururajj77 (approved):
This looks good, thanks for refining this @ariellalgilmore

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -797,12 +795,6 @@ const ComboBox = forwardRef(
       downshiftSetInputValue,
       toggleMenu,
     ]);
-    useEffect(() => {
-      if (inputRef.current) {
-        inputRef.current.setSelectionRange(cursorPosition, cursorPosition);
-      }
-    }, [inputValue, cursorPosition]);
-
```
tay1orjones, line 805: Any idea what this was used for originally? It's wild that this whole useEffect was in there that not only wasn't needed, but actively caused a bug!

tay1orjones (approved):

riddhybansal (commented):
This LGTM !!

packages/react/src/components/ComboBox/ComboBox.tsx
ariellalgilmore: the cursorPosition came from the recent Autocomplete with Typeahead PR, but the odd cursor behavior was happening before that. When trying to fix the bug though, these additions seemed to probably help a weird cursor position when adding the autocomplete, so I realized fixing the original cursor bug also meant we could get rid of these new additions

PR #16617, opened 2024-05-29
Stylelint16

lee-chase:
Closes #16604 

Updates to stylelint 16

#### Testing / Reviewing

Running the script `lint:styles`

NOTE: Still work to do, many stylistic rules deprecated in 15 gone in 16.

See - https://stylelint.io/migration-guide/to-15#deprecated-stylistic-rules

PR #17770, opened 2024-10-16
fix(deps): update stylelint and related deps to v16

### LINKED ISSUE #16604: [Feature Request]: Update stylelint-config-carbon for Stylelint 16
### The problem

Stylelint has switched to ESModules as have many maintained packages.

This has a knock on effect of requiring various other packages to be updated including some internal ones to replace module.exports, require etc.

It may even expand to update Carbon to ESM.

### The solution

Update the Carbon packages to use Stylelint 16.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/16604

Migrated to `Stylelint 16.0.0`
Adjusted the `stylelint-config-carbon` package to reflect changes in accessibility rules.
Converted all relevant code from `CommonJS` to `ESModules`
Removed Deprecated Rule Sets
Removed Unnecessary Disabled Rules
Reference PR- https://github.com/carbon-design-system/carbon/pull/16617

#### Testing / Reviewing

 yarn lint:styles


2nikhiltom:
[[web-components]: deprecate side-panel and tearsheet components](https://github.com/carbon-design-system/carbon/pull/17772) - This PR removed the side-panel and tearsheet components

packages/styles/scss/_reset.scss
```diff
@@ -150,8 +150,6 @@
   blockquote::after,
   q::before,
   q::after {
-    content: '';
-    // stylelint-disable-next-line declaration-block-no-duplicate-properties
```
tay1orjones, line 154: I was curious why this was this way. Interesting: https://stackoverflow.com/a/6803278 - looks good to remove this 👍

packages/web-components/package.json
```diff
@@ -127,6 +127,9 @@
     "extends": [
       "../../config/stylelint-config-carbon"
     ],
+    "rules": {
+      "max-nesting-depth": 3
+    },
```
tay1orjones: This is set to null like 8 lines below this, which of the two should be removed?

Outside of this PR we should see if/how we could get this to follow the coding style standards defined in the dev handbook.

https://github.com/carbon-design-system/carbon/blob/c00b460545cd0376e857bf7ed176b01ef73d4c7c/docs/developer-handbook.md#L324-L345

The default value for this within `stylelint-config-carbon` is zero with exceptions for some `@` rules

package.json
```diff
@@ -143,7 +144,10 @@
   "stylelint": {
     "extends": [
       "./config/stylelint-config-carbon"
-    ]
+    ],
+    "rules": {
+      "max-nesting-depth": 3
+    }
```
tay1orjones: Could you share more on why this is in package.json and if/how it interacts with where it's defined within `styleling-config-carbon`?

https://github.com/carbon-design-system/carbon/blob/e063a4c4478227e48e28c71592779b9184df46e6/config/stylelint-config-carbon/rules/limit-language-features.js#L162-L169

packages/styles/scss/components/tag/_tag.scss
```diff
@@ -401,7 +401,7 @@
   }
 
   // Windows HCM fix
-  /* stylelint-disable */
```
tay1orjones, line 404: So awesome to see stuff like this removed 🎉

packages/styles/scss/components/popover/_popover.scss
```diff
@@ -62,7 +62,7 @@ $popover-text-color: custom-property.get-var(
 // The distance between the popover container and the triggering element
 // Specify the distance between the popover and the trigger. This value must
 // have a unit otherwise the `calc()` expression will not work
-// stylelint-disable-next-line length-zero-no-unit
+//  length-zero-no-unit
```
tay1orjones:
```suggestion
```

config/stylelint-config-carbon/index.js
```diff
@@ -7,17 +7,16 @@
...
   extends: [
-    require.resolve('stylelint-config-standard'),
-    require.resolve('stylelint-no-unsupported-browser-features'),
-    require.resolve('stylelint-config-idiomatic-order'),
-    require.resolve('stylelint-config-standard-scss'),
-    require.resolve('./rules/possible-errors'),
-    require.resolve('./rules/limit-language-features'),
-    require.resolve('./rules/stylistic-issues'),
-    require.resolve('./plugins/a11y'),
-    require.resolve('./plugins/prettier'),
```
tay1orjones, line 20: Why is `./plugins/prettier` being removed?

config/stylelint-config-carbon/package.json
```diff
@@ -3,6 +3,7 @@
   "description": "Stylelint configuration for Carbon",
   "version": "1.19.0",
   "license": "Apache-2.0",
+  "type": "module",
```
tay1orjones, line 6: We'll want to cut a new major for this package once this PR is in. We can pair up and do it together if you'd like!

config/stylelint-config-carbon/package.json
```diff
@@ -24,22 +25,21 @@
     "provenance": true
   },
   "peerDependencies": {
-    "stylelint": "^15.0.0"
+    "stylelint": "^16.0.0"
   },
   "dependencies": {
-    "stylelint-a11y": "^1.2.3",
+    "@double-great/stylelint-a11y": "^3.0.2",
```
tay1orjones, line 31: I looked up the context for this for posterity: https://github.com/developer-stylechain/stylelint-a11y/issues/4#issuecomment-1927321522

config/stylelint-config-carbon/package.json
```diff
@@ -24,22 +25,21 @@
...
     "stylelint-config-idiomatic-order": "^10.0.0",
     "stylelint-config-prettier": "^9.0.3",
-    "stylelint-config-standard": "^34.0.0",
-    "stylelint-config-standard-scss": "^11.0.0",
-    "stylelint-no-unsupported-browser-features": "^7.0.0",
+    "stylelint-config-standard": "^36.0.0",
+    "stylelint-config-standard-scss": "^13.1.0",
+    "stylelint-no-unsupported-browser-features": "^8.0.1",
     "stylelint-order": "^6.0.0",
-    "stylelint-prettier": "^2.0.0",
```
tay1orjones, line 37: Is this going to be worked on in a separate PR? Or do we have an issue for it?

config/stylelint-config-carbon/rules/stylistic-issues.js
```diff
@@ -225,70 +65,298 @@ module.exports = {
...
+  //   // Specify short or long notation for hex colors (Autofixable).
+  //   'color-hex-length': 'long',
+
+  //   // Font family
+  //   // Specify whether or not quotation marks should be used around font family
+  //   // names.
+  //   'font-family-name-quotes': 'always-where-recommended',
+
+  //   // Font weight
+  //   // Require numeric or named (where possible) font-weight values.
```
tay1orjones: Could you put all these commented-out ones back into their original line placement so the diff is more clear and the blame history is more intact?

tay1orjones: Actually I'm realizing you mentioned in the PR body

> Removed Deprecated Rule Sets
> Removed Unnecessary Disabled Rules

Are these commented out rules deprecated or unnecessary/disabled? Should we just remove them instead? That would get the diff more clean.

Also curious, how do we know that these rules are deprecated and no longer needed? Were there like a ton of console warnings or something? I just want to be sure we're still covering all these stylistic pieces.

package.json
```diff
@@ -75,7 +75,8 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "rimraf": "^5.0.0",
-    "stylelint": "^15.0.0"
+    "stylelint": "^16.0.0",
+    "stylelint-scss": "^6.0.0"
```
tay1orjones: Is this needed if it's already installed in `styleling-config-carbon`'s package.json?

package.json
tay1orjones: As posted elsewhere, we should remove this I think

packages/grid/scss/_css-grid.scss
```diff
@@ -75,6 +76,7 @@
         }
 
         @include breakpoint($key) {
+          /* stylelint-disable-next-line max-nesting-depth */
```
tay1orjones: Shouldn't this be covered by the exceptions for `@` rules?

tay1orjones (changes requested):
Hey this looks really solid! I'm very out of the loop on this one so sorry if some of the feedback here isn't valid. Let me know what you think, I'm happy to pair up on stuff if you'd like.

packages/web-components/package.json
2nikhiltom:
```
 "rules": {
      "max-nesting-depth": 3
    },
```
This isn't required- `**/packages/web-components/scss/**/*.scss` at [.stylelintignore](https://github.com/carbon-design-system/carbon/pull/17770/files#diff-36ae73c99ebd1b4072a3164e014df72eb094703df4548dd51938aa7faaa7f8d2) covers for web-component package

config/stylelint-config-carbon/rules/stylistic-issues.js
2nikhiltom: I removed the usage of [ deprecated-stylistic-rules ](https://stylelint.io/migration-guide/to-15#deprecated-stylistic-rules) to avoid below errors 

image

package.json
2nikhiltom: Absolutely correct this isn't required- `**/packages/web-components/scss/**/*.scss` at [.stylelintignore](https://github.com/carbon-design-system/carbon/pull/17770/files#diff-36ae73c99ebd1b4072a3164e014df72eb094703df4548dd51938aa7faaa7f8d2) covers for web-component package

packages/grid/scss/_css-grid.scss
2nikhiltom: added this because here the nesting-depth exceeded 3 ->  but as mentioned this isn't required anymore

2nikhiltom:
Hey @tay1orjones ! Thank you for the detailed review and for pairing up on this, made the necessary changes to be reviewed

config/stylelint-config-carbon/index.js
2nikhiltom: Majorly because of incompatibility between Stylelint 16 and `prettier: "^2.8.8"` and since Lee's PR also removed the ./plugins/prettier file. we planed bumping up prettier separately, it might make sense to deliver both updates in the same PR - Thoughts ? 

I tried moving to "prettier": "^3.2.5" - [draft_PR](https://github.com/carbon-design-system/carbon/pull/17799) - looks like some jest config needs to be addressed

tay1orjones (approved):

kennylam (approved):
Nice!

PR #17787, opened 2024-10-17
Update Menu.mdx

tomlyn:
I added the `open` attribute to each of the `<Menu>` objects because the sample code will not display anything without that. 

I also added a closing slash to these entries in one of the sample
```
<MenuItem label="Image">
<MenuItem label="Movie">
 <MenuItem label="Text">
``` 
because they are syntactically incorrect without the closing /.

Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


ariellalgilmore:
thanks @tomlyn ! can you sign the DCO: https://github.com/carbon-design-system/carbon/pull/17787#issuecomment-2420334843

tomlyn:

ariellalgilmore (approved):
LGTM! thank you

kennylam (dismissed):

kennylam:
Hm, this change sort of makes it seem like `open` is the default state, even though it isn't. Since the [component API](https://react.carbondesignsystem.com/?path=/docs/components-menu--overview#component-api) lists all available props, I wonder if `open` is actually needed in the examples? @tay1orjones what do you think?

tay1orjones (approved):
I agree with Kenny, though I think I'm actually cool with this change. Reason being is that these snippets are not the actual source of the stories and they're intended to be copied and pasted. You could argue it'd be even better if `open` was from a `useState` so it's totally clear the intent is for it to be managed by the consumer.

tay1orjones:
Also wow this is from October 🤯 so sorry this slipped through my notifications!

PR #17806, opened 2024-10-18
Move Feature Flag out of Experimental

### LINKED ISSUE #17762: FeatureFlags: Refactor story and doc locations
### Acceptance Criteria

- [ ] Refactor the location of feature flag specific stories for each component (move out of experimental, into component folders)
- [ ] Dedupe/remove component stories currently in the feature flag section that do not apply to demonstrating the feature flag (structuredlist feature flag section does not need another "default" because it's no different than the non-feature-flag version)
- [ ] Rename feature flag "Overview" mdx pages to something like "Flag details"
- [ ] Link from the main feature flags Overview page to each "Flag details" page

guidari:
Closes #17762

In this PR it was moved all Feature Flag stories out of experimental into their respective components folder under a folder called Feature Flag.

In a few cases you will notice that was deleted a few stories if the exampless could be seen only in one stories.

#### Changelog

**Changed**

- Moved all Feature Flag stories to their respective component's folder.

#### Testing / Reviewing

- Make sure the stories deleted make sense by comparing with the current storybook


alisonjoseph (approved):
These all LGTM, would we be able to move the "Flag details" page to be at the top the same way overview is? Or should we just name that page overview also to match? 

Related, do we have an issue to update the website live demos once this is live? I think Tile,Tag and StrutcturedList will need to get updated with the new urls.

annawen1:
> These all LGTM, would we be able to move the "Flag details" page to be at the top the same way overview is? Or should we just name that page overview also to match?

I agree, if we can have the "Flag details" page above the feature flag stories that would be ideal!

annawen1 (approved):

PR #17814, opened 2024-10-20
fix: remove loading animation in inlineloading inactive state

### LINKED ISSUE #17793: [Bug]: InlineLoading shows animation on inactive state
### Package

@carbon/react

### Browser

Chrome

### Package version

1.68.0

### React version

16.14.0

### Description

The component `<InlineLoading/>` has a short spinning animation when the status is inactive. I wonder why this is the case.
In my case (testing a connection) it should not show anything while the connection has not yet been tested. For this state i would naturally use the inactive state. But that would then bring up an irritating animation, where i would simply expect to see nothing.  This corresponds with what the  the [specification](https://carbondesignsystem.com/components/inline-loading/usage/) says: “Inactive has no visual indicator”. But maybe i am not fully understanding the inactive state.
The unwanted animation can be seen on the [Story board](https://react.carbondesignsystem.com/?path=/story/components-inlineloading--playground) as well, e.g. when you switch the status from 'finished' to 'inactive'.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-inlineloading--playground

### Steps to reproduce

Look at the [Story board](https://react.carbondesignsystem.com/?path=/story/components-inlineloading--playground) and switch the status from 'finished' to 'inactive'. That will show you the animation.

### Suggested Severity

### Application/PAL

DataGate

### Code of Conduct

dedanade:
Closes #17793

This is to Fix the bug of loading animation visible beside the title when `inlineloading` is in an `inactive` status. 
#### Changelog

**Changed**

- I separated the `if` condition of an active and inactive state.
- I added a new `classname` to the title of an inactive state following other classname guidelines.
- I changed the test cases to `expect` new `classname` to be in the Document.

**Removed**

- Nothing

#### Testing / Reviewing

From the [Story board](https://react.carbondesignsystem.com/?path=/story/components-inlineloading--playground), switch the status from 'finished' to 'inactive'. The loading animation will not show and we have a static text.


dedanade:

emyarod (commented):
could you remove the `.history` files from the PR? I believe that is generated by VS Code right? it may also be worth adding `.history` to the `.gitignore` as well

dedanade:
> could you remove the `.history` files from the PR? I believe that is generated by VS Code right? it may also be worth adding `.history` to the `.gitignore` as well

Hi @emyarod, Thanks for the heads up. This has been fixed.

emyarod (approved):
looks good to me!

alisonjoseph:
Want to double check with @carbon-design-system/design that this is intended for the inactive state and that the animate out wasn't done on purpose. I also want to check on if we should keep the space for the loading icon we had before.
Before PR
Screenshot 2024-10-30 at 12 25 17 PM


After changes
Screenshot 2024-10-30 at 12 24 32 PM

.gitignore
```diff
@@ -61,3 +61,6 @@ blob-report
 
 # Local token
 .env
+
+#Virtual Studio Code
```
alisonjoseph:
```suggestion
#VS Code
```

thyhmdo:
> Want to double check with @carbon-design-system/design that this is intended for the inactive state and that the animate out wasn't done on purpose. I also want to check on if we should keep the space for the loading icon we had before. Before PR Screenshot 2024-10-30 at 12 25 17 PM
> 
> After changes Screenshot 2024-10-30 at 12 24 32 PM

I think the inactive state should not be available at all. It will take some research to validate, but for the quick fix, I'd say that we can lose the space for the `inactive state`

alina-jacob (approved):
Based on the above discussion, removing the loading animation and the space for in-active state makes sense, so good to go! 🚀

PR #17815, opened 2024-10-21
feat:  callout implementation

### LINKED ISSUE #17691: [Callout] Implement the Callout update
There's an update for Callout, refer to [this issue](https://github.com/carbon-design-system/carbon/issues/17083) for visual spec.

1. Callout is not available in success & warning states
   * ask design what the migration path is here so the change is backwards compatible. success -> info and warning -> error?
2. Callout does not have action buttons, only links 
   * this is a storybook-only change, we will still allow `children` and `node` props to be fully configurable

Gururajj77:
Closes #17691 

implemented the callout update.

#### Changelog

**Changed**

- used the helped function `deprecateValuesWithin` to deprecate the values `success` and `warning` in prop types
- created a new `kindProps` type for `kind` which is an union of both deprecated values and newly accepted values
- changed the default value for `kind` from `error` to `info` as `error` is not a supported prop value for callout

#### Testing / Reviewing

check the callout component in unstable section.


alisonjoseph:
@Gururajj77 do you want to add a couple tests for this functionality?

guidari (approved):
LGTM! Thanks for testing that!

guidari (approved):
LGTM! 🚀

alisonjoseph (changes requested):
Couple things to make this a little clearer.

- Can you clarify in the PR description what changed, I believe we talked about it no longer redirecting to the new prop, but just having the deprecation warning.
- Update the playground story so the `kind` prop displays `info` and `warning`.
- Update the Overview Component API documentation, so correctly reflects the `kind` prop options.

packages/react/src/components/Notification/Notification.tsx
```diff
@@ -1348,14 +1368,18 @@ Callout.propTypes = {
...
+  kind: deprecateValuesWithin(
+    PropTypes.oneOf([
+      'error',
+      'info',
+      'info-square',
+      'success',
+      'warning',
+      'warning-alt',
+    ]),
+    ['success', 'info', 'warning', 'error'],
```
alisonjoseph: Should these just be `info` and `warning`?

alisonjoseph:
@Gururajj77 its still showing error as the default kind

Gururajj77:
> @Gururajj77 its still showing error as the default kind

Ill fix it

alisonjoseph:
> to deprecate the values success and warning in prop types
I think the deprecated props should be `error` and `success` based on design spec/docs

alisonjoseph:
@Gururajj77 its still showing error state in storybook

Screenshot 2024-11-07 at 9 18 24 AM


Screenshot 2024-11-07 at 9 18 36 AM

alisonjoseph (approved):

PR #17834, opened 2024-10-22
fix: update align prop value across various components

### LINKED ISSUE #17585: [Bug]: CodeSnippet suggests align value that isn't available yet
### The problem

I currently use the component `<CodeSnippet>` with `align` property "bottom-right".

However, during test runs, `<CodeSnippet>` logs a console warning to use `align` property "bottom-end" instead. However, `<CodeSnippet>` does not currently support this align property.

Unnecessary/unaddressable warnings make it harder to see fixable errors and slow down CI workflows.

### The solution

Either:
- Support the "bottom-end" property in `<CodeSnippet>`
- Do not log the warning until `<CodeSnippet>` supports "bottom-end"

### Examples

Code sample:
```
<CodeSnippet
        wrapText
        align="bottom-right"
        type="multi"
        onClick={() => {
          copyDetail(value);
        }}
      >
        {value}
      </CodeSnippet>
```

Logs from test running this code:
```
console.warn
    Warning: "bottom-right" is a deprecated value for the "align" prop on the "Popover" component. Use "bottom-end" instead. "bottom-right" will be removed in the next major release.

```
      

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

Shankar-CodeJunkie:
Closes #17585

Support the newly added properties like `bottom-end`, `bottom-start`, `left-end`, `left-start` in <CodeSnippet>, <Copy> and <CopyButton> components. 

#### Changelog

**Changed**

```
top
bottom
left
right
top-start
top-end
bottom-start
bottom-end
left-end
left-start
right-end
right-start
```


tay1orjones (changes requested):
Thanks for taking a look at this! The old values must remain for backwards compatibility. I think you could mirror the approach taken in popover https://github.com/carbon-design-system/carbon/blob/4b4fe0209f6868a6f1903c8f5ad199234175a47b/packages/react/src/components/Popover/index.tsx#L50-L79

The proptypes use a helper to distinguish them as deprecated 
https://github.com/carbon-design-system/carbon/blob/4b4fe0209f6868a6f1903c8f5ad199234175a47b/packages/react/src/components/Popover/index.tsx#L457-L465

If `align` is actually used anywhere other than passed through to the child component you can use the prop mapper function to help handle the new values. This might not apply to these components though, I'm not sure. https://github.com/carbon-design-system/carbon/blob/4b4fe0209f6868a6f1903c8f5ad199234175a47b/packages/react/src/components/Popover/index.tsx#L81-L93

Shankar-CodeJunkie:
Hi @tay1orjones :
I have pushed the changes to the PR. I have refactored the files by taking similar changes done on  popover component.

emyarod (approved):
looks good to me!

tay1orjones (approved):
Thanks, sorry for the delay!

PR #17837, opened 2024-10-22
chore: update react package telemetry config

mattrosno:
A while back, the `FeatureFlags` component was updated from a single prop of type object, to one prop per feature, so this data can be captured by IBM Telemetry. These new props need added to the `telemetry.yml`'s `allowedAttributeNames`.

I ran this command in the `/packages/react` directory:

```bash
npx -y @ibm/telemetry-js-config-generator generate --id fd89d12b-6a39-48b4-adf4-30c94dc0dddd --endpoint https://www-api.ibm.com/ibm-telemetry/v1/metrics --files ./src/components/**/*.(tsx|js|jsx)
```

The Telemetry team maintains `@ibm/telemetry-js-config-generator`, although it's up to you how you use it. E.g. manually if you add a net-new prop or prop value, or pro mode would be automatically somehow.

#### Changelog

**New**

- N/A

**Changed**

- `telemetry.yml` re-generated

**Removed**

- N/A

#### Testing / Reviewing

I can't figure out why the new YML files doesn't have quotes anymore, making the PR more difficult to diff. This view seems best for this PR:

image

You should see new props included if no components used them the last time this file was generated. For example, `enableV12Overflowmenu`.


tay1orjones (approved):
Thanks for this! Instead of running it automatically I added a commit that updates ci to check if the file needs updated. It's basically the same idea as the `format` check where it fails if changes are needed - you have to make the change locally and push up the changes to get the status check to pass.

Let me know what you think - I can revert that commit if there's some gaps I missed.

mattrosno:
@tay1orjones I like it. I didn't make any changes to the `telemetry.yml` after the config generator did its thing, so post merge, the check should continue working.

mattrosno:
@tay1orjones I guess this could be used in the 3 other React-based packages where each script command uses its own project ID: https://github.com/search?q=repo%3Acarbon-design-system%2Fcarbon%20allowedAttributeNames&type=code

But if these other packages rarely add/remove component props, it might not be worth the effort.

alisonjoseph (approved):

ariellalgilmore (approved):

tay1orjones:
@mattrosno great point - I'll push an update and fix the new check too, it needs a `yarn install`

tay1orjones:
I couldn't figure out why the CI check was reporting a difference in the config - whenever I would run it locally it was fine. I tried a bunch of stuff but gave up. Left in the script though so it's easier to update this!

PR #17840, opened 2024-10-23
fix: popover is clipping in modal

### LINKED ISSUE #17749: [Bug]: Popover is clipping in a modal
### Package

@carbon/react

### Browser

Chrome

### Package version

1.68.0

### React version

18.3.1

### Description

Opening a popover inside a modal results in the content being clipped. The alignment setting is also not being respected most likely due to the underlying issue causing the clipping.

Image

Using a toggle tip, I do not see the same issue.

Image

**Hacky work-around:**
On the element with the `cds-popover` class if you add a css style` position: fixed` it forces the popover to break out of the clip.
However, it doesn't truly fix the underlying issue and popover positioning based on the parent does not work (ie. align: left).

### Reproduction/example

https://stackblitz.com/edit/github-ukeceq?file=src%2FApp.jsx

### Steps to reproduce

1. Open the reproduction example
2. Click the open modal button
3. In the modal, click the "Popover" button.

Result: the popover which appears is clipped in the modal - notice the red border which encloses the "I am some content" popover is clipped
Expected: the popover is not clipped.

Notice clicking the toggle tip text shows the same popover as expected (ie. not clipped).

### Suggested Severity

### Application/PAL

Automation Builder Workflow Designer

### Code of Conduct

preetibansalui:
Closes #17749 

popover was clipping in modal

#### Changelog

**Changed**

The issue was caused by the drop-shadow, as it creates a stacking effect. So I applied the drop-shadow to its direct child .#`{$prefix}--popover-content `instead of `.#{$prefix}--popover`

#### Testing / Reviewing

1. Go to Storybook > Modal > Default
2. Open popover by clicking checkbox
3. Scroll within the modal. the popover should not be clipped


alisonjoseph (approved):
LGTM

kennylam (approved):

PR #17842, opened 2024-10-23
17492: Increase SideNavMenu unit test coverage 

### LINKED ISSUE #17492: Increase `SideNavMenu` unit test coverage

mikdhadc:
Added testcases for SideNavMenu Component that increased the test coverage

BREAKING CHANGE: The SideNavMenu now has a test coverage greater than 90%

Closes #17492 

#### Changelog

**Changed**

- Added tests case to properly test and increase the test coverage for SideNavMenu

#### Testing / Reviewing
Screenshot 2024-10-24 at 7 10 14 PM


packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
```diff
@@ -98,4 +102,118 @@ describe('SideNavMenu', () => {
...
+        <SideNavMenuItem isActive={false} aria-current="page">
+          a
+        </SideNavMenuItem>
+      </SideNavMenuItem>,
+      <SideNavMenuItem isActive={true} aria-current="page">
+        b
+      </SideNavMenuItem>,
+    ];
+    expect(hasActiveDescendant(child)).toBe(true);
+  });
```
tay1orjones: Rather than test the internal implementation like this - how could we reframe the tests from the perspective of someone using the component?

hasActiveDescendant is only called in one place under certain circumstances
https://github.com/carbon-design-system/carbon/blob/061970723f64d1376ea28109328c1aec1f192ecc/packages/react/src/components/UIShell/SideNavMenu.tsx#L91-L92

I think this would mean that to test this from a user-interaction viewpoint, each test could `render` the component with a different type of children which would get `hasActiveDescendant` to be called, and then the assertion would be `toHaveClass('${prefix}--side-nav__item--active')`.

packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
```diff
@@ -98,4 +102,118 @@ describe('SideNavMenu', () => {
...
+
+  it('should return true if the child is a invalid React element and has isActive props set to true', () => {
+    const child = <div isActive={true} />;
+    expect(hasActiveDescendant(child)).toBe(true);
+  });
```
tay1orjones: Same kind of idea here. I don't think we need to directly assert that the function call returns true. Instead it's the end result of the function call we want to ensure happens correctly.


Again I think in nearly all cases this would be an assertion of `toHaveClass` or `.not.toHaveClass`. The classes are admittedly still an implementation detail, but they're the closest thing to what the user perceives without having to actually assert that certain styles are applied (background, etc. or whatever is in selectors with `${prefix}--side-nav__item--active`

packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
```diff
@@ -98,4 +102,118 @@ describe('SideNavMenu', () => {
...
+    const child = ['abc', 'xyz'];
+    render(
+      <SideNavMenu title="test-title">
+        <SideNavMenuItem>a</SideNavMenuItem>
+        <SideNavMenuItem>b</SideNavMenuItem>
+        <SideNavMenuItem>c</SideNavMenuItem>
+      </SideNavMenu>
+    );
+    expect(hasActiveDescendant(child)).toBe(false);
+  });
```
tay1orjones: This one doesn't use anything that's being rendered.

packages/react/src/components/UIShell/SideNavMenu.tsx
```diff
@@ -200,7 +200,9 @@ SideNavMenu.propTypes = {
 Defining the children parameter with the type ReactNode | ReactNode[]. This allows for various possibilities:
 a single element, an array of elements, or null or undefined.
 **/
-function hasActiveDescendant(children: ReactNode | ReactNode[]): boolean {
+export function hasActiveDescendant(
```
tay1orjones: If the tests are refactored to be [user centered](https://testing-library.com/docs/#:~:text=What%20you%20should,Child%20components) this won't need to be exported.

packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
```diff
@@ -98,4 +102,118 @@ describe('SideNavMenu', () => {
...
+          ref={mockRef}
+          title="test-title">
+          <SideNavMenuItem>a</SideNavMenuItem>
+          <SideNavMenuItem>b</SideNavMenuItem>
+          <SideNavMenuItem>c</SideNavMenuItem>
+        </SideNavMenu>
+      </SideNavContext.Provider>
+    );
+
+    fireEvent.keyDown(screen.getByText(/a/i), {
```
tay1orjones: Could/should this use `userEvent` instead?

tay1orjones (changes requested):
Thanks for taking this on! I have a couple points of feedback below, let me know what you think.

mikdhadc:
Thanks for your feedback. Will work on them.

mikdhadc:
@tay1orjones , I have updated the PR based on your feedback. Let me know if any other modifications are needed. Thank you.

packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
```diff
@@ -98,4 +101,116 @@ describe('SideNavMenu', () => {
...
+    );
+  });
+
+  it('sets isExpanded and prevExpanded when sideNav is expanded, prevExpanded is true and isRail is true', () => {
+    render(
+      <SideNavContext.Provider value={{ isRail: true }}>
+        <SideNavMenu
+          isSideNavExpanded={true}
+          defaultExpanded={true}
+          ref={mockRef} //uses the value of mockRef from previous test
```
tay1orjones: Could you share more on why this is needed? Generally tests should be isolated from one another and if you need to mock a ref you can do it with jest.fn() like in the `should support a `ref` on the <button> element` test.

mikdhadc: Thanks. It was unnecessary there. I have removed them.

packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
```diff
@@ -98,4 +99,111 @@ describe('SideNavMenu', () => {
...
+          <SideNavMenuItem>b</SideNavMenuItem>
+          <SideNavMenuItem>c</SideNavMenuItem>
+        </SideNavMenu>
+      </SideNavContext.Provider>
+    );
+    expect(screen.getByRole('button')).toHaveAttribute(
+      'aria-expanded',
+      'false'
+    );
+  });
```
tay1orjones: I won't hold up the merging of this as the tests pass and do cover the code properly, but the naming and structure of these was really confusing to me. I had to open up the code coverage report to discern what you're aiming to cover. I found this:

2024-10-29 at 13 36 55-MAIN-Google Chrome

Refactoring the structure to be something like below would help

```js

describe('properly detects active descendants and applies active styling', () => {
  it('when a child in children array has isActive', () => { ... });
  it('when a child in children array has aria-current', () => { ... });
  it('when a child in nested children array has isActive', () => { ... });
  it('when a child in nested children array has aria-current', () => { ... });
  it('when a single child has isActive', () => { ... });
  it('when a single child has aria-current', () => { ... });
});

describe('properly detects non-active descendants and does not apply active styling', () => {
  it('when a child in children array is not a valid element', () => { ... });
  it('when all children have no props given', () => { ... });
  it('when a single child has no props', () => { ... });
});

```

Each of these would need a render block with `SideNavMenu defaultExpanded={true}`. Each assertion would still be `toHaveClass` of the active class, even when given `aria-current`. For the "children array" ones, pass in multiple SideNavMenuItem components rather than a dummy array.

tay1orjones (approved):

packages/react/src/components/UIShell/__tests__/SideNavMenu-test.js
mikdhadc: Thanks for the feedback. I have updated the code. Please let me know if anything else is to be done. thank you
@tay1orjones @emyarod

emyarod (approved):
looks good to me!

PR #17858, opened 2024-10-23
fix(combobox): controlled combobox onchange

### LINKED ISSUE #17847: [Bug]: onChange fires twice when selecting item on ComboBox
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.68.0

### React version

_No response_

### Description

When selecting an item on a combobox, the onChange function is called twice.

### Reproduction/example

https://stackblitz.com/edit/github-yb3are?file=src%2FApp.jsx%3AL34-L34,package-lock.json

### Steps to reproduce

Run the slackblitz sample with devtools open so you can see the console logs.
Select an item on the ComboBox
Problem: The onChange function is called twice. You can see the debug log printed 2 times to the console

### Suggested Severity

None

### Application/PAL

IBM Cloud - IAM UI

### Code of Conduct

neuescodes:
Closes https://github.com/carbon-design-system/carbon/issues/17847

In cases where a selection is made to a fully controlled combobox component, and the selectedItem prop updates on this selection, such as in cases where this state is managed externally (probably just about all cases where `selectedItem` is being used), the onChange was firing twice. This PR adds a check to see if the selectedItem prop was changed externally before calling the onChange in the useEffect pertaining to controlled comboboxes. It also ensures that onChange is not fired for controlled comboboxes when a selection is made that is the same as the current selection.

* updated useEffect hook for controlled combobox so onChange doesn't fire twice when selection is made and selectedItem prop changes
* updated onSelectedItemChange to not fire onChange if selectedItem is unchanged
* unit tests

#### Testing / Reviewing

* add log to [onChange in fullyControlled combobox in ComboBox.stories.js](https://github.com/Neues/carbon/blob/ef9f8b0418431d885f464effeff4ab98bbea78e8/packages/react/src/components/ComboBox/ComboBox.stories.js#L257)
* select different options to see onChange fires only once
* select the same option to see that onChange does not fire


packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -746,7 +750,10 @@ const ComboBox = forwardRef(
         setHighlightedIndex(indexToHighlight(normalizedInput));
       },
       onSelectedItemChange({ selectedItem }) {
-        onChange({ selectedItem });
+        // only call onChange if new selection is updated from previous
+        if (JSON.stringify(selectedItem) !== JSON.stringify(selectedItemProp)) {
```
tay1orjones: FilterableMultiSelect and MultiSelect use `isEqual` from `react-fast-compare` for this. It's faster and handles a bunch of edge cases that would break or be wrong with JSON.stringify.

tay1orjones (changes requested):
This looks awesome, thanks for the super fast turnaround! Tests look great too. Just one minor piece of feedback from me

packages/react/src/components/ComboBox/ComboBox.tsx
neuescodes: Nice, good info, updated!

guidari (approved):
Awesome! Thanks for having this fix up! 🚀

alisonjoseph (approved):

tay1orjones (approved):

PR #17868, opened 2024-10-24
feat(DataTable): add double toolbar prop

### LINKED ISSUE #17844: [Feature Request]: Add batch actions bar below the Data table tool bar
### The problem

In the current DataTables, when selecting rows and generating the Batch Actions blue bar is rendered on top of the Action toolbar. With that, the user can't do actions like update a search query.

Without the Batch Actions bar (search bar and actions button shown):
Image

With the Batch Actions bar (search bar and actions button are hidden):
Image

### The solution

Have a new prop that allows the Batch Actions bar to be displayed bellow the Actions Toolbar. 

Example: 
Image

### Examples

_No response_

### Application/PAL

Carbon for IBM Products

### Business priority

### Available extra resources

_No response_

### Code of Conduct

pilarfernandezz:
Closes #17844

This PR allows both Batch Actions and Actions toolbars to be displayed simultaneously.
Right now the Batch Actions toolbar overrides the Actions toolbar when there is some row selected. With that, the search bar and actions buttons can not be changed/clicked.

image

#### Changelog

**New**

- Added ```useDoubleToolbar``` property to Table and TableToolbar.  

**Changed**

- The default value for the prop is ``` false```, maintaining the current behavior. 
- If the prop is ```true``` the Batch Actions toolbar will be displayed bellow the Actions toolbar.

#### Testing / Reviewing

Change the prop to true and select any row in the DataTable. The blue Batch Actions toolbar should be displayed bellow the Actions toolbar. Should be possible to click in the Actions buttons and edit the search bar input.
image

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1670,6 +1670,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "pilarfernandezz",
+      "name": "Pilar Fernandez",
+      "avatar_url": "https://avatars.githubusercontent.com/u/38540218?v=4",
+      "profile": "https://github.com/pilarfernandezz",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "none"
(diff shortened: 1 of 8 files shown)


pilarfernandezz:

pilarfernandezz:
recheck

tay1orjones:
Before we can review this we need a finalized spec from design, https://github.com/carbon-design-system/carbon/issues/17844#issuecomment-2441750479

tay1orjones (changes requested):
Before we can review this we need a finalized spec from design, https://github.com/carbon-design-system/carbon/issues/17844#issuecomment-2441750479

Issue #17499, opened 2024-09-18
Increase `OverflowMenu` unit test coverage


tay1orjones:

divya-281:
Hi @alisonjoseph  @tay1orjones I'd like to work on this issue. Can you assign this to me?

PR #15375, opened 2023-12-12
`OverflowMenu` fix when oppening two menus

guidari:
Closes #12942 

Fixing the css selection by using the `first-child`

#### Testing / Reviewing

- Go to the Default stories of the `OverflowMenu` and try to open both menus
- [x] Delete duplicated `OverflowMenu`;
- [x] Tests should pass after deleting the duplicate `OverflowMenu`

Issue #12942, opened 2023-01-10
[Bug]: OverflowMenuItem doesn't take focus if disabled


pre-ska:
### Package

carbon-components

### Package version

10.52.0

### React version

17.0.2

### Description

if Overflow menu has disabled items, it doesn't close previously opened menu

### Suggested Severity

Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.

### Reproduction/example

https://codesandbox.io/s/naughty-thunder-7p6n40?file=/index.js

### Steps to reproduce

there are two overflow menus  on codesanbox, top and bottom
the bottom one has menu items disabled

1) click top menu icon - it shows menu
(don't close top menu by clicking somewhere on the page)
2) click bottom menu icon- it shows bottom menu but top menu didn't close

if it's other way around, first bottom then top, top menu will take focus and close bottom one

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tw15egan:
Thanks for reporting, @pre-ska! You're right; The issue is that there are no focusable items when the second menu is opened, so the previous menu does not close properly.  It seems to work when at least one item is not disabled.

PR #17875, opened 2024-10-24
[unit-test]: overflow menu

divya-281:
increases test coverage for overflow menu
#17499


guidari:
Hey @divya-281, thanks for opening that PR!! 🚀
Do you think we could raise the coverage to over 80%? Currently it is 67,41%. But if there is no way to tests those parts, don't worry.

https://app.codecov.io/gh/carbon-design-system/carbon/pull/17875/blob/packages/react/src/components/OverflowMenu/OverflowMenu.tsx

divya-281:
@guidari  I was able to increase to 74%

tay1orjones (commented):
This is looking good, thanks for taking it on! 

Looking at the coverage report, a few ideas to get this to 80%:

* Test that pressing <kbd>escape</kbd> closes the menu
* Find the path to get `this.closeMenuAndFocus();` to be called. The steps to reproduce [this bug](https://github.com/carbon-design-system/carbon/issues/12942) could be helpful, as [the fix](https://github.com/carbon-design-system/carbon/pull/15375) was within the final condition before it's called. I think the assertion would be that the menu closed and focus returned to the trigger element.

https://github.com/carbon-design-system/carbon/blob/796f838c420d0356922322dd7f62cc395c527061/packages/react/src/components/OverflowMenu/OverflowMenu.tsx#L585-L593

divya-281:
@tay1orjones @guidari I added a test to check if the menu closes when pressing the esc key

divya-281:
@tay1orjones @guidari Thanks for the suggestions, I checked locally and the coverage was 80%.

divya-281:
@tay1orjones  @guidari Thanks for the suggestions, I was able to get the coverage upto 80%.

guidari:
Thank you @divya-281 
One question, is there any reason why we can't test the `left` `right` in th switch case? I see `top` `bottom` are being tested.
Screenshot 2024-11-12 at 09 50 32

divya-281:
@guidari  
const triggerButtonPositionProps = {
  [DIRECTION_TOP]: 'bottom',
  [DIRECTION_BOTTOM]: 'top',
};
 I only see bottom and top here, Can you let me know when the case left is triggered? Am I missing something?

guidari:
Hey @divya-281 Interesting catch! I was taking a look in the code and seems like the `left` and `right` are never called.
The different types of direction are only called in the FeatureFlag OverflowMenu with `menuAlignment`
I guess we can comment or get rid of this line of code.

divya-281:
@guidari  I've commented those lines

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -81,15 +82,16 @@ export const getMenuOffset = (menuBody, direction, trigger, flip) => {
...
+    // case left and right is not called
+    // case 'left':
+    // case 'right': {
+    //   // TODO: Ensure `trigger` is there for `<OverflowMenu open>`
+    //   const triggerHeight = !trigger ? 0 : trigger.offsetHeight;
+    //   return {
+    //     left: 0,
+    //     top: (!flip ? 1 : -1) * (menuHeight / 2 - triggerHeight / 2),
+    //   };
+    // }
```
tay1orjones:
```suggestion
```

tay1orjones (approved):
LGTM, thanks!

guidari (approved):
Thank you @divya-281

PR #17878, opened 2024-10-25
feat(utilities/date-and-time/relative): add localization for "just now"

janhassel:
Localizes the "just now" variant for relative time formats occurring when an event has happened <=59s ago. The reference (en-US) wis changed to "now" as this allows us to utilize the Internationalization API for translating into other locales.

#### Changelog

**Changed**

- If an event has happened <=59s ago, use a relative formatter set up with `numeric: 'auto'` and pretend the event happened 0 seconds ago to return the localized "now" string.
  - Added test for this case

**Removed**

- Unnecessary npm scripts in `packages/utilities` and `packages/utilities-react`


ariellalgilmore (approved):
LGTM!

tay1orjones (approved):

PR #17889, opened 2024-10-28
Fix: Reset date range calendar when value is empty string or array

### LINKED ISSUE #17702: [Bug]: Resetting Datepicker range with calendar using value property doesn't work when assigning either an array of two emtpy strings ["", ""] or an empty string ""
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox

### Package version

v1.67.0

### React version

v18.2.0

### Description

Resetting Datepicker range with calendar using value property doesn't work when assigning either an array of two emtpy strings ["", ""] or an empty string ""

Below are our observations on the issue to summarize,

1. If we use an empty array [], the end date is not getting reset
2. If we use empty string "", when manually enter dates instead of picking from calendar opens up the calendar on end date input upon resetting.
3. We tried using array of two empty strings ["", ""], this fixes the issue with manual input, however, it's causing another issue, which is end date pertaining to previously selected value immediately after selecting the start date without opening calendar.

### Reproduction/example

https://stackblitz.com/edit/github-cmfaqv?file=src%2FApp.jsx

### Steps to reproduce

1. Select date range using calendar
2. Click on "Reset range" button
3. Dates would be reset
4. Now select "Start date" again through calendar
5. You observe "End date" would automatically appear from your previous selection even after you reset both dates

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

dedanade:
Closes #17702

Fix issue when a user resets datepicker calendar using an empty string, array, or two empty strings in an array.
#### Changelog

**New**

- add condition to reset date ranges
- added tests for date range reset
- extracted locales to  `DatePickerLocales.ts` constants file

**Changed**

- various type and unused var fixes to remove `eslint-disable` (not all, some seem more involved)

**Removed**

- unused code

#### Testing / Reviewing

Create a button to reset the date range in DatePicker.stories to reset the calendar with a daterange value like this [https://stackblitz.com/edit/github-cmfaqv?file=src%2FApp.jsx.](url)

Select the start and end date, click on the reset button you created above and reselect start date again to notice the end date doesn't automatically populate from the previous selection.


dedanade:

emyarod (commented):
it looks like VS Code is adding `.history` to the repo but we probably don't need to commit those files!

dedanade:
> it looks like VS Code is adding `.history` to the repo but we probably don't need to commit those files!

Hi @emyarod, Apologies, This is coming from a different commit that has been fixed and unrelated to this task. It looks good to me now.

guidari:
Hey @dedanade The Stackblitz does not work.
Can you create an example story for us to test, please

dedanade:
Hi @guidari , to test for the issue, kindly check  https://stackblitz.com/edit/github-cmfaqv?file=src%2FApp.jsx

guidari (approved):
I pull down locally and created a storybook example to test the fix.

LGTM 🚀 Thanks for opening that.

tay1orjones (approved):
LGTM 👍

tay1orjones:
@dedanade your commits were made with username/email creds that don't match your github profile, so the DCO bot doesn't see this "user" as having signed the DCO. We can merge this manually to override the failing check, but just something to know for the future. Thanks again for this fix!

dedanade:
recheck

heloiselui:
Thanks for looking into this! I was wondering about one detail: in addition to handling `"" and ["", ""]`, should we also cover cases `like [] or null`? Based on the issue description, it seems those might also cause the end date to persist unexpectedly. What do you think about extending the condition to reset the calendar in those cases as well?

riddhybansal (approved):

kennylam:
> Thanks for looking into this! I was wondering about one detail: in addition to handling `"" and ["", ""]`, should we also cover cases `like [] or null`? Based on the issue description, it seems those might also cause the end date to persist unexpectedly. What do you think about extending the condition to reset the calendar in those cases as well?

@heloiselui Yes, great catch. I've added your suggestions to the condition (it matches the pattern in the `useEffect` a few lines above it). I also added tests for this as well as a few other small changes. I've updated the PR description with the changes.

PR #17907, opened 2024-10-29
feat: exposed tabIndex to Table component

### LINKED ISSUE #17896: [Bug]: Table component not tab-able (for accessibility)
### Package

@carbon/react

### Browser

Chrome

### Package version

1.68.0

### React version

18.3.1

### Description

We need to be able to configure the `tabIndex` for the div with class `cds--data-table-content`. Can we get this exposed or configurable through sometime like the `Table` component? not sure where is best.. Not sure if you can simply move `tabIndex` from `cds--data-table-content` to the `table` element.. (i think it worked for me but i dont know all the use cases). 
Or maybe `Table` can have a new prop like `tabAlways={true|false}`, and if we set to `true`, the code will always add `tabIndex="0"` to the `cds--data-table-content` div?

It's a long story, but try to summarize.. 
We are using Carbon's `Table` to show log entries and we need to be able to Tab to it. Generally, Tabbing to focus/highlight the whole table is not available unless the table content's width is fairly wide and the horizontal scrolling appears. So basically, when horizontal scrolling does appear, the div with class `cds--data-table-content` gets `tabIndex="0"` added to it, and this allows us to tab to it. 
But if content isn't wide enough (no horizontal scrollbar), we're not able to tab to it... but we still need to tab to it because the table height is big (lots of rows) and we want to vertically scroll it. 

So imagine a table with lots of rows.. some times the message are long and some times the message are short (not predictable). We want to be able to scroll the table always (which we can right now with a mouse/trackpad, but currently for accessibility, we cannot tab to it and cannot scroll with keyboard). The restriction/limitation/condition currently really depends on the browser's screen width, the content of the table, and if the horizontal scrollbar appears or not.. But we want to be able to tab to it always.. so we need Carbon's Table to somehow allow us to configure tabbing to the entire table.

There is a conversation here that provides more details:
https://ibm-analytics.slack.com/archives/C2K6RFJ1G/p1730127982591899

### Reproduction/example

N/A

### Steps to reproduce

Not tab-able table:
Any carbon `Table` component where screen width is wide and the table content is short.. not enough to create horizontal scrollbar.

Tab-able table:
Any Carbon `Table` when table content isnt wide enough to fit the page and causes table to have horizontal scrollbar.

We want carbon Table to be configurable so that the currently not tab-able scenario can be tab-able.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

guidari:
Closes #17896 

Exposing `tabIndex` for `Table` component.

#### Testing / Reviewing

- Table component should work as expected (you can test the Datatable With AI Label)


packages/react/src/components/DataTable/Table.tsx
```diff
@@ -226,7 +232,9 @@ export const Table = ({
     <div
       className={`${prefix}--data-table-content`}
       // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-      tabIndex={isScrollable ? 0 : undefined}>
+      tabIndex={
+        tabIndex === undefined ? (isScrollable ? 0 : undefined) : tabIndex
```
alisonjoseph:
```suggestion
      tabIndex={tabIndex ?? (isScrollable ? 0 : undefined)}
```

this feels slightly cleaner without having to nest ternarys

alisonjoseph (approved):
this LGTM, small code suggestion

riddhybansal (approved):

ariellalgilmore (commented):
looks like some tests are failing before this can get merged

PR #17908, opened 2024-10-29
fix(inline-loading): fix inactive styles

### LINKED ISSUE #17793: [Bug]: InlineLoading shows animation on inactive state

alisonjoseph:
Closes #17793

#### Changelog

**Changed**

- update inline loading component when in inactive state to no longer show loading svg

#### Testing / Reviewing

### DIFF
--- a/packages/react/src/components/InlineLoading/InlineLoading.tsx
+++ b/packages/react/src/components/InlineLoading/InlineLoading.tsx
@@ -71,7 +71,9 @@ const InlineLoading = ({
   ...rest
 }: InlineLoadingProps) => {
   const prefix = usePrefix();
-  const loadingClasses = classNames(`${prefix}--inline-loading`, className);
+  const loadingClasses = classNames(`${prefix}--inline-loading`, className, {
+    [`${prefix}--inline-loading__inactive`]: status === 'inactive',
+  });
   const getLoading = () => {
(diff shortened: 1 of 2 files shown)


alisonjoseph:
duplicate PR

Issue #13138, opened 2023-02-13
Enhance issue triage process via automation


tay1orjones:
There's a lot of pieces of our issue triage process that we should automate to reduce the load on maintainers when triaging issues.

### Ideas to explore
- [x] Ensure "needs triage" label and project field are always applied to new issues, no matter who opens it
- [ ] Ensure "carbon champions" (org members) issues are always addressed and do not fall through any cracks in triage automation
- [ ] Triage automation: apply `status: needs reproduction` label when no sandbox, stackblitz, or storybook url is provided in issue body, remove label if issue body or any comments have a sandbox, stackblitz, or storybook link
- [ ] Triage automation: Auto close issues that have `needs reproduction` label for x days (7 days ?)
- [ ] Triage automation: apply `package: *` labels based on the response of the `Package` field in the issue template
- [ ] ~Triage automation: apply `version: 10`/`version: 11` label based on the response of the `Version` field in the issue template~
- [ ] Triage automation: apply `browser: *` labels based on the response of the `Browser` field in the issue template
- [ ] Triage automation: apply `component: *` labels based on any references to components in the issue or comment body
- [ ] Triage automation: apply `screen reader: *` labels based on the response of the field in the issue template
- [x] A11y issue template should also apply a `bug` label
- [ ] Triage automation: when `severity: 1` label is applied, notify in team slack channel

### Open questions
- Is there a way we can keep this automation in sync with if/when our issue templates change?

ghost:
Hey! I maintain an open-source tool that might help with this.

**GitScope** uses AI agents to automatically triage GitHub issues:
- Auto-classify issues (bug/feature/question/docs)
- Detect semantic duplicates (catches 3x more than keyword matching)
- Flag sentiment issues before they escalate

Free for public repos: https://gitscope.dev

Here's your current health badge:
[![Repository Health](https://api.gitscope.dev/api/badges/carbon-design-system/carbon.svg)](https://gitscope.dev)

Check out how simonmichael/hledger organized 8 different GitScope badges in their README: https://github.com/simonmichael/hledger

Happy to give you a walkthrough if useful! 🙌

PR #17910, opened 2024-10-29
ci(actions): close issues with no reproduction

frankharkins:
Part of #13138, specifically:

> Triage automation: Auto close issues that have needs reproduction label for x days (7 days ?)

### Implementation

There's no easy way to work out when a label was applied from the GitHub API. Instead, we just check for issues with the `status: needs reproduction` label that haven't been updated at all in the last X days.

This action would currently close the following two issues, but may apply to more if you start automatically adding the "need reproduction" label to new issues.

```
ID      TITLE                                                                          LABELS                                                              UPDATED           
#17321  [Bug]: Problems when update from carbon-components to @Carbon/react            type: bug 🐛, status: needs triage 🕵️, status: needs reproduction    about 27 days ago
#16083  [a11y]: SideNavMenuItem is not keyboard operable, unable to focus using tab    type: a11y ♿, status: needs reproduction                            about 2 months ago
```

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/stale@28ca1036281a5e5922ead5184a1bbf96e5fc984e #v9.0.0
         with:
           any-of-issue-labels:
-            "status: waiting for author's response 💬,status: needs more info 🤷‍♀️"
+            "status: waiting for author's response 💬,status: needs more info 🤷‍♀️,status: needs reproduction"
           only-pr-labels: ''
           stale-issue-message: |
             This issue has been marked as stale because it has required additional


tay1orjones (changes requested):
Thanks for taking a look at this! My only hesitation with this idea is it would probably end up closing issues that shouldn't be closed. Something like this is very likely:

1. The issue receives the label
2. They provide a repro link
3. We miss it and/or we forget to remove the label
4. Issue is closed when it shouldn't have

One way around that problem could be to parse all the issue body and comments when an issue is commented/updated/etc. to determine if a link to stackblitz or codesandbox exists. 

Another way would be to turn off the issue _closing_ and then we could manually as a team review issues with both the "needs repro" and "stale" labels to close them if needed.

The goal isn't to close issues anyway, it's to prompt a response with a reproduction. I think the second option is maybe the route to go.

Additionally
- Remove the other label triggers so it's just the repro one
- Update stale issue message
  > This issue has been marked as stale because it needs a reproduction link. When you get the chance, use a [sandbox template](https://github.com/carbon-design-system/sandboxes/tree/main?tab=readme-ov-file#carbon-design-system-sandboxes) to reproduce the problem and post the link back here.
- Remove the close issue message and the days-before-close (or whatever needs done to turn issue closing off)
- Remove the `debug-only`

tay1orjones:
Hey thanks again for this PR! I'm going to close for now based on lack of activity but I'm happy to reopen, just let me know.

PR #17920, opened 2024-10-30
fix(Tabs): handle activation mode scrolling in separate `useIsomorphicEffect`s

### LINKED ISSUE #17274: [Bug]: Tab does not scrollIntoView when selected tab is changed programmatically and activation is manual
### Package

@carbon/react

### Browser

_No response_

### Package version

^1.65.0-rc.0 (from Carbon Vite template)

### React version

18.2 (from Carbon Vite template)

### Description

When not all the tabs are visible and you programmatically change the selected tab to one that is not visible, the tab scrolls into view.  However, if activation = manual, then this functionality does not work.

### Reproduction/example

https://stackblitz.com/edit/github-xfxgew

### Steps to reproduce

Initially, the first tab is selected.
Click the button to change to the 4th tab.
The tab content changes but the 4th tab is not visible.

If you remove the activation="manual", the functionality works correctly.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

emyarod:
**NOTE: there is a test story that should be removed before merge**

---

Closes #17274

This PR updates the Tab component to scroll programmatically selected tabs into view, regardless of tab `activation` type

Prior to this PR, Tabs in manual `activation` mode would only scroll when the active tab changed. Normally this is not an issue but when the selected tab is modified programmatically, the lack of user interaction means the selection does not occur on the active tab.

By splitting the existing `useIsomorphicEffect` into two separate hooks, they can handle `manual` and `auto` activation modes separately. This ensures that the component will scroll as needed both when changing the selected tab and when changing the active tab in manual `activation mode

#### Changelog

**New**

- `scrollTabIntoView` Tab method

**Changed**

- split existing `useIsomorphicEffect` that handled scrolling into two distinct `useIsomorphicEffect`s for each `activation` mode

#### Testing / Reviewing

- visit the temporary Test Tabs story
- confirm that both activation modes will scroll the correct tabs into view as needed

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.tsx
+++ b/packages/react/src/components/Tabs/Tabs.tsx
@@ -606,13 +606,13 @@ function TabList({
 
   useIsomorphicEffect(() => {
     if (ref.current) {
-      //adding 1 in calculation for firefox support
+      // adding 1 in calculation for firefox support
       setIsScrollable(ref.current.scrollWidth > ref.current.clientWidth + 1);
     }
 
     function handler() {
       if (ref.current) {
-        //adding 1 in calculation for firefox support
+        // adding 1 in calculation for firefox support
         setIsScrollable(ref.current.scrollWidth > ref.current.clientWidth + 1);
       }
     }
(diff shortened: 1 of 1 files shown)


kennylam (approved):
LGTM!

alisonjoseph (approved):
LGTM!

PR #17944, opened 2024-11-01
fix: fixed hover on multiselect

### LINKED ISSUE #17820: [Bug]: Readonly state in Multiselect and FilterableMultiselect has hover interaction on the close button.
### Package

@carbon/react

### Browser

Chrome

### Package version

v11

### React version

v18

### Description

from the comment in this [PR](https://github.com/carbon-design-system/carbon/pull/17662#pullrequestreview-2367110367), we found out that the hover state is still present on the button when in readonly state. This is from the components `ListBox.Selection` and `ListBoxSelection` (which is in the next folder). 

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-multiselect--playground&args=readOnly:!true

### Steps to reproduce

go to the above link and hover on the 'X' button.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

guidari:
Closes #17820

Setting background-color to transparent when `Multiselect` is set to `readOnly`

#### Testing / Reviewing

- Go to playground and set `Multiselect` or `FilterableMultiselect` to `readOnly`


Gururajj77 (approved):

preetibansalui (approved):

preetibansalui:
It looks good to me but I think it's better to take a design review on this one.

laurenmrice (changes requested):
@guidari Looking good! Just  two small things:

**MultiSelect and Filterable MultiSelect**
- The border around the tag is not the correct color. The [design spec](https://github.com/carbon-design-system/carbon/issues/17820#issuecomment-2427496780) says it should be color token `$border-subtle`.

**MultiSelect**
- On the "Playground" story of MultiSelect, when the readonly prop is set to true, when hovering over the field and tag, there are still areas where the mouse cursor turns to a pointer instead of the arrow. It should always be the arrow in the readonly state.
Nov-05-2024 15-14-19

tay1orjones (approved):
LGTM once the cursor is updated 👍

guidari:
> @guidari Looking good! Just two small things:
> 
> **MultiSelect and Filterable MultiSelect**
> 
> * The border around the tag is not the correct color. The [design spec](https://github.com/carbon-design-system/carbon/issues/17820#issuecomment-2427496780) says it should be color token `$border-subtle`.

Thanks @laurenmrice ! My bad, I didn't the design spec down there in the issue 😅

laurenmrice (approved):
Looks good to me. Thanks Gui!

PR #17952, opened 2024-11-01
fix(dropdown): stabilize internal references to reduce re-renders

### LINKED ISSUE #17897: [Bug]: [Dropdown] Uncaught Error: Maximum update depth exceeded.
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.68.0

### React version

v18.2.0

### Description

We started seeing error `Uncaught Error: Maximum update depth exceeded. ` in our application when changing dropdown menu options. Following screen recording shows the error in our application -

https://github.com/user-attachments/assets/c79f6796-52e7-492e-8696-2df1e21a3ec6

```
Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

we recently added this code -
```
<Dropdown
	itemToElement={this.renderItem}
	renderSelectedItem={this.renderItem}
/>
```

I am seeing `renderItem()` is called multiple times when I hover over dropdown menu options and also when I select new dropdown options.
Wondering if the error is caused due to so many calls to the renderItem() method?

### Reproduction/example

https://stackblitz.com/edit/github-enf21q?file=src%2FApp.jsx

### Steps to reproduce

- Go to https://stackblitz.com/edit/github-enf21q?file=src%2FApp.jsx
- Open browser console
- Change dropdown options
- renderItem method is called several times
Image

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #17897

This makes Dropdown's internal references (variables, functions, props, etc) more stable to reduce re-renders by ~98% 🤯 

#### Changelog

**Changed**

- Move `stateReducer` to be delcared outside of the component
- Wrap `onSelectedItemChange` in `useCallback`
- Memoize the `readOnlyEventHandlers`, and wrap the `onKeyDownHandler` it uses with `useCallback`
- Memoize the `normalizedSlug` that clones the `slug` prop and forces the `mini` size
- -
- Prevent unnecessary state updates by updating `stateReducer` by returning `state` directly instead of a new object every time (this alone was a huge improvement)
- Memoize `items`
- Memoize `selectProps` and all the props passed in (`isItemDisabled`,`onHighlightedIndexChange`)
- Memoize `toggleButtonProps`
- Add `enableFloatingStyles` to the dependency array to ensure `menuProps` will get updated

#### Testing / Reviewing

- Ensure all stories for Dropdown work as expected, including keyboard interaction, cursor interaction, and selection

<details><summary>Click to expand test results</summary>
<p>

The benchmark I tested against is the original bug repro with the following interactions, resulting in **4674 renders**.

https://github.com/user-attachments/assets/a7a891d6-74ba-4642-a567-cfeeeac846d5

The first commit reduces this same interaction to **2145 renders**.

https://github.com/user-attachments/assets/47c427f7-01e1-46c2-be85-9d5cbe3a00b4

The second commit is even better, down to just **89 renders**.

https://github.com/user-attachments/assets/8e1398e0-af41-4008-bdaa-61939e4aee6e

The first two counts are somewhat of an incorrect count because they're tied to mousemove interaction. The update to `stateReducer` is what pulled that down so significantly.

</p>
</details>


emyarod (approved):
looks good to me!

alisonjoseph (approved):
🔥 LGTM!

PR #17964, opened 2024-11-05
fix: definition tooltip breaks with large text

### LINKED ISSUE #17848: Definition tooltip : tooltip and the caret icon is not aligned when used in larger texts
When we use Definition tooltip for larger texts and set `align` to `bottom-right | bottom-left` , the tooltip content is aligned accordingly , but the caret icon remains at the centre.

Image

preetibansalui:
Closes #17848 

Definition Tooltip breaks with large text

#### Changelog

1. Update default `align` value to `bottom` instead of `bottom-start`
2. Added autoAlign prop

#### Testing / Reviewing

1. Go to Storybook > Definition Tooltip > Default
2. Hover on text, it should open the tooltip.


2nikhiltom (approved):
LGTM

Gururajj77 (approved):
Looks good!

laurenmrice (changes requested):
@preetibansalui When the `autoAlign` prop is set to True, the "bottom" alignment looks fine, but the "bottom-left" and "bottom-right" caret is visually off. I also think that if the user is choosing a "bottom-left" or "bottom-right" alignment, the tooltip container needs to be vertically flush with the definition text depending on which alignment is being used. 

Definition tooltip (auto align_ true)

---

- Also unrelated, but there seems to be some more spacing alignment bugs with some of the alignment options that is unrelated to this PR, and I can make a separate issue for that in our repo.

preetibansalui:
> @preetibansalui When the `autoAlign` prop is set to True, the "bottom" alignment looks fine, but the "bottom-left" and "bottom-right" caret is visually off. I also think that if the user is choosing a "bottom-left" or "bottom-right" alignment, the tooltip container needs to be vertically flush with the definition text depending on which alignment is being used.
> 
> Definition tooltip (auto align_ true)
> * Also unrelated, but there seems to be some more spacing alignment bugs with some of the alignment options that is unrelated to this PR, and I can make a separate issue for that in our repo.

Hey @laurenmrice , Sorry little confusion here. 
Do you want me to add a padding from left/right in case of `bottom-left` or `bottom-right` as you shown me in the image attached?

In that case will it be similar for `top-left` or `top-right` as well?

Also, do we have to tackle it in `tooltip component` as well to make consistency?

laurenmrice:
@preetibansalui

> Do you want me to add a padding from left/right in case of bottom-left or bottom-right as you shown me in the image attached?

When the tooltip container appears, it needs to align vertically flush with the beginning (bottom-left) or end (bottom-right) of the definition tooltip text on the page. If adding padding from left/right helps resolve that then yes.

> In that case will it be similar for top-left or top-right as well?

Yes, if that is possible to do for the `top-left` and `top-right` alignments then those could be adjusted as well with the same logic but on the top instead of the bottom.

> Also, do we have to tackle it in tooltip component as well to make consistency?

Would that be a separate PR from this? I see some spacing issues in the Tooltip component, but it's mostly around where the caret is being placed on the tooltip container.

guidari (commented):
Hey Preeti! 

I guess the caret is in the wrong position, right?

The design specs:
image

Here it is the deploy preview with `bottom-right`. It happens with the other postions like `bottom-left/right` and `top-left/right`

Screenshot 2025-01-16 at 11 28 39

preetibansalui:
> ens with the other postions

Hmm.. I think it's taking the tooltip to the position we are assigning, not sure about what logic it's applying to calculate the arrow position. Let me check how we are showing for other tooltip/popover with large target text.

packages/react/src/components/Popover/index.tsx
```diff
@@ -98,6 +98,11 @@ interface PopoverBaseProps {
    */
   align?: PopoverAlignment;
 
+  /**
+   * Will add padding value to arrow in case of bottom-left, bottom-right or top-left, top-right
+   */
+  arrowPadding?: number;
```
tay1orjones:
```suggestion
  caretPadding?: number;
```

This is minor, but the website docs exclusively refer to it as a caret and not an arrow.

packages/react/src/components/Popover/index.tsx
```diff
@@ -98,6 +98,11 @@ interface PopoverBaseProps {
    */
   align?: PopoverAlignment;
 
+  /**
+   * Will add padding value to arrow in case of bottom-left, bottom-right or top-left, top-right
```
tay1orjones: I think we should avoid specifying the deprecated alignment options. Keeping it experimental for now seems prudent, too. Maybe revising to something like this:

```suggestion
   * **Experimental**: Specify padding between the caret and the popover edge. 
   * This helps avoid the caret visually separating from Popovers with rounded 
   * corners when aligned bottom-start, bottom-end, top-start, or top-end.
```

packages/react/src/components/Tooltip/DefinitionTooltip.stories.js
```diff
@@ -50,11 +50,14 @@ export const Default = (args) => {
     <p>
       Custom domains direct requests for your apps in this Cloud Foundry
       organization to a{' '}
-      <DefinitionTooltip openOnHover definition={definition} {...args}>
-        URL
+      <DefinitionTooltip
+        openOnHover
+        autoAlign
```
tay1orjones: Using experimental props on default stories could be troublesome because devs use these as a snippet to copy/paste. A new `with autoAlign` story could be added instead that has the ability to scroll and validate the autoAlign functionality is working correctly.

packages/react/src/components/Toggletip/index.tsx
```diff
@@ -225,9 +225,9 @@ Toggletip.propTypes = {
   ]),
 
   /**
-   * Provide an offset value for alignment axis.
+   * Will add padding value to arrow in case of bottom-left, bottom-right or top-left, top-right
    */
-  alignmentAxisOffset: PropTypes.number,
+  arrowPadding: PropTypes.number,
```
tay1orjones: Mirror the `caretPadding` name change and updated prop/type comment from Popover to here

tay1orjones (changes requested):
A few small things I saw in the comments below. Does this fix the root issue of DefinitionTooltip's misplaced caret with long text _without_ having to turn on `autoAlign`?

packages/react/src/components/Popover/index.tsx
preetibansalui: Sounds good to me, changes applied as suggested.

packages/react/src/components/Tooltip/DefinitionTooltip.stories.js
preetibansalui: Made a separate story 'With Large Text'.

preetibansalui: Enabled the controls of this new story for testing purpose, will remove those controls before merge.

tay1orjones:
@preetibansalui just wanted to confirm this can be fixed without having to turn on `autoAlign`?

preetibansalui:
> @preetibansalui just wanted to confirm this can be fixed without having to turn on `autoAlign`?

let me take a look without autoAlign

preetibansalui:
> > @preetibansalui just wanted to confirm this can be fixed without having to turn on `autoAlign`?
> 
> let me take a look without autoAlign

Hey @tay1orjones, Looks like we can do it without autoAlign as well. I have pushed the changes. Thanks for the suggestion. Now it will work fine with/without autoAlign. In the storybook, I have not set autoAlign to true.

Will remove the controls of 'With Large Text' story before merge.

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -9485,15 +9488,15 @@ Map {
         ],
         "type": "oneOf",
       },
-      "alignmentAxisOffset": Object {
-        "type": "number",
-      },
```
tay1orjones: Is this a breaking change?

packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -9485,15 +9488,15 @@ Map {
...
       "as": Object {
         "type": "elementType",
       },
       "autoAlign": Object {
         "type": "bool",
       },
+      "caretPadding": Object {
+        "type": "number",
+      },
```
tay1orjones: Is this still needed now that we have the css-based fix?

tay1orjones (approved):
LGTM! 🎉  Just need to remove the test story before merge

laurenmrice (approved):

PR #17972, opened 2024-11-05
fix(ai-label): update classes from slug to ai-label

### LINKED ISSUE #17714: Rename `slug` classes to `ai-label`
When Slug was renamed to AILabel(https://github.com/carbon-design-system/carbon/pull/16803), the props and classnames were not included in the rename. This issue is to fix that.

This rename needs to be backwards compatible. The approach we could take could be:

- [ ] Add new duplicate classname selectors to the `ailabel.scss` styles
  ```diff
  - .#{$prefix}--slug .#{$prefix}--slug__button {
  
  + .#{$prefix}--slug .#{$prefix}--slug__button,
  + .#{$prefix}--ai-label .#{$prefix}--ai-label__button { 
  ```
- [ ] Update classnames within AILabel components to use the new `#{$prefix}--ai-label` names
- [ ] Create a new sub-issue/tasklist item on https://github.com/carbon-design-system/carbon/issues/15803 for removing these classes as part of v12

> [!NOTE]
> **On casing:** for situations like react where we have to use PascalCase we decided on `AILabel`
> 
> For non-pascalcase situations like classnames I think the codebase convention would be "sausage-case" (or whatever you like to call it :upside_down_face: ): `ai-label`

ariellalgilmore:
Closes #17714 

Update `AILabel` to use classes `ai-label` instead of `slug`

#### Changelog

**Changed**

- updated `AILabel` component classes and updated old doc links
- updated all components using `AILabel` to have duplicate scss selectors for both `ai-label` and `slug` classes
   -  the `slug` classes will eventually be removed in v12
- updated date-picker fluid, so that styles would match regular date-picker to have the ai-gradient be applied on the input level

#### Testing / Reviewing

Confirm components using AILabel still show/hide ai gradients when using AILabel inside the component


kennylam (approved):
LGTM!

guidari (approved):
LGTM

Issue #17003, opened 2024-07-19
`autoAlign`: Enable consumers to configure `autoAlign`-related props from anywhere in their react tree


tay1orjones:
Multiple components recently gained support for `autoAlign` behavior utilizing floating-ui. Oftentimes these components rely on one another, all the way down to a primitive, e.g.

* `IconButton` ➡️ `Tooltip` ➡️ `Popover`

While `autoAlign` is a single prop, there are multiple other props related to the positioning of floating elements that should be configurable by a consumer. One example is the `align` prop. We've had to add this prop to all these components and ensure they're passed down properly. 

```js
<IconButton align="top"/>
// |__ passes to <Tooltip align="top" />
//    |__ passes to <Popover align="top" />
```

This approach of passing props, sometimes called "prop drilling", is fine but there are some downsides:

* The prop API of all components becomes cluttered with configuration options
* The consumer must configure these props over and over for each instance of say, `<IconButton/>`.

## Goal

Provide a consumer a way to configure one/many popover-based elements in a single place, from anywhere within their react tree.

## Solution options

The [react docs](https://react.dev/learn/passing-data-deeply-with-context#before-you-use-context) suggest to start with prop drilling, then "extract components and pass JSX as children to them", before finally exploring using context.

### 1. Prop drilling

This is the existing approach.

### 2. Composition

Components such as IconButton could be updated to modify how they handle `children`. If a Tooltip is provided as `children`, there could be special treatment for that element all the way down the tree.

```js
<IconButton>
  <Tooltip align="top"
</IconButton>
// |__ clones the tooltip child, adds any required extra props, and ends up rendering <Tooltip align="top" {...requiredExtraProps} />
//    |__ which renders a <Popover align="top" {...requiredExtraProps} />
```

This gives consumers more direct control of the underlying popover-based elements.

This would need to work alongside the existing prop-drilling approach, so the prop API would not change. We could update all the documentation though to show this new approach as an option, maybe the "recommended" option.

### 3. Context

Expose a new context to consumers that can be configured anywhere in their react tree. This context would house all the popover-related props that a consumer would want to configure all at the same time - `align`, `autoAlign`, `autoAlignBoundary`, etc. Props like `label` may not make sense to include in this context since consumers are likely going to need to configure content on a per-popover basis. 

This would require a new (`PopoverContext`? `PopoverPlacementContext`?) context provider component for consumers to set the values:

```js
[...]

PR #17982, opened 2024-11-06
fix: arrow breaking in AILabel popover

### LINKED ISSUE #17965: [Bug]: Arrow looks breaking in the AI explainability popover
### Package

@carbon/react

### Browser

_No response_

### Package version

11.69.0

### React version

_No response_

### Description

Arrow looks breaking in the AI explainability popover in case of align values like 'bottom-left', 'bottom-right', 'top-right', 'top-left' etc. Also this is for AILabel icon size is 'mini', 'xs' or '2xs'. 

To resolve the issue, we are planning to add a cross-axis offset of 24px in case of above conditions only.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-ailabel--default

### Steps to reproduce

Image

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

preetibansalui:
Closes #17965 

Arrow looks breaking in the AI explainability popover in case of small icons.

#### Changelog

added alignmentAxis in offset for popover.

**Changed**

passing a prop from AILabel to Toggletip component in case of 'XS', '2xs' or 'mini' icon.

#### Testing / Reviewing

1. Go to Storybook > AILabel > Default
2. hover on AI Label, the popover should not looks breaking.


aagonzales (commented):
The caret looks like its 1px too low.
image

preetibansalui:
> The caret looks like its 1px too low. image

Thanks Anna for noticing this. It was caused by recent renaming of slug to ai-label, which is fixed now.

aagonzales (approved):
Looks good to me!

guidari (approved):
LGTM! 🚀

ariellalgilmore (approved):
LGTM!

packages/react/src/components/Popover/index.tsx
```diff
@@ -172,6 +172,7 @@ export const Popover: PopoverComponent = React.forwardRef(
       highContrast = false,
       onRequestClose,
       open,
+      alignmentAxisOffset,
```
tay1orjones, line 175: Does this produce an proptypes or ts error/warning because it's not defined in proptypes or the interface?

packages/react/src/components/Popover/index.tsx
```diff
@@ -257,7 +260,14 @@ export const Popover: PopoverComponent = React.forwardRef(
 
             // Middleware order matters, arrow should be last
             middleware: [
-              offset(!isTabTip ? popoverDimensions?.current?.offset : 0),
+              offset(
+                !isTabTip
+                  ? {
+                      alignmentAxis: alignmentAxisOffset,
+                      mainAxis: popoverDimensions?.current?.offset,
```
tay1orjones, line 267: This isn't something blocking this PR and doesn't need changed, but one concern I just mentioned a comment on https://github.com/carbon-design-system/carbon/issues/17003#issuecomment-2514805003 is that as we add more of these config values, they bloat the interface/api of these components. Especially when we haven't identified a use case from a consumer for this, it's just something we need internally for our use w/ AILabel.

The mainAxis offset is already pulled from a custom property instead of a prop value. Doing this for other values as well might be one way to scale these types of config values without having to have a prop for each one. Could be used in tandem w/ however we end up providing a way for consumers to config floating-ui from anywhere in their react tree.

packages/react/src/components/Popover/index.tsx
preetibansalui: Hi Taylor, As discussed, its not giving any warning.

packages/react/src/components/Popover/index.tsx
preetibansalui: I understood your point, I added a prop here instead of some calculations so that it could be changed to different cases as well which might come in future as we have now for AILabel.

tay1orjones (approved):

PR #18014, opened 2024-11-07
fix(inline-dropdown): remove unnecessary column with props.hideLabel

janhassel:
This fixes a bug where an inline dropdown with `props.hideLabel` is trailed with an unnecessary grid column, leading to dead space and layout issues.

Before
image

After
image

#### Changelog

**Changed**

- Added selector to change the `grid-template` when the inline dropdown has a hidden label.

#### Testing / Reviewing

```jsx
<Dropdown
  …
  type="inline"
  hideLabel
/>
```


alisonjoseph (approved):
LGTM! thanks

ariellalgilmore (approved):
LGTM! just need to do a `yarn format`

PR #18015, opened 2024-11-07
fix(pagination-nav): enforce a minimum of 4 items shown

janhassel:
The minimum number of items shown on the PaginationNav was intended to be 4. Many parts of the code referenced this but in the `useEffect` that triggers when `props.itemsShown` or `props.items` changes, one of the two functions wasn't provided with this limitation. Therefore, in some scenarios, the component was able to be set to `3` which causes visual and interaction issues.

This PR enforces the intended minimum of 4 and adds this note to the prop types.

#### Changelog

**Changed**

- Ensure a minimum of 4 items (unless `props.items` is less than 4 and therefore no truncation would happen).

#### Testing / Reviewing

1. Start local storybook
2. Navigate to the Playground story for PaginationNav
3. Change `props.itemsShown` to 1, 2, or 3.
4. It should still show 4 items (incl. overflow indicators)

### DIFF
--- a/packages/react/src/components/PaginationNav/PaginationNav.tsx
+++ b/packages/react/src/components/PaginationNav/PaginationNav.tsx
@@ -292,7 +292,7 @@ interface PaginationNavProps
   disableOverflow?: boolean;
 
   /**
-   * The number of items to be shown.
+   * The number of items to be shown (minimum of 4).
    */
   itemsShown?: number;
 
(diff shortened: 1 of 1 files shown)


tay1orjones (approved):
LGTM!

emyarod (approved):
may introduce a conflict with [this PR](https://github.com/carbon-design-system/carbon/pull/18028) but after discussing with Gui, it should be a simple conflict to resolve. looks good to me!

PR #18019, opened 2024-11-07
fix(date-picker): check properties on first render of calendar

### LINKED ISSUE #18018: [Bug]: date-picker inputting min and max date range doesn't work
### Package

@carbon/web-components

### Browser

_No response_

### Package version

v2.16.0

### React version

_No response_

### Description

When applying the `min-date` and `max-date` properties, I expect to see the calendar grayed out / disabled for all dates except for the date range given. However there is no change in the calendar.

### Reproduction/example

https://stackblitz.com/edit/github-7fmqdw?file=index.html

### Steps to reproduce

Open the stackblitz and see that the calendar displays all dates even though the date range has been supplied.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

annawen1:
Closes #18018

Properties applied to the `date-picker` that don't have default values aren't applied to on the first render and are only checked in the `updated` cycle. Hence why it works in storybook when the properties are applied once the `date-picker` has been rendered.

#### Changelog

**Changed**

- add `_setCalendar()` function that sets the calendar
- add checks for properties in the `_instantiateDatePicker()` that will call `_setCalendar()` and apply them on the first render
- use `_setCalendar()` in the `updated()` cycle that checks for changedProperties

#### Testing / Reviewing

run storyook locally and apply the properties (`open`, `min-date`, etc) directly in the playground story code in `date-picker.stories.ts` and see if they are applied when `date-picker` first renders

example:

```     
<cds-date-picker
        ?disabled="${disabled}"
        date-format="${dateFormat}"
        ?readonly="${readonly}"
        min-date="11/09/2024" <----- set a value here
        max-date="11/23/2024" <----- set a value here and see if calendar applies this range on first render

### DIFF
--- a/packages/web-components/src/components/date-picker/date-picker.ts
+++ b/packages/web-components/src/components/date-picker/date-picker.ts
@@ -275,6 +275,93 @@ class CDSDatePicker extends HostListenerMixin(FormMixin(LitElement)) {
   };
 
   /**
+   * Sets calendar options
+   * @param property property to set
+   * @param calendar calendar object
+   */
+  private _setCalendar(property, calendar) {
+    const { disabled, dateFormat, open, readonly, minDate, maxDate, value } =
+      this;
+
+    const { selectorInputFrom, selectorInputTo } = this
+      .constructor as typeof CDSDatePicker;
+    const inputFrom = this.querySelector(
+      selectorInputFrom
+    ) as CDSDatePickerInput;
+    const inputTo = this.querySelector(selectorInputTo) as CDSDatePickerInput;
+
+    if (property === 'dateFormat') {
+      calendar.set({ dateFormat });
+    }
+    if (property === 'date') {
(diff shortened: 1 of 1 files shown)


ariellalgilmore (approved):
LGTM! noticing a weird thing though that the year gets unbolded when i put the min and max locally
Screenshot 2024-11-08 at 10 23 02 AM

annawen1:
> LGTM! noticing a weird thing though that the year gets unbolded when i put the min and max locally Screenshot 2024-11-08 at 10 23 02 AM

Ah! I think this might be intended - if I set the max and min date to be a year appart (ie. 11/12/2024 - 11/12/2025), the year is bolded because there is a little year adjuster. I guess if the dates are within the same year, that functionality gets disabled and the year is unbolded.
Screenshot 2024-11-11 at 2 04 58 PM

kennylam (approved):
LGTM!

PR #18041, opened 2024-11-11
fix(combobox): controlled combobox clear

### LINKED ISSUE #18027: [Bug]: Controllled combobox fires onchange twice when input field is cleared
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.70.0

### React version

_No response_

### Description

When I click the x in the combo box input field to clear the content, the onchange event is fired twice. The first time it fires with a selectedItem value of null which is expected. However then it fires again with a selectedItem value of the previous selection. It should not be called a 2nd time.
This is likely related to https://github.com/carbon-design-system/carbon/issues/17847 which was recently fixed.

### Reproduction/example

https://stackblitz.com/edit/github-k8ejla?file=src%2FApp.jsx&preset=node=

### Steps to reproduce

Run the slackblitz sample with devtools open so you can see the console logs.
Select an item on the ComboBox.
Click the x to clear the input field
Problem: The onChange function is called twice. You can see the debug log printed 2 times to the console with the different selectedItem values

Image

### Suggested Severity

None

### Application/PAL

IBM Cloud IAM-UI

### Code of Conduct

neuescodes:
Closes https://github.com/carbon-design-system/carbon/issues/18027

* update the way onChange is called in comboBox to prevent duplicate calls
* add unit test cases for controlled combobox managed by parent component
* fix console errors related to missing onChange prop from ComboBox.mdx
* fix console errors for components using WithLayer related to wrong proptypes
Screenshot 2024-11-14 at 11 15 19

#### Testing / Reviewing

- add log to [onChange in fullyControlled combobox in ComboBox.stories.js](https://github.com/Neues/carbon/blob/ef9f8b0418431d885f464effeff4ab98bbea78e8/packages/react/src/components/ComboBox/ComboBox.stories.js#L257)
- select different options to see onChange fires only once
- select the same option to see that onChange does not fire
- add log to onChange other ComboBox stories to ensure it is firing the appropriate amount of times
- verify console errors are no longer present


tay1orjones (approved):
LGTM, thanks for taking a second look at this!

guidari (approved):
Hey @Neues Thanks for opening that!
I'll close my [PR](https://github.com/carbon-design-system/carbon/pull/18048) that has a similar fix. I'll fix the format in this one, it is the same error I got.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -770,6 +759,20 @@ const ComboBox = forwardRef(
...
       },
+      onStateChange: ({ type, selectedItem: newSelectedItem }) => {
+        if (
+          type === '__item_click__' &&
+          !isEqual(selectedItemProp, newSelectedItem)
+        ) {
+          onChange({ selectedItem: newSelectedItem });
+        }
+        if (
+          type === '__function_select_item__' ||
```
k02pradeep, line 770: Should use enum

k02pradeep (changes requested):

packages/react/src/components/ComboBox/ComboBox.tsx
neuescodes: This pull request is already merged, and a fix for this has also been merged, see https://github.com/carbon-design-system/carbon/issues/18145

PR #18057, opened 2024-11-13
fix: checkbox invalid message to match figma kit

### LINKED ISSUE #18033: [Bug]: invalidText of horizontal CheckboxGroup is rendered on wrong place
### Package

@carbon/react

### Browser

Firefox

### Package version

1.17.0

### React version

18.3.0

### Description

`CheckboxGroup` component renders the `invalidText` at the different position from the figma library.
With `orientation="horizontal"`, `invalidText` shows at the right of the `Checkbox`s, and I expected it to be shown at the bottom.

Here is the screenshot from our Figma sketch:
Image

And here's the screenshot of the example react component:
Image

### Reproduction/example

https://stackblitz.com/edit/github-emxaww?file=src%2FApp.jsx

### Steps to reproduce

1. Create figma checkbox group with Horizontal=on, and state=Invalid
2. Compare the rendered component with the provided stackblitz example

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #18033 

CheckboxGroup component renders the invalidText at the different position from the figma library.
With orientation="horizontal", invalidText shows at the right of the Checkboxs, and I expected it to be shown at the bottom.

#### Changelog

**Changed**

- Seperated the container of error message from the flex container of radio buttons and hence modified the classes

#### Testing / Reviewing

Go to checkbox and check if invalid message is rendered correctly at bottom


guidari (changes requested):
Hey Riddhi, It is looking good! Just one thing.
I think we should keep the `invalid` and `readonly` message inside the `fieldset`. If you test with VO in the [current Storybook](https://react.carbondesignsystem.com/?path=/story/components-checkbox--playground&args=invalid:!true), you’ll notice that they are grouped together, which helps the screen reader user understand that the message is part of the same group label. 
We can also confirm that with Raghu if needed.

riddhybansal:
> Hey Riddhi, It is looking good! Just one thing. I think we should keep the `invalid` and `readonly` message inside the `fieldset`. If you test with VO in the [current Storybook](https://react.carbondesignsystem.com/?path=/story/components-checkbox--playground&args=invalid:!true), you’ll notice that they are grouped together, which helps the screen reader user understand that the message is part of the same group label. We can also confirm that with Raghu if needed.

Oh I see what you are saying !!

riddhybansal:
Hey @guidari Look at this found an alternative way to fix that bug . Lemme know what do you think about it ?

packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -65,13 +65,16 @@ const CheckboxGroup: React.FC<CheckboxGroupProps> = ({
...
     [`${prefix}--checkbox-group--readonly`]: readOnly,
     [`${prefix}--checkbox-group--invalid`]: !readOnly && invalid,
     [`${prefix}--checkbox-group--warning`]: showWarning,
     [`${prefix}--checkbox-group--slug`]: slug,
   });
 
+  const orientationClass =
+    orientation === 'horizontal'
+      ? `${prefix}--checkbox-group--${orientation}`
+      : '';
```
guidari:
```suggestion
      : undefined;
```
LGTM! Let's just change to `undefined` to hide the empty class.

Screenshot 2024-11-18 at 10 38 00

guidari (approved):
LGTM! 🚀

thyhmdo (approved):

alisonjoseph (changes requested):
This does technically fix the bug, however I think it would be preferable if we just update the styles vs. changing the underlying html structure. I believe you should be able to add a  inline-size (width) of 100% to the invalid or warn text when its horizontal and it will display correctly below the checkboxes.

alisonjoseph (approved):

PR #18061, opened 2024-11-13
fix(select): a11y: stop SR announcement for hidden labels 

### LINKED ISSUE #16613: [a11y]: Select - Playground - JAWS is announcing the name of the label even after hiding the label from playground
### Package

@carbon/react

### Browser

Firefox

### Operating System

_No response_

### Package version

v1.58.0

### React version

18.02

Manual QA

### Assistive technology

JAWS

### Description

From the Select - playground, set the hide label Boolean to "True" and observe the JAWS announcement.
JAWS will announce the name of the label which is visually hidden.

### Recommend Fix
Ensure that visually hidden label is hidden from screen reader as well

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-select--playground&args=hideLabel:!true;readOnly:!false

### Steps to reproduce

none

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/16613

SR should not announce the label when `hideLabel` is true

#### Changelog

**New**
Added `aria-hidden` attribute 

#### Testing / Reviewing

Open From the Select - playground story
set the hide label Boolean to "True" and observe the SR do not announce label.


2nikhiltom:
Gower confirmed in CAG call that the label has to be announced, otherwise that might lead to a11y issues.

PR #18077, opened 2024-11-14
feat(tag): new decorator prop

### LINKED ISSUE #18002: Decorator: Tag

ariellalgilmore:
Closes #18002 

Updates Tag to include new decorator prop

#### Changelog

**New**

- new decorator prop, classes, and styles

**Changed**

- deprecated slug prop
- update rendering slug to render the decorator component but still set the size if it's an AILabel
- updated tests
- note I only updated Tag and Dismissible Tag and not the Tag using the `filter` prop since it's deprecated

#### Testing / Reviewing

Check tag with AILabel storybook ( i think the tooltip still looks a little odd, but maybe something to fix once people start using it and we can have design take a better look)

TODO:

- [ ] remove tag with tooltip before merging!


alisonjoseph (approved):
LGTM!

tay1orjones (approved):

PR #18096, opened 2024-11-18
fix(selectable-tile): don't require deprecated property "value"

### LINKED ISSUE #16977: [Bug]: SelectableTile: deprecated `value` prop
### Package

@carbon/react

### Browser

_No response_

### Package version

1.61.0

### React version

18.0.0

### Description

The value prop is marked as `deprecated` despite being required.

### Reproduction/example

https://stackblitz.com/edit/vitejs-vite-dbavtk?file=src%2FApp.tsx

### Steps to reproduce

1. Use the SelectableTile component with TypeScript,
2. Add value prop
3. Have it marked as deprecated

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

wkeese:
Fixes #16977.
Refs #13537, #13631.

Update PropTypes so "name" and "value" are reported as deprecated. (The descriptions already included @deprecated, but that's different.)

More importantly, stop marking "value" as required. Properties cannot be deprecated and required at the same time.

#### Changelog

**Changed**

- declarations for SelectableTile "name" and "value" properties

#### Testing / Reviewing

Tested locally.  Updated snapshot, and updated test file to stop using the deprecated parameters.

### DIFF
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -6984,9 +6984,7 @@ Map {
         "type": "string",
       },
       "light": [Function],
-      "name": Object {
-        "type": "string",
-      },
+      "name": [Function],
       "onChange": Object {
         "type": "func",
       },
@@ -7008,20 +7006,7 @@ Map {
       "title": Object {
         "type": "string",
       },
-      "value": Object {
-        "args": Array [
-          Array [
-            Object {
-              "type": "string",
-            },
-            Object {
-              "type": "number",
-            },
-          ],
-        ],
-        "isRequired": true,
-        "type": "oneOfType",
-      },
+      "value": [Function],
     },
     "render": [Function],
   },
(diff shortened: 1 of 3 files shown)


wkeese:
I saw that, but isn't that an infrastructure problem, some issue with Jenkins etc.?

For me, the tests are all failing locally even on the main/ branch, with an error that

> Cannot find module '@carbon/feature-flags' from 'packages/react/src/components/FeatureFlags/index.tsx'

That may be unrelated though.

tay1orjones (approved):
LGTM, just need to update the public api snapshot and push it up. `yarn test -u` at the root will do it

wkeese:
Ah thanks, I got it to work via running this from the root:

```
yarn install
yarn run build
yarn test -u
```

(It's never clear to me when to run from root vs. when to run from packages/react.)

I force-pushed the latest.

wkeese:
The visual diffs are unrelated to this PR, you can see the same thing happening in #18106.

kennylam (approved):

Issue #18084, opened 2024-11-15
[Menu and Menu buttons] Nested menus: Research


laurenmrice:
## Acceptance criteria

Design needs to decide if we want to allow nested menus within the Menu button component. We allow this for the context menu variant when the need for these menus seems higher for this particular use case, But is this also acceptable for Menu buttons?

- [ ] Research if the industry allows this. What is the best practice for including nested menus in a component like Menu button. Should we also allow this for Combo button? Should we allow this for Overflow menu as well (A team brought this up recently in Slack ([thread](https://ibm-studios.slack.com/archives/C0M053VPT/p1775842833851819))? This is in a feature flag, but lets review it.
- [ ] Consult a11y to ensure this is acceptable from an a11y perspective.
- [ ] Playback findings and make a design decision of how to move forward.

----

Related: [#15875](https://github.com/carbon-design-system/carbon/issues/15875)

alina-jacob:
Also investigate usage of caret vs chevron based on [this discussion](https://ibm-studios.slack.com/archives/C0M053VPT/p1776078058923249?thread_ts=1775842833.851819&cid=C0M053VPT)!

Issue #17725, opened 2024-10-11
[Menu and Menu buttons] Nested menus: Implementation


dianasanborn:
### The problem

Add submenu example and guidance to the menu button 🌳

**This is part of the effort to consolidate the [Sustainability Software PAL buttons](https://www.figma.com/design/DXsHD2rqGTcHRfsRKMkW86/(v10)-PAL-Components-%E2%80%93%C2%A0Sustainability-Software?node-id=18808-16544&node-type=section&t=Vb5KENG6DVKUJYKc-11) guidance with [Carbon's menu button](https://carbondesignsystem.com/components/menu-buttons/usage/) guidance.**

Figma file: [Menu button comparison](https://www.figma.com/design/CtxMZE9hcDJeU3QBRMk5QW/SSW-PAL-%26-Carbon-consolidation?node-id=348-62&node-type=section&t=4gxPxSSLChQG3iYq-11)

Currently there are two very similar components available in both Sustainability PAL and Carbon: the menu button and combo button. Carbon's menu button was created based on Sustainability's button extension and other use cases. 

Functionally, they both work the same. There are some slight stylistic differences but the Sustainability org will inherit Carbon's menu button styling. 

One scenario that is not included in Carbon's guidance is the use of a submenu from the menu. This uses [Carbon's menu](https://carbondesignsystem.com/components/menu/usage/#submenus) component.

![Image](https://github.com/user-attachments/assets/9d834172-39c0-4e8c-9f89-9601f59a2c21)

### The solution

Based on conversations with the Carbon team:
- This is likely already possible to do since the menu button uses the menu component, but it's not available OOTB
- Consider putting an image example and guidance in the docs to let teams know that this capability is available
- Consider putting a prop or example of this in the storybook in code
- Diana had a discussion with Michael Gower on the accessibility considerations and there are no concerns with the use of a submenu in the menu button. 

### Examples

(Ignore button styling for these screenshots — we will inherit Carbon's button styling)

Menu button
![Image](https://github.com/user-attachments/assets/51e8d0ee-17dd-4340-943d-376a7bbffd8f)

Combo button
![Image](https://github.com/user-attachments/assets/db800a50-d5d0-463a-a3c3-2afc388d40b8)

### Application/PAL

Sustainability Software PAL

### Business priority

None

### Available extra resources

[...]

Issue #15875, opened 2024-03-04
[Bug]: Menu button nested menu items warning


misiekhardcore:
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.51.1

### React version

17.0.2

### Description

When using the `MenuButton` with nested `MenuItem`s we get the warning in the console:
![Screenshot from 2024-03-04 08-51-08](https://github.com/carbon-design-system/carbon/assets/58469680/5bce1fa5-3047-44ef-9260-15c415c45822)
I can see that in `MenuButton` implementation the inner `Menu`s prop `mode` is set to `basic`, so this is the reason I assume, but what is it for a purpose or is it a bug?

### Reproduction/example

https://stackblitz.com/edit/github-rztleo?file=src%2FApp.jsx

### Steps to reproduce

- just open the menu button and you should see the mentioned error in the browsers console

### Suggested Severity

None

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
@misiekhardcore Hey, thanks for reporting this. The stackblitz link you posted isn't working for me.

misiekhardcore:
hey @tay1orjones I think it was set to private. Please let me know if it works now

misiekhardcore:
Hey @guidari I've seen that description but it does not really explain why in `MenuButton` this is set to basic (the description says it will be determined automatically but in this case it is not)

janhassel:
Hey Michał, the behavior is working as expected. Both MenuButtons and ComboButtons are not intended to be used for complex scenarios like nested or selectable items. This is reserved for OverflowMenu (only with `enable-v12-overflowmenu` flag) and the core Menu (used by `useContextMenu`).

@thyhmdo Did we ever clarify that to the design guidance? I can't find it right now.

misiekhardcore:
@janhassel I see, thanks for the clarification. For me the `OverflowMenu` is not a solution because I need it to be primary and with text. Something like this
![Screenshot from 2024-03-11 15-23-32](https://github.com/carbon-design-system/carbon/assets/58469680/d8ba5e53-c5ad-469b-ab2b-41c0370bdbdc)
[...]

PR #18153, opened 2024-11-25
feat(menu): deprecate props.mode / always support full capabilities + icons

### LINKED ISSUE #18148: [Feature Request]: v12 overflowMenu support for `renderIcon` prop on `MenuItem`. and `checkbox`
### The problem

Migration from v11 to v12 `OverflowMenu` (experimental). doesn't provide an option to render icons in `MenuItem` AFAIK.

We have an [issue](https://github.com/carbon-design-system/ibm-products/issues/6328) in C4IP where

In v11, we utilized the `OverflowMenu` component with `OverflowMenuItem` as its child, which supported an `itemText` prop that accepted a React node. This allowed us to pass both an icon and text together.

To render the checkbox, we implemented a [workaround](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/AddSelect/AddSelectColumn.js#L149), though it was not ideal, as shown below. This approach needs to be updated to align with Carbon's implementation. However, I could not find an option to achieve the same result with the `MenuItem` component. Do you have any recommendations on how to approach this?

Image

In v12 of `OverflowMenu`, the `MenuItem` component is used as a child element with properties such as `label` and `shortcut`, both of which are strings. Additionally, the `renderIcon` prop does not function as expected on `MenuItem` unless the parent component is set to `mode='basic'`. However, this configuration is only applicable to the `Menu` component and cannot be used with `OverflowMenu`.

### The solution

ability to render checkboxes and icons for v12 `OverflowMenu`, like we did in v11.
<table>
  <tr>
    <td>Image 1</td>
    <td>Image 2</td>
  </tr>
</table>

### Examples

v11 implementation (go to folder 1 to see 2 overflow menu use cases in action)
https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-patterns-add-and-select-multiaddselect--with-hierarchy&globals=viewport:basic

### Application/PAL

c4ip

### Business priority

### Available extra resources

_No response_

### Code of Conduct

janhassel:
Closes #18148, #15875, #17725
Related #18084, https://ibm-analytics.slack.com/archives/C0M053VPT/p1715933314490569

Deprecates `props.mode` for `Menu` and always support the full list of capabilities:
- Selectable items (single and multi select)
- Nested menus
- Icons

The original concern of having a checkmark icon and a custom icon on the same visual level is being addressed by placing them in separate columns. The original concern was that users wouldn't be able to distinguish quickly what's a decorative icon and what is a status icon. The design was discussed with @thyhmdo and @ariellalgilmore. 

#### Changelog

**Changed**

- Updated styles to support a status icon column in menu items
- Moved checkmark icon logic from MenuItemSelectable and MenuItemRadioGroup into MenuItem based on `aria-checked`

**Removed**

- Removed all usage of `props.mode`
- Removed mode logic from menu context
- Removed warnings for invalid mode and feature pairings

---

image

@thyhmdo Side note: even if technically selectable items inside of menu and combo buttons would be allowed with this approach, we should probably add some details in the design guidance discouraging this. Also, here's our related discussion: https://ibm-analytics.slack.com/archives/C07HHD0CNR4/p1725459515044009

General side note: this adds support for icons for `MenuItemSelectable`. We should probably add support for `MenuItemRadioGroup` as well but it's not compatible with the current structure. My long-term thought is to replace the `props.items` with something like a `MenuItemRadioOption` component that should be used as children of `MenuItemRadioGroup`. But that's probably better discussed in a separate issue.

---

Opening as draft as an issue occurs with the nested `MenuButton`. For some reason, the `overflow-y: auto` causes the nested menu to be clipped in `MenuButton` but not `Menu`. I'll need to investigate further.


janhassel:
@tay1orjones It seems that using `transform` on the root menu causes the child menu to base it's boundaries on the parent for some reason. I couldn't find any related spec detailing this and I haven't seen this before (I think).

As a test, I wrote a janky middleware for floating-ui that adds the intended transform to `top` and `left` and it resolves the problem. Do you have an idea how to proceed here?

```js
floatingSize({
  apply({ elements }) {
    const { transform, top: topPx, left: leftPx } = elements.floating.style;

    if (transform) {
      const [_, xPx, yPx] = transform.match(/translate\((.+), (.+)\)/);
      const values = [topPx, yPx, leftPx, xPx].map((px) => Number(px.replace('px', '')));

      Object.assign(elements.floating.style, {
        top: `${values[0] + values[1]}px`,
        left: `${values[2] + values[3]}px`,
        transform: 'unset',
        willChange: 'unset',
      });
    }
  }
})
```

With middleware:
image

Without middleware (Safari):
image

You can see the box-shadow of the nested menu on the right side of "Delete", showing that it renders but is clipped.

In Edge, a scroll is added to the parent (without middleware):
image

tay1orjones:
@janhassel I dug around a bit and found this https://github.com/floating-ui/floating-ui/issues/2858

It might be worth trying to set the boundary to `[]` instead of the default. Might be why you're seeing that specifically with nesting.

janhassel:
@tay1orjones Thank you for the link! After further research I found the culprit: when an element is using `transform` it creates a new containing block which all (also fixed-positioned) descendants reference.

> For elements whose layout is governed by the CSS box model, any value other than none for the transform property also causes the element to establish a containing block for all descendants. Its padding box will be used to layout for all of its absolute-position descendants, fixed-position descendants, and descendant fixed background attachments.

https://www.w3.org/TR/2019/CR-css-transforms-1-20190214/#current-transformation-matrix-computation

This caused the submenu to fall into the `overflow` clipping of the parent menu which we need in order to offer vertical scroll. I updated the configuration for `useFloating` to use absolute positioning instead of transform. All demo stories seem to work as intended now.

ariellalgilmore (approved):
this looks great to me!

tay1orjones (approved):
Sorry for the delay, LGTM!

Issue #18145, opened 2024-11-22
[Bug]: ComboBox: onChange is not called when building the code in production mode


flannanl:
### Package

@carbon/react

### Package version

v1.71

### Description

This is a regression caused by https://github.com/carbon-design-system/carbon/pull/18041. When using webpack to build the code in "production" mode (not sure if it is caused by the minimizer or something), the downshift type is behaving different, and this is causing the `onChange` not to be fired at all. It works, however, if I build the code for local development without minimizing / ugilify the JS.

So I have added a few `console.log` statements to prove the behaviour.

```
    onStateChange: _ref6 => {
      let {
        type,
        selectedItem: newSelectedItem
      } = _ref6;
      console.log('type', type)
      console.log('selectedItemProp', selectedItemProp)
      console.log('newSelectedItem', newSelectedItem)
      console.log('fire onChange 1', type === '__item_click__' && !isEqual(selectedItemProp, newSelectedItem))
      console.log('fire onChange 2', type === '__function_select_item__' || type === '__input_keydown_enter__')
      if (type === '__item_click__' && !isEqual(selectedItemProp, newSelectedItem)) {
        onChange({
          selectedItem: newSelectedItem
        });
      }
      if (type === '__function_select_item__' || type === '__input_keydown_enter__') {
        onChange({
          selectedItem: newSelectedItem
        });
      }
    },
```

When building the code for local development, I see these printed in the console as:

```
type __input_click__
ComboBox.js:460 selectedItemProp null
ComboBox.js:461 newSelectedItem undefined
ComboBox.js:462 fire onChange 1 false
ComboBox.js:463 fire onChange 2 false

ComboBox.js:459 type __item_click__
ComboBox.js:460 selectedItemProp null
ComboBox.js:461 newSelectedItem {id: '471af655-c123-52a6-5c79-aed05c34aaf7', name: 'flo21-apikey', name_escaped: 'flo21-apikey', group_id: 'default', group_name: 'Default', …}
ComboBox.js:462 fire onChange 1 true
ComboBox.js:463 fire onChange 2 false

ComboBox.js:459 type __function_set_highlighted_index__
ComboBox.js:460 selectedItemProp null
ComboBox.js:461 newSelectedItem undefined
ComboBox.js:462 fire onChange 1 false
ComboBox.js:463 fire onChange 2 false
```

When building the code for production, I see these printed in the console instead:

```
type 10
app.js:2 selectedItemProp null
[...]

PR #18154, opened 2024-11-25
fix(combobox): use enum for state change types

a88zach:
Closes #

[#18145](https://github.com/carbon-design-system/carbon/issues/18145)

Webpack will optimize code size and will sometimes convert string enums to integers.  This change uses the enum from Downshift so that if Webpack converts the string enum to integers, all code will be converted correctly

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -3,7 +3,9 @@
   "projectOwner": "carbon-design-system",
   "repoType": "github",
   "repoHost": "https://github.com",
-  "files": ["README.md"],
+  "files": [
+    "README.md"
+  ],
   "imageSize": 100,
   "contributorsPerLine": 7,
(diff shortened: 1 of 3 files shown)


guidari (approved):
Thanks for doing that! LGTM 🚀

The `.all-contributorsrc` file has some changes, mostly it is format. Can you just rollback that, just to keep as it was. You can add your lines and save without format.

a88zach:
> Thanks for doing that! LGTM 🚀
> 
> 
> 
> The `.all-contributorsrc` file has some changes, mostly it is format. Can you just rollback that, just to keep as it was. You can add your lines and save without format.

I can,  but I'm out of office until next week

Gururajj77 (approved):
Looks good

PR #18155, opened 2024-11-25
fix: fixed aira-current in breadcrumb

### LINKED ISSUE #18113: [a11y]: Breadcrumbs don't inform screen reader about current page
### Package

@carbon/react

### Browser

_No response_

### Operating System

_No response_

### Package version

1.70.0

### React version

_No response_

AVT3 - JAWS screen reader

### Assistive technology

JAWS

### Description

The breadcrumbs component should inform screen reader users which is the current page in the breadcrumbs.

### WCAG 2.1 Violation

4.1.2 Name, Role, Value

### Reproduction/example

https://stackblitz.com/edit/github-dupbrt?file=src%2FApp.jsx

### Steps to reproduce

Use a screen reader with a Breadcrumb component, and it will not distinguish which breadcrumb item is for the current page making it more difficult for a screen reader user to understand where they are.

### Suggested Severity

None

### Code of Conduct

guidari:
Closes #18113 

Added the `aria-current` to the `span` tag that it is generated in the DOM

#### Testing / Reviewing

- Open Breadcrumb With Overflow Menu stories and turn on the VO.
- You should hear "Breadcrumb 6, current item"

Screenshot 2024-11-25 at 14 10 15


kennylam:
Hm, I'm unable to focus on text elements no matter the screen reader used (MacOS/iOS VoiceOver, JAWS).

guidari:
> Hm, I'm unable to focus on text elements no matter the screen reader used (MacOS/iOS VoiceOver, JAWS).

@kennylam 
Yeah! You can't use tabs on the Breadcrumb that has the `aria-current`. You should use arrow navigation and you can hear the VO announcing the `aria-current`.

kennylam:
> You should use arrow navigation and you can hear the VO announcing the `aria-current`.

But mobile users would not be able to navigate this way, right?

guidari:
Good question Kenny, I'm not sure. I know it does not receive foxu because it should not be a link.

I know that we have this guidelines on our website.

https://carbondesignsystem.com/components/breadcrumb/accessibility/#keyboard-interactions

Screenshot 2024-12-04 at 12 31 07

kennylam:
Maybe we can check with Ragu, otherwise this looks fine to me.

guidari:
Here on [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/) says that it should still be an intereactive element. Maybe we should change our specs. 🤔

tay1orjones (approved):

kennylam (approved):

PR #18162, opened 2024-11-26
fix: changed from lodash to es-toolkit

### LINKED ISSUE #17731: [Feature Request]: Drop per-method lodash dependencies
### The problem

You are using several [single-method lodash package dependencies](https://github.com/search?q=repo%3Acarbon-design-system%2Fcarbon+lodash.&type=code) in your codebase, which is [discouraged even by the authors](https://lodash.com/per-method-packages).

The reason I came up to this is that we have an app where size is a pretty big issue and inspection of the built artifact shows duplication of lodash code... (where the per-method ones are also needlessly big).

Image

### The solution

- drop all dependencies [where native ECMAScript methods exist](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore)
- consider if other methods could be much simplified by using a simplified and project-shared utility that only uses the aforementioned ES6 methods
- replace the rest with a tree-shake-able import from (`lodash-es`)[https://www.npmjs.com/package/lodash-es]… ideally with a very permissive range on `lodash` version to make it more likely that it will get deduplicated.

They also have some kind of code mod in the linked docs page, but I have not used it, so I cannot vouch for it.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

### LINKED ISSUE #18097: @carbon/react uses lodash.findlast@4.6.0 which has security issues
Summary
@carbon/react uses lodash.findlast@4.6.0  which has security issues

Details
Criticial and High priority security issues
https://github.com/advisories/GHSA-jf85-cpcp-j695 (BDSA-2019-2112)
https://github.com/advisories/GHSA-p6mc-m468-83gw (BDSA-2020-1674)
https://github.com/advisories/GHSA-35jh-r3h4-6jhm (BDSA-2021-0392)

Impact
Because the dependency of the @carbon/react v1.70.0 has dependency on lodash v4.6.0, we are not to release the product

guidari:
Closes #17731 

We are moving from `lodash` to [es-toolkit](https://es-toolkit.slash.page/) which is a modern library. That change will solve security issues we were having in `lodash` and also it will lead to a smaller final bundle and better performance.

Replaced a few lodash code to use pure JavaScript. Here it is a [Stackblitz](https://stackblitz.com/edit/github-mnqpei?file=src%2FOmit.jsx,src%2FApp.jsx,src%2Findex.scss,src%2FFindLast.jsx,src%2Fmain.jsx) with the code changes examples to show that it works.

#### Changelog

**New**

- Added `es-toolkit` dependency

**Changed**

- Removed `lodash.findlast` to use pure Javascript code
- Removed `lodash.omit` to use pure Javascript code

**Removed**

- Removed `lodash` dependencies

#### Testing / Reviewing

- Build should work
- CI should pass


packages/react/src/components/Slider/Slider.tsx
```diff
@@ -13,7 +13,7 @@ import React, {
 import PropTypes from 'prop-types';
 
 import classNames from 'classnames';
-import throttle from 'lodash.throttle';
+import { throttle } from 'es-toolkit/compat';
```
preetibansalui, line 16: Hey Gui, Thanks for the changes, its Awesome!

I tried to read about throttle from es-toolkit and found that it will accept only 2 parameters like `throttle(func, wait)` while for lodash it's like `throttle(func, wait, options)`

In this file `Slider.tsx` we were using throttle from lodash as
 ```
onDrag = throttle(this._onDrag, EVENT_THROTTLE, {
    leading: true,
    trailing: false,
  });
```

Which means the function `this._onDrag` is executed immediately when the event is triggered , but it will not be executed again at the end of the throttling interval.

Can you please make sure to change the code as well as per es-toolkit to achieve the same functionality.

guidari: Hey Preeti! 
I tested with lodash and es-toolkit and the behaviour is the same. We are using the `es-toolkit/compat` that has [full compatibility](https://es-toolkit.slash.page/reference/function/throttle.html#compatibility-with-lodash) with lodash, so we don't have to change the code. All `es-toolkit` functions have a "compat" version to ensure we don’t need to change anything in our code.

preetibansalui: Oh yes, the link is saying it supports `leading` and `trailing` options as well. Thanks!

preetibansalui (approved):

ariellalgilmore (approved):
Great! LGTM!

ariellalgilmore:
looks like just need to resolve conflicts

PR #18188, opened 2024-12-03
fix(composed-modal): don't close on mouse drag and release outside modal

### LINKED ISSUE #18187: [Bug]: ComposedModal closes when user clicks inside modal, drags outside and releases mouse button (Chrome only)
### Package

@carbon/react

### Browser

Chrome

### Package version

v11.71.0

### React version

v18.2.0

### Description

When using ComposedModal to implement custom modal dialogs, if the user presses left mouse button inside modal window, drags the cursor outside the modal to the background layer, and then releases mouse button, the modal closes.

### Reproduction/example

https://stackblitz.com/edit/github-znrzua?file=src%2FApp.jsx

### Steps to reproduce

Open a link to Stackblitz example on Google Chrome or a Chrome-like browser; make the preview window wider, so modal is not filling the entire window. Then click inside modal body, and while not releasing the mouse button, drag the cursor away from modal window to background layer, and then release the mouse button. Modal window will close, which is not the expected behavior for the user.

On Firefox, this does not reproduce. The problem may be that on Chrome an `event.target` element for onClick event is the element that was under the cursor when the mouse button was released, instead of the element, on which the user held the mouse button initially.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

vsvsv:
Closes #18187

Fixed incorrect behavior of ComposedModal closing when user clicks inside modal window, drags cursor outside and releases the mouse button in Chrome.

#### Changelog

**Changed**

- instead of checking only `event.target` node of 'onClick' event also check `event.target` of 'onMouseDown' event, and close the modal only if both are not inside the modal window node.


vsvsv:

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -299,14 +302,22 @@ const ComposedModal = React.forwardRef<HTMLDivElement, ComposedModalProps>(
...
 
+    function handleOnMouseDown(evt: React.MouseEvent<HTMLDivElement>) {
+      const target = evt.target as Node;
+      onMouseDownTarget.current = target;
+    }
+
     function handleOnClick(evt: React.MouseEvent<HTMLDivElement>) {
       const target = evt.target as Node;
+      const mouseDownTarget = onMouseDownTarget.current as Node;
+      console.log({ target, mouseDownTarget });
```
tay1orjones:
```suggestion
```

tay1orjones (changes requested):
Thanks for taking a look at this. Looks good to me, but could you add some tests to cover this functionality?

vsvsv:
@tay1orjones Sure, added unit tests to cover this fix.

ariellalgilmore (approved):
LGTM!

tay1orjones (approved):

PR #18229, opened 2024-12-09
chore: added .nx/cache to gitignore

Gururajj77:
Added `.nx/cache` folder to .gitignore file, as suggested by the Nx engineer under a question in [Nx repository](https://github.com/nrwl/nx/discussions/20813)

#### Testing / Reviewing

make sure CI passes


tay1orjones (approved):
After this we can remove the existing files, too. So they're no longer included in the upstream repo

.gitignore
```diff
@@ -61,3 +61,6 @@ blob-report
 
 # Local token
 .env
+
+# nx local buiild cache
+.nx/cache
```
tay1orjones:
```suggestion
.nx/cache
.nx/workspace-data
```

tay1orjones: from https://github.com/nrwl/nx/discussions/20813#discussioncomment-9764011

tay1orjones (approved):

PR #18236, opened 2024-12-10
chore: swap assert syntax with 'with' to accommodate Node >=22

annawen1:
When running the web-components build on Node v23, the following line errors out

```
import * as packageJson from '../package.json' assert { type: 'json' };
```

[due to the "import assertions" no longer being supported](https://stackoverflow.com/a/78876692). Fix is to use `with` instead. This will break in older Node versions like 16 || 17, but that doesn't matter as we specify Carbon needs >=18. 

#### Changelog

**Changed**

- swap `assert` keyword with `with`

#### Testing / Reviewing

See that the ci-checks, especially the build, pass


Gururajj77 (approved):
Works with node 23 🚀

2nikhiltom (approved):

ariellalgilmore (approved):

PR #18250, opened 2024-12-12
Storybook Playground and Stackblitz

### LINKED ISSUE #16816: Revise "Playground" story approach in the `@carbon/react` storybook
## Context

> As a developer using `@carbon/react`,
> I need a way to copy and paste valid/working component code snippets from storybook,
> So that I can get a jumpstart using a component in my project, experiment in a sandbox/stackblitz, etc.
> 

The `@carbon/react` storybook has supported this use case so far via the [storysource plugin](https://storybook.js.org/addons/@storybook/addon-storysource/). It displays the code source of the current story being shown in the canvas. It meets this need fairly well if the story uses explicit values within the story.

<details><summary>Click to expand example</summary>

CleanShot 2024-06-18 at 13 42 30@2x

</details> 

Although when controls/args are used, the source code doesn't work well for copy/paste because it contains storybook-specific code (`...args`) that would cause an error in any environment outside of our storybook:

<details><summary>Click to expand example</summary>

CleanShot 2024-06-18 at 13 38 46@2x

</details> 

## As-is experience

To combat this we have so far tried to follow a strategy where we only use controls/args on a single specific story named "Playground". Each component typically has one - controls are fully enabled, and other stories have explicit prop usage instead of controls. This way the story source is copy/paste-able for every story except the Playground story.

In some cases Playground stories can't be used because the component has variants that require different stories and aren't able to be contained in one Playground story. Button and StructuredList are an example of this currently. In these cases controls are enabled on every story and there is no Playground story.

This strategy and it's inconsistency is confusing for both maintainers and users of the storybook.

## Proposal

Stories are [intended to be authored using args](https://storybook.js.org/docs/writing-stories#how-to-write-stories). Every story should be updated to use args where possible. We could then delete the playground stories. 

Another way to think about this is that we're renaming `Playground` stories to be `Default`. The goal isn't to encourage adding a new story for every variant of a component - instead this change encourages further reduction of stories towards greater reliance on args. 

There are a few ways we can continue supporting the copy/paste use case:
* Enhance our docs pages to include [Source](https://storybook.js.org/docs/api/doc-block-source) or [Canvas](https://storybook.js.org/docs/api/doc-block-canvas) blocks, which appear to support dynamic code snippet generation when using args via the [`type` option](https://storybook.js.org/docs/api/doc-block-source#type):
  * > dynamic: Renders the story source with dynamically updated arg values
* Provide an "Open in Stackblitz" link on every story via [the stackblitz addon](https://storybook.js.org/addons/@stackblitz/storybook-addon-stackblitz), where the current story source can be opened in a sandbox for experimentation

Additional benefits of this approach include:
* Having one less story in every component folder
* A simplified story authoring experience that would match the storybook documentation
* Every story would support args, so we could use args for configuration in AVT and VRT tests, which would simplify the "arrange" portion of these tests

guidari:
Closes #16816

Storybook does not offer a good integration with third party packages like `@stackblitz/sdk`. There is no clean way to grab code from stories, that's why the `codePreviewer.tsx` file has a lot of regex.

We are gonna add the "Open in Stackblitz" option in this [issue](https://github.com/carbon-design-system/carbon/issues/18257).

#### Changelog

- Added the `@stackblitz/sdk` package. With this package now we are able to add an "Open in Stackblitz" for components.

**Changed**

- Added `args` to stories that make sense to use it.
- Changed avt tests to match the new story name.

**Removed**

- Removed the `Playground` stories from 99% os components. 

#### Testing / Reviewing

- Verify if the stories are working as expected.
- Percy will throw a few "errors" as expected. We can use that to review those specific components.


tay1orjones (approved):
LGTM!

annawen1 (approved):

alisonjoseph (approved):

PR #18283, opened 2024-12-27
fix: added href to iconButton

### LINKED ISSUE #17955: [Bug]: `IconButton` types do not include `href`
### Package

@carbon/react

### Browser

_No response_

### Package version

1.67.1

### React version

_No response_

### Description

The types for `IconButton` (`IconButtonProps`) do not include `href`. However, the types for `Button` (`ButtonProps` which use `ButtonBaseProps`) do include an `href`.

### Reproduction/example

https://stackblitz.com/edit/github-azd4us?file=src%2FApp.tsx

### Steps to reproduce

Paste the following code in an IDE with support for TypeScript (e.g., VS Code):

```tsx
import { Button, IconButton } from '@carbon/react';
import { Information } from '@carbon/react/icons';

const Temp = () => (
  <>
    <Button
      id="help-button"
      hasIconOnly
      renderIcon={Information}
      iconDescription="Help"
      href="/help"
      target="_blank"
      rel="noreferrer"
    />
    <IconButton
      id="help-icon-button"
      label="Help"
      href="/help"
      target="_blank"
      rel="noreferrer"
    >
      <Information />
    </IconButton>
  </>
);
```

See the following error:

```txt
Type '{ children: Element; id: string; label: string; href: string; target: string; rel: string; }' is not assignable to type 'IntrinsicAttributes & IconButtonProps & RefAttributes<unknown>'.
  Property 'href' does not exist on type 'IntrinsicAttributes & IconButtonProps & RefAttributes<unknown>'. Did you mean 'ref'?
```

`Button` accepts `href`. `IconButton` does not. They both should.

### Suggested Severity

### Application/PAL

DataPower

### Code of Conduct

riddhybansal:
Closes #17955 

Added href to iconButton

#### Changelog

**New**

-Added href to iconButton

#### Testing / Reviewing

Try using IconButton in .tsx file it should work as intended and should not show any typescript error.


guidari (approved):
LGTM! 🚀 You just have to run `yarn test -u` to update the snapshots.

packages/react/src/components/IconButton/index.tsx
```diff
@@ -77,6 +77,11 @@ interface IconButtonProps
    */
   autoAlign?: boolean;
 
+  /**
+   * Optionally specify an href for your Button to become an `<a>` element
```
adamalston: Should these JSDoc comments say 'IconButton' instead of 'Button'?

```suggestion
   * Optionally specify an href for the IconButton to become an `<a>` element
```

packages/react/src/components/IconButton/index.tsx
```diff
@@ -253,6 +258,11 @@ IconButton.propTypes = {
    */
   autoAlign: PropTypes.bool,
 
+  /**
+   * Optionally specify an href for your Button to become an `<a>` element
```
adamalston: https://github.com/carbon-design-system/carbon/pull/18283#discussion_r1900499521

adamalston (approved):

Gururajj77 (approved):

packages/react/src/components/IconButton/index.tsx
tay1orjones:
```suggestion
   * Optionally specify an href for your IconButton to become an `<a>` element
```

tay1orjones (approved):

adamalston (commented):
One thing that's not clear to me, and maybe I should've mentioned it in my review, is how adding this `href` attribute is going to fix the type issues for `rel` and `target` attributes. If you delete the `href` attribute in the reproduction, the type error just moves down to the next untyped prop which is `target`. With that being said, are there more changes that should be made here?

PR #18295, opened 2025-01-03
fix: Updates the interface for DataTableHeader to include isSortable

### LINKED ISSUE #18172: [Bug]: `DataTableHeader` doesn't have an `isSortable` property
### Package

@carbon/react

### Browser

_No response_

### Package version

1.67.1

### React version

_No response_

### Description

`DataTableHeader` doesn't have an `isSortable` property even though `getHeaderProps` accepts one. Ideally, the problematic code in the reproduction should be able to simplified like the following:

```diff
-  {...getHeaderProps({
-    header,
-    isSortable: header.isSortable,
-  })}
+  {...getHeaderProps({ header })}
```

I'm not sure whether this issue should be a bug for a missing property or a feature request to add it.

My workaround at the moment is to just ignore the type error. Everything seems to work correctly.

### Reproduction/example

https://stackblitz.com/edit/github-anyd1u?file=src%2FApp.tsx

### Steps to reproduce

See type error in reproduction.

### Suggested Severity

### Application/PAL

DataPower

### Code of Conduct

cknabe:
Closes #18172

Minor update to the DataTableHeader interface to handle a header having its own isSortable, instead of needing to be redeclared when passing into getHeaderProps.

I picked a random bug/feature to refresh my knowledge of contributing to Carbon. If this fix isn't wanted/needed, no problem.

#### Changelog

- DataTableHeader now has isSortable in its interface
- getHeaderProps now checks for isSortable in the header first ~ there may be backwards incompatibilities or side effects

#### Testing / Reviewing
- I added a minor test that verifies a column does not allow sorting. Adding testing updates for this change seems like overkill, or possibly unwarranted. We can remove the testing updates if desired.


packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -474,7 +476,7 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
       ...rest,
       key: header.key,
       sortDirection,
-      isSortable,
+      isSortable: header.isSortable || isSortable, // header check first, since every header defaults to the table isSortable prop
```
adamalston, line 479: I think the nullish coalescing operator you had prior to https://github.com/carbon-design-system/carbon/commit/f34935f7fb9a2930a273eecf39b4ad90f4a66b44 was correct. Otherwise, if someone sets `header.isSortable` to `false` and `isSortable` is `true`, the expression `false || true` will evaluate to `true`, which I don't think is optimal. `header.isSortable` should take precedence over `isSortable`.

Also, should this scenario be tested?

tay1orjones:
Hi there! Thanks for the effort on this so far 🙏 Since it’s been inactive for a bit, I’m going to close this PR for now. If you’d like to pick it back up, awesome—just reopen and continue!

PR #18301, opened 2025-01-07
feat: enabling type imports directly from `@carbon/react`

### LINKED ISSUE #16436: [Feature Request]: As a Developer, I would like to import types from `@carbon/react`
### The problem

Users are often using reusable components to render Tags with different colors (type) based on some status.
To achieve this, a TagType enum needs to be defined because we can not import it from carbon/react.

For instance:

```
enum TagType {
        red = 'red'
        green....
        ... 
}

interface CustomInterface {
  type: TagType,
  title: string,
}
```

### The solution

Could be nice to define and export a `TagType` containing all supported types and then import it so that we can use it in our interfaces, couple of examples below.

### Examples

Example of an interface for Props, with tagType
```
import {TagType} from '@carbon/react';

interface CustomInterface {
  tagType: TagType,
  title: string,
}

`<MyCustomCardWithTag tagType={TagType.red} title="Example" />`

```

### Application/PAL

_No response_

### Business priority

### Available extra resources

I'm happy to contribute if this change can be useful.

### Code of Conduct

Gururajj77:
Closes #16436 

enabling component's type imports directly from library's root.

#### Changelog

**Changed**

- `index.ts` - imported component's prop type interfaces to root index.ts

#### Testing / Reviewing
 can be tested only after merge, issues can be raised accordingly


tay1orjones (commented):
Looking good so far!

tay1orjones (approved):

ariellalgilmore (approved):

riddhybansal (approved):
This LGTM 💯

PR #18304, opened 2025-01-07
feat: codemod to enable v12 overflowmenu

### LINKED ISSUE #18121: Research/Write codemod for `enable-v12-overflowmenu `

riddhybansal:
Closes #18121 

Add codemod for enable-v12-overflowmenu feature flag migration

#### Changelog

**New**
- Add codemod to automatically wrap OverflowMenu components with FeatureFlags
- Add support for both JavaScript and TypeScript files
- Add comprehensive test cases to verify transform behavior

**Changed**
- Transform modifies imports to include FeatureFlags from @carbon/react when needed
- Preserves existing FeatureFlags wrapping to avoid duplicate transformations

**Removed**
- N/A

#### Testing / Reviewing

1. Manual Testing:
   ```bash
 
   npx @carbon/upgrade migrate enable-v12-overflowmenu --write

2. Test Cases to Verify:

Basic OverflowMenu components should be wrapped with FeatureFlags
FeatureFlags import should be added if not present
Already wrapped OverflowMenu components should remain unchanged
TypeScript files should be properly handled
Props and children should be preserved
Component structure and formatting should be maintained

3. Files to Review:

transforms/enable-v12-overflowmenu.js - Main transform logic
transforms/__tests__/enable-v12-overflowmenu-test.js - Test setup
transforms/__testfixtures__/enable-v12-overflowmenu.input.js - JS test input
transforms/__testfixtures__/enable-v12-overflowmenu.output.js - JS test output
transforms/__testfixtures__/enable-v12-overflowmenu.input.tsx - ts test input
transforms/__testfixtures__/enable-v12-overflowmenu.output.tsx - ts test output


tay1orjones (changes requested):
Take another look at the [existing API usage](https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--overview) in storybook and compare it with the [new API usage](https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu-feature-flag--flag-details) behind the flag. This should be the rubric for transformation - you'll see there are a few things that need to change here including:

- `OverflowMenu.Item` doesn't exist
- Input files should use `OverflowMenuItem`
- Output files should always have `OverflowMenuItem` migrated to `MenuItem`
- Good call on covering both spread and explicit props! 👍  Double check that all the props available on `OverflowMenuItem` are migrated properly to the props available on `MenuItem`.

I also think it would be a good idea to add a new api usage into the input files to ensure that those aren't migrated. Because it's possible someone runs the codemod against a file that uses both the old and the new. The old should be migrated without touching the new.

tay1orjones:
Something else I just thought of - a consumer could use a single FeatureFlags at the root of their app and turn this flag on for their entire application. They wouldn't need anything wrapped in the `FeatureFlag` component. What are some ways we could handle this?

* Don't include the wrapping of FeatureFlag at all, and just migrate the OverflowMenu api usages
* Conditionally wrap new api use in FeatureFlag based on a command line parameter
* Have one codemod with FeatureFlag wrapping, and another one without
* ... 

Curious what you think, and if maybe there's another approach I haven't thought of?

riddhybansal:
> Something else I just thought of - a consumer could use a single FeatureFlags at the root of their app and turn this flag on for their entire application. They wouldn't need anything wrapped in the `FeatureFlag` component. What are some ways we could handle this?
> 
> * Don't include the wrapping of FeatureFlag at all, and just migrate the OverflowMenu api usages
> * Conditionally wrap new api use in FeatureFlag based on a command line parameter
> * Have one codemod with FeatureFlag wrapping, and another one without
> * ...
> 
> Curious what you think, and if maybe there's another approach I haven't thought of?

Possible Approaches:

1. **Separate Codemods**:
```
transforms/
  ├── enable-v12-overflowmenu-with-flags.js    // With FeatureFlags wrapping
  └── enable-v12-overflowmenu-api-only.js      // Only API migration
```
Pros:
- Clear separation of concerns
- Users can choose exactly what they need

Cons:
- Maintenance of two similar codemods
- Users need to run two commands if they want to mix approaches

2. **CLI Parameter**:
```bash
# With FeatureFlags wrapping
yarn jscodeshift -t transforms/enable-v12-overflowmenu.js --wrap=true 'src/**/*.{js,tsx}'

# API migration only
yarn jscodeshift -t transforms/enable-v12-overflowmenu.js --wrap=false 'src/**/*.{js,tsx}'
```
Pros:
- Single codemod with flexible usage
- Users can mix approaches in different parts of their codebase

Cons:
- More complex codemod logic
- Need to document CLI options

Recommended Solution:

I recommend going with the CLI Parameter approach because:
1. It keeps everything in one codemod
2. Provides flexibility without over-engineering
3. Follows existing jscodeshift patterns
4. Easy to understand and use

This solution:
1. Makes FeatureFlags wrapping optional via CLI parameter
2. Maintains a single codemod
3. Easy to understand and use
4. Allows mixing approaches in different parts of codebase
5. Well-documented in both the codemod and config
6. Follows existing Carbon upgrade CLI patterns

tay1orjones:
Nice! Sounds good to me. I don't know if we've used cli params like that before, might require some research how to get them passed through to the script and condition the logic off of it.

packages/upgrade/src/cli.js
```diff
@@ -75,13 +75,14 @@ export async function main({ argv, cwd }) {
       );
     },
     run(async (args) => {
-      const { verbose, migration, write, paths } = args;
+      const { verbose, migration, write, paths, wrap } = args;
```
tay1orjones: Could we make this more specific? `wrapWithFeatureFlag` or `wrapWithFlag` maybe

packages/upgrade/src/upgrades.js
```diff
@@ -331,6 +331,49 @@ export const upgrades = [
...
+                    '**/storybook-static/**',
+                  ],
+                });
+
+          await run({
+            dry: !options.write,
+            transform,
+            paths,
+            verbose: options.verbose,
+            wrap: options.wrap, // Pass wrap in parserConfig
```
tay1orjones:
```suggestion
            wrap: options.wrap,
```

packages/upgrade/transforms/__tests__/enable-v12-overflowmenu-test.js
```diff
@@ -0,0 +1,18 @@
+// __tests__/enable-v12-overflowmenu-test.js
```
tay1orjones: To match the others in this folder
```suggestion
/**
 * Copyright IBM Corp. 2025
 *
 * This source code is licensed under the Apache-2.0 license found in the
 * LICENSE file in the root directory of this source tree.
 */

'use strict';

```

packages/upgrade/transforms/enable-v12-overflowmenu.js
```diff
@@ -0,0 +1,220 @@
+/**
+ * Copyright IBM Corp. 2024
```
tay1orjones:
```suggestion
 * Copyright IBM Corp. 2025
```

tay1orjones (changes requested):
Looking good! Just a couple minor things

tay1orjones:
The tests are failing because jscodeshift has a minor bug where it puts parens `(` `)` around things even though the output technically doesn't need it. You can just update the output files to match this and have the parens/formatting. You can grab it from the logs of the test failures.

tay1orjones:
Could you share the steps and command you're using to test locally?

riddhybansal:
Hey there , To test this codemod I have created test files that can be deleted later when the PR is reviewed . There are two files in tranform test.tsx and test.js. 

You can test the codemod by following these steps 
1). Yarn build
2). Run this command 

```
npx jscodeshift -t packages/upgrade/transforms/enable-v12-overflowmenu.js 'packages/upgrade/transforms/test/test.{js,tsx}' --wrapWithFeatureFlag=false 
```

There will be changes to test.tsx and test.js OverflowMenu will be changed to new overflowmenu without wrapping it with feature flag since we have passed --wrapWithFeatureFlag=false.

3). Run this command 

```
npx jscodeshift -t packages/upgrade/transforms/enable-v12-overflowmenu.js 'packages/upgrade/transforms/test/test.{js,tsx}' --wrapWithFeatureFlag=true
```

There will be changes to test.tsx and test.js OverflowMenu will be changed to new overflowmenu with wrapping it with feature flag since we have passed --wrapWithFeatureFlag=true.

tay1orjones (approved):

heloiselui (approved):
LGTM! 👍🏼

PR #18322, opened 2025-01-10
fix(utilities/date-time-format): prevent "0 years" display

janhassel:
Since we define months as four-week periods instead of calendar months, the current implementation of relative date and time format can result in "0 years ago" displays. This code fixes it [according to the guidelines](https://pages.github.ibm.com/carbon/ibm-products/guidelines/content/date-and-time/#units-of-time-and-when-to-use-each-type).

#### Changelog

**Changed**

- Use "months" preset if 364 days or less.
- Updated test to catch this

#### Testing / Reviewing

The updated automated test would fail with the previous implementation as it would result in "0 years ago" instead of "12 months ago"


guidari (approved):

riddhybansal (approved):
This LGTM 💯

kennylam (approved):

PR #18323, opened 2025-01-10
fix(tablist): make aria-label optional

tay1orjones:
It looks like this was a minor oversight during the initial refactor of the Tabs API. After discussing with IBM Accessibility team it was clear this should optional, not required. Plenty of context is provided through the existing labels as each `Tab` has an accessible name (even when IconTab is used). Also the TabList doesn't receive focus or have a tab stop. In some limited cases it might make sense to give the user some additional context via the TabList aria-label but it shouldn't be required.

#### Changelog

**Changed**

- modified aria-label prop to be optional
- update associated stories

#### Testing / Reviewing

- Prop tables for TabList on the storybook should show it as optional and a TabList should render in the stories without a typescript/proptypes warning.


kennylam (approved):
LGTM!

2nikhiltom (approved):

PR #18324, opened 2025-01-11
Control ArrowUp for Dropdown

### LINKED ISSUE #18277: [Bug]: Dropdown shouldn't open with Up arrow key press
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.72.0 

### React version

v18.2.0

### Description

The closed default & fluid dropdown input should not open when the user presses the `Up` arrow key, as unlike the combobox, there is no ability to wrap around the options list (i.e., pressing the Up arrow at the top of the open list does nothing). 
For more details, refer to [this comment](https://github.com/carbon-design-system/carbon-website/pull/4362#issuecomment-2489282181).

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-dropdown--playground

### Steps to reproduce

-

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

cknabe:
Closes #18277

- Only allow ArrowUp if the Dropdown is already open
- Allow all other key presses regardless if open or closed

Testing
1. Open up the Storybook for this PR
2. On the left under Components, expand the Dropdown component entry
3. Select the Default Dropdown from the expanded Dropdown component options in the left nav
4. Tab to the Dropdown component that appears in the content section on the right
5. Press the Up arrow key on your keyboard
6. The Dropdown should not open at all
7. After confirming that the Dropdown did not open, press the Down arrow key on your keyboard
8. The Dropdown component should open and you should see the following:
image
9. Press the Up arrow key on your keyboard and you should see the `Lorem, ipsum dolor sit amet consectetur adipisicing elit.` option become the highlighted option.
image
10. Press the Enter key on your keyboard, and the Dropdown should close and you should see the following:
image


annawen1:
Hi @cknabe, trying to following the ticket linked in the corresponding bug issue - are the rules listed in your PR description coming from another issue?

```
Disallows ArrowUp if the first item is selected.
Disallows ArrowDown if the last item is selected.
Allows all key-presses when the Dropdown is open
Allows all key-presses that are not ArrowUp or ArrowDown
```

cknabe:
@annawen1 No, https://github.com/carbon-design-system/carbon/issues/18277 states this:

> The closed default & fluid dropdown input should not open when the user presses the Up arrow key, as unlike the combobox, there is no ability to wrap around the options list (i.e., pressing the Up arrow at the top of the open list does nothing).

This PR causes the Dropdown to not open when the user is at the top selection and they press the up arrow key.
When the user is at the bottom selection and presses the down arrow key, it also does not open.

annawen1:
> @annawen1 No, #18277 states this:
> 
> > The closed default & fluid dropdown input should not open when the user presses the Up arrow key, as unlike the combobox, there is no ability to wrap around the options list (i.e., pressing the Up arrow at the top of the open list does nothing).
> 
> This PR causes the Dropdown to not open when the user is at the top selection and they press the up arrow key. When the user is at the bottom selection and presses the down arrow key, it also does not open.

I see, I had understood the issue to mean that the Dropdown should not open at all on `arrowUp` regardless of which item is selected... @Kritvi-bhatia17 could you clarify?

cknabe:
I'll wait for @Kritvi-bhatia17 to comment before implementing the tests. I did make an assumption that the only reason the arrow up should be disabled is if it is already the top selected item, since Dropdown does not wrap. In the same way, I assumed the arrow down should not open if at the last selected item, because Dropdown does not wrap. I concluded that if a middle item is selected, then arrow up/arrow down should work, since you can move up or down in the list from that position.

tay1orjones:
@mbgower could you re-clarify here the changes you were wanting to see wrt arrowup/down keypresses on Dropdown?

Kritvi-bhatia17:
Hi team! 
Thanks for opening that, @cknabe. 
Based on my understanding of what you've mentioned, @annawen1, it seems correct that the dropdown should not open on arrowUp, regardless of which item is selected. However, I’d still request @mbgower to clarify further, especially regarding the latter part of the question.

> I see, I had understood the issue to mean that the Dropdown should not open at all on arrowUp regardless of which item is selected... @Kritvi-bhatia17 could you clarify? And on the disabling of opening Dropdown on arrowDown when last item is selected?

mbgower:
> @mbgower could you re-clarify here the changes you were wanting to see wrt arrowup/down keypresses on Dropdown?

@tay1orjones I think this PR wording is making things a bit more complicated than need be. I merely stated that the **default** Dropdown component should not open on `up arrow` since the arrow key behaviour when the component is open is that arrow keys do not wrap from bottom to top (or vice versa).

There is NO need to alter any behaviour for `down arrow`. It is working as expected.

I also do not think the list of 7 test steps is clear or correct.

---

Since there seems to be some confusion in this PR, I will say for added context that the only one of the variants of Dropdown that wraps with arrow keys is Combobox -- and it also opens on `up arrow` to match that behaviour. Given combobox can have a LOT of options, that seems reasonable to me (and it's supported by the [APG](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) which lists it as optional and uses that behaviour in its demos). We have that documented as such on the a11y tab. So everything for combobox seems to be as expected. It should NOT be changed by this PR.

Further, everything for multi-select seems correct already. It doesn't open on `up arrow`. The default dropdown behaviour should match this variant.

guidari (approved):
Based on the conversations this LGTM 🚀
It just have to fix a small conflict.
Thanks for working on this!

annawen1 (approved):

PR #18342, opened 2025-01-14
feat(badge-indicators): new component

### LINKED ISSUE #17538: [Badge indicator] React|WC Implementation
----

### Storybook implementation

Dev consideration for Badge

---

### Badge indicator spec 
FINAL

ariellalgilmore:
Closes #17538

worked on with @2nikhiltom !!!!

Implement badge indicator

#### Changelog

**New**

- new internal badge indicator component in react/wc

**Changed**

- react
   - added badgeCount prop to iconButton (which trickles down to button)
- web component
  - added badge-indicator slot to button (which trickles down to icon-button)

#### Testing / Reviewing

- check icon button and button storybook for react and web component
- shouldn't display any badge when disabled
- implementations in other components will be in a different PR

Note: seems like some styles are off in button for web-components, but are unrelated to the badge-indicator and should be fixed in the parity work


tay1orjones:
Today we decided to pivot on the following:

Based on the [guidance from the open PR](https://carbondesignsystem-git-fork-thyhmdo-de4bbb-carbon-design-system.vercel.app/patterns/status-indicator-pattern/#badge-indicator), the badge indicator is intended to only be used on a Button/IconButton. That supports:
1. Not exporting BadgeIndicator
2. Baking BadgeIndicator into Button/IconButton directly, exposing control through a `badgeCount` (or similar) number prop
3. Removing experimental/BadgeIndicator stories and instead just have a "with badge indicator" story for both Button and IconButton. 
   * This would essentially be the same story you already have, just splitting it into two different locations (Button using hasIconOnly and IconButton)
   * If/when we determine additional uses for BadgeIndicator outside of Button/IconButton we can explore exporting it and if it should have it's own stories. 

It feels to me like this should live under `packages/react/src/components` because it renders elements to the DOM and has styles associated with it. The components currently in `packages/react/src/internal` don't render any UI and don't have any styles associated with them.

tay1orjones:
Another requirement we talked about - the badge indicator should only show on icon-only buttons that have `size="lg"` and `kind="ghost"`. This is our understanding of the intent from design @thyhmdo. 

Rather than doing a bunch of conditional types, we'll just mention it in the prop description for `badgeCount`.

```js
  /**
   * Display a badge on the button. An empty/dot badge if 0, a numbered badge if >0.
   * Must be used with size=lg and kind=ghost
   */
  badgeCount?: number;
```

We can also provide a dev-only console warning if `badgeCount` is configured and `size` or `kind` are wrong.

> The IconButton badgeCount prop must be used with size="lg" and kind="ghost". This requirement is documented in the design specification: https://carbondesignsystem.com/patterns/status-indicator-pattern/#badge-indicator

packages/web-components/src/components/button/button.stories.ts
```diff
@@ -301,6 +308,18 @@ export const IconButton = {
     </cds-button>`,
 };
 
+export const iconButtonWithBadge = {
+  argTypes: controls,
+  args: defaultArgs,
+  render: ({ badgeCount }) =>
```
annawen1: think we need to add the `disabled` arg here since I'm still seeing the badge count when I have the button disabled in storybook

packages/web-components/src/components/button/button.stories.ts
```diff
@@ -65,13 +66,19 @@ const defaultArgs = {
...
   tooltipPosition: BUTTON_TOOLTIP_POSITION.TOP,
+  badgeCount: 4,
 };
 
 const controls = {
   buttonClassName: {
     control: 'text',
     description: 'Specify an optional className to be added to your Button',
   },
+  badgeCount: {
```
annawen1: not seeing the `badgeCount` control appear in the `Icon Button` story in React, only appears in `Icon Button with Badge` - should we align with that too for WC storybook?

thyhmdo:
In Web Component, the icon color inside the button is incorrect. Should be `$icon-primary`
image

packages/web-components/src/components/button/button.stories.ts
ariellalgilmore: thanks updated!

packages/web-components/src/components/button/button.stories.ts
ariellalgilmore: updated!

thyhmdo (approved):
lgtm :)

packages/react/src/index.ts
```diff
@@ -189,6 +189,7 @@ export * from './components/Tooltip/DefinitionTooltip';
 export * from './components/Theme';
 export * from './internal/usePrefix';
 export { useIdPrefix } from './internal/useIdPrefix';
+export { IconIndicator as unstable__IconIndicator } from './components/IconIndicator';
```
ariellalgilmore:
```suggestion
```

annawen1 (approved):
LGTM

Issue #18164, opened 2024-11-26
[Bug]: Select component ignores "selected" property on page load


thekevinhunt:
### Package

@carbon/web-components

### Browser

Chrome

### Package version

v2.18.0

### Description

When using the cds-select component and including "selected" on a cds-select-item element, the option is not selected on page load.  Interestingly, if you modify the DOM in the developer toolbar to add "selected" AFTER the page initially loads, it properly selects the element.

Interestingly, the "disabled" property doesn't have the same problem on cds-select-item elements.

### Reproduction/example

https://stackblitz.com/github/carbon-design-system/carbon/tree/main/packages/web-components/examples/components/select?file=index.html

### Steps to reproduce

1. Use the existing stackblitz build provided in the docs (https://web-components.carbondesignsystem.com/?path=/docs/components-select--overview)
2. Simply add "selected" as a property to any cds-select-item in the index.html file.
3. The page will automatically reload, but the option won't be selected
4. Edit the live HTML directly in the web developer tools and the option will get selected fine

### Suggested Severity

Severity 2 = User cannot complete task, and/or no workaround within the user experience of a given component.

### Application/PAL

IBM Industry Compliance PIMS Portal

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

ariellalgilmore:
Hi @thekevinhunt did you try setting the `value` on `cds-select` like the following:

```
    <cds-select
      helper-text="Optional helper text"
      label-text="Select"
      placeholder="Optional placeholder text"
      value="cloudFoundry"
    >
      <cds-select-item-group label="Category 1">
        <cds-select-item value="all">Option 1</cds-select-item>
        <cds-select-item value="cloudFoundry">Option 2</cds-select-item>
      </cds-select-item-group>
      <cds-select-item-group label="Category 2">
        <cds-select-item value="staging">Option 3</cds-select-item>
        <cds-select-item value="dea">Option 4</cds-select-item>
        <cds-select-item value="router">Option 5</cds-select-item>
      </cds-select-item-group>
    </cds-select>
```

On page load i see the select now starting with "Option 2"

thekevinhunt:
[...]

PR #18367, opened 2025-01-16
fix(select): resolve parity issues

### LINKED ISSUE #17931: React|WC Parity: Select
See parent issue #17992 for acceptance criteria.

- [ ] [#18164](https://github.com/carbon-design-system/carbon/issues/18164) clean up docs so that it's clear how `value` and `selected` work

> [!IMPORTANT]  

### LINKED ISSUE #17932: React|WC Parity: Skeleton
See parent issue #17992 for acceptance criteria.

> [!IMPORTANT]  

makafsal:
Closes #17931 #18164 

Resolved a few parity issues

#### Changelog

**Changed**

- Resolved the issue #18164 
- Changed AI gradient to match the React version
- Changed AI button focus to match the React version

#### Testing / Reviewing

> Verify issue #18164 
Please see the acceptance criteria table [here](https://github.com/carbon-design-system/carbon/issues/18164#issuecomment-2580147879)
1. Open `select` playground storybook
2. Inspect the `select` element using the browser dev tools and select any `option` by putting the `selected` attribute
3. Also, we can utilize the `value` prop from controls to give the `value` prop to the component
This way we can verify all cases mentioned in the table.

> Verify AI Label gradient
1. Open the `with AI label` sub-story from the `select` storybook
2. Then compare the gradient style with the corresponding React storybook

> Verify AI button focus style
1. Open the `with AI label` sub-story from the `select` storybook
2. Then press the `tab` button and compare the focus border thickness with the corresponding React storybook

<b>Note</b> Also verify this style from the `AI Label` component storybook


laurenmrice (changes requested):
Hey @makafsal , this is looking good! Just had a couple of comments:

### Default story
- [x] There should be no helper text shown in the field in WC to match React.
- [x] When the select menu is open, can we match the menu options to be exactly the same as the React one? It would be good to try to keep them the same.

Screenshot 2025-01-27 at 4 25 33 PM

- ~I do not see the `decorator`, `noLabel`, or `slug` props in WC but they are present in React. Just clarifying if that is intentional?~
  - Intentional.

- ~I see `onInput`, `placeholder`, and `value` props in Web components but not in React. Also clarifying if that is intentional that they do not match between the two?~
  - Intentional.

- ~When the prop `inline` is set to true and the prop `disabled` is set to true, there should be no background on the field, it should be transparent.~
  - Design make a separate issue for this because its happening in React as well. 

- [x] There is a bug when the prop `inline` is set to true and the prop `warn` is set to true. The warning icon should be 8px to the left of the chevron icon. It should match the spacing that in the invalid state. I do want to note that this scenario in React is currently really broken, which I am making a separate issue for. But I feel that we can address the WC spacing problem here in this PR.

### Inline story
- [x] There should be no helper text shown in the field in WC to match React.
- [x] When the select menu is open, can we match the menu options to be exactly the same as the React one? It would be good to try to keep them the same.
- ~When the prop `disabled` is set to true, there should be no background on the field, it should be transparent.~
  - Design make a separate issue for this because its happening in React as well.

- [x] There is a slight bug when the prop `warn` is set to true. The warning icon should be 8px to the left of the chevron icon. It should match the spacing that in the invalid state.

### With AI Label
- [x] There should be no helper text shown in the field in WC to match React.
- [x] The input field is missing the vertical divider between the AI slug and the chevron icon. You can view the design specs [here on our website](https://carbondesignsystem.com/components/select/style/#ai-presence).
- ~I do not see the `decorator`, `noLabel`, or `slug` props in WC but they are present in React. Just clarifying if that is intentional?~
  - Intentional.

- ~I do see the `onInput`, `placeholder`, and `value` props in WC but not React. Also is this intentional?~ 
  - Intentional.

- [x] The WC AI explainability popover design isn’t matching the React one. I have documented the design specs in this image below.
Frame 4

-----

*Note: I will be making a separate bug issue to tackle some unrelated visual bugs that were already in React.

makafsal:
@laurenmrice 
> 1. I do not see the decorator, noLabel, or slug props in WC but they are present in React. Just clarifying if that is intentional?

Yes, it is an experimental prop, so it can be excluded.

> 2. I see onInput, placeholder, and value props in Web components but not in React. Also clarifying if that is intentional that they do not match between the two?

I assume those are new props in web components.

> 3. When the prop inline is set to true and the prop disabled is set to true, there should be no background on the field, it should be transparent.

This is the exact behavior in React as well. Because both WC and React components shares the same style. So could you please raise a separate issue for this because this looks exact with React and it is out of the scope of parity.

#### NOTE: All other issues are resolved

laurenmrice (changes requested):
Thanks @makafsal for making some of those updates. I will add that one point you mentioned above into a separate bug issue. 👍

I did notice a couple thing still haven't been addressed and wanted to clarify:

**Open select menu**
Can we further customize what items show up in the open select menu so it matches Reacts open select menu? I am hesitant to show "Categories" in the menu, because we currently don't support or recommend that in our website guidance and I don't want to potentially confuse our users. I know that the menu is browser based, but I thought we could customize it more? This is the difference right now between the two:

Frame 6

-----

**Warning icon padding bug**
I still see the warning icon padding bug when the `warn` prop is set to true in the Default and Inline stories. It should follow the spacing that happens when the `invalid` prop is set to true currently. Here are some specs below:

select-style-structure-default

makafsal:
> Thanks @makafsal for making some of those updates. I will add that one point you mentioned above into a separate bug issue. 👍
> 
> I did notice a couple thing still haven't been addressed and wanted to clarify:
> 
> **Open select menu** Can we further customize what items show up in the open select menu so it matches Reacts open select menu? I am hesitant to show "Categories" in the menu, because we currently don't support or recommend that in our website guidance and I don't want to potentially confuse our users. I know that the menu is browser based, but I thought we could customize it more? This is the difference right now between the two:
> 
> Frame 6
> **Warning icon padding bug** I still see the warning icon padding bug when the `warn` prop is set to true in the Default and Inline stories. It should follow the spacing that happens when the `invalid` prop is set to true currently. Here are some specs below:
> 
> select-style-structure-default

Done.

laurenmrice (approved):

makafsal:
@laurenmrice Decided to fix those issues (in WC) you've mentioned in this [issue](https://github.com/carbon-design-system/carbon/issues/18558) to address in this PR. So I've made those changes in this PR. Can you please verify again?

packages/styles/scss/components/form/_form.scss
```diff
@@ -163,6 +163,20 @@ $input-label-weight: 400 !default;
...
+    .#{$prefix}--select-input--inline__wrapper,
+  .#{$prefix}--select-input--inline__wrapper[data-invalid] {
+    ~ .#{$prefix}--form-requirement {
+      display: inline-flex;
+      overflow: visible;
+      margin: 0;
+      inline-size: 100%;
+      margin-block-end: 0;
+      max-block-size: 100%;
+      padding-inline-start: 0.5rem;
```
tay1orjones:
```suggestion
      padding-inline-start: $spacing-03;
```

tay1orjones (approved):
One minor correction, otherwise LGTM!

laurenmrice (changes requested):
Hey @makafsal, most of the changes look good now, thankyou! There are just three things below:

**Default story (inline set to true) and Inline story**
- In general, there needs to be 8px on the left of the text in the input for all states. Right now it’s 16px. It is only set at 8px for the inline style. The right side looks good and is currently at 8px.

Frame 1

- Warning state: There is some extra padding in the Inline warning state between the chevron and warning icon. There should be 8px between the chevron and the warning icon.
- Error state: The error border around the input is not appearing.

heloiselui (approved):

makafsal:
> Hey @makafsal, most of the changes look good now, thankyou! There are just three things below:
> 
> **Default story (inline set to true) and Inline story**
> 
> * In general, there needs to be 8px on the left of the text in the input for all states. Right now it’s 16px. It is only set at 8px for the inline style. The right side looks good and is currently at 8px.
> 
> Frame 1
> * Warning state: There is some extra padding in the Inline warning state between the chevron and warning icon. There should be 8px between the chevron and the warning icon.
> * Error state: The error border around the input is not appearing.

Done.

laurenmrice (approved):
Looks good! 🎉

packages/web-components/src/components/select/select.ts
```diff
@@ -471,20 +501,21 @@ class CDSSelect extends FormMixin(LitElement) {
             </label>`
```
riddhybansal:
```
<div class="${selectClasses}">
        ${!hideLabel
          ? html`<label class="${labelClasses}" for="input">
              <slot name="label-text"> ${labelText} </slot>
```
The div wrapper does appear unnecessary here. You could apply these styles directly to the host element using `:host` selectors in your SCSS rather than creating an extra div with classes. 

This approach would reduce DOM nesting and take better advantage of Web Components' capabilities. You could use SCSS extensions to apply the necessary styles to the host element based on its attributes:

```scss
:host(#{$prefix}-time-picker[disabled]) { 
  @extend .#{$prefix}--select--disabled; 
}
```

This would produce cleaner markup while maintaining the same visual appearance and functionality.

packages/web-components/src/components/select/select.ts
```diff
@@ -443,15 +471,17 @@ class CDSSelect extends FormMixin(LitElement) {
...
                 class="${prefix}--select-option"
-                value="${placeholderItemValue}"
-                selected>
+                value="${placeholderItemValue}">
                 ${placeholder}
               </option>
             `}
         ${this._renderItems(this)}
       </select>
       ${ChevronDown16({ class: `${prefix}--select__arrow` })}
```
riddhybansal: ${ChevronDown16({ class: `${prefix}--select__arrow` })}

I feel this needs aria-hidden="true"

riddhybansal (changes requested):
A minor documentation update is needed for the Select component. We should add a "Form Item" section to the Overview, similar to what exists in the Text Input docs, since we're already wrapping the component in <cds-form-item> in our stories. This would maintain consistency across component documentation.

Look at [Text Input docs](<cds-form-item>)

makafsal:
@heloiselui Resolved merge conflicts.

packages/web-components/src/components/select/select-story.scss
```diff
@@ -0,0 +1,103 @@
+/**
+* @license
```
heloiselui:
```suggestion
```

packages/web-components/src/components/select/select-story.scss
```diff
@@ -0,0 +1,103 @@
...
+*
+* Copyright IBM Corp. 2019, 2024
```
heloiselui:
```suggestion
* Copyright IBM Corp. 2025, 2025
```

packages/web-components/src/components/select/select-story.scss
heloiselui: @makafsal one more detail, based on this [PR](https://github.com/carbon-design-system/carbon/pull/18949), we are removing the licenses from Web Components.

riddhybansal (approved):
LGTM 💯

PR #18409, opened 2025-01-23
feat(tree-links): Adds support for link nodes in the TreeView component

### LINKED ISSUE #18408: [Feature Request]: Add support for links in the treeview/treenode component
### The problem

Currently there is no way to make a tree node contain an accessible link, meaning the TreeView cannot be used for navigation.

### The solution

If a node contains a link, it should contain an anchor tag which directs the user to the page.
Following the APG example, a tree node containing a link should follow the structure:
```
<li role="none">
    <a role="treenode" href="/link" .... />
    <ul role="group">
        children....
```

### Examples

[APG navigation example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/)

### Application/PAL

IBM z/OS Connect

### Business priority

### Available extra resources

Raised PR

### Code of Conduct

Tweakified:
Closes #18408 

Adds support for link nodes in the TreeView component 

#### Changelog

**New**

- Adds `With Links` story to the TreeView component

**Changed**

- TreeView - changed the view so that links can now be considered nodes
- TreeNode - Added updated link node structure 

**Removed**

- N/A

#### Testing / Reviewing

Added unit tests:
- checks if a TreeNode with a link correctly renders with an anchor tag
- checks if a parent TreeNode with a link correctly renders with an anchor tag

e2e:
- new avt test which runs `toHaveNoACViolations` on `With Links` story
- new vrt test which snapshots the `With Links` story


Tweakified:
Not sure how I'd implement this, but could it be worth utilising the Link component?

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -328,43 +369,104 @@ const TreeNode = React.forwardRef<HTMLLIElement, TreeNodeProps>(
     };
 
     if (!children) {
+      if (href) {
+        return (
+          <li role="none">
+            <Link
+              {...treeNodeProps}
```
tay1orjones: Something that @mbgower [surfaced in slack](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1738180729073459?thread_ts=1736333013.380109&cid=C2K6RFJ1G) was that the link should be **wrapped** in the role of treeitem:

> Okay, so you’re just using a link mechansim to populate the main content area, and I assume that link is wrapped in the role of treeitem so that it works in the tree as expected.
> @tay1orjones overall i can’t think of an issue with this approach.

Here `treeNodeProps` contains the `role="treeitem"`, so it's not wrapped but placed directly on the link.

tay1orjones (changes requested):
Thanks again for working on this! I do want to confirm

> * ~`href` should not take the place of or override `active` or `selected`?~ 

_Update: It should not_

> * ~We don't need any visual indication that an item is a link?~ _No styling/visual indication is needed_
>   * ~Right now it feels odd to me that in the story some of them are links and others are not with no way to tell them apart.~
>   * ~@mbgower @laurenmrice I'm curious what you both think.~
>   * ~Should guidance be that TreeView can mix and match href and non-href nodes like this, or should all nodes be required to be the same?~ 

_Update: Yes they can mix and match_

-----

There's also a couple things that would be good to have included here
- [ ] Add a new avt e2e test that visits the `with links` story and calls `toBeAccessible()` on it. 
- [ ] Add a new vrt e2e test that visits the `with links` story and snapshots it

I'll also enable the ci workflows for this PR, codecov might point out some codepaths that still need covered by tests. It looks like there might need to be one more unit test in `TreeNode-test.js` covering the expanded conditions in TreeNode, specifically when `children` and `href` are both `true`.

packages/react/src/components/TreeView/TreeNode.tsx
tay1orjones: I confirmed with @mbgower just now that this isn't a concern.

tay1orjones:
@Tweakified I updated my earlier review comment with the result of a conversation with the team today. With no remaining a11y and visual concerns, the only thing the team identified was that we need the example to be more intentional and show common use.

For the `with links` story, what we propose is adding a content section to the right of the existing TreeView that renders the title of the currently activated TreeNode. As you click/activate other links in the TreeView, the content section updates. 

The anchor tags will need their default behavior prevented so the browser doesn't actually navigate, as well as adding some new state around it all to make it work. This will result in an example that mirrors the one you shared in slack albeit much more simplified.

This way the story example code uses the `href` prop and provides a more in-flight example that clears up the UX and visual concerns I had previously. The avt and vrt tests can still run against this as the first item in the TreeView should be selected by default on initial render.

What do you think?

DanielSmithUI:
Hiya @tay1orjones! Thanks for the update.
Sounds good about the changes, I'll update the PR shortly.

Tweakified:
Hiya @tay1orjones! Thanks for the update.
The changes sound good. I'll update the PR shortly

tay1orjones:
Could the first item in the TreeView be selected by default? Would like to avoid this situation if we can
2025-02-04 at 14 56 30-MAIN-Google Chrome

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -21,6 +21,7 @@ import { useControllableState } from '../../internal/useControllableState';
 import { usePrefix } from '../../internal/usePrefix';
 import uniqueId from '../../tools/uniqueId';
 import { useFeatureFlag } from '../FeatureFlags';
+import { navigate } from '@reach/router';
```
tay1orjones:
```suggestion
```

packages/react/package.json
```diff
@@ -56,6 +56,7 @@
     "@carbon/styles": "^1.74.0",
     "@floating-ui/react": "^0.26.0",
     "@ibm/telemetry-js": "^1.5.0",
+    "@reach/router": "^1.3.4",
```
tay1orjones:
```suggestion
```

packages/react/package.json
```diff
@@ -98,6 +99,7 @@
     "@storybook/react": "^8.4.7",
     "@storybook/react-webpack5": "^8.4.7",
     "@storybook/theming": "^8.4.7",
+    "@types/reach__router": "^1",
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -266,6 +294,25 @@ const TreeNode = React.forwardRef<HTMLLIElement, TreeNodeProps>(
...
+     * @param {*} event select event
+     * @param {TreeNodeProps} node
+     */
+    const onNodeSelect = (event, node?: TreeNodeProps) => {
+      if (onSelect) {
+        onSelect(event, node);
+      }
+      if (href) {
+        event.preventDefault();
+        navigate(href);
```
tay1orjones: We'll need to rely on the default browser functionality here and not integrate with any library directly like this.

Earlier when I mentioned preventing the default behavior of the anchor tag links I meant just within the storybook's example story. Sorry for the confusion!

tay1orjones (changes requested):
One other minor thing here, otherwise LGTM. Thanks for adding the avt and vrt tests!

Tweakified:
Thank you for the review @tay1orjones! I believe I've responded to all the points you raised! 
Please could I get an additional review when able?

kennylam:
I'm experiencing some strange toggle interactions:

1. expand (not select) `Business automation` and/or `Data & Analytics`
2. clicking any other link node, including their own child nodes, will collapse the items. this behavior does not apply to the `Cloud Computing` node - it always stays expanded

Is this intended?

DanielSmithUI:

Tweakified:
@kennylam Thanks for taking a look!
I've managed to resolve that issue now 🥳

tay1orjones (approved):

kennylam (approved):

PR #18416, opened 2025-01-23
test(percy): snap 2 widths on white theme, config network-idle-timeout

guidari:
~Testing if Percy will catch this change in the white theme in the mobile screen under 360px~

See [comment below ⬇️ ](https://github.com/carbon-design-system/carbon/pull/18416#issuecomment-2613188471)

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -8,3 +8,4 @@ snapshot:
 discovery:
   allowed-hostnames:
     - 1.www.s81c.com
+  network-idle-timeout: 1000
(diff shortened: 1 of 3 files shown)


tay1orjones:
@gui I think I figured it out! The widths condition was using `context.themes` instead of `context.theme` 🤦‍♂️ 

I also increased the network-idle-timeout from the default of 500 to 1000, this should help ensure the font assets to finish/settle during the asset discovery phase on percy's servers.

tay1orjones:
@guidari I think I figured it out! The widths condition was using `context.themes` instead of `context.theme` 🤦‍♂️ 

I also configured the `network-idle-timeout` to 1000ms instead of the default of 500ms, this should hopefully help ensure the font assets finish/settle during the asset discovery phase on percy's servers so we don't see snaps changing from Plex not being available.

tay1orjones:
239 new snaps @ mobile width and one snap updated with plex 👍

tay1orjones (approved):

annawen1 (approved):

PR #18447, opened 2025-01-27
fix(tabs): make IconTab's tooltip visible again

tay1orjones:
Part of https://github.com/carbon-design-system/carbon/issues/18477

A change in #16738 inadvertently made it so the tooltip on IconTab wasn't visible.

#### Changelog

**Changed**

- added the intended prop usage to the story
- added to the styles to target the correct selector
- added new vrt tests to cover the focus state of the icon only tabs stories

#### Testing / Reviewing

- Ensure the tooltip is visible on https://deploy-preview-18447--v11-carbon-react.netlify.app/?path=/story/components-tabs--icon-20-only
- Ensure the tooltip is visible on https://deploy-preview-18447--v11-carbon-react.netlify.app/?path=/story/components-tabs--icon-only
- Percy should show new snaps for this showing the focused state w/ visible tooltip


emyarod (approved):
looks good to me!

ariellalgilmore (approved):

kennylam (approved):

e2e/components/Tabs/Tabs-test.e2e.js
```diff
@@ -9,6 +9,7 @@
 
 const { test } = require('@playwright/test');
```
kennylam:
```suggestion
const { expect, test } = require('@playwright/test');
```

Looks like this is missing and causing VRT to fail.

PR #18452, opened 2025-01-28
fix: stopped `onChange` throwing undefined

### LINKED ISSUE #18286: [Bug]: Controlled ComboBox onStateChange calling onChange with selectedItem as 'undefined'
### Package

@carbon/react

### Browser

Chrome

### Package version

1.72.0

### React version

19.0.0

### Description

- onStateChange of combobox receive `changes` as a function parameter. `onStateChange?: (changes: UseComboboxStateChange<Item>) => void`
- `changes` object can't have `selectedItem` as a property all the time.
- So, It is passing `undefined` as a `selectedItem` for steps mentioned below.

### Reproduction/example

https://stackblitz.com/edit/github-zurxxwnj?file=src%2FApp.jsx,package-lock.json&preset=node=

### Steps to reproduce

1. Select value '1' from combobox.
2. Click on combobox and remove value '1' with backspace/delete.
3. Now click on MenuItem with value '1'.
4. combobox input value becomes empty.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

### LINKED ISSUE #18361: [Bug]: Toggling the same option in ComboBox dropdown doesn't always trigger the onChange event
### Package

@carbon/react

### Browser

Chrome

### Package version

1.72.0

### React version

18.3.1

### Description

Try to select the same option in ComboBox multiple time, onChange event is not getting called on every toggle.

I think, similar kind of issue already registered here - https://github.com/carbon-design-system/carbon/issues/18286

### Reproduction/example

https://codesandbox.io/p/sandbox/nlzr7m

### Steps to reproduce

You select one option from the dropdown, and then toggle it, that option will be deselected. After that if you try to select that option again, no such onChange event gets triggered even though the option will be selected.
You need to clear the selected item by clicking the cross icon to get it selected again properly.

### Suggested Severity

### Application/PAL

IBM OpenPages

### Code of Conduct

Gururajj77:
Closes #18286 
Closes #18361

This PR resolves two things:
- stops `onChange` from getting called again when an already selected item is selected again.
- stops `onChange` from throwing `undefined`.

#### Changelog

**Changed**

- added the line `!isEqual(selectedItemProp, newSelectedItem)` to the keyboard event handler inside the `onStateChange` to make sure we don't call the `onChange` even when we select the same item.

#### Testing / Reviewing

- Go to the "On Change Test" story.
- open the console.
- select an item
- there should be log with the selected item
- try to select the same item
- we should not see the same item getting logged again.
- try this using both keyboard and mouse


emyarod (approved):
looks good to me!

packages/react/src/components/ComboBox/ComboBox.stories.js
```diff
@@ -379,6 +379,52 @@ export const _fullyControlled = (args) => {
   );
 };
 
+export const OnChangeTest = (args) => {
```
kennylam: This story can probably be removed, no?

kennylam (commented):
@Gururajj77 Nice! Maybe it would be worth it to add this check to the unit tests?

packages/react/src/components/ComboBox/ComboBox.stories.js
Gururajj77: Ill remove these after I got 2 approvals as its needed for tests.

riddhybansal (approved):
LGTM!! Thanks for the test cases !!

PR #18495, opened 2025-01-31
feat(actionable-notification): remove focus wrap when role is provided

### LINKED ISSUE #17392: [Feature Request]: Add a "disableFocusTrap" prop to the ActionableNotification component
### The problem

The ActionableNotification component is currently implemented to require a user to dismiss it before they can interact with the rest of the application. When an actionable notification is opened, it will immediately grab focus and then trap focus inside the notification until the notification is dismissed.

We have a use case where the notification is used for information purposes about asynchronous processes. The user triggers some action that completes later and the notification is used to inform the user the action is complete and it includes a button to take the user to that action to see the results. When an action is complete, is should be informational to the user and should not forcibly move or trap focus. There may even be multiple notifications. We know the current behavior is as designed and we would like a prop to allow us to control the behavior.

There is currently a `hasFocus` prop on the component now that disables the initial grabbing of focus but this prop is deprecated and even with that prop, the notification will still trap focus if focus is moved inside of it.

Here are a few related issues discussing this problem:
https://github.com/carbon-design-system/carbon/issues/15319
https://github.com/carbon-design-system/carbon/issues/16231

### The solution

Add something like a `disableFocusTrap` prop to the component that will disable both the initial grabbing of focus and the focus trap that occurs when elements inside the notification get focus.

### Examples

Here's a video showing a simplified example. The user is interacting with a chat bot that starts an asynchronous process. When the process complete, focus is moved away from the field while the user is in the middle of typing. Just recording this video was hard because I kept pressing the space bar as I was typing which just dismissed the notification immediately when it got focus.

https://github.com/user-attachments/assets/34d0ef97-3c99-4721-bf02-2cc0be2c680d

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

tay1orjones:
Closes #17392 

#### Changelog

**New**

- a new section in the storybook docs outling the additional work required when configuring `role`

**Changed**

- only wrap focus if `role === 'alertdialog'`, otherwise don't wrap focus

#### Testing / Reviewing

- Go the the actionable notification story
- Set the `role` prop to `status` in the controls
- Test keyboard focus by tabbing, observe that focus no longer wraps within the component. Focus can now leave the component. 
- Remove the `role` prop control value to restore the default and focus should wrap again


packages/react/src/components/Notification/Notification.mdx
```diff
@@ -65,6 +65,27 @@ an `aria-describedby` on the interactive element with a value that matches the
 
 <Canvas id="components-notifications-actionable--default" />
 
+### Using the `role` prop
+
+ActionableNotification's `role` prop allows you to provide a different
+`aria-role` than the default, `alertdialog`.
+
+This prop should be used with caution as it will disable the wrapping of focus
```
ariellalgilmore: minor possible wording update. also, noticed it in the prop description too
```suggestion
This prop should be used with caution as it will disable the focus wrapping
```

ariellalgilmore (approved):
LGTM!

guidari (approved):

PR #18496, opened 2025-01-31
fix(17119): ComboBox not displaying selected option

### LINKED ISSUE #17119: [Bug]: Carbon Combobox component not displaying selected option once search query is cleared
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox

### Package version

1.63.1

### React version

18.0.0

### Description

I noticed an issue in combobox component. Once we searches and select an item from list, it will be selected. But when we clear the search box and click outside, it seems the field has no value selected. Only when we expand the dropdown, we can see the item as checked.
When we search for an item which is not in the list and click outside, the selected value is populated back in the search field.
Attaching a video for reference.

https://github.com/user-attachments/assets/3aa0d2dc-15a6-4766-a46d-7b334bfafe93

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-combobox--playground&args=allowCustomValue:!false

### Steps to reproduce

- Select a value from combobox. The dropdown will be collapsed and the item will be displayed in search field.
- Now remove the search term from the field. and click outside. The value won't be displayed in the input field. but when expanded, there we can see the prev item is already in selected state.

When we search for some other item which is not in the list and click outside, the selected item will be populated back in the input field. Same behaviour is expected on clearing the input field when an item is already selected.

### Suggested Severity

### Application/PAL

IBM Data Product Hub

### Code of Conduct

enricobguedes:
Closes https://github.com/carbon-design-system/carbon/issues/17119

Added some new checks for combo box to make it work as intended.

#### Changelog

**New**

- Added new checks within combo box for empty inputs and no text highlights to invalidate user selection (return to empty state as intented).
- Added check for the blur functionality to do the same as the above, returning the component state to empty/not selected
- Both mimic the clearSelection state.

**Changed**

- Changed

**Removed**

- {{removed thing}}

#### Testing / Reviewing

Using the storybook Carbon component reviewers can test as shown in the original issue reported.


enricobguedes:

enricobguedes:
recheck

PR #18498, opened 2025-02-01
fix(17119): combobox not displaying selected option once search query is cleared

### LINKED ISSUE #17119: [Bug]: Carbon Combobox component not displaying selected option once search query is cleared

enricobguedes:
Closes https://github.com/carbon-design-system/carbon/issues/17119

#### Changelog

**New**

- Added new checks within combo box for empty inputs and no text highlights to invalidate user selection (return to empty state as intented).
- Added check for the blur functionality to do the same as the above, returning the component state to empty/not selected
- Both mimic the clearSelection state.

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

Using the Carbon storybook for react components, reviewers can test as shown in the original issue reported.


packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1013,6 +1016,15 @@ const ComboBox = forwardRef(
...
                   }
+                  if (!inputValue && highlightedIndex == -1) {
+                    if (event.key == 'Enter') {
+                      if (!event.currentTarget.value) {
+                        if (!isOpen) toggleMenu();
+                        selectItem(null);
+                        event.preventDownshiftDefault = true;
+                      }
+                    }
+                  }
```
guidari: Can we simplify this if clause into a single condition? I might be wrong, but it looks like `inputValue` and `event.currentTarget.value` are the same, making this check a bit redundant.

I wonder if something like that would work:

```
if (!inputValue && highlightedIndex === -1 && event.key === 'Enter') {
  if (!isOpen) toggleMenu();
  selectItem(null);
  event.preventDownshiftDefault = true;
}
```

We would have to cover that with Jest and Playwright tests as well.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -843,6 +843,9 @@ const ComboBox = forwardRef(
 
     const handleFocus = (evt: FocusEvent<HTMLDivElement>) => {
       setIsFocused(evt.type === 'focus');
+      if (!inputRef.current?.value && evt.type === 'blur') {
```
guidari, line 846: Would `inputValue` work in this case? Just wondering...

e2e/components/ComboBox/ComboBox-test.avt.e2e.js
```diff
@@ -91,7 +91,9 @@ test.describe('@avt ComboBox', () => {
     await expect(menu).toBeHidden();
     await expect(combobox).toBeFocused();
     await page.keyboard.press('Enter');
-    await expect(menu).toBeVisible();
+    // Expect to be hidden if no highlight or text in combobox.
+    await expect(menu).toBeHidden();
+    await page.keyboard.press('Space');
```
guidari: Remove this test and keep the `await expect(menu).toBeVisible();`

guidari (changes requested):
Looking good so far! Thanks for opening that! 🚀 

A couple things. Looks like the Combobox is not opening when pressing `Enter` anymore. Combobox should open when pressing `Enter`. The test should remaing the same in this case.

packages/react/src/components/ComboBox/ComboBox-test.js
```diff
@@ -764,5 +764,17 @@ describe('ComboBox', () => {
 
       expect(input).toHaveDisplayValue('Apple');
     });
+
+    it('should remove input and enter new conditions', async () => {
```
guidari, line 769: I'm not sure about this test. Is it testing the functionality you created?

Otherwise we could add a test like that, to test step by step the interaction with the Combobox.

You will noticed that if you remove the changes you made in the `Combox.tsx` this test will fail **(as expected)**, however it will pass with the current changes **(as expected)**

```
 it('should open the menu and select null when Enter is pressed with no input and no highlighted item', async () => {
      const onInputChange = jest.fn();

      render(<ComboBox {...mockProps} onInputChange={onInputChange} />);

      await userEvent.type(findInputNode(), 'apple');
      expect(findInputNode()).toHaveDisplayValue('apple');
      await userEvent.keyboard('[Enter]');

      // Delete the selected item
      await userEvent.keyboard('[Backspace]');
      await userEvent.keyboard('[Backspace]');
      await userEvent.keyboard('[Backspace]');
      await userEvent.keyboard('[Backspace]');
      await userEvent.keyboard('[Backspace]');
      // check for an empty value
      expect(findInputNode()).toHaveDisplayValue('');

      // blur
      await userEvent.keyboard('[Tab]');
      assertMenuClosed(mockProps);

      // open the menu
      await userEvent.click(findInputNode());
      assertMenuOpen(mockProps);

      // check if the `li` item are all false
      for (let i = 0; i < mockProps.items.length; i++) {
        const item = findMenuItemNode(i);
        console.log({ item });
        expect(item).toHaveAttribute('aria-selected', 'false');
      }
    });
    ```

enricobguedes: Yes, this is the test I've created for this addition! I will add this extra evaluation then, thanks!

guidari (approved):

kennylam (approved):
LGTM, thanks!

PR #18506, opened 2025-02-03
fix: add autoAlign to inner IconButton

### LINKED ISSUE #17305: [Bug]: `OverflowMenu` tooltip should support auto align (like the stand alone `Tooltop`), currently it gets cropped off when the container ends
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

@carbon/react@1.65.0

### React version

17.0.2

### Description

**Problem**

When the `OverflowMenu` is placed at the corner of an container the tooltip that is now with Carbon 11 rendered by default is cropped of. There is no option to bypass this problem, except ugly JavaScript or CSS hacking.

Here is an example with `DataTable`:

image

Here is an example with the component right in the story book:

image

**Expectation**

From the carbon `Tooltip` component I saw the `autoAlign` prop that does this magic for us: Always shows the tooltip proper. I adore it really much, since it always lets it display proper, no matter where this tooltip is or how small or big the browser window is. But here for the `OverflowMenu` it is not there. Only manual alignments. Since in the underlaying component you anyhow use a `Tooltip` provide also a `autoAlign` prop for the `OverflowMenu` and bubble it down.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--playground&args=aria-label:A+very+long+text+that+gets+cropped+off;iconDescription:A+very+long+text+that+gets+cropped+off

### Steps to reproduce

1. Open the link
2. Hover the `OverflowMenu` button.
3. Experience what you see above in the image.

### Suggested Severity

### Application/PAL

IKC

### Code of Conduct

richardpilot:
Closes #17305

Added the autoAlign prop to allow the inner IconButton to use the floating-ui component. This allows for the tooltip popover to render over the table rather than inside it, preventing the overflow scrollbar from appearing.

#### Changelog

**Changed**

- Added `autoAlign={true}` prop to inner IconButton in OverflowMenu

#### Testing / Reviewing

- Verified that the autoAlign property on the IconButton within a OverMenu within a DataTable is honored.


packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -719,6 +719,7 @@ class OverflowMenu extends React.Component<
             aria-haspopup
             aria-expanded={open}
             aria-controls={open ? menuBodyId : undefined}
+            autoAlign={true}
```
tay1orjones, line 722: I think this should be added as an optional experimental boolean prop instead of hardcoding to `true`. It's a sizable behavior change that has a decent chance of breaking things in consuming projects.

tay1orjones (changes requested):
Thanks for doing this! I just have one suggestion

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
richardpilot: I'm not familiar with "optional experimental boolean prop", can you give me an example of it's usage. It feels a little odd to have a bug fix locked behind a prop.

PR #18544, opened 2025-02-10
fix: added condition to clear input when closed using chevron

### LINKED ISSUE #18484: [Bug]: Combobox allows custom value even when `allowCustomValue = false`
### Package

@carbon/react

### Browser

Chrome

### Package version

1.74.0

### React version

18

### Description

Through a specific interaction the combobox will allow a custom value to be retained in the input even when `allowCustomValue` is off/false.

There are many similar issues that may be related or wrapped up with this one:

### Reproduction/example

### Steps to reproduce

1. Open the combobox
2. Enter a value not in the listbox options
3. Click the chevron
4. The combobox closes, focus can move away, the unique value is retained.

https://github.com/user-attachments/assets/481d2f96-1860-497b-99b8-c073dff244c9

By contrast, do these same steps but instead of clicking the chevron, click away from the component altogether. The unique value is cleared.

https://github.com/user-attachments/assets/d5b0d635-237c-4845-9c03-eab2dd8597dc

The input should be cleared in both instances.

---

This is very similar to https://github.com/carbon-design-system/carbon/issues/17290, maybe even a duplicate though the interaction there is slightly different. 

It also may be related to or wrapped up with other combobox value selection issues:

* https://github.com/carbon-design-system/carbon/issues/17192
* https://github.com/carbon-design-system/carbon/issues/18158
* https://github.com/carbon-design-system/carbon/issues/17119

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #18484 

Fixed the bug, when chevron is closed, the input wasn't getting cleared if the typed text doesn't match any items.

#### Changelog

**New**

- added this new condition in the `stateReducer`'s  `FunctionToggleMenu` case

```            
         if (
              !changes.isOpen &&
              state.inputValue &&
              highlightedIndex === -1 &&
              !allowCustomValue
            ) {
              return {
                ...changes,
                inputValue: '', // Clear the input
              };
            }
```
- added a test to make sure this condition is checked.

#### Testing / Reviewing
- open the default story from Combobox component
- type in the input, and make sure it doesn't match with any items from the list
- click the chevron to trigger the close event
- see that the text you typed in the input is cleared.


emyarod (approved):
looks good to me!

2nikhiltom (approved):
LGTM, just has a small conflict in test file

riddhybansal (approved):

alisonjoseph (approved):

PR #18584, opened 2025-02-13
fix: added condition to handle selection clearing

### LINKED ISSUE #17290: [Bug]: ComboBox allow non-existing value after a set of actions
### Package

@carbon/react

### Browser

Chrome

### Package version

1.64.1

### React version

18.2.0

### Description

By going through a certain set of actions, you can select a non existing option from the `ComboBox` with `allowCustomValue={false}`

Same behavior is observed from the official Storybook.

### Reproduction/example

### Steps to reproduce

1. Select an option 
2. Type a random value
3. Enter (return) and blur

Error reproduction:

https://github.com/user-attachments/assets/1dcb9c60-8cea-4540-a4f7-227b0e8b9fa0

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #17290 

Fixed the bug which keeps the selection in the list, even when the input value doesn't find any match with the items.

#### Changelog

**New**

- added new condition in the `InputKeydownEnter` case of the `stateReducer` to clear the preset selected item, when the user presses enter key with an input value which doesn't have any match with the list of items.

- tests to cover this case

#### Testing / Reviewing

- ci should pass
- open the deploy preview
- go to the `ComboBox` default story
- open the list and select an item
- click on the input and type anything after the selected item's text, make sure it doesn't match with any of the items
- press enter
- see that the combobox doesn't have any selected item.


2nikhiltom (approved):
LGTM, input clears when chevron is clicked and onChange is called with {selectedItem: null}

kennylam (approved):

PR #18599, opened 2025-02-15
fix: update compareItems type in MultiSelect

### LINKED ISSUE #18199: [Bug]: `MultiSelect` `compareItems` parameter types are not correct
### Package

@carbon/react

### Browser

_No response_

### Package version

1.67.1

### React version

_No response_

### Description

The types for the `compareItems` prop's parameters are not correct. Each parameter is a `string`, not an `ItemType`.

### Reproduction/example

https://stackblitz.com/edit/github-79u5xe?file=src%2FApp.tsx

### Steps to reproduce

See reproduction and type error.

### Suggested Severity

### Application/PAL

DataPower

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18199

Updated `compareItems` type in `MultiSelect`.

#### Changelog

**New**

- Added `defaultCompareItems` tests.

**Changed**

- Updated `compareItems` type in `MultiSelect`.

#### Testing / Reviewing

- Run the test that was added.
- Use `compareItems` in a TypeScript file similar to the reproduction in the issue and check for errors.

### DIFF
--- a/packages/react/src/components/MultiSelect/MultiSelectPropTypes.ts
+++ b/packages/react/src/components/MultiSelect/MultiSelectPropTypes.ts
@@ -62,11 +62,7 @@ export interface MultiSelectSortingProps<ItemType> {
    * Provide a compare function that is used to determine the ordering of
    * options. See 'sortItems' for more control.
    */
-  compareItems?(
-    item1: ItemType,
-    item2: ItemType,
-    options: SharedOptions
-  ): number;
+  compareItems?(item1: string, item2: string, options: SharedOptions): number;
 
   /**
    * Provide a method that sorts all options in the control. Overriding this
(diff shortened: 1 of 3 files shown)


packages/react/src/components/MultiSelect/MultiSelectPropTypes.ts
```diff
@@ -46,14 +46,14 @@ interface SharedOptions {
   locale: string;
 }
 
+interface CompareItems {
+  (itemA: string, itemB: string, options: SharedOptions): number;
```
tay1orjones, line 50: Are we losing some strictness here by moving the items from `ItemType` to `string`?

tay1orjones (approved):
One question but nothing to hold up a merge. Thanks for this contribution!

packages/react/src/components/MultiSelect/MultiSelectPropTypes.ts
adamalston: Wasn't that the problem, that the items had type `ItemType` instead of `string`? Aren't the items going through `itemToString` before they go to the comparison function?

ariellalgilmore (approved):
LGTM

PR #18602, opened 2025-02-15
fix: use initialSelectedItem only on initial render in ComboBox

### LINKED ISSUE #17192: [Bug]: Carbon ComboBox does not reset displayed "selectedItem" value after setting to null
### Package

@carbon/react

### Browser

Firefox

### Package version

@carbon/react: 1.60.3

### React version

react: 18.2.0

### Description

I am using a `ComboBox` populated with values from my store. After the user submits the value selected in the combo box, I update the selectedItem to "null", which is also derived from my store. The displayed value in the combo box does not reset and retains the last selected item, even after resetting the selectedItem to "null" in the store.

A workaround for this issue is to use `useRef` for the key of the of combo box, and `useEffect` to update the `current` value of the key any time the selectedItem changes. This will force a re-render of the combo box, resetting its value. Example of workaround:

```
const comboboxKeyRef = useRef(0)

useEffect(() => {
    comboboxKeyRef.current = `combobox_${Date.now()}`
  }, [selectedItem])
  
  
<ComboBox
            ...other props
            items={items}
            key={comboboxKeyRef.current}
            selectedItem={selectedItem}
/>
```

### Reproduction/example

https://stackblitz.com/edit/github-wkihg5?file=src%2FApp.jsx

### Steps to reproduce

Reproduction example provided in link above. Have `selectedItem` derived from state and some mechanism that resets `selectedItems` after user action. Combo box will not reset its displayed item, even after `selectedItem` is set to null.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/17192

Updated `ComboBox` to use `initialSelectedItem` only on initial render.

#### Changelog

**Changed**

- Updated `ComboBox` to use `initialSelectedItem` only on initial render.

#### Testing / Reviewing

```sh
yarn test packages/react/src/components/ComboBox/ComboBox-test.js
```

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox-test.js
+++ b/packages/react/src/components/ComboBox/ComboBox-test.js
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2016, 2023
+ * Copyright IBM Corp. 2016, 2025
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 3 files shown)


Gururajj77:
@adamalston , the problem mentioned in the issue you have tagged, isn't getting rectified by this PR, please re-check your code.

adamalston:
> the problem mentioned in the issue you have tagged, isn't getting rectified by this PR

How are you checking if the issue was addressed? Unless I'm misunderstanding the issue, when I checked locally using the Storybook, it had been addressed with the changes in this pull request.

https://github.com/user-attachments/assets/3fd52435-66ca-435c-8e7e-9c76d25104db

Gururajj77:
@adamalston, the video shows its working as expected. The "Fully Controlled" story is where I tested it, can you add a test story so that we can verify and see how the code is written?

tay1orjones (approved):
This is awesome, thanks 👍 

Please remember to remove the test story before merging

heloiselui (approved):
🚀🎉

adamalston:
I don't see a button to re-request a review from @Gururajj77 but that's what I'm awaiting before deleting the story. I don't want to regress anything and this component seems very finicky.

Gururajj77 (approved):
Hey @adamalston , thanks for fixing this PR, this looks good and works in the scenario I wanted to check.

The issue with why "Fully Controlled" story isn't working even after this fix, is that it has storybook's `args` that is spread and ` _fullyControlled.argTypes = { ...sharedArgTypes };` is also being used to override the props we pass. 

And the scenario Which I wanted to test is with these items:

```
  const options = [
    {
      id: 'option-1',
      text: 'Option 1',
    },
    {
      id: 'option-2',
      text: 'Option 2',
    },
    {
      id: 'option-3',
      text: 'Option 3',
    },
  ];
  ```
  
  Wanted to make sure that this fix works when we pass each item as an object.

adamalston:
I'm not sure why this pull request was merged. I didn't delete the story which I mentioned in https://github.com/carbon-design-system/carbon/pull/18602#issuecomment-2672908488 that I still needed to do. @Gururajj77, how do I reassign myself to https://github.com/carbon-design-system/carbon/issues/17192?

PR #18606, opened 2025-02-16
fix: update FilterableMultiSelect onMenuChange behavior

### LINKED ISSUE #18270: [Bug]: FilterableMutiselect "onMenuChange" triggering multiple times on render
### Package

@carbon/react

### Browser

Chrome

### Package version

1.72

### React version

18.3

### Description

After _@carbon/react_ v1.65, _onMenuChange_ started triggering multiple times on the first render. I have a state update that is based on the _onMenuChage_ callback which is now causing my application to have an odd behavior.

Also, now I need to use a _useCallback_ so the application doesn't start looping and crashes. This is also reproducible by just removing the _useCallback_ from the function passed to _onMenuChange_.

I tracked it down to this bug fix: https://github.com/carbon-design-system/carbon/issues/16896

Previous behavior the _onMenuChange_ callback was only triggered after the user interacted with the dropdown.

### Reproduction/example

https://stackblitz.com/edit/github-ehxczfmb?file=src%2FApp.tsx

### Steps to reproduce

1. Start the application

expected result: count should only be updated after the menu open/closes
actual result: count state is updated after render

Downgrading _@carbon/react_ to v1.65 will make it work as expected.

### Suggested Severity

### Application/PAL

RPA

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18270

Updated `FilterableMultiSelect`'s `onMenuChange` behavior.

#### Changelog

**Changed**

- Updated `FilterableMultiSelect`'s `onMenuChange` behavior.

#### Testing / Reviewing

```sh
yarn test packages/react/src/components/MultiSelect/__tests__/FilterableMultiSelect-test.js
```


Gururajj77:
Hey @adamalston , could you write more about what are the changes you did and how it affects the component's behaviour??

adamalston:
I've updated the logic so that:

- When the component is controlled (i.e., the `open` prop is provided), `onMenuChange` is called exactly once on mount (if `open` is true) and doesn't trigger again on re-renders with the same `open` value.
- In uncontrolled mode, the callback isn't called on mount, but will be triggered when a user interacts with the components.

The tests reflect these scenarios.

packages/react/src/components/MultiSelect/__tests__/FilterableMultiSelect-test.js
```diff
@@ -536,6 +536,12 @@ describe('FilterableMultiSelect', () => {
           item ? (
             <span className="test-element" data-testid="custom-id-item">
               {item.text}{' '}
+              {/*
+              // TODO: Should `alt` be changed to `aria-label`? There appears to
+              // be an accessibility issue with this component.
+              // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#description
+              // https://dequeuniversity.com/rules/axe/4.10/role-img-alt?application=axe-linter
```
tay1orjones: Interesting, yeah looks like that would make sense

tay1orjones (approved):
LGTM! 👍

adamalston:
I see the security scan is failing. I guess it didn't like me merging `main` into the branch. Doesn't seem like something I should need to revert though.

ariellalgilmore (approved):
LGTM!

Issue #17234, opened 2024-08-21
Evaluate and consolidate `useMergedRefs` helpers


tay1orjones:
There's been a `useMergedRefs` helper in the codebase for quite a while. 

https://github.com/carbon-design-system/carbon/blob/ab9cd8c843d64427f496ac4ca781e165a3d07d2e/packages/react/src/internal/useMergedRefs.ts#L10-L26

With the addition of floating-ui we now have access to it's helper by the same name. This brings up a few things to consider:

* Are these helpers equivalent?
* Should we use the floating-ui one and remove our internal one?
* Would we prefer to instead remain using our own and add an eslint rule to help enforce not using the floating-ui one?

adamalston:
> Are these helpers equivalent?

They're similar but not equivalent. [`useMergedRefs`](https://github.com/carbon-design-system/carbon/blob/fe90ac8ddb8364e229ecb3ef8037e368466c3ee6/packages/react/src/internal/useMergedRefs.ts) is simpler by just assigning the node to each reference. [`useMergeRefs`](https://github.com/floating-ui/floating-ui/blob/b9ad108d9b0257fe1d7d25a3b187fd5189062b9a/packages/react/src/hooks/useMergeRefs.ts#L3) takes it a step further by also managing the lifecycle, calling any cleanup functions returned by function references, and clearing stale references. Also, `useMergeRefs` has more flexible types.

Overall, `useMergeRefs` appears to be a more robust hook. I'm testing some changes and may open a pull request.

PR #18611, opened 2025-02-18
fix: update ref handling in ListBoxMenuItem useIsTruncated hook

### LINKED ISSUE #18459: [Bug]: ComboBox does not display the full name for items with truncated text (eg.tooltip) when shouldFilterItem is true
### Package

@carbon/react

### Browser

Chrome

### Package version

10.5.0

### React version

18.2.0

### Description

Using ComboBox from @carbon/react,
When the search box is empty, the item with truncated text shows the tooltip as expected.
Image
However, when you type something in the search box, the tooltip stops showing.
Image

### Reproduction/example

https://stackblitz.com/edit/jwernaba-awct9mvr?file=src%2FApp.jsx

### Steps to reproduce

load the above example,
type `e` in search box, 
the list should be filtered to 3 items.
hover over on the third one, no tooltip would be showing.

App.jsx as follows:

import React from 'react';

import { ComboBox } from '@carbon/react';

export default function App() {
  const items = [
    {
      id: 'option-0',
      text: 'An example option that is really long to show what should be done to handle long text',
    },
    {
      id: 'option-1',
      text: 'Option 1',
    },
    {
      id: 'option-2',
      text: 'Option 2',
    },
    {
      id: 'option-3',
      text: 'Option 3 - a disabled item',
    },
    {
      id: 'option-4',
      text: 'Option 4',
    },
    {
      id: 'option-5',
      text: 'Option 5 really long to show what should be done to handle long text',
    },
  ];

  const comboBoxFilterCallback =({
    inputValue,
    item,
    itemToString,
  }) => {
    const label = item.text;
    const searchLabel = label?.toLocaleLowerCase() || '';
  
    const searchToken = inputValue ? inputValue.toLocaleLowerCase() : '';
  
    return searchLabel.includes(searchToken);
  };
  return (
    <div
      style={{
        width: 300,
      }}
    >
      <ComboBox
        id="carbon-combobox"
        items={items}
        itemToString={(item) => (item ? item.text : '')}
        titleText="ComboBox title"
        helperText="Combobox helper text"
        onChange={() => {}}
        shouldFilterItem={comboBoxFilterCallback}
        downshiftProps= {
          {isOpen: true}
        }
      />
    </div>
  );
}

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18459

Updated `ref` handling in `ListBoxMenuItem` `useIsTruncated` hook.

#### Changelog

**Changed**

- Converted `useIsTruncated` to a generic hook with explicit `HTMLElement` typing.
- Improved forwarded ref handling by ensuring proper fallback to a local ref.
- Included dependencies in `useIsTruncated` to accurately detect truncation.

#### Testing / Reviewing

~~I'm not sure how to write a test for this functionality. I did take the code from https://github.com/carbon-design-system/carbon/issues/18459 and test it to ensure that `title`s were added to the appropriate items.~~

```sh
yarn test packages/react/src/components/ListBox
```


packages/react/src/components/ListBox/ListBoxMenuItem.tsx
```diff
@@ -12,22 +12,28 @@ import React, {
...
   useEffect(() => {
     const element = ref.current;
-    const { offsetWidth, scrollWidth } = element;
-    setIsTruncated(offsetWidth < scrollWidth);
-  }, [ref, setIsTruncated]);
+
+    if (element) {
+      setIsTruncated(element.offsetWidth < element.scrollWidth);
+    }
+  }, [ref, ...deps]);
```
tay1orjones: At first glance I thought this was an issue because changing the value of the ref won't trigger a re-render, but [turns out that's not the case](https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array#:~:text=What%20about%20a%20custom%20hook%20that%20accepts%20a%20ref). TIL!

packages/react/src/components/ListBox/ListBoxMenuItem.tsx
```diff
@@ -74,18 +80,27 @@ export type ListBoxMenuItemComponent = ForwardRefReturn<
...
+    { children, isActive = false, isHighlighted = false, title, ...rest },
     forwardedRef: ListBoxMenuItemForwardedRef
   ) {
     const prefix = usePrefix();
-    const ref = useRef(null);
-    const isTruncated = useIsTruncated(forwardedRef?.menuItemOptionRef || ref);
+    const localRef = useRef<HTMLDivElement>(null);
+
+    // Try to get a RefObject from `forwardedRef.menuItemOptionRef`.
+    // If it's a callback ref (or not provided), fall back to `localRef`.
```
tay1orjones: Would it make sense to test this with a mocked callback ref function?

https://github.com/carbon-design-system/carbon/blob/7b9a05be22b9bee34f70299f7568f1aeb6e7c06a/packages/icon-build-helpers/src/builders/react/components/__tests__/Icon-test.js#L40-L47

tay1orjones (commented):
Should `useIsTruncated` house the ref inspection? I'm anticipating it's soon going to be moved out of this file to be something we can import into various other components that need it.

adamalston:
> Should `useIsTruncated` house the ref inspection? I'm anticipating it's soon going to be moved out of this file to be something we can import into various other components that need it.

Let me know if what's in https://github.com/carbon-design-system/carbon/commit/dd403b6b10de3d8a6ccd8d2e21174c908c1e890a is better or worse. I tried to make it more composable but I'm not sure if I hit the mark.

tay1orjones:
@adamalston Yep, we actually have a useMergedRefs helper already! haha

https://github.com/carbon-design-system/carbon/blob/009351acad1227291c713ebc18073995da20aeba/packages/react/src/internal/useMergedRefs.ts#L15-L28

There's one available through floating-ui too, if you have an opinion for https://github.com/carbon-design-system/carbon/issues/17234

tay1orjones (approved):
Once the mergedRefs helper is deduped I think this looks good to go 👍

adamalston:
I will have to add TypeScript expect error directives to be able to use either of those.

2nikhiltom (approved):

Issue #11347, opened 2022-05-04
[Bug]: theme neutral color token for popover bg when overlapping zoned theming


jnm2377:
### Summary

This issue was brought up when migrating the Carbon tutorial to v11.
In the tutorial, we have the header as `g100` and the content as `white`. The UI shell icon buttons all use tooltips (since all our icon only buttons do for a11y). We realized that when the tooltip was showing, the `g100` theme tooltip bg color was a light color, which blended in too much with the `white` bg content. 

This might not be the only time something like this happens when zoned themes have popover content that overlap with the different themes. 
<img width="204" alt="Screen Shot 2022-04-28 at 1 58 12 PM" src="https://user-images.githubusercontent.com/32556167/166816733-a45be4db-88b0-4439-8f6f-3e1b6063849d.png">

<img width="234" alt="Screen Shot 2022-04-28 at 1 57 43 PM" src="https://user-images.githubusercontent.com/32556167/166816746-587380b0-e460-479a-b303-3cd6cce15edb.png">

One of the options we considered was having a theme neutral color token for these situations. But then also need a way to be able to select that or configure that.

For the time being, we used `$gray-20` as the tooltip bg color (per design guidance).

### Justification

Better visibility and a11y. When the color blends in too much with one theme, it can be easy to miss the content.

### Required functionality

Ability to be able to easily configure (via prop or something) a theme neutral bg color

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

sstrubberg:
Just chatted about this one in Backlog Cleaning. We decided that this was more of a bug than a feature and prioritized it as such. We want to see if other v11 adopters experience this before prioritizing the work because there will need to be some research sprint(s) to figure out a path forward. Thanks for logging this!

emyarod:
> Is this resolved by [#18636](https://github.com/carbon-design-system/carbon/pull/18636)? Sort of seems like it. [@alisonjoseph](https://github.com/alisonjoseph) [@emyarod](https://github.com/emyarod) if you agree we could close this

yes along with https://github.com/carbon-design-system/carbon/pull/18725 I believe this issue is resolved

PR #18636, opened 2025-02-19
feat: make `highContrast` and `dropShadow` props configurable

emyarod:
Related https://github.com/carbon-design-system/carbon-labs/issues/418 #11347

This PR makes the `highContrast` and `dropShadow` Popover props configurable in the Tooltip and IconButton. The default values will remain as-is, but this change will allow the new UI shell side nav variant to inherit these styles as needed without the style overrides.

#### Changelog

**Changed**

- expose `highContrast` and `dropShadow` props in the Tooltip and IconButton


alisonjoseph (approved):
LGTM

riddhybansal (approved):

PR #18639, opened 2025-02-19
fix(es-toolkit): include as internal to avoid module resolution errors

### LINKED ISSUE #18403: [Bug]: Components using Copy fails unit tests fail due to missing module 'es-toolkit/compat'
### Package

@carbon/react

### Browser

_No response_

### Package version

1.74.0

### React version

18.3.1

### Description

When running unit tests, components using Copy produces the following error and prevents test suite from running:

`Test suite failed to run                                                                                                                                                                                        
                                                                                                                                                                                                                    
    Cannot find module 'es-toolkit/compat' from 'node_modules/@carbon/react/lib/components/Copy/Copy.js'                                                                                                            
                                                                                                                                                                                                                    
    Require stack:                                                                                                                                                                                                  
      node_modules/@carbon/react/lib/components/Copy/Copy.js                                                                                                                                                        
      node_modules/@carbon/react/lib/components/CodeSnippet/CodeSnippet.js                                                                                                                                          
      node_modules/@carbon/react/lib/index.js`

I have attached the project:
[react-failure.zip](https://github.com/user-attachments/files/18509084/react-failure.zip)

### Reproduction/example

NIL

### Steps to reproduce

1. `npm install`
2. `npm test`

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Closes #18403, [Internal slack thread](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1736923058543039) for reference

~Updating this import path to be more specific should remove the possibility that a consuming app's bundler can't resolve this module.~ The updated import path broke nextjs' module resolution because it _does_ follow proper behavior wrt `exports`, so I'm trying another approach.

I've retooled this to now include es-toolkit in the final bundle. This increases our bundle size very slightly, but should bypass any potential module resolution errors. 

#### Changelog

**Changed**

- update usages of `'es-toolkit/compat'` to `'es-toolkit/dist/compat'`

#### Testing / Reviewing

- Pull down and look at the build output
- You should see the compat modules we use (and not everything else) included in a new node_modules folder
- You should see in components that use it, like ComposedModal, that the path properly resolves to that internal location and does not just have `import ... from 'es-toolkit/compat'`

2025-02-20 at 15 06 49-debounce mjs js — monorepo-Code
2025-02-20 at 15 07 10-ComposedModal js — monorepo-Code


adamalston (approved):

ariellalgilmore (approved):
LGTM!

annawen1 (approved):

PR #18642, opened 2025-02-20
refactor: rewrite selection in typescript

### LINKED ISSUE #18575: [multiselect]: fix incorrectly typed props
The Downshift props `filteredItems` and `initialSelectedItems` are incorrectly typed and result in the following build error:

```
Type 'ItemType' is not assignable to type 'never'. ts(2322)
```

Affected files are:
- [FilterableMultiSelect.tsx](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx#L379)
- [MultiSelect.tsx](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/MultiSelect/MultiSelect.tsx#L435-L442)

Currently bypassed with `@ts-ignore` so doesn't interfere with build.

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18575

Rewrote Selection.js in TypeScript.

#### Changelog

**Changed**

- Rewrote Selection.js in TypeScript.
- Converted `Selection` from a `class` component to a functional component.

#### Testing / Reviewing

```sh
yarn test packages/react/src/components/MultiSelect packages/react/src/internal
```


packages/react/src/internal/Selection.tsx
```diff
@@ -0,0 +1,247 @@
...
+    onItemChange,
+    selectedItems,
+  };
+};
+
+interface SelectionRenderProps {
+  clearSelection: () => void;
+  onItemChange: (item: any) => void;
+  selectedItems: any[];
+}
```
preetibansalui, line 152: Hey, can we try to avoid any here?

```
interface SelectionRenderProps<ItemType> {
  clearSelection: () => void;
  onItemChange: (item: ItemType) => void;
  selectedItems: ItemType[];
}
```

packages/react/src/internal/Selection.tsx
```diff
@@ -0,0 +1,247 @@
+/**
+ * Copyright IBM Corp. 2016, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useCallback, useEffect, useState, useRef } from 'react';
+import PropTypes from 'prop-types';
```
preetibansalui: Can we remove this import if not using it?

preetibansalui (commented):
Thanks for the changes. It looks good to me and all the tests are passing successfully. I am just little concerned about use of any in various places in `Selection.tsx` file.

adamalston:
> I am just little concerned about use of any in various places in `Selection.tsx` file.

Some of the usages of `any` align with that's in the `MultiSelect` component:

https://github.com/carbon-design-system/carbon/blob/62b5c1e2d58f7ad4f45fc0563c860c350a520327/packages/react/src/components/MultiSelect/MultiSelect.tsx#L356

https://github.com/carbon-design-system/carbon/blob/62b5c1e2d58f7ad4f45fc0563c860c350a520327/packages/react/src/components/MultiSelect/MultiSelect.tsx#L468-L470

If you want to open an issue, I can investigate getting those removed too. Seems out of scope for the issue that this pull request is addressing, unless you disagree?

adamalston:
Also, would you or anyone else happen to know the answer to the `TODO` I left in Selection.tsx?

packages/react/src/internal/Selection.tsx
```diff
@@ -0,0 +1,242 @@
...
+      } else {
+        let selectedIndex: number | undefined;
+        selectedItems.forEach((selectedItem, index) => {
+          if (isEqual(selectedItem, item)) {
+            // TODO: Should this functionality be changed so that the loop stops
+            // iterating after finding a match? If not, should the loop be
+            // iterated in reverse? Either way, seems like there's a more
+            // performant way to do this.
+            selectedIndex = index;
+          }
```
tay1orjones: Yeah I'm not sure if it _needs_ to be the last selected index, but that's what the result of this ends up being. Feels like it could instead be this?
```suggestion
        selectedIndex = selectedItems.findLastIndex(selectedItem => isEqual(selectedItem, item));
```

tay1orjones (approved):
Thanks for this! Moving downshift-related _anything_ to ts is a very welcome change. Stability in those components is challenging and type safety will help 👍

packages/react/src/internal/Selection.tsx
adamalston: Yea, a solution with `findLastIndex` worked locally. I was able to convert the variable to a `const` which is better too.

kennylam (approved):

PR #18644, opened 2025-02-20
chore: overflow utility initial commit

### LINKED ISSUE #6883: Rewrite `useOverflowItems` utility as vanilla JS
Currently, `useOverflowItems` is written as a React hook, but to support the many acquisitions teams looking to use the TagSet, we’d want to rewrite this as a vanilla utility provided through `@carbon/utilities`.

### Acceptance criteria

- [ ] Rewrite `useOverflowItems` as framework agnostic
- [ ] Provide through `@carbon/utilities` package

devadula-nandan:
Closes [#6883](https://github.com/carbon-design-system/ibm-products/issues/6883) on ibm-products

adds framework agonistic overflow utility, to the utility package.

#### Changelog

**New**

- adds overflow utility to utilities package
- adds tests to it

#### Testing / Reviewing

automated tests written in the test file

existing ibm-products actionbar and tagset example
https://stackblitz.com/edit/github-yxe1kwrz?file=src%2FExample%2FExample.jsx

Iteration 1
[React usage](https://stackblitz.com/edit/github-nufrullm?file=src%2FExamples%2FExamples.tsx) example of this utility
[Lit usage](https://stackblitz.com/edit/vitejs-vite-hdurjrpy?file=src%2Ftagset.ts) example of this utility

iteration 2
[React usage](https://stackblitz.com/edit/github-nufrullm-mz2zx3zy?file=src%2FExamples%2FExamples.tsx) example of this utility
[Lit usage](https://stackblitz.com/edit/vitejs-vite-sksyz7me?file=src%2Ftagset.ts) example of this utility


devadula-nandan:
@makafsal to be working on typescript suppoort / configurations

packages/utilities/src/overflowHandler/overflowHandler.ts
```diff
@@ -0,0 +1,152 @@
...
+   */
+  items: HTMLElement[];
+  /**
+   * The size offset value used when calculating overflow.
+   */
+  offsetSize: number;
+  /**
+   * Maximum number of visible items. If provided, only this number of items will be shown.
+   */
+  maxVisible?: number;
```
tay1orjones:
```suggestion
  maxVisibleItems?: number;
```

tay1orjones (commented):
This is awesome! 🔥 

Some pieces of the react usage example have me wondering if the abstractions you have here could be modified to improve performance and usage. 

An entirely new `overflowHandler` is created anytime the `actions` change. `actions` may be unlikely to change, but it seems like we should be able to reuse the one we already have. The container isn't changing, none of the other "props" are changing, the existing resize observer doesn't need to be destroyed, etc. 

Also the amount of boilerplate stood out to me. It would be best if the consumer didn't have to bear that burden.

If the internal functions (getDimension, update, etc) were available by themselves in addition to createOverflowHandlder, we could make a useOverflowHandler hook for react users that uses these internally. It would streamlines the usage for consumers while also giving us a chance to make it more performant. We could memoize the actions and the container, contain the `hiddenItems` state, selectively provide return values if they were to need them, etc.

It might even be that we could avoid handling the `requestAnimationFrame` ourselves and use useLayoutEffect in the hook. 

You've maybe considered all this though 😅 if so, we can divert to a new issue. Either way I'm curious to hear what you think

devadula-nandan:
hey @tay1orjones , i like to do some more changes to this, to reduce the boilerplate, i see some potential for it, and yes, actionbar example doesnt need any memoization, i just copy pasted the example from tagset and forgot to clean up.
we are destroying and recreating the overflowHandler on tagset overflow tag width changes which usually happens  at number 9->10, 99->100.
i am also making use of data-hidden and data-offset and data-fixed attributes which we can use to do custom styling through pure css, and dropping onVisible and onHidden callbacks, we can still access them in onChange callback.

ill move this to draft and will open with new changes soon. thanks for the review, it enabled few more insights to this.

devadula-nandan:
Hi @tay1orjones ,

this is done, and open for re-review

iteration 2 change log

1. seperated internal functions `getSize` , `updateOverflowHandler` and exported (interfaces as well)
2. removed `onVisible`, `onHidden` callbacks, instead we are setting `data-hidden` attributes, which we use to style with plain css. we can still access hidden and visible items from onChange callback
3. added `data-offset` attribute for the offset element and removed `offsetSize` prop
4. removed items prop, since we can construct this array from children
5. introduced `data-fixed` attribute. any item that has data-fixed attribute, doesn't participate in any overflowing activity
6. added interface docs at most places
7. stackblitz examples updated

tay1orjones (approved):

2nikhiltom (approved):
LGTM 💯 
could we add Docs/README around this and usage examples to notify users?
https://github.com/carbon-design-system/ibm-products/issues/6716 has pointers for creating the docs and examples within Storybook.

PR #18658, opened 2025-02-21
fix: update icon button types and docs

### LINKED ISSUE #17956: [Bug]: `IconButton` `size` docs do not align with component usage
### Package

@carbon/react

### Browser

_No response_

### Package version

1.67.1

### React version

_No response_

### Description

The `IconButton` docs say that the button defaults to `md`. However, the linked reproduction shows that it defaults to `lg`.

https://react.carbondesignsystem.com/?path=/docs/components-iconbutton--overview:

> Specify the size of the Button. Defaults to `md`.
> `'sm'` `'md'` `'lg'`

Another issue is that the types for `IconButton` `size` list five options. However, only three are listed in the docs.

### Reproduction/example

https://stackblitz.com/edit/github-ai8zhx?file=src%2FApp.jsx

### Steps to reproduce

1. Render buttons in all three listed sizes. 
2. Render one button that has no `size` prop. 
3. See that the button without the `size` prop is not the same size as the `md` button.

### Suggested Severity

### Application/PAL

DataPower

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/17956

Updated `IconButton` types and documentation.

#### Changelog

**Changed**

- Updated `IconButton` types and documentation.
- Updated `Button` tests.

#### Testing / Reviewing

My understanding of the `Button` tests was that there was a bug in them. The test titles say 'should set the expected classes for the button of <prop>: `%s`'. Doesn't that mean if the prop is set on the `Button` that the code would apply the right class to it? If so, I'm not sure how these tests were ever going to fail considering the `className`s were explictly set in them.

```sh
yarn test packages/react/src/components/Button packages/react/src/components/IconButton
```

If the changes in this pull request are correct and the `IconButton` is supposed to be `lg` by default, does the following code need to be changed too?

https://github.com/carbon-design-system/carbon/blob/a60a4d3071663b240c3122d8bc2c3558c788346f/packages/web-components/src/components/icon-button/icon-button.ts#L64-L68

### DIFF
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -193,6 +193,14 @@ const Button: ButtonComponent = React.forwardRef(
       }
 
       return (
+        // @ts-expect-error - `IconButton` does not support all `size`s that
+        // `Button` supports.
+        //
+        // TODO: What should be done here?
+        // 1. Should the `IconButton` not be rendered if the `size` is not
+        //    supported?
+        // 2. Should an error be thrown?
+        // 3. Something else?
         <IconButton
           {...rest}
           ref={ref}
--- a/packages/react/src/components/Button/ButtonBase.tsx
+++ b/packages/react/src/components/Button/ButtonBase.tsx
@@ -45,6 +45,7 @@ const ButtonBase = React.forwardRef(function ButtonBase<
     [`${prefix}--btn`]: true,
     [`${prefix}--btn--sm`]: size === 'sm' && !isExpressive, // TODO: V12 - Remove this class
     [`${prefix}--btn--md`]: size === 'md' && !isExpressive, // TODO: V12 - Remove this class
+    [`${prefix}--btn--lg`]: size === 'lg' && !isExpressive, // TODO: V12 - Remove this class
     [`${prefix}--btn--xl`]: size === 'xl', // TODO: V12 - Remove this class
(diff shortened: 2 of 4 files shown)


kennylam:
> If the changes in this pull request are correct and the `IconButton` is supposed to be `lg` by default, does the following code need to be changed too?

@adamalston Thanks for this contribution! Yes, what you referenced would need to be changed as well. There's an open epic for React/web-component parity, i.e. updating web components to mirror React.

packages/react/src/components/Button/Button.tsx
```diff
@@ -193,6 +193,14 @@ const Button: ButtonComponent = React.forwardRef(
...
 
       return (
+        // @ts-expect-error - `IconButton` does not support all `size`s that
+        // `Button` supports.
+        //
+        // TODO: What should be done here?
+        // 1. Should the `IconButton` not be rendered if the `size` is not
+        //    supported?
+        // 2. Should an error be thrown?
+        // 3. Something else?
```
tay1orjones, line 203: > What should be done here?

To avoid a breaking change, nothing. To help encourage best practices we could update the typing to be a mix of what you have and what we had before. In other components we define a "deprecated" type and then union it with the non-deprecated types. `align` in Popover does this as an example

In the next major the deprecated ones would go away and the types would limit it to only the supported sizes.

tay1orjones (approved):

riddhybansal (approved):

PR #18585, opened 2025-02-13
fix(tabs): move react-is to peer dependencies

tay1orjones:
Closes #18574 

In #18207 we've found that the original issue reporting problems with Tabs and react 19 (#16620) was fixed simply by updating the react-is version to v19. We've opted to delay #18207 for now and instead just open up the react peer dependency ranges to include v19, https://github.com/carbon-design-system/carbon/pull/18458. 

The issue with tabs came up again because `react-is` was a hard dep and locked to v18. This PR updates `react-is` to be a peer dependency allowing both v18 and v19. I also added `react-is` to devDeps so that internally the project still uses v18.

#### Changelog

**Changed**

- moved react-is to peer dependencies
- expanded the ranges for allowed react-is versions

PR #18660, opened 2025-02-21
fix: TabList children rendering

remimarechal:
Affect react 19.0.0

isElement test on Tab component causing a regression on react 19.0.0
the TabList children are not rendered

### Changed

- Fixed not null child test in TabList

### Testing / Reviewing

`yarn test packages/react/src/components/Tabs`


remimarechal:

tay1orjones:
@remimarechal Thanks for opening this up, we found this issue can be resolved simply by using v19 of `react-is` when using react 19. We moved `react-is` to be a peer dep so you can do this in your environment. https://github.com/carbon-design-system/carbon/pull/18585 There's some linked issues there w/ related info too.

I'm going to close this one as I think the peer dep update should resolve the issue for you without having to add this additional condition. It went out yesterday in `@carbon/react@v1.77.0`

tay1orjones:
Let us know if the peer dep approach isn't working, we can reopen this PR or look for a more in depth solution. Thanks again for the PR!

remimarechal:
@tay1orjones sorry, i've missed the issue report and the associated fix, during my search
it works, tabs are visibles, thank you.

PR #18612, opened 2025-02-18
feat(Menu): integration of menu web component to carbon repo

amal-k-joy:
Contributes #18338 ,#17917

This PR is the cherry pick of [Sangeethas PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11740) from dotcom repo

#### Changelog

**New**

- Moved the changes of `Menu` component from dotCom to the folder carbon-webcomponents as part of cherry pick
- Copied the changes from `packages/carbon-webcomponents` folder to `packages/webcomponents` folder
- committed yarn cache for @lit/context addition 

#### Testing / Reviewing

The basic functionality is verified in local storybook

PR #18703, opened 2025-02-26
Menu : Integration of Menu webcomponent to carbon

### LINKED ISSUE #18338: React|WC Parity: Menu
This was initially started before WC was moved to the monorepo and just needs some final items addressed:

### Acceptance Criteria

- [ ] See parent issue for baseline criteria
- [ ] Upgrade the implementation to `lit@v3`
- [ ] Cherry pick the PR over from the old repo
- [ ] Update/remove `mode`, see https://github.com/carbon-design-system/carbon/pull/18153

> [!IMPORTANT]  

amal-k-joy:
Closes  #18338

 Changes from  [Sangeethas PR](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/11740) to dotcom repo with cherry picked every commits 
Adopt changes from [jans PR](https://github.com/carbon-design-system/carbon/pull/18153) for deprecating mode
Keyboard navigation fixes

**New**

- Moved the changes of `Menu` component from dotCom to the folder carbon-webcomponents as part of cherry pick
- Copied the changes from `packages/carbon-webcomponents` folder to `packages/webcomponents` folder
- committed yarn cache for @lit/context addition 
- mode deprecation changes and support to render icons
- breaking changes from reviews

#### Testing / Reviewing


tay1orjones (approved):

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * @element cds-menu-item
+ */
+@customElement(`${prefix}-menu-item`)
+class CDSmenuItem extends LitElement {
+  @consume({ context: MenuContext })
+  context;
+  // @provide({ context: MenuContext })
+  // @consume({ context: MenuContext, subscribe: true })
+  // contextChild;
```
annawen1: Can we clean up the commented out code if it's not needed?

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * Provide an optional function to be called when the MenuItem is clicked.
+   */
+  @property()
+  onClick?: (event: KeyboardEvent | MouseEvent) => void;
+
+  checkForIcon() {
+    if (this.renderIcon) {
+      //this.context.updateFromChild?.({ hasIcons: true },this);
+      // if(!this.context.hasIcons)
```
annawen1:
```suggestion

```

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+      this.parentElement?.shadowRoot
+        ?.querySelector(`.${prefix}--menu`)
+        ?.classList.add(`${prefix}--menu--with-icons`);
+    } else {
+      // this.hasIcon=false;
+    }
```
annawen1:
```suggestion
    } 
```

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+      isDanger,
+      boundaries,
+      childElements,
+      isRtl,
+    } = this;
+    const menuItemClasses = classMap({
+      [`${prefix}--menu-item`]: true,
+      [`${prefix}--menu-item--disabled`]: isDisabled,
+      [`${prefix}--menu-item--danger`]: isDanger,
+      // [`${prefix}--menu-item--with-icon`]:!this.hasIcon
```
annawen1:
```suggestion
```

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+        ${shortcut &&
+        !hasChildren &&
+        html` <div class="${prefix}--menu-item__shortcut">${shortcut}</div> `}
+        ${hasChildren
+          ? html`
+              <div class="${prefix}--menu-item__shortcut">
+                ${isRtl ? CaretLeft16() : CaretRight16()}
+              </div>
+              <cds-menu
+                .isChild="${hasChildren}"
```
annawen1: If `hasChildren` is a boolean we should be able to apply the attribute or not based on its value

```suggestion
                ?isChild="${hasChildren}"
```

packages/web-components/src/components/menu/menu.mdx
```diff
@@ -0,0 +1,178 @@
...
+  { weight: 600, label: 'Semibold' },
+  { weight: 700, label: 'Bold' },
+];
+
+<cds-menu-item-radio-group
+  label="Font weight"
+  items={items}
+  selectedItem={items[3]}
+  itemToString={(item) => item.label}
+/>;
```
annawen1:
```suggestion
<cds-menu-item-radio-group
  label="Font weight"
  items={items}
  selectedItem={items[3]}
  itemToString={(item) => item.label}></cds-menu-item-radio-group>;
```

packages/web-components/src/components/menu/menu.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { html } from 'lit';
+import './index';
+import FolderShared16 from '@carbon/icons/lib/folder--shared/16.js';
+import Cut16 from '@carbon/icons/lib/cut/16.js';
+import Copy16 from '@carbon/icons/lib/copy/16.js';
+//import TrashCan16 from '@carbon/icons/lib/trash-can/16.js';
```
annawen1:
```suggestion

```

packages/web-components/src/components/menu/menu.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+    },
+  },
+  args,
+  argTypes,
+  render: ({ open, size }) => {
+    const itemlist = ['None', 'Overline', 'Line-through', 'Underline'];
+    const subitemlist = ['None', 'Product team', 'Organization', 'Company'];
+    return html`
+      <cds-menu ?open=${open} size=${size}>
+        <cds-menu-item label="Share with" .renderIcon=${FolderShared16}>
```
annawen1: Instead of passing in the icon through the `renderIcon` property, can the icon be slotted in instead? Having to pass in the icon this way might be difficult for folks using CDN or are not using Lit in their own application. 

We can detect if icon has been passed in the slot using the `_handleSlotChange()` event (ex: https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/text-input/text-input.ts#L52-L75)

annawen1:
@amal-k-joy with the 2 other Menu PRs (https://github.com/carbon-design-system/carbon/pull/18718 and https://github.com/carbon-design-system/carbon/pull/18612), can you clarify which one we should be focusing on first?

packages/web-components/src/components/menu/menu.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+          <cds-menu-item-radio-group
+            label="Share with list"
+            .items="${subitemlist}"
```
annawen1: Same here, we should probably try to reorganize this as slotted items instead - having the adopters pass in the elements themselves

```
<cds-menu-item-radio-group>
  <cds-menu-item> .... </cds-menu-item>
  <cds-menu-item> .... </cds-menu-item>
  <cds-menu-item> .... </cds-menu-item>
  <cds-menu-item> .... </cds-menu-item>
</cds-menu-item-radio-group>
```

packages/web-components/src/components/menu/menu-item-divider.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @element cds-menu-item-divider
+ */
+@customElement(`${prefix}-menu-item-divider`)
+class CDSmenuItemDivider extends LitElement {
+  render() {
+    return html`
+      <li class="${prefix}--menu-item-divider" role="separator"></li>
+    `;
+  }
```
ariellalgilmore: In general we try to avoid having unessarry rendering of html if we can apply the attributes (ie class and role) to the the custom element itself. 

Here is an example:
https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/ui-shell/side-nav-divider.ts

We could additionally remove the `"${prefix}--menu-item-divider"` class since we can just extend the styles like so: https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/ui-shell/side-nav.scss#L47

When a adopter then uses `cds-menu-item-divider` then they can apply a class like:
```
<cds-menu-item-diviver class="custom-class" />
```
and then they would be able to apply custom styles to the same level as the `menu-item-divider` styles vs the way it's implemented now the adopter would have a custom class at a higher level then the `<li class="${prefix}--menu-item-divider"/>` level

This should be applied to all the components and not just this one

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * Checks if document direction is rtl.
+   */
+  @property({ type: Boolean })
+  isRtl = false;
+  /**
+   * Checks if document direction is rtl.
+   */
+  @property()
+  kind: 'default' | 'danger' = 'default';
```
ariellalgilmore: can we make a `defs` file for things like this? here's an example: https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/ui-shell/defs.ts

and double check the comments accurately describe the right property. might be a copy paste error

packages/web-components/src/components/menu/menu-item.ts
```diff
@@ -0,0 +1,339 @@
...
+        role="menuitem"
+        tabindex="0"
+        class="${menuItemClasses}"
+        aria-disabled="${isDisabled ?? undefined}"
+        aria-haspopup="${hasChildren ?? undefined}"
+        aria-expanded="${hasChildren ? submenuOpen : undefined}"
+        @click="${handleClick}"
+        @mouseenter="${hasChildren ? handleMouseEnter : undefined}"
+        @mouseleave="${hasChildren ? handleMouseLeave : undefined}"
+        @keydown="${handleKeyDown}">
```
annawen1: Instead of applying all these attributes to the `li` element, we can perhaps apply it to the `menu-item` component itself so we won't need the `li` element.

We have a `HostListener` mixin that we use to listen to events: https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/globals/mixins/host-listener.ts

Used in the `button` component like so: https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/button/button.ts#L63-L104

packages/web-components/src/components/menu/menu-item-selectable.ts
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Whether the menu item is selected or not.
+   */
+  @property({ type: Boolean })
+  selected = false;
+
+  /**
+   * Provide an optional function to be called when the selection state changes.
+   */
+  onChange?;
```
annawen1: We usually let adopters tap into event changes by emitting a custom event like so in `tag`:

https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/tag/tag.ts#L64-L97

On click of the close icon in `cds-tag`, a custom event is emitted (`eventClose` or `${prefix}-tag-closed`), which for the adopter they can listen for like so

```
document.addEventListener('cds-tag-closed', (event) => {
  // use values here
});
```

packages/web-components/src/components/menu/menu.mdx
```diff
@@ -0,0 +1,178 @@
+import { ArgTypes, Markdown, Meta } from '@storybook/blocks';
+import { cdnJs, cdnCss } from '../../globals/internal/storybook-cdn';
+import * as MenuStories from './menu.stories';
+
+<Meta of={MenuStories} /> 
+
+# Menu
+
+[Source code](https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/Menu)
```
ariellalgilmore:
```suggestion
[Source code](https://github.com/carbon-design-system/carbon/tree/main/packages/web-components/src/components/Menu)
```

packages/web-components/src/components/menu/menu-item-radio-group.ts
```diff
@@ -0,0 +1,97 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { LitElement, html } from 'lit';
+import { property } from 'lit/decorators.js';
+import { prefix } from '../../globals/settings';
+import styles from './menu-item.scss?lit';
+import { carbonElement as customElement } from '../../globals/decorators/carbon-element';
+import { ChangeEventHandler } from 'react';
```
annawen1: shouldn't have to import from react if we follow the comment I made below in `menu-item-selectable` about events

packages/web-components/src/components/menu/menu.mdx
```diff
@@ -0,0 +1,178 @@
...
+  selectedItem={items[3]}
+  itemToString={(item) => item.label}
+/>;
+```
+
+## Feedback
+
+Help us improve this component by providing feedback, asking questions on Slack,
+or updating this file on
+[GitHub](https://github.com/carbon-design-system/carbon/edit/main/packages/react/src/components/Menu/Menu.mdx).
```
ariellalgilmore:
```suggestion
[GitHub](https://github.com/carbon-design-system/carbon/edit/main/packages/web-components/src/components/Menu/Menu.mdx).
```

packages/web-components/src/components/menu/menu-item.ts
amal-k-joy: Committed to the [latest PR](https://github.com/carbon-design-system/carbon/pull/18718)

packages/web-components/src/components/menu/menu-item.ts
amal-k-joy: Fixed in the latest [PR](https://github.com/carbon-design-system/carbon/pull/18718)

packages/web-components/src/components/menu/menu-item.ts
amal-k-joy: Fixed in the latest [PR](https://github.com/carbon-design-system/carbon/pull/18718)

packages/web-components/src/components/menu/menu-item.ts
amal-k-joy: Fixed in the latest [PR](https://github.com/carbon-design-system/carbon/pull/18718)

packages/web-components/src/components/menu/menu.stories.ts
amal-k-joy: Fixed

packages/web-components/src/components/menu/menu.stories.ts
amal-k-joy: used slot to render the icon as well as submenu

packages/web-components/src/components/menu/menu.stories.ts
amal-k-joy: Hey, When i checked the [react implementation ](https://react.carbondesignsystem.com/?path=/docs/components-menu--overview&globals=theme:g10), they have accepted an array of strings as items instead of using children.

packages/web-components/src/components/menu/menu-item-divider.ts
amal-k-joy: Modified as advised

packages/web-components/src/components/menu/menu-item.ts
amal-k-joy: Modified as advised

packages/web-components/src/components/menu/menu-item.ts
amal-k-joy: Modified as advised

packages/web-components/src/components/menu/menu-item-selectable.ts
amal-k-joy: Modified as advised

packages/web-components/src/components/menu/menu.stories.ts
annawen1: This is one of the instances where React implementation and Web Components implementation would differ. We generally try to avoid having the component handle object / array data, reasons why explained here: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/web-components/docs/coding-conventions.md#limiting-components-that-works-with-complex-data

packages/web-components/src/components/menu/menu-item-divider.ts
```diff
@@ -0,0 +1,32 @@
...
+class CDSmenuItemDivider extends LitElement {
+  connectedCallback() {
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'separator');
+    }
+    super.connectedCallback();
+  }
+  render() {
+    return html` <li></li> `;
+  }
```
ariellalgilmore: i believe this shouldn't be needed nowo and the styles can just be applied to the host element
```suggestion
```

packages/web-components/src/components/menu/menu-item-group.ts
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Label for the menu item.
+   */
+  @property({ type: String })
+  label;
+
+  render() {
+    const { label } = this;
+    return html`
+      <li class="${prefix}--menu-item-group" role="none">
```
ariellalgilmore: is the `li` needed here or can we use the host element and apply `${prefix}--menu-item-group` styles to host as well?

packages/web-components/src/components/menu/menu-item-radio-group.ts
```diff
@@ -0,0 +1,117 @@
...
+    const defaultItemToString = (item) => item.toString();
+    const {
+      label,
+      items,
+      itemToString = defaultItemToString,
+      selectedItem,
+      _handleClick: handleClick,
+    } = this;
+    return html`
+      <li class="${prefix}--menu-item-radio-group" role="none">
```
ariellalgilmore: same as here

amal-k-joy: modified as advised

packages/web-components/src/components/menu/menu-item-group.ts
amal-k-joy: modified as advised

packages/web-components/src/components/menu/menu.stories.ts
amal-k-joy: I also felt the same thing, that we should allow user to pass their items rather than accepting an array we build items for them. But thinking deeper in this particular scenario, may be its worth to become more restrictive. since `cds-menu-item-radio-group` has to behave only like a radio group, and only one item can be selectable at a  time, that means the children of `cds-menu-item-radio-group` has to be only `cds-menu-item`s. 
But if we allow user to pass any children , this can cause issues in the behaviour,  for instance if they pass `cds-menu-item-selectable`, this act more like a check box which is weird.
May be this was the reason they opt to accept items as string array even in react.

amal-k-joy: I have updated the cds-menu-item-radio-group to accept children.

packages/web-components/src/components/menu/menu-item-divider.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2024
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { LitElement, html } from 'lit';
```
ariellalgilmore:
```suggestion
import { LitElement } from 'lit';
```

packages/web-components/src/components/menu/menu.ts
```diff
@@ -0,0 +1,516 @@
...
+      [`${prefix}--menu--open`]: open,
+      [`${prefix}--menu--shown`]: position[0] >= 0 && position[1] >= 0,
+      [`${prefix}--menu--with-selectable-items`]:
+        this.context.hasSelectableItems,
+    });
+    return html`
+      <ul
+        class="${menuClasses}"
+        aria-label="${label}"
+        tabindex="1"
```
ariellalgilmore: Double check the accessibility checker. When looking I saw this and compared to react and it looks like it should be set to "-1"
```suggestion
        tabindex="-1"
```

packages/web-components/src/components/menu/menu-item-selectable.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.context.updateFromChild({ hasSelectableItems: true });
+  }
+  render() {
+    const { label, selected, _handleClick: handleClick } = this;
+
+    return html`
+      <cds-menu-item
+        label="${label}"
+        class="${prefix}--menu-item-selectable--selected"
+        role="menuitemcheckbox"
```
ariellalgilmore, line 90: maybe not using the right component in the story? not seeing the right role here
Screenshot 2025-03-12 at 3 06 45 PM

ariellalgilmore (commented):
getting close! thanks for the updates! I would just recommend now going through the accessibility checker and making sure it aligns with Reacts output.

packages/web-components/src/components/menu/menu-item-selectable.ts
```diff
@@ -0,0 +1,102 @@
...
+        shortcut=${this.shortcut}
+        aria-checked="${selected}"
+        @click="${handleClick}">
+        <slot slot="render-icon" name="render-icon"></slot>
+
+        ></cds-menu-item
```
ariellalgilmore:
```suggestion
        </cds-menu-item
```

ariellalgilmore (approved):
LGTM

riddhybansal (approved):

laurenmrice (commented):
This visually looks good @amal-k-joy.

I had one question about a11y when it comes to navigating with focus to and within the secondary menu:
- In React, it seems that the whole secondary menu gets focus first by pressing Tab, and then you can Tab or Arrow down to move focus to individual menu items. 
- In WC, you Arrow right to move focus to individual menu items, and the secondary menu does not get focus around it as a whole. 

Do you know what the preferred a11y for focus for this would be? I would think it should be the same for WC and React?

Mar-18-2025 14-03-25

laurenmrice:
@amal-k-joy Okay thanks, if the current WC implementation for this is more desirable, should I open a separate issue for React, to make sure React has the same implementation for this as WC?

amal-k-joy:
> @amal-k-joy Okay thanks, if the current WC implementation for this is more desirable, should I open a separate issue for React, to make sure React has the same implementation for this as WC?

Hey @laurenmrice ,
I think it would be good to raise an issue in react that allows team to verify the behaviour during their triage . We can come back to WC based on their feedback if required.

laurenmrice (approved):

PR #18718, opened 2025-02-27
fix(Menu): fix extra gap issue when the menu doesnt contain any item with icon

amal-k-joy:
Contributes #18338 

We have cds menu as parent and menu items as children inside a slot. We need to apply a particular style when none of the items has rendered with an icon. If any of the items is rendered with icon, we need to reserve a blank space for all menu items in that parent menu, otherwise not. Styling from parent does not work as in react. 

Here I am dispatching an event from menu item when there is an icon  and listened by parent menu and class is added to the parent menu accordingly. 
The child items style is then handled using css variables.

#### Testing / Reviewing

Local storybook


packages/web-components/src/components/menu/menu-item.scss
```diff
@@ -0,0 +1,118 @@
...
+@use '@carbon/styles/scss/utilities' as *;
+@use '@carbon/styles/scss/components/button/tokens' as button-tokens;
+@use '@carbon/styles/scss/utilities/convert' as *;
+
+.#{$prefix}--menu-item {
+  @include type-style('body-short-01');
+
+  display: grid;
+  align-items: center;
+  block-size: 2rem;
```
heloiselui, line 25:
```suggestion
  block-size: $spacing-07;
```

packages/web-components/src/components/menu/menu.stories.ts
```diff
@@ -0,0 +1,151 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { html } from 'lit';
+import './index';
+import FolderShared16 from '@carbon/icons/lib/folder--shared/16.js';
+import Cut16 from '@carbon/icons/lib/cut/16.js';
+import Copy16 from '@carbon/icons/lib/copy/16.js';
+//import TrashCan16 from '@carbon/icons/lib/trash-can/16.js';
```
heloiselui: Should we clean up the commented out code if it's not needed?

heloiselui (commented):
Hi Amal, I believe there are some changes that need to be addressed in this PR: https://github.com/carbon-design-system/carbon/pull/18703. I’ll put this review on hold until the updates are completed.

PR #18732, opened 2025-02-28
feat: enable configuration of highcontrast and dropshadow props

### LINKED ISSUE #11347: [Bug]: theme neutral color token for popover bg when overlapping zoned theming

alisonjoseph:
Closes #11347

Related PR for IconButton https://github.com/carbon-design-system/carbon/pull/18636

Exposes tooltip props to Button and HeaderGlobalAction, there are cases where the Header using a different theme than the main body content so the default tooltip theme color will blend into the background. 

#### Changelog

**New**

- add `tooltipDropShadow` and `tooltipHighContrast` prop to `Button`
- add `tooltipDropShadow` and `tooltipHighContrast` prop to `HeaderGlobalAction`

#### Testing / Reviewing

Check test story for UI Shell header - tooltips should render correctly.
https://deploy-preview-18732--v11-carbon-react.netlify.app/?path=/story/components-ui-shell-header--test

- [ ] Remove test story before merging


kennylam (approved):

2nikhiltom (approved):
Just the test story needs to be removed

PR #18736, opened 2025-03-01
ci: use .nvmrc for node versioning in workflows

### LINKED ISSUE #18735: Use .nvmrc for Node versioning in workflows
This change should simplify the process of using Node in workflows by making it so that there is a single source of truth for the version that should be used, .nvmrc. For the next Node upgrade, all that should need to be done to change the version of Node used in CI is to update .nvmrc.

https://github.com/actions/setup-node#usage

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18735

Updated workflows to use .nvmrc for Node versioning.

#### Changelog

**Changed**

- Updated workflows to use .nvmrc for Node versioning.

#### Testing / Reviewing

I added a `TODO` that I'm curious to hear thoughts on.

### DIFF
--- a/.github/workflows/publish-web-components-cdn.yml
+++ b/.github/workflows/publish-web-components-cdn.yml
@@ -21,12 +21,10 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: '0'
-      - name: Use Node.js 18.x
-        uses: actions/setup-node@v4
+      - name: Use Node.js version from .nvmrc
+        uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a
         with:
-          # TODO: Should this Node.js version be upgraded?
-          node-version: '18.x'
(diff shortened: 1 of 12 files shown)


.github/workflows/publish-web-components-cdn.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       - name: Use Node.js 18.x
         uses: actions/setup-node@v4
         with:
+          # TODO: Should this Node.js version be upgraded?
           node-version: '18.x'
           cache: 'yarn'
       - name: Install dependencies
```
kennylam:
```suggestion
      - name: Use Node.js version from .nvmrc
        uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a
        with:
          node-version-file: '.nvmrc'
      - name: Install dependencies
```

kennylam (commented):
You're right, this should be updated as well. I think we can scrap `cache` as well. Thanks!

kennylam (approved):

ariellalgilmore (approved):

PR #18737, opened 2025-03-02
fix(ComboBox): update Enter key handling

### LINKED ISSUE #18728: [Bug]: ComboBox `onChange` yields `null` `selectedItem` when pressing Enter
### Package

@carbon/react

### Browser

_No response_

### Package version

1.77.0

### React version

_No response_

### Description

ComboBox `onChange` yields `null` `selectedItem` when pressing Enter.

This behavior appears to be a regression, possibly introduced in https://github.com/carbon-design-system/carbon/pull/18584. Everything seemed to work fine in 1.76.0 before updating to 1.77.0.

### Reproduction/example

https://stackblitz.com/edit/github-8wkviptr?file=src%2FApp.tsx

### Steps to reproduce

First issue:

1. Click the `ComboBox`.
2. Type <kbd>Option 1</kbd>.
3. Press Enter.
4. See that the field is marked as invalid.
5. Press Backspace.
6. Type <kbd>1</kbd>.
7. Press Enter.
8. See that the field is not marked as invalid even though the value matches the previous invalid value.

Second issue:

1. Click the `ComboBox`.
2. Type <kbd>Option 2</kbd>.
3. Press Enter.
4. See that the selected item is 'Option 2'.
5. Press Backspace.
6. Press Enter.
8. See that the selected item is 'Option ' which isn't a valid option.

Logging `selectedItem` in the `onChange` shows that `null` is being returned which isn't right because Enter should be selecting a value in the field.

### Suggested Severity

### Application/PAL

DataPower

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18728

Fixed `ComboBox` Enter key handling.

#### Changelog

**Changed**

- Fixed `ComboBox` Enter key handling.

#### Testing / Reviewing

```sh
yarn test packages/react/src/components/ComboBox
```

A test story can be created by copying and pasting this code in ComboBox.stories.js:

```js
export const _test = (args) => {
  const [invalidText, setInvalidText] = useState('');

  return (
    <div style={{ width: 300 }}>
      <ComboBox
        helperText="Combobox helper text"
        id="carbon-combobox"
        itemToString={(item) => item?.text ?? ''}
        invalid={!!invalidText}
        invalidText={invalidText}
        items={[
          { id: 'option-1', text: 'Option 1' },
          { id: 'option-2', text: 'Option 2' },
          { id: 'option-3', text: 'Option 3' },
        ]}
        onChange={({ selectedItem }) => {
          setInvalidText(selectedItem?.id === 'option-1' ? 'Invalid' : '');
        }}
        onToggleClick={() => {}}
        titleText="ComboBox title"
      />
    </div>
  );
};
```

Reference commits: https://github.com/carbon-design-system/carbon/compare/448a2b7c2419d970926d77f878fd453e7cd56c54...3fbba63546e19131f940c47c17eeb2c53dca0de8


heloiselui (approved):
Looks good Adam!

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -50,6 +50,8 @@ import { autoUpdate, flip, hide, useFloating } from '@floating-ui/react';
 import { TranslateWithId } from '../../types/common';
 import { useFeatureFlag } from '../FeatureFlags';
 
+// TODO: Is there supposed to be handling for `InputKeyDownEscape` in this
+// component?
```
tay1orjones: I don't think so? Escape is handled here:

https://github.com/carbon-design-system/carbon/blob/381edb24bc38a243f0a1997da0274c85bff68ab8/packages/react/src/components/ComboBox/ComboBox.tsx#L1023-L1029

Usually that means the default downshift behavior isn't what we want and so it's handled on keydown before downshift gets a chance to run it's callbacks/behavior. You could look into it though, it might be possible to reduce or simplify the code by handling it in the state manager.

tay1orjones (approved):
LGTM once the comment and test story is removed

packages/react/src/components/ComboBox/ComboBox.tsx
adamalston: Noted. The reason I asked is that the Escape behavior doesn't match the accessibility documentation.

tay1orjones:
@heloiselui could you approve percy if needed to merge?

PR #18740, opened 2025-03-03
fix: Select component on hover shows invalid title/tooltip

### LINKED ISSUE #15752: [Bug]: Select component on hover shows invalid title/tooltip

Bhas-kar:
Closes https://github.com/carbon-design-system/carbon/issues/15752, https://github.com/carbon-design-system/carbon/issues/18769

Fixed the invalid title/no title in Select component on hovering

#### Changelog

**New**

- The default title for the component was not being set correctly, hence used `useState` to set the title by the default or the `selectedItem` value. onChange updates these title values.

**Changed**

To address the issue, the fix implements the following logic.
- `Prioritised Default Value`: The code now first attempts to find a child option whose value matches the defaultValue prop. If found, this option's value is used as the default title.
- `Fallback to Selected/First Option`: If no matching defaultValue is found, it falls back to using the value which is already selected, essentially the first in the valueSet, as the default title. This ensures that a title is always set.

**Removed**

- None

#### Testing / Reviewing

- Hover on Select component, it should show the selected Option as tooltip

- If there's no option selected, tooltip should not show


Bhas-kar:

Bhas-kar:
recheck

Bhas-kar:

Bhas-kar:
recheck

Bhas-kar:
Hi @2nikhiltom ,
I have fixed the issue now and the build is successful 
image
image

tay1orjones:
Just a heads up that the DCO is failing because some of the commits weren't made with the username and email that matches your github profile, so the DCO sees this as a different "person" that hasn't signed the DCO. 

We can override it and merge but something to know for next time.

packages/react/src/components/Select/Select.tsx
```diff
@@ -169,9 +167,27 @@ const Select = React.forwardRef(function Select(
...
+    React.isValidElement(child)
+  );
+
+  // Initialize the title with the `text` prop
+  const [title, setTitle] = useState(
+    (React.isValidElement(defaultOption) && defaultOption.props?.text) ||
+      (React.isValidElement(firstOptionInSet) &&
+        firstOptionInSet.props?.text) ||
+      ''
+  );
```
tay1orjones: Reducing the duplication and simplifying the logic a bit:

```suggestion
  // Convert children to an array of valid elements once
  const validChildren = React.Children.toArray(children).filter(React.isValidElement);
  // Find the default option based on the specified defaultValue
  const defaultOption = validChildren.find(
    (child) => child.props?.value === other?.defaultValue
  );
  // Use the default option's text if available; otherwise, fallback to the first option's text
  const initialTitle = defaultOption?.props?.text || validChildren[0]?.props?.text || '';
  
  const [title, setTitle] = useState(initialTitle);
```

What do you think?

Bhas-kar:
Hey @tay1orjones ,
I have added the suggested code with few changes to validate the React child element too.
requesting to re-review.
thanks

packages/react/src/components/Select/Select.tsx
```diff
@@ -169,9 +167,18 @@ const Select = React.forwardRef(function Select(
...
   const selectInstanceId = useId();
 
+  // Convert children to an array of valid elements once
+  const validChildren = React.Children.toArray(children).filter(React.isValidElement);
+  // Find the default option based on the specified defaultValue
+  const defaultOption = validChildren.find(
+    (child) => child.props?.value === other?.defaultValue
+  );
+  // Use the default option's text if available; otherwise, fallback to the first option's text
+  const initialTitle = defaultOption?.props?.text || validChildren[0]?.props?.text || '';
```
2nikhiltom: The TypeScript error Property 'text' does not exist on type '{}' is occurring because TypeScript doesn't recognize the text property on the props of child elements. 

```suggestion
 // Define an interface for the expected props
interface SelectItemProps {
  value: any;
  text: string;
}

// Use type narrowing
const validChildren = React.Children.toArray(children)
  .filter((child): child is React.ReactElement<SelectItemProps> => 
    React.isValidElement<SelectItemProps>(child));

// Now TypeScript knows the shape of each child's props
const defaultOption = validChildren.find(
  (child) => child.props.value === other?.defaultValue
);
/ Use the default option's text if available; otherwise, fallback to the first option's text
  const initialTitle =
    defaultOption?.props?.text || validChildren[0]?.props?.text || '';
```

What are your views ?

Bhas-kar: included this with minor changes

Bhas-kar: the code has been changed to validate the child element and build error got resolved

Bhas-kar: UPDATE: addressed the said issue, pushed the suggested change.

2nikhiltom (approved):
Looks great! 🎉 .Thanks for the prompt follow-up.

tay1orjones (approved):

PR #18765, opened 2025-03-05
fix: selectableTile onChange event giving old selected value

### LINKED ISSUE #18478: [Bug]: SelectableTile onChange method is presented with stale 'selected' value
### Package

@carbon/react

### Browser

Chrome

### Package version

1.74.0

### React version

v18.3.1

### Description

When the SelectedTile component is in the 'unselected' state and we trigger a change (via click or keyboard), the onChange method is presented with a `selected` value that is stale. In this case, it is presented with `false.` The UI redraws correctly and properly updates all the data as far as I can tell.

The defect stems from calling `onChange` after calling `setIsSelected` and assuming that `setIsSelected` will impact the value of `isSelected` during this call. It will not, it will only drive a re-render.
https://react.dev/learn/state-as-a-snapshot#state-over-time

A suggested fix would be to use the same value that is passed to `setIsSelected` to also be passed to `onChange`

### Reproduction/example

https://stackblitz.com/edit/github-grcad4ps?file=src%2FApp.jsx

### Steps to reproduce

Access the example provided and click the selectable tile in the preview area. Observe the incorrect value presented to the method.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

preetibansalui:
Closes #18478 

SelectableTile onChange event was giving old selected value.

#### Changelog

Pass the latest selected value to onChange.

#### Testing / Reviewing

Go to Storybook > Tile > Selectable
Select/unselect the tile and check console for correct value.

PS - I will remove the console value before merge. Added it for testing purpose only.


riddhybansal (changes requested):
LGTM! Can a test case be added to verify the onChange handler receives the correct updated state value? And please remember to remove the console.log from the Storybook example before merging.

guidari (approved):
LGTM! 🚀

heloiselui (approved):
Looks good 👍🏽

riddhybansal (approved):

PR #18788, opened 2025-03-06
feat(pageheader): initial components, exports, stories, styles, tests

### LINKED ISSUE #18649: PageHeader: Research, add implementation scaffolding (storybook, blank stories, initial empty zone components)
- [ ] Create experimental zone components PageHeader, PageHeaderBreadcrumbBar, PageHeaderContent, PageHeaderTabBar. Render empty content divs initially
- [ ] Set up stories for each
  - [ ] put these in a whole new section in storybook - "Patterns (experimental)" (maybe "blocks", but we'll have to build consensus around that
  - [ ] Set up empty testing file(s) - unit, vrt, avt

tay1orjones:
Closes #18649

This adds the initial scaffolding needed to begin work on the PageHeader.

#### Changelog

**New**

- unstable__PageHeader component files, test files, story file, documentation, style files

#### Testing / Reviewing

- Verify the stories work
- CI/tests should pass

@kennylam @ariellalgilmore @annawen1 @riddhybansal To facilitate the component namespacing I moved all the components into one file. This will reduce boilerplate significantly as we won't have to import shared modules (like an eventual context we'll need every component to use) into a ton of different files. Same thing with the tests being in a single file. Also the components are exported under both their "direct"/"real" names as well as the namespaced ones for convenience, and theres a test and story demonstrating that.

--- 


tay1orjones:
An interesting related read: https://ivicabatinic.from.hr/posts/multipart-namespace-components-addressing-rsc-and-dot-notation-issues

tay1orjones:
@2nikhiltom it looks like this is failing to generate a new inline snapshot due to an error about a method not existing in prettier. This is something you and I worked on when upgrading to prettier 3, when we added the prettier2 alias in `@carbon/react`.

Jest is installed in the workspace root though, so it's importing Prettier (v3) and using a method that no longer exists.

One way around this is to upgrade to Jest 30. It's in alpha though. Do you have any ideas or know of another way we could tackle this?

packages/react/.storybook/main.js
```diff
@@ -151,6 +151,7 @@ const config = {
                 path.resolve(__dirname, '..', 'node_modules'),
                 path.resolve(__dirname, '..', '..', '..', 'node_modules'),
               ],
+              // quietDeps: true,
```
heloiselui, line 154: Just a question, are we keeping this line commented out for now to catch possible issues?

heloiselui (commented):
Just the CI is not passing, but looks really good!

packages/react/.storybook/main.js
tay1orjones: Yeah it silences the sass deprecation warnings. It should remain commented out so we see the warnings in CI but we can temporarily enable it locally if we'd like to cut down on the warnings during development.

ariellalgilmore (approved):
LGTM!

heloiselui (approved):

riddhybansal (approved):

packages/react/src/components/PageHeader/PageHeader.stories.js
```diff
@@ -0,0 +1,71 @@
...
+export default {
+  title: 'Patterns/unstable__PageHeader',
+  component: PageHeader,
+  subcomponents: {
+    PageHeaderBreadcrumbBar,
+    PageHeaderContent,
+    PageHeaderTabBar,
+  },
+  // uncomment includeStories before merging so the stories aren't visible in prod
+  // includeStories: [],
```
ariellalgilmore: looks like we just need to uncomment this and then we can merge

PR #18796, opened 2025-03-09
refactor: update useMergedRefs and delete useMergeRefs usage

### LINKED ISSUE #17234: Evaluate and consolidate `useMergedRefs` helpers

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/17234

**Short description**

Updated `useMergedRefs` types and deleted `useMergeRefs` usage from `@floating-ui/react`.

#### Changelog

**Changed**

- Updated `useMergedRefs` types.

**Removed**

- Deleted `useMergeRefs` usage from `@floating-ui/react`.

#### Testing / Reviewing

```
yarn test packages/react
```


tay1orjones (commented):
Hey, I'm curious what you think about this - my only concern is that they just dropped support for react 16, so while this might work w/ 16 today it could unexpectedly break in the future. We don't test against react 16, and usage is the lowest between v16-18.

adamalston:
That's a good point. I guess our only real option if we're not going to pin `@floating-ui/react` is to use `useMergedRefs`, unless you had something else in mind? If not, I can revert all of the code changes. On this point from https://github.com/carbon-design-system/carbon/issues/17234#issue-2478444659:

> Would we prefer to instead remain using our own and add an eslint rule to help enforce not using the floating-ui one?

I could add the following ESLint rule but it won't do anything given ESLint is disabled which I mentioned in Slack a few weeks ago:

```mjs
  {
    rules: {
      'no-restricted-imports': [
        'error',
        {
          paths: [
            {
              name: '@floating-ui/react',
              importNames: ['useMergeRefs'],
              message:
                'Do not use `useMergeRefs` from `@floating-ui/react`. Use `useMergedRefs` from useMergedRefs.ts instead.',
            },
          ],
        },
      ],
    },
  },
```

Any thoughts?

tay1orjones:
Yeah, I don't have any other ideas. I suppose we just stick to what we have for now unless you had some ideas on how to improve it now. It's been a while since I looked at how floating-ui and others implement their ref merging helpers. It might be quite similar, I know they used to be.

adamalston:
After https://github.com/carbon-design-system/carbon/pull/18611 is merged, I will investigate updating `useMergedRefs`.

adamalston:
I have updated the pull request title and description to reflect the latest changes.

adamalston:
I updated the pull request title and description to reflect the latest changes.

Gururajj77 (approved):
Looks good!

kennylam (approved):

PR #18810, opened 2025-03-11
feat(grid): add polymorphicProps in CSSGrid and FlexGrid

### LINKED ISSUE #18429: Adding strong PolymorphicProps to Grid
Here is a PR https://github.com/carbon-design-system/carbon/pull/18274 that serves as an example on how to implement this.

heloiselui:
**Closes #18429**  

Implementing polymorphic for Grid component by updating `CSSGrid` and `FlexGrid`.

**Changelog**

**New**  
- Added support for rendering `CSSGrid` and `FlexGrid` as different elements using the `as` prop.  
- Now both components support refs properly with `React.forwardRef`.  
- Improved TypeScript types to handle polymorphic behavior better.  

**Changed**  
- Updated `CSSGrid` and `FlexGrid` to use `PolymorphicComponentPropWithRef`.  
- Refactored components to use a dynamic `BaseComponent` instead of fixed elements.  

#### Testing / Reviewing  

- CI should pass.
- Make sure `CSSGrid` and `FlexGrid` still work as expected.  
<details open >
<summary>Test polymorphic capabilities:</summary>

image
</details>

>[!Note]
>Test if they can render as different elements using the as prop. A test story was added just for review purposes and will be removed after the PR is reviewed.


packages/react/src/components/Grid/FlexGrid.tsx
```diff
@@ -11,40 +11,41 @@ import React from 'react';
...
+    );
+  }
+);
 
 FlexGrid.propTypes = {
-  /**
-   * Provide a custom element to render instead of the default <div>
-   */
-  as: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
-
```
tay1orjones: We have to keep proptypes in for now for the ~40% of users who use javascript and aren't using typescript.

packages/react/src/components/Grid/CSSGrid.tsx
```diff
@@ -9,74 +9,78 @@ import cx from 'classnames';
...
   /**
    * Specify grid alignment. Default is center
    */
   align: PropTypes.oneOf(['start', 'center', 'end']),
 
-  /**
-   * Provide a custom element to render instead of the default <div>
-   */
-  as: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
-
```
tay1orjones: > We have to keep proptypes in for now for the ~40% of users who use javascript and aren't using typescript.

same here

tay1orjones (approved):
I don't want to hold this up because I'll be out all next week but I think this looks okay to merge after adding the `as` proptypes back in

riddhybansal (approved):
LGTM !! Just remove test story before merging

2nikhiltom (approved):

PR #18829, opened 2025-03-14
fix(icon-button): ensure aria-pressed is set according to isSelected

### LINKED ISSUE #18824: `aria-pressed` is no longer emitted on icon-only ghost buttons
### Package

@carbon/react

### Browser

Safari

### Package version

1.78.0

### React version

18.3.1

### Description

When using icon-only ghost buttons (either by using `<Button hasIconOnly>` or `<IconButton>`) with `props.isSelected`, the required aria attribute `aria-pressed` is not added to button element. I believe this is a regression but don't know when exactly it broke.

### Reproduction/example

https://stackblitz.com/edit/github-aaidzqdo?file=src%2FApp.jsx

### Steps to reproduce

Inspect the buttons from the reproduction above. Neither have `aria-pressed="true|false"` in their attributes.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

janhassel:
Closes #18824 

#### Changelog

**New**

- Forwarded `props.isSelected` from `IconButton` to `ButtonBase`
  - Added tests to verify `props.isSelected` properly controls `aria-pressed` if `kind="ghost"`

**Changed**

- Removed manual `--icon-only` classname from `IconButton` in favor of providing `props.hasIconOnly` to `ButtonBase` which in turn already adds that classname.
  - Updated DataTable snapshots as this affected the order of classnames for toolbar actions
- Spread all storybook controls to "IconButton/Default" story

#### Testing / Reviewing

- Automated tests
- Inspect the "IconButton/Default" and "Button/IconButton" stories with the following configurations:

| `kind` | `isSelected` | Expected `aria-pressed` on the `<button>` element |
| :- | :- | :- |
| `primary` | `undefined` | (not present) |
| `ghost` | `undefined` | (not present) |
| `ghost` | `false` | `"false"` |
| `ghost` | `true` | `"true"` |

### DIFF
--- a/packages/react/src/components/Button/ButtonBase.tsx
+++ b/packages/react/src/components/Button/ButtonBase.tsx
@@ -51,8 +51,7 @@ const ButtonBase = React.forwardRef(function ButtonBase<
     [`${prefix}--btn--${kind}`]: kind,
     [`${prefix}--btn--disabled`]: disabled,
     [`${prefix}--btn--expressive`]: isExpressive,
-    [`${prefix}--btn--icon-only`]:
-      hasIconOnly && !className?.includes(`${prefix}--btn--icon-only`),
+    [`${prefix}--btn--icon-only`]: hasIconOnly,
     [`${prefix}--btn--selected`]: hasIconOnly && isSelected && kind === 'ghost',
   });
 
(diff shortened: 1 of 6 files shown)


tay1orjones (approved):
LGTM 👍

riddhybansal (approved):

PR #18846, opened 2025-03-18
feat: added polymorphic props support for column

### LINKED ISSUE #18422: Adding strong PolymorphicProps to Column
Here is a PR https://github.com/carbon-design-system/carbon/pull/18274 that serves as an example on how to implement this.

Gururajj77:
Closes #18422 

added polymorphic props support for `Column`

#### Changelog

**New**

- Used `PolymorphicComponentPropWithRef` to properly type component props, allowing for type-safe prop checking based on the chosen element type.
- Implemented the generic `<T extends ElementType>` approach to maintain proper typing across different element types
- Created a proper `ColumnComponentType` interface that includes both the function signature and static properties, fixing the TypeScript error with PropTypes

#### Testing / Reviewing

- Text component should work as expected
- It should not accept wrong attributes


heloiselui (approved):
LGTM!

riddhybansal (approved):
Good to go !!

packages/react/src/components/Grid/Column.tsx
```diff
@@ -138,13 +146,12 @@ function Column<T extends React.ElementType>({
   });
 
   // cast as any to let TypeScript allow passing in attributes to base component
```
tay1orjones:
```suggestion
```

tay1orjones (approved):

PR #18878, opened 2025-03-20
feat(overflow-menu): add custom event to item click

### LINKED ISSUE #18487: [Feature Request]: Overflow Menu Item Click Event
### The problem

I am experiencing a problem when trying to capture the event where an overflow menu item is clicked. I would expect there would be a custom event emitted or similar when the menu item is clicked. 

There does not appear to be any example that involves emitting or capturing an overflow menu item click event. In my case, I can't just use the onclick attribute because I want to call a custom function without exposing it globally. I can use a document level event listener, but I was hoping there was a more elegant or native way to capture this event.

### The solution

I am aware that I may be missing something fundamental here or misunderstanding the expected functionality of the carbon design system. I would love to be corrected in my understanding of how to use the overflow menu item component.  However, my expectation is that there would be some kind of custom event fired or really *anything* to hook into to support custom functionality for when this element is clicked.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

kennylam:
Closes #18487 

Adds custom event when overflow menu item is clicked.

#### Changelog

**New**

- add `cds-overflow-menu-item-clicked` custom event to overflow menu item click event.

**Changed**

- add `@element` JSDoc to `cds-tag`

#### Testing / Reviewing

Add a simple code snippet to `overflow-menu.stories.ts` to see the click event fire.
```javascript
document.addEventListener('cds-overflow-menu-item-clicked', (event) => {
  const { currentTarget } = event;
  const { detail } = event as CustomEvent;
  const { evt } = detail;
  console.log('item clicked', currentTarget, evt);
});
```

### DIFF
--- a/packages/web-components/src/components/overflow-menu/overflow-menu-item.ts
+++ b/packages/web-components/src/components/overflow-menu/overflow-menu-item.ts
@@ -19,9 +19,28 @@ import styles from './overflow-menu.scss?lit';
  * Overflow menu item.
  *
  * @element cds-overflow-menu-item
+ * @fires cds-overflow-menu-item-clicked - The custom event fired when an overflow menu item is clicked.
  */
 @customElement(`${prefix}-overflow-menu-item`)
 class CDSOverflowMenuItem extends FocusMixin(LitElement) {
+  /**
+   * Handles `click` event on this element.
+   */
+  private _handleClick(event: Event) {
+    this.dispatchEvent(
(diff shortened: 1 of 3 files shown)


annawen1 (approved):
LGTM! I see the event firing in the console 👍

riddhybansal (approved):

PR #18884, opened 2025-03-21
feat(button-set): add stacked property to web component `button-set`

kennylam:
Adds `stacked` feature to the `button-set` component.

#### Changelog

**New**

- add `stacked` property to `button-set` to allow vertical layout of buttons

#### Testing / Reviewing

Go to Button story in deploy preview, `Set of Buttons` variant and toggle `stacked` control.

### DIFF
--- a/packages/web-components/src/components/button/button-set.ts
+++ b/packages/web-components/src/components/button/button-set.ts
@@ -7,7 +7,9 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { classMap } from 'lit/directives/class-map.js';
 import { html } from 'lit';
+import { property } from 'lit/decorators.js';
 import { BUTTON_KIND } from './defs';
 import styles from './button.scss?lit';
 import { prefix } from '../../globals/settings';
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/button/button.stories.ts
```diff
@@ -372,11 +377,12 @@ export const SetOfButtons = {
     isSelected,
     linkRole,
     size,
+    stacked,
     tooltipAlignment,
     tooltipPosition,
     type,
   }) =>
-    html` <cds-button-set>
+    html` <cds-button-set .stacked="${stacked}">
```
ariellalgilmore, line 385:
```suggestion
    html` <cds-button-set ?stacked="${stacked}">
```
i don't think the . is necessary here

packages/web-components/src/components/button/button-set.ts
```diff
@@ -21,6 +23,12 @@ import { carbonElement as customElement } from '../../globals/decorators/carbon-
  */
 @customElement(`${prefix}-button-set`)
 class CDSButtonSet extends CDSButtonSetBase {
+  /**
+   * `true` if the button should have input focus when the page loads.
```
ariellalgilmore, line 27:
```suggestion
   * `true` if the Button should be stacked, or not. Only applies to the button-set variant.

packages/web-components/src/components/button/button.stories.ts
kennylam: Ordinarily not, but that will get pulled into the `Canvas` source panel (in the overview) as an unrendered prop and it might be confusing to people.

ariellalgilmore (approved):

annawen1 (approved):

Issue #10896, opened 2022-03-01
[Bug]: useMatchMedia hook add Safari <14 version support


1mku:
### Package

carbon-components-react

### Browser

Safari

### Package version

v7.52.0

### React version

v17.0.2

### Description

After update `carbon-components-react` to v7.52.0
useMatchMedia hook which is used in SideNav component not supports Safari <14 versions

![Screenshot 2022-03-01 at 23 07 32](https://user-images.githubusercontent.com/20165886/156258945-83074890-0ab5-45e8-a252-e03c2a87f2f1.png)

useMatchMedia hook were added here (https://github.com/carbon-design-system/carbon/pull/10087) 

_Before Safari 14, MediaQueryList is based on EventTarget, so you must use addListener() and removeListener() to observe media query lists. (https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener)_

Maybe it will be useful (https://www.designcise.com/web/tutorial/how-to-fix-the-javascript-typeerror-matchmedia-addeventlistener-is-not-a-function)

### CodeSandbox example

https://codesandbox.io/s/weathered-violet-u0cx7h

### Steps to reproduce

Open in Safari version <14 storybook example of SideNav component
https://react.carbondesignsystem.com/?path=/story/components-ui-shell--side-nav-rail

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

PR #10897, opened 2022-03-02
fix(react): update useMatchMedia to fallback to addListener

joshblack:
Closes #10896 

Update `useMatchMedia` to fallback to `addListener` for older browser support

#### Changelog

**New**

**Changed**

- Update `useMatchMedia` to fallback to `addListener` for older browser support

**Removed**

#### Testing / Reviewing

- Verify our SideNav behavior works in older versions of Safari (<v14)

PR #18890, opened 2025-03-22
refactor: rewrite useMatchMedia in typescript

### LINKED ISSUE #18889: Rewrite `useMatchMedia` in TypeScript

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18889

Ported `useMatchMedia` to TypeScript.

#### Changelog

**Changed**

- Ported `useMatchMedia` to TypeScript.

#### Testing / Reviewing

```sh
yarn test packages/react
```

Curious if anyone has an answer to the question in the `TODO`. That fallback was added more than 3 years ago to address https://github.com/carbon-design-system/carbon/issues/10896 and it appears to have been targetted for Carbon v10. Further, there is zero usage of those methods outside of this one hook in the codebase.


packages/react/src/internal/useMatchMedia.ts
```diff
@@ -8,39 +8,46 @@
...
     setMatches(mediaQueryList.matches);
 
     return () => {
-      if (mediaQueryList.addEventListener) {
+      if (mediaQueryList.removeEventListener) {
         mediaQueryList.removeEventListener('change', listener);
       } else {
         mediaQueryList.removeListener(listener);
       }
     };
```
riddhybansal: Hey This looks good to me just a question!!

What do you think about the change from checking mediaQueryList.addEventListener to mediaQueryList.removeEventListener in the cleanup function? While it may be unusual, browsers could potentially implement these methods inconsistently, which might lead to memory leaks. Would it make more sense to maintain the original approach for consistency, ensuring we use the same API for both adding and removing listeners?

riddhybansal (changes requested):

packages/react/src/internal/useMatchMedia.ts
adamalston: I changed the condition from `mediaQueryList.addEventListener` to `mediaQueryList.removeEventListener` because it seemed to me that the code should check for the function it is about to execute (`mediaQueryList.removeEventListener`), rather than its counterpart (`mediaQueryList.addEventListener`).

In theory, yes, if a browser were to implement `addEventListener` and `removeEventListener` inconsistently, it could lead to problems such as memory leaks and unexpected event behavior. These issues would severely impact both performance and usability in the browser. However, modern browsers adhere to strict web standards, so this scenario seems unlikely. Why should Carbon need to compensate for a browser deficiency? And why is the compensation mechanism to only check for a related method? If Carbon was going to compensate, I would think that the mechanism would be a little more robust.

I did not see an explanation in https://github.com/carbon-design-system/carbon/pull/10897 or https://github.com/carbon-design-system/carbon/issues/10896 detailing why `addEventListener` was checked instead of `removeEventListener`.

Let me know if you still want me to change it and I will.

tay1orjones: Sounds good. I agree it makes sense to change it to check for the method it's about to call. addEventListener was probably added to this condition in error from a copy/paste of the addEventListener code above in the main body of the hook.

packages/react/src/internal/useMatchMedia.ts
```diff
@@ -8,39 +8,46 @@
 import { useState, useEffect } from 'react';
 import { canUseDOM } from './environment';
 
-export function useMatchMedia(mediaQueryString) {
+// TODO: When can support for `addListener` be removed?
```
tay1orjones: I think it's safe to remove 👍 

`addEventListener` already had near-full browser support as of the v11 release (March 2021). We probably could have removed it then but didn't. The only thing was the `signals` property wasn't supported until Safari 15, which was September 2021. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility

Not a huge deal and this is an internal-only helper so no concerns about backwards incompatibility.

tay1orjones (approved):

packages/react/src/internal/useMatchMedia.ts
adamalston: That brings up another question: should there even be checks for the `addEventListener` and `removeEventListener` methods or can they just be called? There are no checks for these methods in other hooks.

I have deleted the checks for the methods in ~~https://github.com/carbon-design-system/carbon/commit/0c8b6dcb6789523b1e7c33898073d2d806f74c65~~ https://github.com/carbon-design-system/carbon/commit/1ae2399d75ee57525b607d01872951fc74f4b1b8. Curious to hear thoughts on that.

tay1orjones: Yeah it's probably safe to remove the checks and just call them. If something comes up it's easy to put them back in.

riddhybansal: Makes sense!! Thanks for clarity !!

riddhybansal (approved):

PR #18902, opened 2025-03-24
refactor: delete remnants of `MultiSelect.Filterable`

### LINKED ISSUE #18899: Delete code related to `MultiSelect.Filterable`
There is leftover code in the codebase related to this component. Per https://github.com/carbon-design-system/carbon/blob/f574ad0e46ed2f7d84ba50c8561bd9cb20ffa1b3/packages/react/docs/migration/11.x-namespaced-exports.md#changes and other docs, `FilterableMultiSelect` should be used instead of `MultiSelect.Filterable`.

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18899

Delete code related to `MultiSelect.Filterable`

#### Changelog

**Changed**

- Updated the `displayName` of the `FilterableMultiSelect` component.

**Removed**

- Deleted remnants of `MultiSelect.Filterable`.
- Deleted duplicate sentences for `FilterableMultiSelect` in the docs.

#### Testing / Reviewing

```sh
yarn test packages/react
```

There are more code related to `Multiselect.Filterable` in these files:

- packages/react/docs/migration/11.x-namespaced-exports.md
- packages/upgrade/transforms/size-prop-update.js
- packages/upgrade/transforms/_\_testfixtures__/size-prop-update.input.js
- packages/upgrade/transforms/_\_testfixtures__/size-prop-update.output.js

The code in 11.x-namespaced-exports.md is necessary because it's documenting the Carbon v10 to v11 migration. However, I'm not sure whether the code in the packages/upgrade files is necessary. Should those codemods be converting `MultiSelect.Filterable` to `FilterableMultiSelect`? Should the `MultiSelect.Filterable` component be changed to some other arbitrary component? Or should we just leave the code in those files as is?


packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -5299,262 +5299,6 @@ Map {
   },
   "MultiSelect" => Object {
     "$$typeof": Symbol(react.forward_ref),
-    "Filterable": Object {
```
tay1orjones, line 5302: I just want to call out that this isn't a breaking change because it was deprecated all the way back in v10.41.0, https://github.com/carbon-design-system/carbon/pull/9008 and stated that it would be removed in v11. I don't know of a reason why this was left in, we probably just missed it in the lead-up to v11.

tay1orjones (approved):
The codemod doesn't handle `FilterableMultiSelect` because it was written two months before the flat export was added, https://github.com/carbon-design-system/carbon/pull/8570

Overall I don't think there's a worthwhile benefit worth updating `size-prop-update`. Migration from v10 to v11 is largely a thing of the past and the lack of `FilterableMultiSelect` support in the codemod has never been mentioned/reported since it was added in https://github.com/carbon-design-system/carbon/pull/9008

kennylam (approved):

PR #18905, opened 2025-03-24
fix(labels): update needs triage label automations

tay1orjones:
This updates issue templates and triage automations to use the updated label value for the needs triage label to avoid duplicate labels being created.

<details><summary>Full context</summary>
<p>

I updated the label a couple weeks ago to use the "colon-notation" for the emoji instead of hardcoding the emoji in the label. I was having lots of issues trying to query the github api with the emoji and the "colon-notation" fixed it. 

The issue templates and workflow automations _create_ any label that doesn't exist, which caused a duplicate label to be made

2025-03-24 at 15 08 23-MAIN-Google Chrome

by updating these files we shouldn't get a duplicate label created anymore.

</p>
</details> 

#### Changelog

**Changed**

- updated all usages of the needs triage label in the codebase

#### Testing / Reviewing

Once this is merged I'll clean up the old label and make sure issues with it are given the proper label

### DIFF
--- a/.github/ISSUE_TEMPLATE/ACCESSIBILITY_ISSUE.yaml
+++ b/.github/ISSUE_TEMPLATE/ACCESSIBILITY_ISSUE.yaml
@@ -2,7 +2,8 @@ name: Accessibility Issue ♿
 description: Report an accessibility or usability issue.
 title: '[a11y]: '
 type: 'bug'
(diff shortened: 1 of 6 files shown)


alisonjoseph (approved):
🕵️‍♂️

heloiselui (approved):
👍🏽

PR #18913, opened 2025-03-25
refactor: rewrite `wrapFocus` utils in typescript

### LINKED ISSUE #18911: Rewrite `wrapFocus` utils in TypeScript

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18911

Rewrote the `wrapFocus` utilities in TypreScript.

#### Changelog

**New**

**Changed**

- Rewrote the `wrapFocus` utilities in TypreScript.
- Updated types in usages of the `wrapFocus` utilities.

#### Testing / Reviewing

```sh
yarn test packages/react
```


packages/react/src/internal/wrapFocus.ts
```diff
@@ -0,0 +1,168 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { useEffect } from 'react';
+import { tabbable } from 'tabbable';
+import {
+  // TODO: Should the `DOCUMENT` constants be put into this file since they are
+  // only used here?
+  DOCUMENT_POSITION_BROAD_FOLLOWING,
```
heloiselui: Indeed, looks like those constants are only used in `wrapFocus` and aren’t keyboard-specific.
Might be cleaner to move them into this file.

packages/react/src/internal/wrapFocus.ts
```diff
@@ -0,0 +1,168 @@
...
+  /** The event that triggered this function. */
+  event: React.KeyboardEvent | KeyboardEvent;
+}) => {
+  if (!containerNode) return;
+
+  if (
+    ['blur', 'focusout', 'focusin', 'focus'].includes(event.type) &&
+    process.env.NODE_ENV !== 'production'
+  ) {
+    // TODO: Why do the hooks rules need to be ignored here? Is it not possible
```
heloiselui: Good point, not 100% sure, but I think the idea here was to show an error when wrapFocusWithoutSentinels gets called during events like blur or focusout, instead of keydown. Also, using useEffect like this seems a bit off since it's not in a React component, and it’s a bit outside the usual hook usage pattern in the [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks).
Maybe @tay1orjones or @kennylam can share some background on this approach if there is any?

tay1orjones: I can't think of a reason for this setup. Moving the condition inside the useEffect sounds good to me 👍

heloiselui (approved):

tay1orjones (approved):

adamalston:
Looks like someone is going to have to restart https://github.com/carbon-design-system/carbon/actions/runs/14541187128/job/40799369554?pr=18913.

Issue #17329, opened 2024-09-04
[Low contrast content switcher] Final design specs


laurenmrice:
## Acceptance criteria

- [x] Create the final design specs for the low contrast content switcher with the updated visuals and color tokens. 
- [x] Add an image of the spec here.

## Deliverables:
- [Spec in issue](https://github.com/carbon-design-system/carbon/issues/17329#issuecomment-2566710362)
- [Spec in Figma](https://www.figma.com/design/3h9TcywkocMGZpEsvGYTzQ/Content-Switcher?node-id=1853-53433&t=JZGOENhKf81Hbszr-1)

aagonzales:
# Low contrast content switcher

### New component level color tokens 
Question for dev: Can we nest layer token inside of a component level token?

<img width="925" alt="Image" src="https://github.com/user-attachments/assets/765d32ae-9d85-4cf3-b306-91b02dbabca7" />


### Spec

![Image](https://github.com/user-attachments/assets/5264b719-1bef-4c34-92d8-3dd3140c333f)
![Image](https://github.com/user-attachments/assets/f049a226-5441-45de-8dbd-83fa92a60196)

kiran-ravindra-ibm:
Updates being made in [this Figma branch](https://www.figma.com/design/YAnB1jKx0yCUL29j6uSLpg/branch/aj2AEofucmDdeX9wqQPi0c/(v11)-Carbon-Design-System?m=auto&node-id=68-3091&t=0z1RTRkJm8b1KXET-1). 

- Corner radius modified on all variants incl. as-is/Default (high contrast)
- Selected, unselected, focus, disabled states created for Subtle
- A blend mode issue that was causing darkened unselected states has been resolved
- After discussion with @aagonzales, @laurenmrice, we are proceeding with rounded focus states for now

Hover states are still in discussion. Need to determine what new color tokens need to be added, if any. Subtle switcher with as-is (Default/high contrast) hover values shown below: 
![Image](https://github.com/user-attachments/assets/efbde9d5-872f-4f83-b09b-a53a7aabdbe4)

Added to schedule for discussion tomorrow 1/28 at design crit

PR #18923, opened 2025-03-26
feat: low contrast content switcher implementation

### LINKED ISSUE #17332: [Low contrast content switcher] React Implementation
### Acceptance Criteria

- Define new color tokens for low contrast content switcher
- Final design specs https://github.com/carbon-design-system/carbon/issues/17329
- Update React - add lowContrast prop
- Update Web Components
- Add tests

> [!IMPORTANT]  
> Web components side of this is dependent on https://github.com/carbon-design-system/carbon/pull/18815 merging first before starting work there.

Gururajj77:
Closes #17332 

Worked with @ariellalgilmore 🚀 !!!

 React implementation on Low contrast content switcher

#### Changelog

**New**

- New low contrast versions for both default and icon only.
- `lowContrast` prop for toggling between the versions
- new color tokens, `$content-switcher-background`, `$content-switcher-background-hover` and `$content-switcher-selected`

#### Testing / Reviewing

Here is the [design spec](https://github.com/carbon-design-system/carbon/issues/17329) for the content switcher's low contrast version.


alisonjoseph (changes requested):
This looks great! 

Couple little things
- lowContrast control isn't working on the main Low Contrast story in storybook. Working fine for the other stories. 
- Do we want to update this doc to include content switcher https://github.com/carbon-design-system/carbon/blob/main/packages/styles/docs/sass.md#component-tokens
- also looks like a test is failing

Do we have an issue to update the website with these new tokens? https://carbondesignsystem.com/elements/color/tokens/#component-tokens

laurenmrice (changes requested):
Hey @Gururajj77 this is looking very good! ⭐️ Just a couple of small things I saw below:

----

## React 

**All content switchers**
- In the hover state on an unselected section that is next to a selected section, there is a small area at the top and bottom that does not get the hover state color. The hover state color should be in this area as well.

Screenshot 2025-04-24 at 5 31 41 PM

----

**Default and Low contrast icon only content switchers**
- When switching the selection between different sections in the icon only content switchers, the icon is jumping slightly upwords. The icon should always remain in the same place. It seems like it is only happening when the selected state also has focus on it, if you click off of the selected section and the focus is removed, then the icon jumps back.

icon jump

**Default and Low contrast content switchers**
- Similarly as above, there is a slight jump of the text when a selected state also has focus, once focus is removed and you click off of the selected section, then the text jumps back to normal. This is most noticeable right now in the low contrast version, but I noticed it happens on the first and third sections of the default version, while the second section does not do this and looks fine.

jumping text

----

**Low contrast content switchers**
In the unselected states, when there is also focus, there should not be a white border inside of the focus state. The white border should be removed. This is only for low contrast switchers.

white border

laurenmrice:
@alisonjoseph I have already added the new color tokens to the color token table on the website. I've got a PR up for it which I will merge once the code is done. https://github.com/carbon-design-system/carbon-website/pull/4487

laurenmrice (changes requested):
These are the last visual bugs I see, I think after this it should be good to go. They are very small things. The gifs are a little hard to see but you can follow it in the React story for it.

### Default, Icon only, Icon Only With Layer, and With Layer stories

**SelectionMode prop on `manual`**
- When placing focus on an unselected section and hovering on the section while it has focus, there should not be an inset space around it, it should just be the hover state background color touching the focus border.
- When hovering on an unselected section next to a section that has a focus border, the hover background color does not extend into the corners of the section.

This gif below shows both of these bugs:
Apr-30-2025 11-59-55

-----

### Low Contrast and Low Contrast Icon Only stories

**SelectionMode prop on `manual`**
Similarly like above, when hovering on an unselected section next to a section that has a focus border, the hover background color does not extend into the corners of the section.

This gif below shows both of these bugs:
Apr-30-2025 12-11-05

-----

### Low Contrast Icon Only story
- On hover, the icon should be` $icon-primary` color token.

laurenmrice (approved):
This looks good 👍, we will resolve the rest of the changes in a separate PR.

packages/react/src/components/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -238,6 +248,29 @@ export default class ContentSwitcher extends React.Component<
               selected: index === this.state.selectedIndex,
               ref: this.handleItemRef(index),
               size,
+              onMouseEnter: () => {
+                const { selectedIndex } = this.state;
+                // Check if the hovered index is adjacent to the selected button
+                if (Math.abs(index - selectedIndex!) === 1) {
+                  const selectedBtn = this._switchRefs[selectedIndex!];
```
adamalston: Suggestion to make this implementation type safe. I'm leaving this suggestion because I have rewritten this component in https://github.com/carbon-design-system/carbon/pull/19062 to have accurate types. Making the types here accurate is going to make it easier to merge these changes with the ones in that pull request.

```suggestion
                if (
                  typeof selectedIndex !== 'undefined' &&
                  Math.abs(index - selectedIndex) === 1
                ) {
                  const selectedBtn = this._switchRefs[selectedIndex];
```

packages/react/src/components/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -238,6 +248,29 @@ export default class ContentSwitcher extends React.Component<
...
+                    selectedBtn.classList.add(
+                      `${prefix}--content-switcher__selected-hovered`
+                    );
+                  }
+                }
+              },
+              onMouseLeave: () => {
+                const { selectedIndex } = this.state;
+                if (Math.abs(index - selectedIndex!) === 1) {
+                  const selectedBtn = this._switchRefs[selectedIndex!];
```
adamalston: Suggestion to make this implementation type safe.

```suggestion
                if (
                  typeof selectedIndex !== 'undefined' &&
                  Math.abs(index - selectedIndex) === 1
                ) {
                  const selectedBtn = this._switchRefs[selectedIndex];
```

packages/react/src/components/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -238,6 +248,29 @@ export default class ContentSwitcher extends React.Component<
               selected: index === this.state.selectedIndex,
               ref: this.handleItemRef(index),
               size,
+              onMouseEnter: () => {
+                const { selectedIndex } = this.state;
+                // Check if the hovered index is adjacent to the selected button
+                if (Math.abs(index - selectedIndex!) === 1) {
+                  const selectedBtn = this._switchRefs[selectedIndex!];
+                  if (selectedBtn) {
```
adamalston: `selectedBtn` is a `HTMLButtonElement` according to its type. Why is this condition necessary? The same question also applies to the condition in the other event handler.

packages/styles/scss/components/content-switcher/_tokens.scss
```diff
@@ -0,0 +1,110 @@
+//
+// Copyright IBM Corp. 2020
```
adamalston: https://github.com/carbon-design-system/carbon/issues/18954#issuecomment-2813348278

```suggestion
// Copyright IBM Corp. 2025
```

packages/themes/src/component-tokens/content-switcher/tokens.js
adamalston, line 1: Should this file have a copyright notice?

packages/themes/scss/_component-tokens.scss
adamalston, line 1: Do the copyright notices in this file and others in this pull request (excluding the autogenerated copyright notices) need to be updated?

packages/themes/tasks/builders/modules-content-switcher-tokens.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright IBM Corp. 2015, 2023
```
adamalston: https://github.com/carbon-design-system/carbon/issues/18954#issuecomment-2813348278

```suggestion
 * Copyright IBM Corp. 2025
```

packages/themes/src/component-tokens/content-switcher/index.js
adamalston, line 1: Should this file have a copyright notice?

packages/react/src/components/ContentSwitcher/ContentSwitcher-test.js
```diff
@@ -134,5 +134,41 @@ describe('ContentSwitcher - RTL', () => {
...
+        </ContentSwitcher>
+      );
+
+      // Should have both the low-contrast class and the icon-only class
+      expect(container.firstChild).toHaveClass(
+        'cds--content-switcher--low-contrast'
+      );
+      expect(container.firstChild).toHaveClass(
+        'cds--content-switcher--icon-only'
+      );
```
adamalston: Would it be better to assert all of the classes that the element has? Something similar to https://github.com/carbon-design-system/carbon/pull/19192#discussion_r2052651636. The same question applies to the earlier test as well.

packages/styles/scss/components/content-switcher/_tokens.scss
adamalston: This comment was marked as resolved but I don't see a change. Was there supposed to be one?

packages/styles/scss/components/content-switcher/_content-switcher.scss
adamalston, line 1: https://github.com/carbon-design-system/carbon/pull/18923#discussion_r2073938296

packages/styles/scss/components/content-switcher/_index.scss
adamalston, line 1: https://github.com/carbon-design-system/carbon/pull/18923#discussion_r2073938296

alisonjoseph (approved):
LGTM!

Issue #18206, opened 2024-12-05
Feature Flags: Notify consumers of available flags


tay1orjones:
When consumers use a component or feature that has been improved or replaced via a feature flag, we should notify them of the change. 

### Justification

Notifying consumers of changes to the system in their development environment has many benefits:

* **Early access to enhancements**, providing their users with a better experience sooner rather than later
* **Reduced technical debt**, encouraging a controlled, phased transition without the pressure of last minute changes when a flag is turned on by default
* **Input on the system**, ensuring new features are more tailored to specific needs of consumers' teams and end users
* **Lower friction during breaking changes**, giving more time and flexibility to test and implement changes before they become mandatory
* **Opportunity to innovate**, allowing teams to experiment with new capabilities to better solve their product and end user's needs. This encourages a culture of innovation within the ecosystem by leveraging the latest tools before they're fully rolled out.
* **Competitive advantage**, as features behind flags improve the visual design, accessibility, performance, etc. This ultimately allows teams to give their end users a better experience compared to their competitors. 
* **Overall alignment with the future of the system**, reducing friction and helping teams make more informed choices about their user experience to ensure future compatibility.

Being notified will give teams the knowledge to take advantage of faster iteration, smoother upgrades, and more control over their adoption timeline. Without these messages we risk declining overall product quality by leaving enhancements on the table with little adoption. 

### Risks

* Notifications will very easily be seen as too noisy. We should really nail down how to not be a nuisance.
* ...

### Additional requirements

- [ ] Notify `unstable_FeatureFlags` users that FeatureFlags has been marked stable and they can should update their imports

devadula-nandan:
js docs could be an idea. shows up right in the IDE.

tay1orjones:
Could we attach a popover or something to part of UIShell that would render in browser in development environments? Is that too invasive? Is there a way to gate this effectively?

tay1orjones:
Could `FeatureFlags` components spit out console warning messages for old flags, or flags that move to be committed to a major version? It could also have a "new flags" message announcing new ones, we'd periodically manually update it as new ones are created?

devadula-nandan:
### Options for placing feature flag information for discoverability

#### 1. Website Code tab

Since this is the primary discovery point for our component library, we could surface feature flags here as well.

Maybe we can reuse the Live Demo area to showcase feature flags by referencing feature-flag stories. We could document the flags in a table on the website, or somehow render the same MDX content that already exists in Storybook for each component.

[...]

PR #18942, opened 2025-03-27
feat(react): export FeatureFlags as stable

tay1orjones:
Part of the reason why feature flag usage is so low could be that the export is still marked `unstable`, so I've added a "stable" export without the prefix. 

This also ensures that as we add new flags the telemetry.yml file will updated as part of the release to reflect those changes. This should improve our telemetry picture of feature flag usage as we add and modify them over time.

#### Changelog

**Changed**

- add FeatureFlags and related exports, "legacy" `unstable_FeatureFlags` exports remain for backwards compatibility. We'll directly notify users of this in https://github.com/carbon-design-system/carbon/issues/18206
- update snaps to reflect these changes
- update docs to reflect these changes
- update version.yml workflow to run the script to update telemetry.yml. This means telemetry.yml changes will be included in the release PRs.
- update telemetry.yml via the same script used in the workflow


tay1orjones:
@sstrubberg @mattrosno @laurenmrice Huge thanks for figuring out we needed to do this! 🙏

alisonjoseph (approved):
LGTM!

heloiselui (approved):
LGTM!

PR #18951, opened 2025-03-28
fix(modal): allow data-modal-primary-focus in danger

### LINKED ISSUE #18904: [Bug]: unable to set focus within a danger Modal
### Package

@carbon/react

### Browser

Chrome

### Package version

^1.56.0 and ^1.65.0 (replit)

### React version

16.14.0 and ^18.2.0 (replit)

### Description

I'm unable to set focus to an element within a danger modal using `data-modal-primary-focus` or `selectorPrimaryFocus` . The AVT team opened [an issue](https://github.ibm.com/mhub/ui/issues/2003) saying focus should be on the `<TextInput />` when the modal opens but focus always goes to the cancel button. However, I'm able to set focus if I remove the danger prop. Related slack thread in the [carbon/react channel](https://ibm-cloudplatform.slack.com/archives/C2K6RFJ1G/p1742576031107969).

### Reproduction/example

https://replit.com/@bruceBanner/Setting-focus-within-a-danger-modal?v=1

### Steps to reproduce

1. create a project with a <Modal /> that has a <TextInput /> as the modal's child
2. add `data-modal-primary-focus` to the <TextInput />
3. open modal
4. focus should be on secondary button instead of <TextInput />

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

ariellalgilmore:
Closes #18904 

data-modal-primary-focus was not being set when using kind danger

#### Changelog

**Changed**

- updated so that a user can use `data-modal-primary-focus` with `danger` modal. but if they aren't applying `data-modal-primary-focus` then it will still focus on the secondary button for danger. 

#### Testing / Reviewing

go to modal stories 
- default (which has `data-modal-primary-focus`) then set the `Danger` boolean to true
  -  see if the focus state stays on the first text input
- danger modal story
  - check if focus is on the secondary button `Cancel`


alisonjoseph (approved):
LGTM!

kennylam (approved):

PR #18955, opened 2025-03-28
Add entry points for Sass

ibacher:
Trying to fix an issue that arose using Rspack, sass-loader, and Carbon. sass-loader configures the bundler to use the `sass`, `style`, or `main` properties (in that order) to resolve Sass imports. Rspack's resolver thus resolves directives like:

```scss
@forward @carbon/types;
```

to `@carbon/types/lib/index.js` instead of `@carbon/types/index.scss` as Webpack does. Adding a `sass` entry point resolves this issue.

#### Changelog

**New**

- Add sass entrypoints to modules with code entrypoints and an index.scss file.

#### Other

This shouldn't have any major impacts, but if you want to reproduce and verify a fix, I have a minimal example repo that demonstrates the issue [here](https://github.com/ibacher/sass-loader-issue).


packages/elements/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "Apache-2.0",
   "main": "lib/index.js",
   "module": "es/index.js",
+  "sass": "index.scss",
```
tay1orjones, line 8: Should the entrypoints contain the current directory? The [examples from the docs](https://sass-lang.com/documentation/js-api/classes/nodepackageimporter/) have it. 

```suggestion
  "sass": "./index.scss",
```

tay1orjones (approved):
I took a dive on this and found this matches the expectation of the new [package importers](https://sass-lang.com/blog/rfc-package-importer/). Also for reference, [here's the docs on NodePackageImporter](https://sass-lang.com/documentation/js-api/classes/nodepackageimporter/)

I have one minor question about the paths but otherwise this looks good to me. Thanks for taking a look at it!

packages/elements/package.json
ibacher: My read on it is that it's similar to Node's rules for resolving `main` and `module`, so I followed the same style that was already used for those. 

In the actual [Dart code](https://github.com/sass/dart-sass/blob/52221c0e1587c745d073eac961bc665dcbb6075f/lib/src/importer/node_package.dart#L147) it uses `path.join()` using the directory containing the `package.json` and the path part specified by `sass` entry, so I think this is safe.

tay1orjones: Awesome, thanks!

2nikhiltom (approved):

PR #18956, opened 2025-03-28
fix(datepicker): add picker type to dependencies of initialization hook

### LINKED ISSUE #18828: [Bug]: Date Picker does not handle type change
### Package

@carbon/react

### Browser

Firefox, Chrome

### Package version

1.78.0

### React version

18.3.14

### Description

When the `datePickerType` prop of the `DatePicker` component is changed programatically, the component does not behave properly:
* If the type was `single` and was changed to `range`, it does not allow to select an end date
* If the type was `range`, had a date range selected, and was changed to `single`, when opening again a range is still selected, and if you select a new date it is handled as a start date

### Reproduction/example

https://stackblitz.com/edit/github-hs9y3qpy?file=src%2FApp.jsx

### Steps to reproduce

1. Select values for the date pickers, it should work as expected.
2. Click the toggle button
3. Try to select a range for the first input, it does not allow you to select an end date
4. Try to select a single date for the second input, if the date you selected is before the end date, it keeps selecting a date range

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

ychavoya:
Closes #18828 

Added the `datePickerType` prop to the dependencies array of the initialization hook so the component can change the behavior after the initial render.

#### Changelog

**Changed**

- Modified the dependency array of the `useEffect` used for initialization of the calendar and event listeners to include the `datePickerType` property.

#### Testing / Reviewing

1. Open the [default DatePicker story](http://localhost:3000/?path=/story/components-datepicker--default) in storybook
2. Verify you can open the calendar and select one date
3. In the controls, change the `datePickerType` to a different type
4. If you selected `simple`, it should not show a calendar when focusing on the element
5. If you selected `range`, it should show the calendar and allow you to select a date range instead of only one date
6. If you go back to `single` it should behave as it initially did

Previous behavior:
Kapture 2025-03-18 at 09 53 06

New behavior:
Kapture 2025-03-18 at 09 59 10


annawen1 (approved):
LGTM!

heloiselui (approved):
That's great! Thanks!

PR #18971, opened 2025-03-31
refactor: replace `displayName` checks with type comparisons

### LINKED ISSUE #18969: Replace `displayName` checks with direct type comparisons
Example of code that should be updated: https://github.com/carbon-design-system/carbon/blob/c5899b98721e7c6d9f48a4ae2bb66d6244cbcde6/packages/react/src/components/Checkbox/Checkbox.tsx#L169

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18969

Replaced `displayName` checks with component type comparisons.

#### Changelog

**New**

- Added testing for the code that was updated in HeaderPanel.tsx.

**Changed**

- Replaced `displayName` checks with component type comparisons.

#### Testing / Reviewing

```sh
yarn test packages/react
```

Reference commits:

- https://github.com/carbon-design-system/carbon/compare/efbce4e304e03bb520a138f31ed317be47d32246...675239c749de1992e007bd7e5414338d2a3cab4c
- https://github.com/carbon-design-system/carbon/compare/2601d361089b0afd734f87defa0c5039ec8b943c...f8495e85e298fe2897c44bb2021a64905954b74f

The old code compared the display name of the decorator element (i.e., `normalizedDecorator['type']?.displayName === 'SomeComponentName'`), which is fragile because it relies on the display name not changing. The updated code uses a direct type comparison (i.e., `normalizedDecorator?.type === SomeComponent`), which is more robust because it compares the actual component references rather than a mutable string. These changes make the component type checks more reliable.


annawen1 (approved):

riddhybansal (approved):

tay1orjones:
@adamalston @riddhybansal @annawen1 I think it's possible this could break some folks. With `displayName` consumers could use their own components in place of the ones provided through `@carbon/react` as long as the API matched and they set `displayName` to be the one expected by the parent component internals. How could consumers still do that with this change?

adamalston:
They wouldn't be able to. Comparing `displayName` does allow for more flexible "duck typing," where consumers can plug in their own components as long as they match the expected API and explicitly set `displayName`. Switching to reference-based comparisons prevents that kind of substitution, since it requires the exact component identity.

That said, using `displayName` for component checks is inherently fragile which is the main issue that this pull request was attempting to address. It creates a contract that's undocumented (or at least, I haven't found documentation for it — does it exist somewhere?). It also limits the benefits of using TypeScript.

Do we need to make a change to preserve flexibility? If so, there may be a few options to do that while improving robustness:

- Fall back to a `displayName` check if the direct reference comparison fails.
- Explicitly document a supported pattern for overriding components via props or context, rather than relying on internal implementation details. This option is contingent on this behavior being supported.

Do you have any thoughts, especially on whether this kind of override is meant to be part of the public API, or more of an unsupported escape hatch?

tay1orjones:
@adamalston Yeah, there's never been a statement of support for `displayName` spoofing. I agree removing that fragility is for the best. If this change ends up being disruptive I think the fallback is a good stopgap measure. If it's not disruptive then there's no need to do anything.

In the long term we'll probably still run into the many inherent shortcomings with checking for specific children and cloning them. We're working on one right now:
* https://github.com/carbon-design-system/carbon/issues/19522

Eventually using context instead like you mention would be a good option. It's worked well for Tabs, and if they're exported it could aid consumers in creating their own "override" components that reimplement ones provided through `@carbon/react`. I outlined what that approach could look like to fix the above issue in https://github.com/carbon-design-system/carbon/issues/19522#issuecomment-2932177171. 

FWIW Spectrum's gone pretty far down this road with their [advanced customization](https://react-spectrum.adobe.com/react-aria/advanced.html). The alternative would be something like the [`asChild` pattern](https://www.radix-ui.com/primitives/docs/guides/composition) popularized by Radix.

cc @Gururajj77

Issue #17279, opened 2024-08-27
Move from webpack to vite for storybook config


guidari:
We will have much better performance by moving from webpack to vite.

This [PR](https://github.com/carbon-design-system/carbon/pull/16910/) have tried to that along side with the Storybook 8 migration. [This commit](https://github.com/carbon-design-system/carbon/pull/16910/#:~:text=2%20weeks%20ago-,fix%3A%20added%20babel%20back,-75cc693) was the last one using vite, we can use that as a start point, since it has a lot of steps done.

Just in case we face the same issue, [here](https://github.com/vitejs/vite-plugin-react/issues/357) it is an explanation about `__DEV__` being undefined.

PR #18978, opened 2025-03-31
fix(warning): replace `__DEV__` with `process.env.NODE_ENV` condition

tay1orjones:
### Background

Pieces of the codebase that should only run in non-production environments are currently guarded by a global constant, `__DEV__`.

This works in concert with a babel plugin, [`babel-plugin-dev-expression`](https://www.npmjs.com/package/babel-plugin-dev-expression), seen as `dev-expression` in the babel config.

https://github.com/carbon-design-system/carbon/blob/787d3f252fad59ea13ca4ed06e83b0eabfc1a038/config/babel-preset-carbon/index.js#L25

The plugin transforms usages of `__DEV__` into `process.env.NODE_ENV !== "production"`. 

```diff
- if (__DEV__) {
+ if (process.env.NODE_ENV !== "production") {
  console.log(`something that should only happen in development environments`)
}
```

One example is `useNoInteractiveChildren`: https://app.unpkg.com/@carbon/react@1.79.0/files/es/internal/useNoInteractiveChildren.js#L12

---

The usage of `__DEV__` is problematic when a build pipeline is used without babel (like in the utilities package, and the vite storybook migration https://github.com/carbon-design-system/carbon/issues/17279)

This PR replaces all `__DEV__` usage and adds a eslint rule to avoid it's use in the future.

I didn't remove `babel-plugin-dev-expression` because it does two other transformations regarding `warning()` and `invariant()` calls that we rely on (check out the package readme for more info). This could be removed in the future though.

#### Changelog

**New**

- add eslint rule to prevent `__DEV__` usage

**Changed**

- replace `__DEV__` usages with `process.env.NODE_ENV` condition
- update dev docs

#### Testing / Reviewing

- CI should pass
- Look at the stories for the components and try to get the warning to pop - it should not pop warnings in the deploy previews because this is considered a "production" deployment environment.
- Pull locally and run a build - look at the `es` build output in `packages/react` and compare the components changed in this PR to what's on unpkg, it should be the same without any change.

### DIFF
--- a/config/eslint-config-carbon/base.js
+++ b/config/eslint-config-carbon/base.js
@@ -39,7 +39,4 @@ module.exports = {
     jest: true,
     jasmine: true,
   },
-  globals: {
-    __DEV__: true,
-  },
 };
--- a/config/eslint-config-carbon/rules/best-practices.js
+++ b/config/eslint-config-carbon/rules/best-practices.js
@@ -11,5 +11,12 @@ module.exports = {
   rules: {
     // @see https://eslint.org/docs/rules/curly
     curly: 'error',
+    'no-restricted-globals': [
+      'error',
+      {
+        name: '__DEV__',
+        message: "Use process.env.NODE_ENV !== 'production' instead",
+      },
+    ],
   },
 };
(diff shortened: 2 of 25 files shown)


kennylam (approved):

heloiselui (approved):
🎉🎉

PR #19020, opened 2025-04-03
fix(react): nextjs hydration error (#19003)

### LINKED ISSUE #19003: [Bug]: UI Shell Sidenav is unclickable on Lg breakpoint
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.79.0

### React version

v18.2.0

### Description

https://github.com/user-attachments/assets/01f5d786-d6d9-44ab-8f29-b54fff50244a

The links and menu items on the side nav are not clickable (see the video) when the view port is at `Lg` breakpoint or above (the side nav is visible by default)

The issue seems only to exist only on NextJS (I tried NextJS 13 and 15, with React 18 and 19).

Tried with Vite + React but not reproducible.

The issue vanishes when shink viewport side to hide the side nav then extend it again.

### Reproduction/example

https://stackblitz.com/edit/stackblitz-starters-n9jgxuuv

### Steps to reproduce

2. `yarn` & `yarn dev`
3. Open in browser with large view port (>= `Lg` breakpoint)
4. Load the page
5. Try to click on side nav items

### Suggested Severity

### Application/PAL

A minimal POC with Carbon

### Code of Conduct

maisonsmd:
Closes #19003

Fix useMatchMedia that causes hydration mismatch in NextJS because provided initial value in useState() is different between server and client.

#### Changelog

**New**

- Add optional default state for `useMatchMedia`, it should not affect existing code.

**Changed**

**Removed**

- Removed initial media match logic for `useMatchMedia`, the logic will run in useEffect() when the component mounts

#### Testing / Reviewing

This issue can be verified using minimal code (https://stackblitz.com/edit/stackblitz-starters-n9jgxuuv) provided in #19003 description


maisonsmd:

tay1orjones (approved):
It's not a blocker, but it's worth mentioning that I think this means that the initial render can no longer obtain/use the actual value of `mediaQueryList.matches` since the effect is ran after the initial render.

A couple references:
* https://www.joshwcomeau.com/react/the-perils-of-rehydration/
* https://react.dev/reference/react/useEffect#displaying-different-content-on-the-server-and-the-client

heloiselui (approved):

maisonsmd:
> It's not a blocker, but it's worth mentioning that I think this means that the initial render can no longer obtain/use the actual value of `mediaQueryList.matches` since the effect is ran after the initial render.
> 
> A couple references:
> 
> * https://www.joshwcomeau.com/react/the-perils-of-rehydration/
> * https://react.dev/reference/react/useEffect#displaying-different-content-on-the-server-and-the-client

I agree with you, the [react-use](https://github.com/streamich/react-use) package had similar concern and provided a way to let user choose whether to use default value or the result of `mediaQueryList.matches`

https://github.com/streamich/react-use/blob/ad33f76dfff7ddb041a9ef74b80656a94affaa80/src/useMedia.ts#L17

But to make use of it, you have to provide default value in frameworks that have SSR support, and `undefined` in CSR-only ones.

PR #19025, opened 2025-04-03
feat: add base config for eslint v9

### LINKED ISSUE #19006: Add base config for ESLint

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/19006

Added a base config for ESLint v9.

#### Changelog

**New**

- Added a base config for ESLint v9.

**Changed**

- Updated the `lint` script in `package.json`.
- Updated the `lint` script in `package.json` and the lint-staged config. These changes will prevent files with issues from being commited and will result in CI failing if commit verification is skipped.

**Removed**

- Deleted unused ESLint directives. I will note that these deletions occurred automatically due to this code: https://github.com/carbon-design-system/carbon/blob/6095667260a8a74e4725fe3a424aa2facd2017a5/eslint.config.mjs#L26-L36

#### Testing / Reviewing

Reference commits: 
- https://github.com/carbon-design-system/carbon/commit/93063499a20dd1dba3a63d04e6f02aa1e619a0c1
- https://github.com/carbon-design-system/carbon/compare/25aa770ae1957eeb9b116316306445689157b9f6...5c2af08a193857dcc2ac8051528db716768ffc77

I need an answer to the `TODO` in `.gitignore`. All of the others can be addressed in future pull requests.

Curious to hear thoughts on the script and config changes in `package.json` if anyone has any. I don't think these changes will prevent someone from doing something that they're allowed to do so if someone could confirm, that would be helpful. For reference, I have set these things up the same way I have them set up in other internal and external projects.

I started with the ESLint config at the root of the project because that's what was added in https://github.com/carbon-design-system/carbon/commit/3b7067c22857856d467ef398473f4f101f3a165c in https://github.com/carbon-design-system/carbon/pull/16577, the last change related to ESLint. Eventually, this work will need to encompass `eslint-config-carbon`. If anyone thinks that wasn't a good starting point, let me know.


2nikhiltom:
Hey, thanks for another PR. This looks good to me, I have a question regarding the TODO about nested .gitignore files:

I am trying to understand what specific limitations in our current ignore configuration would be meaningfully addressed by a comprehensive approach to handling nested .gitignore files, while ensuring we maintain our existing workflow's performance and predictability?

Could we develop a targeted utility to collect .gitignore patterns across nested directories? The approach would balance pattern identification with performance constraints, focusing on controlled recursive traversal and minimizing file system overhead

Curious to hear your thoughts on balancing comprehensiveness with performance and predictability.

adamalston:
The `TODO` came from noticing that ESLint was still processing files ignored in nested `.gitignore` files. That led me to add extra ignores in the config. The idea behind the `TODO` was to explore whether we could better align ESLint's behavior with Git’s by honoring those nested ignore files. Doing so would speed up linting by avoiding unnecessary work. There's also a chance that linting those files, which may be compiled build files, could break a local build. To fix that, a user may have to re-run an entire build.

I should also note that I added the `TODO` after trying to integrate https://eslint.org/docs/latest/use/configure/ignore#including-gitignore-files, which didn’t quite work.

> Could we develop a targeted utility to collect .gitignore patterns across nested directories? The approach would balance pattern identification with performance constraints, focusing on controlled recursive traversal and minimizing file system overhead.

Can you clarify you mean by "targeted utility"? Some kind of utility would be needed to support this change, but I’m not sure what's being emphasized with "targeted" here.

.gitignore
```diff
@@ -16,8 +16,10 @@ css
 # Config files
 .npmrc
 
-# Cache folders
+# Caches
 .cache
+.eslintcache
+# TODO: Where is Parcel used?
```
tay1orjones: We used to use it in a few random areas - for examples mainly. We removed it a while back because of https://github.com/parcel-bundler/parcel/issues/7923

config/eslint-config-carbon/plugins/eslint-plugin-ssr-friendly.js
```diff
@@ -7,6 +7,9 @@
 
 'use strict';
 
+// TODO: `ssr-friendly` doesn't support ESLint v9.
```
tay1orjones, line 10: IMO this one isn't critical and we can remove it

eslint.config.mjs
```diff
@@ -70,6 +85,52 @@ export default [
 
       // Templates
       'packages/cli/src/component/templates/**',
+
+      // Generated files.
+      'packages/feature-flags/src/generated',
```
tay1orjones: Honestly this could be something like `**/generated/**`

package.json
```diff
@@ -21,7 +21,7 @@
     "doctoc": "doctoc --title '## Table of Contents'",
     "format": "prettier --cache --write '**/*.{js,md,scss,ts,tsx}' '!**/{build,es,lib,storybook,ts,umd}/**'",
     "format:diff": "prettier --list-different '**/*.{js,md,scss,ts,tsx}' '!**/{build,es,lib,storybook,ts,umd}/**' '!packages/components/**'",
-    "lint": "eslint actions config packages www",
+    "lint": "eslint --cache --fix --max-warnings 0 actions config packages www",
```
tay1orjones: I don't think `--cache` and `--fix` should be on by default so `yarn lint` remains deterministic in CI. Both can/should be turned on within a dev's IDE though. Maybe it would make sense for the precommit hook could use them. A new `lint:fix` script with both these options could be nice too.

Will `--max-warnings 0` cause it to always fail, even when there are 0 warnings? I assume the default `-1` means it's turned off and always passes, but not sure. Would we want `--max-warnings 1`?

tay1orjones (approved):
A couple minor things but overall LGTM!

config/eslint-config-carbon/plugins/eslint-plugin-ssr-friendly.js
adamalston: I opened an issue to delete this functionality (https://github.com/carbon-design-system/carbon/issues/19212) since deleting it will mean uninstalling a dependency.

package.json
adamalston: Regarding `--max-warnings 0`, it fails only if there are warnings, but not if there are no warnings. So it should still pass when everything is clean. The default is `-1`, which disables the check entirely.

We use `--max-warnings 0` internally in a pre-commit hook to prevent commits from being created when warnings are present. The idea is I want the lint to fail if there are warnings or errors, not just errors.

2nikhiltom (approved):

PR #19043, opened 2025-04-07
fix: integrates carbon ToolTip in TreeNode

### LINKED ISSUE #18919: [Bug]: TreeNode to support truncation when label text gets too big, consider integrating <Tooltip> within TreeNode
### Package

@carbon/react

### Browser

_No response_

### Package version

v11.78.0

### React version

_No response_

### Description

1. Label text does not get truncated in treeview/treeNode

2. The component does not have `tooltips` integrated within it (like -IconButton, ContentSwitcher with icons, CopyButton, Slider )
Image

### Reproduction/example

https://stackblitz.com/edit/github-ndgnpet1?file=src%2FApp.jsx&preset=node=

### Steps to reproduce

Pass a really long `label` to TreeNode 
Note: 
A workaround is to pass a Tooltip in a label - so that label is displayed inside of a tooltip but here too no truncation happens 
```
 <TreeNode
            key="1"
            id="1"
            value="Artificial intelligence"
            label={
              <span>
                this does not truncate : Artificial intelligence Artificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligenceArtificial
                intelligenceArtificial intelligence
              </span>
            }
          />
```

Image

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/18919

This PR adds  support for  truncation in  TreeNode  when label text gets too long.
Integrates "IconButton" component to have Tooltip support  when label text is truncated

#### Changelog

**New**
Added a `useEllipsisCheck` custom hook that detects when text is being truncated
Created `extractTextContent` utility function to handle various React node types
added styles
Adds avt test cases
Adds autoalign and align prop to treenode API
 
**Changed**
Updated tests to accommodate the new DOM structure - Made minimal changes to maintain test intent while fixing failures

#### Testing / Reviewing
- Open the deploy preview
- Hover over TreeNode items with longer labels to verify truncation detection and tooltip appearance
- Test with both parent and child nodes
- Verify keyboard navigation and focus states work correctly
- check if style change makes sense


laurenmrice (changes requested):
This is looking pretty good Nikhil, just a couple comments below. I created a spec image that you can refer to for some of these points.

Frame 4

- [x] Can we use the [icon button tooltip](https://react.carbondesignsystem.com/?path=/story/components-iconbutton--default) here? I think it would be good to keep the padding within the tooltip condensed like it is in the button icon tooltip, so it has less potential to cover up other elements.
- [x] We should use real content with truncation for the treeview node label text. I have an example in the spec image below. We could use the term "Application development and integration solutions," which should still truncate at this treeview container width size.
- [x] The label text should start truncating with the "..." 16px before the end of the treeview container. 
- The label text does not look completely centered within the row and sits slightly high. I would also make sure it uses the `$body-compact-01` type token like the other label text in rows.
- [ ] ~On hover and focus states, the whole tooltip container should be center-aligned next to the text label. Right now the tooltip is positioned slightly low, so the caret isn't pointing directly to the label text.~ Looks like the implementation is going to just use the center top alignment which I am okay with. People can change this alignment if needed for their tree view use case.
Screenshot 2025-04-08 at 12 10 25 PM

---

- [x] The tooltip currently has a vertical line between the caret and the tooltip container, which should not be appearing.
Screenshot 2025-04-08 at 12 12 34 PM

----

- [x]  In the focus state, the label text at the beginning gets a little cut-off. Is there a way to add some more padding to the focus border to avoid that?
Screenshot 2025-04-08 at 12 09 28 PM

----

- [x] The text in the first row looks off. The row is not the right height size and it should be 32px in height, currently it looks larger than that. You can reference the row below it called "Blockchain" which looks correct.

----

- [x] I noticed that there is now another truncated node with the tooltip opening above. I feel like we don't need two examples of this and I would rather keep the first example. Also, the text here is siting too high in the row  and needs to be center in the row.

Screenshot 2025-06-04 at 12 57 09 PM

Gururajj77 (approved):
Looks good, great work!

riddhybansal (changes requested):
Hey I feel you should add test cases for codeCov !!

riddhybansal (approved):
LGTM 💯

laurenmrice (changes requested):
**Focus**
- [x] There should not be an inner border of white in the light themes and black in the dark themes. It should just be the focus border.
Frame 2

----

**Tooltip**
- [ ] There should be 16px space on the right after the text like there is on the left side according to the design spec.
Frame 3

2nikhiltom:
Hey @laurenmrice 

> There should be 16px space on the right after the text like there is on the left side according to the design spec.

I looked at the [structure of tooltip ](https://carbondesignsystem.com/components/tooltip/style/#structure)in cds website there is padding on top and bottom as well for multi-line tooltip . 
Do we want to keep it as is in treeview implementation  or override/change the styles for tooltips inside of treeview?
image

The padding issues seems to be bug in the `iconButton` component itself  
https://stackblitz.com/edit/github-npbpavky?file=src%2FApp.jsx&preset=node=

- Do you think this needs to be fixed in `Iconbutton` component 
image

laurenmrice:
@2nikhiltom Okay, use the multi-line design spec with 16px padding on all sides for now. Eventually, I think design needs to come up with a different spec for all ellipsed text situations that is a bit more condensed when it comes to the internal padding in the tooltip; this version just feels a little large to me and gives more potential to cover up other important information on the page. 

The padding issues of the icon button tooltip can be a separate issue. Because yes, once the text wraps to two lines, the padding should change to the multi-line design spec.

2nikhiltom:
https://github.com/carbon-design-system/carbon/issues/19623 issue for padding  and Text not filling the available width properly

laurenmrice (approved):

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -97,8 +99,122 @@ export type TreeNodeProps = {
...
+    const children = element.props.children;
+    return extractTextContent(children);
+  }
+
+  return '';
+};
+
+type HTMLElementOrAnchor = HTMLElement | HTMLAnchorElement | null;
+
+const useEllipsisCheck = (
```
annawen1, line 150: Just a comment, I see we check if ellipsis is applied or will be applied in a few components (Tag, Dismissible Tag, Operational Tag, Tabs, FileUploaderItem, etc) we might want to pull this out in the future as an internal hook to share amongst these components

annawen1 (approved):

PR #19050, opened 2025-04-07
feat(PageHeader): Page Header Content collapsible page actions behavior

### LINKED ISSUE #18903: PageHeader: Content zone - right actions collapsible behavior
Complete collapsible behavior: actions fold into menubutton, ensure proper focus order

annawen1:
Closes #18903 

This PR sets up the collapsible behavior for the PageHeaderContent page actions. The actions are collapsed into a MenuButton component when PageHeader is minimized. 

#### Changelog

**New**

- `PageHeader.ContentText` component to handle the max-width styling
- `PageHeader.ContentPageActions` component to separate out logic and increase composability 
- add `@carbon/utilities` to react dependencies as the `overflowHandler` utility is being used

**Changed**

- measure the width of the menu button and apply the width to the css variable for the `grid-template-columns` property

#### Testing / Reviewing

Review PageHeader Content with Actions story to ensure the collapsible behavior of the page actions works


kingtraceyj (changes requested):
it's looking great! one small change and two questions.

**Changes:**
- Take out the underline below the truncated title

**Questions:**
1. Could teams have control over what elements show at smaller view ports? Although I think most will use our responsive patterns, there are use cases for certain things to not be visible at smaller views.
2. With the content area below the subtitle, when there is text inside we recommend the 640px max width. if a user decides to not put text there and instead another component, will it be constrained by the 640px max-width?
image

annawen1:
@kingtraceyj 

> it's looking great! one small change and two questions.
> 
> **Changes:**
> 
> * Take out the underline below the truncated title

Done!

> 
> **Questions:**
> 
> 1. Could teams have control over what elements show at smaller view ports? Although I think most will use our responsive patterns, there are use cases for certain things to not be visible at smaller views.

I think we can leave that to the teams to handle where they can apply styles (`display:none`) on their own which will hide the component from view.

> 2. With the content area below the subtitle, when there is text inside we recommend the 640px max width. if a user decides to not put text there and instead another component, will it be constrained by the 640px max-width?

I changed this to make the text its own component, that way if a user doesn't want to be constrained by the 640px max-width, they can just pass in whatever components on their own in!

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,186 @@ PageHeaderContent.propTypes = {
...
+>(function PageHeaderContentPageActions(
+  {
+    className,
+    children,
+    menuButtonLabel = 'Actions',
+    pageActions,
+    ...other
+  }: PageHeaderContentPageActionsProps,
+  ref
+) {
```
adamalston: Should this function be simplified to arrow function which is what most other components use? I have this question for other components in this file as well because function names are being duplicated which doesn't seem necessary.

```suggestion
>((
  {
    className,
    children,
    menuButtonLabel = 'Actions',
    pageActions,
    ...other
  }: PageHeaderContentPageActionsProps,
  ref
) => {
```

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,186 @@ PageHeaderContent.propTypes = {
...
+  PageHeaderContentPageActionsProps
+>(function PageHeaderContentPageActions(
+  {
+    className,
+    children,
+    menuButtonLabel = 'Actions',
+    pageActions,
+    ...other
+  }: PageHeaderContentPageActionsProps,
+  ref
```
adamalston: `ref` is unused here and in other components in this file. Should `forwardRef` still be used or should this component be a regular component?

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,186 @@ PageHeaderContent.propTypes = {
...
+  const containerRef = useRef<HTMLDivElement>(null);
+  const offsetRef = useRef<HTMLDivElement>(null);
+  const [menuButtonVisibility, setMenuButtonVisibility] = useState(false);
+  const [hiddenItems, setHiddenItems] = useState<any[]>([]);
+
+  // need to set the grid columns width based on the menu button's width
+  // to avoid overlapping when resizing
+  useLayoutEffect(() => {
+    if (menuButtonVisibility && offsetRef.current) {
+      console.log('hello');
```
adamalston:
```suggestion
```

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,186 @@ PageHeaderContent.propTypes = {
...
+                    hiddenItems.length > 0 &&
+                    hiddenItems
+                      .reverse()
+                      .map((item) => (
+                        <MenuItem
+                          key={item.id}
+                          label={item.label}
+                          onClick={item.onClick}
+                        />
+                      ))}
```
adamalston: If the types are correct and `hiddenItems` is always an array, I don't think the conditions here should be necessary. It seems like you could simplify quite a bit.

```suggestion
                  {hiddenItems.reverse().map((item) => (
                    <MenuItem
                      key={item.id}
                      label={item.label}
                      onClick={item.onClick}
                    />
                  ))}
```

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,176 @@ PageHeaderContent.propTypes = {
...
+  HTMLDivElement,
+  PageHeaderContentPageActionsProps
+>(
+  ({
+    className,
+    children,
+    menuButtonLabel = 'Actions',
+    pageActions,
+    ...other
+  }: PageHeaderContentPageActionsProps) => {
```
adamalston: I think this can be simplified more:

1. `children` is unused.
2. `other` is unused.
3. `PageHeaderContentPageActionsProps` is inferred from the generics on `forwardRef`.

Also, if `children` is unused, should it be removed from the component's interface, or is the intention that it will be used in the future?

```suggestion
  ({
    className,
    menuButtonLabel = 'Actions',
    pageActions,
  }) => {
```

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,176 @@ PageHeaderContent.propTypes = {
...
+  /**
+   *  The PageHeaderContent's page actions collabpsible Menu button label
+   */
+  menuButtonLabel?: string;
+  /**
+   * The PageHeaderContent's page actions
+   */
+  pageActions?: React.ReactNode;
+}
+const PageHeaderContentPageActions = React.forwardRef<
```
adamalston: Regarding the question in https://github.com/carbon-design-system/carbon/pull/19050#discussion_r2039839176, could this be a regular component since there is no `ref`? This question also applies to other components in this file that don't have a `ref` and use `forwardRef`.

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,176 @@ PageHeaderContent.propTypes = {
...
+  /**
+   * Provide child elements to be rendered inside PageHeaderContentPageActions.
+   */
+  children?: React.ReactNode;
+  /**
+   * Specify an optional className to be added to your PageHeaderContentPageActions
+   */
+  className?: string;
+  /**
+   *  The PageHeaderContent's page actions collabpsible Menu button label
```
adamalston:
```suggestion
   *  The PageHeaderContent's page actions collapsible Menu button label
```

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -17,8 +18,11 @@ import { usePrefix } from '../../internal/usePrefix';
 import { breakpoints } from '@carbon/layout';
 import { useMatchMedia } from '../../internal/useMatchMedia';
 import { Text } from '../Text';
+import { MenuButton } from '../MenuButton';
+import { MenuItem } from '../Menu';
 import { DefinitionTooltip } from '../Tooltip';
 import { AspectRatio } from '../AspectRatio';
+import { createOverflowHandler } from '@carbon/utilities/es/overflowHandler';
```
adamalston: Is it safe to import this from the `es` directory? Or should it be the following instead:

```suggestion
import { createOverflowHandler } from '@carbon/utilities';
```

I will note that when it's imported from the `es` directory, the types for the function are not picked up when it's used. That's not the best because you're never going to reap the benefits of using TypeScript.

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,176 @@ PageHeaderContent.propTypes = {
...
+    const classNames = classnames(
+      {
+        [`${prefix}--page-header__content__page-actions`]: true,
+      },
+      className
+    );
+    const containerRef = useRef<HTMLDivElement>(null);
+    const offsetRef = useRef<HTMLDivElement>(null);
+    const [menuButtonVisibility, setMenuButtonVisibility] = useState(false);
+    const [hiddenItems, setHiddenItems] = useState<any[]>([]);
```
adamalston: Can the `any` type be replaced with a fully qualified type? Asking because later on, there is the following code:

https://github.com/carbon-design-system/carbon/blob/a9f98e4d454a476acd8454c9e68f4ded8423520b/packages/react/src/components/PageHeader/PageHeader.tsx#L328-L334

Based on that code, it seems that `hiddenItems` entries should have at least `id`, `label`, and `onClick` properties.

packages/react/src/components/PageHeader/PageHeader.tsx
```diff
@@ -241,6 +235,176 @@ PageHeaderContent.propTypes = {
...
+                  <div key={action.id} className="action">
+                    {action.body}
+                  </div>
+                ))}
+                <span data-offset data-hidden ref={offsetRef}>
+                  <MenuButton
+                    menuAlignment="bottom-end"
+                    label={menuButtonLabel}
+                    size="md">
+                    {hiddenItems.reverse().map((item) => (
```
adamalston: Suggestion to avoid mutating `hiddenItems`:

```suggestion
                    {[...hiddenItems].reverse().map((item) => (
```

packages/react/src/components/PageHeader/PageHeader.tsx
annawen1: Ah I'm used to importing from the `es` folder when working on the web components package and did the same here - yep I'll adjust that

packages/react/src/components/PageHeader/PageHeader.tsx
annawen1: It's most likely going to change, but I'll define the type for now

packages/react/src/components/PageHeader/PageHeader.tsx
annawen1: yeah I'll convert the components created in this PR to regular components and if needed in the future we can adjust

packages/react/src/components/PageHeader/PageHeader.tsx
annawen1: We'll most likely change the way the page actions are consumed, instead of having them passed in through the `pageActions` prop I would like to have them be passed in through `children` so I'll keep that in for now. There will be another issue opened to address that

kingtraceyj (changes requested):
Spacing:
It would be nice when the subtitle and body copy is present to have the spacing between subtitle and text to be $spacing-03 Would this be feasible? We could include subtitle in the text component because they have the same max width but should teams still put other components underneath the subtitle? Happy to jump on call to talk through options.

image


Secondary actions at small breakpoints:
They’re currently not collapsing in the example which I do think could be up to teams how to handle them at small break points. I can include those guidelines within Design docs but maybe for the final storybook example we take out the dropdown and leave the tag for simplicity. 
image

annawen1:
@kingtraceyj 
> Spacing:
> It would be nice when the subtitle and body copy is present to have the spacing between subtitle and text to be $spacing-03 Would this be feasible? We could include subtitle in the text component because they have the same max width but should teams still put other components underneath the subtitle? Happy to jump on call to talk through options.

Yep, I've included the subtitle inside the text component so now both the subtitle and body text are within the same max-width. Also added the $spacing-03 under the subtitle. 

Teams are able to place components under the subtitle - if they want the components to be contained in the max-width they can pass the components as children to the text component. Otherwise they can place the components outside of the text component and they would fall under the subtitle and/or body text.

> Secondary actions at small breakpoints:
> They’re currently not collapsing in the example which I do think could be up to teams how to handle them at small break points. I can include those guidelines within Design docs but maybe for the final storybook example we take out the dropdown and leave the tag for simplicity.

Gotcha, I've removed the dropdown from contextual actions in storybook!

kingtraceyj (approved):
looks good to go! 💥

ariellalgilmore (approved):
LGTM!

heloiselui (approved):
LGTM!

packages/styles/scss/components/page-header/_page-header.scss
```diff
@@ -108,10 +130,15 @@
...
 
+  [data-hidden]:not([data-fixed]) {
+    display: none;
+  }
+
   .#{$prefix}--page-header__hero-image {
     display: flex;
     overflow: hidden;
     align-items: center;
+    justify-content: end;
```
a88zach, line 141: @annawen1  This should be changed to `flex-end` to remove autoprefixer warnings

`(28864:3) autoprefixer: end value has mixed support, consider using flex-end instead`

PR #19056, opened 2025-04-08
fix: update icon usage to use relative sizing

### LINKED ISSUE #18539: [Bug]: Icons not scaling with fonts
### Package

@carbon/react

### Browser

Chrome

### Package version

1.73.0

### React version

_No response_

### Description

Most of the icons are not responsive which is causing issues. When font size is scaled up, the icons should also be scaled.

## Components where I noticed that bug
- [Icon-only Button](https://react.carbondesignsystem.com/?path=/story/components-button--icon-button)
- [Icon Button](https://react.carbondesignsystem.com/?path=/story/components-iconbutton--default) 
- [Date Picker](https://react.carbondesignsystem.com/?path=/story/components-datepicker--default)
- [Combo Button](https://react.carbondesignsystem.com/?path=/story/components-combobutton--default)
- [Multi Select](https://react.carbondesignsystem.com/?path=/story/components-multiselect--default)
- [Notification](https://react.carbondesignsystem.com/?path=/story/components-notifications-actionable--default)
- [Number Input](https://react.carbondesignsystem.com/?path=/story/components-numberinput--default)
- [Password Input](https://react.carbondesignsystem.com/?path=/story/components-passwordinput--default)
- [Slider](https://react.carbondesignsystem.com/?path=/story/components-select--default)
- [Pagination](https://react.carbondesignsystem.com/?path=/story/components-paginationnav--default)
- [Tooltip](https://react.carbondesignsystem.com/?path=/story/components-tooltip--default)
- [Toggletip](https://react.carbondesignsystem.com/?path=/story/components-toggletip--default)

## Screenshot
Image

## Related issues
- https://github.com/carbon-design-system/carbon/issues/18374
- https://github.com/carbon-design-system/carbon/issues/18579

### Reproduction/example

https://stackblitz.com/edit/github-m4d6tvtu?file=src%2Findex.scss,src%2FApp.jsx,src%2Fmain.jsx&preset=node=

### Steps to reproduce

In Chrome go to **Settings** > **Appearance** > **Font Size** > **Very large**

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

heloiselui:
Closes #18539

Updated icons to use relative sizing (`size="1rem"`) to support font scaling and accessibility.  
In components where icons are passed via `renderIcon`, I switched to using an inline function to allow setting the icon size.  
> [!NOTE]  
> This update has been applied only to specific components for now. At the moment, I haven’t found any better alternatives (happy to hear what you all think), but if we validate this approach, we can apply it to more icons.

#### Changelog

**Changed**

- Updated icon usage to use `size="1rem"` instead of fixed pixel sizing  
- Refactored `renderIcon` props to use inline functions for flexibility (e.g. `renderIcon={() => <Icon size="1rem" />}`)  
- Applied updates only to components directly related to the issue  

#### Testing / Reviewing

Please check that icons scale correctly with font-size in the components listed below:

- [X] [Button - Icon-only]  
- [X] [Icon Button]
- [X] [Date Picker]
- [X] [Combo Button]
- [X] [Multi Select]
- [X] [Notification]  
- [X] [Number Input]
- [X] [Password Input]
- [X] [Pagination]
- [X] [Tooltip]
- [X] [Toggletip]

1. Open the components listed in Storybook.
2. In your browser settings (Chrome), increase the **font size** to **Large** or **Very Large**.
   - On Chrome: Settings → Appearance → Font size
3. Reload the Storybook page and check that the icons scale proportionally.

> [!IMPORTANT]  
> Based on this feedback below, we decided not to change the icon defaults. Instead, PR #19148 documents how to achieve responsive sizing using `rem` units.


tay1orjones (commented):
It's cool that this works. I forgot the `size` prop accepts a string! It's a good example of how a consumer can fix this today in their applications - take the current number they're using for `size`, divide it by 16 and add the `rem` unit. e.g. `32 / 16 = 2rem`

I've been thinking on this and I don't think we should modify the icon defaults in the build system. It would be a large systematic change with lots of churn.

Rather than updating all these examples, what if we documented this approach in a new story in the `Elements/Icons` section of the storybook? We could show both examples you have here, with and without the inline function. Explain that the default on icons is a unitless value and if you want it to respond to font size changes you'll have to use `rem` units. We don't have an icons.mdx (the `Elements/Icons/Overview` story is currently autogenerated by storybook) but we could make one to house all this info. What do you think?

heloiselui:
> It's cool that this works. I forgot the `size` prop accepts a string! It's a good example of how a consumer can fix this today in their applications - take the current number they're using for `size`, divide it by 16 and add the `rem` unit. e.g. `32 / 16 = 2rem`
> 
> I've been thinking on this and I don't think we should modify the icon defaults in the build system. It would be a large systematic change with lots of churn.
> 
> Rather than updating all these examples, what if we documented this approach in a new story in the `Elements/Icons` section of the storybook? We could show both examples you have here, with and without the inline function. Explain that the default on icons is a unitless value and if you want it to respond to font size changes you'll have to use `rem` units. We don't have an icons.mdx (the `Elements/Icons/Overview` story is currently autogenerated by storybook) but we could make one to house all this info. What do you think?

Thanks @tay1orjones for the feedback! Based on your suggestion, I created a new PR (#19148) that documents how to achieve responsive icon sizing using rem units. It includes both examples -using the size prop directly and the renderIcon function. Let me know what you think! I'll close this PR.

Issue #19038, opened 2025-04-07
[Bug]: Tooltip issue with large text without space


d-tejaswi:
### Package

@carbon/react

### Browser

Chrome

### Package version

1.79.0

### Description

![Image](https://github.com/user-attachments/assets/61a0a9a8-d9fe-4ddb-84ce-7b5fbbf97990)

![Image](https://github.com/user-attachments/assets/9e08480b-eb95-4f09-9941-3c8588c38601)

The tooltip contains large text without spaces, causing it to overflow in a weird manner.

Adding word-break: break-word to .cds--tooltip-content or .cds--popover-content resolves the issue.

A similar fix is needed here, as was done for https://github.com/carbon-design-system/carbon/issues/18161

### Reproduction/example

https://stackblitz.com/edit/k1kse93i?file=src%2FApp.jsx

### Steps to reproduce

<Tooltip
      label="Occasionally,servicesareupdatedinaspecifiedtimewindowtoensurenodowntimeforcustomers."
      align="bottom-left"
    >
      <Button>This button has a tooltip</Button>
    </Tooltip>

### Suggested Severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

PR #19058, opened 2025-04-08
fix(tooltip): wrap unbroken text to prevent overflow

heloiselui:
**Closes** #19038
Fixes an issue where long text without spaces inside a Tooltip would overflow instead of wrapping correctly.

#### Changelog

**New**
- Added`overflow-wrap: break-word` to `.cds--tooltip-content` to support line breaks in long unbroken text, because it is the [standard](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) and works well for breaking long words.

**Changed**
- Tooltip now handles long strings without breaking layout or overflowing.

#### Testing / Reviewing

To test this fix:

1. Go to the **Tooltip** story in Storybook.
2. Pass a long string without spaces as the `label` prop, for example:  
>`Loremipsumdolorsitamet,consecteturadipiscingelit,seddoeiusmodtemporincididuntutlaboreetdoloremagnaaliqua.Utenimadminimveniam,quisnostrudexercitationullamcolaborisnisiutaliquipexeacommodoconsequat.Duisauteiruredolorinreprehenderitinvoluptatevelitessecillumdoloreeufugiatnullapariatur.Excepteursintoccaecatcupidatatnonproident,suntinculpaquiofficiadeseruntmollitanimidestlaborum.`
3. Observe that the tooltip content wraps correctly and does **not overflow** its container.
4. Confirm that other tooltips still behave and look as expected.

Below is a screenshot showing the result after the fix:
Pasted Graphic 2


Gururajj77 (approved):
Looks good!

kennylam (approved):

Issue #18997, opened 2025-04-02
[Bug]: MenuButton does not always respect menuTarget


kcpru:
### Package

@carbon/react

### Browser

Chrome

### Package version

1.78.2

### Description

When assigning an element as `menuTarget`, the menu should render inside this element. However, the menu always renders near MenuButton, event if MenuButton is outside the assigned `menuTarget`. `menuTarget` is only respected once when its content is changed. After a page refresh, the assigned `menuTarget` is ignored, and the menu renders near `MenuButton` instead

### Reproduction/example

https://stackblitz.com/edit/github-ivpbinfa?file=src%2FExample%2FExample.jsx

### Steps to reproduce

1. Assing a element as `menuTarget`
2. Observer where the menu renders - it should appear inside `menuTarget`, but it renders near `MenuButton`
3. Change `menuTarget` content - the menu correctly updates its position
4. Refresh the page - `menuTarget` is ignored, and the menu renders near `MenuButton`

### Suggested Severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

2nikhiltom:
At first look I think storing menuTarget in a local ref with useEffect to track changes, then passing this ref to Menu instead of passing menuTarget directly might resolve this issue

kcpru:
I tried that approach, but unfortunately it didn't work

adamalston:
Can you check if the issue is still present in https://stackblitz.com/edit/github-aq8alsec?file=src%2FApp.tsx? I made a change that should ensure `menuTarget` is set after the ref is populated. It seems the issue is that it was being set before the `ref` was populated unless I'm misunderstanding https://github.com/carbon-design-system/carbon/issues/18997#issue-2965578406.

kcpru:
I've checked, seems to be working correctly now

2nikhiltom:
@adamalston  There was a connecting [issue](https://github.com/carbon-design-system/carbon-labs/issues/536) around theme and focus order which you can see in the original [stackblitz](https://stackblitz.com/edit/github-ivpbinfa?file=src%2FExample%2FExample.jsx ) which was used to report this issue, the first menuItem in the list is not focus by default when it opens
[...]

Issue #18935, opened 2025-03-27
[Feature Request]: Mount and unmount Modal/ComposedModal


timfuhrmann:
### The problem

As of right now `Modal`/`ComposedModal` are immediately mounted, even if closed. This is a bottleneck in terms of performance and state management. For example:

1. Form states within in modals are not reset
2. Open state has to be forwarded to children to delay any data fetching
3. Huge data tables with overflow menus are mounting a lot of modals
4. Virtual lists will unmount and mount modals and create portals very frequently

That's why I have seen many teams either conditionally render its children or the entire modal. Both approaches are not optimal:
1.  Conditionally rendering the children will remove the content before the exit animation has ended and the wrapper will still be mounted, the portal sill created
2. Conditionally rendering the modal will remove all animations

### The solution

I assume the component is mounted incoherent to the open state for the enter/exit animation. Hence I would suggest to utilise the HTML element's animation event listeners to delay the unmount until the animation has finished. 

### Examples

1. Radix is using `<Presence />`: https://github.com/radix-ui/primitives/blob/main/packages/react/presence/src/presence.tsx, see [popover](https://github.com/radix-ui/primitives/blob/6e75e117977c9e6ffa939e6951a707f16ba0f95e/packages/react/popover/src/popover.tsx#L196).
2. React Aria is using `useEnterAnimation` and `useExitAnimation`: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/utils/src/animation.ts, see [popover](https://github.com/radix-ui/primitives/blob/6e75e117977c9e6ffa939e6951a707f16ba0f95e/packages/react/popover/src/popover.tsx#L196).

### Package

@carbon/react

### Business priority

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

timfuhrmann:
@tay1orjones Thanks for getting back to me on this so fast! 

> It uses a new internal `Dialog` component that uses `@starting-style` for the animations, which supports [transitioning elements on DOM addition and removal](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style#transitioning_elements_on_dom_addition_and_removal).

[...]

Issue #16883, opened 2024-06-27
[Feature Request]: Stop using a portal for `Menu`


tay1orjones:
### The problem

The Menu primitive uses a portal to render the `ul` list. This was added in https://github.com/carbon-design-system/carbon/pull/8829 to: 

> avoid it being misplaced, cut off or invisible in certain setups (such as the DataTable toolbar) 

Those issues may be solved now by the new autoAlign feature using floating ui, https://github.com/carbon-design-system/carbon/issues/15521.

The primary reason for this is that it's a risk of failing WCAG [1.3.2: Meaningful Sequence](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence.html) (Level A.) Specifically, [Failure of Success Criterion 1.3.2 due to changing the meaning of content by positioning information with CSS](https://www.w3.org/WAI/WCAG21/Techniques/failures/F1).

### The solution

Explore removing the portal in favor of rendering the menu adjacent to the trigger in the DOM and determine if floating-ui solves all the use cases for positioning and overlaying wrt Menu. This includes MenuButton, OverflowMenu, ContextMenu, etc. This probably will need to utilize the native Popover API as well so it gets promoted to the top layer. There could be some difficulties with this though, see #17017 

This could be put behind a flag for now and included by default in the next major.

#### Workarounds

~Short of this https://github.com/carbon-design-system/carbon/issues/16882 should at least be fixed so the portal destination is configurable.~  

* Menu's `target` prop can be used to control what element the menu is portalled into
* Menu-based components OverflowMenu and MenuButton allow configuration of the target via the `menuTarget` prop, https://github.com/carbon-design-system/carbon/pull/17145

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

adamalston:
Possibly related: https://github.com/carbon-design-system/carbon/pull/19059#issuecomment-2795702883 and issue linked in that comment.

jainpranay20:
Hi @alina-jacob @tay1orjones, this is blocking our current feature delivery, and we are waiting for the fix. How soon can we expect the fix?

jainpranay20:
Hi @alina-jacob do we have any alternative?

wkeese:
[...]

PR #19059, opened 2025-04-08
feat(MenuButton): match menu parent theme

### LINKED ISSUE #536: [Platforms]: focus issue with MenuButton when using menuTarget in Header
Related to https://github.com/carbon-design-system/carbon-labs/pull/465

There are a few things going on here to look into, without the menuTarget prop the Menu renders outside of the correct theme. When adding the prop it renders in the correct theme, however the focus is incorrect. When you open the MenuButton the focus should go to the first item. The Header component also doesn't currently forward the ref, so the menuTarget is null. 

```jsx

.....
<MenuButton
  menuTarget={headerRef.current}
  .....>
  ......
</MenuButton>
```

We need to MenuButton to open with the focus on the first item, in the correct theme. This may need to be fixed upstream in Carbon.

### LINKED ISSUE #537: [Platforms]: HeaderMenu sometimes doesn't close via outside click event
I noticed a weird behavior clicking around with the menus:

https://github.com/user-attachments/assets/cbc21f2e-1452-434a-8c4f-f3519792aad2

To reproduce:
1. First click the `MenuButton`
2. Click the `HeaderMenu`, the `MenuButton` closes
3. Click the `MenuButton` again, `HeaderMenu` does not close
4. `HeaderMenu` can't be closed via outside click events but `MenuButton` does. Only upon clicking the menu again will it close

*This might be an upstream Carbon bug, appears to possibly only be in Chrome

IgnacioBecerra:
Closes https://github.com/carbon-design-system/carbon-labs/issues/536
Closes https://github.com/carbon-design-system/carbon-labs/issues/537

This PR introduces a new prop for `MenuButton` and `Menu` called `matchTheme`. 

In `@carbon-labs`, for the [Default UI Shell story](https://labs.carbondesignsystem.com/react/?path=/story/components-uishell--default), upon opening the `MenuButton` labeled `Select Category`, the focus is directed to the new `Menu` itself rather than the first `MenuItem`.

This is due to the Default story using `target={headerRef.current}` for the `MenuButton`, referring to the `HeaderContainer` housing the component. Removing it means the menu renders as the last element in the `document.body`, fixing this focus issue, as seen in the [Header story](https://labs.carbondesignsystem.com/react/?path=/story/components-uishell--header-story). 

However there is the side effect of the menu not rendering with the correct parent theme, as it's being render outside of its parent's theme scope. I'm not entirely sure why this fixes the focus error, but for the moment, this change allows for the menu to match its parent's theme.

#### Changelog

**New**

- added `matchTheme` boolean prop to `MenuButton` and `Menu`

#### Testing / Reviewing
Go to the new `With matched theme` story under `MenuButton`, and the menu will display in it's parent's theme. Note that the button itself is still the default kind.


alisonjoseph:
@IgnacioBecerra I think you need to run `yarn test -u` to fix the failing test / update the snapshot

alisonjoseph:
This appears to be working as intended, however I'm wondering if its needed if we are able to solve this [bug](https://github.com/carbon-design-system/carbon/issues/19076) around the focus and the `menuTarget` prop.

IgnacioBecerra:
@alisonjoseph The way I understand it, the `menuTarget` prop is so the menu doesn't render behind a modal by default. If `menuTarget={modalRef.current}`, the focus would work as intended within the modal as it has focus wrap. I'm assuming the focus is weird in our use case due to this, but I am still not 100% sure as to why it happens. I haven't been able to find out the main reason behind it.

And if we're not specifying the `menuTarget`, it does work properly, but the Header theme doesn't get carried over to the `Menu` component as it's rendered outside the header DOM-wise.

So I believe the focus "bug" isn't a bug per se, but rather we're using it incorrectly. I think it's meant to render in `document.body` and not within another component, so this new prop would still deliver the correct parent theme to it.

I'd like to hear the other devs' thoughts on this though!

adamalston:
I think there's a bug with the focus behavior.

Using the `menuTarget` prop causes the menu to render in a portal rather than inline. Because of this rendering method, `focusItem` runs before the menu items are fully mounted and their refs are attached. As a result, when the code attempts to focus the first menu item, it fails to find it. Without `menuTarget`, the menu items are rendered inline, and their refs are available in time for the focus logic to work correctly.

I wonder if the fix here is to delay `focusItem` (using `setTimeout` or `requestAnimationFrame`) which should allow React to finish rendering and attaching refs to all menu items.

CI for https://github.com/carbon-design-system/carbon/commit/42d2702a3a236c6ba40d181954130ded2c590a2d passed and I can't reproduce https://github.com/carbon-design-system/carbon/issues/19076 in the test story in that commit. I suppose I could open a PR for that unless someone sees something inherently wrong with it or has other thoughts.

Side note: the JSDoc comment for `target` in `MenuProps` is not correct.

tay1orjones:
There's lots of issues related to this. @2nikhiltom is working on https://github.com/carbon-design-system/carbon/issues/18997, and we've discussed the idea of removing portal usage from Menu in https://github.com/carbon-design-system/carbon/issues/16883

There's also been discussion in https://github.com/carbon-design-system/carbon/issues/18935 of a helper that maybe could be used to help here?

In any case I think it would be best if we could find another way to fix this without having to introduce this new prop to conditionally wrap it in a `Theme`.

2nikhiltom:
I feel removing the portal rendering and instead leveraging Floating UI for positioning in Menu would fix should fix majority of bugs  like Focus Management, MenuTarget inconsistency and themes related issues. 
 z-index and overflow benefits portals gave us can be implemented using Floating UI middleware

PR #19068, opened 2025-04-08
fix(number-input): min/max step value bug

### LINKED ISSUE #18842: [Bug]: NumberInput enforces min/max only on steppers if both values are set
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.78.0

### React version

v18

### Description

The NumberInput only applies a limit set by "min" if also "max" is set.
If max is missing, clicking on "-" sets the input to values smaller than min.

This is a change to the behaviour in v1.77.0

### Reproduction/example

https://stackblitz.com/edit/github-2hf2p8dq?file=src%2FApp.tsx

### Steps to reproduce

Add a `NumberInput` and set only `min={0}`.
Clicking the steppers it is possible to go to values less than 0.
In earlier versions, the stepper stopped at 0.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

ariellalgilmore:
Closes #18842 

NumberInput re-inforce if only one max or min is set then using the steppers will still block the numbers from increasing/decreasing past the min/max value

#### Changelog

**Changed**

- combine the new work done with floatValue and the old work where we would inforce min/max separetly 

#### Testing / Reviewing

make sure all tests pass
go to storybook
if only setting min value then check if using the "-" stepper that you can't go past the min value


packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -372,10 +372,14 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
...
-          typeof min !== 'undefined' && typeof max !== 'undefined'
-            ? Math.min(Math.max(floatValue, min), max)
-            : floatValue;
+        let newValue = floatValue;
+        if (typeof min !== 'undefined') {
+          newValue = Math.max(newValue, min);
+        }
+        if (typeof max !== 'undefined') {
+          newValue = Math.min(newValue, max);
+        }
```
adamalston: This functionality is similar to `clamp` in packages/react/src/components/Slider/Slider.tsx. Would it be better to abstract a `clamp` function into a utils file (i.e., packages/react/src/internal/utils.ts like I've done in https://github.com/carbon-design-system/carbon/pull/18971) and import it for use here and in packages/react/src/components/Slider/Slider.tsx (deleting the `clamp` function that's already in that file)?

```tsx
const clamp = (value: number, min?: number, max?: number) =>
  Math.min(max ?? Infinity, Math.max(min ?? -Infinity, value));
```

```suggestion
        const newValue = clamp(floatValue, min, max);
```

2nikhiltom (approved):
clamp could be reused . LGTM otherwise

packages/react/src/components/NumberInput/NumberInput.tsx
ariellalgilmore: thanks updated!

annawen1 (approved):

PR #19085, opened 2025-04-10
fix: fixed the onchange firing undefined 

### LINKED ISSUE #18361: [Bug]: Toggling the same option in ComboBox dropdown doesn't always trigger the onChange event

Gururajj77:
Closes #18361 

bug fixes on the combobox
- silenced the `onChange` firing with `undefined`
- now, mouse and keyboard interactions won't fire `onChange`, if the user selects an already selected it.
- this holds good for both with and without `allowCustomValue` prop.

#### Changelog

**Changed**

- added `isManualClearingRef` to keep track of the user clearing the items using the crossmark (X).
- rewrote the `InputKeydownEnter` case in the reducer, to handle the selections accordingly for both scenarios of with and without `allowCustomValue`
- added the `undefined` guards in the `onStateChange` to make sure it doesn't fire `onChange` with `undefined` which triggers other parts of the code.
- rewrote the `clearSelection` function that is in the `<ListBoxSelection />` to fire onChange with `null`.

#### Testing / Reviewing

- there are two test stores, "Undefined Checks Testing" and "Undefined Checks with Allow Custom Value"
- In both of the test stories check for these interactions:
   - Open the console
   - select and item using keyboard
   - see that console logs with that selected item in `selectedItem: <your-selected-item>`
   - select the same item again in the list
   - see that `onChange` doesn't fire again with that same item
   - select another item
   - see that `onchange` fires with that new selected item.
   - perform the same interactions using the mouse
   - play around the buttons to test the fully controlled version also.
- Make sure that in the two test stories
   - In the "Undefined Checks Testing" story, when you type custom value and close the combobox, it clears off that input
   - In the "Undefined Checks with Allow Custom Value", when you type custom value and close the combobox, the input stays the same without clearing the value, and an `onChange` is fired with `{ selectedItem: null, inputValue: "<your-custom-value"`
- Delete the two test stories after approval, before merging.


packages/react/src/components/ComboBox/ComboBox.stories.js
```diff
@@ -15,6 +15,7 @@ import { IconButton } from '../IconButton';
 import { View, FolderOpen, Folders } from '@carbon/icons-react';
 import { action } from '@storybook/addon-actions';
 import mdx from './ComboBox.mdx';
+import { disable } from '@carbon/feature-flags';
```
adamalston:
```suggestion
```

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -817,16 +829,21 @@ const ComboBox = forwardRef(
...
+          newSelectedItem !== undefined
         ) {
           onChange({ selectedItem: newSelectedItem });
         }
         if (
           (type === useCombobox.stateChangeTypes.FunctionSelectItem ||
             type === useCombobox.stateChangeTypes.InputKeyDownEnter) &&
-          !isEqual(selectedItemProp, newSelectedItem) // Only fire if there's an actual change
+          !isEqual(selectedItemProp, newSelectedItem) &&
+          newSelectedItem !== undefined
```
adamalston: https://stackoverflow.com/a/2703122/11106650

```suggestion
          typeof newSelectedItem !== 'undefined'
```

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1094,7 +1111,14 @@ const ComboBox = forwardRef(
             {inputValue && (
               <ListBoxSelection
                 clearSelection={() => {
+                  isManualClearingRef.current = true;
+                  setInputValue('');
+                  onChange({ selectedItem: null });
                   selectItem(null);
+                  setTimeout(() => {
+                    isManualClearingRef.current = false;
+                  }, 0);
```
adamalston: Does using a mutable ref together with a `setTimeout` to manage state introduce a race condition given the ordering of events may not be constant?

Also, the fact that the flag reset is scheduled via `setTimeout` means that there is a short window during which events are ignored. If another change happens during that gap, there may be some unexpected behavior here.

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -817,16 +829,21 @@ const ComboBox = forwardRef(
...
           selectedItem: newSelectedItem,
         });
+        if (isManualClearingRef.current) {
+          return;
+        }
         if (
           type === useCombobox.stateChangeTypes.ItemClick &&
-          !isEqual(selectedItemProp, newSelectedItem)
+          !isEqual(selectedItemProp, newSelectedItem) &&
+          newSelectedItem !== undefined
```
adamalston: https://stackoverflow.com/a/2703122/11106650

```suggestion
          typeof newSelectedItem !== 'undefined'
```

tay1orjones:
@Gururajj77 were you able to find a different approach that doesn't use a `setTimeout`? 

Also a reminder that this needs tests covering the interactions you describe in the PR body and in the originally linked issue.

Gururajj77:
> @Gururajj77 were you able to find a different approach that doesn't use a `setTimeout`?
> 
> Also a reminder that this needs tests covering the interactions you describe in the PR body and in the originally linked issue.

Yes, I've figured out a way to make it work without `setTimeout`. It's using a combination of state and ref. Added some tests for them to make sure the functionality is covered.

adamalston:
When I revert all of the code changes, the tests still pass. Shouldn't they be failing?

```sh
git checkout combobox-onchange-again
git checkout main packages/react/src/components/ComboBox/ComboBox.tsx
yarn test packages/react/src/components/ComboBox/ComboBox-test.js

**Comment by Gururajj77** [comment]
> When I revert all of the code changes, the tests still pass. Shouldn't they be failing?
> 
> ```shell
> git checkout combobox-onchange-again
> git checkout main packages/react/src/components/ComboBox/ComboBox.tsx
> yarn test packages/react/src/components/ComboBox/ComboBox-test.js
> ```

Rewritten the tests, and added cases for both cases where the items are array of strings, and array of objects, to make sure all cases are covered.

**Review by tay1orjones — CHANGES_REQUESTED** [review]
Listbox selection is not reflected in the list after clicking "controlled" button

https://github.com/user-attachments/assets/2cb8955f-0f2c-4925-b085-b0c2329c8229

Additionally in this same flow, it will eventually stop responding to the "controlled" buttons altogether


https://github.com/user-attachments/assets/5118a11d-8e4e-486a-b1b1-e11afc1e5583



Listbox selection shows old selected value after clicking "controlled" button

https://github.com/user-attachments/assets/657f3f51-16b0-4804-82d5-707bd1bcf21c

These all seem to persist with allowCustomValue on as well.

**Review by tay1orjones — APPROVED** [review]
I moved my previous findings to that new issue

File: packages/react/src/components/ComboBox/ComboBox-test.js [hunk 5]
```diff
@@ -1130,4 +1130,580 @@ describe('ComboBox', () => {
...
+          {
+            id: 'option-5',
+            text: 'Option 5',
+          },
+        ];
+
+        const handleChange = ({ selectedItem }) => {
+          setSelectedItem(selectedItem);
+          setChangeCount((prev) => prev + 1);
+        };
```
adamalston: This `items` variable and the `handleChange` are duplicated throughout many tests in this file. Can either of them be abstracted and reused?

packages/react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -817,16 +838,21 @@ const ComboBox = forwardRef(
...
         }
         if (
           (type === useCombobox.stateChangeTypes.FunctionSelectItem ||
             type === useCombobox.stateChangeTypes.InputKeyDownEnter) &&
-          !isEqual(selectedItemProp, newSelectedItem) // Only fire if there's an actual change
+          !isEqual(selectedItemProp, newSelectedItem) &&
+          typeof newSelectedItem !== 'undefined'
         ) {
           onChange({ selectedItem: newSelectedItem });
         }
```
adamalston: Performance suggestion to consolidate these conditions. `type` is mutually exclusive and both blocks share the same logic. Combining them reduces redundancy and avoids running duplicate checks unnecessarily, especially since an `if...else` isn't being used.

```diff
diff --git a/packages/react/src/components/ComboBox/ComboBox.tsx b/packages/react/src/components/ComboBox/ComboBox.tsx
index 0c7e5d85f..fdfefc433 100644
--- a/packages/react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/react/src/components/ComboBox/ComboBox.tsx
@@ -59,6 +59,8 @@ const {
   InputKeyDownArrowUp,
   InputKeyDownArrowDown,
   MenuMouseLeave,
+  ItemClick,
+  FunctionSelectItem,
 } = useCombobox.stateChangeTypes;
 
 const defaultItemToString = <ItemType,>(item: ItemType | null) => {
@@ -842,17 +844,12 @@ const ComboBox = forwardRef(
           return;
         }
         if (
-          type === useCombobox.stateChangeTypes.ItemClick &&
-          !isEqual(selectedItemProp, newSelectedItem) &&
-          typeof newSelectedItem !== 'undefined'
-        ) {
-          onChange({ selectedItem: newSelectedItem });
-        }
-        if (
-          (type === useCombobox.stateChangeTypes.FunctionSelectItem ||
-            type === useCombobox.stateChangeTypes.InputKeyDownEnter) &&
-          !isEqual(selectedItemProp, newSelectedItem) &&
-          typeof newSelectedItem !== 'undefined'
+          (type === FunctionSelectItem ||
+            type === InputKeyDownEnter ||
+            type === ItemClick) &&
+          typeof newSelectedItem !== 'undefined' &&
+          !isEqual(selectedItemProp, newSelectedItem)
         ) {
           onChange({ selectedItem: newSelectedItem });
         }
```

packages/react/src/components/ComboBox/ComboBox-test.js
Gururajj77: I moved the `ControlledCombobox` outside tests, under the `describe` scope, so that we can reuse the whole component itself.

packages/react/src/components/ComboBox/ComboBox.tsx
Gururajj77: I have considered these improvements, with some subtle change in using the `isManualClearingRef`, @adamalston, please check the changes

2nikhiltom (approved):

packages/react/src/components/ComboBox/ComboBox.tsx
adamalston: Changes look good.

PR #19106, opened 2025-04-11
feat(Tag): add dismissTooltipLabel prop to DismissibleTag

### LINKED ISSUE #18517: [Bug]: Allow custom dismiss tooltip label in `DismissibleTag` when ellipsis applied
### The problem

Currently if ellipsis is applied, the hard-coded 'Dismiss' string will be added to the tooltip label.

https://github.com/carbon-design-system/carbon/blob/2f99a479550f3bc52636d65f816cd3ddc3e26207/packages/react/src/components/Tag/DismissibleTag.tsx#L150

https://github.com/carbon-design-system/carbon/blob/2f99a479550f3bc52636d65f816cd3ddc3e26207/packages/react/src/components/Tag/DismissibleTag.tsx#L177-L178

### The solution

Allow the component to receive custom label to be used instead of `dismissLabel` const.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

### Tasks

- [ ] React Implementation

heloiselui:
Closes #18517

Add a new prop to customize the tooltip and aria-label for the dismiss button on the DismissibleTag component.

> [!NOTE]  
> I created a separate PR for web-component.

#### Changelog

**New**
- Added `dismissTooltipLabel` prop to `DismissibleTag` component.

**Changed**
- The tooltip and `aria-label` now use `dismissTooltipLabel` when text is truncated (ellipsis).

#### Testing / Reviewing
- CI should pass

To test this in Storybook (Deploy Preview):
1. Go to `Tag > Dismissible` story.
2. Set a long `text` value (so it's truncated).
4. Hover over the close button (`×`) and check:
   - The tooltip shows your custom message.
   - The button’s `aria-label` matches the custom message (check via dev tools).

Below is a screenshot to show the result:
screenshot


packages/react/src/components/Tag/DismissibleTag.tsx
```diff
@@ -161,6 +167,10 @@ const DismissibleTag = forwardRef(
 
     const dismissLabel = `Dismiss "${text}"`;
 
+    const dismissActionLabel = isEllipsisApplied
+      ? dismissTooltipLabel || dismissLabel
```
adamalston: Should a nullish coalescing operator be used to ensure that falsy values outside of `null` and `undefined` are respected?

```suggestion
      ? dismissTooltipLabel ?? dismissLabel
```

heloiselui: Thanks for the suggestion, I intentionally used || instead of ?? to fallback on any falsy value like empty strings, ensuring we avoid blank tooltips in the UI.

packages/react/src/components/Tag/DismissibleTag.tsx
```diff
@@ -110,6 +115,7 @@ const DismissibleTag = forwardRef(
       text,
       tagTitle,
       type,
+      dismissTooltipLabel,
```
tay1orjones:
```suggestion
      dismissTooltipLabel = `Dismiss "${text}"`,
```

tay1orjones (commented):
Would adding the `Dismiss "${text}"` string as a default for the new prop make it so we don't have to alter any other component code? Like the conditionals wouldn't need revised at all this way, I think? 🤔

packages/react/src/components/Tag/DismissibleTag.tsx
heloiselui: While testing your suggestion, I noticed that when text is empty, it was rendering as Dismiss "".
To avoid that, I added a fallback:  `dismissTooltipLabel = text ? ``Dismiss "${text}"`` : 'Dismiss',`
I also simplified the logic by removing the extra `dismissLabel` variable, since the fallback is now handled in the default value directly.

preetibansalui (approved):

packages/react/src/components/Tag/DismissibleTag.tsx
```diff
@@ -110,6 +115,7 @@ const DismissibleTag = forwardRef(
       text,
       tagTitle,
       type,
+      dismissTooltipLabel = text ? `Dismiss "${text}"` : 'Dismiss',
```
kennylam: Should this prop have one default value, i.e. `Dismiss tooltip` or something similar? Otherwise `Dismiss` is hard-coded in English. Users would be responsible for any custom labels, including localization.

thyhmdo:
@laurenmrice so you can also read about this later. 

I like Heloise's approach to this, providing opt-in flexibility for those who need to localize or customize it. And also a good recommendation from Kenny because I see the same pattern (Action + Component name) in Tab (Remove [Name] tab) and Notification (Close notification), but it'd be "Dismiss tag" instead. 

So in short: 

- Keep the default as "Dismiss tag".
- Let users override it via `dismissTooltipLabel`

----
**Additional considerations for later:**

Inconsistent current labels of other components with the same action: 
- Dismiss – Tag
- Remove – Tab  (Remove, it makes sense but we call it dismisable) 
- Close – Notification

image
image
image

kennylam (approved):
LGTM thanks @heloiselui!

PR #19120, opened 2025-04-14
fix(MenuButton): do not steal focus on keyboard click of MenuItem

### LINKED ISSUE #18418: [Bug]: Incorrect focus behavior for MenuButton and MenuItem components
### Package

@carbon/react

### Browser

Chrome

### Package version

1.74.0

### React version

18.0.0

### Description

We’ve encountered an accessibility issue with the MenuButton and MenuItem components in the Carbon library. Specifically, when navigating using the keyboard, the focus is returned to the MenuButton after a menu selection. This behavior is problematic for our use case, where the focus should instead move to a specific input element within a Tearsheet following a menu selection. Mouse interactions work as expected.

The current implementation appears to explicitly set focus back to the MenuButton on the menu close event, which disrupts the navigation flow for keyboard users.

### Reproduction/example

https://stackblitz.com/edit/github-6ufyydgb-d1phig1l?file=src%2FApp.jsx,package.json,src%2Fmain.jsx

### Steps to reproduce

Render a MenuButton and MenuItem and Tearsheet (with primaryFocus for input) from IBM Products with input element
Selecting MenuItem has to open the tearsheet
Navigate the menu using the keyboard and select an item.
Observe that the focus returns to the MenuButton instead of the intended input field within the Tearsheet.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

wkeese:
Closes #18418.

When you click a MenuItem, the Menu closes and focus goes back to the MenuButton. This makes sense except when the MenuItem's onClick() callback already focused somewhere else.  The typical case is when the MenuItem opened a Modal or Tearsheet.

Actually, this was already working when the user clicked the MenuItem with the mouse. The MenuButton was only stealing focus when the MenuItem was clicked via keyboard. In that case, MenuItem's onClick() ran on keydown, and the Menu closed on keydown, but the refocus() call didn't happen until keyup.

This PR changes MenuItem so the "click" happens on "keyup" rather than on "keydown". So nothing happens on keydown, but on keyup, the following events happen in this order:

1. Refocus the MenuButton.
2. Close the Menu dropdown.
3. Execute onClick() callback, which may (or may not) focus somewhere else.

Rationale: When working on dijit, I found that "click" event on native elements like BUTTON happened on "keyup", so it made sense to do it that way for DIV too.  Otherwise, IIRC, if you change focus on keydown, and the newly focused element gets the keyup event, then bad things can happen.  see https://github.com/dojo/dijit/blob/master/a11yclick.js.  

See related #8829, which added the initial code for closing the menu on keyup rather than keydown.  Maybe that code was to fix the same problem I listed above.

#### Changelog

**Changed**

- Changed MenuItem onClick() to run on keyup rather than keydown, so that the MenuButton is focused before the MenuItem's onClick callback runs.

#### Testing / Reviewing

I tested by modifying MenuButton.stories.js as follows, and then testing both mouse and keyboard:

```jsx
export const Default = (args) => {
  return (
    <>
      <MenuButton {...args} onClick={action('onClick')} label="Actions">
        <MenuItem
          label="First action with a long label description"
          onClick={() => { document.getElementById("my-input").focus(); }}
        />
        <MenuItem label="Second action" onClick={action('onClick')} />
        <MenuItem label="Third action" onClick={action('onClick')} disabled />
      </MenuButton>
      <input id="my-input"/>
    </>
  );
};
```


annawen1:
@wkeese This looks good to me, works locally when I tested it out with the given storybook change - could you add a test for this scenario?

wkeese:
@annawen1 - Sure... you're talking about the plain Jest tests (MenuButton-test.js) rather than the E2E tests?

wkeese:
MenuButton didn't have any keyboard tests.  I added a basic test and a test for then the MenuItem's onClick changes the focus.

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -181,7 +181,13 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
...
 
       if (match(e, keys.Enter) || match(e, keys.Space)) {
-        handleClick(e);
+        e.target.addEventListener(
+          'keyup',
+          (e) => {
+            handleClick(e as unknown as React.KeyboardEvent<HTMLLIElement>);
+          },
+          { once: true }
+        );
```
adamalston: Should this listener be removed via `removeEventListener` at some point?

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -181,7 +181,13 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
       }
 
       if (match(e, keys.Enter) || match(e, keys.Space)) {
-        handleClick(e);
+        e.target.addEventListener(
+          'keyup',
+          (e) => {
+            handleClick(e as unknown as React.KeyboardEvent<HTMLLIElement>);
```
adamalston: Is there a type safe way to do this so that type assertions aren't necessary? It does seem odd that a function named `handleClick` is being called with a keyboard event but that's not a result of anything you've done.

packages/react/src/components/Menu/MenuItem.tsx
wkeese: That was necessary in the old days but now we can just use `{once: true}`, see https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once.

packages/react/src/components/Menu/MenuItem.tsx
wkeese: Actually there was a deeper issue there, about the different between react events (like React.KeyboardEvent) and events on native DOM elements.

I refactored the code to set up the onKeyUp listener with React, rather than directly calling addEventListener().

I also expanded the keyboard tests as I'm finding it pretty tricky to deal with click events (and keyboard) on a plain node like LI instead of BUTTON.  I actually think it would be better to rewrite MenuItem to render a BUTTON, but I didn't do that as part of this PR.

packages/react/src/components/MenuButton/MenuButton-test.js
```diff
@@ -115,6 +115,102 @@ describe('MenuButton', () => {
       expect(screen.getByRole('menu')).toBeInTheDocument();
       expect(screen.getByRole('menuitem')).toHaveTextContent(/^Action$/);
     });
+
+    it('has basic keyboard support', async () => {
```
tay1orjones, line 119: This looks good and I don't expect you to make changes related to this, but want to mention that we've started putting keyboard navigation tests in playwright e2e tests. 

It ensures keyboard nav is tested in an environment that resembles an actual user's environment as closely as possible. It's also way more debuggable since you can run them locally in a visible browser. 

Colocating them w/ other a11y tests has also allowed us to use the playright test reporter to populate component a11y info on carbondesignsystem.com

All that said, the e2e test we had for Menu keyboard nav was incredibly flaky and we turned it off. 

https://github.com/carbon-design-system/carbon/blob/eab657f4f6ea42d9934bdd0936677a5220a26dbc/e2e/components/Menu/Menu-test.avt.e2e.js#L25

I'd like to see if we could bring this test over there to supplement or replace the existing one. You can take a look at it if you'd like, but no worries if not, I can open an issue for it after this merges.

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -174,21 +174,35 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
...
         rest.onKeyDown(e);
       }
     }
 
+    function handleKeyUp(e: KeyboardEvent<HTMLLIElement>) {
+      if (pendingKeyboardClick && keyboardClickEvent(e)) {
+        handleClick(e);
+      }
+
+      setPendingKeyboardClick(false);
```
tay1orjones: I was curious if we should be concerned here with react batching these updates between handleKeyUp and handleKeyDown. 

React [docs](https://react.dev/learn/queueing-a-series-of-state-updates#:~:text=React%20waits%20until%20all%20code%20in%20the%20event%20handlers%20has%20run%20before%20processing%20your%20state%20updates) say 

> React waits until all code in the event handlers has run before processing your state updates.

but then later [seems to convey](https://react.dev/learn/queueing-a-series-of-state-updates#:~:text=React%20does%20not%20batch%20across%20multiple%20intentional%20events%20like%20clicks%E2%80%94each%20click%20is%20handled%20separately) this is fine and will be stable.

> React does not batch across multiple intentional events like clicks—each click is handled separately

tay1orjones (approved):
Thanks for updating to use react events vs direct listeners. LGTM!

wkeese:
@tay1orjones - Thanks for the comments and explanation.

I think all the issues you mentioned have the same root cause: `MenuItem` is reinventing the wheel by simulating a button using `<li>` (instead of `<button>`).   With a `<button>`, we would just set up a click handler and there would be no race condition.

Maybe I'm oversimplifying though, as that doesn't explain Playwright's flakiness dealing with the keyboard **arrow** keys.

In any case, I pushed another change to eliminate any potential race condition due to the delay of `setPendingClick()`.

In principal, I agree it's nice to have the tests in Playwright so you can see what's going on.  Relatedly, it's nice to have the tests in Storybook, for the same reason (i.e. easy debugging).  I'm reluctant to move my new tests though since Playwright is being flaky.   So let's open a separate ticket for that?

annawen1 (approved):

PR #19122, opened 2025-04-14
wc: fix(tag): align tooltip logic and a11y with React

### LINKED ISSUE #18517: [Bug]: Allow custom dismiss tooltip label in `DismissibleTag` when ellipsis applied

heloiselui:
Closes #18517

Fixes the tooltip behavior on `<cds-dismissible-tag>` to match the logic in the React implementation.

> [!NOTE]  
> I created a separate [PR](https://github.com/carbon-design-system/carbon/pull/19106) for React.

#### Changelog

**New**
- Added `dismissTooltipLabel` property

**Changed**
- Updated `dismissActionLabel` logic in `<cds-dismissible-tag>` to show:
  - `Dismiss "Label"` when ellipsis/truncation is applied
  - `Dismiss` when not
- Removed native `title` attribute from close button to avoid tooltip overlap
image

#### Testing / Reviewing
- CI should pass

To test this in Storybook (Deploy Preview):

1. Go to Tag > Dismissible story.
2. Set a long text value (so it's truncated).
3. Hover over the close button (×) and check:
   - The tooltip shows your custom message.
image

### DIFF
--- a/packages/web-components/src/components/tag/dismissible-tag.ts
+++ b/packages/web-components/src/components/tag/dismissible-tag.ts
@@ -100,16 +100,16 @@ class CDSDismissibleTag extends HostListenerMixin(FocusMixin(CDSTag)) {
   };
 
   /**
-   * Text to show on tag "close" buttons
-   */
-  @property({ type: String, reflect: true })
-  title = 'Dismiss';
-
-  /**
    * `true` if the tag should be disabled
    */
   @property({ type: Boolean, reflect: true })
   disabled = false;
+
+  /**
+   * Provide a custom tooltip label for the dismiss button
+   */
+  @property({ type: String, attribute: 'dismiss-tooltip-label', reflect: true })
+  dismissTooltipLabel?: string;
 
   /**
    * `true` if the tag should be open.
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/tag/dismissible-tag.ts
```diff
@@ -169,13 +171,13 @@ class CDSDismissibleTag extends HostListenerMixin(FocusMixin(CDSTag)) {
           <button
             class="sb-tooltip-trigger"
             role="button"
-            aria-labelledby="content"
+            aria-label="${dismissActionLabel}"
```
tay1orjones: Is aria-label what we typically do here or could this remain `aria-labelledby="content"` since `dismissActionLabel` is rendered in the `#content`?

tay1orjones (approved):
Just one question

packages/web-components/src/components/tag/dismissible-tag.ts
heloiselui: Thanks for pointing that out! I'll use aria-labelledby="content" because the label text is already rendered in #content, and it follows the pattern used in other components as well.

2nikhiltom (approved):

PR #19081, opened 2025-04-09
refactor: consolidate `Popover` `align` mapping functionality

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/19079

Consolidated `Popover` `align` mapping functionality.

#### Changelog

**Removed**

- Consolidated `Popover` `align` mapping functionality.
- Deleted things in packages/react/src/tools/createPropAdapter.js that weren't used anywhere.

#### Testing / Reviewing

```sh
yarn test packages/react
```

PR #19131, opened 2025-04-14
fix(overflowmenu): added typing to align prop

### LINKED ISSUE #18957: [Bug]: The `OverflowMenu` `align` prop isn't typed
### Package

@carbon/react

### Browser

_No response_

### Package version

1.79.0

### React version

_No response_

### Description

The `OverflowMenu` `align` prop isn't typed even though it's used the code in https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--overview#customizing-the-tooltip.

### Reproduction/example

https://stackblitz.com/edit/github-zwvqaxgz?file=src%2FApp.tsx

### Steps to reproduce

See type error in reproduction.

### Suggested Severity

### Application/PAL

DataPower

### Code of Conduct

jose-biescas:
Closes #18957 

Adds a type to the `align` prop in `OverflowMenu`

#### Changelog

**New**

- Added `align` property

#### Testing / Reviewing

Go to the `Default` storybook for `OverflowMenu` and the `align` property can be controlled via a dropdown with allowed prop values


jose-biescas:

jose-biescas:
recheck

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -120,7 +121,59 @@ export const getMenuOffset: MenuOffset = (
...
+    'right-bottom': 'right-end',
+    'right-top': 'right-start',
+  };
+  return mapping[deprecatedValue];
+};
+
+type IconProps = {
+  className?: string;
+  'aria-label'?: string;
+};
```
adamalston: What's this `type` for?

tay1orjones: Is this needed?
```suggestion
```

packages/react/src/components/OverflowMenu/OverflowMenu.stories.js
```diff
@@ -51,6 +51,18 @@ Default.args = {
...
+  align: {
+    options: [
+      'top',
+      'top-left',
+      'top-right',
+      'bottom',
+      'bottom-left',
+      'bottom-right',
+      'left',
+      'right',
```
tay1orjones: I think these could be just the "new" values instead of the old to encourage use of the new ones.

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -120,7 +121,59 @@ export const getMenuOffset: MenuOffset = (
...
+  | 'bottom-start'
+  | 'bottom-end'
+  | 'left-end'
+  | 'left-start'
+  | 'right-end'
+  | 'right-start';
+
+export type OverflowMenuAlignment =
+  | DeprecatedOverflowMenuAlignment
+  | NewOverflowMenuAlignment;
```
adamalston: Suggestion to delete these types. You could use `PopoverAlignment` since this `align` prop is just being passed to `IconButton` which passes it to `Tooltip` which passes it to `Popover`.

For reference, I've consolidated all of this functionality in https://github.com/carbon-design-system/carbon/pull/19081 but until that goes in, I think things here could be simplified.

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -120,7 +121,59 @@ export const getMenuOffset: MenuOffset = (
...
+    'top-right': 'top-end',
+    'bottom-left': 'bottom-start',
+    'bottom-right': 'bottom-end',
+    'left-bottom': 'left-end',
+    'left-top': 'left-start',
+    'right-bottom': 'right-end',
+    'right-top': 'right-start',
+  };
+  return mapping[deprecatedValue];
+};
```
adamalston, line 139: As with the other comment, suggestion to simplify here by exporting `propMappingFunction` from `packages/react/src/components/Popover/index.tsx` and using it in this file instead of create another function.

Again, all of this logic will be consolidated pending https://github.com/carbon-design-system/carbon/pull/19081.

tay1orjones (changes requested):
A couple comments but otherwise looks good to me. Thanks for taking this on!

---

Having `align` alongside `direction` and `flipped` is probably going to lead to some confusion. Prop/type comments help though since they explicitly mention "tooltip". The new OverflowMenu implementation calls it `tooltipAlignment` to avoid this confusion.

This being the old implementation though I don't think it's necessary to try and get a new `tooltipAlignment` to work alongside the existing `align` that some might be using. Plus it would add another layer of complexity having to iron out the types for all that. 

I just want to make note of this, there's no action for you to take here.

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
jose-biescas: Would you want me to wait until #19081 gets merged, so I can export the function from `mapPopoverAlign.ts`? Instead of exporting `propMappingFunction`, since it will be changing.

adamalston: I don't think you have to wait necessarily. I guess another thing you could do instead of making these changes is to add a `TODO` like the following in the code:

```tsx
// TODO: Replace this code when align mapping is consolidated.
// https://github.com/carbon-design-system/carbon/pull/19081
```

That `TODO` will just let me know to go and update this code later.

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
jose-biescas: The `type` has been removed

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
jose-biescas: Using `PopoverAlignment` as the type for `align` prop

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
jose-biescas: Added the TODO

packages/react/src/components/OverflowMenu/OverflowMenu.stories.js
jose-biescas: Updated the list to the latest options

tay1orjones (approved):

jose-biescas:
@tay1orjones Apologies for that, not sure what happened. Will make sure future commits are also verified/signed. Thank you!

kennylam (approved):

2nikhiltom (approved):

PR #19114, opened 2025-04-12
refactor: convert `ToggleSmallSkeleton` to functional component

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/19112

Converted `ToggleSmallSkeleton` to a functional component.

#### Changelog

**Changed**

- Converted `ToggleSmallSkeleton` to a functional component.
- Updated the deprecation message `ModalWrapper.tsx` because the incorrect version was specified.

#### Testing / Reviewing

```sh
yarn test packages/react
```

PR #19132, opened 2025-04-14
fix(ModalWrapper): correct deprecation version from v12.x to v2.x

heloiselui:
Closes #19114
- Fixes incorrect deprecation version reference in ModalWrapper from this PR (19114)

#### Changelog
**Changed**
- Updated the deprecation message to correctly refer to removal in `@carbon/react@v2.x` instead of `v12.x`.

Testing / Reviewing
`yarn test packages/react`

### DIFF
--- a/packages/react/src/components/ModalWrapper/ModalWrapper.tsx
+++ b/packages/react/src/components/ModalWrapper/ModalWrapper.tsx
@@ -54,7 +54,7 @@ export default class ModalWrapper extends React.Component<
   if(isDev) {
     warning(
       didWarnAboutDeprecation,
-      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in the next major version, `@carbon/react@v12.x`'
+      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in the next major version, `@carbon/react@v2.x`'
     );
     didWarnAboutDeprecation = true;
   }


packages/react/src/components/ModalWrapper/ModalWrapper.tsx
```diff
@@ -54,7 +54,7 @@ export default class ModalWrapper extends React.Component<
   if(isDev) {
     warning(
       didWarnAboutDeprecation,
-      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in the next major version, `@carbon/react@v12.x`'
+      '`<ModalWrapper>` has been deprecated in favor of `<ComposedModal/>` and will be removed in the next major version, `@carbon/react@v2.x`'
```
adamalston, line 57: Hmm, I don't understand how the update I made in https://github.com/carbon-design-system/carbon/pull/19114 was incorrect. If you look at https://github.com/carbon-design-system/carbon/pull/12263, where this component was deprecated, migations.mdx includes this code:

https://github.com/carbon-design-system/carbon/blob/813a001e5dc109d419ea00ec807455293b341588/packages/react/src/components/ModalWrapper/migration.mdx?plain=1#L5-L6

So, either `migration.mdx` had an incorrect version or `ModalWrapper.tsx` had an incorrect version. Given there's no v2 of `@carbon/react` coming up, I concluded that `ModalWrapper.tsx` had the incorrect version.

I even made a mention of this change in https://github.com/carbon-design-system/carbon/pull/19114#issue-2990483128 when I opened the pull request.

tay1orjones: When talking about the next major, we use "v12" as a sort of "thematic name" for the release. Lots of packages will actually move from v11.x to v12.x, but others will have different numbers because we use an independent versioning strategy.

The next major ("v12") for `@carbon/react` will be `v2.x`. For `@carbon/web-components` it'll be `v3.x`, etc. Because of this I think it makes sense that when messages use specific package names they should specify the actual version number.

tay1orjones (approved):

packages/react/src/components/ModalWrapper/ModalWrapper.tsx
adamalston: Okay. I guess another confusing part here is that when I search for <kbd>v2</kbd> in packages/react/src, there is exactly 1 search result and it's in `ModalWrapper.tsx`. When I search for <kbd>v12</kbd>, there are more than 150 search results. So it seems like we're calling the next major release of `@carbon/react` 2 different things.

kennylam (approved):

PR #18553, opened 2025-02-10
feat(modal): refactor to use native dialog element

tay1orjones:
Closes #17086

<details><summary>Original implementation notes</summary>
<p>

#### Todo
- [x] Refactor Modal to use `Dialog` behind a flag, and do not use the new Dialog* components to prevent changes in classnames for consumers.
- [x] Refactor ComposedModal to use `Dialog` behind a flag, and do not use the new Dialog* components to prevent changes in classnames for consumers.

#### Concerns beyond the scope of this PR that were moved to https://github.com/carbon-design-system/carbon/issues/18243
- [ ] Determine if we should deprecate Modal and ComposedModal in favor of using Dialog directly 
- [ ] Assess a strategy of migration for classnames. e.g. If someone updates to dialog-based modals and has existing modal override styles, what could/should they do? Can we help?
  - [ ] Maybe Modal and ComposedModal should use their own existing styles and not use any new dialog-based components or classes, even when using dialog under the hood. Just put everything that already exists into the empty dialog box. The migration path for consumers would be to directly use Dialog moving forward, and update existing usages of Modal to use `<Dialog modal>`. We'll keep Modal and ComposedModal around until usage drops super low and then we could remove them in a future major.

</p>
</details> 

Huge thanks to @heloiselui for pairing up with me on this one 🎉 

#### Changelog

**New**

- Add `enableDialogElement` feature flag

**Changed**

- Refactor Modal to use the native `<dialog>` behind the `enableDialogElement` flag
- Refactor ComposedModal to use the native `<dialog>` behind the `enableDialogElement` flag
- Update documentation for both components
- Update storybook docs for the feature flags
- Clarify `enableExperimentalFocusWrapWithoutSentinels` has effectively been renamed/superseded by `enableDialogElement`
- Modified styles to conditionally emit only when the flag is not in use
  - Updated the modal story styles to bring this in under a custom selector and prefix to avoid style collisions

**Removed**

- `Dialog-story.js` is now `Dialog.stories.js`

#### Testing / Reviewing

- Review the unstable__Dialog stories, they're not fully baked but you can see how the base component works for modal and non modal 
- Review the Modal feature flag story and docs
- Review the ComposedModal feature flag story and docs
### ⚠️ BEFORE MERGE ⚠️ 
* we should uncomment the `includeStories: []` in Dialog.stories.js to ensure they're not shown in the storybook. We'll eventually show them but it would be something for #18243 or it's parent #13807

Issue #18243, opened 2024-12-11
Export `<Dialog>` as experimental


sstrubberg:
- [ ] Evaluate how dialog fits into the system - It's a primitive, minimal styling. How do people know when to use it. When would i use dialog over popover. Why would i use modal dialog over the actual modal component?
  - [x] Based on work leftover from #17086, #18553, we should build out a family of Dialog components, fully composable, that do include full styling to follow the design spec for both modal and non-modal dialogs. Eventually this could replace Modal & ComposedModal.
  - [x] Add more composable inner components, Dialog.Title, Dialog.Subtitle, etc. that follow [what the PageHeader stuff is doing](https://github.com/carbon-design-system/carbon/pull/18788/files#diff-30749304ea5a669921bae44b34f076145d09b4260de7cf75cfd5ffdd6ed231eaR40-R44)
  - [x] Have Dialog.Header, Dialog.Content/Dialog.Body, Dialog.Footer take some inspiration from the components from ComposedModal
  - [ ] Could `<Dialog />` be used in place of Modal and ComposedModal? Maybe those become just a re-export of `<Dialog/>`?
- [x] Implement the [scroll shadow](https://css-tricks.com/books/greatest-css-tricks/scroll-shadows/) (already present in Modal but needs pulled over into Dialog to follow the spec
- [x] Add VRT for Dialog
- [x] Add AVT for Dialog
- [x] Add unit tests for Dialog
- [ ] Finalize storybook docs for Dialog

2nikhiltom:
https://github.com/carbon-design-system/carbon/issues/19496 -> issue for Modal v Composed Modal Audit and to figure out how dialog fits into system
https://github.com/carbon-design-system/carbon/issues/19484 -> Test cases are ready - will be merged with Dialog is made available in storybook

PR #19134, opened 2025-04-14
test(dialog): initial unit tests, mock `HTMLDialogElement` in jest config, revert feature flag unification

tay1orjones:
In https://github.com/carbon-design-system/carbon/pull/18553 when we merged `enable-experimental-focus-wrap-without-sentinels` into `enable-dialog-element` we should've included an update to the jest config to mock `HTMLDialogElement` because jsdom doesn't implement it, https://github.com/jsdom/jsdom/issues/3294

Additionally, unifying these two flags has caused some unanticipated problems, so I've separated these again with updated docs and console warnings if both are used at the same time.

#### Changelog

**New**

- add initial unit tests for internal/unstable Dialog component to validate HTMLDialogElement mock works as expected
- console warnings if both are used at the same time

**Changed**

- update jest config to mock HTMLDialogElement
- update modal composedmodal to support `enable-experimental-focus-wrap-without-sentinels` separately from `enable-dialog-element` (this reinstates the old behavior of `enable-experimental-focus-wrap-without-sentinels`)
- update feature flags docs for Modal and ComposedModal

#### Testing / Reviewing

- Double check my syntax in the tests and docs
- CI tests should pass

> [!NOTE]  
> @2nikhiltom There's two tests included here that are skipped because they can't be tested with the simple mocks I added. For #18243 we should test the portions of the API that rely on actual HTMLDialogElement functionality in a real browser in e2e tests.


2nikhiltom (approved):
Awesome! thank you

packages/react/src/components/Dialog/Dialog-test.js
```diff
@@ -0,0 +1,107 @@
+/**
+ * Copyright IBM Corp. 2023
```
heloiselui, line 2:
```suggestion
 * Copyright IBM Corp. 2025
```

heloiselui (approved):
Looks great!

tay1orjones:
I'm going to make one more change here based on some user feedback we got today

tay1orjones:
@2nikhiltom @heloiselui @kennylam I've updated this PR with some additional pieces [based on user feedback](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1744743640390829?thread_ts=1744630233.564899&cid=C2K6RFJ1G). 

@wkeese @AlanGreene Once this is merged I'll cut a patch containing the changes from this PR

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -486,12 +510,14 @@ const ComposedModal = React.forwardRef<HTMLDivElement, ComposedModalProps>(
         aria-label={ariaLabel ? ariaLabel : generatedAriaLabel}
         aria-labelledby={ariaLabelledBy}>
         {/* Non-translatable: Focus-wrap code makes this `<button>` not actually read by screen readers */}
-        <button
-          type="button"
-          ref={startSentinel}
-          className={`${prefix}--visually-hidden`}>
-          Focus sentinel
-        </button>
+        {!focusTrapWithoutSentinels && (
```
wkeese, line 513: Shouldn't this condition be `!enableDialogElement && !focusTrapWithoutSentinels` like you did for Modal.tsx?  Likewise on line 534.

wkeese (commented):
Thanks for the quick response on this!

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -312,6 +322,19 @@ const Modal = React.forwardRef(function Modal(
...
+      if (
+        focusTrapWithoutSentinels &&
+        !enableDialogElement &&
+        match(evt, keys.Tab) &&
+        innerModal.current
+      ) {
+        wrapFocusWithoutSentinels({
+          containerNode: innerModal.current,
+          currentActiveNode: evt.target,
+          event: evt as any,
```
adamalston, line 337:
```suggestion
          // TODO: Delete type assertion following util rewrite.
          // https://github.com/carbon-design-system/carbon/pull/18913
          event: evt as any,
```

packages/react/src/components/ComposedModal/ComposedModal.tsx
2nikhiltom: Hi I feel there is no need to check enableDialogElement here . We're in the else branch of the enableDialogElement ternary,
  so only render sentinel buttons when not using the focus-wrap-without-sentinels approach.
```
  const modalBody = enableDialogElement ? (
  <Dialog>
    {/* Dialog content - no sentinel buttons needed */}
  </Dialog>
) : (
  <div>
    {!focusTrapWithoutSentinels && (
      <button>Focus sentinel</button>
    )}
```

2nikhiltom (approved):

packages/react/src/components/ComposedModal/ComposedModal.tsx
wkeese: Oh OK, makes sense.

heloiselui (approved):

kennylam (approved):

wkeese (approved):

PR #19141, opened 2025-04-15
refactor: rewrite `IconSwitch` in typescript and add types

adamalston:
No issue.

Rewrote `IconSwitch` in TypeScript.

#### Changelog

**New**

- Added types to `IconSwitch`.

**Changed**

- Rewrote `IconSwitch` in TypeScript.
- Replace the references to `Switch` in the component docs with IconSwitch`.

#### Testing / Reviewing

I left some `TODO`s with questions that hopefully someone can answer.

```sh
yarn test packages/react
```


packages/react/src/components/IconButton/IconButton.stories.js
```diff
@@ -57,6 +57,10 @@ Default.args = {
 
 Default.argTypes = {
   align: {
+    // TODO:
+    // 1. Should the deprecated options be deleted?
+    // 2. The list doesn't include all of the options available in the
+    //    component. Is it supposed to?
```
kennylam, line 63: Yes, the deprecated options should be deleted and use the new options instead. I also checked with @thyhmdo on caret alignment and confirmed that [toolip placements](https://carbondesignsystem.com/components/tooltip/usage/#placement) should not include `-start`/`-end` for `left`/`right` and they should default to the center alignment.

Those alignments should be used for [popover](https://carbondesignsystem.com/components/popover/usage/#caret-tip).

kennylam (approved):

Gururajj77 (approved):

PR #19146, opened 2025-04-16
fix(Modal): only focus launcherButtonRef on close

### LINKED ISSUE #18709: [Bug]: Modal's "launcherButtonRef" usage grabs focus on page load.
### Package

@carbon/react

### Browser

Chrome

### Package version

1.76.0

### React version

_No response_

### Description

**What I saw**

On page load, a button using the `launcherButtonRef` prop is grabbing focus.

**What I expected to see**

On page load, a button using the `launcherButtonRef` prop _should not_ be grabbing focus.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/docs/components-modal--overview

### Steps to reproduce

1. Load the above URL.
2. See how the "Launch modal" button has focus.

This behavior can also been seen with **Carbon for IBM Products'** `ExportModal` [Storybook page](http://localhost:3000/?path=/docs/ibm-products-components-export-exportmodal--docs&globals=viewport:basic). This component uses the Carbon Modal as its base component.

Scroll to to the top of the page and reload the page
1. The page auto-scrolls to the button with focus.
2. Reload multiple times and see how the second or third button (both using `launcherButtonRef`) randomly grabs focus.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

wkeese:
Closes #18709.

Fix bug where `launcherButtonRef` is focused when `Modal` renders, even though it has never even been opened.

Also, add usePreviousValue() hook based on the one in @carbon/ibm-products. 

Side note: I noticed that the `launcherButtonRef` code is disabled when the `enableDialogElement` flag is set to true.  That seems overly optimistic, especially since `launcherButtonRef` currently supports focusing any element, not just the actual launch button. Regardless, that's unrelated to this PR.

#### Changelog

**Changed**

- Fixed Modal logic to only focus launcherButtonRef when the Modal is closed, not when it's initially rendered.

#### Testing / Reviewing

I tested in Storybook after modifying Modal.stories.tsx to add an `input` focus point and a `launcherButtonRef`:

```jsx
export const Default = ({ numberOfButtons, ...args }) => {
  const [open, setOpen] = useState(false);
  const [popoverOpen, setPopoverOpen] = useState(false);
  const launcherButtonRef = useRef(null);
  return (
    <>
      <input/>
      <Button ref={launcherButtonRef} onClick={() => setOpen(true)}>Launch modal</Button>
      <Modal
        onRequestClose={(e) => {
          action(e);
          setOpen(false);
        }}
        modalHeading="Add a custom domain"
        primaryButtonText="Add"
        secondaryButtonText="Cancel"
        aria-label="Modal content"
        open={open}
        launcherButtonRef={launcherButtonRef}
        {...args}
```


packages/react/src/internal/usePreviousValue.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright IBM Corp. 2021, 2021
```
adamalston:
```suggestion
 * Copyright IBM Corp. 2025
```

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -73,7 +73,7 @@ export interface ModalProps extends ReactAttr<HTMLDivElement> {
   className?: string;
 
   /**
-   * Specify an label for the close button of the modal; defaults to close
+   * Specify label for the close button of the modal; defaults to close
```
adamalston, line 78: Does the following code need to be updated too?

https://github.com/carbon-design-system/carbon/blob/0338c6c99ee54f521fc0038c67ec27e890975521/packages/react/src/components/Modal/Modal.tsx#L782

packages/react/src/internal/usePreviousValue.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param {object} value - The current value of any type.
+ * @returns {T | undefined} - The previous value of the same type, or undefined if there is none
+ */
+export const usePreviousValue = (value) => {
+  const ref = useRef(undefined);
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
```
adamalston: The JSDoc comment doesn't have the correct types (the comment doesn't even need the types as long as the function includes them) and the function is missing generics.

```suggestion
/**
 * Stores the previous value of a given input.
 *
 * @param value - The current value.
 * @returns The value before the current render.
 *
 * @example
 * const [count, setCount] = useState(0);
 * const prevCount = usePreviousValue(count);
 */
export const usePreviousValue = <T,>(value: T): T | undefined => {
  const ref = useRef<T | undefined>(undefined);

  useEffect(() => {
    ref.current = value;
  }, [value]);

  return ref.current;
};
```

adamalston (commented):
Should a functional test be added in packages/react/src/components/Modal/Modal-test.js? If so and you don't want to write one, feel free to use the one that I wrote when I was playing around with a fix for this issue earlier: https://github.com/carbon-design-system/carbon/commit/d44e13f4a72f663abc9cb7ab069d1ae9cd80773b

packages/react/src/components/Modal/Modal.tsx
wkeese: thanks, updated

packages/react/src/internal/usePreviousValue.ts
wkeese: Make sense.  (I just copied the code from @carbon/ibm-products, but this is an improvement.)

wkeese:
> Should a functional test be added in packages/react/src/components/Modal/Modal-test.js? If so and you don't want to write one, feel free to use the one that I wrote when I was playing around with a fix for this issue earlier: [d44e13f](https://github.com/carbon-design-system/carbon/commit/d44e13f4a72f663abc9cb7ab069d1ae9cd80773b)

Oh thanks, I didn't realize you were working on a fix.  I copied over your code, although I'm not sure what should be in the Jest "unit" tests vs. what should be in the E2E playwright tests.

packages/react/src/internal/usePreviousValue.ts
adamalston: Maybe their code needs to be updated too.

packages/react/src/internal/usePreviousValue.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export const usePreviousValue = <T>(value: T): T | undefined => {
+  const ref = useRef<T | undefined>(undefined);
+
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+
+  return ref.current;
+};
```
adamalston, line 28: If this util is going to be added, do tests need to be written for it?

wkeese: Hmm, you're making me regret copying over this function as opposed to using a library like https://usehooks.com/useprevious.  But I added a unit test.

packages/react/src/internal/__tests__/usePreviousValue.test.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * Copyright IBM Corp. 2020
```
adamalston:
```suggestion
 * Copyright IBM Corp. 2025
```

packages/react/src/internal/usePreviousValue.ts
adamalston: Apologies. If https://github.com/uidotdev/usehooks/blob/90fbbb4cc085e74e50c36a62a5759a40c62bb98e/index.js#L1017-L1027 is the code for that hook, I would be hesitant to use it. It doesn't follow the rules of hooks. But yea, sorry about that.

wkeese: I guess it technically breaks the rules of hooks, but 

```js
  if (value !== current) {
    setPrevious(current);
    setCurrent(value);
  }
```

is essentially equivalent to

```js
useEffect(() => {
  if (value !== current) {
    setPrevious(current);
    setCurrent(value);
  }
}, [current, value]);
```

The more interesting thing to me, assuming I'm reading this correctly, is that the behavior is different.  

`open` and `usePreviousValue(open)` are only different for a single render, after which `usePreviousValue(open)` catches up to the current value.

Whereas `open` and `usePrevious(open)` are always different, no matter how many times the component renders.

2nikhiltom (commented):
LGTM

2nikhiltom (approved):

tay1orjones (approved):

tay1orjones:
> I noticed that the `launcherButtonRef` code is disabled when the `enableDialogElement` flag is set to true. That seems overly optimistic, especially since `launcherButtonRef` currently supports focusing any element, not just the actual launch button. Regardless, that's unrelated to this PR.

The rationale right now is we're aiming to defer all focusing behavior to the specc'ed/implemented native `<dialog>` behavior by the browser, which puts focus on the previously focused element on close. If they open up a way to configure it we'll aim to support it.

PR #19152, opened 2025-04-16
fix(`Toggletip`): prevent escape key from propagating when open

### LINKED ISSUE #19065: [Bug]: Pressing Esc key while Toggletip content is open will also close the Modal it's in
### Package

@carbon/react

### Browser

Chrome

### Package version

1.79.0

### React version

16.14.0

### Description

When rendering a `Toggletip` inside a `Modal`, pressing the `Esc` key in attempt to close the `Toggletip` will also close the `Modal` it's sitting inside. The expected behavior would be that the `Toggletip` content would close, but the `Modal` still stays open when an user presses the `Esc` key.

### Reproduction/example

https://stackblitz.com/edit/github-qcbokazr?file=src%2FApp.jsx

### Steps to reproduce

1. Render a `Toggletip` inside a `Modal` with content
2. Click on `Toggletip` to open content
3. Press `Esc` key on keyboard

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/19065

Prevented Escape key from propagating when `ToggleTip` was open.

#### Changelog

**Changed**

- Prevented Escape key from propagating when `ToggleTip` was open.

#### Testing / Reviewing

I left some `TODO`s with questions around code organization. Hopefully someone can answer them.

```sh
yarn test packages/react
```


packages/react/src/components/Toggletip/__tests__/Toggletip-test.js
```diff
@@ -242,6 +244,12 @@ describe('Toggletip', () => {
       });
     });
 
+    // TODO: There is ToggletipLabel-test.js. Should this describe block be
```
kennylam: Moving the one block from `ToggletipLabel` to this file makes sense, IMO. The same goes for the other files (`actions`, `button`, etc), maybe.

packages/react/src/components/Toggletip/Toggletip.stories.js
```diff
@@ -137,6 +138,27 @@ Default.argTypes = {
   },
 };
 
+export const Test = () => {
```
kennylam: Can remove this, I think.

kennylam (approved):

packages/react/src/components/Toggletip/__tests__/Toggletip-test.js
adamalston: I made some changes. Let me know if I misunderstood or if something else should be changed.

Edit: Just to confirm, you meant move the tests from the other files to this file, correct, and not the other way around?

Gururajj77 (approved):

packages/react/src/components/Toggletip/__tests__/Toggletip-test.js
```diff
@@ -54,6 +55,8 @@ describe('Toggletip', () => {
     expect(container.firstChild).not.toHaveClass(`${prefix}--toggletip--open`);
     expect(container.firstChild).not.toHaveClass(`${prefix}--popover--open`);
   });
+  // TODO: Should these close tests be moved inside of `Toggletip Closing
+  // Behavior`?
```
tay1orjones: They could, not a huge deal to leave them here by the other "visibility" tests though imo

PR #19157, opened 2025-04-17
feat(formatDate): add locale option to be extensible for absolute format

dkaushik95:
Closes #

Currently `absolute.format*` doesn't work for our product since we allow users the option to switch to UTC time. Adding `timeZone` as a prop allows us to override the default timezone that `Intl` chooses. 

#### Changelog

**New**

- @carbon/utilities - `Datetime.absolute` now allows you to choose a `timeZone` which can be used to override the default timeZone while formatting datetime. This prop will be available to: `formatTime`, `formatTime`, `format`, and `formatRange`

#### Testing / Reviewing

- Testers can try out the following use case:
1. timestamp: `1744835151846`
2. use this timestamp to check the time difference between your locale and `UTC`
3. eg:
```
datetime: 1744835151846
utc string: Apr 16, 2025, 8:25:51 PM
local string: Apr 16, 2025, 1:25:51 PM
```


packages/utilities/src/dateTimeFormat/absolute.ts
```diff
@@ -24,6 +26,7 @@ export function formatDate(
   options?: Partial<{
     locale: string;
     style: Intl.DateTimeFormatOptions['dateStyle'];
+    timeZone?: Intl.DateTimeFormatOptions['timeZone'];
   }>
```
nchevsky: `options` is already wrapped in `Partial<>`, so the `?` after `timeZone` is redundant.

packages/utilities/src/dateTimeFormat/absolute.ts
```diff
@@ -40,6 +43,7 @@ export function format(
     style: Intl.DateTimeFormatOptions['timeStyle'] | 'tooltip';
     timeStyle: Intl.DateTimeFormatOptions['timeStyle'];
     dateStyle: Intl.DateTimeFormatOptions['dateStyle'];
+    timeZone?: Intl.DateTimeFormatOptions['timeZone'];
   }>
```
nchevsky: `options` is already wrapped in `Partial<>`, so the `?` after `timeZone` is redundant.

packages/utilities/src/dateTimeFormat/absolute.ts
```diff
@@ -68,6 +73,7 @@ export function formatRange(
     style: Intl.DateTimeFormatOptions['timeStyle'];
     timeStyle: Intl.DateTimeFormatOptions['timeStyle'] | null;
     dateStyle: Intl.DateTimeFormatOptions['dateStyle'] | null;
+    timeZone?: Intl.DateTimeFormatOptions['timeZone'];
   }>
```
nchevsky: `options` is already wrapped in `Partial<>`, so the `?` after `timeZone` is redundant.

nchevsky (changes requested):
@dkaushik95 Can you add unit test coverage in `absolute-test.js`?

dkaushik95:
Thanks for the review @nchevsky . I have addressed all the review comments. Can you re-review it? Thanks!

packages/utilities/src/dateTimeFormat/absolute-test.js
adamalston, line 1: Can you update the copyright notice in this file?

```diff
- * Copyright IBM Corp. 2024
+ * Copyright IBM Corp. 2024, 2025
```

packages/utilities/src/dateTimeFormat/absolute.ts
adamalston, line 1: Can you update the copyright notice in this file?

packages/utilities/src/dateTimeFormat/absolute-test.js
```diff
@@ -23,6 +24,23 @@ describe(locale, () => {
...
+
+        test(`${style} with timeZone ${timeZone} → ${expectedOutput}`, () => {
+          const actualOutput = absolute.formatTime(date, {
+            locale,
+            style,
+            timeZone,
+          });
+          expect(actualOutput).toBe(expectedOutput);
+        });
+      });
```
nchevsky: I wouldn't put this within the `timeStyle` loop, as (a) it multiplies the number of iterations 4 → 12 and (b) `timeStyle` and `timeZone` aren't dependent on each other.

We don't even need to test three different time zones, for `timeZone` is simply passed through to `Intl.DateTimeFormat` regardless of its value. Some alternatives, in ascending order of complexity:

1. Test any two time zones so that, even if one happens to match the host's, the other won't; or
2. spy on `Intl.DateTimeFormat` and simply assert that `absolute.format*()` instantiated it with the given `timeZone` option; or
3. dynamically select a time zone other than the host's, and test just that one.

I'd personally go with # 2 as it's more of a true unit test, but seeing how none of the preexisting tests are written this way, it might be less controversial to stick with # 1.

Whichever you choose, please replicate to all tests. 🙏🏻

packages/utilities/src/dateTimeFormat/absolute-test.js
```diff
@@ -54,6 +89,26 @@ describe(locale, () => {
...
+            const actualOutput = absolute.format(date, {
+              locale,
+              style: timeStyle,
+              timeZone,
+            });
+            expect(actualOutput).toBe(expectedOutput);
+          });
+        });
+      });
+
```
nchevsky: A couple of consistency nitpicks:

1. In `formatTime()` and `formatDate()`, the time zone `test()`s are top-level.
2. In `format()` and `formatRange()`, the time zone `test()`s are wrapped in `describe()` blocks.

packages/utilities/src/dateTimeFormat/absolute-test.js
dkaushik95: Hey, I chose the 1st option. I chose two timezones `Tokyo` and `UTC` so that it's always different for any host for at least one test. I also addressed the 2nd concern of using `describe` for timeZones so it should be consistent across all tests. I was trying different approaches and accidentally added both approaches :) 

This will how the tests would show up. 

At the end there will be 2 test cases for each function (8 iterations in all)

Screenshot 2025-04-21 at 8 58 21 AM

nchevsky (approved):

preetibansalui (approved):

preetibansalui:
@dkaushik95 I approved the PR and then realized we missed updating the documentation for the newly added option. Could you please add the relevant details to the [README](https://github.com/carbon-design-system/carbon/tree/main/packages/utilities/src/dateTimeFormat) file?

dkaushik95:
@preetibansalui Hello! I have made changes to add documentation to the readme. Can you give it a review? Thanks!

packages/utilities/src/dateTimeFormat/README.md
```diff
@@ -142,3 +142,22 @@ dateTimeFormat.absolute.formatRange(startDate, sameDayEndDate, {
 });
 // 3:47 – 4:29 PM
 ```
+
+#### Timezone
+
+For `absolute` functions, you can provide `timeZone` as an optional property.
+This is usefull when (for example) you want to display utc time instead of a
+local timezone.
```
preetibansalui, line 150: This is **useful** when (for example) you want to display **UTC** time instead of a
local time zone.

preetibansalui (changes requested):
Looks good to me—just a small spelling correction needed.

packages/utilities/src/dateTimeFormat/README.md
dkaushik95: thanks! Fixed :)

preetibansalui (approved):
Thanks for the changes, This LGTM :+1:

ariellalgilmore (approved):

PR #19158, opened 2025-04-17
fix(OverflowMenu): fix types

wkeese:
Fix `OverflowMenuProps` to include properties that get passed through to `IconButton`, for example `align`, `autoAlign`, and `tooltip`.

The alternative is to list the properties manually, which is what #19131 did for the `align` prop, but not for the other props.

Relatedly, fix `TableToolbarMenuProps` to extend `OverflowMenuProps`. I also removed the `aria-label` as it's redundant with `iconDescription`.

#### Changelog

**Changed**

- Fixed Typescript types of `OverflowMenu` and `TableToolbar`.

#### Testing / Reviewing

Built carbon, and also built my app with the new carbon.


heloiselui (approved):
Looks good!

kennylam (approved):

PR #19163, opened 2025-04-17
feat(Dialog): Implement composable pattern with Modal like functionalities

### LINKED ISSUE #19057: <Dialog> to have  composable inner components
parent issue: https://github.com/carbon-design-system/carbon/issues/18243
Add more composable inner components, Dialog.Title, Dialog.Subtitle, etc. that follow [what the PageHeader stuff is doing](https://github.com/carbon-design-system/carbon/pull/18788/files#diff-30749304ea5a669921bae44b34f076145d09b4260de7cf75cfd5ffdd6ed231eaR40-R44)

Have Dialog.Header, Dialog.Content/Dialog.Body, Dialog.Footer take some inspiration from the components from ComposedModal

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/19057  

closes related sub-tasks from :  https://github.com/carbon-design-system/carbon/issues/18243 

This PR transforms the Dialog component by implementing a composable pattern architecture and adds functionalities  similar to the Modal component

#### Changelog

**New**

- Added composable component API with Dialog.Title, Dialog.Subtitle, Dialog.Body, Dialog.Footer
- Added Playwright test cases
- Implemented scrollable content support with automatic detection
- Added accessibility improvements (aria attributes, proper focus management)
- Added support for danger variant with automatic focus on cancel button
- Added styles for Dialog component

**Changed**

- Refactored to use named exports for all subcomponents
- Refactored TypeScript definitions for all component props

#### Testing / Reviewing
CI should pass
[deploy preview](https://deploy-preview-19163--v11-carbon-react.netlify.app/?path=/docs/experimental-unstable-dialog--overview)
Check both modal and non-modal dialog variants
Verify scrollable content works correctly
verify Test keyboard navigation and focus  

⚠️ BEFORE MERGE ⚠️
we should uncomment the includeStories: [] in Dialog.stories.js to ensure they're not shown in the storybook. We'll eventually show them but it would be something for  https://github.com/carbon-design-system/carbon/issues/13807


e2e/components/Dialog/Dialog-test.avt.e2e.js
```diff
@@ -0,0 +1,128 @@
...
+    await expect(page.locator('.cds--dialog__close')).toBeFocused();
+    await expect(page.locator('.cds--dialog__header')).toBeVisible();
+    await expect(page.locator('.cds--dialog-header__heading')).toBeVisible();
+    await expect(page.locator('.cds--dialog-header__label')).toBeVisible();
+    await expect(page.locator('.cds--dialog-content')).toBeVisible();
+    await expect(page.locator('.cds--dialog-footer')).toBeVisible();
+    await page.locator('.cds--dialog__close').click();
+    await expect(dialog).toBeHidden();
+  });
+  test('@avt-native-dialog-behavior non-modal', async ({ page }) => {
```
tay1orjones:
```suggestion
  test('@avt-advanced-states dialog non-modal behavior', async ({ page }) => {
```

e2e/components/Dialog/Dialog-test.avt.e2e.js
```diff
@@ -0,0 +1,128 @@
...
+    await expect(dialog).toHaveClass(/cds--dialog cds--dialog--modal/);
+    await expect(page.locator('.cds--dialog__close')).toBeFocused();
+    await page.keyboard.press('Tab');
+    // In dialog the focus circules through the close button and URL
+    await page.keyboard.press('Tab');
+    await expect(page.locator('.cds--dialog__close')).toBeFocused();
+    await page.locator('.cds--dialog__close').click();
+    await expect(dialog).toBeHidden();
+  });
+  test('@avt-alertdialog-role danger dialog', async ({ page }) => {
```
tay1orjones:
```suggestion
  test('@avt-advanced-states dialog danger alertdialog role', async ({ page }) => {
```

e2e/components/Dialog/Dialog-test.avt.e2e.js
```diff
@@ -0,0 +1,128 @@
...
+    await visitStory(page, {
+      component: 'unstable_Dialog',
+      id: 'experimental-unstable-dialog--modal',
+      globals: {
+        theme: 'white',
+      },
+    });
+    await expect(page).toHaveNoACViolations('Dialog');
+  });
+  test('@avt-native-dialog-behavior modal', async ({ page }) => {
```
tay1orjones:
```suggestion
  test('@avt-advanced-states dialog modal behavior', async ({ page }) => {
```

e2e/components/Dialog/Dialog-test.avt.e2e.js
```diff
@@ -0,0 +1,128 @@
...
+      },
+    });
+    const dialog = page.locator('dialog');
+    await expect(dialog).toBeVisible();
+    await expect(dialog).toHaveClass(/cds--dialog cds--dialog--modal/);
+    await expect(page.locator('.cds--dialog__close')).toBeFocused();
+    await expect(dialog).toBeVisible();
+    await page.keyboard.press('Tab');
+    await expect(
+      page.locator('.cds--btn.cds--btn--secondary:has-text("Close")')
```
tay1orjones: These selectors could be `getByRole('button', { name: "Close" })` instead of using the classnames

e2e/components/Dialog/Dialog-test.avt.e2e.js
```diff
@@ -0,0 +1,128 @@
...
+    await visitStory(page, {
+      component: 'unstable_Dialog',
+      id: 'experimental-unstable-dialog--passive-dialog',
+      globals: {
+        theme: 'white',
+      },
+    });
+    const toggleButton = page.getByRole('button', { name: 'Toggle open' });
+    await toggleButton.click();
+    const dialog = page.locator('dialog');
```
tay1orjones: Similar here
```suggestion
    const dialog = page.getByRole('dialog');
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -66,38 +69,53 @@ export const Modal = ({ open: _open, ...args }) => {
...
+            laborum, ratione.
+          </p>
+          <br></br>
+        </Dialog.Body>
+        <Dialog.Footer>
+          <Button
+            type="button"
+            kind="secondary"
+            onClick={closeDialog}
+            autoFocus>
```
tay1orjones: I think for this example it might be more appropriate to place `autofocus` on the DialogCloseButton. IIRC it's just the "danger" variant with the requirement to focus the footer Cancel button by default.

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -66,38 +69,53 @@ export const Modal = ({ open: _open, ...args }) => {
...
-          {/* <DialogSubtitle></DialogSubtitle>
-          <DialogTitle></DialogTitle> */}
+      <Dialog
+        {...args}
+        open={open}
+        onRequestClose={handleRequestClose}
+        aria-label="Dialog Title">
+        <Dialog.Header>
+          <Dialog.Subtitle>Configure dialog settings</Dialog.Subtitle>
+          <Dialog.Title>Modal Dialog Example</Dialog.Title>
```
tay1orjones: Would this be better since we already have the Title as a "label"?

```suggestion
        aria-labelledby="title">
        <Dialog.Header>
          <Dialog.Subtitle>Configure dialog settings</Dialog.Subtitle>
          <Dialog.Title id="title">Modal Dialog Example</Dialog.Title>
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -132,37 +150,59 @@ export const NonModal = ({ open: _open, ...args }) => {
...
         Toggle open
       </Button>
-      <Dialog {...args} open={open} onRequestClose={handleRequestClose}>
-        <DialogHeader>
-          <p>TODO: Modal example</p>
+      <Dialog
+        {...args}
+        open={open}
+        onRequestClose={handleRequestClose}
+        aria-label="Dialog Title">
```
tay1orjones, line 199: same thing here w/ aria-labelledby

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -66,38 +69,53 @@ export const Modal = ({ open: _open, ...args }) => {
...
+          </p>
+          <br></br>
+          <p>
+            Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
+            autem officiis dolores facilis nulla earum! Neque quia nemo sequi
+            assumenda ratione officia Voluptate beatae eligendi placeat nemo
+            laborum, ratione.
+          </p>
+          <br></br>
+        </Dialog.Body>
```
tay1orjones:
```suggestion
        <Dialog.Body>
          <VStack>
            <p>
              Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
              autem officiis dolores facilis nulla earum! Neque quia nemo sequi
              assumenda ratione officia Voluptate beatae eligendi placeat nemo
              laborum, ratione.
            </p>
            <p>
              Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
              autem officiis dolores facilis nulla earum! Neque quia nemo sequi
              assumenda ratione officia Voluptate beatae eligendi placeat nemo
              laborum, ratione.
            </p>
            <p>
              Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
              autem officiis dolores facilis nulla earum! Neque quia nemo sequi
              assumenda ratione officia Voluptate beatae eligendi placeat nemo
              laborum, ratione.
            </p>
          </VStack>
        </Dialog.Body>
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -132,37 +150,59 @@ export const NonModal = ({ open: _open, ...args }) => {
...
+            <SelectItem value="us-south" text="US South" />
+            <SelectItem value="us-east" text="US East" />
+          </Select>
+          <p>
+            Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
+            autem officiis dolores facilis nulla earum! Neque quia nemo sequi
+            assumenda ratione officia Voluptate beatae eligendi placeat nemo
+            laborum, ratione.
+          </p>
+        </Dialog.Body>
```
tay1orjones: Inline styles aren't CSP compliant so we should avoid them in our examples

```suggestion
        <Dialog.Body>
          <VStack>
            <p>
              Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
              autem officiis dolores facilis nulla earum! Neque quia nemo sequi
              assumenda ratione officia Voluptate beatae eligendi placeat nemo
              laborum, ratione.
            </p>
            <TextInput
              id="dialog-text-input"
              labelText="Name"
              placeholder="Enter your name"
            />
            <Select
              id="dialog-select"
              labelText="Region"
              defaultValue="us-south">
              <SelectItem value="us-south" text="US South" />
              <SelectItem value="us-east" text="US East" />
            </Select>
            <p>
              Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
              autem officiis dolores facilis nulla earum! Neque quia nemo sequi
              assumenda ratione officia Voluptate beatae eligendi placeat nemo
              laborum, ratione.
            </p>
          </VStack>
        </Dialog.Body>
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -171,3 +211,238 @@ NonModal.args = {
...
+            elementum vel vehicula vitae, porttitor ac lorem. Sed viverra elit
+            ac risus tincidunt fermentum. Ut sollicitudin nibh id risus ornare
+            ornare. Etiam gravida orci ut lectus dictum, quis ultricies felis
+            mollis. Mauris nec commodo est, nec faucibus nibh. Nunc commodo ante
+            quis pretium consectetur. Ut ac nisl vitae mi mattis vulputate a at
+            elit. Nullam porttitor ex eget mi feugiat mattis. Nunc non sodales
+            magna. Proin ornare tellus quis hendrerit egestas. Donec pharetra
+            leo nec molestie sollicitudin.
+          </p>
+          <br></br>
```
tay1orjones: same thing here w/ using `VStack` instead of line break tags

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -171,3 +211,238 @@ NonModal.args = {
...
+          <TextInput
+            id="dialog-text-input"
+            labelText="Name"
+            placeholder="Enter your name"
+            style={{ marginBottom: '1rem' }}
```
tay1orjones: etc., will need done elsewhere in these examples.
```suggestion
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -171,3 +211,238 @@ NonModal.args = {
...
+    action('Dialog onRequestClose');
+    closeDialog(e);
+  }
+
+  useEffect(() => {
+    setOpen(_open);
+  }, [_open]);
+
+  return (
+    <div>
```
tay1orjones: These could be fragments instead of wrapper divs
```suggestion
    <>
```

packages/react/src/components/Dialog/index.tsx
```diff
@@ -122,21 +174,90 @@ export const unstable__Dialog = React.forwardRef(
       }
     }, [modal, open]);
 
+    useEffect(() => {
+      if (modal && open) {
+        document.body.classList.add(`${prefix}--body--with-modal-open`);
```
tay1orjones: Why do we need to place classes on the body?

packages/react/src/components/Dialog/index.tsx
```diff
@@ -122,21 +174,90 @@ export const unstable__Dialog = React.forwardRef(
...
+    // For danger dialogs, focus the secondary/cancel button
+    useEffect(() => {
+      if (open && ref.current && danger) {
+        setTimeout(() => {
+          if (ref.current) {
+            const secondaryButton = ref.current.querySelector<HTMLElement>(
+              `.${prefix}--btn.${prefix}--btn--secondary`
+            );
+            if (secondaryButton) {
+              secondaryButton.focus();
```
tay1orjones: This can all be removed in favor of placing `autofocus` on the desired element in the stories

packages/react/src/components/Dialog/index.tsx
```diff
@@ -171,8 +292,43 @@ unstable__Dialog.propTypes = {
...
+
+  /**
+   * Specify the ID of an element that labels this dialog
+   */
+  ariaLabelledBy: PropTypes.string,
+
+  /**
+   * Specify the ID of an element that describes this dialog
+   */
+  ariaDescribedBy: PropTypes.string,
```
tay1orjones: Is there a use case for each of these? If not let's remove them. If so, could we briefly specify why/when to use it? Also these should use the "native" sausage cased values like `aria-label` instead of the camel cased `ariaLabel`. https://github.com/carbon-design-system/carbon/issues/12802

packages/react/src/components/Dialog/index.tsx
```diff
@@ -122,21 +174,90 @@ export const unstable__Dialog = React.forwardRef(
...
+
+    // Find and set IDs for title and subtitle when dialog opens
+    useEffect(() => {
+      if (ref.current && open) {
+        const title = ref.current.querySelector(
+          `.${prefix}--dialog-header__heading`
+        );
+        const subtitle = ref.current.querySelector(
+          `.${prefix}--dialog-header__label`
+        );
```
tay1orjones: Instead of using selectors like this we can:

- Defer the setup of these id's to the consumer where they compose them directly on the title/subtitle
- Have Dialog.Root set up a context. Have Dialog.Title and Dialog.Subtitle put whatever info is needed into that context (id's?), so we can tap into that context from any of the other components.

It might be best to do both, too. This way we could provide default id's while also allowing them to override them directly on components.

packages/react/src/components/Dialog/index.tsx
```diff
@@ -242,20 +414,411 @@ export const DialogCloseButton = React.forwardRef<
...
+
+  /**
+   * Specify whether the content has overflow that should be scrollable
+   */
+  hasScrollingContent: PropTypes.bool,
+};
+
+/**
+ * -------------
+ * DialogContent (alias for DialogBody)
```
tay1orjones: Why is this alias needed?

packages/react/src/components/Dialog/index.tsx
```diff
@@ -242,20 +414,411 @@ export const DialogCloseButton = React.forwardRef<
...
+
+export const DialogFooter = React.forwardRef<HTMLDivElement, DialogFooterProps>(
+  (
+    {
+      children,
+      className,
+      onRequestClose,
+      onRequestSubmit,
+      primaryButtonText = 'Save',
+      secondaryButtonText = 'Cancle',
```
tay1orjones:
```suggestion
      secondaryButtonText = 'Cancel',
```

packages/react/src/components/Dialog/index.tsx
```diff
@@ -242,20 +414,411 @@ export const DialogCloseButton = React.forwardRef<
...
+(unstable__Dialog as DialogComponent).Header = Header;
+(unstable__Dialog as DialogComponent).Controls = Controls;
+(unstable__Dialog as DialogComponent).CloseButton = CloseButton;
+(unstable__Dialog as DialogComponent).Title = Title;
+(unstable__Dialog as DialogComponent).Subtitle = Subtitle;
+(unstable__Dialog as DialogComponent).Body = Body;
+(unstable__Dialog as DialogComponent).Content = Content;
+(unstable__Dialog as DialogComponent).Footer = Footer;
+
+export const Dialog = unstable__Dialog as DialogComponent;
```
tay1orjones: This is a bit weird, but I think it's okay for now. Once we add Dialog as an export to the main index.ts export file we can handle exporting Dialog as unstable__Dialog. For now though this makes sense

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -564,7 +564,7 @@ const Modal = React.forwardRef(function Modal(
       open={open}
       modal
       ref={innerModal}
-      role={isAlertDialog ? 'alertdialog' : ''}
+      role={isAlertDialog ? 'alertdialog' : 'dialog'}
```
tay1orjones: The spec recommends against Dialog having `role='dialog'` https://www.w3.org/TR/html-aria/#docconformance:~:text=Role%3A-,alertdialog,-.%20(dialog%20is

```suggestion
      role={isAlertDialog ? 'alertdialog' : ''}
```
I put that in the comments up above this a few lines where isAlertDialog is defined. Maybe it would be better to fully encapsulate that prop so there's no confusion.

```js
// Dialog elements can only have an `aria-role` of `alertdialog` or `dialog`, and `dialog` is not recommended, so only `alertdialog` is used here.
// https://www.w3.org/TR/html-aria/#docconformance:~:text=Role%3A-,alertdialog,-.%20(dialog%20is
const role = alert && !passiveModal ? 'alertdialog' : '';

...and then down here at the usage
<Dialog
  ...
  role={role}
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -13,130 +13,171 @@
...
+@use '../../utilities/update_fields_on_layer' as *;
+@use '../../utilities/z-index' as *;
 
 /// Dialog styles
 /// @access public
 /// @group dialog
 @mixin dialog {
   .#{$prefix}--dialog {
+    position: fixed;
+    z-index: z('modal');
```
tay1orjones: this is not needed, dialogs are automatically promoted to the [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer)
```suggestion
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -13,130 +13,171 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
     padding: 0;
     border: none;
-    background-color: $layer;
-    block-size: fit-content;
+    background-color: transparent;
```
tay1orjones: Why not `$layer`?

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -13,130 +13,171 @@
...
-    transform: translateY(calc(-1 * #{$spacing-06}));
-
-    /** opening and closing is used in as allow-discrete is not currently supported wide enough
-     * https://caniuse.com/mdn-css_properties_display_is_transitionable
-     */
-    transition:
-      opacity $duration-moderate-02 motion(exit, expressive),
-      transform $duration-moderate-02 motion(exit, expressive),
-      overlay $duration-moderate-02 motion(exit, expressive) allow-discrete,
-      display $duration-moderate-02 motion(exit, expressive) allow-discrete;
```
tay1orjones: Usage of `allow-discrete` was specifically included in the original implementation, we should keep this approach. Did you run into an issue with it?

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -13,130 +13,171 @@
...
+    &:not([open]) {
+      display: none;
     }
-    @include breakpoint(xlg) {
-      max-inline-size: 64%;
-    }
-    @include breakpoint(xlg) {
-      max-inline-size: 60%;
+
+    &::backdrop {
```
tay1orjones: Remember to [avoid nesting selectors](https://github.com/carbon-design-system/carbon/blob/0f1e73e5777089301af89fff52f2d72a7eb2b72c/docs/style.md#L901) and instead use a fully qualified selector where possible. IIRC this was something the previous iteration of this file did fairly well.

```suggestion
  .#{$prefix}--dialog::backdrop {
```

tay1orjones (changes requested):
Hey I found a few more things in addition to what we discussed earlier today.

I'm slightly concerned with the amount of removal in the styling. The existing styles had a good base to work with - I expected to see mostly additions and not such an overhaul. Could we take another pass at the styles preferring to keep the existing Dialog styles as much as possible, and only pulling over from Modal styles what's absolutely needed? Just one example is the addition of `display: flex` on `.#{$prefix}--dialog`, the background color being transparent, the loss of default sizing w/ `max-block-size` and `max-inline-size`. We recently received feedback leading to the addition of `block-size: fit-content` and that was removed as well. I don't think there should be a reason to remove `@starting-style`. 

These styles are being used in product, so I want to ensure the diff stays as small as possible to avoid regressions, even though these styles are behind a feature flag.

packages/react/src/components/Dialog/Dialog.stories.js
2nikhiltom: yikes ! thanks for the catch

packages/react/src/components/Dialog/index.tsx
2nikhiltom: image

this is not required , I added this because the issue had "Dialog.Content/Dialog.Body" 
Dialog.Body is good

packages/react/src/components/Dialog/index.tsx
2nikhiltom: Not required, i removed them.

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -8,24 +8,30 @@
...
+import Select from '../Select';
+import SelectItem from '../SelectItem';
 import { action } from '@storybook/addon-actions';
 import mdx from './Dialog.mdx';
 
 export default {
   title: 'Experimental/unstable_Dialog',
   component: Dialog,
-  includeStories: [],
+  // includeStories: [],
```
tay1orjones: The playwright tests won't work unless we include these in the storybook. I think it would be weird to put them in there when the Dialog stuff is not exported. I wonder if we could hide them from the sidebar with `tags: ['!dev']`?

* https://github.com/storybookjs/storybook/issues/9209#issuecomment-2189400640
* https://storybook.js.org/docs/writing-stories/tags#docs-only-stories

If not we could consider exporting them. Another option would be to add `.todo` to the playwright tests for now. That's probably the easiest option and we could make an issue as part #18243 to re-enable them.

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -128,46 +145,265 @@ export const NonModal = ({ open: _open, ...args }) => {
...
-        </Button>
+        </Dialog.Header>
+        <Dialog.Body>
+          <p>
+            Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
+            autem officiis dolores facilis nulla earum! Neque quia nemo sequi
+            assumenda ratione officia Voluptate beatae eligendi placeat nemo
+            laborum, ratione.
+          </p>
+          <br></br>
```
tay1orjones:
```suggestion
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -128,46 +145,265 @@ export const NonModal = ({ open: _open, ...args }) => {
...
+
+export const WithScrollingContent = ({ open: _open, ...args }) => {
+  const [open, setOpen] = useState(_open);
+
+  function toggleDialog() {
+    setOpen(!open);
+  }
+
+  function closeDialog(e) {
+    console.log('e', e);
```
tay1orjones:
```suggestion
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -128,46 +145,265 @@ export const NonModal = ({ open: _open, ...args }) => {
...
+    setOpen(false);
+  }
+
+  function handleRequestClose(e) {
+    console.log('ee', e);
```
tay1orjones:
```suggestion
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -128,46 +145,265 @@ export const NonModal = ({ open: _open, ...args }) => {
...
+              odio, elementum vel vehicula vitae, porttitor ac lorem. Sed
+              viverra elit ac risus tincidunt fermentum. Ut sollicitudin nibh id
+              risus ornare ornare. Etiam gravida orci ut lectus dictum, quis
+              ultricies felis mollis. Mauris nec commodo est, nec faucibus nibh.
+              Nunc commodo ante quis pretium consectetur. Ut ac nisl vitae mi
+              mattis vulputate a at elit. Nullam porttitor ex eget mi feugiat
+              mattis. Nunc non sodales magna. Proin ornare tellus quis hendrerit
+              egestas. Donec pharetra leo nec molestie sollicitudin.
+            </p>
+            <br></br>
```
tay1orjones:
```suggestion
```

packages/react/src/components/Dialog/Dialog.stories.js
```diff
@@ -128,46 +145,265 @@ export const NonModal = ({ open: _open, ...args }) => {
...
+              assumenda ratione officia Voluptate beatae eligendi placeat nemo
+              laborum, ratione.
+            </p>
+            <p>
+              Elit hic at labore culpa itaque fugiat. Consequuntur iure autem
+              autem officiis dolores facilis nulla earum! Neque quia nemo sequi
+              assumenda ratione officia Voluptate beatae eligendi placeat nemo
+              laborum, ratione.
+            </p>
+            <br></br>
```
kennylam:
```suggestion

```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -97,7 +97,10 @@
     position: relative;
     overflow: visible;
     inline-size: 100%;
+    margin-block-end: 0.5rem;
     min-block-size: $spacing-09;
+    padding-block-start: 1rem;
+    padding-inline: 1rem 3rem;
```
kennylam:
```suggestion
    margin-block-end: $spacing-03;
    min-block-size: $spacing-09;
    padding-block-start: $spacing-05;
    padding-inline: $spacing-05 $spacing-09;
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -145,4 +148,121 @@ because the nesting selector cannot represent pseudo-elements. */
...
+    grid-template-rows: auto 1fr auto;
+    inline-size: 100%;
+  }
+
+  .#{$prefix}--dialog-content {
+    position: relative;
+    grid-column: 1 / -1;
+    grid-row: 2 / -2;
+    padding-block: 0.5rem 3rem;
+    padding-inline: 1rem;
```
kennylam:
```suggestion
    padding-block: $spacing-03 $spacing-09;
    padding-inline: $spacing-05;
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -145,4 +148,121 @@ because the nesting selector cannot represent pseudo-elements. */
...
+
+    mask-image: linear-gradient(
+        to bottom,
+        $layer calc(100% - 80px),
+        transparent calc(100% - 48px),
+        transparent 100%
+      ),
+      linear-gradient(to left, $layer 0, 16px, transparent 16px),
+      linear-gradient(to right, $layer 0, 2px, transparent 2px),
+      linear-gradient(to top, $layer 0, 2px, transparent 2px);
```
kennylam:
```suggestion
    border-block-end: $spacing-01 solid transparent;

    mask-image: linear-gradient(
        to bottom,
        $layer calc(100% - $spacing-11),
        transparent calc(100% - $spacing-09),
        transparent 100%
      ),
      linear-gradient(to left, $layer 0, $spacing-05, transparent $spacing-05),
      linear-gradient(to right, $layer 0, $spacing-01, transparent $spacing-01),
      linear-gradient(to top, $layer 0, $spacing-01, transparent $spacing-01);
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -145,4 +148,121 @@ because the nesting selector cannot represent pseudo-elements. */
...
+  }
+
+  .#{$prefix}--dialog-scroll-content:has(.#{$prefix}--autoalign) {
+    mask-image: none;
+  }
+
+  .#{$prefix}--dialog-footer {
+    display: flex;
+    justify-content: flex-end;
+    block-size: convert.to-rem(64px);
```
kennylam:
```suggestion
    block-size: $spacing-10;
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -145,4 +148,121 @@ because the nesting selector cannot represent pseudo-elements. */
...
+    grid-column: 1/-1;
+    grid-row: -1/-1;
+    margin-block-start: auto;
+  }
+
+  .#{$prefix}--dialog-footer .#{$prefix}--btn {
+    flex: 0 1 50%;
+    align-items: baseline;
+    margin: 0;
+    block-size: convert.to-rem(64px);
```
kennylam:
```suggestion
    block-size: $spacing-10;
```

packages/styles/scss/components/dialog/_dialog.scss
```diff
@@ -145,4 +148,121 @@ because the nesting selector cannot represent pseudo-elements. */
...
+    .#{$prefix}--dialog-content.#{$prefix}--dialog-scroll-content {
+    mask-image: linear-gradient(
+        to bottom,
+        $layer calc(100% - 80px),
+        transparent calc(100% - 48px),
+        transparent 100%
+      ),
+      linear-gradient(to left, $layer 0, 16px, transparent 16px),
+      linear-gradient(to right, $layer 0, 2px, transparent 2px),
+      linear-gradient(to top, $layer 0, 2px, transparent 2px);
```
kennylam:
```suggestion
      linear-gradient(to left, $layer 0, $spacing-05, transparent $spacing-05),
      linear-gradient(to right, $layer 0, $spacing-01, transparent $spacing-01),
      linear-gradient(to top, $layer 0, $spacing-01, transparent $spacing-01);
```

packages/styles/scss/components/dialog/_dialog.scss
2nikhiltom: Updated to use tokens

packages/react/src/components/Dialog/Dialog.stories.js
2nikhiltom: @tay1orjones  ~The Dialog and other inner components are being exported, I think it will be better to export them and have avt test cases enabled here in this PR. I have moved the  [browser-based test cases ](https://github.com/carbon-design-system/carbon/blob/04965e3c9be0bd3374ec139f9b3635c03ca1e599/packages/react/src/components/Dialog/Dialog-test.js#L78C1-L105C8 ) to Dialog-test.avt.e2e.js file. thoughts ?~
I agree it’s best not to include it in Storybook at this stage. I’ll keep it for now to support design reviews, and will remove it from Storybook before merging

tay1orjones (approved):

heloiselui (approved):
Looks good, just need to uncomment the includeStories: []

PR #19180, opened 2025-04-20
fix(Dropdown): fix type declarations to include ref

wkeese:
Refs #13022.

Like most Carbon components, `Dropdown` can take a `ref`, but for some reason that was hidden in its public type.

Also cleared up some contradicting code: The `ref` is on an `HTMLButtonElement` not an `HTMLDivElement`.

#### Changelog

**Changed**

- Exported type of Dropdown now includes `ref`.

#### Testing / Reviewing

Ran lint and build, and then tested against my own to confirm that it let me get rid of the `@ts-expect-error` directives when passing `ref` to a `Dropdown`.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -284,7 +284,7 @@ function stateReducer(state, actionAndChanges) {
   }
 }
 
-const Dropdown = React.forwardRef<HTMLButtonElement, DropdownProps<any>>(
+const Dropdown = React.forwardRef(
   <ItemType,>(
     {
       autoAlign = false,
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -11,6 +11,7 @@ import React, {
   isValidElement,
   MouseEvent,
   ReactNode,
+  Ref,
   useCallback,
   useContext,
   useEffect,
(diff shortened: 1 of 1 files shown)


Gururajj77 (approved):
Thanks for this contribution!

riddhybansal (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -284,7 +284,7 @@ function stateReducer(state, actionAndChanges) {
   }
 }
 
-const Dropdown = React.forwardRef(
+const Dropdown = React.forwardRef<HTMLButtonElement, DropdownProps<any>>(
```
adamalston: The usage of `any` here appears to override `ItemType` to a degree. Paste this code at the bottom of packages/react/src/components/Dropdown/Dropdown.Skeleton.tsx:

```tsx
const Test = () => (
  <Dropdown
    label="placeholder"
    id="dropdown"
    items={[]}
    titleText={<div />}
  />
);
```

When I hover over `items`, the intellisense tooltips says:

```
(property) DropdownProps<any>.items: any[]
```

Then, if I revert all of the changes in this file, the intellisense tooltip says:

```
(property) items: unknown[]
```

So it seems like the types are changing unexpectedly. Is `ItemType` supposed to be `unknown` instead of `any` in this scenario?

I also tried the following code to verify that a type error would be thrown if something was way off and that appears to be happening in both cases which is good.

```tsx
const Test = () => (
  <Dropdown
    label="placeholder"
    id="dropdown"
    items={[{ name: 'item1', title: '' }, { name: 'item2' }]}
    initialSelectedItem={{ display: 'item1' }}
    selectedItem={{ hmm: 'item1' }}
    titleText={<div />}
  />
);
```

wkeese: I'll just revert the change on that line.  The important thing is the type that's exported from DropDown.tsx, which is hardcoded with `as`:

```
export default Dropdown as DropdownComponent;
```

in order to workaround https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref.

PR #18643, opened 2025-02-20
fix: update Dropdown and MultiSelect to handle interactive labels

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/16987

Updated `Dropdown` and `MultiSelect` to handle interactive labels.

#### Changelog

**Changed**

- Updated `Dropdown` and `MultiSelect` to handle interactive labels.

#### Testing / Reviewing

```sh
yarn test packages/react/src/components/Dropdown packages/react/src/components/MultiSelect
```

I tried to write a test for these changes. I couldn't get it to pass. I think it has something to do with the changing visibility of the popover and its visibility from the test's perspective. Curious to hear if anyone has suggestions to test these changes. Alternatively, I created Storybook entries.

```js
it('should not steal the focus from interactive elements in titleText', async () => {
  render(
    <Dropdown
      {...mockProps}
      titleText={
        <>
          <ToggletipLabel>Dropdown label</ToggletipLabel>
          <Toggletip>
            <ToggletipButton>Info</ToggletipButton>
            <ToggletipContent>Tooltip content</ToggletipContent>
          </Toggletip>
        </>
      }
    />
  );
  await waitForPosition();

  // Initially, the toggletip content should not be visible
  expect(screen.queryByTestId('toggletip-content')).not.toBeVisible();

  // Click the toggletip button
  await userEvent.click(screen.getByTestId('toggletip-button'));

  // Wait for any asynchronous updates (if needed)
  await waitFor(() => {
    expect(screen.getByTestId('toggletip-content')).toBeVisible();
  });
});
```

Issue #16987, opened 2024-07-17
[Bug]: Interactive elements inside of Dropdown's titleText prop do not work correctly


jared-schwalbe:
### Package

@carbon/react

### Browser

Chrome

### Package version

1.60.3

### React version

16.14.0

### Description

When using a Toggletip inside of the node passed to the titleText prop of a Dropdown, the Toggletip refuses to open and instead simply passes focus to the Dropdown when clicked.

My use case is that I would like to have a tooltip next to the label for a dropdown and inside that tooltip I want to provide a link to some documentation.

A valid workaround is instead of using the titleText prop on the Dropdown, simply use a FormLabel as a sibling to the Dropdown.

### Reproduction/example

https://stackblitz.com/edit/github-c68srd-hwqb9s?file=src%2FApp.jsx&preset=node=

### Steps to reproduce

1. Pass a node to the titleText prop of Dropdown and include a Toggletip inside that node
2. Click on the Toggletip button

### Suggested Severity

None

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

tay1orjones:
Thanks again for opening this, I think you should be able to put interactive elements into `titleText` and have them work as expectd. It looks like the Toggletip contents never shows because focus is immediately being moved to the Dropdown for some reason.

jose-biescas:
Hi, just happened to encounter this same bug still in v1.75.0

I believe the cause of the bug is because the `Dropdown` component is utilizing the html element `<label>` tag to wrap the `titleText` property. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label from the documentation: 

When a user clicks or touches/taps a label, the browser passes the focus to its associated input (the resulting event is also raised for the input). That increased hit area for focusing the input provides an advantage to anyone trying to activate it — including those using a touch-screen device.

This is most likely why, the focus initially is on the `Tooltip`, but it is quickly overlapped by the `label`'s property of shifting that focus over to the input.

Those lines are here: 613-617

```
        {titleText && (
          <label className={titleClasses} {...getLabelProps()}>
            {titleText}
          </label>
        )}
```

Whereas other components like the `TextInput` re-use the `Text` component for their `labelText`.

[...]

PR #19192, opened 2025-04-21
fix(dropdown): add id to label for titleText elements

### LINKED ISSUE #19101: [a11y]: The 'id' "downshift-:r2:-label" specified for the ARIA property 'aria-labelledby' value is not valid
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

v1.40.0

### React version

_No response_

### Assistive technology

_No response_

### Description

The 'id' "downshift-:r2:-label" specified for the ARIA property 'aria-labelledby' value is not valid.

FYI: I have found that if we are using the Tooltip/ ToggleTip component inside the Dropdown, it causes a violation.

Image

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://stackblitz.com/edit/github-jdzenu?file=src%2FApp.jsx

### Steps to reproduce

https://stackblitz.com/edit/github-jdzenu?file=src%2FApp.jsx

### Suggested Severity

### Code of Conduct

annawen1:
Closes #19101 

When we pass in an element to the `titleText` prop, the `id` and `for` attributes provided by downshift don't get applied to the `label` element which results in a violation with the IBM Equal Access Accessibility Checker

example:
`The 'id' "downshift-:r2:-label" specified for the ARIA property 'aria-labelledby' value is not valid.`

This PR adds only `id` to the `label` element when the `titleText` is an element which allows for [interactive labels to still work](https://github.com/carbon-design-system/carbon/pull/18643)

#### Changelog

**Changed**

- change tests to check for `for` attribute instead of `id`
- in `Dropdown` and `Multiselect`, add `id` to `labelProps` when `titleText` is an HTMLElement

#### Testing / Reviewing

Go to `With Toggletip Label` stories in `Dropdown` and `Multiselect` and check that you are still able to open and interact with the Toggletip component in the labels

Run the Accessibility Checker to confirm the violation mentioned above doesn't appear


annawen1:
Hi @adamalston can you review this PR and double check it doesn't undo the fix you had made to address https://github.com/carbon-design-system/carbon/issues/16987?

packages/react/src/components/Dropdown/__tests__/Dropdown-test.js
```diff
@@ -472,6 +472,6 @@ describe('Test useEffect ', () => {
 
     const label = screen.getByText('Dropdown Title').closest('label');
 
-    expect(label).not.toHaveAttribute('id');
+    expect(label).not.toHaveAttribute('for');
```
adamalston: I notice that we're checking for the element not to have a `for` attribute. However, I don't see any other testing that an element should have that attribute in this file. So when should and element have it and when should it not? And should both be tested? I probably should've considered these things more in https://github.com/carbon-design-system/carbon/issues/16987.

Suggestion to replace the last two tests in this file with the following code or something similar to check all attributes. Note that I've also changed the test titles given the component updates.

```js
it('should add certain label props when `titleText` is a string', () => {
  render(<Dropdown {...mockProps} titleText="Dropdown Title" />);

  const label = screen.getByText('Dropdown Title').closest('label');
  const attributes = Array.from(label.attributes).reduce(
    (acc, { name, value }) => ({ ...acc, [name]: value }),
    {}
  );

  expect(attributes).toEqual({
    class: 'cds--label',
    for: 'downshift-:r2d:-toggle-button',
    id: 'downshift-:r2d:-label',
  });
});

it('should add certain label props when `titleText` is an element', () => {
  render(<Dropdown {...mockProps} titleText={<span>Dropdown Title</span>} />);

  const label = screen.getByText('Dropdown Title').closest('label');
  const attributes = Array.from(label.attributes).reduce(
    (acc, { name, value }) => ({ ...acc, [name]: value }),
    {}
  );

  expect(attributes).toEqual({
    class: 'cds--label',
    id: 'downshift-:r2g:-label',
  });
});
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -608,7 +608,9 @@ const Dropdown = React.forwardRef(
       return React.isValidElement(element) ? element : null;
     }, [slug, decorator]);
 
-    const labelProps = !isValidElement(titleText) ? getLabelProps() : null;
+    const labelProps = !isValidElement(titleText)
+      ? getLabelProps()
+      : { id: getLabelProps().id };
```
adamalston:
```suggestion
    const allLabelProps = getLabelProps();
    const labelProps = isValidElement(titleText)
      ? { id: allLabelProps.id }
      : allLabelProps;
```

adamalston (commented):
I checked the stories and the tests I added in https://github.com/carbon-design-system/carbon/issues/16987. They all still appear to be working correctly.

Separately, it seems like a good bit of code is shared between `Dropdown` and `MultiSelect`. Should some of that code be abstracted? When thinking about the future of these components, do we run the risk of someone updating something in one of them and forgetting to put that same update in the other? Nothing that needs to change in this PR if at all, just curious to hear your thoughts and others.

packages/react/src/components/Dropdown/__tests__/Dropdown-test.js
annawen1: Good point, I'll do the same for MultiSelect as well

annawen1:
> I checked the stories and the tests I added in #16987. They all still appear to be working correctly.
> 
> Separately, it seems like a good bit of code is shared between `Dropdown` and `MultiSelect`. Should some of that code be abstracted? When thinking about the future of these components, do we run the risk of someone updating something in one of them and forgetting to put that same update in the other? Nothing that needs to change in this PR if at all, just curious to hear your thoughts and others.

Yeah that makes sense to me to abstract those common parts!

adamalston (approved):

2nikhiltom (approved):

kennylam (approved):

CCCCChloe:
Hi I verified v11.81.0 but seeing the accessibility issue is still there. Could you check this code piece to see if there is any changes need to made? https://stackblitz.com/edit/github-7bh8iprp?file=src%2FApp.jsx

Screenshot 2025-04-30 at 10 59 56 AM

PR #19196, opened 2025-04-21
fix(tabs): call debounce function to update scrollLeft

### LINKED ISSUE #19061: [Bug]: Scroll buttons for tabs not working as expected when scrolling manually
### Package

@carbon/react

### Browser

Chrome

### Package version

1.76.0

### React version

18.2.0

### Description

I saw a similar issue https://github.com/carbon-design-system/carbon/issues/18719 but this case is slightly different.

When there are many tabs that don't have enough space scroll buttons appears and they work fine as long as we're using only them to scroll.
When we are using scroll and we are NOT using the arrows the left arrow will not appear (when we will scroll to the right). 
The right scroll arrow button won't disappear either when we reach the end of tabs. 
There will also be a problem with disappearing when we will scroll to the left.

Briefly: I think scrolling manually should be taken into account within scroll buttons behaviour. 

### Reproduction/example

https://stackblitz.com/edit/github-hvon3jwa?file=src%2FApp.jsx&preset=node=

### Steps to reproduce

1. Open the example
2. Scroll tabs manually, not using arrow buttons

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

warrenmblood:
Closes #19061 

This update fixes a bug in `Tabs` where the visibility of the scroll arrow buttons is not updated correctly when scrolling manually. The button visibility is determined by the state variable `scrollLeft`, which should be updated by the debounce function inside of the `onScroll` handler. The debounce function was previously being returned but not actually called by the handler.

I also added a new playwright test to verify that the scroll button disappears after scrolling manually to the end of tabs.

#### Changelog

**New**
- Playwright test for `Tabs` checking if scroll button disappears after reaching end of tabs via manual scroll

**Changed**

- `debouncedOnScroll` now calls the debounce function which updates the state variable `scrollLeft`
- `scrollLeft` is updated to the value of `ref.current.scrollLeft` to replace the deprecated declaration of `event`

#### Testing / Reviewing

- Run playwright test for manual horizontal scroll
- Verify that the arrow button visibility updates appropriately when scrolling manually- the right arrow should disappear after reaching end of tabs and the left one should disappear at the beginning of tabs


tay1orjones (approved):
LGTM!

riddhybansal (approved):
LGTM

PR #19205, opened 2025-04-22
fix(code-connect): update to 1.3.2 and component fixes

alisonjoseph:
Updates to latest version of code connect, and fixes for components that have changed within Figma since the last update.

#### Changelog

- update to v1.3.2 of @figma/code-connect
- remove hideLabel to match with Figma no longer having a show label prop for Text Inputs
- update prop naming to match latest from Figma
- update Breadcrumb url (it was broken for some reason)


kennylam (approved):

2nikhiltom (approved):
yarn dedupe --check should fix the ci failure

PR #19154, opened 2025-04-16
ci(release): add patch version workflow automation

tay1orjones:
This adds a new workflow we can invoke manually to release a patch version.

#### Changelog

**New**

- version-patch.yml

#### Testing / Reviewing

- There's no way to test these until they're merged in
- Take a cursory look for any syntax issues or misspellings

PR #19226, opened 2025-04-23
fix: update version patch workflow

adamalston:
Related to https://github.com/carbon-design-system/carbon/pull/19154.

Updated version patch workflow.

#### Changelog

**New**

- Added instructions to the contributing doc detailing how to create a patch release.

**Changed**

- Updated version patch workflow.

#### Testing / Reviewing

I left some comments on https://github.com/carbon-design-system/carbon/pull/19154 after it was merged. I've addressed some of them here.

The original version of this workflow attempted to cherry-pick commits directly in GitHub Actions. However, GitHub Actions cannot resolve merge conflicts without a lot of configuration.

This PR refactors the workflow to offload cherry-picking and conflict resolution to the author of the patch. Instead of taking commits as inputs, the updated workflow assumes a release branch has already been created and pushed, with the appropriate commits already applied via `git cherry-pick`.

As with https://github.com/carbon-design-system/carbon/pull/19154, I don't think this workflow can be fully tested until this PR is merged. If someone is aware of a way to fully test it before this PR is merged, let me know.


.github/CONTRIBUTING.md
```diff
@@ -382,3 +382,52 @@ services:
 
 - [`@carbon/react` template](https://codesandbox.io/s/github/carbon-design-system/carbon/tree/v11.2.0/examples/codesandbox)
 - [`@carbon/styles` template](https://codesandbox.io/s/github/carbon-design-system/carbon/tree/v11.2.0/examples/codesandbox-styles)
+
+## Creating a patch release
```
annawen1: I think we can probably update the release docs instead: https://github.com/carbon-design-system/carbon/blob/main/docs/release.md#manual-patch-release

adamalston: Thanks, I knew that existed somewhere.

My only question: do you think I should replace the existing content with this content, or find a way to update what's already there? Another option might be to hold off on documenting this workflow for now, since it's not yet confirmed to be working. If we can confirm that it does work, then we could document it in https://github.com/carbon-design-system/carbon/blob/dc698e29ce768b0415fe6ade56db2723a938a155/docs/release.md#manual-patch-release.  

Any thoughts on which approach I should take?

annawen1: I'm leaning towards holding off on updating until we can confirm this workflow works. When we do update the doc, it would probably be best to keep a reference of the manual steps in there as well

tay1orjones (commented):
Most of the time there's no conflicts with the cherry-pick. This most recent release was the first time that had happened to me while cutting a patch.

My goal in making this workflow was to get as close as we could to a single button press to cut a patch. Having to manually make the release branch, cherry pick commits, etc. opens up the possibility of mistakes during the release process. 

I'm not sure how valuable this workflow ends up being without handling those steps.

adamalston:
Ah, okay. Do you think any of the comments I left in https://github.com/carbon-design-system/carbon/pull/19154 should be addressed? If so, I can address them here. Otherwise, I can close this PR.

tay1orjones:
Sorry this fell off my radar! Your comments on the old PR are valid and could be addressed if you'd like

tay1orjones (approved):

annawen1 (approved):

PR #19231, opened 2025-04-24
fix: controlled tag component for selected prop #19228

### LINKED ISSUE #19228: [Bug]: Selectable tags are not controlled. Not responding to selected prop once mounted
### Package

@carbon/react

### Browser

Chrome

### Package version

1.81.0

### React version

18

### Description

Component not reading selected prop once mounted. - https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tag/SelectableTag.tsx#L98
When updating the selected prop value, the component is not updating.
I have created an example - https://stackblitz.com/edit/ktw9c81w
I am trying to load several tags in a table; tags may be duplicated in every row. When selecting a tag in one row, all other row tags should behave the same.

https://github.com/user-attachments/assets/027bbd46-966e-4f0d-83b6-0324eddd50ed

### Reproduction/example

https://stackblitz.com/edit/ktw9c81w

### Steps to reproduce

Try to update selected prop if a tag once tag is mounted in DOM

### Suggested Severity

### Application/PAL

Event automation

### Code of Conduct

sojinantony01:
Closes #19228

Controlled selectable tag component

#### Changelog

**New**

- Controlled Tab component

#### Testing / Reviewing
https://stackblitz.com/edit/github-g93j9mdu?file=src%2FApp.jsx  - component should be controllable with latest version


sojinantony01:
recheck

sojinantony01:

packages/react/src/components/Tag/SelectableTag.tsx
```diff
@@ -101,6 +102,12 @@ const SelectableTag = forwardRef(
     });
     const [isEllipsisApplied, setIsEllipsisApplied] = useState(false);
 
+    useEffect(() => {
+      if (selected !== selectedTag) {
+        setSelectedTag(selected);
+      }
+    }, [selected]);
```
tay1orjones: Does this need an effect? 

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

tay1orjones (changes requested):
Thanks for writing a test to cover this! I made a more isolated repro for what it's worth: https://stackblitz.com/edit/github-g93j9mdu?file=src%2FApp.jsx

packages/react/src/components/Tag/SelectableTag.tsx
sojinantony01: I think we still need useEffect here. `selected` and `onChange`  are optional props to the component, so if parent component not passing `selected` prop Tag component should handle is inside own state.
Do you agree? 🤔 

```
      if (selected !== selectedTag) {
        setSelectedTag(selected);
      }
```
If we add this to render, i think it is good to remove the state and use value directly from prop, but this case props needs to be mandatory.


Otherwise i can check if props passed or not in the condition

```
      if (typeof selected == "boolean" && onChange && selected !== selectedTag) {
        setSelectedTag(selected);
      }
```
🤔.

sojinantony01: @tay1orjones  Please have a look

tay1orjones: @sojinantony01 I forgot we have the `useControllableState` hook you can use for this.

https://github.com/carbon-design-system/carbon/blob/15b90a42ed5d35b269396669f3dd72328b58b8dc/packages/react/src/internal/useControllableState.ts#L38

It'll handle the whole thing for you. We'll need to add a `defaultSelected` boolean prop to SelectableTag to pass to the hook. Probably the closest usage example you could follow is in Toggle:

https://github.com/carbon-design-system/carbon/blob/a6eb26586b0531ff65150398adebdee7cdd6e4e2/packages/react/src/components/Toggle/Toggle.tsx#L121-L125

We'll want to make sure there's clear passing tests for both controlled and uncontrolled modes here too.

sojinantony01: Thats really a useful hook. Thanks @tay1orjones.
I have updated the component using the `useControllableState` hook. we already have tests for uncontrolled state and i have added couple of tests for the controlled case - All passing.

And have manually tested the component using storybook.
Please have a look in the latest code.

packages/react/src/components/Tag/Tag-test.js
```diff
@@ -308,4 +308,34 @@ describe('Tag', () => {
...
+  it('Controlled selectable tag, should call onChange', async () => {
+    const onChange = jest.fn();
+
+    const { container } = render(
+      <SelectableTag text="Tag content" onChange={onChange} selected={true} />
+    );
+
+    const selectableTag = container.firstChild;
+
+    await userEvent.click(selectableTag);
```
tay1orjones: I'd prefer the tests to use `screen.getByRole` instead of `container.firstChild` and assert `aria-pressed` in addition to the `toHaveClass` assertions (since a user doesn't directly interact with the classnames), but it's not a blocker to merge. 

```suggestion
    await userEvent.click(screen.getByRole("button", {name: "Tag content"}));
```

tay1orjones (approved):

packages/react/src/components/Tag/SelectableTag.tsx
```diff
@@ -95,7 +97,11 @@ const SelectableTag = forwardRef(
     const prefix = usePrefix();
     const tagRef = useRef<HTMLButtonElement>(null);
     const tagId = id || `tag-${useId()}`;
-    const [selectedTag, setSelectedTag] = useState(selected);
+    const [selectedTag, setSelectedTag] = useControllableState({
+      value: selected,
+      onChange: onChange,
+      defaultValue: false,
```
tay1orjones: Shouldn't this be configurable through a new `defaultSelected` prop, similar to what's done in other usages like Toggle and Tabs? https://github.com/carbon-design-system/carbon/blob/a6eb26586b0531ff65150398adebdee7cdd6e4e2/packages/react/src/components/Tabs/Tabs.tsx#L152

It would have a default function param of `false`

tay1orjones (changes requested):
One minor thing, I think the default should be configurable

packages/react/src/components/Tag/SelectableTag.tsx
sojinantony01: Sure, Thanks @tay1orjones 
Added a new defaultSelected prop and tests for same.

packages/react/src/components/Tag/Tag-test.js
sojinantony01: Sure, Updated the tests.

tay1orjones (approved):
LGTM, thanks again!

riddhybansal (approved):
LGTM!!

PR #19239, opened 2025-04-25
fix(date-picker): clear flatpickr instance on reset

### LINKED ISSUE #18982: [Bug]: DatePicker Range: End Date Retains Old Value After Resetting
### Package

@carbon/react

### Browser

Chrome

### Package version

1.79.0

### React version

18.2.0

### Description

When using the Carbon DatePicker component in range mode, I encountered an issue where the end date input retains its old value after the component is reset. This occurs when both start and end dates are selected, and the inputs are reset (e.g., by clearing the values in the state). Upon selecting a new start date after the reset, the end date automatically reverts to its previous value instead of remaining blank.

This behavior suggests that the DatePicker does not fully synchronize its internal state with the value prop after reset. A fix or workaround to enforce proper synchronization is necessary

### Reproduction/example

https://stackblitz.com/edit/github-hs9y3qpy-jnjsw5zr?file=src%2FApp.jsx

### Steps to reproduce

1. Use the DatePicker with datePickerType="range" in a React project.

2. Populate both the start date and end date inputs with valid dates.

3. Reset the component by clearing the state (e.g., setting value to [null, null]).

4. Select a new start date.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

warrenmblood:
Closes #18982 

Fixed behavior in the above issue. The flatpickr date instance was retaining date values after resetting the DatePicker `value` prop to [null, null]. This was due to the flatpickr instance only being cleared when the `value` prop was manually set to an empty string. This change will also clear the flatpickr instance when all dates in `value` are set to an empty string, `null`, `undefined`, or other JS falsy values.

#### Changelog

**New**

- Added unit test checking if end date value is retained after manually setting to `null`

**Changed**

- `useEffect` with `value` prop as dependency now clears the flatpickr instance when dates are set to empty values (null, undefined, "", 0)

#### Testing / Reviewing

- New unit test should pass
- Verify that we should allow users to manually reset this way. Given that users are trying to do this and running into issues, I would think so, but I'm not sure if there is a specific reason we require a reset using an empty string.


packages/react/src/components/DatePicker/DatePicker.tsx
```diff
@@ -853,10 +853,18 @@ const DatePicker = React.forwardRef(function DatePicker(
...
+
   useEffect(() => {
-    //when value prop is set to empty, this clears the flatpicker's calendar instance and text input
-    if (value === '') {
+    // when value prop is set to empty, this clears the flatpickr calendar instance and text input
+    // run if both:
+    // 1. value prop is set to an empty value (any falsy value in JS) OR an array with all empty values
+    // 2. flatpickr instance contains values in its selectedDates property so it hasn't already been cleared
+    if (
+      (!value || (Array.isArray(value) && value.every((date) => !date))) &&
```
adamalston, line 863: `value` can be a `number`. Is this condition, and more specifically `!value`, going to work as expected if `value` is `0`?

warrenmblood: Yes, if `value` is set to `0`, the condition will evaluate to `true` and the flatpickr instance will be cleared the same as if `value` were `""`, `undefined`, or `null`

adamalston: Okay, I asked because of this statement in https://github.com/carbon-design-system/carbon/pull/19239#issue-3021030876:

> `useEffect` with `value` prop as dependency now clears the flatpickr instance when dates are set to empty values (null, undefined, "", 0)

`0` isn't an "empty" value. It's a valid Unix timestamp which will be parsed as 1 January 1970 00:00:00 (UTC).

warrenmblood: Ok, "falsy" would be more accurate than "empty" there. I updated the file comments and storybook docs to provide better clarity around these changes. I think manually setting `value` to `0` would be a rare case, but it would work as a reset here.

heloiselui (approved):
LGTM!

packages/react/src/components/DatePicker/DatePicker.tsx
```diff
@@ -853,10 +853,18 @@ const DatePicker = React.forwardRef(function DatePicker(
...
     }
   }, [inline]);
+
   useEffect(() => {
-    //when value prop is set to empty, this clears the flatpicker's calendar instance and text input
-    if (value === '') {
+    // when value prop is manually reset, this clears the flatpickr calendar instance and text input
+    // run if both:
+    // 1. value prop is set to a falsy value ("", undefined, null, etc) OR an array of all falsy values
+    // 2. flatpickr instance contains values in its selectedDates property so it hasn't already been cleared
```
kennylam:
```suggestion
    // 1. value prop is set to a falsy value (`""`, `undefined`, `null`, etc) OR an array of all falsy values
    // 2. flatpickr instance contains values in its `selectedDates` property so it hasn't already been cleared
```

tay1orjones (approved):

PR #18774, opened 2025-03-06
feat(tag): web component parity

annawen1:
Closes #18653 

This updates the `cds-tag` component with the latest features from React.

New components include:
- `cds-tag-skeleton`
- `cds-dismissible-tag`
- `cds-selectable-tag`
- `cds-operational-tag`

**NOTE:** The tag component incorporates the `definition-tooltip` that's used when the tag content results in an ellipsis. However since `definition-tooltip` doesn't exist in Web Components yet, I'm going to create a ticket to add the `definition-tooltip` to the tag component once completed.

#### Changelog

**New**

- New tag components 
  - `cds-tag-skeleton`
  - `cds-dismissible-tag`
  - `cds-selectable-tag`
  - `cds-operational-tag`

**Changed**

- modified the `tooltip` event methods to accommodate for tag functionality where the tooltip content doesn't appear outside of keyboard interactions

#### Testing / Reviewing

Review with the `tag` stories on Storybook and compare with React version

PR #19245, opened 2025-04-28
fix: box sizing on the tag

devadula-nandan:
Contributes to #18774

the react version of tag returns a button, that inherits the user agent styles of button the browser provides (box-sizing included).
however wc tags (those having the border) being a custom element with role="button". doesn't get any user agent styles unless it extends `HTMLButtonElement`. they couldn't inherit as well due to shadow dom encapsulation. so manually adding box-sizing would fix this

| before | After |
| ------ | ------ |
| Screenshot 2025-04-28 at 1 38 54 PM | Screenshot 2025-04-28 at 1 39 00 PM |

#### Changelog

**Changed**

- added box sizing to variants that has borders

#### Testing / Reviewing

visually through storybook

### DIFF
--- a/packages/web-components/src/components/tag/tag.scss
+++ b/packages/web-components/src/components/tag/tag.scss
@@ -20,6 +20,8 @@ $css--plex: true !default;
 :host(#{$prefix}-tag),
 :host(#{$prefix}-dismissible-tag) {
   @extend .#{$prefix}--tag;
+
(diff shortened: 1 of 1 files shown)


annawen1 (approved):
Nice! LGTM

heloiselui (approved):
LGTM!

Issue #19076, opened 2025-04-09
[Bug]: focus breaks with menuTarget prop on MenuButton


alisonjoseph:
### Package

@carbon/react

### Package version

v1.79

### React version

v1.79

### Description

It appears when using menuTarget on MenuButton the focus breaks. When opening focus should go to the first child, however when adding menuTarget that is no longer the case. 

With menuTarget (click on select category)
https://labs.carbondesignsystem.com/?path=/story/react_components-uishell--default
Without menuTarget  (click on select category)
https://labs.carbondesignsystem.com/?path=/story/react_components-uishell--header-story

You can also see the difference here https://stackblitz.com/edit/jgpfytus-9dpdpkm7?file=src%2FApp.jsx

Related issue in labs https://github.com/carbon-design-system/carbon-labs/issues/536

### Reproduction/example

https://stackblitz.com/edit/jgpfytus-9dpdpkm7?file=src%2FApp.jsx

### Steps to reproduce

Add MenuButton with and without menuTarget prop

### Suggested Severity

None

### Application/PAL

platforms/labs

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

adamalston:
Possible fix in https://github.com/carbon-design-system/carbon/pull/19059#issuecomment-2795702883.

2nikhiltom:
This happened because during initial render, containerRef.current is null. After mounting, the ref contains the DOM element, but React doesn't detect this change because the ref object itself hasn't changed and the menu renders at the default location (document.body) instead of the target element.

A better approach to pass `menuTarget` could be  to Use state to capture the ref value. When the ref populates after mounting, setTarget() triggers a re-render with the correct DOM element reference, ensuring the MenuButton receives the proper target.

here is the updated stackbliz :  https://stackblitz.com/edit/jgpfytus-wt3mzlgu?file=src%2FApp.jsx

adamalston:
Was this issue supposed to be addressed by https://github.com/carbon-design-system/carbon/pull/19246? If so, the Closed keyword wasn't added before the issue link in the PR's description.

maradwan26:
Hey @alisonjoseph it looks like this has been fixed in later versions.

[...]

PR #19246, opened 2025-04-28
docs: updates menuTarget prop usage docs

### LINKED ISSUE #18997: [Bug]: MenuButton does not always respect menuTarget

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/18997,  https://github.com/carbon-design-system/carbon-labs/issues/536,https://github.com/carbon-design-system/carbon/issues/19076,  https://github.com/carbon-design-system/carbon/issues/19076

There are multiple bugs reported around the usage of MenuTarget prop in MenuButton/Menu component
[Render issue](https://github.com/carbon-design-system/carbon/issues/18997 )
[focus breaking issue](https://github.com/carbon-design-system/carbon/issues/19076)
[focus/theme](https://github.com/carbon-design-system/carbon-labs/issues/536)

```
 <MenuButton
        defaultopen
        label="with MenuTarget"
        menuTarget={targetRef.current}
      >
```
Issues occurred when` targetRef.current `was passed to `menuTarget`

The functionality broke because during initial render, containerRef.current is null. After mounting, the ref contains the DOM element, but React doesn't detect this change because the ref object itself hasn't changed and the menu renders at the default location instead of the target element.

A better approach to pass menuTarget could be to use React.useState to capture the ref value. When the ref populates after mounting, setTarget() triggers a re-render with the correct DOM element reference, ensuring the MenuButton receives the proper target.

using the states to keep track to changes fixes inconsistent rendering and focus order issues  : https://stackblitz.com/edit/github-aq8alsec-sczywbg7?file=src%2FApp.tsx

#### Changelog

**New**
Adds menuTarget prop usage docs 

#### Testing / Reviewing

verify that usage of states fixes the issues  https://stackblitz.com/edit/github-aq8alsec-sczywbg7?file=src%2FApp.tsx
verify that docs makes sense


riddhybansal (approved):

packages/react/src/components/MenuButton/MenuButton.mdx
```diff
@@ -115,6 +115,26 @@ components with focus management.
...
+    setTarget(containerRef.current);
+  }
+}, []);
+<div ref={containerRef}>
+  <MenuButton label="Actions" menuTarget={target}>
+    <MenuItem label="First action" />
+  </MenuButton>
+</div>
+
+```
```
adamalston: Suggestion to format this doc, fix the code snippet (missing `return`), and convert the code snippet to a valid TypeScript component so it's easier to copy and paste code:

```suggestion
## Managing the `menuTarget` prop with React refs

Because React does not trigger re-renders when a ref's `.current` value changes,
you must capture the ref value in state to update the `menuTarget` prop.

```tsx
const Component = () => {
  const containerRef = useRef<HTMLDivElement>(null);
  const [target, setTarget] = useState<HTMLDivElement | undefined>(undefined);

  useEffect(() => {
    if (containerRef.current) {
      setTarget(containerRef.current);
    }
  }, []);

  return (
    <div ref={containerRef}>
      <MenuButton label="Actions" menuTarget={target}>
        <MenuItem label="First action" />
      </MenuButton>
    </div>
  );
};
```
```

2nikhiltom: Makes sense, thanks for the suggestion

annawen1 (approved):

PR #17248, opened 2024-08-22
fix(overflowmenu): improve `autoAlign` fallback placements

tay1orjones:
Based on a [slack conversation](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1724344719376389?thread_ts=1723645080.434069&cid=C2K6RFJ1G), this improves the fallback placements for OverflowMenu's `autoAlign`. Here's an example of the problem:

https://github.com/user-attachments/assets/2d8747ae-2d63-4ced-a038-4c083ff3e1e6

With this PR it avoids vertically flipping unless it absolutely has to:

https://github.com/user-attachments/assets/46c9b497-d7a8-43e4-a3a2-fd255403fec3

#### Changelog

**New**

- reorder the `fallbackPlacements` based on the value of `menuAlignment`

**Changed**

- updated/improved code comments

**Removed**

- removed `fallbackAxisSideDirection` because it's [ignored](https://floating-ui.com/docs/flip#:~:text=The%20options%20flipAlignment%20and%20fallbackAxisSideDirection%20no%20longer%20have%20an%20effect%20if%20this%20option%20is%20explicitly%20specified) when `fallbackPlacements` is used

#### Testing / Reviewing

- Check out the feature flagg overflow auto align story, ensure it works in a more logical way to what you'd expect as shown in the video.

PR #19247, opened 2025-04-28
fix: Tab tip popover alignment to follow spec

### LINKED ISSUE #18379: [Bug]: Popover with `isTabTip` - `autoAlign` uses invalid alignments
### Package

@carbon/react

### Browser

Chrome

### Package version

11.74.0

### React version

18

### Description

The spec shows two alignments, but I think there's technically four allowed alignments for tab tip popovers. What's shown in the spec, plus a vertical inverse, so top-start (left) and top-end (right)

Horizontally it's using a 'centered' alignment, which is off spec

Image

Vertically it's allowing alignments to the side of the trigger, which shouldn't happen at all. It should just flip upwards.

Image

This is basically the same spec/config for OverflowMenu that was fixed/added in https://github.com/carbon-design-system/carbon/pull/17248. I suppose that config just needs ported to Popover itself when `isTabTip`.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-popover--tab-tip-experimental-auto-align

### Steps to reproduce

Scroll around on the story to recreate what is shown above.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #18379 

Fixing tab tip popover alignment to follow spec

#### Changelog

**Changed**

- Updated Popover component to restrict tab tip popover alignments to follow spec (only top-start, top-end, bottom-start, bottom-end)
- Fixed issue where tab tip popovers were using a centered alignment (off-spec)
- Prevented tab tip popovers from aligning to the sides of the trigger

#### Testing / Reviewing

- [ ]  Navigate to the tab tip popover story

- [ ] Scroll around and verify tab tip popovers only align in the four correct positions:
      Top-start (left)
      Top-end (right)
      Bottom-start (left)
      Bottom-end (right)

- [ ] Verify popover only flips vertically (top to bottom or bottom to top), not horizontally
- [ ] Ensure there's no "centered" alignment visible
- [ ] Compare behavior to OverflowMenu which was fixed in PR #17248

### DIFF
--- a/packages/react/src/components/Popover/index.tsx
+++ b/packages/react/src/components/Popover/index.tsx
@@ -277,35 +277,39 @@ export const Popover: PopoverComponent & {
             ),
             autoAlign &&
               flip({
-                fallbackPlacements: align.includes('bottom')
-                  ? [
-                      'bottom',
-                      'bottom-start',
-                      'bottom-end',
-                      'right',
-                      'right-start',
-                      'right-end',
-                      'left',
-                      'left-start',
-                      'left-end',
-                      'top',
-                      'top-start',
-                      'top-end',
-                    ]
-                  : [
-                      'top',
-                      'top-start',
-                      'top-end',
-                      'left',
-                      'left-start',
-                      'left-end',
-                      'right',
-                      'right-start',
-                      'right-end',
-                      'bottom',
-                      'bottom-start',
-                      'bottom-end',
-                    ],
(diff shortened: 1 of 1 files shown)


2nikhiltom (approved):
config on `TabTip` does the trick ! LGTM

Gururajj77 (approved):

PR #19256, opened 2025-04-30
fix(Menu): treat Tab key like Escape

### LINKED ISSUE #17983: [a11y]: OverflowMenu: browser focus lost on tab key
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.63

### React version

React

N/A

### Assistive technology

N/A

### Description

* If the focus is on a dropdown menu item, and the user hits the tab key, focus moves to the dropdown menu itself. It should go to the element after the ellipsis button for opening the overflow menu.

* If the user tabs again, focus gets lost (it goes back to the browser address bar or something like that). It should go to the element after the ellipsis button for opening the overflow menu.

`MenuButton` does better than `OverflowMenu`.  With `MenuButton`, hitting the tab key from a `MenuItem` just doesn't do anything.  Not ideal, but not too bad either.

You can compare the behavior of `MenuButton` and `OverflowMenu` at https://stackblitz.com/edit/github-ihxqde-xrt7yt?file=src%2FApp.jsx,package.json.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default

### Steps to reproduce

You can reproduce on https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default or https://stackblitz.com/edit/github-ihxqde-xrt7yt?file=src%2FApp.jsx,package.json.

Behavior varies slightly depending on `focusTrap`, see #13115, but with either setting focus eventually jumps back to the browser address bar.

### Suggested Severity

None

### Code of Conduct

wkeese:
Closes #17983.

Address problems when users press the Tab key when on a dropdown menu item, and focus ends up going somewhere weird. Especially in the case when the menu is in a Modal/Tearsheet.

Updated both MenuButton and OverflowMenu.

The Tab key should arguably move focus to the element after the MenuButton/ OverflowMenu, or for shift-tab, the element before.  But to keep things simple, I just return focus to the OverflowMenu/MenuButton itself.  Directing focus before or after the MenuButton would be hard/easy depending on whether the menu was rendered using portalling or the fixed strategy.  See https://floating-ui.com/docs/misc#clipping for details.

About tests, it would be good to add Playwright tests for this, but I didn't because:

1. yarn run test:e2e doesn't run any of the relevant tests anyway, it only runs *-test.js files whereas the relevant files have names like OverflowMenu-test.e2e.js.

2. As Taylor wrote in #19120, the Menu keyboard test Menu-test.avt.e2e.js is currently disabled.

Also:

* Deprecated focusTrap.  I think it's no longer (or never was) necessary.

* Fix race condition preventing focus when opening submenus (with the right arrow key).  For some reason, when the user used the right-arrow to open a submenu, it called focus() before the menu was actually displayed, thus failing.  See the menu example in Storybook (and test it with keyboard rather than mouse).

#### Changelog

**Changed**

- Handle Tab key on menu items same way as Escape key.
- Deprecated Menu's `focusTrap` attribute.
- Fix focus when opening submenu using the right arrow.

#### Testing / Reviewing

Tested in Storybook (MenuButton, OverflowMenu, and Menu stories).


wkeese:
I'm not sure what that "merge playwrights reports" error is; it seems like it's just some problem with the Github CI rather than my PR.

Relatedly, I discovered that "yarn playwright" is a sort-of "hidden" command to run tests (not listed by "yarn run"), so I guess that's how you run the AVT tests.

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -383,9 +385,12 @@ const Menu = forwardRef<HTMLUListElement, MenuProps>(function Menu(
     return [fitValue(ranges.x, 'x') ?? -1, fitValue(ranges.y, 'y') ?? -1];
   }
 
+  // When a menu is opened, focus the first item.
   useEffect(() => {
     if (open && focusableItems.length > 0) {
-      focusItem();
+      // setTimeout() necessary for nested menus.  Otherwise, we try to focus before the menu is visible.
+      // Unclear why that happens.
+      setTimeout(focusItem, 0);
```
preetibansalui: It’s working for me even without using `setTimeout`. I'm testing it under the Menu > Default story with nested menus. Could you please check again on your end?

wkeese: Here's what happens for me, using keyboard (right arrow) to open the menu.   This is on Chrome/Mac.
https://github.com/user-attachments/assets/be816178-9dcb-4cdf-aba5-976c8f690d9e

That's with this code:

```js
// When a menu is opened, focus the first item.
useEffect(() => {
  if (open && focusableItems.length > 0) {
    focusItem();
  }
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [open, focusableItems]);
```

How about you?

preetibansalui: Ah, it didn’t focus for me today either, maybe it was a cache issue that day.
BTW, did you try using `useLayoutEffect`? This way you don't need to add setTimeout.

wkeese: I didn't try useLayoutEffect(), but I think there are [good reasons not to](https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a), plus we're using setTimeout in other places like:

```
useEffect(() => {
  if (!enableDialogElement && !open && launcherButtonRef) {
    setTimeout(() => {
      if ('current' in launcherButtonRef) {
        launcherButtonRef.current?.focus();
      }
    });
  }
}, [open, launcherButtonRef, enableDialogElement]);
```

preetibansalui: Yes, we're using setTimeout in a few other places as well, but I wanted to understand why it's needed here.
useLayoutEffect gives us the same result as setTimeout, but in a cleaner and React-supported way. Since both approaches work similarly in this case, would you mind removing the comment `// Unclear why that happens`

wkeese: > Yes, we're using setTimeout in a few other places as well, but I wanted to understand why it's needed here.

I'd like to understand that too.   It's a mystery.  That's why I added the comment.
 
> useLayoutEffect gives us the same result as setTimeout, but in a cleaner and React-supported way.

I think you're confused.   The focus() call is already inside a useEffect(), and that's not good enough.  So a useLayoutEffect() is definitely not good enough.   You need a setTimeout() or something more.

If you're suggesting to put a useLayoutEffect() inside of a useEffect()... that's just not allowed by React.

preetibansalui: I'm not suggesting using `useLayoutEffect()` inside a `useEffect()`. Instead, we can simply replace` useEffect()` with `useLayoutEffect()` to avoid setTimeout like this:

```
 useLayoutEffect(() => {
    if (open && focusableItems.length > 0) {
      focusItem()
    }
  }, [open, focusableItems]);
```

Check the recording to see how it’s working for me.


https://github.com/user-attachments/assets/d16761e2-081f-4bed-bcf4-ecd717701536


it's because useLayoutEffect works where we need to interact with the DOM after it’s updated but before the browser paints.

wkeese: If that's working for you, then it's probably just a race condition, and thus unreliable.  useLayoutEffect() executes **before** useEffect(), so if useEffect() doesn't work without setTimeout(), then useLayoutEffect() won't work without setTimeout() either.

And I said above, useLayoutEffect() is only supposed to be used for special layout-related rendering.  It should be avoided in general as it causes server warnings etc.

heloiselui:
Hi @wkeese, sorry for the delay, there are some conflicts in this PR now. Whenever you have a moment, could you take a look so we can move it forward? Thank you!

wkeese:
Wow, that's a long delay.  OK, I rebased against the latest code and fixed the conflict.

I noticed that the `MenuButton` doesn't get focus by (mouse) clicking on it, so you can't test this fix using the mouse.  You need to tab to the `MenuButton` and then use the space key to open the menu.  I tested on:

* http://localhost:3000/iframe.html?args=&globals=&id=components-menubutton--default&viewMode=story
* http://localhost:3000/iframe.html?globals=&id=components-overflowmenu--default&viewMode=story

heloiselui:
Tested locally and the fix behaves as described: Tab now closes the menu and returns focus to the trigger. That’s different from the APG expectation (Tab moves to the next/previous focusable), but the PR calls this out as an intentional trade‑off. @tombrunet could you help us reviewing this PR and confirm this approach is acceptable from an a11y perspective?

heloiselui (approved):
LGTM. I checked with Tom, and his feedback was that the current implementation is indeed an improvement. He mentioned there's probably an even better way where focus just remains on the button and `aria-activedescendant` is used to identify the active menu item, but he thinks this PR is an improvement at least, especially for the `OverflowMenu` since focus seems to go somewhere completely weird.

kennylam (approved):

PR #19287, opened 2025-05-02
fix: breadcrumb with single character fails wcag 2.2 target size

### LINKED ISSUE #19266: [a11y]: Breadcrumb with single-character link names fails WCAG 2.2 target size
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.81.0

### React version

_No response_

N/A

### Assistive technology

_No response_

### Description

When Breadcrumb size is set to small and more than one adjacent links are named with a single character, WCAG 2.2 requirement for target size will fail because two interactive elements do not have sufficient spacing. I noticed this will not fail in the WC version as the margin after the link name is 8px, whereas in React version the margin decreases to 4px for small size

Image

### WCAG 2.1 Violation

WCAG 2.2: 2.5.8 Target Size (Minimum)

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-breadcrumb--default&args=size:sm

### Steps to reproduce

1. Go to Breadcrumb React storybook: https://react.carbondesignsystem.com/?path=/story/components-breadcrumb--default&args=size:sm
2. Set size=small in Controls
3. Change the names of two adjacent breadcrumb items to a single character (i.e. numbers)
4. Inspect the spacing between the breadcrumb items. Per WCAG 2.2 guidelines, undersized targets (those less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the bounding box of each, the circles do not intersect another target or the circle for another undersized target

### Suggested Severity

### Code of Conduct

preetibansalui:
Closes #19266 

Breadcrumb with single character in small size fails to wcag 2.2 target size, it should be minimum 24 px.

#### Changelog

**Changed**

Provide a minimum width of 20px to breadcrumb item

#### Testing / Reviewing

1. Go to Storybook > Breadcrumb > Default
2. change text to single corrector, eg: 1,2
3. Inspect and check the size, it's width should be more than 24px including margins.


packages/styles/scss/components/breadcrumb/_breadcrumb.scss
```diff
@@ -49,6 +49,11 @@
     margin-inline-end: $spacing-02;
   }
 
+  .#{$prefix}--breadcrumb--sm .#{$prefix}--breadcrumb-item .#{$prefix}--link {
+    justify-content: center;
+    min-inline-size: $spacing-04;
```
annawen1, line 54: @warrenmblood had mentioned this issue with the Target size doesn't exist in Web Components because the margin is 8px instead of 4px in React - I wonder if we need to discuss with design to see if it's just an issue of adjusting the margins? or if this fix needs to be applied to WC too (setting the margin to 4px to align with React and then applying this same fix)?

Screenshot 2025-05-02 at 10 18 40 AM

preetibansalui: I thought to avoid changing the margins because it would affect all breadcrumbs, not just the single-character ones, which could potentially break the existing design for current users. I think it's a good idea to confirm with Design.

warrenmblood (commented):
I agree with discussing this with design first and deciding how to match WC and React, but these changes look good to me for preventing the WCAG target size failure!

thyhmdo:
Hi, according to our guidance:

- 4px/.25rem is applied to margins of the small size 
- 8px/.5 rem is applied to margins of the medium size

image
image

preetibansalui:
> Hi, according to our guidance:
> 
> * 4px/.25rem is applied to margins of the small size
> * 8px/.5 rem is applied to margins of the medium size
> 
> image image

Thanks @thyhmdo !

@annawen1 : Small and medium breadcrumbs already have guided margins, so the minimum width solution should be fine.

annawen1:
> > Hi, according to our guidance:
> > 
> > * 4px/.25rem is applied to margins of the small size
> > * 8px/.5 rem is applied to margins of the medium size
> > 
> > image image
> 
> Thanks @thyhmdo !
> 
> @annawen1 : Small and medium breadcrumbs already have guided margins, so the minimum width solution should be fine.

Ok great, let's adjust the margin on WC breadcrumb since it's 8px and confirm the min-width also get applied there too!

preetibansalui:
> > > Hi, according to our guidance:
> > > 
> > > * 4px/.25rem is applied to margins of the small size
> > > * 8px/.5 rem is applied to margins of the medium size
> > > 
> > > image image
> > 
> > 
> > Thanks @thyhmdo !
> > @annawen1 : Small and medium breadcrumbs already have guided margins, so the minimum width solution should be fine.
> 
> Ok great, let's adjust the margin on WC small breadcrumb since it's currently 8px and confirm the min-width also gets applied there too!

Applied the same changes to web-components as well to make them align. Please review

packages/web-components/src/components/breadcrumb/breadcrumb.ts
```diff
@@ -66,6 +66,18 @@ class CDSBreadcrumb extends LitElement {
     `;
   }
 
+  updated(changedProperties) {
+    if (changedProperties.has('size')) {
+      const items = document
+        .querySelector(`${prefix}-breadcrumb`)
+        ?.querySelectorAll(`${prefix}-breadcrumb-item`);
```
annawen1:
```suggestion
      const items = this.querySelectorAll(`${prefix}-breadcrumb-item`);
```

You'll want to scope the querySelector to the current instance of breadcrumb instead of `document` otherwise if there are other instances of breadcrumb on the page, it will affect those other instances as well

packages/web-components/src/components/breadcrumb/breadcrumb.ts
```diff
@@ -29,7 +29,7 @@ class CDSBreadcrumb extends LitElement {
    * Specify the size of the Breadcrumb. Currently
    * supports the following: `sm` & `md` (default: 'md')
    */
-  @property()
+  @property({ type: BREADCRUMB_SIZE, reflect: true, attribute: 'size' })
```
annawen1:
```suggestion
  @property({ type: BREADCRUMB_SIZE, reflect: true })
```

Also no need to set the `attribute` here since the name is the same

annawen1 (approved):

warrenmblood (approved):
LGTM

PR #19288, opened 2025-05-02
fix: modal footer buttons were not fully visible in zoom to 400

### LINKED ISSUE #19139: [a11y]: Modal: Footer buttons not fully in viewport at 400% zoom
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

v1.80.0

### React version

_No response_

n/a

### Assistive technology

_No response_

### Description

At 400% zoom, longer modal heading texts can wrap multiple lines and push the footer buttons outside of the viewport. In extreme cases, the buttons are pushed completely outside of the viewport.

### WCAG 2.1 Violation

1.4.10 Reflow

### Reproduction/example

https://react.carbondesignsystem.com/iframe.html?globals=&args=&id=components-modal--default&viewMode=story

### Steps to reproduce

1. Change zoom level to 400%.
2. Edit the modal heading text until it wraps twice or more.
3. See footer buttons are no longer fully in the viewport. In the example below, button labels are now hidden.

Example screenshot:
Image

### Suggested Severity

None

### Code of Conduct

preetibansalui:
Closes #19139 

Modal footer were not fully visible when zoom to 400%

#### Changelog

**New**

**Changed**

The modal uses `grid-template-rows: auto 1fr auto`. I set a `max-height: 50vh `on the header with scroll for overflow in case of large content. This keeps the body and footer visible even with large headers.

#### Testing / Reviewing
1. Go to Storybook > Modal > Test
2. Click on 'Open Canvas in New Tab' button to open story in a new page 
3. Zoom the screen to 400%
4. Footer buttons should be visible

Note: I will remove Test story before merge.


ariellalgilmore (commented):
LGTM, but im just not sure this something we need to solve.

The docs: https://carbondesignsystem.com/components/modal/usage/#title
`Use a brief verb phrase that describes the modal’s task or purpose.`. The Modal title's should be kept short, or 

> Sometimes, you might need to provide contextual information for the modal, such as when the actions performed using the modal apply to a specific item and you want to make it clear which item the actions apply to.
In this case, you can use the optional label that is displayed above the title.
For example, if the user needs to know the path to the object that they are editing, you might put the path in the optional label.

So maybe this is a non issue?

preetibansalui:
> LGTM, but im just not sure this something we need to solve.
> 
> The docs: https://carbondesignsystem.com/components/modal/usage/#title `Use a brief verb phrase that describes the modal’s task or purpose.`. The Modal title's should be kept short, or
> 
> > Sometimes, you might need to provide contextual information for the modal, such as when the actions performed using the modal apply to a specific item and you want to make it clear which item the actions apply to.
> > In this case, you can use the optional label that is displayed above the title.
> > For example, if the user needs to know the path to the object that they are editing, you might put the path in the optional label.
> 
> So maybe this is a non issue?

Yes, it should ideally be kept short, but we can't strictly enforce a maximum length. That's why I decided to set a max size, just in case someone enters a very long title.

ariellalgilmore (approved):

heloiselui (approved):
Looks good! Just need to remove the test story before merging.

Gururajj77 (approved):

PR #19292, opened 2025-05-05
feat: side nav button

### LINKED ISSUE #18562: [Feature Request]: Make it possible for SideNavLink and SideNavMenuItem components to render buttons
### The problem

At the moment, if you want to create a sidebar navigation, you would import the components and render menu items as `SideNavLink` or `SideNavMenu` with a nested `SideNavMenuItem`. The code in this case looks something like this:
```
<SideNavMenuItem href="https://www.carbondesignsystem.com/">
    This is a link and it renders well
</SideNavMenuItem>
```
 It works well, if those are links, meaning, if you pass it the `href` prop. In that the HTML is the following:
```
<a href="https://www.carbondesignsystem.com/" class="cds--side-nav__link">
    <span class="cds--side-nav__link-text">This is a link and it renders well</span>
</a>
```

However, not in all the cases the sidebar navigation menu item is a link. Sometimes it can be a button performing an action. For example, a button opening a modal with information like current version of the software or showing internal release notes. In that case you don't want to pass an `href` prop and you want to pass the `onClick` event handler instead. The component in that case looks like: 
```
<SideNavMenuItem
    onClick={() => {
        // a proper action would live here
        console.log('I should open a modal');
    }}
>
    This is a button that opens a modal. It renders as a link without the cursor pointer, since the href prop is missing
</SideNavMenuItem>
```

The HTML looks the following way in that case:
```
<a class="cds--side-nav__link"><span class="cds--side-nav__link-text">
    This is a button that opens a modal. It renders as a link without the cursor pointer, since the href prop is missing</span>
</a>
```

It has 2 problems:

1. It is semantically incorrect, since it is a button
2. It is missing a cursor pointer

Here is a StackBlitz showing the issue: https://stackblitz.com/edit/github-rd2s5bph?file=src%2FApp.jsx
I would also upload the video, but the limit of 10mb doesn't allow me to do that.

### The solution

1. `SideNavMenuItem` component with no `href` prop and `onClick` prop should render a button that has the same styling as the link
2. As far as I understand, `SideNavMenuItem` has to be nested within a `SideNavMenu` component, which you cannot always do. If you want a button to be on top level of a menu, you have to use the `SideNavLink` component. In that case `SideNavLink` should also render a button, just like in point 1.
3. All of these components should have the same styles as the link component would have

### Examples

_No response_

### Application/PAL

Instana

### Business priority

### Available extra resources

_No response_

### Code of Conduct

riddhybansal:
Closes #18562 

Creates SideNavButton component for the Carbon Design System

#### Changelog

**New**

- Added SideNavButton component that renders a button element with consistent SideNav styling
- Provides an alternative to SideNavLink when button semantics are needed

#### Testing / Reviewing

Verify the SideNavButton renders correctly within the SideNav component


riddhybansal:
SideNavButton ia created, but am not sure how to display this component in storybook! @laurenmrice can you verify if its fine on design perspective

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
+/**
+ * Copyright IBM Corp. 2025,
```
adamalston:
```suggestion
 * Copyright IBM Corp. 2025
```

packages/react/src/components/UIShell/__tests__/SideNavButton-test.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright IBM Corp. 2025,
```
adamalston:
```suggestion
 * Copyright IBM Corp. 2025
```

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+import PropTypes from 'prop-types';
+import React, {
+  ElementType,
+  ForwardedRef,
+  useContext,
+  Ref,
+  ComponentProps,
+} from 'react';
```
adamalston: - `ElementType` is unused.
- `Ref` is unused.
- `type` imports can be used.

```suggestion
import React, {
  useContext,
  type ComponentProps,
  type ForwardedRef,
} from 'react';
```

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+import Button from '../Button';
+import SideNavItem from './SideNavItem';
+import cx from 'classnames';
+import { SideNavContext } from './SideNav';
+
+export type SideNavButtonProps = ComponentProps<typeof Button> & {
+  /**
+   * Specify the children to be rendered inside of the `SideNavMenuItem`
```
adamalston, line 21: Why does this JSDoc comment reference `SideNavMenuItem`?

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  isSideNavExpanded?: boolean;
+
+  /**
+   * Optional prop to specify the tabIndex of the button. If undefined, it will be applied default validation
+   */
+  tabIndex?: number;
+};
+
+const SideNavButton = React.forwardRef<HTMLButtonElement, SideNavButtonProps>(
+  function SideNavButton(props, ref: ForwardedRef<HTMLButtonElement>) {
```
adamalston: - `ref` is already typed.
- An anonymous function can be used. `SideNavButton` is never gonna be called.

```suggestion
  (props, ref) => {
```

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const {
+      children,
+      className: customClassName,
+      isSideNavExpanded,
+      onClick,
+      tabIndex,
+      ...rest
+    } = props;
+    const className = cx({
+      [customClassName as string]: !!customClassName,
```
adamalston, line 54: The `string` type assertion shouldn't be necessary if the component is typed correctly. Like when I delete the type assertion here and `ComponentProps<typeof Button> &` above, a type error shows up. That type error is valid.

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+    return (
+      <SideNavItem>
+        <Button className={className} {...rest} onClick={onClick} ref={ref}>
+          {children}
+          tabIndex=
+          {tabIndex === undefined
```
adamalston: https://stackoverflow.com/a/2703122/11106650

```suggestion
          {typeof tabIndex === 'undefined'
```

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+      ...rest
+    } = props;
+    const className = cx({
+      [customClassName as string]: !!customClassName,
+    });
+    return (
+      <SideNavItem>
+        <Button className={className} {...rest} onClick={onClick} ref={ref}>
+          {children}
+          tabIndex=
```
adamalston, line 60: Is this text supposed to be here?

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+              ? -1
+              : 0
+            : tabIndex}
+        </Button>
+      </SideNavItem>
+    );
+  }
+);
+
+SideNavButton.displayName = 'SideNavButton';
```
adamalston, line 72: What's this `displayName` needed for?

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Provide an optional class to be applied to the containing node
+   */
+  className: PropTypes.string,
+
+  /**
+   * Provide an optional function to be called when the button element
+   * is clicked
+   */
+  onClick: PropTypes.func,
```
adamalston, line 88: This property isn't included in `SideNavButtonProps`. Is it supposed to be?

packages/react/src/components/UIShell/SideNavButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+    return (
+      <SideNavItem>
+        <Button className={className} {...rest} onClick={onClick} ref={ref}>
+          {children}
+          tabIndex=
+          {tabIndex === undefined
+            ? !isSideNavExpanded && !isRail
+              ? -1
+              : 0
+            : tabIndex}
```
adamalston: Is all of this `tabIndex` functionality supposed to be in a prop on the `Button` component? Further, is the current implementation accessible?

packages/react/src/components/UIShell/__tests__/SideNavButton-test.js
```diff
@@ -0,0 +1,39 @@
...
+  it('should call the `onClick` handler when clicked', async () => {
+    const onClick = jest.fn();
+    const user = userEvent.setup();
+
+    render(<SideNavButton onClick={onClick}>test</SideNavButton>);
+
+    await user.click(screen.getByRole('button'));
+    expect(onClick).toHaveBeenCalledTimes(1);
+  });
+});
```
adamalston, line 39: Should testing be added for these things:

1. `tabIndex` and other accessibility functionlity.
2. The button's children, both static (e.g., text) and dynamic (e.g., a link).
3. `isRail` functionality.
4. Visual testing since one of the key points in https://github.com/carbon-design-system/carbon/issues/18562#issuecomment-2653961793 was broken styles.

adamalston (commented):
Should `SideNavButton` be exported in packages/react/src/components/UIShell/index.ts?

thyhmdo:
I’d suggest we hold off on this for now. There’s no issue with having a designer involved to make sure the design specs and considerations are aligned. We’d like some time to talk through the use case with the design team and make sure any decisions are backed by research before we bring this into the system.

alisonjoseph (changes requested):
@riddhybansal @thyhmdo I believe the intent of the original issue was to match the styling of the SideNavLink, the only difference being having the underlying html element is a button, so that they can add onClick events instead of hrefs.

riddhybansal:
> @riddhybansal @thyhmdo I believe the intent of the original issue was to match the styling of the SideNavLink, the only difference being having the underlying html element is a button, so that they can add onClick events instead of hrefs.

So may be we dont need to add story for it just the component ?

alisonjoseph:
> So may be we dont need to add story for it just the component ?
We still would want to show it in one of the stories, but rihgt now you have it styled like a Carbon button, it should look like the SideNavLink

alisonjoseph:
After speaking with @thyhmdo we're going to close this until we can do some more research on the design side about this interaction.

PR #19308, opened 2025-05-05
fix: tooltip and icon-button respect child aria attributes

### LINKED ISSUE #18565: [a11y]: IconButton does not always render aria-labelledby and aria-describedby correctly
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.70.0

### React version

17.0.2

### Assistive technology

None

### Description

When using the IconButton component, it always adds `aria-labelledby=IdOfTooltip`, if the label is present and `aria-describedby=IdOfTooltip`, if the label is not present. The code lives here: `packages/react/src/components/Tooltip/Tooltip.tsx` and is the following:

```
  if (label) {
    triggerProps['aria-labelledby'] = id;
  } else {
    triggerProps['aria-describedby'] = id;
  }
```

The issue is that sometimes it doesn't work properly. There are some cases when you want to pass the `aria-label` manually and `label` would serve more as a description. There are cases when you manually want to pass `aria-describedby` with the label present as well. In all these cases `aria-labelledby` is added to the element in HTML, even though it should be `aria-describedby`.

The solution seems to be the following:
1. When no accessibility props (no `aria-label`, no `aria-describedby`) are passed, follow the existing logic.
2. When `aria-describedby` is passed manually, render the one that user passes instead of ID.
4. When `aria-label` is passed, never render `aria-labelledby` and check, if `aria-describedby` is passed manually. If it is, render it, if it isn't, assign it manually.

There may be some more cases that I didn't think about, but at the moment it seems to overlook the cases mentioned above.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://stackblitz.com/edit/github-qqpwtach?file=src%2FApp.tsx

### Steps to reproduce

1. Use `IconButton` component
2. Manually pass it the a`aria-label` prop
3. Observe that `aria-labelledby` is rendered in HTML instead of `aria-describedby`

### Suggested Severity

### Code of Conduct

heloiselui:
Closes #18565

Fixes an accessibility issue where `Tooltip` and `IconButton` were always adding `aria-labelledby` or `aria-describedby`, even when these attributes were already set on the button. This could cause problems for screen readers.

Now, `Tooltip` only adds ARIA attributes if they’re not already present. `IconButton` also respects a custom `aria-describedby` passed via props.

#### Changelog

**New**

- Support for custom `aria-describedby` on `IconButton`.

**Changed**

- Tooltip now respects existing ARIA attributes on its child (`aria-label`, `aria-labelledby`, `aria-describedby`).
- Added unit tests for ARIA behavior.

#### Testing / Reviewing

- CI should pass

You can also test manually in Storybook:

Go to `React Deploy Preview` > `IconButton` > `Default`
Use DevTools to inspect the button and check the ARIA attributes in these cases:
1. **Button has `aria-label`**
   - Tooltip should **not** add `aria-labelledby`.
2. **Button has `aria-labelledby`**
   - Tooltip should **not** overwrite it.
3. **Button has no ARIA props**
   - Tooltip should set `aria-labelledby` with its own ID.
4. **Button has both `aria-label` and `aria-describedby`**
   - Tooltip should **not** overwrite `aria-describedby`.
5. **Button has no ARIA props, and tooltip uses only `description`**
   - Tooltip should set `aria-describedby` with its own ID.

> [!NOTE] 
> All these cases are covered by unit tests.


packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -161,10 +164,38 @@ const Tooltip: TooltipComponent = React.forwardRef(
       return eventHandlers;
     }
 
-    if (label) {
-      triggerProps['aria-labelledby'] = id;
-    } else {
-      triggerProps['aria-describedby'] = id;
+    const hasAriaLabel = child?.props?.['aria-label'];
```
adamalston: Suggestion to avoid the reassignments and to abstract some of the common logic.

```diff
diff --git a/packages/react/src/components/Tooltip/Tooltip.tsx b/packages/react/src/components/Tooltip/Tooltip.tsx
index 0125047b6..a3c409c85 100644
--- a/packages/react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react/src/components/Tooltip/Tooltip.tsx
@@ -133,6 +133,24 @@ const Tooltip: TooltipComponent = React.forwardRef(
     const prefix = usePrefix();
     const child = React.Children.only(children);
 
+    const {
+      'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledBy,
+      'aria-describedby': ariaDescribedBy,
+    } = child?.props ?? {};
+
+    const hasLabel = !!label;
+    const hasAriaLabel = typeof ariaLabel !== 'undefined' && ariaLabel !== null;
+
+    const labelledBy = hasAriaLabel
+      ? null
+      : hasLabel
+        ? (ariaLabelledBy ?? id)
+        : undefined;
+    const describedBy = hasAriaLabel
+      ? ariaDescribedBy
+      : (ariaDescribedBy ?? (!hasLabel && !ariaLabelledBy ? id : undefined));
+
     const triggerProps = {
       onFocus: () => !focusByMouse && setOpen(true),
       onBlur: () => {
@@ -146,6 +164,8 @@ const Tooltip: TooltipComponent = React.forwardRef(
       onMouseDown,
       onMouseMove: onMouseMove,
       onTouchStart: onDragStart,
+      'aria-labelledby': labelledBy,
+      'aria-describedby': describedBy,
     };
 
     function getChildEventHandlers(childProps: any) {
@@ -164,40 +184,6 @@ const Tooltip: TooltipComponent = React.forwardRef(
       return eventHandlers;
     }
 
-    const hasAriaLabel = child?.props?.['aria-label'];
-    const hasAriaLabelledBy = child?.props?.['aria-labelledby'];
-    const hasAriaDescribedBy = child?.props?.['aria-describedby'];
-    const hasLabel = !!label;
-
-    let labelledBy: string | null | undefined;
-    let describedBy: string | undefined;
-
-    // aria-labelledby logic
-    if (hasAriaLabel) {
-      labelledBy = null; // ignore labelledBy when aria-label is present
-    } else if (hasLabel && hasAriaLabelledBy) {
-      labelledBy = hasAriaLabelledBy;
-    } else if (hasLabel) {
-      labelledBy = id;
-    }
-
-    // aria-describedby logic
-    if (hasAriaLabel) {
-      describedBy = hasAriaDescribedBy; // always show if explicitly passed
-    } else if (hasAriaDescribedBy) {
-      describedBy = hasAriaDescribedBy;
-    } else if (!hasLabel && !hasAriaLabelledBy) {
-      describedBy = id;
-    }
-
-    // apply to triggerProps
-    if (labelledBy !== undefined) {
-      triggerProps['aria-labelledby'] = labelledBy;
-    }
-    if (describedBy !== undefined) {
-      triggerProps['aria-describedby'] = describedBy;
-    }
-
     const onKeyDown = useCallback(
       (event: KeyboardEvent) => {
         if (open && match(event, keys.Escape)) {
```

tay1orjones: "Nested" ternaries make me wince with how debatable their formatting and readability are https://github.com/prettier/prettier/issues/9561

I think it's a nice streamlining of the logic though since there's only single assignments happening within the `if` statements.

My only suggestion would be a code comment to help smooth out understanding.

```diff
+ // An aria-label takes precedence over aria-labelledby, but when it's
+ // needed and the user doesn't specify one, the fallback `id` is used.
  const labelledBy = hasAriaLabel
        ? null
        : hasLabel
          ? (ariaLabelledBy ?? id)
          : undefined;
```

Something similar for `describedBy`, too.

heloiselui: @adamalston  and @tay1orjones, thanks for the feedback and suggestions! I’ve applied the recommended changes 🎉

adamalston: > "Nested" ternaries make me wince with how debatable their formatting and readability are [prettier/prettier#9561](https://github.com/prettier/prettier/issues/9561)

That's fair, sorry about that.

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -130,6 +133,29 @@ const Tooltip: TooltipComponent = React.forwardRef(
...
     const child = React.Children.only(children);
 
+    const {
+      'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledBy,
+      'aria-describedby': ariaDescribedBy,
+    } = child?.props ?? {};
+
+    const hasLabel = !!label;
+    const hasAriaLabel = typeof ariaLabel !== 'undefined' && ariaLabel !== null;
```
2nikhiltom: Overall this PR looks soild

```suggestion
  
const hasAriaLabel = typeof ariaLabel === 'string' ? ariaLabel.trim() !== '' : false;
```
This would prevent empty or whitespace-only aria-labels from being treated as valid but would be ignored by screen readers

2nikhiltom (approved):

packages/react/src/components/Tooltip/Tooltip.tsx
heloiselui: Thanks for check that!

tay1orjones (approved):

adamalston:
Not sure if the following change is expected. Prior to this pull request, `aria-labelledby` attributes were added to `HeaderGlobalAction` components. These attributes connected the underlying `IconButton` to the `Popover` via an `id`. Now, `aria-labelledby` is no longer being added.

For example, before this pull request, the element with `aria-label="Search"` in https://react.carbondesignsystem.com/?path=/story/components-ui-shell-header--header-w-actions-and-right-panel also had an `aria-labelledby` attribute. That's no longer the case.

Is this change expected? The only reason I ask is that typically, there are `id` references connecting `Popover` elements to their corresponding triggers. Now, the `Popover` that appears when hovering over the **Search** action in the header has an `id` that doesn't correspond to any other element.

I can open an issue if there's more investigation that needs to take place. I thought it would be good to ask here first though.

PR #19319, opened 2025-05-06
fix(Copy): set leaveDelayMs during animation feedback

### LINKED ISSUE #16811: [Bug]: feedbackTimeout doesn't work
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.59.0

### React version

v18.3.1

### Description

Hi I am using the <CopyButton/> with feedbackTimeout={2000} which indeed is a default value. However, the feedback text doesn't goes away under the copy button on the UI. See below. Please note that the feedback is not shown on the tootip as in your playground. But it only is written down underneath the copy button. Which is okay, but it should go away after the feedback timeout. 
Screenshot 2024-06-18 at 12 14 42 p m

### Reproduction/example

Please see steps below.

### Steps to reproduce

import copy from "copy-to-clipboard";
import { CopyButton } from "@carbon/react";

                  <Col>
                    <CopyButton
                      feedback="Copied!"
                      align="top-right"
                      feedbackTimeout={2000}
                      iconDescription=""
                      onClick={() => copy(configMap)}
                    />
                  </Col>

### Suggested Severity

### Application/PAL

IBM's internal app.

### Code of Conduct

emyarod:
Closes #16811

This PR adds a condition for the Copy component feedback timeout delay on mouse leave so that the feedback tooltip does not disappear on mouseleave while an animation is active

#### Changelog

**New**

- `leaveDelayMs` on the `IconButton` in `Copy`

#### Testing / Reviewing

click the Copy/CopyButton component to activate the feedback tooltip. move the mouse on and off the element and ensure the feedback timeout is respected


annawen1 (approved):

packages/react/src/components/Copy/Copy-test.js
```diff
@@ -146,4 +146,21 @@ describe('Feedback', () => {
...
+        <CopyIcon />
+      </Copy>
+    );
+
+    const button = screen.getByTestId('copy-button-8');
+    expect(screen.getAllByText('Copy to clipboard').length).toBe(1);
+    await user.click(button);
+    fireEvent.mouseLeave(button);
+    expect(screen.getAllByText('Copied!').length).toBe(1);
+  });
```
adamalston, line 165: If I revert the code change, this test still passes. Should it be failing?

```console
git checkout emyarod-16811-fix/feedback-mouseleave-persist
git checkout main packages/react/src/components/Copy/Copy.tsx
yarn test packages/react/src/components/Copy/Copy-test.js
```

adamalston (commented):
When I revert the code change, the test still passes. Should it be failing?

```console
git checkout emyarod-16811-fix/feedback-mouseleave-persist
git checkout main packages/react/src/components/Copy/Copy.tsx
yarn test packages/react/src/components/Copy/Copy-test.js
```

2nikhiltom (approved):
I don't want to be a blocker here, the fix LGTM.  In test case I think we could have `jest.advanceTimersByTime()` calls that verify the feedback message remains visible throughout the timeout period.

emyarod:
> I don't want to be a blocker here, the fix LGTM. In test case I think we could have `jest.advanceTimersByTime()` calls that verify the feedback message remains visible throughout the timeout period.

I have been stuck on how to properly test the original regression so thank you for the suggestion, I will look into it!

PR #19346, opened 2025-05-08
chore(release): v11.82.1

annawen1:
Release PR for v11.82.1, which bumps up the version of `@carbon/utilities` package and the versions of the packages depending on it as well


annawen1:
This was a manual release and that the PR doesn’t need to be merged

PR #19351, opened 2025-05-08
fix(use-resize-observer): remove dependency with local hook

### LINKED ISSUE #19350: [Bug]: `use-resize-observer` does not support React 19
### Package

@carbon/react

### Browser

Chrome

### Package version

1.81.0

### React version

19

### Description

Reported via @ocornec, teams within Cloud noticed build failures after including his AI chat library components from `@carbon/labs`. However, they noticed build failures because of version compatibility issues with `use-resize-observer` since they are using React 19. Even though these are just warnings, it is causing their builds to fail.

`@carbon/ibm-products` has a `useResizeObserver` hook that we use internally, I can add this internally as well to `@carbon/react` and remove `use-resize-observer` as a dependency in `@carbon/react`. This overall sounds like a win-win as it will resolve the warnings/build failures for teams and remove a potentially extraneous dependency.

### Reproduction/example

https://stackblitz.com/edit/q1yb8s5s?file=package.json

### Steps to reproduce

Run `npm install` from stackblitz provided (thanks @ariellalgilmore for putting the stackblitz together!)

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

matthewgallo:
Closes #19350 

This PR removes the `use-resize-observer` package from `@carbon/react`. This package is causing peer dependency warnings for teams using React 19, and in some cases those warnings will cause builds to fail. As an alternative approach, I included a local/internal hook (`useResizeObserver`) that is used in `@carbon/ibm-products`. I verified the behavior remains the same in the `CodeSnippet` multiline stories.

### Changelog

**New**

- `packages/react/src/internal/__tests__/useResizeObserver-test.js`
- `packages/react/src/internal/useResizeObserver.ts`
- `packages/react/src/internal/__mocks__/mockHTMLElement.ts`

**Changed**

- `packages/react/src/components/CodeSnippet/CodeSnippet.tsx`
- `packages/react/package.json`
- `yarn.lock`

#### Testing / Reviewing

Added a unit test that covers the basic usage of the hook.

## PR Checklist

- [x] Updated documentation and storybook examples (N/A, couldn't find any other docs on the previous resize observer implementation)


tay1orjones (commented):
Any idea how is CodeSnippet still working without updating the import?

https://github.com/carbon-design-system/carbon/blob/97343be5a105ba1f28d01c7af5805e7174262847/packages/react/src/components/CodeSnippet/CodeSnippet.tsx#L17

Kind of wild the build didn't fail here.

matthewgallo:
That is a great question 😄 I don't know if I accidentally reverted that import or what?

kennylam (approved):

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,97 @@
...
+    cb.current = onResize;
+  }, [onResize]);
+
+  useEffect(() => {
+    const getInitialSize = () => {
+      if (ref.current) {
+        const refComputedStyle = window.getComputedStyle(ref.current);
+        const initialWidth =
+          (ref.current?.offsetWidth || 0) -
+          (typeof refComputedStyle?.paddingLeft === 'string' &&
```
tay1orjones, line 35: I wondered here about logical property support but it looks like this is taken care of by the [spec](https://www.w3.org/TR/css-logical-1/#:~:text=APIs%20that%20return%20computed%20values%20(such%20as%20getComputedStyle())%20must%20return%20the%20same%20value%20for%20equivalent%20pairs%20of%20such%20properties.)

> APIs that return computed values (such as `getComputedStyle()`) must return the same value for equivalent pairs of such properties

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,97 @@
+/**
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { useRef, useState, useLayoutEffect, useEffect } from 'react';
+
+export const useResizeObserver = ({ ref, onResize, deps = [] }) => {
```
tay1orjones:
```suggestion
export const useResizeObserver = ({ ref, onResize }) => {
```

If we don't have a need for this, I'd leave it out. Makes it really easy for someone to accidentally have these useEffects re-run on every render. If there's a solid usecase for it though I think it'd make sense to write a test for it. Sort of "documenting" that use case via the test, if that makes sense. Non blocking, just my 2c

tay1orjones (approved):

packages/react/src/internal/__mocks__/mockHTMLElement.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const hep = HTMLElement.prototype;
+export const mockHTMLElement = (options) => {
```
adamalston: Can you add types?

packages/react/src/internal/__tests__/useResizeObserver-test.js
```diff
@@ -0,0 +1,93 @@
...
+import { mockHTMLElement } from '../__mocks__/mockHTMLElement';
+
+const sizes = (base) => ({
+  offsetWidth: {
+    ['observed-el']: base,
+  },
+  offsetHeight: { ['observed-el']: base / 2 },
+});
+
+const testSizes = (el, property) => {
```
adamalston: Can you add types for the function parameters?

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,97 @@
...
+  useEffect(() => {
+    const getInitialSize = () => {
+      if (ref.current) {
+        const refComputedStyle = window.getComputedStyle(ref.current);
+        const initialWidth =
+          (ref.current?.offsetWidth || 0) -
+          (typeof refComputedStyle?.paddingLeft === 'string' &&
+          refComputedStyle?.paddingLeft.length
+            ? parseFloat(refComputedStyle?.paddingLeft)
+            : 0,
```
adamalston: Is the comma intentional? With the comma operator, only the last value is used in the expression, so the result would ignore the left padding entirely and only subtract the right padding. The two blocks of code below are effectively equivalent. If the comma is intentional, I would think that you should be able to simplify your implementation.

```ts
const initialWidth =
  (ref.current?.offsetWidth || 0) -
  (typeof refComputedStyle?.paddingLeft === 'string' &&
  refComputedStyle?.paddingLeft.length
    ? parseFloat(refComputedStyle?.paddingLeft)
    : 0,
  typeof refComputedStyle?.paddingRight === 'string' &&
  refComputedStyle?.paddingRight.length
    ? parseFloat(refComputedStyle?.paddingRight)
    : 0);
```

```ts
const initialWidth =
  (ref.current?.offsetWidth || 0) -
  (typeof refComputedStyle?.paddingRight === 'string' &&
  refComputedStyle?.paddingRight.length
    ? parseFloat(refComputedStyle?.paddingRight)
    : 0);
```

I also have this question for the `initialHeight` variable.

---

The code below demonstrates that the first part of the expression in `complex` is ignored. Only the last part is used, which is exactly what the comma operator does.

```ts
const something = true;
const somethingElse = '20';

const complex =
  100 -
  (something && somethingElse ? parseFloat(somethingElse) : 0,
  somethingElse && somethingElse ? parseFloat(somethingElse) : 0);
const simple =
  100 - (somethingElse && somethingElse ? parseFloat(somethingElse) : 0);

console.log({ complex, simple });
```

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,97 @@
...
+        setWidth(initialWidth);
+        setHeight(initialHeight);
+      }
+    };
+    if (!ref?.current || (width >= 0 && height >= 0)) {
+      return;
+    }
+    getInitialSize();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [width, height, ref.current, ...deps]);
```
adamalston: Any reason why `ref.current` is in the dependency array? I'm pretty sure that's an anti-pattern.

https://stackoverflow.com/q/60476155/11106650
https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (!ref?.current) {
+      return;
+    }
+
+    const doCallbacks = () => {
+      if (!ref?.current || !Array.isArray(entriesToHandle?.current)) {
+        return;
+      }
+
+      const entry = entriesToHandle.current[0];
```
adamalston, line 75: What if multiple elements are being observed? Or is that not a supported use case?

packages/react/src/internal/useResizeObserver.ts
matthewgallo: Removed

packages/react/src/internal/useResizeObserver.ts
matthewgallo: Thanks for pointing this out! I think this was a typo while migrating this to TS.

packages/react/src/internal/useResizeObserver.ts
matthewgallo: As it currently stands, this hook only supports observing a single element

packages/react/src/internal/useResizeObserver.ts
matthewgallo: Agreed, I removed it since there currently isn't a need.

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    // observe all refs passed
+    observer.observe(ref.current);
+
+    return () => {
+      observer?.disconnect();
+      observer = null;
+    };
+  }, [ref.current]);
```
adamalston: https://github.com/carbon-design-system/carbon/pull/19351#discussion_r2093728105

packages/react/src/internal/__mocks__/mockHTMLElement.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const hep = HTMLElement.prototype;
+export const mockHTMLElement = (options: HTMLElement) => {
+  const originals = {};
```
adamalston: May be worth typing `originals`.

```suggestion
  const originals: Record<string, PropertyDescriptor | undefined> = {};
```

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        const initialHeight =
+          (ref.current?.offsetHeight || 0) -
+          (typeof refComputedStyle?.paddingTop === 'string' &&
+          refComputedStyle?.paddingTop.length
+            ? parseFloat(refComputedStyle?.paddingTop)
+            : 0) -
+          (typeof refComputedStyle?.paddingLeft === 'string' &&
+          refComputedStyle?.paddingLeft.length
+            ? parseFloat(refComputedStyle?.paddingLeft)
```
adamalston: Should `paddingLeft` be `paddingBottom` instead? A height is being calculated. If it should be `paddingBottom`, should a test be failing given it's not?

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { useRef, useState, useLayoutEffect, useEffect } from 'react';
+
+export const useResizeObserver = ({ ref, onResize }) => {
```
adamalston: Can you type the options?

```suggestion
export const useResizeObserver = ({
  ref,
  onResize,
}: {
  ref: RefObject<HTMLElement>;
  onResize?: (rect: DOMRectReadOnly) => void;
}) => {
```

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,96 @@
...
+        doCallbacks();
+      });
+    });
+
+    // observe all refs passed
+    observer.observe(ref.current);
+
+    return () => {
+      observer?.disconnect();
+      observer = null;
```
adamalston: Why does the observer need to be nullified if it's disconnected?

matthewgallo: Good point, don't think it needs to be. I've updated this.

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      const entry = entriesToHandle.current[0];
+
+      setWidth(entry.contentRect.width);
+      setHeight(entry.contentRect.height);
+
+      cb.current && cb.current(entry.contentRect);
+    };
+
+    const observer: ResizeObserver = new ResizeObserver((entries) => {
```
adamalston: The type is inferred so it can be omitted.

```suggestion
    const observer = new ResizeObserver((entries) => {
```

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,101 @@
...
+        // do callbacks
+        doCallbacks();
+      });
+    });
+
+    // observe all refs passed
+    observer.observe(ref.current);
+
+    return () => {
+      observer?.disconnect();
```
adamalston: Suggestion to delete the optional chaining operator given https://github.com/carbon-design-system/carbon/pull/19351#discussion_r2096690981.

```suggestion
      observer.disconnect();
```

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,101 @@
...
+    cb.current = onResize;
+  }, [onResize]);
+
+  useEffect(() => {
+    const getInitialSize = () => {
+      if (ref.current) {
+        const refComputedStyle = window.getComputedStyle(ref.current);
+
+        const initialWidth =
+          (ref.current?.offsetWidth || 0) -
```
adamalston: Suggestion to use a nullish coalescing operator here and in `initialHeight` to avoid `0 || 0`.

```suggestion
          (ref.current?.offsetWidth ?? 0) -
```

packages/react/src/internal/useResizeObserver.ts
```diff
@@ -0,0 +1,101 @@
...
+  const [width, setWidth] = useState(-1);
+  const [height, setHeight] = useState(-1);
+  const entriesToHandle = useRef<ResizeObserverEntry[] | null>(null);
+  const cb = useRef(onResize);
+
+  useEffect(() => {
+    // ref for onResize removes it as dependency from useLayoutEffect
+    // This significantly reduces repeated calls if a function is redefined on every
+    // render
+    cb.current = onResize;
```
adamalston, line 25: A future consideration if performance becomes an issue may be to debounce the `onResize` callback.

adamalston (commented):
Just a few more cleanup suggestions. Thanks for adding types. They should guide future changes and prevent regressions.

packages/react/src/internal/useResizeObserver.ts
matthewgallo: Agreed, happy to add that as an enhancement after this initial PR.

Issue #19023, opened 2025-04-03
[a11y]: [Windows High Contrast Mode]: [Tooltip] The icon is barely visible in High Contrast Mode and g10 theme


xrissot-ibm:
### Package

@carbon/react

### Browser

Chrome

### Operating System

Windows

### Package version

1.79

### Automated testing tool and ruleset

Manual testing for Review recommandations section

### Description

Expect that the icon is visible in high contrast mode

### Reproduction/example

use carbon-react playground

### Steps to reproduce

Open tooltip playground and select g10 theme: https://react.carbondesignsystem.com/?path=/story/components-tooltip--default&globals=theme:g10

Activate Windows HCM,
The tooltip icon is barely visible:
![Image](https://github.com/user-attachments/assets/9e7605a8-5fe4-4574-9bd8-ae0d8f1c8807)

### Suggested Severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

heloiselui:
This was fixed in PR https://github.com/carbon-design-system/carbon/pull/19352

Issue #19015, opened 2025-04-03
[a11y]:  [Windows High Contrast Mode]: CopyButton, Close Button in InlineNotification with 'lowContrast=true' and IconButton with kind='ghost' are not visible in windows 10 High-Contrast Mode


SushanthTanariIBM:
### Package

@carbon/react

### Browser

Chrome

### Operating System

Windows

### Package version

1.79.0

### React version

18.0.0

### Automated testing tool and ruleset

Manual Verification

### Assistive technology

NVDA

### Description

"CopyButton", "Close Button in InlineNotification with lowContrast=true" and "IconButton with kind='ghost'" are not visible in windows 10 High-Contrast Mode

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-iconbutton--default&args=kind:ghost

### Steps to reproduce

Urls to @carbon/react storybook -

https://react.carbondesignsystem.com/?path=/docs/components-copybutton--overview&args=kind:ghost
https://react.carbondesignsystem.com/?path=/story/components-iconbutton--default&args=kind:ghost
https://react.carbondesignsystem.com/?path=/story/components-notifications-inline--default&args=lowContrast:!true

IconButton:

![Image](https://github.com/user-attachments/assets/4d088a81-d7ca-48ed-9701-d54f2cfede1a)

CopyButton:

![Image](https://github.com/user-attachments/assets/4cc7f1ce-dcec-47e7-b917-794691e4dc59)

"Close Button in InlineNotification with lowContrast=true" :
![Image](https://github.com/user-attachments/assets/5a59f53c-34b0-4752-a80b-fec358c8644f)

### Suggested Severity

Severity 3 = User can complete task, and/or has a workaround within the user experience of a given component.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

wkeese:
I'm looking into this.

I reproduce this (or at least some of it) on Windows 11 too, using the "Aquatic" theme.  Note that you can work around it by setting your Carbon theme to G100.  

Basically, it's a CSS precedence problem.   When you are in a high-contrast theme, you need to set your `fill` to a system color.  And Carbon does that twice for the Copy icon via:

```css
[...]

Issue #18830, opened 2025-03-14
[a11y]: [Windows High Contrast Mode]: [OverflowMenu] The icon is not visible in high contrast mode with g10 theme


xrissot-ibm:
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.78

### Description

OverflowMenu icon is not visible when activating Windows High Contrast Mode with Gray10 theme

### Reproduction/example

use carbon-react playground

### Steps to reproduce

Open OverflowMenu playground https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--default&globals=theme:g10
Activate Windows HCM, the icon is not visible.

![Image](https://github.com/user-attachments/assets/30f2ab4d-4569-4d7f-a6d1-8961ff51c51a)

### Suggested Severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

heloiselui:
This was fixed in PR https://github.com/carbon-design-system/carbon/pull/19352

PR #8489, opened 2021-04-22
feat(HCM): add HCM mixin, use `forced-colors` 

tw15egan:
Closes https://github.com/carbon-design-system/carbon/issues/8413
Closes https://github.com/carbon-design-system/carbon/issues/8458

Updates the HCM media queries around the codebase to use a mixin, and moves the HCM to the bottom of the `.scss` files to ensure they are not overwritten.

Also uses `forced-colors` which is fully supported in Chrome 🎉 

#### Changelog

**New**

- `@mixin high-contrast-mode` which has a few built-in styles to handle various HCM fixes. This mixin makes it easier to update the underlying media query used for HCM. 
- added `forced-colors` support, which is supported in Chrome, Edge, Safari (opt-in), and Firefox (opt-in).

**Changed**

- Moved HCM queries to the end of the file so that other styles do not override them

**Removed**

- Media queries throughout the repo for HCM have been removed, and replaced with the HCM mixin at the bottom of `.scss` files 

#### Testing / Reviewing

In Windows, enable High Contrast Mode, and open up Firefox, Chrome, and Edge. In Firefox, you will need to type `about:config` in the address bar, then search for `forced-colors` and set that to `true`. Then, ensure each component is usable when in HCM and that all icons are displayed.

PR #19352, opened 2025-05-08
fix(icons): fix icons invisible in high contrast mode

### LINKED ISSUE #17147: [a11y]: [Windows High Contrast Mode]: [IconButton] The icon is not visible in high contrast mode
### Package

@carbon/react

### Browser

Chrome

### Operating System

Windows

### Package version

1.57.0

### React version

17.0.1

Manual Verification

### Assistive technology

_No response_

### Description

Manual verification of the above requirement brought up this issue. 
Ref: 
Requirement - [1.4.5 Images of Text](https://www.ibm.com/able/requirements/requirements/?version=7_2#1_4_5)
Manual Verification - [Check use of color, contrast, and images of text](https://www.ibm.com/able/toolkit/verify/manual/#color)

### WCAG 2.1 Violation

1.4.5 Images of Text

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-iconbutton--playground&args=kind:ghost

### Steps to reproduce

Go to Carbon storybook URL - https://react.carbondesignsystem.com/?path=/story/components-iconbutton--playground&args=kind:ghost

image

You can see that the icon is not visible. 

### Suggested Severity

### Code of Conduct

### LINKED ISSUE #19140: [a11y]: [Windows High Contrast Mode]: [ContentSwitcher] Icons are not visible and selected item has no borders
### Package

@carbon/react

### Browser

Chrome

### Operating System

Windows

### Package version

v1.79

### React version

_No response_

IBM Accessibility tool - needs review section 

### Assistive technology

_No response_

### Description

Icon should be visible, borders also and way to determine which entry is the selected one

### WCAG 2.1 Violation

Verify High contrat mode on windows

### Reproduction/example

Use the playground for the component and activate HCM

### Steps to reproduce

Use the playground for the component and activate HCM.

Selected item icon is not visible (in Icons only mode) + borders and background for the selected value;
Image

Image

### Suggested Severity

None

### Code of Conduct

wkeese:
Closes #17147, #18830, #19015, #19023, and partially fixes #19140.
Also fixes problems like showing the selected tab in UIShell / SidePanel.
Likely also fixes similar problems in @carbon/ibm-products.
Refs #8489.

If the system has a dark high-contrast theme, then icons need to be displayed in white (or a light color), and vice-versa.  Otherwise, icons become invisible (black on black or white on white). Importantly, the icon color needs to be set relative to the system's high contrast theme _regardless of which Carbon theme is selected_.

Although Carbon already has a mixin for this, it is not used consistently, and it would be labor-intensive, bloat the code, and error-prone to do so.   Because you basically would need to call the mixin every time you set a `fill`, which is done hundreds of times.   For example:

```scss
.#{$prefix}--accordion__arrow,
.#{$prefix}--accordion__item--active .#{$prefix}--accordion__arrow {
  fill: $icon-primary;
  @include high-contrast-mode('icon-fill');
}
```

And as I wrote in #19015, you also need to worry about CSS precedence, so you might think you've fixed a bug, but it isn't fixed in all cases.

So instead, this PR goes to the root of the issue by just overriding the value of `--cds-icon-primary` CSS property referenced by the `$icon-primary` SCSS  variable.   Likewise for similar CSS properties.   For example, consider [this story](https://react.carbondesignsystem.com/?path=/story/components-iconbutton--default&args=kind:ghost&globals=theme:g100), in the G100 theme.  There's this existing CSS:

```css
[data-carbon-theme=g100] {
  --cds-icon-primary: #f4f4f4;
  --cds-icon-secondary: ...
  ...
}
```

But in high contrast mode, it's now overridden by:

```css
@media screen and (-ms-high-contrast: active), (forced-colors: active) {
    [data-carbon-theme=g100] {
        --cds-icon-primary: ButtonText;
        --cds-icon-secondary: ButtonText;
        --cds-icon-disabled: GrayText;
        --cds-icon-on-color-disabled: GrayText;
        --cds-icon-inverse: SelectedItemText;
        --cds-icon-on-color: SelectedItemText;
        ...
    }
}
```

I also removed the current 27 calls to:

```scss
@include high-contrast-mode('icon-fill');
```

Plus a few other places that were manually setting `fill` to `ButtonText`, but now no longer need to do so.

### Changelog

**New**

- Added CSS to override the values of icon CSS properties (ex: `--cds-icon-primary`) when the system is in high contrast mode.

#### Testing / Reviewing

Tested in Storybook on Windows 11, using the Aquatic and Desert themes.

Note that you can emulate forced-mode on Mac in Chrome (see https://cssence.com/2024/forced-colors-mode-strategies/), which is useful if you don't have a PC handy.   Just open Chrome DevTools, hit Ctrl/Cmd+Shift+P, type “forced”, and you’ll find the option to "Emulate CSS forced-colors: active".

## PR Checklist

- [N/A] Wrote passing tests that cover this change


heloiselui (commented):
Great, thank you so much for working on this!

I've been reviewing all the issues and your updates:

For `@include high-contrast-mode('icon-fill') `- try removing it in this PR if possible, so I can review it here.

I tested #17147, #18830, #19015, and #19023 - they all look good, thanks 🎉

I tested #19140 and it still shows the same behavior as the issue. I recorded a quick video so you can take a look. I also saw that @mariyageorge01 made some changes in her PR, I tested it and it works as expected. Maybe we can wait for your PR to be merged first, and then she can continue with her updates? What do you think?

https://github.com/user-attachments/assets/b6858bcf-0454-43e9-850e-ca3096aeedfa

wkeese:
> For @include high-contrast-mode('icon-fill') - try removing it in this PR if possible, so I can review it here.

OK, great, done.  

Note: I didn't convert $support-error and $support-warning usages
to a system color. Previously the inline notification icon 🚫 was converted
from red to ButtonText.  Also,  $support-error and $support-warning usages in
_list-box.scss.
    
Also, InlineNotification has a dodgy use of $black-100 that I didn't change.

> I tested https://github.com/carbon-design-system/carbon/issues/19140 and it still shows the same behavior as the issue. I recorded a quick video so you can take a look.  I also saw that @mariyageorge01 made some changes in her PR, I tested it and it works as expected. 

Right, I only addressed the `fill` color for icons, so @mariyageorge01's PR to adjust borders is still valid.

> Maybe we can wait for your PR to be merged first, and then she can continue with her updates? What do you think?

I think that makes sense.

heloiselui:
Hey @wkeese, 

I did some new tests, and with your changes we were able to fix not only the issues you were working on, but also other issues that weren’t even on our radar. Thanks for that!

Just one detail: the `Checkbox` component is now a bit hard to see the check mark ☑️, including in all components that use checkboxes. 

Checkbox in DataTable:
image

Also, the `Grid` is not visible anymore. Although, I might need to open a new issue to check the `Grid`, because even without your changes it was already not showing what it should.

Grid:

https://github.com/user-attachments/assets/6122ef40-fbec-429b-9d35-db5da9bbff36

wkeese:
> Just one detail: the Checkbox component is now a bit hard to see the check mark ☑️, including in all components that use checkboxes.

Good catch. 

This is complicated because Carbon has the idea of inverted colors, but that's not a thing in high-contrast mode.  MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/system-color) suggests that checkmarks should use `SelectedItem` and `SelectedItemText`, so I pushed a change to do that.  It looks like this in Chrome's forced-color emulation:

Screenshot 2025-05-11 at 9 27 32

I **could** get the white-on-black inverted color effect by (for example) setting the foreground-color to `Canvas` and the background-color to `ButtonText`... but that seems counter to how high-contrast mode is supposed to work.

> Also, the Grid is not visible anymore. Although, I might need to open a new issue to check the Grid, because even without your changes it was already not showing what it should.

Yah, I think that was broken before my changes.  The problem is that the grid columns in the story are suddenly missing the `cds--css-grid-column` class, which sets the height to 80px... thus their height is now 0px.

wkeese:
BTW @tay1orjones, do you know why [at least in some settings], `background-color: SelectedItem` doesn't work for where there's text, but it works in other places? 

Screenshot 2025-05-13 at 20 46 18

heloiselui (approved):
I tested this again and everything works as expected. Thanks again @wkeese! 🚀

annawen1 (approved):
tested and LGTM as well!

PR #19355, opened 2025-05-09
fix: Fix focus management for dismissible tags

### LINKED ISSUE #19269: [Bug]: When closing the dismissible tag, the focus doesn't jump to next tag
### Package

@carbon/web-components

### Browser

_No response_

### Package version

latest

### React version

_No response_

### Description

When user closes a dismissible tag, the focus does not shift to the next dismissible tag like it does in the React version

https://github.com/user-attachments/assets/1f081d85-cc6e-4ea6-aea4-4bfd38072978

### Reproduction/example

https://web-components.carbondesignsystem.com/?path=/story/components-tag--dismissible

### Steps to reproduce

1. Go to https://web-components.carbondesignsystem.com/?path=/story/components-tag--dismissible
2. tab to first dismissible tag and hit the enter key to close
3. see that the focus does not shift to the close icon in the next dismissible tag

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #19269 

Fix focus management for dismissible tags

### Changelog

**New**

- Improved accessibility by ensuring focus moves to next available tag when dismissing 

#### Testing / Reviewing

- Go to dismissable tags
- Close the tags  and see if focus shifts to next tag


Gururajj77 (approved):
Works well

packages/web-components/src/components/tag/dismissible-tag.ts
```diff
@@ -90,6 +110,15 @@ class CDSDismissibleTag extends HostListenerMixin(FocusMixin(CDSTag)) {
           )
         ) {
           this.open = false;
+          if (nextFocusableTag) {
+            const nextCloseIcon = (nextFocusableTag as CDSDismissibleTag)
+              ._buttonNode;
+            if (nextCloseIcon) {
+              setTimeout(() => {
+                nextCloseIcon.focus();
+              }, 0);
```
preetibansalui: Hey, Could you check if `setTimeout` is really needed here? It seems to be working fine for me without it. Do we need it for any specific case or scenario?

riddhybansal: @preetibansalui Without the timeout, the focus might be set before other event handlers have completed processing the current event.

preetibansalui: Can we add some checks to prevent it from being set before the others? I don't have an issue with using `setTimeout` but I just want to make sure it's truly necessary in this case.

preetibansalui (approved):
LGTM

PR #19372, opened 2025-05-14
fix: slider track background does not reset

### LINKED ISSUE #18608: [Bug]: Slider track background doesn't reset when set to minimum value and switched between hidden state
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.71.0

### React version

18.2.0

### Description

When the Slider is used with a checkbox to hide/unhide it, the Slider track does not reset when the slider value is set to the minimum value and then the slider state is changed from hidden to visible. Please refer to the recording below.

https://github.com/user-attachments/assets/92b6cb90-1d99-4936-a20c-e19a30e35169

Expected:
Slider track background should be applied correctly when minimum value is set and slider state is changed.

### Reproduction/example

https://stackblitz.com/edit/github-kwcuseew?file=src%2FApp.jsx

### Steps to reproduce

1. Navigate to https://stackblitz.com/edit/github-kwcuseew?file=src%2FApp.jsx

2. Set the slider to minimum value i.e 0
3. Select checkbox to hide and then unhide the slider.
4. Slider track background is incorrect.

Image

### Suggested Severity

None

### Application/PAL

Elyra Canvas

### Code of Conduct

preetibansalui:
Closes #18608 

Slider track background doesn't reset when set to minimum value and switched between hidden state

### Changelog

**Changed**

The issue occurred because the code didn’t enter componentDidUpdate when the value was zero. I’ve added the logic to componentDidMount to ensure the track colour is set when the component mounts.

#### Testing / Reviewing

1. Go to Storybook > Slider > Test
2. Set the slider value to 0
3. Now hide the slider using checkbox and click again to show it.
4. The slider track should display correct background colour.

Note: I will remove the Test story before merge.

## PR Checklist

- [ ] <s>Updated documentation and storybook examples</s>
- [ ] <s>Wrote passing tests that cover this change</s>
- [ ] <s>Addressed any impact on accessibility (a11y)</s>


2nikhiltom (approved):

packages/react/src/components/Slider/Slider.stories.js
```diff
@@ -174,6 +175,33 @@ Default.args = {
...
+  return (
+    <>
+      <Checkbox
+        labelText="Show Slider"
+        id="showSlider"
+        checked={showSlider}
+        onChange={(checked) => setShowSlider(!showSlider)}
+      />
+      {showSlider
+        ? showSlider && (
```
emyarod: wouldn't `showSlider` on L190 be redundant since it's being evaluated on L189?
```suggestion
      {showSlider
        ? (
```

preetibansalui: Thanks for the suggestion @emyarod but this is a test story, I will remove it before merge. Do you still wants me to make changes?

Gururajj77 (approved):

Issue #19380, opened 2025-05-14
[Feature Request]: Link component removes Carbon class names when `as` prop is truthy


sjbeatle:
### The problem

Thank you so much for the convenient `as` prop inclusion for your `Link` component, allowing us to utilize things like `react-router` logic within a Carbon `Link` 🎉

Unfortunately, when using the `as` prop, you are currently removing all Carbon `cds--link` classnames 😞

I see that is intentional, although I'm not entirely sure why. Ideally I think they should remain no matter what, but perhaps it's more appropriate to include an additional property allowing consumers to choose whether or not to keep the Carbon class names?

### The solution

As a `@carbon/react` user, I would like the option to include Carbon link styles in the `Link` component, even if I'm using the `as` property.

This could be accomplished with an additional boolean property user's could leverage. Property name, TBD.

### Package

@carbon/react

### Application/PAL

z/OSMF

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

sjbeatle:
I will work on a PR for this feature... I'll do my best to choose a property name, but by all means feel free to change it once the PR is submitted. Thanks!

PR #19384, opened 2025-05-14
fix(Link): class name logic refactor

### LINKED ISSUE #19379: [Bug]: Link component does not maintain custom classname when `as` prop is truthy
### Package

@carbon/react

### Browser

Chrome, Safari, Firefox, Edge

### Package version

latest

### React version

latest

### Description

Thank you so much for the convenient `as` prop inclusion for your `Link` component, allowing use to utilize things like `react-router` logic within a Carbon `Link` 🎉

Unfortunately, when using the `as` prop, you are currently removing any custom `className` I may've provided to the component. 😞 

### Reproduction/example

N/A

### Steps to reproduce

With the `Link` component:
* add a `className` prop value of "foobar"
* inspect source and see that the custom class name is present 🎉 
* Now, add an `as` prop value, setting it to whatever you'd like, doesn't matter as long as it's truthy
* BUG! inspect source and see that the custom class name is no longer present 😞 

### Suggested Severity

### Application/PAL

z/OSMF

### Code of Conduct

sjbeatle:
Closes #19379 #19380

When using the `as` property, we currently lose all class names, even any custom one added by the user. We want to maintain ~the custom~ all class name(s) in all cases ~, _and_ allow an option for users to keep the Carbon class names~ when using the `as` property.

### Changelog

**New**

- ~`asWithCarbonClasses`: By default, including the `as` property will remove Carbon class selectors: `cds--link`, etc. This property will allow you to continue to use Carbon styles while also using the `as` property.~

**Changed**

- _Do not_ remove any custom class name(s), regardless of `as` being present or not.
- _Do not_ remove any Carbon class name(s), regardless of `as` being present or not.

**Removed**

- N/A

#### Testing / Reviewing

Regression tests were added to the `Link-test.js` file, testing all new/changed use/cases and passing.

## PR Checklist


packages/react/src/components/Link/Link-test.js
```diff
@@ -49,6 +49,36 @@ describe('Link', () => {
...
+    expect(screen.getByRole('heading').classList.length).toBeFalsy();
+  });
+
+  it('should keep Carbon classes on the element with the asWithCarbonClasses property', () => {
+    render(
+      <Link href="https://carbondesignsystem.com" as="h1" asWithCarbonClasses>
+        test without carbon classes
+      </Link>
+    );
+    expect(screen.getByRole('heading').classList.length).toBeTruthy();
```
sjbeatle, line 70: I didn't want to explicitly reference the `cds` class names with static text in the event that they are ever altered. I figured this spec in tandem with the one prior, it pretty much handles a potential regression. The exact class name(s) aren't really relevant to the test, as they are for the custom class test case.

packages/react/src/components/Link/Link.stories.js
```diff
@@ -5,8 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
 import { ArrowRight } from '@carbon/icons-react';
+import React from 'react';
 import Link from './Link';
```
sjbeatle: Some auto-cleanup by Prettier on-save

packages/react/src/components/Link/Link.tsx
```diff
@@ -14,12 +14,8 @@ import React, {
...
 } from 'react';
+import { PolymorphicComponentPropWithRef } from '../../internal/PolymorphicProps';
 import { usePrefix } from '../../internal/usePrefix';
-import { PolymorphicProps } from '../../types/common';
-import {
-  PolymorphicComponentPropWithRef,
-  PolymorphicRef,
-} from '../../internal/PolymorphicProps';
 
 export interface LinkBaseProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
```
sjbeatle, line 20: Some auto-cleanup by Prettier on-save

packages/react/src/components/Link/Link.tsx
```diff
@@ -69,6 +65,11 @@ export interface LinkBaseProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
    * Specify whether you want the link to receive visited styles after the link has been clicked
    */
   visited?: boolean;
+
+  /**
+   * Specify whether you want to use Carbon styles when using the `as` property.
+   */
+  asWithCarbonClasses?: boolean;
```
sjbeatle: I am totally open to this name being changed. Maintainers have permission to update at will.

packages/react/src/components/Link/Link.tsx
```diff
@@ -88,6 +89,7 @@ const LinkBase = React.forwardRef<
   (
     {
       as: BaseComponent,
+      asWithCarbonClasses = false,
```
sjbeatle: default is `false` to maintain current behavior (will not affect current use-cases)

tay1orjones (changes requested):
Let me know what you think, but I don't think we need the prop. The classnames could always be passed to what's rendered.

packages/react/src/components/Link/Link.tsx
```diff
@@ -110,7 +112,8 @@ const LinkBase = React.forwardRef<
     });
     const rel = target === '_blank' ? 'noopener' : undefined;
     const linkProps: AnchorHTMLAttributes<HTMLAnchorElement> = {
-      className: BaseComponent ? undefined : className,
+      className:
+        BaseComponent && !asWithCarbonClasses ? customClassName : className,
```
tay1orjones: From what I can tell this is the only place in the entire codebase that we don't apply the default classes when using the `as` prop. 

https://github.com/carbon-design-system/carbon/blob/a6eb26586b0531ff65150398adebdee7cdd6e4e2/packages/react/src/components/AspectRatio/AspectRatio.tsx#L69
https://github.com/carbon-design-system/carbon/blob/a6eb26586b0531ff65150398adebdee7cdd6e4e2/packages/react/src/components/Grid/Column.tsx#L259
https://github.com/carbon-design-system/carbon/blob/a6eb26586b0531ff65150398adebdee7cdd6e4e2/packages/react/src/components/Grid/FlexGrid.tsx#L45
https://github.com/carbon-design-system/carbon/blob/09985f68c0a1dea07a77a2523ad38b8dfc54ab79/packages/react/src/components/Popover/index.tsx#L474

There's like 20 total of basically the same pattern. It's always up to the user configuring `as` to handle what to do with those classes.

I think this was added in https://github.com/carbon-design-system/carbon/pull/15233 because the assumption was `as` would be most/always be used to convert this to `button`. That's not always the case though and I think it's more likely `as` is used as a way to use a link/anchor tag component from a routing library. That or some custom `<MyLinkComponent>` that renders an anchor tag.

So my view is this is a bug - it's out of convention and counterintuitive. I'd support just making this change without guarding it behind a prop. Anyone relying on those classes not being there could re-remove the styling with something like:

```js
<Link as={Something} className="my-class" />
```

```scss
.my-class[attr*="cds--link--"] {
  all: initial;

  &:hover
  &:active,
  &:active:visited,
  &:active:visited:hover
  &:focus
  &:visited {
    all: initial;
  }
}
```

sjbeatle: On it! Thanks!

sjbeatle: Thanks @tay1orjones!
Changes implemented... I kept two tests for any potential regressions that could be introduced.

Anything else, just let me know! Cheers!

tay1orjones (approved):
LGTM, thank you! 🙏

heloiselui (approved):
Looks good!

PR #19363, opened 2025-05-12
fix: header's right panel storybook fix (React version)

Gururajj77:
Closes #19289 

Right panel's close on loss of focus is fixed, this is for React version, and only a storybook change.

### Changelog

**New**

- added a new state to track the panel's toggle, and close on focus
- added the `aria-label` switch up to work for accessibility

#### Testing / Reviewing

- in the deploy preview, go to the ui-shell's header story, Header w/ Actions and Right Panel.
- click on the header, and tab towards the right panel, and open it
- after you tested the open and close toggle, tab away from it, make sure it closes after the loss of focuss

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
- [ ] Updated documentation and storybook examples
- [ ] Wrote passing tests that cover this change
- [x] Addressed any impact on accessibility (a11y)
- [x] Tested for cross-browser consistency
- [x] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #19400, opened 2025-05-15
fix: header's right panel storybook (19363)

heloiselui:
Closes #19363

Fixed implementation of the header's right panel toggle behavior in Storybook. Focus and panel state logic were mistakenly applied to the App Switcher button instead of the Notifications button.

### Changelog

**Changed**

- Moved `isPanelExpanded` state and handlers from the App Switcher to the Notifications button.
- Updated Storybook story to reflect correct accessibility labels and behavior.

#### Testing / Reviewing

1. Go to React `Deploy Preview` > `UIShell` > `Header w/ Actions and Right Panel`
2. Click on the `Bell icon (notification)`, check if open and close
3. After you tested the open and close, tab away from it, make sure it closes after the loss of focuss

- [N/A] Updated documentation and storybook examples
- [N/A] Wrote passing tests that cover this change

### DIFF
--- a/packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
+++ b/packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
@@ -534,16 +534,16 @@ export const HeaderWActionsAndRightPanel = (args) => {
           <HeaderGlobalAction
             aria-label="Notifications"
             badgeCount={args.badgeCount}
-            isActive
-            onClick={action('notification click')}>
+            isActive={isPanelExpanded}
+            onClick={togglePanel}
+            tooltipAlignment="center"
+            id="notification-button">
             <Notification size={20} />
           </HeaderGlobalAction>
           <HeaderGlobalAction
-            aria-label={isPanelExpanded ? 'Close panel' : 'Open panel'}
-            isActive={isPanelExpanded}
-            onClick={togglePanel}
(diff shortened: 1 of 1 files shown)


annawen1 (commented):
Looks great! One thing, when the focus goes to to the associated trigger action - should the right panel stay open?


https://github.com/user-attachments/assets/be4e5a66-8faa-441e-b313-60c95bf8005a

heloiselui:
> Looks great! One thing, when the focus goes to to the associated trigger action - should the right panel stay open?
> 
>  Screen.Recording.2025-05-19.at.10.49.57.AM.mov

@annawen1, thanks! That really does seem like a bug, I think it’s worth handling that in a separate issue, I can open for us.
I think here, we can keep the focus on the notification. What do you think?

annawen1:
> > Looks great! One thing, when the focus goes to to the associated trigger action - should the right panel stay open?
> > Screen.Recording.2025-05-19.at.10.49.57.AM.mov
> 
> @annawen1, thanks! That really does seem like a bug, I think it’s worth handling that in a separate issue, I can open for us. I think here, we can keep the focus on the notification. What do you think?

Sure, opening a separate issue for it sounds good to me!

annawen1 (approved):

kennylam (approved):

PR #19405, opened 2025-05-16
chore(TabPanel): remove vestigial tabindex setting code

### LINKED ISSUE #19083: [a11y]: TabPanel code to set tabIndex doesn't work
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.80.0

### React version

20

accessibility-checker

### Assistive technology

_No response_

### Description

TabPanel has a bunch of code to set the tabIndex based on whether the content is interactive or not, including this main block:

```ts
useEffect(() => {
  if (!panel.current) {
    return;
  }

  const content = getInteractiveContent(panel.current);
  if (content) {
    setInteractiveContent(true);
    setTabIndex(-1);
  }
}, []);

// tabindex should only be 0 if no interactive content in children
useEffect(() => {
  const node = panel.current;
  if (!node) {
    return;
  }

  function callback() {
    const content = getInteractiveContent(node as HTMLElement);
    if (content) {
      setInteractiveContent(true);
      setTabIndex(-1);
    } else {
      setInteractiveContent(false);
      setTabIndex(0);
    }
  }
  const observer = new MutationObserver(callback);

  observer.observe(node, {
    childList: true,
    subtree: true,
  });
  return () => observer.disconnect();
}, []);
```

However, if you look at https://react.carbondesignsystem.com/?path=/story/components-tabs--contained, you can see that that tabIndex is always -1, both for tabs with and without interactive content:

Image

### Solution

I suggest to fix that by just removing all that code, as I did in #19405, although I understand that's a bit controversial because you're worried about supporting Mac Safari.

### WCAG 2.1 Violation

https://www.w3.org/WAI/standards-guidelines/act/rules/0ssw9k/proposed/

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tabs--contained

### Steps to reproduce

Just look in Chrome developer tools.

### Suggested Severity

None

### Code of Conduct

wkeese:
Fixes #19083, refs #19789.

So remove the complicated `TabPanel` code that tries to do that.

### Changelog

**Removed**

- Removed code to set TabPanel's tabindex to 0 or 1 depending on content.

#### Testing / Reviewing

Storybook.

## PR Checklist


adamalston (commented):
> As [Chrome](https://developer.chrome.com/blog/keyboard-focusable-scrollers) and Firefox now both support keyboard scrolling on divs, we no longer need to set tabindex on scrollable nodes, even if they don't contain any interactive content.

What about Safari and all of the other browsers that are covered in https://github.com/carbon-design-system/carbon/blob/5a866937257a6b524ad004a0a3e426afcb423172/config/browserslist-config-carbon/index.js?

wkeese:
> > As [Chrome](https://developer.chrome.com/blog/keyboard-focusable-scrollers) and Firefox now both support keyboard scrolling on divs, we no longer need to set tabindex on scrollable nodes, even if they don't contain any interactive content.
> 
> What about Safari and all of the other browsers that are covered in https://github.com/carbon-design-system/carbon/blob/5a866937257a6b524ad004a0a3e426afcb423172/config/browserslist-config-carbon/index.js?

I'm going by the comment in https://github.com/carbon-design-system/ibm-products/issues/7043#issuecomment-2886321557, which implies that we don't need to worry about them.

But to be more specific:

* Safari - Tab key on Safari doesn't even focus on buttons, unless you do a special advanced setting.  I know that we don't support VoiceOver on Safari and I assume we don't officially support keyboard either.
* Edge - It's based on Chrome so it should have the same fix.

wkeese:
I'll mark this as a draft though until there's an official decision on what Carbon wants to do. I filed https://github.com/IBMa/equal-access/issues/2281 to talk about removing the requirement from accessibility-checker, and then we can decide on Carbon after that.

heloiselui:
Hey @mbgower, could you give some insights about this topic?

miriam-g:
@heloiselui @ariellalgilmore any news on that? We're waiting for this fix to eliminate accessibility issues in our product.

heloiselui:
> @heloiselui @ariellalgilmore any news on that? We're waiting for this fix to eliminate accessibility issues in our product.

@miriam-g Let us double-check with our accessibility team and get back to you.

mbgower:
Sorry, I was queried on this while I was on vacation, so missed it.
My understanding is that @wkeese is correct and this is the recommended approach, but I'll reconfirm with the a11y team and get back to you.

mbgower:
I confirm that the accessibility team supports this.

heloiselui:
> I confirm that the accessibility team supports this.

Thank you @mbgower, we can move forward with this PR though.

heloiselui (approved):
I tested, also tested this locally and everything looks good. I checked the keyboard navigation, focus behavior, and also ran the IBM Accessibility Checker before and after the change. No accessibility or functional regressions were found, the Tabs still work as expected. Thanks a lot for taking the time to do this!

ariellalgilmore (approved):

wkeese:
Thanks everyone.  I also filed https://github.com/carbon-design-system/carbon/pull/19414, it's basically the same thing except for Table instead of TabPanel.

PR #19414, opened 2025-05-17
chore(Table): remove tabindex setting code

wkeese:
Refs #19789.

So remove the complicated `Table` code that tries to do that.

### Changelog

**Removed**

- Code for setting tabindex
- 
#### Testing / Reviewing

You can test this on
http://localhost:3000/?path=/story/components-datatable-basic--default. If you make your browser narrow enough, the table gets a horizontal scrollbar, and you can tab to it.

## PR Checklist


wkeese:
I'll mark this as a draft though until there's an official decision on what Carbon wants to do. I filed https://github.com/IBMa/equal-access/issues/2281 to talk about removing the requirement from accessibility-checker, and then we can decide on Carbon after that.

packages/react/src/components/DataTable/Table.tsx
```diff
@@ -229,10 +207,7 @@ export const Table = ({
   }, [setTableAlignment, size]);
 
   const table = (
-    <div
-      className={`${prefix}--data-table-content`}
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-      tabIndex={tabIndex ?? (isScrollable ? 0 : undefined)}>
+    <div className={`${prefix}--data-table-content`}>
```
emyarod, line 211: does the `tabIndex` prop become unused after removing this logic? if so, it may throw a lint error. but if we can allow it to be spread directly into the `table` element then we may be able to remove line 116

wkeese: Agreed, I removed the tabindex parameter.

emyarod (approved):
looks good to me!

heloiselui (approved):
LGTM!! Thanks @wkeese!

PR #19415, opened 2025-05-17
chore(TabContent): remove tabindex setting code

wkeese:
Refs #19789.

So remove the complicated `TabContent` code that tries to do that.

I'd also point out that the old code didn't work very well because it only checks the tab content once, so it can't handle content that changes dynamically.

`TabContent` itself seems deprecated and replaced by `TabPanel`. Not sure why it's still here, but I can't just remove it because it's exported.

### Changelog

**Removed**

-  TabContent tabindex setting code

#### Testing / Reviewing

There's no storybook but I ran the automated tests.

## PR Checklist


riddhybansal (commented):
Should we add test cases to verify the keyboard scrolling behavior works correctly without the tabindex?

riddhybansal:
Does it works with Safari as well ?

wkeese:
I thought I marked this as a draft while waiting for https://github.com/IBMa/equal-access/issues/2281 to resolve.

Regardless, to answer your questions:

> Should we add test cases to verify the keyboard scrolling behavior works correctly without the tabindex?

Well, I don't think so.  The purpose of the automated tests is to confirm that Carbon code works correctly, but in this case, there is no Carbon code to test.   There _used to be_ Carbon code, but this PR removes it.

In other words, any test we wrote would simply test that native browser functionality (against plain DOM nodes)... and I don't think that's our job.

PS: Or, you could just say it's not necessary because they already added a test case to https://github.com/IBMa/equal-access/pull/2311/files.

> Does it works with Safari as well ?

As I wrote in the accessibility-checker ticket:

> Safari on Mac doesn't support [automatic tab navigation to scrollable nodes], but we don't really support accessibility on Safari. See https://pages.github.ibm.com/IBMa/able/Test/verify/#macos-support, which essentially says that we don’t support accessibility on MacOS itself (and by corollary, we don’t support Mac OS Safari). At least, we don't support VoiceOver. Plus, Safari's default settings don't let the user tab to anything except inputs.

kennylam (approved):

riddhybansal (approved):

Issue #19366, opened 2025-05-13
[a11y]: Many violations when there's more than one table with expansions


jinhua-ibm:
### Package

@carbon/react

### Package version

1.76.0

### Automated testing tool and ruleset

IBM Equal Access Accessibility Checker

### Description

While having 2 tables with `<TableExpandRow>` and `<TableExpandHeader>`, it will have lots of violations 

<img width="2539" alt="Image" src="https://github.com/user-attachments/assets/4b3e5978-4554-4275-88c7-bf41cf780464" />

### Reproduction/example

https://stackblitz.com/edit/github-u7nfga-b9nzyzrg?file=src%2FApp.jsx,package.json

### Steps to reproduce

Create 2 tables, with <TableExpandHeader> and <TableExpandRow> and perform a scan.
Also, the placement could affect it as well, the demo is showing the case where they are showing on the z-index. But, there could be an issue too when z-index is different, one covers another.

### Suggested Severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon/issues) for duplicate problems

Rudra1402:
Hello @preetibansalui, 
The screenshot attached to the issue shows multiple duplicate ID problems, such as `The <tr> element has the id "expanded-row-a" that is already in use`. I would like to contribute towards fixing them. I have tested the changes locally and am just waiting for approval to make a partial contribution to this issue.

preetibansalui:
Hi @Rudra1402, Thanks for looking into it! Please go ahead and raise a PR.

tay1orjones:
Fixed via https://github.com/carbon-design-system/carbon/pull/20633

PR #19416, opened 2025-05-18
fix(DataTable): fix duplicate expanded row id for multiple tables

Rudra1402:
### Issue:
- Partially Addresses: https://github.com/carbon-design-system/carbon/issues/19366

### Changelog

**Changed**

- Refactored the way the `id` and `aria-controls` attributes are set for expanded rows when multiple tables with expandable rows are present.
- Appended the `instanceId` to each row’s `id` to ensure uniqueness within the DOM.
- Appended the `instanceId` to the `aria-controls` attribute to correctly associate it with the corresponding row `id`.

#### Before making changes:
- Use multiple tables with row expansion.
- Inspect the elements and search for `expanded-row-a`. You’ll notice that multiple rows have the same `id` value.
#### After making changes:
- You’ll notice that appending the `instanceId` makes each `id` unique.

## PR Checklist

### DIFF
--- a/packages/react/src/components/DataTable/DataTable.tsx
+++ b/packages/react/src/components/DataTable/DataTable.tsx
@@ -610,9 +610,10 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
     row: DataTableRow<ColTypes>;
     [key: string]: unknown;
   }) => {
+    const expandedRowId = `expanded-row-${row.id}-${this.instanceId}`;
     return {
       ...rest,
-      id: `expanded-row-${row.id}-${this.instanceId}`,
+      id: expandedRowId,
     };
   };
 
--- a/packages/react/src/components/DataTable/DataTable.tsx
+++ b/packages/react/src/components/DataTable/DataTable.tsx
@@ -590,7 +590,7 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
       onExpand: composeEventHandlers([this.handleOnExpandRow(row.id), onClick]),
       isExpanded: row.isExpanded,
       'aria-label': t(translationKey),
-      'aria-controls': `expanded-row-${row.id}`,
+      'aria-controls': `expanded-row-${row.id}-${this.instanceId}`,
       isSelected: row.isSelected,
       disabled: row.disabled,
     };
@@ -612,7 +612,7 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
   }) => {
     return {
       ...rest,
-      id: `expanded-row-${row.id}`,
+      id: `expanded-row-${row.id}-${this.instanceId}`,
     };
   };
 


Rudra1402:

tay1orjones:
Please do not remove checklist items from the pr template. If some do not apply, just `~strike out the text with tildes~`

jinhua-ibm:
Thanks for submitting the PR! Since I initially reported the issue, I wanted to point out that the problem isn’t limited to just id attributes—there are also violations related to `headers` attributes. I'm not entirely sure how Carbon performs its checks, but I'd recommend running an IBM Equal Access Accessibility Checker scan to catch and address all remaining violations comprehensively.

Rudra1402:
> Thanks for submitting the PR! Since I initially reported the issue, I wanted to point out that the problem isn’t limited to just id attributes—there are also violations related to `headers` attributes. I'm not entirely sure how Carbon performs its checks, but I'd recommend running an IBM Equal Access Accessibility Checker scan to catch and address all remaining violations comprehensively.

Thanks for the note! I actually had a local fix ready for the specific `id` attribute issue, so I submitted that as a starting point. I agree that a full scan would be helpful to catch broader violations like `headers` attributes, but I’d suggest treating those as a separate effort to keep the changes focused. Happy to help where I can!

packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -612,7 +612,7 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
   }) => {
     return {
       ...rest,
-      id: `expanded-row-${row.id}`,
+      id: `expanded-row-${row.id}-${this.instanceId}`,
```
adamalston: Suggestion to store this ID in a variable to keep the implementation DRY and ensure consistency.

Rudra1402: Good catch! I’ve updated to use a variable for the expanded row ID.

Gururajj77:
Hey @Rudra1402 , could you please add the steps to test this PR

Rudra1402:
> Hey @Rudra1402 , could you please add the steps to test this PR

Hello, create 2 DataTables with expandable rows for comparative analysis use case. Run the IBM A11y Checker, you'll notice duplicate id violations.
After the fix, run the IBM scanner again, the `instanceId` will never allow 2 rows from 2 different tables with same `id` even if they use the same dataset.

tay1orjones (commented):
Thanks for taking a look at this. This wasn't caught by our IBM A11y Checker (accessibility-checker) based tests in `e2e/components/DataTable/DataTable-test.avt.e2e.js` because we run the checker against the storybook stories and we don't have a story with multiple `DataTable`s. 

Those are ran in a real browser environment to avoid quirks with jsdom, but the checker _can_ be used in unit tests. I think it's worth trying adding a test for this with something like what's below in `packages/react/src/components/DataTable/__tests__/DataTable-test.js`

```js
it('should have no Accessibility Checker violations with multiple instances', async () => {
  const { container } = render(
    <>
      <DataTable {...mockProps} />
      <DataTable {...mockProps} />
    </>
  );
  await expect(container).toHaveNoACViolations(
    'DataTableMultipleInstances'
  );
});
```

packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -610,9 +610,10 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
     row: DataTableRow<ColTypes>;
     [key: string]: unknown;
   }) => {
+    const expandedRowId = `expanded-row-${row.id}-${this.instanceId}`;
```
adamalston, line 613: It looks like https://github.com/carbon-design-system/carbon/pull/19416#discussion_r2101421980 was only partially addressed. I suggest applying the changes in the following diff to make the implementation more robust by ensuring the same ID is used in both places:

```diff
diff --git a/packages/react/src/components/DataTable/DataTable.tsx b/packages/react/src/components/DataTable/DataTable.tsx
index c761873d1..2780be735 100644
--- a/packages/react/src/components/DataTable/DataTable.tsx
+++ b/packages/react/src/components/DataTable/DataTable.tsx
@@ -561,6 +561,9 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
     return (e) => onClick(e, expandParams);
   };
 
+  getExpandedRowId = (rowId: string) =>
+    `expanded-row-${rowId}-${this.instanceId}`;
+
   /**
    * Get the props associated with the given row. Mostly used for expansion.
    *
@@ -590,7 +593,7 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
       onExpand: composeEventHandlers([this.handleOnExpandRow(row.id), onClick]),
       isExpanded: row.isExpanded,
       'aria-label': t(translationKey),
-      'aria-controls': `expanded-row-${row.id}-${this.instanceId}`,
+      'aria-controls': this.getExpandedRowId(row.id),
       isSelected: row.isSelected,
       disabled: row.disabled,
     };
@@ -610,10 +613,9 @@ class DataTable<RowType, ColTypes extends any[]> extends React.Component<
     row: DataTableRow<ColTypes>;
     [key: string]: unknown;
   }) => {
-    const expandedRowId = `expanded-row-${row.id}-${this.instanceId}`;
     return {
       ...rest,
-      id: expandedRowId,
+      id: this.getExpandedRowId(row.id),
     };
   };
```

tay1orjones:
Superseded by https://github.com/carbon-design-system/carbon/pull/20633

PR #19417, opened 2025-05-19
fix(TreeView): treeview not passing props to treenode children

### LINKED ISSUE #19136: [Bug]: TreeView not passing props to TreeNode children when in controlled state
### Package

@carbon/react

### Browser

Firefox, Safari, Chrome

### Package version

1.76, 1.80.0

### React version

17.0.2, 18.3.1

### Description

A recent carbon upgrade seems to have broken the TreeView controlled state from passing certain props like `selected` and `active` down to its children. I can't programmatically control either state from the outside.

I can reproduce it in several different places, including Stackblitz; however I can only reproduce it in your starter template for Stackblitz (linked in this bug report), whereas if I navigate to Stackblitz.com directly and write the code from scratch, I cannot reproduce the issue, regardless of react versions, carbon versions, etc.

I can also reproduce the issue in our production application, as well as a brand new vite + react application with nothing else but the TreeView code with the bug, hoping to isolate the issue from anything else.

### Reproduction/example

https://stackblitz.com/edit/github-5qgc2mb2?file=src%2FBrokenTreeView.tsx

### Steps to reproduce

- Click on each TreeNode. They should all know which one is the `selected` at any given time. External state is updated correctly, and everything is in sync.
- Click on one of the buttons above the TreeNodes. They should be updating the TreeNodes via props, but they're not. They update external state, but that external state is not showing in the TreeNodes themselves.
- If you hover over the TreeNode to see a tooltip, you can see the `active` and `selected` items remain as whatever the TreeNodes set them to, and they ignore what the outside components set them to.

### Suggested Severity

### Application/PAL

Apptio

### Code of Conduct

Gopi-Agasthia-S-005CIU744:
Closes #19136

Addressed the issue where TreeView doesn't pass props to TreeNode if TreeNode is wrapped by a component

### Changelog

**Changed**

- Refactored the way TreeView and TreeNode passes the prop to its children to recursively check for TreeNode Component in children

#### Testing / Reviewing

- In the TreeView story wrap the TreeNode with a div to check if props are passed correctly

## PR Checklist


Gopi-Agasthia-S-005CIU744:

heloiselui (commented):
Thank you so much for doing this! The code looks great! 

Just one thing, can you add some tests for lines [188–189] and [211–212], as they are not currently covered by our tests.

Also, a small suggestion for clarity (totally optional, both are functionally equivalent):
I would replace in `TreeNode.tsx` file, lines [182–184]: 
> tabIndex: (!(node as ReactElement<TreeNodeProps>).props.disabled && -1) || null,

you could use:
> tabIndex: node.props.disabled ? null : -1,

tay1orjones:
Please do not remove checklist items from the pr template. If some do not apply, just `~strike out the text with tildes~`

Gopi-Agasthia-S-005CIU744:
> Thank you so much for doing this! The code looks great!
> 
> Just one thing, can you add some tests for lines [188–189] and [211–212], as they are not currently covered by our tests.
> 
> Also, a small suggestion for clarity (totally optional, both are functionally equivalent): I would replace in `TreeNode.tsx` file, lines [182–184]:
> 
> > tabIndex: (!(node as ReactElement).props.disabled && -1) || null,
> 
> you could use:
> 
> > tabIndex: node.props.disabled ? null : -1,

 - [ ] Added test case wherein a Treenode is wrapped in another component

 - [ ] Refactored the TreeNode.tsx file -> **used tabIndex: node.props.disabled ? null : -1** instead of **tabIndex: (!(node as ReactElement).props.disabled && -1) || null**

heloiselui (approved):
LGTM!

tay1orjones (approved):

PR #19421, opened 2025-05-19
chore(mend): update repo for issues

alisonjoseph:
I believe this is the correct config to direct issues to the https://github.com/carbon-design-system/mend-issues repo based off of this [slack message](https://ibm-studios.slack.com/archives/C01GWMKLEP7/p1659042269825759)

### DIFF
--- a/.whitesource
+++ b/.whitesource
@@ -1,6 +1,6 @@
 {
   "settingsInheritedFrom": "ibm-mend-config/mend-config",
   "issueSettings": {
-    "minSeverityLevel": "NONE"
+     "issueRepoName": "mend-issues"
   }
 }


annawen1 (approved):

.whitesource
```diff
@@ -1,6 +1,6 @@
 {
   "settingsInheritedFrom": "ibm-mend-config/mend-config",
   "issueSettings": {
-    "minSeverityLevel": "NONE"
+     "issueRepoName": "mend-issues"
```
preetibansalui, line 4: Hey Alison, Is it ok to remove `"minSeverityLevel": "NONE"` to add `issueRepoName`. From what I understand, they serve different purposes, so I just wanted to confirm.

alisonjoseph: Yes I believe so, we currently have it configured to NOT open issues, this should change things so it does open issues, but in the new repo we created, not this repo. Or at least I think that is what will happen.

preetibansalui: Ah, got it. Found it on Mend docs website that default configuration is '`Low`' for '`minSeverityLevel`' if we won't add it here. Thanks for clarifying it.

preetibansalui (approved):

tay1orjones (approved):

PR #19425, opened 2025-05-19
fix(with-layer): update stories

### LINKED ISSUE #19377: update with  Layer stories
add a background box to clarify what the original first background is and use the new Layer prop `withBackground`

There's 28 usages of background: transparent across 17 components we'd need to look into
some are hover styles and such, so lots of stuff that won't matter but some like content switcher will

ariellalgilmore:
Closes #19377

With layer stories are now updated so that it's clear which tokens are being applied at each layer

### Changelog

**Changed**

- new background type for our Storybook's Annotation template
- with-layer stories
   - first rendering of the component is wrapped in a `background` box
   - second/third rendering are wrapped in `layer-01/02` box with a `<Layer withBackground>` wrapped around the components
   - removed `withLayer.scss` since the `withBackground` prop is handling the background now

#### Testing / Reviewing

Check the withLayer stories for each componeont
confirm the Theme withLayer story is still outputting as expected

## PR Checklist


tay1orjones (approved):
There were 118 snapshot changes in percy, I thumbed through them all and they looked correct to me 👍

annawen1 (approved):

PR #19469, opened 2025-05-22
fix: InlineLoading changes browser title to not loading due to missing svg

### LINKED ISSUE #19467: [Bug]: InlineLoading makes brower title as "not loading" when it is set to "inactive"
### Package

@carbon/react

### Browser

Edge, Firefox, Safari, Chrome

### Package version

v1.83.0

### React version

v19.1.0

### Description

Image

### Reproduction/example

https://githubq68tgxvp-4ngh--5173--55edb8f4.local-corp.webcontainer.io/

### Steps to reproduce

1. Place a `InlineLoading` component in the page
2. Make status as `inactive`
3. Observe browser title

### Suggested Severity

Severity 2

### Application/PAL

_No response_

### Code of Conduct

mihui:
Closes #19467

Fixed the changed title of the tab when the status is set to `inactive`

### Changelog

**New**

- N/A

**Changed**

- Added the `svg` tag to wrap the title tag in order to fix the changes to the tab's title

**Removed**

- N/A

#### Testing / Reviewing

1. Use the URL mentioned in the issue #19467 to test the result and see if the browser's title is changed
https://gqairvnmmrgithub-aszi--5173--4d9fd228.local-corp.webcontainer.io/
2. Click on the button to toggle the InlineLoading status
3. Check the tab's title

## PR Checklist


packages/react/src/components/InlineLoading/InlineLoading.tsx
```diff
@@ -121,33 +129,34 @@ const InlineLoading = ({
...
-      return (
-        <title className={`${prefix}--inline-loading__inactive-status`}>
-          {iconLabel}
-        </title>
+      loading.animation = (
+        <svg width="0" height="0" viewBox="0 0 0 0">
+          <title className={`${prefix}--inline-loading__inactive-status`}>
+            {iconLabel}
+          </title>
+        </svg>
```
adamalston: Is it correct to use a custom svg here that's effectively blank? Are there any accessibility implications related to the latter?

packages/react/src/components/InlineLoading/InlineLoading.tsx
```diff
@@ -93,26 +93,34 @@ const InlineLoading = ({
...
+    let loading = {
+      text: description ? (
+        <div className={`${prefix}--inline-loading__text`}>{description}</div>
+      ) : undefined,
+      animation: undefined,
+    };
+
     if (status === 'error') {
-      return (
+      element = (
```
adamalston: Per https://github.com/carbon-design-system/carbon/issues/19467, if the issue only affects the inactive status, why modify the handling for all other statuses? Is the issue more pervasive than reported?

packages/react/src/components/InlineLoading/InlineLoading.tsx
mihui: To not effect the title element in the page that can cause page title change, it is essential to surround the svg to indicate it is title of SVG element, also it is consistent with the other spinning icons, you'll know when you change the status of the InlineLoading component.

packages/react/src/components/InlineLoading/InlineLoading.tsx
mihui: All others are in the animation wrapper, which "inactive" does not necessarily have an animation wrapper, to do so I had two options: 1) Not to change this return statement, but I have to additionally add another "if" to determine if it is not "inactive"; 2) change the whole piece of code to save 1 more logic check for only "inactive" status. So you can see in `loading.animation = (<svg...</svg>)` removes ```<div className={`${prefix}--inline-loading__animation`}>{loading}</div>``` wrapper for `inactive` loading animation.

kennylam:
@mihui, I agree with @adamalston that most of these changes aren't necessary. Looks like `<title>` was introduced incorrectly in https://github.com/carbon-design-system/carbon/pull/17814. I've removed that, as well as the empty svg, so the loader simply does and renders nothing when it's inactive (confirmed by one of the designers in https://github.com/carbon-design-system/carbon/pull/17814#issuecomment-2489303750.

I've also set `aria-live=off` when the loader is inactive, as well as updated the tests.

adamalston (approved):
These changes look good. I reverted the fix and confirmed that the test failed as expected.

heloiselui (approved):
LGTM!

maradwan26 (approved):
LGTM

PR #19491, opened 2025-05-28
Exports Dialog as preview__Dialog  and adds  avt test for Dialog

### LINKED ISSUE #19484: Add VRT for Dialog, Add AVT for Dialog once is exported as experimental
- [ ] Add VRT for Dialog
- [ ] Add AVT for Dialog

### LINKED ISSUE #20109: Exports Dialog as preview__Dialog and make available in storybook

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/19484
Closes https://github.com/carbon-design-system/carbon/issues/20109

This PR Exports Dialog as preview__Dialog and adds avt test cases for Dialog

### Changelog

**New**

Adds test cases for Dailog

#### Testing / Reviewing

ci should pass and verify that test cases makes sense  

## PR Checklist


packages/react/src/components/Dialog/index.tsx
```diff
@@ -114,7 +114,7 @@ interface DialogProps extends HTMLAttributes<HTMLDialogElement> {
   ariaDescribedBy?: string;
 }
 
-const unstable__Dialog = React.forwardRef(
+const preview__Dialog = React.forwardRef(
```
tay1orjones: If we're going to expose this in storybook I think it needs to be exported from the main entrypoint. In doing so, this file should only use the real, non-prefixed names. This will mirror the setup used for PageHeader, #18788

I think this file will need to be renamed `Dialog.tsx` so `index.tsx` can house the barrel export and types. `Dialog.tsx` is where we should use the real non-prefixed names. `Then in packages/react/src/index.(js|ts)` we export with the `preview__` prefix, again like PageHeader:

https://github.com/carbon-design-system/carbon/blob/42d3f0bac50f2a0b26f9a81fede2cdfb8ada1618/packages/react/src/index.js#L127

tay1orjones (changes requested):

packages/react/src/components/Dialog/index.tsx
2nikhiltom: Great catch , Thank you so much. I made necessary change.

tay1orjones (approved):

maradwan26 (approved):

PR #19504, opened 2025-05-28
fix(date-picker): fix focus management and calendar closing

### LINKED ISSUE #12594: [a11y]: Datepicker opens calendar on focus, but does not dismiss on loss of focus; also cannot dismiss with Esc
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

https://react.carbondesignsystem.com/?path=/story/components-datepicker--single-with-calendar#storybook-preview-wrapper

### React version

_No response_

n/a

### Assistive technology

_No response_

### Description

I'm spinning this issue off from the larger issue on keyboard affordances in the datepicker, https://github.com/carbon-design-system/carbon/issues/11800. It looks like that work will be rolled into the V12 datepicker update; however, I think the on-focus behaviour is different enough, and of a higher order of disruption, that it should be given its own assessment and a higher priority, especially if the v12 datepicker release isn't impending.

The following is pretty much what I documented in a [comment in the other issue](https://github.com/carbon-design-system/carbon/issues/11800#issuecomment-1184900142) back in July.

Carbon's [react implementation of a date input with a calendar](https://react.carbondesignsystem.com/?path=/story/components-datepicker--single-with-calendar#storybook-preview-wrapper) departs significantly from this stock HTML interaction, in the following ways:
1. The calendar opens on date input focus
2. The calendar does not close on loss of focus of the date input
3. The calendar cannot be closed by pressing Esc from the date input

**These three items compounded constitute a real possible problem for keyboard users. The calendar could be covering other content info and there is no apparent way to dismiss.**

It IS closed when the keyboard is in the calendar itself and  Esc is pressed, however, it is not intuitive or consistent how a user moves from the input to the calendar.
1. The date input takes only one tab focus. When the user presses Tab again, focus moves to the next item.
2. The user needs to press the Down arrow to move to the calendar. This isn't a traditional operation.
3. As well, the user has to Shift+Tab to get back to the date input, even though the user did not Tab to get to the calendar. 

The solution to this issue is fairly simple:
1. If you are going to have the date picker appear on focus of the date input, make it dismissible with Esc from the date input
2. I would also make Tab be the key that moves to the calendar from the input, but I would rate this as a 'should' not a 'must'. It's a much lower priority thing to fix than dismissing the calendar. 

---

The behaviour is definitely not delightful. It is a debatable failure of 2.1.1 Keyboard. There _is_ a method of operating by keyboard. It's just not expected, and is very cumbersome. It also verges on a failure of [1.4.13 Content on Hover of Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html), but since the calendar does not go away on loss of focus, it fails the spirit of the SC without the normative wording. Finally, it is debatable whether it represents a potential failure of [Focus Visible](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible) (since the calendar can entirely obscure an item with focus). It will definitely fail WCAG 2.2's [Focus Not Obscured](https://w3c.github.io/wcag/understanding/focus-not-obscured-minimum.html), but that will not be required until the second half of 2023 at the earliest. 

### WCAG 2.1 Violation

see note in description

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datepicker--single-with-calendar#storybook-preview-wrapper

### Steps to reproduce

Persistent and non-dismissible calendar
1. Using https://react.carbondesignsystem.com/?path=/story/components-datepicker--single-with-calendar#storybook-preview-wrapper tab to date input.
2. Attempt to dismiss the calendar that appears with the Esc key.
3. Press Tab. Focus will leave the component, yet the calendar does not react.

Inconsistent date input <> calendar operation
1. Having established above that Tab will not move focus from the date input to the calendar, with focus in the date input, press the Down Arrow to move to the calendar (an unintuitive move)
2. Now try to return to the datepicker by pressing the Up Arrow. Instead the selected date in the calendar moves up a week.
3. To return to the datepicker, the user must press Shift+Tab. So, arrow key in one direction, Tab in the other.

### Code of Conduct

warrenmblood:
Closes #12594

Closes calendar on loss of focus via both Tab and Shift+Tab. Also makes Tab the key to move focus to calendar instead of Down Arrow (this makes pressing Shift+Tab to return to the date input field more intuitive). Confirmed that Escape works to close calendar from both the calendar and input field.

### Changelog

**New**
-  Listeners calling `handleInputFieldKeyDown` and `handleCalendarKeyDown` to handle focus navigation and calendar closing on `keydown ` events

**Changed**

- Keyboard focus navigation to calendar (and reopening it) by pressing Tab instead of ArrowDown
- Updated existing tests for calendar closing on loss of focus to include additional cases
- Updated ISO 8601 format test to account for changes in keyboard navigation
- Organized existing unit tests with `datePickerType="range"` into a new test group (some were organized under `minDate` and `maxDate` group)
- Updated existing playwright tests with keyboard nav changes

**Removed**
- `handleArrowDown` function
- `useEffect` containing `handleKeyDown`

#### Testing / Reviewing

- New focus management and calendar closing tests should pass
- Verify that keyboard navigation functions as expected for all DatePicker variants in storybook

## PR Checklist

- [N/A] ~Updated documentation and storybook examples~

### DIFF
--- a/e2e/components/DatePicker/DatePicker-test.avt.e2e.js
+++ b/e2e/components/DatePicker/DatePicker-test.avt.e2e.js
@@ -97,7 +97,7 @@ test.describe('@avt DatePicker', () => {
     // avoid flaky test failures from the keyboard press happening too quickly
     // this retries the keypress along with the focus assertion until it passes
     await expect(async () => {
-      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Tab');
       const today = await page.locator('span.today');
       await expect(today).toBeVisible();
       await expect(today).toBeFocused();
@@ -128,7 +128,7 @@ test.describe('@avt DatePicker', () => {
       page.locator('input#date-picker-input-id-start')
     ).toBeFocused();
     await expect(page.locator('div.flatpickr-calendar')).toHaveClass(/open/);
-    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('Tab');
     await expect(page.locator('span.today')).toBeFocused();
     await page.keyboard.press('Enter');
     await expect(page.locator('span.today')).toBeFocused();
--- a/e2e/components/FluidDatePicker/FluidDatePicker-test.avt.e2e.js
+++ b/e2e/components/FluidDatePicker/FluidDatePicker-test.avt.e2e.js
@@ -76,7 +76,7 @@ test.describe('@avt FluidDatePicker', () => {
     // avoid flaky test failures from the keyboard press happening too quickly
     // this retries the keypress along with the focus assertion until it passes
     await expect(async () => {
-      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Tab');
       await expect(today).toBeVisible();
       await expect(today).toBeFocused();
     }).toPass();
@@ -105,7 +105,7 @@ test.describe('@avt FluidDatePicker', () => {
       page.locator('input#date-picker-input-id-start')
     ).toBeFocused();
     await expect(page.locator('div.flatpickr-calendar')).toHaveClass(/open/);
-    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('Tab');
     await expect(page.locator('span.today')).toBeFocused();
     await page.keyboard.press('Enter');
     await expect(page.locator('span.today')).toBeFocused();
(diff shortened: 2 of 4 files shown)


annawen1 (approved):
LGTM!

heloiselui (approved):
Looks good, just a quick note, lines 753–754 and 852 in `DatePicker.tsx` aren't currently covered by tests. Might be worth adding coverage there.

PR #10060, opened 2021-11-05
feat(react): refactor composed modal to functional component

abbeyhrt:
Closes #10012 

Refactors `ComposedModal` to functional component and updates tests. 

#### Testing / Reviewing

Pull down PR and confirm that ComposedModal works as expected

PR #19531, opened 2025-05-30
feat(modal): mount/unmount on open change

### LINKED ISSUE #18935: [Feature Request]: Mount and unmount Modal/ComposedModal

timfuhrmann:
Closes https://github.com/carbon-design-system/carbon/issues/18935

> [!NOTE]
> **Most LOC come from indentation, slightly changing composition, tests and stories. I recommend to activate "Hide whitespace" for a code review.**

As of right now all modals remain mounted in disregard of its open state. This is due to the exit animation, as an unmount will interrupt any transition/animation. Instead of having the modal mounted at all times, this PR proposes to defer an unmount until the exit animation has finished instead. 

The current behaviour is a bottleneck in terms of state management and performance. Also, it makes devs grab for bad react practices (e.g. resetting state on close, introduction of `useEffects` to sync initial values, bad composition, etc.) which tremendously increase complexity, the probability of bugs, visual jank and all sorts of side effects, leading to a deep cut in quality. Other component libs use an `isExiting` state to defer unmounting the modal. This PR follows a similar approach. 

This also will tremendously reduce the complexity of modal or other related components, since any logic included, such as event handlers, useEffects, etc. will never run as long as the modal is not opened. 

Happy to elaborate on the reasons why and their results, if necessary. 

### Changelog

**New**

- `usePresence`: Hook to defer the update of `isOpen` to `false` until all of the ref's animations have finished
```javascript
const [isPresent, isExiting] = usePresence(ref, open);
```
- `ModalPresence/ComposedModalPresence`: This greatly improves the possibilities of composition together with monolithic exports. Because while the modals' children are unmounted, its calling parent is not. This is fine for components where composition is not a problem (e.g. `ComposedModal`), since everything can be moved to their children already. But when the modal is monolithic and calling component stateful, its states would not be unmounted. Exposing `ModalPresence` helps to offer this flexibility for the monolithic modal as well. However, this introduces a lot more complexity than simply using `usePresence` would, as a single context must only be consumed by one modal, hence this needs to handle exclusivity.
- `withModalPresence/withComposedModalPresence`: TBD/Up to carbon team (Higher-Order Component)

```javascript
// use it just like before, when no state is necessary or you are fine with not unmounting it
const FooModal = ({ onSubmit }) => {
  const [barState, setBarState] = useState()
  return <Modal onRequestSubmit={() => onSubmit(barState)}  />;
}
```
```javascript
// or move presence boundary above calling parent if you want state to be unmounted
const FooModal = ({ open, onSubmit }) => {
  return (
    <ModalPresence open={open}>
      <CustomModal onSubmit={onSubmit} />
    </ModalPresence>
  );
}

const CustomModal = ({ onSubmit }) => {
  const [barState, setBarState] = useState()

  // queries, mutations, local data management etc. will be cleaned up properly on close

  return <Modal onRequestSubmit={() => onSubmit(barState)} />;
}
```
```javascript
// or use HOC
const CustomModal = withModalPresence(({ onSubmit }) => {
  const [barState, setBarState] = useState()

  // queries, mutations, local data management etc. will be cleaned up properly on close

  return <Modal onRequestSubmit={() => onSubmit(barState)} />;
})
```

**Changed**

- Updates tests to run for all implementations: without presence, opt-in by `FeatureFlags` and opt-in by `ModalPresence`
- Updates feature flag stories for both `Modal` & `ComposedModal`

#### Testing / Reviewing

1. Go to modal's story with state manager
2. Modal is not present in DOM
3. Open modal
4. Modal is present in DOM
5. Fill input
6. Close modal
7. Exit animation finishes
8. Modal is removed from DOM
9. Open modal again
10. Input is empty

## PR Checklist

- [x] Made sure previous test are passing

cc @tay1orjones


packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
...
+    isOpen ? 'idle' : 'finished'
+  );
+  const isExiting = exitState === 'active';
+
+  if (!isOpen && exitState === 'idle') {
+    setExitState('active');
+  }
+  if (isOpen && exitState !== 'idle') {
+    setExitState('idle');
+  }
```
adamalston: Should this functionality be inside of an effect? Otherwise, how would the act of calling `setState` during a render be prevented?

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (!animations) {
+      handleAnimationEnd();
+      return;
+    }
+
+    try {
+      Promise.all(animations.map((animation) => animation.finished)).then(
+        handleAnimationEnd
+      );
+    } catch {}
```
adamalston: Any reason not to log the error?

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
...
+  }, [isAnimating, handleAnimationEnd]);
+
+  return [isOpen || exitState !== 'finished', isExiting] as const;
```
adamalston: Suggestion to return an object instead of an array so that the entities are named.

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+  if (isOpen && exitState !== 'idle') {
+    setExitState('idle');
+  }
+
+  const handleAnimationEnd = useCallback(() => {
+    setExitState('finished');
+  }, []);
+
+  const isAnimating = isExiting;
```
adamalston: Is having both `isExiting` and `isAnimating` necessary?

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
...
+  useIsomorphicEffect(() => {
+    if (!ref.current || !isAnimating) return;
+    if (!('getAnimations' in ref.current)) {
+      handleAnimationEnd();
+      return;
+    }
+
+    const animations = ref.current.getAnimations();
+    if (!animations) {
```
adamalston: Per the types, `animations` is `Animation[]`. If that's correct, this condition will never be satisfied. In other words, it's dead code. So, is the condition incorrect or is the type incorrect?

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
+import { RefObject, useCallback, useState } from 'react';
```
adamalston: This file needs a copyright notice.

```suggestion
/**
 * Copyright IBM Corp. 2025
 *
 * This source code is licensed under the Apache-2.0 license found in the
 * LICENSE file in the root directory of this source tree.
 */

import { RefObject, useCallback, useState } from 'react';
```

packages/react/src/internal/usePresence.ts
adamalston, line 1: Can you add testing for this hook?

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (isOpen && exitState !== 'idle') {
+    setExitState('idle');
+  }
+
+  const handleAnimationEnd = useCallback(() => {
+    setExitState('finished');
+  }, []);
+
+  const isAnimating = isExiting;
+  useIsomorphicEffect(() => {
```
adamalston: Is there a noticable difference between using `useIsomorphicEffect` here vs. `useEffect`? Just wondering why `useIsomorphicEffect` was used.

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -234,7 +240,40 @@ export interface ModalProps extends HTMLAttributes<HTMLDivElement> {
...
+  )
+);
+
+type ModalPresenceContextType = {
+  presenceRef: RefObject<HTMLDivElement | null>;
+  isExiting: boolean;
+};
+const ModalPresenceContext = createContext<ModalPresenceContextType>(
+  undefined as unknown as ModalPresenceContextType
+);
```
adamalston: Any reason not to initialize the context with a valid value?

```suggestion
const ModalPresenceContext = createContext<ModalPresenceContextType>({
  presenceRef: { current: null },
  isExiting: false,
});
```

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -13,6 +13,11 @@ import React, {
   type HTMLAttributes,
   type ReactNode,
   type Ref,
+  createContext,
+  RefObject,
+  PropsWithChildren,
+  useMemo,
+  useContext,
```
adamalston: Suggestion to sort these new imports and use `type` imports wherever applicable.

packages/react/src/internal/usePresence.ts
timfuhrmann: When calling a set function during render, react will immediately re-render this component, before rendering the children. This helps to properly pick up the open state, without having to wait for a second re-render of the whole tree and avoids the component ever being in a state where `isOpen` is `false`, but `exitState` still in `idle` mode. This pattern also described [here](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes), if you are interested.

packages/react/src/internal/usePresence.ts
timfuhrmann: Good point

packages/react/src/internal/usePresence.ts
timfuhrmann: Removed, thanks

packages/react/src/internal/usePresence.ts
timfuhrmann: Updated the condition to check length instead, thanks

packages/react/src/components/Modal/Modal.tsx
timfuhrmann: Just a habit to have occurrences throw if not called within a provider. Will update.

packages/react/src/internal/usePresence.ts
timfuhrmann: I was under the assumption there might be. However after some testing there is no difference. Changing to useEffect.

packages/react/src/internal/usePresence.ts
timfuhrmann: It's not really an error, but `animation.finished` rejects when an animation is canceled manually or implicitly

packages/react/src/internal/usePresence.ts
```diff
@@ -0,0 +1,54 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (!ref.current || !isExiting) return;
+    if (!('getAnimations' in ref.current)) {
+      handleAnimationEnd();
+      return;
+    }
+
+    const animations = ref.current.getAnimations();
```
tay1orjones: When applied to the Dialog, I think this will need to include the subtree option which will return animations on pseudo-elements like the `::backdrop`.

```suggestion
    const animations = ref.current.getAnimations({ subtree: true });
```

packages/react/src/internal/usePresence.ts
adamalston: Got it, makes sense that it rejects on cancel. That said, should we still log and/or handle the rejection? Even if it’s not a "real" error, ignoring it could make debugging it in the future more difficult if unexpected cancellations happen.

packages/react/src/components/Modal/Modal-test.js
```diff
@@ -18,7 +18,7 @@ const prefix = 'cds';
 describe('Modal', () => {
   it('should add extra classes that are passed via className', () => {
     render(
-      <Modal data-testid="modal-1" className="custom-class">
```
tay1orjones, line 21: This would be a breaking change, right? Needing to add `open` to these tests so they pass

packages/styles/scss/components/modal/_modal.scss
```diff
@@ -43,55 +43,51 @@
...
       content: '';
       inline-size: 100vw;
       inset-block-start: 0;
       inset-inline-start: 0;
-      opacity: 0;
-      transition:
-        opacity $duration-moderate-02 motion(exit, expressive),
-        visibility 0ms linear $duration-moderate-02;
-      visibility: hidden;
-    }
```
tay1orjones, line 57: `@carbon/styles` is used in all the framework variants. These style changes break the web-components modal, for instance: [deploy preview](https://deploy-preview-19531--v11-carbon-web-components.netlify.app/?path=/story/components-modal--with-state-manager) vs [prod](https://web-components.carbondesignsystem.com/?path=/story/components-modal--with-state-manager) (can no longer be closed). It could also break `carbon-components-svelte`, `carbon-components-angular`, etc.

So to ensure backwards compatibility we'll need to guard these changes behind something the other framework authors can opt-in to. A new class to scope selectors is probably the easiest way.

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -234,7 +240,41 @@ export interface ModalProps extends HTMLAttributes<HTMLDivElement> {
...
 }
 
-const Modal = React.forwardRef(function Modal(
+const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
+  ({ open, ...props }, ref) => (
+    <ModalPresence open={open}>
+      <ModalDialog {...props} ref={ref} />
+    </ModalPresence>
+  )
+);
```
tay1orjones: I'm not confident we can do this without breaking consumers. Tests could be written based on the modal always being in the DOM, for instance.

tay1orjones (commented):
The direction here looks great. I'm mostly concerned about backwards compatibility and ensuring these changes can be taken by all consumers without any impact. If we can't ensure that, we'll need to guard it somehow. 

* A new or existing feature flag
  * Dialog is already marked unstable and behind a flag in Modal/ComposedModal
* A new prop, sometimes can be an existing one if it's an enum or something. 
* Opt in through composition
  * Rough idea could be <ModalPresence> setting a boolean in a new context that Modal could then read from and modify it's internal logic based on if it's wrapped with ModalPresence or not.

I'm curious how that sounds, or what you think might make more sense

packages/react/src/components/Modal/Modal.tsx
adamalston: I can confirm that this change would break testing for some of the UI's I've worked on. Our tests check if the modal is either hidden or visible. Regardless, they assume that it is always present in the DOM.

packages/styles/scss/components/modal/_modal.scss
timfuhrmann: Makes sense. Thank you for pointing this out!

packages/react/src/components/Modal/Modal.tsx
timfuhrmann: Yes this will break tests, will come back to you soon on backwards compatibility.

packages/react/src/components/Modal/Modal-test.js
timfuhrmann: Yes, previously you could test your modal's contents even if it was closed, moving forward it would have to be opened first.

timfuhrmann:
> The direction here looks great. I'm mostly concerned about backwards compatibility and ensuring these changes can be taken by all consumers without any impact. If we can't ensure that, we'll need to guard it somehow.

@tay1orjones Good point, backwards compatibility is something I have not considered with this PR yet. Without providing a way to opt in, this indeed is a breaking change, especially for tests, as most teams have not been necessarily opening the modal, before testing its contents.

I like your suggestions and I am always a fan of composition, a feature flag also sounds like a good option. However the direction might depend on whether you are planning to deprecate the version that's always present in the DOM in the future or if you prefer to keep both versions? 

Edit:
I actually like a combination of both composition and feature flag. A feature flag makes sense in terms of future adoption as well as generally opting into the new behaviour of the modal. If a user decides to use `ModalPresence` as described above though, I would assume that he willingly opts in too. This should improve DX as the user will be able to use ModalPresence without having to explicitly opt in every time again (in case of incremental migration). What do you think?

packages/react/src/internal/usePresence.ts
timfuhrmann: Good to know. Thanks!

timfuhrmann:
Update: I updated this PR to provide backwards compatibility. As of right now, the feature has to be enabled via feature flag. However, I am still thinking about a way to auto opt-in as mentioned above. 

All tests are updated to cover both versions of the modal.

timfuhrmann:
Update: `ModalPresence` now also supports auto opt-in when used standalone. @tay1orjones What do you think about this approach to backwards compatibility? Will add something similar to ComposedModal.

```javascript
/**
* Backwards compatible solution, no API changes
*/
const WithoutPresence = () => {
  const [isOpen, setIsOpen] = useState(false);
  return <Modal open={isOpen} onRequestClose={() => setIsOpen(false)} />;
};
```
```javascript
/**
* Modal will manually opt in via feature flags
*/
const WithFeatureFlags = () => {
  const [isOpen, setIsOpen] = useState(false);
  return (
    <FeatureFlags enablePresence>
      <Modal open={isOpen} onRequestClose={() => setIsOpen(false)} />
    </FeatureFlags>
  );
};
```
```javascript
/**
* Modal will auto opt-in, necessary styles will be applied automatically, 
* all children and their local states will be unmounted
*/
const CustomModal = ({ onClose }) => {
  const [foo, setFoo] = useState();
  return <Modal onRequestClose={() => setIsOpen(false)} />;
}

const WithComposition = () => {
  const [isOpen, setIsOpen] = useState(false);
  return (
    <ModalPresence open={isOpen}>
      <CustomModal onClose={() => setIsOpen(false)} />
    </ModalPresence>
  );
};
```

timfuhrmann:
> Update: `ModalPresence` now also supports auto opt-in when used standalone. @tay1orjones What do you think about this approach to backwards compatibility? Will add something similar to ComposedModal.
> 
> ```js
> /**
> * Backwards compatible solution, no API changes
> */
> const WithoutPresence = () => {
>   const [isOpen, setIsOpen] = useState(false);
>   return <Modal open={isOpen} onRequestClose={() => setIsOpen(false)} />;
> };
> ```
> 
> ```js
> /**
> * Modal will manually opt in via feature flags
> */
> const WithFeatureFlags = () => {
>   const [isOpen, setIsOpen] = useState(false);
>   return (
>     <FeatureFlags enablePresence>
>       <Modal open={isOpen} onRequestClose={() => setIsOpen(false)} />
>     </FeatureFlags>
>   );
> };
> ```
> 
> ```js
> /**
> * Modal will auto opt-in, necessary styles will be applied automatically, 
> * all children and their local states will be unmounted
> */
> const CustomModal = ({ onClose }) => {
>   const [foo, setFoo] = useState();
>   return <Modal onRequestClose={() => setIsOpen(false)} />;
> }
> 
> const WithComposition = () => {
>   const [isOpen, setIsOpen] = useState(false);
>   return (
>     <ModalPresence open={isOpen}>
>       <CustomModal onClose={() => setIsOpen(false)} />
>     </ModalPresence>
>   );
> };
> ```

Update: Both Modal and ComposedModal are now following said approaches to opt in.

timfuhrmann:
#### Summary
This PR implements a hook called `usePresence` that can be applied to whatever element to defer its unmount by subscribing to its ongoing animations. For both Modal & ComposedModal this is hook is integrated via Modal-/ComposedModalPresence, which mounts a React context. 
While the hook could be used directly in both components without having to create a context, this improves DX by a lot, since a developer can shift the presence boundary to a higher level, especially helpful for monolithic exports such as the `Modal` or IBM's `Tearsheet`. However, this addition not mandatory and obviously introduces more complexity than only using the hook would, as it has to handle exclusivity, since only a single modal must read from one context.

TODOs I am aware of:

- [x] Storybook Documentation
- [x] ComposedModal's `isOpen` state: I have not been able to completely understand the purpose of syncing `open` with `isOpen`& `wasOpen` compared to just using `open`. The docs are only mentioning controlling the open state by providing `open`, but nothing in regards to an uncontrolled way of using this component. Anyways, in case it will be necessary, I am happy to adjust my implementation to cover this, if you can fill me in on the expected outcome. Thank you!

tay1orjones:
Storybook docs were something I was going to mention, I think it'll be really helpful to have those 👍 

> ComposedModal's `isOpen` state: I have not been able to completely understand the purpose of syncing `open` with `isOpen`& `wasOpen` compared to just using `open`. The docs are only mentioning controlling the open state by providing `open`, but nothing in regards to an uncontrolled way of using this component. Anyways, in case it will be necessary, I am happy to adjust my implementation to cover this, if you can fill me in on the expected outcome. Thank you!

The internal `isOpen` state exists to support a "semi-controlled" use case where a consumer passes in `open={true}` and they don't have `onClose` wired up. In this case, the ComposedModal still should close on a click outside or pressing <kbd>esc</kbd>. It was initially added as part of refactoring ComposedModal to be a functional component. 

* https://github.com/carbon-design-system/carbon/pull/10060

timfuhrmann:
@tay1orjones I have implemented both support for the uncontrolled ComposedModal as well as added storybook documentation. Please let me know if you think this is sufficient.

tay1orjones (approved):
This looks great, I pushed one small commit adding the flag to the feature flags overview page in storybook.

Thanks so much for taking this on, the multiple reviews and revisions you had to do. Landing such a well thought out and executed enhancement like this is awesome! 🎯💯

ariellalgilmore (approved):
LGTM!!!

PR #19532, opened 2025-05-30
[a11y] ExpandableSearch has unnecessary aria-label="button"

### LINKED ISSUE #19516: [a11y]: ExpandableSearch element: unnecessary inclusion of `aria-label="button"` attribute
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.78.0

### React version

18.3.1

### Assistive technology

MacOS Voiceover

### Description

In IBM OpenPages, we use Carbon's `ExpandableSearch` element. The search icon, which serves as a button to expand the search bar, must have an appropriate label attribute to pass the accessibility checker, which it does. In fact, there are two attributes that, on their own, would cause the element to pass the accessibility checker. There is an `aria-label="button"` attribute, as well as an `aria-labelledby="search-input-:rar:-search"`, which appropriately points to a label that is fitting for this button. Therefore, we are able to pass the accessibility checker with the `ExpandableSearch` element.

However, according to [this documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label), you are not to use both `aria-label` and `aria-labelledby` on the same element, because `aria-labelledby` will take precedence over `aria-label` if both are applied.

`aria-label` and `aria-labelledby` are not fields we can manually set, according to [Carbon's documentation](https://react.carbondesignsystem.com/?path=/docs/components-search--overview), so the attributes `aria-label` and `aria-labelledby` must be automatically set by Carbon.

`aria-labelledby` is the more appropriate attribute to include, because it allows us to point to a label that's more fitting and descriptive, whereas an `aria-label` of "button" isn't informative.

Lastly, though the [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label), mentions that `aria-labelledby` will take precedence, using MacOS Voiceover when the focus is on the `ExpandableSearch` yields a redundant and possibly confusing explanation:

<em>"Search for field name, collapsed, button, Search for field name, search. You are currently on a button. To click this button, press Control-Option-Space."</em>

Again, I think removing the default attribute `aria-label="button"` would solve this.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://stackblitz.com/edit/github-amzubdxd?file=src%2FApp.jsx

### Steps to reproduce

It should already be reproduced. Note that the `aria-label` and `aria-labelledby` attributes are provided even though I don't set them in the code.

### Suggested Severity

None

### Code of Conduct

maradwan26:
Closes #19516

This [documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) states: "Don't use both on the same element because `aria-labelledby` will take precedence over aria-label if both are applied." Therefore, I have removed the `aria-label` and kept `aria-labelledby` since `aria-label` is redundant.

### Changelog

**New**

- Nothing new

**Changed**

- Nothing changed

**Removed**

- Removed aria-label="button"

#### Testing / Reviewing

- Visit deploy preview
- Inspect ExpandableSearch
- Notice that there is no aria-label, we just have aria-labelledby since we should not have both
- Ensure there are no breaking changes for other Search variations


riddhybansal (approved):
LGTM 💯

2nikhiltom (approved):

Gururajj77 (approved):

PR #19554, opened 2025-06-03
fix(OverflowMenu): prevent onOpen infinite loop

### LINKED ISSUE #19153: [Bug]: OverflowMenu onOpen fires multiple times if it changes state
### Package

@carbon/react

### Browser

Chrome

### Package version

1.80.0

### React version

19.1.0

### Description

In Carbon React v1.79.0 and later, the `onOpen` callback of the OverflowMenu component is called multiple times per user interaction if onOpen triggers a state update.
This behaviour differs from previous versions, where onOpen would only be called once per open interaction.
In the latest version (v1.80.0) it can call `onOpen` endlessly (like in the provided example).
(Last good version is v1.78.2 where onOpen is called only once per user open interaction). 

### Reproduction/example

https://stackblitz.com/edit/github-1ljccm-3wwykhcu?file=package.json,src%2FApp.jsx

### Steps to reproduce

1. Click the overflow menu button
2. Check console logs

Expected:
The onOpen message is logged only once if overflow menu opens via user interaction.

Actual:
The onOpen message will be logged endlessly until you close the overflow menu.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/19153

Fixes infinite loop bug where onOpen callback would trigger repeatedly when it caused content changes that affected menu positioning. 

### Changelog

**New**

Moved onOpen from positioning callback (handlePlace) to state change effect to ensure it's called exactly once per menu opening.

**Changed**

OverflowMenu onOpen callback now triggers on state change instead of positioning events
**Removed**

#### Testing / Reviewing
Open deploy preview - and go to "Test" story 
Verify onOpen is called exactly once when menu opens
Verify menu still opens/closes normally
Verify content changes don't trigger additional onOpen calls

- [x] ⚠️ BEFORE MERGE ⚠️  the test story needs to be removed 

## PR Checklist

- ~[ ] Updated documentation and storybook examples~ N/A
- ~[ ] Addressed any impact on accessibility (a11y)~ N/A

### DIFF
--- a/packages/react/src/components/OverflowMenu/OverflowMenu.tsx
+++ b/packages/react/src/components/OverflowMenu/OverflowMenu.tsx
@@ -308,9 +308,7 @@ export const OverflowMenu = forwardRef<HTMLButtonElement, OverflowMenuProps>(
     useEffect(() => {
       if (open && !prevOpenState.current) {
         onOpen();
-      }
-
-      if (!open && prevOpenState.current) {
+      } else if (!open && prevOpenState.current) {
         onClose();
       }
 
--- a/packages/react/src/components/OverflowMenu/OverflowMenu.tsx
+++ b/packages/react/src/components/OverflowMenu/OverflowMenu.tsx
@@ -305,13 +305,17 @@ export const OverflowMenu = forwardRef<HTMLButtonElement, OverflowMenuProps>(
       }
     }, []);
 
-    // Call `onClose` when menu closes.
     useEffect(() => {
+      if (open && !prevOpenState.current) {
+        onOpen();
+      }
+
       if (!open && prevOpenState.current) {
         onClose();
       }
+
       prevOpenState.current = open;
-    }, [open, onClose]);
+    }, [open, onClose, onOpen]);
 
     useOutsideClick(wrapperRef, ({ target }) => {
       if (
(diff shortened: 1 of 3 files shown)


preetibansalui (approved):

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -305,13 +305,17 @@ export const OverflowMenu = forwardRef<HTMLButtonElement, OverflowMenuProps>(
       }
     }, []);
 
-    // Call `onClose` when menu closes.
     useEffect(() => {
+      if (open && !prevOpenState.current) {
+        onOpen();
+      }
+
       if (!open && prevOpenState.current) {
```
adamalston: Should this code be an `else if`?

riddhybansal (approved):
This LGTM !! Just Remember to remove test story!!

packages/react/src/components/OverflowMenu/OverflowMenu.tsx
2nikhiltom: Good catch! these conditions are mutually exclusive and else if makes more sense. thanks

PR #19573, opened 2025-06-05
fix(TimePicker): Assignment of className props only to the container element

### LINKED ISSUE #19518: [Bug]: TimePicker `className` is applied to more than the wrapper
### Package

@carbon/react

### Browser

_No response_

### Package version

latest

### React version

18

### Description

When using the `TimePicker` component, I leveraged the `className` prop. Description states:
>  Specify an optional className to be applied to the container node

But, when the component rendered, styles were applied in an unexpected manner. Further investigation showed the `className` shows up in two additional nodes along with the wrapper.

* Wrapper node: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/TimePicker/TimePicker.tsx#L264
* TimePicker node: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/TimePicker/TimePicker.tsx#L210
* TimePickerInput node: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/TimePicker/TimePicker.tsx#L196

Image

Image

Image

### Reproduction/example

https://stackblitz.com/edit/github-g9rvvspr?file=src%2Findex.scss,src%2FApp.tsx

### Steps to reproduce

Go to referenced StackBlitz URL and see the style applied to three nodes when the `className` prop is used on the root component.

### Suggested Severity

### Application/PAL

z/OSMF

### Code of Conduct

NAS-IBM:
Closes https://github.com/carbon-design-system/carbon/issues/19518

### Changelog

**Changed**

- Updated unit test cases
- Assigning of "className" property to the TimePicker Node and TimePicker Input Node
- Two props (inputClassName and pickerClassName) are introduced to handle the backward compatibility

#### Testing / Reviewing
- Define the "className" attribute for the TimePicker Component
- Inspect the DOM Element

For example, 
````
<TimePicker id="time-picker" className="time-picker-container" />
````
For the above component, DOM would be rendered like,
image

## PR Checklist

~~Addressed any impact on accessibility (a11y)~~
~~Tested for cross-browser consistency~~

### CODE REVIEW

**Comment by NAS-IBM** [comment]

**Comment by NAS-IBM** [comment]
recheck

File: packages/react/src/components/TimePicker/TimePicker.stories.js [hunk 1]
```diff
@@ -30,7 +30,11 @@ export default {
 
 export const Default = (args) => {
   return (
-    <TimePicker id="time-picker" labelText="Select a time" {...args}>
+    <TimePicker
+      id="time-picker"
+      labelText="Select a time"
+      {...args}
+      className="time-picker-container">
```
tay1orjones: This should remain configurable in the controls pane. Anything after `...args` overrides the value from the controls pane.

```suggestion
```

packages/react/src/components/TimePicker/TimePicker-test.js
```diff
@@ -20,8 +20,11 @@ describe('TimePicker', () => {
     });
 
     it('passes classNames as expected', () => {
-      render(<TimePicker id="time-picker" className="🚀" />);
-      expect(screen.getByRole('textbox')).toHaveClass('🚀');
+      const { container } = render(
+        <TimePicker id="time-picker" className="🚀" />
+      );
+      expect(container.querySelector('.🚀')).toBeInTheDocument();
+      expect(screen.getByRole('textbox')).not.toHaveClass('🚀');
```
tay1orjones: Just to avoid unicode parsing issues and keep this closer to real-world usage:

```suggestion
        <TimePicker id="time-picker" className="test-class" />
      );
      expect(container.querySelector('.test-class')).toBeInTheDocument();
      expect(screen.getByRole('textbox')).not.toHaveClass('test-class');
```

tay1orjones (changes requested):
Thanks for taking a look at this! 

Anyone relying on the classname being in/on `timePickerClasses` `timePickerInputClasses` could be effected by this change. We'll need to provide those consumers relying on the previous behavior a path to easily reinstate their custom styling or tests, etc.

My suggestion would be to:

- add `pickerClassName` prop
  - > "Specify an optional className to be applied to the container that wraps the input and TimePickerSelect children"
- add `inputClassName` prop
  - > "Specify an optional className to be applied to the input"
- add associated tests for all this

packages/react/src/components/TimePicker/TimePicker.stories.js
NAS-IBM: @tay1orjones Thank you for the feedback. I have updated the code.

packages/react/src/components/TimePicker/TimePicker-test.js
NAS-IBM: @tay1orjones Thank you for the feedback. I have updated the code.

tay1orjones (approved):
LGTM, thanks!

heloiselui (approved):

PR #19577, opened 2025-06-05
fix(modal): launcherButtonRef support with dialog feature flag

### LINKED ISSUE #19404: [a11y]: Modal's `launcherButtonRef` does not work with `enable-dialog-element` flag
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.82

### React version

18

N/A

### Assistive technology

none

### Description

This is about Modals when using the `enableDialogElement` feature flag.

In some situations, when you close the Modal, focus goes back to `<body>` rather than where the modal was launched from.

This happens even when using `launcherButtonRef`.

So there are two issues:

1. Poor accessibility.  User needs to tabs dozens of times to get back to where they were before.
2. Modal not working according to spec.

Specifically, the spec implies that `launcherButtonRef` should work regardless of any feature flag settings:

```js
/**
 * Provide a ref to return focus to once the modal is closed.
 */
launcherButtonRef?: Ref<HTMLButtonElement>;
```

(I interpret that description to mean that Modal must focus `launcherButtonRef` on close, regardless of what node `launcherButtonRef` points to.)

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://stackblitz.com/edit/github-ykbrpwcg-eyfjd2nc?file=src%2FExample%2FExample.jsx

### Steps to reproduce

Try launching a modal from the button on https://stackblitz.com/edit/github-ykbrpwcg-eyfjd2nc?file=src%2FExample%2FExample.jsx.  When you close it, focus returns to the launch button, as expected.  This is thanks to the native support of the `<dialog>` component to do the right thing.

However, try launching a modal from the overflow menu.  When you close it, the focus goes to `<body>`, even though I specifically set `launcherButtonRef` so that wouldn't happen.  This is not good.

The workaround is to set focus on the `onRequestClose` and `onRequestSubmit` handlers, probably wrapped in a `setTimeout()`.   But I think that (as stated in the documentation), Modal should obey the setting of `launcherButtonRef`.

### Suggested Severity

None

### Code of Conduct

warrenmblood:
Closes #19404

### Changelog

**New**

- Added prop `focusAfterCloseRef` to the Dialog component for the user to specify an element to be focused after the dialog closes

- `launcherButtonRef` prop in Modal and ComposedModal is now internally mapped to `focusAfterCloseRef` in the `enable-dialog-element` flag, which fixes the issue of the prop not working when the flag is enabled

**Changed**

- Removed `launcherButtonRef` and `focusAfterCloseRef` from all storybook controls panes

#### Testing / Reviewing

- CI tests should pass
- New tests in this PR should verify that `focusAfterCloseRef` in Dialog and `launcherButtonRef` in the `enable-dialog-element` flag of Modal & ComposedModal are working properly

## PR Checklist


packages/react/src/components/Dialog/index.tsx
```diff
@@ -178,6 +185,14 @@ const unstable__Dialog = React.forwardRef(
       }
     }, [modal, open]);
 
+    useEffect(() => {
+      if (!open && focusAfterCloseRef) {
+        setTimeout(() => {
+          focusAfterCloseRef.current?.focus();
+        });
```
tay1orjones: Anytime we step outside of react's scheduling I think it's a good idea to add a comment explaining why.

packages/react/src/components/Dialog/index.tsx
```diff
@@ -52,6 +53,11 @@ interface DialogProps extends HTMLAttributes<HTMLDialogElement> {
    */
   className?: string;
 
+  /**
+   * Provide a ref to return focus to once the dialog is closed.
+   */
+  focusAfterCloseRef?: RefObject<HTMLElement | null>;
```
tay1orjones, line 59: This can be hidden in the controls pane in storybook since refs aren't really configurable in the controls pane. In `argTypes`:

```
focusAfterCloseRef: {
  table: {
    disable: true,
  },
},
```

[storybook docs reference](https://storybook.js.org/docs/essentials/controls#disable-controls-for-specific-properties)

tay1orjones (approved):
LGTM, just a couple suggestions

packages/react/src/components/Dialog/index.tsx
kennylam: Probably a good idea to [clear](https://developer.mozilla.org/en-US/docs/Web/API/Window/clearTimeout) that timeout too.

kennylam (approved):

PR #19511, opened 2025-05-29
fix(fonts): unquote unicode-range and font-display values

petersandor:
Closes #19298
Closes #19324

### Changelog

**Changed**

- `unicode-range` and `font-display` values are now without quotes

#### Testing / Reviewing

1. Start any Carbon sample page with the styles loaded (fonts included) in Firefox
2. Open developer tools and toggle CSS logging
3. Check for console messages mentioning unknown descriptor in `unicode-range` and `font-display` 

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
- ~[ ] Updated documentation and storybook examples~
- ~[ ] Wrote passing tests that cover this change~
- ~[ ] Addressed any impact on accessibility (a11y)~
- [x] Tested for cross-browser consistency
- [ ] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #19588, opened 2025-06-07
chore(fonts): revert previous fix, unquote unicode-range and font-display values only

petersandor:
Partially reverts #19511

### Changelog

**Changed**

- use unquoted `unicode-range` and `font-display` values by default

#### Testing / Reviewing

1. Start any Carbon sample page with the styles loaded (fonts included) in Firefox
2. Open developer tools and toggle CSS logging
3. Check for console messages mentioning unknown descriptor in `unicode-range` and `font-display`

## PR Checklist


preetibansalui (changes requested):
Hi @petersandor, could you please remove the `@use 'sass:string'` import if it's no longer being used? I believe it was added in the previous PR.

tay1orjones (approved):
Nice 👍  Thanks again for this revision!

preetibansalui (approved):
Thanks for the changes

PR #18207, opened 2024-12-05
chore(deps): update internal monorepo deps to react v19

tay1orjones:
Closes #16282 

The peerDependency ranges were updated in https://github.com/carbon-design-system/carbon/pull/18458. This gets the monorepo itself running on react 19, including all the necessary [typescript changes](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#typescript-changes)

#### Changelog

**Changed**

- update devDeps to react v19
- update `@types/react` to 19

#### Testing / Reviewing

- ...

PR #6804, opened 2020-09-09
docs(Modal): add storybook MDX example

emyarod:
Ref #6591

This PR adds a custom modal state manager example and MDX docs to the existing Modal docs

#### Testing / Reviewing

Review the new modal state manager example and the storybook Docs tabs updates

PR #19590, opened 2025-06-08
fix: update outside click behavior for modals

### LINKED ISSUE #13712: [a11y]: Modal: Follow guidance on dismissing the modal by clicking elsewhere

> A modal dialog, by definition, is supposed to be the only operable thing on the screen. Theoretically, the area outside the modal should be _inert_. Clicking it should do nothing.
> Modals by design have a Cancel (or No) button as well as the X button to close, both of which allow affordances for a user to easily dismiss. There does not seem to be a good reason to allow the area outside the modal to be interactive.
> 
> The only [functional demonstration in the React storybook](https://react.carbondesignsystem.com/?path=/story/components-modal--with-state-manager) IS dismissable by clicking outside, even though Carbon says such [transactions modals](https://carbondesignsystem.com/components/modal/usage/#transactional-modal) are not dismissed by clicking outside.
> 
> The issue is made a bit problematic since Carbon also introduces the idea of a passive modal, which _can_ be dismissed in this way. In my [discussion on Modals](https://github.com/carbon-design-system/carbon/discussions/12726) I recommend that the Passive modal be called something else. There are lots of non-modal dialog prompts and interactions where clicking outside can dismiss content. I just feel that a modal dialog may not be the place to allow it.
> Note: I am trying to get clarity from ARIA authorities on whether clicking outside is even _allowed_ in a modal dialog. Their response may help inform the approach here. **Sev: discussion**
> 
> At the least, only the passive should be dismissable by clicking outside, to match the Carbon documentation.

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/13712

Updated outside click behavior for modals.

### Changelog

**New**

- Added documentation regarding passive modals.

**Changed**

- Updated outside click behavior for modals.
- Fixed accessibility issues in tests.
- Fixed discrepencies in modal docs.

#### Testing / Reviewing

I left a `TODO` with a question that hopefully someone can answer.

```sh
yarn test packages/react
```

## PR Checklist

### DIFF
--- a/packages/react/src/components/ComposedModal/ComposedModal.tsx
+++ b/packages/react/src/components/ComposedModal/ComposedModal.tsx
@@ -6,6 +6,7 @@
  */
 
 import React, {
+  Children,
   cloneElement,
   useEffect,
   useRef,
@@ -335,8 +336,16 @@ const ComposedModal = React.forwardRef<HTMLDivElement, ComposedModalProps>(
       const { target } = evt;
       const mouseDownTarget = onMouseDownTarget.current;
       evt.stopPropagation();
+      const containsModalFooter = Children.toArray(childrenWithProps).some(
+        (child) => isComponentElement(child, ModalFooter)
+      );
+      const isPassive = !containsModalFooter;
+      const shouldCloseOnOutsideClick =
+        (isPassive && preventCloseOnClickOutside !== false) ||
+        (!isPassive && preventCloseOnClickOutside === true);
+
       if (
-        !preventCloseOnClickOutside &&
+        shouldCloseOnOutsideClick &&
         target instanceof Node &&
         !elementOrParentIsFloatingMenu(target, selectorsFloatingMenus) &&
         innerModal.current &&
--- a/packages/react/src/components/ComposedModal/ComposedModal.tsx
+++ b/packages/react/src/components/ComposedModal/ComposedModal.tsx
@@ -340,9 +340,9 @@ const ComposedModal = React.forwardRef<HTMLDivElement, ComposedModalProps>(
         (child) => isComponentElement(child, ModalFooter)
       );
       const isPassive = !containsModalFooter;
-      const shouldCloseOnOutsideClick =
-        (isPassive && preventCloseOnClickOutside !== false) ||
-        (!isPassive && preventCloseOnClickOutside === true);
+      const shouldCloseOnOutsideClick = isPassive
+        ? preventCloseOnClickOutside !== false
+        : preventCloseOnClickOutside === true;
 
       if (
         shouldCloseOnOutsideClick &&
--- a/packages/react/src/components/Modal/Modal.mdx
+++ b/packages/react/src/components/Modal/Modal.mdx
@@ -1,4 +1,10 @@
-import { Story, ArgTypes, Canvas, Unstyled, Meta } from '@storybook/addon-docs/blocks';
+import {
+  Story,
+  ArgTypes,
+  Canvas,
+  Unstyled,
+  Meta,
+} from '@storybook/addon-docs/blocks';
 import Modal from '../Modal';
 import { InlineNotification } from '../Notification';
 import CodeSnippet from '../CodeSnippet';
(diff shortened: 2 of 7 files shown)


packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -335,8 +336,16 @@ const ComposedModal = React.forwardRef<HTMLDivElement, ComposedModalProps>(
       const { target } = evt;
       const mouseDownTarget = onMouseDownTarget.current;
       evt.stopPropagation();
+      const containsModalFooter = Children.toArray(childrenWithProps).some(
+        (child) => isComponentElement(child, ModalFooter)
+      );
+      const isPassive = !containsModalFooter;
+      const shouldCloseOnOutsideClick =
+        (isPassive && preventCloseOnClickOutside !== false) ||
```
riddhybansal: const shouldCloseOnOutsideClick = isPassive 
  ? preventCloseOnClickOutside !== false  
  : preventCloseOnClickOutside === true;

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -959,7 +959,19 @@ Modal.propTypes = {
...
    */
-  preventCloseOnClickOutside: PropTypes.bool,
+  preventCloseOnClickOutside: (
+    props: ModalProps,
+    propName: string,
+    componentName: string
+  ) => {
+    if (!props.passiveModal && props[propName] === false) {
+      return new Error(
+        `${componentName}: \`${propName}\` should not be \`false\` when \`passiveModal\` is \`false\`. Non-passive ${componentName}s should not be dismissible by clicking outside.`
```
riddhybansal: Should we add test case that mock console.error to verify the error is thrown when passiveModal={false} and preventCloseOnClickOutside={false}, and that valid prop combinations don't trigger errors.

riddhybansal (changes requested):

packages/react/src/components/Modal/Modal.mdx
```diff
@@ -58,6 +64,17 @@ text.
 
 <ArgTypes />
 
+{/* <!-- prettier-ignore-start --> */}
+
+{/*
+TODO: Why are there `ComposedModal` docs in the `Modal` docs?
+The documentation is being duplicated between this file and
+`packages/react/src/components/ComposedModal/ComposedModal.mdx`.
+Currently, there are differences between the `ComposedModal` docs in these two files.
```
tay1orjones: See the discussion at https://github.com/carbon-design-system/carbon/pull/6804

At this point it's probably worth separating the content to only apply to Modal. A nice addition could be to mention ComposedModal near the top of the page. and vice versa in ComposedModal.mdx

```diff
`<Modal>` allows you to use its bespoke set of contents. `children` prop maps to
the modal body content. You can also use `modalLabel`, `modalHeading`,
`secondaryButtonText` and `primaryButtonText` props to change the corresponding
text.
+ 
+ For a more composable approach to modals, see the [ComposedModal](...link/to/composedmodal.mdx).
```

tay1orjones: This can be a separate PR, or create an issue

tay1orjones (approved):
+1 on the test for the proptype validator function, but I don't think it's a blocker to merge. It's somewhat implicitly tested by the allowed values used in the existing and new tests.

packages/react/src/components/Modal/Modal.mdx
adamalston: https://github.com/carbon-design-system/carbon/issues/19624

packages/react/src/components/Modal/Modal.tsx
adamalston: Do prop validators not work in React 19? I see other components that include them and also have error logging functionality. It looks like some of these changes originated in https://github.com/carbon-design-system/carbon/pull/18207, specifically in files like `packages/react/src/components/Button/Button.tsx`, `packages/react/src/components/Tabs/Tabs.tsx`, etc.

Example:

https://github.com/carbon-design-system/carbon/blob/624890b679409d619b1c8c6e99ad5cf041ad7917/packages/react/src/components/Tabs/Tabs.tsx#L142-L146

https://github.com/carbon-design-system/carbon/blob/624890b679409d619b1c8c6e99ad5cf041ad7917/packages/react/src/components/Tabs/Tabs.tsx#L198-L205

riddhybansal (approved):

PR #19597, opened 2025-06-09
fix: `SelectableTile` onChange firing twice in react strict mode

### LINKED ISSUE #19521: [Bug]: SelectableTile onChange callback being called twice when the tile is clicked
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.83

### React version

19

### Description

onChange callback is called twice in dev environments when the tile is clicked or keypressed. This is due to the callback being processed inside the state update callback, which react will call twice in dev mode.

This can lead to many issues, but for me it leads to the following error as I am updating state in the callback.

`Cannot update a component (`xxx`) while rendering a different component (`ForwardRef(SelectableTile)`). To locate the bad setState() call inside `ForwardRef(SelectableTile)`, follow the stack trace as described in https://react.dev/link/setstate-in-render`

I am working around this issue by checking to see if I've already made the state update before trying again.

### Reproduction/example

https://stackblitz.com/edit/vitejs-vite-r8jcjzyb?file=src%2FApp.tsx

### Steps to reproduce

Have a selectableTile component with an onChange callback. Click the selectableTile, watch the callback get control 2x when in dev mode.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #19521 

Fixed the onChange handler in the SelectableTile fires twice under React's strict mode by replacing render-time state sync with useEffect and centralizing selection logic

### Changelog

**New**

- Added useEffect to handle state sync with selected prop instead of doing it during render
- Added centralized handleSelectionChange function to manage all selection updates

**Changed**

- Moved state synchronization from render phase to useEffect to fix strict mode issues
- Updated event handlers to use proper TypeScript types
- Combined duplicate selection logic into single reusable function

**Removed**

- Removed render-time state sync logic that was causing double firing
- Removed unnecessary onChange handler from div element
- Removed prevSelected state that was used for render-time comparisons

#### Testing / Reviewing

To test the bug first, as we can't reproduce this in local directly
- providing the [stackblitz](https://stackblitz.com/edit/github-fdjpjsn2?file=src%2FApp.tsx,package.json,src%2Fmain.tsx&preset=node=) with the bug here to understand how `onChange` is fired twice
- in the main.tsx, remove the `<React.StrictMode>` tag, and test it again.
- see that `onChange` handler fires only once.

To test the fix:
- pull this PR down on local, and build the application.
- add this line in decorators array inside the `packages/react/.storybook/preview.js` file 

```
(Story) => (

<React.StrictMode>

<Story />

</React.StrictMode>

),

```

Image provided for the exact line and place to add this
image
- open the test story "Selectable on change test" and keep the console open.
- click on the tile to make sure that the `onChange` handler is fired once.

## PR Checklist

### DIFF
--- a/packages/react/src/components/Tile/Tile.tsx
+++ b/packages/react/src/components/Tile/Tile.tsx
@@ -7,6 +7,7 @@
 
 import React, {
   cloneElement,
+  useCallback,
   useEffect,
   useRef,
   useState,
@@ -480,7 +481,11 @@ export const SelectableTile = React.forwardRef<
   const keyDownHandler = onKeyDown;
 
   const [isSelected, setIsSelected] = useState<boolean>(selected);
-  const [prevSelected, setPrevSelected] = useState<boolean>(selected);
+
+  // Use useEffect to sync with prop changes instead of render-time logic
+  useEffect(() => {
+    setIsSelected(selected);
+  }, [selected]);
 
   const classes = cx(
     `${prefix}--tile`,
@@ -497,46 +502,39 @@ export const SelectableTile = React.forwardRef<
     className
   );
 
-  function handleClick(evt) {
+  // Single function to handle selection changes
+  const handleSelectionChange = useCallback(
+    (evt: React.SyntheticEvent, newSelected: boolean) => {
+      setIsSelected(newSelected);
+      onChange(evt as React.ChangeEvent<HTMLDivElement>, newSelected, id);
+    },
+    [onChange, id]
+  );
+
+  function handleClick(evt: React.MouseEvent<HTMLDivElement>) {
     evt.preventDefault();
     evt?.persist?.();
     if (
       normalizedDecorator &&
       decoratorRef.current &&
-      decoratorRef.current.contains(evt.target)
+      decoratorRef.current.contains(evt.target as Node)
     ) {
       return;
     }
-    setIsSelected((prevSelected) => {
-      const newSelected = !prevSelected;
-      onChange(evt, newSelected, id);
-      return newSelected;
-    });
+
+    const newSelected = !isSelected;
+    handleSelectionChange(evt, newSelected);
     clickHandler(evt);
   }
 
-  function handleKeyDown(evt) {
+  function handleKeyDown(evt: React.KeyboardEvent<HTMLDivElement>) {
     evt?.persist?.();
     if (matches(evt, [keys.Enter, keys.Space])) {
       evt.preventDefault();
-      setIsSelected((prevSelected) => {
-        const newSelected = !prevSelected;
-        onChange(evt, newSelected, id);
-        return newSelected;
-      });
+      const newSelected = !isSelected;
+      handleSelectionChange(evt, newSelected);
     }
     keyDownHandler(evt);
-  }
-
-  function handleChange(event) {
-    const newSelected = event.target.checked;
-    setIsSelected(newSelected);
-    onChange(event, newSelected, id);
-  }
-
-  if (selected !== prevSelected) {
-    setIsSelected(selected);
-    setPrevSelected(selected);
   }
 
   // AILabel is always size `xs`
@@ -559,7 +557,6 @@ export const SelectableTile = React.forwardRef<
       tabIndex={!disabled ? tabIndex : undefined}
       ref={ref}
       id={id}
-      onChange={!disabled ? handleChange : undefined}
       title={title}
       {...rest}>
       <span
--- a/packages/react/src/components/Tile/Tile.tsx
+++ b/packages/react/src/components/Tile/Tile.tsx
@@ -409,7 +409,7 @@ export interface SelectableTileProps extends HTMLAttributes<HTMLDivElement> {
    * The empty handler of the `<input>`.
    */
   onChange?(
-    event: ChangeEvent<HTMLDivElement>,
+    event: MouseEvent<HTMLDivElement> | KeyboardEvent<HTMLDivElement>,
     selected?: boolean,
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Tile/Tile.tsx
```diff
@@ -497,48 +502,41 @@ export const SelectableTile = React.forwardRef<
...
+  );
+
+  function handleClick(evt: React.MouseEvent<HTMLDivElement>) {
     evt.preventDefault();
     evt?.persist?.();
     if (
       normalizedDecorator &&
       decoratorRef.current &&
-      decoratorRef.current.contains(evt.target)
+      decoratorRef.current.contains(evt.target as Node)
```
adamalston: The type assertion isn't necessary. There's a type-safe way to implement this condition:

```suggestion
      evt.target instanceof Node &&
      decoratorRef.current.contains(evt.target)
```

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -497,48 +502,41 @@ export const SelectableTile = React.forwardRef<
     className
   );
 
-  function handleClick(evt) {
+  // Single function to handle selection changes
+  const handleSelectionChange = useCallback(
+    (evt: React.SyntheticEvent, newSelected: boolean) => {
+      setIsSelected(newSelected);
+      onChange(evt as React.ChangeEvent<HTMLDivElement>, newSelected, id);
```
adamalston: The type assertion shouldn't be necessary. Is the `onChange` type correct? I’m asking since `ChangeEvent` isn’t typically used with `div` elements.

```diff
diff --git a/packages/react/src/components/Tile/Tile.tsx b/packages/react/src/components/Tile/Tile.tsx
index e0a6090e2..1ad036210 100644
--- a/packages/react/src/components/Tile/Tile.tsx
+++ b/packages/react/src/components/Tile/Tile.tsx
@@ -408,11 +408,11 @@ export interface SelectableTileProps extends HTMLAttributes<HTMLDivElement> {
   /**
    * The empty handler of the `<input>`.
    */
-  onChange?(
-    event: ChangeEvent<HTMLDivElement>,
+  onChange?: (
+    event: MouseEvent<HTMLDivElement> | KeyboardEvent<HTMLDivElement>,
     selected?: boolean,
     id?: string
-  ): void;
+  ) => void;
 
   /**
    * Specify the function to run when the SelectableTile is clicked
@@ -502,11 +502,13 @@ export const SelectableTile = React.forwardRef<
     className
   );
 
-  // Single function to handle selection changes
   const handleSelectionChange = useCallback(
-    (evt: React.SyntheticEvent, newSelected: boolean) => {
+    (
+      evt: MouseEvent<HTMLDivElement> | KeyboardEvent<HTMLDivElement>,
+      newSelected: boolean
+    ) => {
       setIsSelected(newSelected);
-      onChange(evt as React.ChangeEvent<HTMLDivElement>, newSelected, id);
+      onChange(evt, newSelected, id);
     },
     [onChange, id]
   );
```

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -497,48 +502,41 @@ export const SelectableTile = React.forwardRef<
...
-      return newSelected;
-    });
+
+    const newSelected = !isSelected;
+    handleSelectionChange(evt, newSelected);
     clickHandler(evt);
   }
 
-  function handleKeyDown(evt) {
+  function handleKeyDown(evt: React.KeyboardEvent<HTMLDivElement>) {
```
adamalston: `KeyboardEvent` is already imported.

```suggestion
  function handleKeyDown(evt: KeyboardEvent<HTMLDivElement>) {
```

packages/react/src/components/Tile/Tile.tsx
```diff
@@ -497,48 +502,41 @@ export const SelectableTile = React.forwardRef<
...
+  // Single function to handle selection changes
+  const handleSelectionChange = useCallback(
+    (evt: React.SyntheticEvent, newSelected: boolean) => {
+      setIsSelected(newSelected);
+      onChange(evt as React.ChangeEvent<HTMLDivElement>, newSelected, id);
+    },
+    [onChange, id]
+  );
+
+  function handleClick(evt: React.MouseEvent<HTMLDivElement>) {
```
adamalston: `MouseEvent` is already imported.

```suggestion
  function handleClick(evt: MouseEvent<HTMLDivElement>) {
```

packages/react/src/components/Tile/Tile.tsx
adamalston, line 1: Several JSDoc comments in this file refer to an `input`, but there isn’t one. Perhaps a new issue should be created to clarify or correct the documentation — I’m not sure. I only started investigating the discrepancy because the `onChange` JSDoc comment mentions an `input` and I couldn't find one. Any thoughts?

packages/react/src/components/Tile/Tile.tsx
Gururajj77: Have added these changes

packages/react/src/components/Tile/Tile.tsx
Gururajj77: Have added these changes

packages/react/src/components/Tile/Tile.tsx
Gururajj77: I think we need to look and update the jsdoc comments

maradwan26 (approved):
Works as expected !!

2nikhiltom (approved):

PR #19612, opened 2025-06-10
fix: pageheader content and tab bar style fixes

### LINKED ISSUE #19546: PageHeader: WC - Implement Content Zone

### LINKED ISSUE #19547: PageHeader: WC - Implement Tabs zone

annawen1:
Closes #19546
Closes #19547

This PR fixes some layout and style bugs around the page-header content and tabs bar zones 

Specifically the tabs overflow button backgrounds, setting them to use `$layer-01` instead of `$background` and adjusts / removes the spacing in the tab bar zone when there are no tags provided.

### Changelog

**Changed**

- set background color for tabs overflow button using variable so that the background can be adjusted in the `page-header-tabs` component styles
- fix tests for the tab zone and regenerate snapshots

**Removed**

- removed the slot name for `tabs`, it should be assumed that the open slot is for the tabs components - one less step for adopter when authoring
- remove div wrapper around tag slotted content, was adding styles that was interfering with layout in tab bar

#### Testing / Reviewing

ci-checks pass

## PR Checklist

### DIFF
--- a/packages/web-components/src/components/combo-button/__tests__/__snapshots__/combo-button-test.snap.js
+++ b/packages/web-components/src/components/combo-button/__tests__/__snapshots__/combo-button-test.snap.js
@@ -1167,3 +1167,35 @@ snapshots['Menu behavior should handle menu item clicks'] = `<cds-button
 </slot>
 `;
 /* end snapshot Menu behavior should handle menu item clicks */
+snapshots['cds-combo-button should support `tooltips-content` attribute'] =
+  `<cds-button
+  has-main-content=""
+  kind="primary"
+  size="lg"
+  tab-index="0"
+  tooltip-alignment=""
+  tooltip-position="top"
+  type="button"
+>
+  Primary action
+</cds-button>
+<cds-icon-button
+  align="top"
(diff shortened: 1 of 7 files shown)


tay1orjones (approved):
It sounds like this is outside the scope of what you're doing here but the tabs are unusable on smaller screens

https://github.com/user-attachments/assets/394a777a-32bd-41de-a232-42043a5c48b4

We can divert that into a new issue if you agree. Otherwise LGTM

annawen1:
> It sounds like this is outside the scope of what you're doing here but the tabs are unusable on smaller screens
> 
>  2025-06-10.at.13.57.53-MAIN-Google.Chrome.mp4 
> We can divert that into a new issue if you agree. Otherwise LGTM

I pushed a quick fix to stack the tags on top of tabs in the small breakpoint. That should hold us up for now until we get the collapsing logic in.


https://github.com/user-attachments/assets/bc9f858c-85b5-47d3-8412-ca51cb955f06

heloiselui (approved):
Looks great @annawen1!

PR #19613, opened 2025-06-10
chore: add overflowhandler readme

### LINKED ISSUE #7632: Add README.md to overflowHandler utility
As a follow on to #7553, we noticed that our Storybook includes docs that would also be beneficial to the overflowHandler utility in `@carbon/utilities`.

### Acceptance criteria

- [x] Copy content from overflowHandler docs[^1] to a README on [`@carbon/utilities` overflowHandler](https://github.com/carbon-design-system/carbon/tree/main/packages/utilities/src/overflowHandler)
- [x] Add an explicit mentioning of the utility being framework agnostic, and supports react and lit, and many other frameworks

[^1]: packages/ibm-products-web-components/src/utilities/overflow-handler/overflow-handler.mdx

davidmenendez:
Closes https://github.com/carbon-design-system/ibm-products/issues/7632

Added a `README.md` in the `overflowHandler` folder. Includes a link to the storybook examples and documentation. The content is the same content found in the [Storybook docs](https://ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-overflowhandler--docs).


packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+The only required argument necessary to consume this utility is `container`.
+Once the handler has been instantiated, an event handler is created that checks
+the width of the container when it's rendered or resized. The handler adds the
+necessary `data-hidden` attributes to the items that should be hidden. These
+items can then be hidden using the following style in the implementation.
+
+```css
+[data-hidden]:not([data-fixed]) { // scope if needed
```
adamalston: Is `// scope if needed` supposed to be a code comment? If so, I suggest using `/* */` which is the official syntax for a comment in CSS. Alternatively, maybe it would be fine to just delete it?

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```js
+import { createOverflowHandler } from '@carbon/utilities';
+
+createOverflowHandler({
+  container, // The document node that holds the items. HTMLElement
+  maxVisibleItems, // Maximum number of visible items. If provided, only this number of items will be shown. number | undefined
+  onChange: (visibleItems: HTMLElement[], hiddenItems: HTMLElement[]) => void, // Callback function invoked when the visible and hidden items change.
+  dimension, // The dimension to consider for overflow calculations. "width" | "height" | undefined
+});
```
adamalston: Any reason that `onChange` is the only property which includes types? If not, I suggest adding types for everything else - types that match the function's interface.

```suggestion
createOverflowHandler({
  container: HTMLElement,
  maxVisibleItems?: number,
  onChange: (visibleItems: HTMLElement[], hiddenItems: HTMLElement[]) => void,
  dimension?: "width" | "height",
});
```

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,51 @@
...
+Using the `onChange` callback, you have access to an array holding the visible
+and hidden elements. This gives you the flexibility to access the hidden items
+and render them elsewhere, such as a modal or popover.
+
+The handler needs to be re-initialized whenever items are added or removed, when
+an item changes its width dynamically, or when the offset changes its width
+dynamically, to update the overflow with the new sizes.
+
+For example implementation, please refer to the example code in
+[Storybook](https://ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-overflowhandler--docs).
```
adamalston: Should these sections be under difference headings? They don't seem to fit under the "Attribute definitions" heading.

```md
## Custom Handling via `onChange`

...

## Re-initialization guidelines

...

## Example implementation

...
```

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+createOverflowHandler({
+  container, // The document node that holds the items. HTMLElement
+  maxVisibleItems, // Maximum number of visible items. If provided, only this number of items will be shown. number | undefined
+  onChange: (visibleItems: HTMLElement[], hiddenItems: HTMLElement[]) => void, // Callback function invoked when the visible and hidden items change.
+  dimension, // The dimension to consider for overflow calculations. "width" | "height" | undefined
+});
+```
+
+The only required argument necessary to consume this utility is `container`.
```
adamalston: Maybe I'm missing something, this statement seems inaccurate as `onChange` is required per `OverflowHandlerOptions`:

https://github.com/carbon-design-system/carbon/blob/4331e5fec7ae539e5b4ac6dac1d0ca8faede8aa6/packages/utilities/src/overflowHandler/overflowHandler.ts#L145

Further, if I change the `onChange` property in a TypeScript file to `undefined`, I get the following type error:

```
Type 'undefined' is not assignable to type '(visibleItems: HTMLElement[], hiddenItems: HTMLElement[]) => void'.
```

davidmenendez:
@adamalston thanks for the review! update based on your suggestions.

adamalston:
> The content is the same content found in the [Storybook docs](https://ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-overflowhandler--docs).

Should that content also be updated?

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,57 @@
...
+```js
+import { createOverflowHandler } from '@carbon/utilities';
+
+createOverflowHandler({
+  container: HTMLElement,
+  maxVisibleItems?: number,
+  onChange: (visibleItems: HTMLElement[], hiddenItems: HTMLElement[]) => void,
+  dimension: 'width': 'height',
+});
+```
```
adamalston: Regarding https://github.com/carbon-design-system/carbon/pull/19613#discussion_r2139174643, and after giving it some more thought, this code is not an example of how to use `createOverflowHandler`, it's just a re-documentation of the function's types. Should an actual example be added, or should this section be removed from the doc?

davidmenendez:
> > The content is the same content found in the [Storybook docs](https://ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-overflowhandler--docs).
> 
> Should that content also be updated?

yes i will totally open a follow up PR in the other repo based on your suggestions. thanks!

packages/utilities/src/overflowHandler/README.md
davidmenendez: yeah that's a valid point. this is something we've also been discussing in products regarding how verbose we want documentation like this to be when live code examples exist in storybook, but i agree with you. i think calling this an "example" isn't exactly accurate. i will update this to make it a more realistic implementation.

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+const initializeHandler = () => {
+  if (handler) {
+    console.log('Handler found. Removing and re-initiating...');
+    document.removeEventListener('DOMContentLoaded', initializeHandler);
+    handler.disconnect();
+    return;
+  }
+
+  // initiate the handler only when the dom settles and is stable, so that the items are at right dimensions before initialization.
```
adamalston:
```suggestion
  // initiate the handler only when the dom settles and is stable, so that the items are at right dimensions before initialization.
```
```suggestion
  // initiate the handler only when the DOM settles and is stable, so that the items are at the correct dimensions before initialization.
```

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,97 @@
...
+  </div>
+`;
+```
+
+Once the handler has been instantiated, an event handler is created that checks
+the width of the container when it's rendered or resized. The handler adds the
+necessary `data-hidden` attributes to the items that should be hidden. These
+items can then be hidden using the following style in the implementation.
+
+In the above example, once the handler has been instantiated, the
```
adamalston: 'Once the handler has been instantiated' is repeated.

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,97 @@
...
+## Custom Handling via `onChange`
+
+Using the `onChange` callback, you have access to an array holding the visible
+and hidden elements. This gives you the flexibility to access the hidden items
+and render them elsewhere, such as a modal or popover.
+
+## Re-initialization guidelines
+
+The handler needs to be re-initialized whenever items are added or removed, when
+an item changes its width dynamically, or when the offset changes its width
```
adamalston: Should it be the following?

```suggestion
an item changes its width dynamically, or when the offset element's width changes
```

adamalston (approved):

annawen1 (approved):

packages/utilities/src/overflowHandler/README.md
```diff
@@ -0,0 +1,96 @@
...
+## Custom Handling via `onChange`
+
+Using the `onChange` callback, you have access to an array holding the visible
+and hidden elements. This gives you the flexibility to access the hidden items
+and render them elsewhere, such as a modal or popover.
+
+## Re-initialization guidelines
+
+The handler needs to be re-initialized whenever items are added or removed, when
+an item changes it's width dynamically, or when the offset element's width
```
heloiselui:
```suggestion
an item changes its width dynamically, or when the offset element's width
```

heloiselui (approved):
Looks good! Just a small suggestion.

riddhybansal (approved):

PR #19592, opened 2025-06-08
fix(DataTable): add support for onClick passed to getRowProps

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/18173

Added support for `onClick` passed to `getRowProps`.

### Changelog

**New**

- Added support for `onClick` passed to `getRowProps`.

#### Testing / Reviewing

```sh
yarn test packages/react
```

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
- [x] Updated documentation and storybook examples
- [x] Wrote passing tests that cover this change
- [x] Addressed any impact on accessibility (a11y)
- [x] Tested for cross-browser consistency
- [x] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #19618, opened 2025-06-11
docs(DataTable): improve render props documentation

adamalston:
No issue. https://github.com/carbon-design-system/carbon/pull/19592#pullrequestreview-2912939918

Updated `DataTable` render props documentation.

### Changelog

**Changed**

- Updated `DataTable` render props documentation.
- Fixed the table of contents.

#### Testing / Reviewing

Check that the docs do not contain spelling, grammatical, or other errors.

I considered documenting all arguments and return values for each getter, but decided against it to avoid coupling the docs too closely to the implementation. Even minor changes to an argument or return signature would require constant updates to keep the docs accurate. I want this documentation to be durable, and in a way, future-proof. If someone disagrees, I can change things.

I think this doc as a whole could benefit from an update. If someone agrees, I can look into it. Otherwise, someone else can.

## PR Checklist

### DIFF
--- a/packages/react/src/components/DataTable/DataTable.mdx
+++ b/packages/react/src/components/DataTable/DataTable.mdx
@@ -33,6 +33,7 @@ import {
 - [Selection](#selection)
   - [Programmatic selection](#programmatic-selection)
 - [Filtering](#filtering)
+  - [Multiple filters with batch updates](#multiple-filters-with-batch-updates)
 - [Batch actions](#batch-actions)
 - [Toolbar](#toolbar)
   - [Overflow Menu](#overflow-menu)
@@ -46,6 +47,7 @@ import {
     - [Prop getters](#prop-getters)
     - [Actions](#actions)
     - [State](#state)
+    - [Example](#example)
 - [Feedback](#feedback)
 
 {/* <!-- END doctoc generated TOC please keep comment here to allow auto update --> */}
@@ -244,7 +246,7 @@ current `row` in order to retrieve selection state about the given row.
 You can access all the selected rows through the `selectedRows` property passed
 into your `render` prop function.
 
-#### Programmatic selection
+### Programmatic selection
 
 You can use either of the following actions from your `render` prop function to
 update the selection status of a row:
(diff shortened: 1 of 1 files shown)


2nikhiltom (approved):

packages/react/src/components/DataTable/DataTable.mdx
```diff
@@ -491,67 +493,105 @@ naming rule
...
-| `expandRow`     | `(rowId: string) => void`     | Expand a specific row by the given `rowId`      |
-| `onInputChange` | `(event: Event) => void`      | Handle the input change of a table search field |
+| Action          | Description                               |
+| --------------- | ----------------------------------------- |
+| `expandAll`     | Toggles expansion for all rows.           |
+| `expandRow`     | Toggles expansion for a row.              |
+| `onInputChange` | Updates the table's filter/search string. |
+| `selectAll`     | Selects or deselects all rows.            |
+| `selectRow`     | Toggles selection for a row.              |
+| `sortBy`        | Sorts the table by a header key.          |
```
preetibansalui, line 541: Thanks for the changes! Any reason for removing the `Type` column? Other files still use it, so just checking for consistency.

adamalston: Yea, that's explained in the description:

> I considered documenting all arguments and return values for each getter, but decided against it to avoid coupling the docs too closely to the implementation. Even minor changes to an argument or return signature would require constant updates to keep the docs accurate. I want this documentation to be durable, and in a way, future-proof. If someone disagrees, I can change things.

To add to that, all of these props are typed, and those types should be accessible in both TypeScript and JavaScript.

Before this pull request, the docs were already out of sync with the code. That happened because keeping them in sync would have required constant updates. This change is meant to prevent that from happening again by reducing the chances of docs drifting from the actual implementation. One way to do that is to lose the type duplication in the docs since they're accessible elsewhere.

preetibansalui: Hmm, that makes sense to me. I’m just a bit concerned about consistency across other components/stories of this component. @kennylam: do you have any thoughts on this?

preetibansalui: Just realised that we are using the same mdx file in different stories of DataTable, so it will update it in all the places. Thanks Adam! 👍

preetibansalui (approved):
LGTM 👍

packages/react/src/components/DataTable/DataTable.mdx
tay1orjones: Yeah, I don't think we should hardcode types into documentation. The same datatable.mdx is used across all the different stories, so this should cover all

kennylam: I'm all for docs from annotations/jsdocs 👍

PR #19620, opened 2025-06-11
feat: added pageheader breadcrumbbar

### LINKED ISSUE #19545: PageHeader: WC - Implement Breadcrumb zone
Implement Breadcrumb zone - lay out only

- [ ] add slots icon, breadcrumb, content actions, and page actions

riddhybansal:
Closes #19545 

Added Pageheader Breadcrumbbar

### Changelog

**New**

- Added Pageheader Breadcrumbbar

#### Testing / Reviewing

Breadcrumbbar in stories in pageheader should work as intended and match with react.
## PR Checklist


packages/web-components/src/components/page-header/page-header-breadcrumb.ts
```diff
@@ -23,7 +23,7 @@ class CDSPageHeaderBreadcrumb extends LitElement {
   /**
    * Specify if breadcrumb bar has bottom border.
    */
-  @property({ reflect: true })
+  @property({ attribute: 'border', type: Boolean })
```
annawen1:
```suggestion
  @property({ reflect: true, type: Boolean })
```
No need to set the `attribute` property since the name is the same. We do need to have the `reflect` property so that the `border` attribute is reflected in the component  can be picked up by the styles.

packages/web-components/src/components/page-header/page-header.scss
```diff
@@ -31,9 +31,39 @@ $css--plex: true !default;
...
+    @extend .#{$prefix}--page-header__breadcrumb__icon;
+  }
+
+  ::slotted([slot='content-actions']) {
+    @extend .#{$prefix}--page-header__breadcrumb__content-actions;
+  }
 }
 
-:host(#{$prefix}-page-header-breadcrumb[border='true']) {
+:host(#{$prefix}-page-header-breadcrumb[border]) {
```
annawen1:
```suggestion
:host(#{$prefix}-page-header-breadcrumb[border='true']) {
```

This is needs to be set to `border='true'` as we need to be able to detect when the border is set to false (see the `Content with Hero Image` story where the breadcrumb border is set to false)

packages/web-components/src/components/page-header/page-header.stories.ts
```diff
@@ -58,7 +63,29 @@ const tags = [
 export const Default = {
   render: () => html`
     <cds-page-header>
-      <cds-page-header-breadcrumb>BreadcrumbBar</cds-page-header-breadcrumb>
+      <cds-page-header-breadcrumb border>
```
annawen1:
```suggestion
      <cds-page-header-breadcrumb>
```
since `border` default value is true, no need to apply here

packages/web-components/src/components/page-header/page-header.stories.ts
```diff
@@ -158,7 +223,17 @@ export const ContentWithHeroImage = {
     <cds-page-header>
       <div class="cds--css-grid">
         <div class="cds--sm:col-span-4 cds--md:col-span-4 cds--lg:col-span-8 cds--css-grid-column">
-          <cds-page-header-breadcrumb within-grid border=false>BreadcrumbBar</cds-page-header-breadcrumb>
+          <cds-page-header-breadcrumb ?border="${false}">
```
annawen1:
```suggestion
          <cds-page-header-breadcrumb border=false>
```

Since `border` is by default true, even if it isn't provided here, it will still render the border styles. Whenever attributes are set to true by default it's a bit tricky to shut off which is why I went with using `:host(#{$prefix}-page-header-breadcrumb[border='true']` in the styles

packages/web-components/src/components/page-header/page-header-breadcrumb.ts
```diff
@@ -33,20 +33,101 @@ class CDSPageHeaderBreadcrumb extends LitElement {
...
+    console.log(border);
     const gridClasses = classMap({
-      [`${prefix}--grid`]: !withinGrid,
+      [`${prefix}--css-grid`]: !withinGrid,
       [`${prefix}--subgrid ${prefix}--subgrid--wide`]: withinGrid,
     });
 
+    const breadcrumbBarClasses = classMap({
+      [`${prefix}--page-header__breadcrumb-bar`]: true,
+      [`${prefix}--page-header__breadcrumb-bar-border`]: border,
```
annawen1:
```suggestion
```
You already have a selector for the border styles based off the attribute so setting this class would be redundant

packages/web-components/src/components/page-header/page-header-breadcrumb.ts
```diff
@@ -33,20 +33,101 @@ class CDSPageHeaderBreadcrumb extends LitElement {
...
+      [`${prefix}--page-header__breadcrumb__actions-flush`]: pageActionsFlush,
+    });
+
+    const contentActionsClasses = classMap({
+      [`${prefix}--page-header__breadcrumb__content-actions`]:
+        !contentActionsFlush,
+    });
```
annawen1:
```suggestion
    const contentActionsClasses = classMap({
      [`${prefix}--page-header__breadcrumb__content-actions`]:
        !contentActionsFlush,
    });
```

I'm thinking we can probably removed this? Since you are applying the styles via the `::slotted` selector?

packages/web-components/src/components/page-header/page-header-breadcrumb.ts
```diff
@@ -33,20 +33,101 @@ class CDSPageHeaderBreadcrumb extends LitElement {
...
+                <slot name="icon"></slot>
+                <slot></slot>
+              </div>
+              <div class="${prefix}--page-header__breadcrumb__actions">
+                <div class="${contentActionsClasses}">
+                  <slot
+                    name="content-actions"
+                    @slotchange=${handleContentActionsSlotChange}>
+                  </slot>
+                </div>
```
annawen1:
```suggestion
                  <slot
                    name="content-actions"
                    @slotchange=${handleContentActionsSlotChange}>
                  </slot>
```

From my comment above, you might be able to remove the wrapper div?

packages/web-components/src/components/page-header/page-header.scss
```diff
@@ -31,9 +31,39 @@ $css--plex: true !default;
...
+
+  .#{$prefix}--page-header__breadcrumb__actions {
+    @extend .#{$prefix}--page-header__breadcrumb__actions;
+
+    max-inline-size: 50%;
+  }
+
+  .#{$prefix}--page-header__breadcrumb__content-actions {
+    @extend .#{$prefix}--page-header__breadcrumb__content-actions;
+  }
```
annawen1:
```suggestion
```
You already have this below
```
::slotted([slot='content-actions']) {
    @extend .#{$prefix}--page-header__breadcrumb__content-actions;
  }
  ```
  so I don't think this is needed here

**Comment by riddhybansal** [comment]
The pr was in draft , I needed to refactor !! But now its up for review !!

File: packages/web-components/src/components/page-header/page-header-breadcrumb.ts [hunk 9]
```diff
@@ -33,20 +33,92 @@ class CDSPageHeaderBreadcrumb extends LitElement {
...
+    const {
+      border,
+      withinGrid,
+      pageActionsFlush,
+      _hasContentActions: hasContentActions,
+      _hasPageActions: hasPageActions,
+      _handleContentActionsSlotChange: handleContentActionsSlotChange,
+      _handlePageActionsSlotChange: handlePageActionsSlotChange,
+    } = this;
+    console.log(border);
```
heloiselui:
```suggestion

```

packages/web-components/src/components/page-header/page-header.stories.ts
```diff
@@ -14,13 +14,18 @@ import '../tag/index';
...
+import '../breadcrumb/index';
 import image1 from './_story-assets/2x1.jpg';
 import image2 from './_story-assets/3x2.jpg';
 import { breakpoints } from '@carbon/layout';
 
 import Add16 from '@carbon/icons/lib/add/16.js';
 import Analytics16 from '@carbon/icons/lib/analytics/16.js';
 import Bee32 from '@carbon/icons/lib/bee/32.js';
+import Bee16 from '@carbon/icons/lib/bee/16.js';
+import Activity16 from '@carbon/icons/lib/activity/16.js';
```
heloiselui, line 25: Are we going to use this?

packages/web-components/src/components/page-header/page-header-breadcrumb.ts
```diff
@@ -33,20 +33,92 @@ class CDSPageHeaderBreadcrumb extends LitElement {
...
+   * Set to `true` if page actions should be flush (no padding)
+   */
+  @property({ attribute: 'page-actions-flush', type: Boolean })
+  pageActionsFlush = false;
+
+  /**
+   * Set to `true` if content actions should be flush (no padding)
+   */
+  @property({ attribute: 'content-actions-flush', type: Boolean })
+  contentActionsFlush = false;
```
heloiselui, line 46: Do we plan to use `contentActionsFlush`? I couldn't find it used in the render logic.

heloiselui (approved):
Looks great! I left a few comments. Also, we’re removing the `license` headers from the web components, so if you could take the opportunity to remove them here as well, that would be great.

packages/web-components/src/components/page-header/page-header.stories.ts
riddhybansal: Yes they are being used !!

annawen1 (approved):

packages/web-components/src/components/page-header/page-header-breadcrumb.ts
annawen1: It's not used in the render logic, but it is used for the styles so I like to keep these properties here just so people know it exists

PR #19626, opened 2025-06-11
fix: Refactored the `Treeview` component to use contexts to communicate with `TreeNode`

### LINKED ISSUE #19522: [Bug]: TreeNode cannot be used from another module, must be in same module as TreeView
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.83

### React version

19

### Description

I am trying to manage TreeNode from my own custom module with it's own state. So I have a component with a TreeView and another component that I want to contain my TreeNode(s). This kind of component nesting runs into trouble with the way the selected tree nodes are maintained and accessed during render. 

If all the TreeNodes are in the same component as the TreeView, everything is fine, but when I put the tree node in another file/component, i get the following error which stops the render...

```
@carbon_react.js?v=ea02a586:108375 Uncaught TypeError: Cannot read properties of undefined (reading 'includes')
    at TreeNode (@carbon_react.js?v=ea02a586:108375:31)
```

This is the following source line

`const isSelected = selected.includes(id);`

I can work around this by adding the following to my component `MyComponent.displayName = 'TreeNode';` to make the TreeNode code think its dealing with a TreeNode directly but that feels wrong.

### Reproduction/example

https://stackblitz.com/edit/vitejs-vite-1mfgqbhw?file=src%2FApp.tsx

### Steps to reproduce

Have a TreeView with a TreeNode that is handled in another file. Simply try to render the TreeView (see stackblitz).

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #19522

Refactored `TreeView` and `TreeNode` to use the Context API instead of `React.cloneElement`. This resolves the issue where wrapping a `TreeNode` in another component would break the tree's state management. This change makes component composition robust and reliable.

### Changelog

**New**

- Added `TreeContext.ts` to house `TreeContext` and `DepthContext`, preventing circular dependencies.

**Changed**

- `TreeView` now acts as a provider for `TreeContext` (for tree-wide state like `active`, `selected`, and handlers) and `DepthContext` (for node depth).
- `TreeNode` now consumes state and handlers from `TreeContext` and its nesting level from `DepthContext`. It also re-provides `DepthContext` to its children with an incremented value.
- Corrected `TreeNode`'s `tabIndex` logic to ensure `disabled` nodes do not have a `tabIndex` attribute, fixing an accessibility regression.
- Restored `onFocus` and `onBlur` event handling on `TreeNode` to ensure local handlers can be attached.

**Removed**

- Removed the `enhanceTreeNodes` recursive cloning function from both `TreeView` and `TreeNode`.
- Removed the `onNodeFocusEvent` prop from `TreeNode` as it is no longer needed with the new architecture.

#### Testing / Reviewing

To verify this fix, a reviewer should:

1.  Run the existing test suite. All tests for `TreeView` and `TreeNode` should now pass.
2.  Inspect `TreeNode-test.js` to see the new `renderWithProviders` test helper. This is how the component is now tested in isolation, confirming its reliance on context.
3.  Manually verify the original bug is fixed by creating a Storybook story or local example where a `TreeNode` is wrapped inside another custom component. The tree must render and function correctly without errors.
4.  Perform a general review of existing `TreeView` stories to ensure there are no visual or functional regressions in keyboard navigation, selection, or style.

## PR Checklist


packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -218,31 +226,32 @@ const useEllipsisCheck = (
...
       value,
       href,
       align = 'bottom',
       autoAlign = false,
+      // These props are destructured to be ignored, as they are now supplied by context
+      active: _active,
+      depth: _depth,
+      onTreeSelect: _onTreeSelect,
+      onNodeFocusEvent: _onNodeFocusEvent,
+      selected: _selected,
```
tay1orjones: This will be a breaking change, right? Is there any way to map these into the context and override the context values if they're provided? It's a large footgun but just trying to figure out how to reduce the impact for folks relying on these props.

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -377,27 +357,38 @@ const TreeNode = React.forwardRef<HTMLElement, TreeNodeProps>(
...
         return;
       }
-      if (matches(event, [keys.ArrowLeft, keys.ArrowRight, keys.Enter])) {
+
+      if (
+        matches(event, [
+          keys.ArrowLeft,
+          keys.ArrowRight,
+          keys.Enter,
+          keys.Space,
```
tay1orjones: isn't space meant to trigger selection?

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -279,163 +273,72 @@ describe('TreeNode - handleKeyDown', () => {
...
-    );
-
-    const treeNode = getByText('Parent Node');
-
-    fireEvent.keyDown(treeNode, { key: 'Enter' });
-    expect(handleClick).toHaveBeenCalled();
-
-    fireEvent.keyDown(treeNode, { key: 'Space' });
-    expect(handleClick).toHaveBeenCalled();
-  });
```
tay1orjones, line 410: and this would need to remain instead of being removed?

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -539,116 +534,81 @@ const TreeNode = React.forwardRef<HTMLElement, TreeNodeProps>(
...
       id,
-      onBlur: handleFocusEvent,
       onClick: handleClick,
-      onFocus: handleFocusEvent,
       onKeyDown: handleKeyDown,
       role: 'treeitem',
+      tabIndex,
+      // FIX: Wire the focus and blur events back to the element
+      onFocus: handleFocusEvent,
+      onBlur: handleFocusEvent,
```
tay1orjones: might need to be addressed

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -173,49 +175,22 @@ const TreeView: TreeViewComponent = ({
...
-  const nodesWithProps = enhanceTreeNodes(children);
+  // Set the first non-disabled node to be tabbable
+  useEffect(() => {
+    const firstNode = treeRootRef.current?.querySelector(
+      `.${prefix}--tree-node:not(.${prefix}--tree-node--disabled)`
+    );
+    if (firstNode instanceof HTMLElement) {
+      firstNode.tabIndex = 0;
+    }
+  }, [prefix]);
```
tay1orjones: Should we do this every render instead of only if `prefix` changes? If so, the useEffect could be removed

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -323,21 +297,10 @@ const TreeView: TreeViewComponent = ({
...
-          setSelected(preselected);
-        }
-        if (prespecifiedActive) {
-          setActive(prespecifiedActive);
-        }
-      }
-    }, []);
-
-  useActiveAndSelectedOnMount();
+  // Removed useActiveAndSelectedOnMount as this is now handled by the state initializers
```
tay1orjones: This makes sense in the context of this PR but I don't think it's valuable to leave in as is

```suggestion
```

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -347,20 +310,38 @@ const TreeView: TreeViewComponent = ({
       </label>
     ) : null;
 
+  // New: Define the context value
```
tay1orjones: This could use a rewording or could be deleted

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -347,20 +310,38 @@ const TreeView: TreeViewComponent = ({
...
-        aria-label={hideLabel ? label : undefined}
-        aria-labelledby={!hideLabel ? labelId : undefined}
-        aria-multiselectable={multiselect || undefined}
-        className={treeClasses}
-        onKeyDown={handleKeyDown}
-        ref={treeRootRef as unknown as React.RefObject<HTMLUListElement | null>}
-        role="tree">
-        {nodesWithProps}
-      </ul>
+      {/* New: Wrap children in Context Providers */}
```
tay1orjones: Same here, can probably remove
```suggestion
```

packages/react/src/components/TreeView/Treeview.stories.js
```diff
@@ -831,3 +831,36 @@ export const WithControlledExpansion = () => {
     </>
   );
 };
+
+const Nested = () => {
+  return <TreeNode key={21} value="Nested" label="Nested" />;
+};
+
+export const WithComplexNesting = (args) => {
```
tay1orjones, line 839: If this stays in and isn't a test story to be removed before merging, it'd be good to add this to the vrt e2e tests just in case

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
 import { render, screen, fireEvent } from '@testing-library/react';
 import TreeNode from '../TreeNode';
-import TreeView from '../TreeView';
 import { keys } from '../../../internal/keyboard';
 import { ArrowDown } from '@carbon/icons-react';
+import { TreeContext, DepthContext } from '../TreeContext';
+
+/**
+ * New: Test helper to render a component with the necessary context providers.
+ * This is crucial for testing TreeNode in isolation now that it relies on context.
```
tay1orjones:
```suggestion
 * This test helper is crucial for testing TreeNode in isolation because it
 * relies on context. It ensures the component is rendered with the necessary
 * context providers.
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
+  return render(
+    <TreeContext.Provider value={providerProps}>
+      <DepthContext.Provider value={depth}>{ui}</DepthContext.Provider>
+    </TreeContext.Provider>,
+    renderOptions
+  );
+};
 
 describe('TreeNode Component', () => {
+  // This test remains unchanged as it doesn't depend on context
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
-
-    const treeNodeElement = container.querySelector('li');
+    renderWithProviders(<TreeNode id="node1" label="Node 1" ref={mockRef} />);
+    const treeNodeElement = screen.getByRole('treeitem');
     expect(treeNodeElement).toBeInTheDocument();
-    expect(mockRef).toHaveBeenCalledWith(treeNodeElement); // Verify if ref is called
+    expect(mockRef).toHaveBeenCalledWith(treeNodeElement);
   });
 
+  // This test remains unchanged as it doesn't depend on context
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
-      <TreeNode
-        id="parent"
-        label="Parent Node"
-        depth={1}
-        active="child1"
-        disabled={false}
-        selected={['child1']}
-        onTreeSelect={() => {}}>
-        <TreeNode id="child1" label="Child Node 1" disabled={false} />
+  // UPDATED: This test now uses renderWithProviders to supply context
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
-    expect(child1).toHaveAttribute('aria-selected', 'true');
+    expect(child1).toHaveAttribute('aria-selected', 'true'); // Verified via context
     expect(child1).toHaveAttribute('tabindex', '-1');
 
     expect(child2).toHaveAttribute('id', 'child2');
     expect(child2).toHaveAttribute('aria-disabled', 'true');
     expect(child2).not.toHaveAttribute('tabindex');
   });
 
+  // UPDATED: This test now uses renderWithProviders to supply depth
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
+      { depth }
     );
-    const treeNode = getByText('Parent Node');
-    expect(treeNode).toBeInTheDocument();
+
+    // FIX: Use .closest() to find the specific styled container
+    const styledNodeContainer = getByText('Parent Node').closest(
+      `.${prefix}--tree-node__label`
+    );
+    expect(styledNodeContainer).toBeInTheDocument();
```
tay1orjones: Looks like this was addressed?

```suggestion
    const styledNodeContainer = getByText('Parent Node').closest(
      `.${prefix}--tree-node__label`
    );
    expect(styledNodeContainer).toBeInTheDocument();
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -8,125 +8,119 @@
...
+      <TreeNode id="parent" label="Parent Node" renderIcon={ArrowDown} />,
+      { depth }
     );
-    const treeNode = getByText('Parent Node');
-    expect(treeNode).toBeInTheDocument();
 
-    const offset = depth + 2 + depth * 0.5;
-
-    const treeNodeLabel = container.querySelector(
+    // FIX: Use .closest() to find the specific styled container
```
tay1orjones, line 109: resolved?

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -186,40 +180,38 @@ describe('TreeNode - handleToggleClick', () => {
 });
 
 describe('TreeNode - handleClick', () => {
-  it('should call onTreeSelect, onNodeSelect, and rest.onClick when the node is not disabled', () => {
+  // UPDATED: This test now uses renderWithProviders to supply onTreeSelect
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -186,40 +180,38 @@ describe('TreeNode - handleToggleClick', () => {
...
+      id: expect.any(String), // ID is generated, so check for type
       label: 'Test Node',
       value: undefined,
     });
-    expect(onClick).toHaveBeenCalled(); // Ensure the rest.onClick handler was triggered
+    expect(onNodeSelect).toHaveBeenCalled();
+    expect(onClick).toHaveBeenCalled();
   });
 
+  // This test remains unchanged
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -235,6 +227,7 @@ describe('TreeNode - handleClick', () => {
   });
 });
 
+// This test remains unchanged
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -254,6 +247,7 @@ it('should support specifying the href for parent nodes', () => {
   );
 });
 
+// These keyboard tests remain largely unchanged
```
tay1orjones:
```suggestion
```

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -279,163 +273,72 @@ describe('TreeNode - handleKeyDown', () => {
     expect(onClick).not.toHaveBeenCalled();
   });
 
-  it('should find the Parent Element when ArrowLeft is clicked ', () => {
```
tay1orjones, line 282: are we losing some coverage with this removal?

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
```diff
@@ -279,163 +273,72 @@ describe('TreeNode - handleKeyDown', () => {
...
-    fireEvent.keyDown(treeNode, { key: 'Enter' });
-    expect(handleClick).toHaveBeenCalled();
-
-    fireEvent.keyDown(treeNode, { key: 'Space' });
-    expect(handleClick).toHaveBeenCalled();
-  });
 });
 
+// UPDATED: These tests are simplified to only check for onFocus/onBlur
+// as onNodeFocusEvent is no longer a prop.
```
tay1orjones:
```suggestion
```

tay1orjones (changes requested):

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
Gururajj77: No, I had missed some tests while adding new tests with `renderWithProviders` and modifying existing tests

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
Gururajj77: Added this back

packages/react/src/components/TreeView/Treeview.stories.js
Gururajj77: Good idea, i'll try to create a better example or use this and add vrt tests.

packages/react/src/components/TreeView/__tests__/TreeNode-test.js
Gururajj77: Yeah this is addressed.

packages/react/src/components/TreeView/TreeNode.tsx
Gururajj77: Have added the props as a primary source now, will use the values passed by the user first, then it will look for the values in the context, this way the component supportsbackwards compatibility

packages/react/src/components/TreeView/TreeView.tsx
Gururajj77: added `children` to the dependency array to run this whenever the nodes change, as this was running only once in the beginning.

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -218,31 +226,37 @@ const useEllipsisCheck = (
...
-    const depth = propDepth as number;
-    const selected = propSelected as (string | number)[];
+    const treeContext = useContext(TreeContext);
+    const contextDepth = useContext(DepthContext);
+
+    // Prioritize direct props, and fall back to context values.
+    const depth = propDepth ?? (contextDepth !== -1 ? contextDepth : 0);
+    const active = propActive ?? treeContext?.active;
+    const selected = propSelected ?? treeContext?.selected ?? [];
+    const onTreeSelect = propOnTreeSelect ?? treeContext?.onTreeSelect;
```
Gururajj77, line 260: these will prioritize user supplied props over the context.

packages/react/src/components/TreeView/TreeView.tsx
```diff
@@ -173,49 +175,22 @@ const TreeView: TreeViewComponent = ({
...
-  const nodesWithProps = enhanceTreeNodes(children);
+  // Set the first non-disabled node to be tabbable
+  useEffect(() => {
+    const firstNode = treeRootRef.current?.querySelector(
+      `.${prefix}--tree-node:not(.${prefix}--tree-node--disabled)`
+    );
+    if (firstNode instanceof HTMLElement) {
+      firstNode.tabIndex = 0;
+    }
+  }, [children, prefix]);
```
Gururajj77, line 193: added the children here to recalculate the tabIndex whenever any nodes change.

tay1orjones (approved):

packages/react/src/components/TreeView/TreeNode.tsx
```diff
@@ -218,31 +226,38 @@ const useEllipsisCheck = (
...
       value,
       href,
       align = 'bottom',
       autoAlign = false,
+      // These props are fallback props if the TreeContext is not available or only TreeNode is used as a standalone component
+      active: propActive,
+      depth: propDepth,
+      selected: propSelected,
+      onTreeSelect: propOnTreeSelect,
+      onNodeFocusEvent,
```
tay1orjones, line 248: Could you make a new issue to track deprecating these?

riddhybansal (approved):

packages/react/src/components/TreeView/TreeNode.tsx
Gururajj77: Done, the issue: https://github.com/carbon-design-system/carbon/issues/19778

2nikhiltom (approved):

heloiselui (approved):

PR #19652, opened 2025-06-14
fix(PopoverContent): Handling CSS based on multi line in the pop over content

### LINKED ISSUE #19623: [Bug]: IconButton tooltip styles breaks for multiline tooltip content
### Package

@carbon/react

### Browser

_No response_

### Package version

v11.84.0

### React version

_No response_

### Description

The IconButton tooltip component does not render multiline content according to design specifications. The tooltip lacks proper padding and text layout

Expected Behavior
1. Tooltip should have 16px padding on all sides
2. Text should utilize the full available width

Actual Behavior
Excessive empty space on the right side (Text not filling the available width properly)
Missing top and bottom padding

Image

Image

### Reproduction/example

https://stackblitz.com/edit/github-npbpavky?file=src%2FApp.jsx&preset=node=

### Steps to reproduce

open https://stackblitz.com/edit/github-npbpavky?file=src%2FApp.jsx&preset=node=
hover on icons to see the style issues 

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

NAS-IBM:
Closes #19623 

### Changelog

**New**

- add unit test cases
- added logic to check if the popover text is single or multi line and based on it adding the class.
- added overrding css to handle multi line scenario for padding

#### Testing / Reviewing

- With the large text, popover content renders in multiple line. The content will have padding of 1rem
- With the small text, popover content renders in single line. The content will have padding of 0.125 rem on the sides. 

## PR Checklist


tay1orjones (changes requested):
The storybook was already correct because of custom styles for the stories that applied a `p-3` class/styling to the content. Could you remove that class & styling from the stories so we can see this works as it should?

Also are there other places in the codebase where we might have a similar set-up that we can remove in favor of this new built in functionality?

packages/react/src/components/Popover/__tests__/Popover-test.js
```diff
@@ -267,5 +267,35 @@ describe('Popover', () => {
...
+          </PopoverContent>
+        </Popover>
+      );
+      await setTimeout(
+        () =>
+          expect(container.firstChild.lastChild.firstChild).toHaveClass(
+            `${prefix}--popover-content--multi-line`
+          ),
+        100
+      );
```
tay1orjones: `setTimeout` shouldn't be used in tests. Use [`findBy`/`waitFor`](https://testing-library.com/docs/dom-testing-library/api-async/) instead.

NAS-IBM: Thanks for your feedback. adapted with waitFor

NAS-IBM:
@tay1orjones I have adapted the changes based on the feedbck.

heloiselui (commented):
Thanks for fixing that! Could you also check line #L619? It looks like it wasn't covered by tests.

NAS-IBM:
> Thanks for fixing that! Could you also check line #L619? It looks like it wasn't covered by tests.

@heloiselui I have adapted the code to remove the unncessary branching condition. The span element will always present when the component is rendered.

heloiselui (approved):
LGTM!

NAS-IBM:
@tay1orjones can you please review this PR?

tay1orjones (approved):

tay1orjones:
@NAS-IBM Thanks again for this! LGTM but there's a couple conflicts to resolve before this can merge

NAS-IBM:
@tay1orjones I have resolved the conflicts. Is there anything I need to process?

alisonjoseph:
This has unintentionally added a breaking change and added padding to popovers where it shouldn't. I'm opening a PR to revert.

emyarod:
since this targets multiline text content, I think this change should target the tooltip specifically rather than the generic popover component

Issue #18991, opened 2025-04-01
Reenable ESLint


adamalston:
ESLint is disabled. Reenable it.

https://github.com/carbon-design-system/carbon/blob/426f1c7734b171dd40b4394a0450e612fa0b385e/eslint.config.mjs#L2-L22

Tasks:

adamalston:
There are actual functional issues that are not being caught because ESLint is disabled. See https://github.com/carbon-design-system/carbon/pull/18971/files#diff-2488ecfa65ff626a3f7ee72ba1295eccb46aac44702cd591f7a5943fd23a8cafR36-R39.

adamalston:
Another issue: https://github.com/carbon-design-system/carbon/blob/c6aec71f71f0031a82ffbb16ac844a2e993d36bf/packages/react/src/components/Tag/Tag.tsx#L174-L175

The intention is to check if `type` is one of the keys in the `TYPES` object. However, using the `in` operator on `Object.keys(TYPES)` doesn’t work as expected because `Object.keys(TYPES)` returns an array of strings, and the `in` operator checks for property keys (which would be numeric indices in this array).

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_operators#in for more info.

adamalston:
`typeof` anything will never equal `null` so these conditions are invalid:

https://github.com/carbon-design-system/carbon/blob/09ef316b1854134af4f697f598e4f902630b90f6/packages/react/src/components/Slider/Slider.tsx#L1328-L1330

https://github.com/carbon-design-system/carbon/blob/09ef316b1854134af4f697f598e4f902630b90f6/packages/react/src/components/Slider/Slider.tsx#L1333-L1335

PR #19658, opened 2025-06-15
feat: add typescript-eslint

### LINKED ISSUE #19008: Install `typescript-eslint`

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/19008

Added `typescript-eslint`.

### Changelog

**New**

- Added `typescript-eslint`.

#### Testing / Reviewing

I left a `TODO` with a question. Also, I'm curious to hear if anyone thinks `packages/themes/src/tokens/Token.js` should be converted from a class to a function.

You may be wondering why this pull request is so small. It's because I haven’t re-enabled TypeScript file processing yet. I plan to open a separate pull request for that. Locally, doing so results in https://github.com/carbon-design-system/carbon/issues/18991#issuecomment-2770412818 and a lot of other problems getting caught which is what we want.

https://github.com/carbon-design-system/carbon/blob/6d6150ce59fbbf0bb1e5ccb7751a84e58643ab89/eslint.config.mjs#L128-L133

```sh
yarn lint
```

## PR Checklist


packages/cli/src/cli.js
```diff
@@ -16,7 +16,10 @@ async function main({ argv }) {
     .version(packageJson.version)
     .usage('Usage: $0 [options]');
 
-  cli
+  // TODO [@carbon-design-system/monorepo-reviewers]: What's the purpose of this
+  // code, especially the `argv` expression at the end? The only reason I had to
+  // add the `void` was because of that expression.
+  void cli
```
kennylam: This file is the main entry point for various project operations, including [builds](https://github.com/carbon-design-system/carbon/blob/main/packages/layout/package.json#L29) (via `src/commands/bundle`, for example), releases, and publishing (CI checks, updating changelogs, etc). Pretty much everything gets piped through `@carbon/cli`.

[argv](https://yargs.js.org/docs/#api-reference) are the options passed to the CLI to parse. Node includes 2 elements by default in `process.argv`, hence the `slice()` to skip those.

I'm not sure why `cli` is broken into 2 blocks, that's before my time, but one chain would also work. And finally, the dangling `argv` can be removed since it's already being returned in `parse()`. I tested `build` and `ci-check` locally and both run fine. This would let you do something like

```js
async function main({ argv }) {
  cli
    .scriptName(packageJson.name)
    .version(packageJson.version)
    .usage('Usage: $0 [options]')
    .commandDir('commands')
    ...
    .parse(argv.slice(2));
  }
```

kennylam (approved):
LGTM!

annawen1 (approved):

Issue #19177, opened 2025-04-19
Export Wrappers: Convert `DataTable` to a functional component


adamalston:
Three parts:

1. Add all of the missing types and clean up the code.
2. Convert the remaining code to a functional component.
3. Address all of the `TODO`s in the component resulting from the conversion.

PR #19659, opened 2025-06-15
refactor(DataTable): convert from class to functional component

adamalston:
Partially addresses https://github.com/carbon-design-system/carbon/issues/19177

Converted `DataTable` from a class to a functional component.

### Changelog

**Changed**

- Converted `DataTable` from a class to a functional component.

#### Testing / Reviewing

I left some `TODO`s with questions.

I planned to address the `TODO`s that are not tagged to the Carbon reviewers in a follow-up pull request, unless someone thinks they should be addressed now. My goal with these changes, given the nature of this component, was to leave as much unchanged as possible. I've added a third part to https://github.com/carbon-design-system/carbon/issues/19177.

Reference commits: https://github.com/carbon-design-system/carbon/compare/6d6150ce59fbbf0bb1e5ccb7751a84e58643ab89...8a1a147c241b13e38c4b2d90d2d1bbd28c1d5d2a

## PR Checklist


Gururajj77 (approved):
Looks good, thanks for the refactor @adamalston

packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -248,6 +253,8 @@ export interface DataTableProps<RowType, ColTypes extends any[]>
   locale?: string;
   overflowMenuOnHover?: boolean;
   radio?: boolean;
+  // TODO [@carbon-design-system/monorepo-reviewers]: Any reason to keep this
+  // prop since it matches `children`? I'm wondering if it should be deprecated.
   render?: (
     renderProps: DataTableRenderProps<RowType, ColTypes>
   ) => ReactElement;
```
tay1orjones: I think it could be deprecated. They'd pass a function as a child instead, right? The message could be something like that - "pass a function as a child instead". Maybe a link to something helpful like [this](https://www.patterns.dev/react/render-props-pattern/#:~:text=Children%20as%20a%20function). The two usages in storybook stories could be updated as well:

https://github.com/carbon-design-system/carbon/blob/9087ae49eac992fc01224c62c76770957c59aef5/packages/react/src/components/DataTable/stories/dynamic-content/DataTable-dynamic-content.stories.js#L166-L179

https://github.com/carbon-design-system/carbon/blob/9087ae49eac992fc01224c62c76770957c59aef5/packages/react/src/components/DataTable/stories/expansion/DataTable-expansion.stories.js#L117-L126

packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -820,89 +805,88 @@ class DataTable<RowType, ColTypes extends any[]> extends Component<
...
-}
 
+  return null;
+};
+
+// TODO [@carbon-design-system/monorepo-reviewers]: Should these namespaced
+// components and properties be removed in a future release? It happened for
+// v11:
+// https://github.com/carbon-design-system/carbon/blob/f574ad0e46ed2f7d84ba50c8561bd9cb20ffa1b3/packages/react/docs/migration/11.x-namespaced-exports.md#changes
+DataTable.translationKeys = Object.values(translationKeys);
```
tay1orjones: For large compound components like this we'll continue to use namespacing for ease of use. The ones removed in v11 were just one-offs like `TextInput.PasswordInput`.

packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -697,48 +670,56 @@ class DataTable<RowType, ColTypes extends any[]> extends Component<
...
-      const row = state.rowsById[rowId];
-      if (this.props.radio) {
+  const handleOnSelectRow = (rowId: string) => () => {
+    setState((prev) => {
+      const row = prev.rowsById[rowId];
+      if (radio) {
+        // TODO:
+        // 1. Should the `reduce` be typed with `<Record<string,
+        // DataTableRow<ColTypes>>>`?
+        // 2. Add better parameter names. What is `p` and `c`?
```
tay1orjones: Probably could just unify with mdn docs and other usages in this file to rename `p` either `acc` or `accumulator`

packages/react/src/components/DataTable/DataTable.tsx
```diff
@@ -286,6 +293,10 @@ interface DataTableState<ColTypes extends any[]> {
 }
 
 /**
+ * TODO [@carbon-design-system/monorepo-reviewers]: Is everything in this
+ * comment still accurate? If not, figure out if it should be updated or
+ * deleted.
```
tay1orjones: Yeah, I think so, but it's a bit convoluted. Trying to parse it out:

* "direct updates to the state" seems like just another way of saying it uses updater functions in setState calls. 
* "normalizing and denormalizing when rendering" is basically the various usages of `reduce`? 
* "accessible through look-up by id" - I assume that's meaning events use `rowId`, `handleOnSelectRow` for example. 
* "updating the state of the single entity will cascade updates to the consumer" aka "we call consumer event handlers on internal state changes"

tay1orjones (approved):

packages/react/src/components/DataTable/DataTable.tsx
adamalston: I can deprecate it in part three of https://github.com/carbon-design-system/carbon/issues/19177 unless you want it deprecated in this pull request.

tay1orjones: Sounds good, no need to put it in here

PR #19694, opened 2025-06-17
ci(chromatic): add chromatic workflow for react storybook

tay1orjones:
This adds a new chromatic github actions steps for testing. Once merged we should see a new CI check on PRs for chromatic.

I did not configure TurboSnap yet, it requires a bit more advanced setup and fine-tuning that should be done after we confirm this workflow works as we expect.

### Changelog

**New**

- chromatic steps to ci.yml

#### Testing / Reviewing

- A new ci status check should be present
- Chromatic status checks should be present
- Deploy preview should be linked in the PR timeline and in the deployments section above the status checks

## PR Checklist


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,75 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  react-storybook:
+    if:
+      ${{ github.repository == 'carbon-design-system/carbon' &&
+      github.event.pull_request.draft == false }}
```
adamalston: Will `github.event.pull_request` exist on a `push` event?

heloiselui (approved):

.github/workflows/chromatic.yml
kennylam: Yep.

kennylam: Or more specifically, it exists here because the `pull_request` event is also in the workflow.

kennylam (approved):

2nikhiltom (approved):

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+          # 👇 Tells the checkout which commit hash to reference
+          # https://www.chromatic.com/docs/github-actions/#recommended-configuration-for-build-events
+          ref: ${{ github.event.pull_request.head.ref }}
```
repomadness: @tay1orjones I believe you're hitting an edge case due to the forked PR.  I'm still in the process of reviewing to determine best recommendations for your use case, but wanted to confirm that I do think the current `ref` is an issue as you suspected.  You'll still need to have it set, but likely a bit differently and with some additional considerations.

repomadness: @tay1orjones would you mind confirming whether or not you're able to have success with this workflow by:
- updating the `ref` to use: `github.event.pull_request.head.sha`
- update your Chromatic step to include all three git env vars (will include below)

```
  env:
    CHROMATIC_BRANCH: ${{ github.event.pull_request.head.ref || github.ref_name }}
    CHROMATIC_SHA: ${{ github.event.pull_request.head.sha || github.ref }}
    CHROMATIC_SLUG: ${{ github.repository }}
```

Using `pull_request.head.sha` should make sure your checkout step doesn't get stuck when the PR is from an external fork.  It does end up leaving you in a detached HEAD state, which could cause issues for TurboSnap calculations, but Chromatic should be able to properly calculate if you include all three of the git env vars in the Chromatic step.  You'll want to note that all three need to be included, otherwise Chromatic will ignore the variables.

PR #19695, opened 2025-06-17
fix(modal): Modal focus trap in Firefox

### LINKED ISSUE #17113: [a11y]: Passive modal not trapping focus in Firefox
### Package

@carbon/react

### Browser

Firefox

### Operating System

MacOS

### Package version

1.63.1

### React version

v18.2.0

Manual

### Assistive technology

_No response_

### Description

The focus is not trapping in the Passive Modal in Firefox. The focus is went outside when pressing the Tab key.

https://github.com/user-attachments/assets/47c63fbc-369a-4d86-8a36-ffc1c95888ca

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-modal--passive-modal

### Steps to reproduce

1. Goto the Passive Modal Storybook
2. Open the Passive Modal
3. Hit the Tab key

### Suggested Severity

### Code of Conduct

warrenmblood:
Closes #17113

Fixes Modal focus not being trapped in Firefox

### Changelog

**New**

- `wrapFocus` called inside a `setTimeout` to ensure focus is set after any browser default focus behavior
- `wrapFocusTimeout` ref added to store setTimeout ID so that it can be cleared

**Changed**

- added docs, a11y to my profile in contributor list

#### Testing / Reviewing

- Open React storybook in Firefox
- Go to "Passive Modal" story and test that focus is trapped inside modal while pressing TAB or SHIFT+TAB

## PR Checklist


packages/react/src/components/Modal/Modal.tsx
```diff
@@ -481,6 +486,15 @@ const Modal = React.forwardRef(function Modal(
...
   }
 
+  useEffect(() => {
+    // clear timeout for wrapFocus
+    return () => {
+      if (wrapFocusTimeout.current) {
+        clearTimeout(wrapFocusTimeout.current);
+      }
+    };
+  }, []);
```
adamalston: After making the changes in the diff and checking the console, I noticed that 'Clearing wrapFocus timeout in Modal' is never logged when visiting http://localhost:3000/?path=/story/components-modal--passive-modal and holding down <kbd>Tab</kbd>. Is that expected?

It makes me think that the cleanup function for `wrapFocusTimeout` isn't being triggered as designed, unless I'm misunderstanding. I was expecting the previous timeout to be cleared before the new one was set. Even after closing and reopening the modal, that log statement doesn't appear. I only see it when clicking the "Remount the component" button.

Steps:

1. Make the changes in the diff below.
2. Go to http://localhost:3000/?path=/story/components-modal--passive-modal.
3. Press <kbd>Tab</kbd> 5 times.
4. Notice that `'Wrapping focus in Modal'` is logged twice per <kbd>Tab</kbd> press (10 times total which seems problematic on its own).
5. Close the modal.
6. Reopen the modal.
7. Notice that 'Clearing wrapFocus timeout in Modal' is never logged.
8. Click the **Remount the component** button.
9. Now the 'Clearing wrapFocus timeout in Modal' message is logged once, despite 'Wrapping focus in Modal' having been logged multiple times (2 × per <kbd>Tab</kbd> press).

```diff
diff --git a/packages/react/src/components/Modal/Modal.tsx b/packages/react/src/components/Modal/Modal.tsx
index 684692d32..a1815c1d1 100644
--- a/packages/react/src/components/Modal/Modal.tsx
+++ b/packages/react/src/components/Modal/Modal.tsx
@@ -383,7 +383,8 @@ const Modal = React.forwardRef(function Modal(
       const { current: endTrapNode } = endTrap;
       // use setTimeout to ensure focus is set after all browser default focus behavior. Fixes issue of
       // focus not wrapping in Firefox
-      wrapFocusTimeout.current = setTimeout(() =>
+      wrapFocusTimeout.current = setTimeout(() => {
+        console.log('Wrapping focus in Modal');
         wrapFocus({
           bodyNode,
           startTrapNode,
@@ -391,8 +392,8 @@ const Modal = React.forwardRef(function Modal(
           currentActiveNode,
           oldActiveNode,
           selectorsFloatingMenus,
-        })
-      );
+        });
+      });
     }
 
     // Adjust scroll if needed so that element with focus is not obscured by gradient
@@ -490,6 +491,7 @@ const Modal = React.forwardRef(function Modal(
     // clear timeout for wrapFocus
     return () => {
       if (wrapFocusTimeout.current) {
+        console.log('Clearing wrapFocus timeout in Modal');
         clearTimeout(wrapFocusTimeout.current);
       }
     };
```

packages/react/src/components/Modal/Modal.tsx
adamalston, line 1: Should there be a condition somewhere in here to limit this functionality to passive modals?

adamalston:
> * don't know how to test cross-browser

Playright can run tests on Firefox: https://playwright.dev/docs/browsers#run-tests-on-different-browsers. Not sure whether that's an option here.

packages/react/src/components/Modal/Modal.tsx
warrenmblood: Thanks for the review! I made changes that move the `clearTimeout` inside the `setTimeout` after `wrapFocus` is called. When I add your console messages, 'Wrapping focus in Modal' and 'Clearing wrapFocus timeout in Modal' are each logged once per time `Tab` is pressed

packages/react/src/components/Modal/Modal.tsx
warrenmblood: I was able to replicate this issue with the other modal types when the close button is the only focusable element. I think this functionality should apply to all modals

warrenmblood:
> > * don't know how to test cross-browser
> 
> Playright can run tests on Firefox: https://playwright.dev/docs/browsers#run-tests-on-different-browsers. Not sure whether that's an option here.

The playwright tests are currently configured to run in Chrome in the playwright.config.js file. I found an existing Modal test `'@avt-keyboard-nav passive modal'` checks this scenario for focus trapping and fails without these PR changes when I change the playwright config to Firefox.

packages/react/src/components/Modal/Modal.tsx
adamalston: The description currently says, 'Fixes Passive Modal focus not being trapped in Firefox.' Should 'Passive' be removed and additional information added clarifying the scope of these changes?

adamalston:
> The playwright tests are currently configured to run in Chrome in the playwright.config.js file. I found an existing Modal test `'@avt-keyboard-nav passive modal'` checks this scenario for focus trapping and fails without these PR changes when I change the playwright config to Firefox.

Given that, I wonder if running all Playwright tests on Firefox a few times might reveal other issues? It's out of scope for this pull request, but it seems like something worth doing, especially if simply running the tests in Firefox could have caught this issue before it reached a user.

tay1orjones (approved):

riddhybansal (approved):

PR #19728, opened 2025-06-20
chore(deps): force brace-expansion to resolve to ^2.0.2

tay1orjones:
Should resolve https://github.com/carbon-design-system/mend-issues/issues/29

@heloiselui and I dug through the `yarn why` of this and found that forcing brace-expansion to resolve to 2.0.2 shouldn't have compat issues (it appears breaking changes were only changes to the supported node versions).

### Changelog

**Changed**

- update root package.json with a `resolutions` key for brace-expansion

#### Testing / Reviewing

- build should pass, style linting should work as expected when writing something that shouldn't pass `stylelint-no-unsupported-browser-features`

## PR Checklist


annawen1 (approved):

heloiselui (approved):

tay1orjones (commented):
If we see issues with this one, brace-expansion does have a 1.11.12 w/ the fix that we could use to resolve all usages of v1.x instead of forcing them all to use v2.x

PR #19736, opened 2025-06-23
Make aria-label and aria-labelledby optional in UIShell's Header

### LINKED ISSUE #19725: [Bug]: "The ARIA attributes "aria-label" are not valid for the element <header> with implicit ARIA role "generic", in the UI Shell's header component
### Package

@carbon/react

### Browser

Chrome

### Package version

v11

### React version

v19

### Description

Context from this [slack thread](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1750333381614129).

The typescript interface doesn't have the enforcement of mandating either `aria-label` or `aria-labelledby`, like how the `propTypes` of the same component has.

### Reproduction/example

https://stackblitz.com/edit/github-sfqmlu1x?file=src%2FApp.jsx&preset=node=

### Steps to reproduce

Open the stackblitz and see the console error that says either of them is requried.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

mariyageorge01:
Closes #19725

This PR makes `aria-label` and `aria-labelledby` optional in UIShell's `Header`.

### Changelog

**New**

- The props: aria-label and aria-labelledby are now optional in UIShell - Header component

**Changed**

- Before this PR, UIShell Header needed either aria-label or aria-labelledby. With the changes in this PR, you can exclude aria-label and aria-labelledby. Excluding them will not emit any console warnings.

**Removed**

- NA

#### Testing / Reviewing

The error in dev console cannot be reproduced in storybook by default. I tested this in my application by updating the node_modules with the changes in this PR and it fixed the issue I was experiencing in https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1750333381614129.

## PR Checklist

- [x] Wrote passing tests that cover this change - updated existing test

### DIFF
--- a/packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
+++ b/packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
@@ -164,13 +164,6 @@ export default {
       hideNoControlsWarning: true,
     },
   },
-  argTypes: {
-    className: {
-      table: {
-        disable: true,
-      },
-    },
-  },
 };
 
 export const HeaderWNavigation = () => (
--- a/packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
+++ b/packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
@@ -164,6 +164,13 @@ export default {
       hideNoControlsWarning: true,
     },
   },
+  argTypes: {
+    className: {
+      table: {
+        disable: true,
+      },
+    },
+  },
 };
 
 export const HeaderWNavigation = () => (
--- a/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
+++ b/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
@@ -4201,8 +4201,12 @@ Map {
   },
   "Header" => Object {
     "propTypes": Object {
-      "aria-label": [Function],
-      "aria-labelledby": [Function],
+      "aria-label": Object {
+        "type": "string",
+      },
+      "aria-labelledby": Object {
+        "type": "string",
+      },
       "className": Object {
         "type": "string",
       },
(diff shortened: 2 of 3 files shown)


packages/react/src/components/UIShell/UIShell.HeaderBase.stories.js
```diff
@@ -164,13 +164,6 @@ export default {
       hideNoControlsWarning: true,
     },
   },
-  argTypes: {
-    className: {
-      table: {
-        disable: true,
-      },
-    },
-  },
```
mariyageorge01: I don't know why this was added. In my understanding, we don't need this.

annawen1: Hi @mariyageorge01, that section is to disable the `className` control from appearing in our storybook. We deemed it unnecessary as adjusting the `className` prop doesn't result in any visual changes. Could you add that section back in?

Screenshot 2025-06-26 at 12 35 26 PM

mariyageorge01: Ahh... so the className prop cannot be used to add styles to the Header. But, to ensure backward compatibility, you cannot remove it from the codebase. So, it was decided to hide it on storybook. Is my understanding right?

mariyageorge01: I have made the change you requested -> added back the section in storybook that hides the className prop.

annawen1: > Ahh... so the className prop cannot be used to add styles to the Header. But, to ensure backward compatibility, you cannot remove it from the codebase. So, it was decided to hide it on storybook. Is my understanding right?

This section of code is specific to storybook only, it does not affect the component - you can use the `className` prop. We decided to not show it in storybook only as changing the `className` value does not result in any visual changes in the storybook

mariyageorge01:
@annawen1 @tay1orjones @riddhybansal Can I get a review on this PR please 🙂?

tay1orjones (approved):

heloiselui (approved):

PR #19742, opened 2025-06-23
fix(numberinput): update onChange value type to Number for type="text"

### LINKED ISSUE #19630: [Feature Request]: NumberInput type="text" functional improvements
### The problem

We've received a few points of feedback since introducing type="text" to NumberInput. These items need fixed.

### The solution

- [ ] The event value from `onChange` and `onBlur` should be a Number, not the locale-formatted string. 
  - If for some reason the consumer wants the locale-formatted string, they'll need to look at the input event.target.value directly
- [ ] When focus is on a stepper button and focus moves away from that stepper button, `onBlur` is not called but should be
- [ ] When the input field is empty, clicking a stepper should reset to `0` or move a `step`, also taking into account `min` and `max`.
  - min/max thoughts: Incrementing from an empty input always starts at `min`, decrementing always starts at `max`. If no `min` or `max` it starts at 0

### Examples

* https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1749474613614339
* https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1749478187315529
* https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1749648029085619

### Package

@carbon/react

tay1orjones:
Closes #19630

Based on user feedback, this updates the `NumberInput` API and logic when `type="text"` in the following ways:

### Changelog

**New**

- Add `onBlur` to NumberInput API

**Changed**

- Update `defaultValue` to default to `0` for type="number" and `NaN` for type="text"
- Harden existing logic to guard against `NaN` values
- Update the event value in `onChange` to be a `Number`, not the locale-formatted `string`
- Ensure `onBlur` is called when the input is blurred, and also when stepper buttons are blurred
- Change stepper logic for when the input is empty:
  - incrementing always starts at min
  - decrementing always starts at max
  - If no min or max it starts at 0
- Storybook docs for the stepper logic
- Added/updated tests

#### Testing / Reviewing

- View the "with type of text ..." stories and ensure the above items work as intended
- The "controlled" example story and helper functions gives you a way to test the functionality is consistent when using `value` over `defaultValue`
- The stories log out event loop values in the console so you can verify their type in addition to their value

> [!WARNING]  
> "Controlled" story and associated story helper/logging functions need to be removed before merge

## PR Checklist


maradwan26 (approved):
Works as intended, LGTM !

kennylam (approved):

PR #19764, opened 2025-06-24
fix(modal): move initial focus to whole modal for passive modal

### LINKED ISSUE #16880: [Bug]: feature flag `OverflowMenu` adds focus on the close button when user clicks on the icon
### Package

@carbon/react

### Browser

Chrome

### Package version

v11.60.0

### React version

v18

### Description

When inside a modal, the v12 OverflowMenu (feature flag enabled) focuses on the modal's close button when user clicks on the overflowmenu icon. This seems to be a bug only on the v12 as the OverflowMenu without feature flag enabled doesn't focus on the close button.

### Reproduction/example

https://stackblitz.com/edit/github-rbge1w-xtdlws?file=src%2FApp.tsx

### Steps to reproduce

- Make sure the feature flag is enabled on the OverflowMenu ( use the stackblitz above) 
- click on the Hello World button
- inside the modal, click on the overflowmenu icon, the menu opens up and you can see the modal's close button is focused.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

katnow:
Avoids opening close button tooltip on passive modal open

Closes https://github.com/carbon-design-system/carbon/issues/16880

The only button in the passive modal is the close button. Once we open it the close button gets focused and we can see the "Close" tooltip open. We don't want to focus on close button in this case, so I moved the focus to modal container. After pressing the tab once the close button gets highlighted, which is expected behaviour.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

## PR Checklist

### DIFF
--- a/packages/react/src/components/Modal/Modal-test.js
+++ b/packages/react/src/components/Modal/Modal-test.js
@@ -893,4 +893,19 @@ describe('events', () => {
     await keyboard('{Enter}');
     expect(onRequestSubmit).toHaveBeenCalledTimes(1);
   });
+
+  it('should set focus on the modal container when the modal is passive', async () => {
+    render(
+      <Modal data-testid="modal-10" passiveModal open>
+        <p>
+          Custom domains direct requests for your apps in this Cloud Foundry
+          organization to a URL that you own. A custom domain can be a shared
+          domain, a shared subdomain, or a shared domain and host.
+        </p>
+      </Modal>
+    );
+
+    const modal = screen.getByRole('dialog');
+    expect(modal).toHaveFocus();
+  });
 });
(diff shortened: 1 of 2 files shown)


tay1orjones (changes requested):
Hey thanks for taking a look at this! Could you elaborate on how this solves the issue you linked? That describes a specific interaction between the feature flagged overflowmenu implementation being used inside a modal. I don't see a unit test for that or a test story demonstrating it. 

Also do you have a source for focusing the modal container being the expected behavior? This doesn't seem right to me - we'd always want to focus an interactive element.

katnow:
@tay1orjones I think I misunderstood the issue at the beginning, so my solution changes something that works as expected, I thought it was a bug. The actual issue described I was not able to reproduce. The focus worked for me as expected, so I'll close the PR

PR #19790, opened 2025-06-27
fix(toggletip): align behavior with React for parity

### LINKED ISSUE #19576: React|WC Parity: Toggletip
### Acceptance Criteria

- [ ] See parent issue for baseline criteria
- [ ] Fix styling differences (Paragraph, Button, Link, etc.)
- [ ] There is a weird bug where the link styling will sometimes not show:
- Image

> [!IMPORTANT]  

heloiselui:
Closes #19576

Aligns the `Toggletip` Web Component behavior with React.

### Changelog

**New**

- Added `alignmentAxisOffset` attribute to `Toggletip`, allowing offset adjustment on the alignment axis (only takes effect when `autoalign` is enabled).
- Added `ExperimentalAutoAlign`.
- Added support for `scrollIntoView()` to ensure visibility of the Toggletip in large scrollable containers (storybook parity).

**Changed**

- Updated Toggletip documentation and stories to reflect new props (`autoalign`, `alignmentAxisOffset`, `defaultOpen`).
- Aligned Web Component behavior with React version for consistency.
- Refactored Storybook argTypes and controls for consistency.

**Removed**

- Removed outdated `Playground` story and replaced it with the new `ExperimentalAutoAlign` example.
- Removed `open` attribute from storybook.

#### Testing / Reviewing

- CI should pass
- In the `WC Deploy Preview`, check the updated `Toggletip` stories:
  - `Default`
  - `ExperimentalAutoAlign`
- Visually compare styles and behavior with React version

> [!IMPORTANT]  
> The paragraph in the toggletip label, as well as the Accordion, is also set to 1rem instead of following the correct token. We have a separate [issue](https://github.com/carbon-design-system/carbon/issues/19847) to fix this.

## PR Checklist


riddhybansal (changes requested):
So In react I can see the caret icon is not always centre alligned it moves with autoalign.
image
But in web components it is always centre aligned , so even if it have space it render it in opposite direction.
Here is the video for reference 

https://github.com/user-attachments/assets/210e2348-d26c-4ef0-9a8d-20be7878771e

while in react 


https://github.com/user-attachments/assets/af47dc63-5415-48d0-8b35-fe484fd806e3

heloiselui:
> So In react I can see the caret icon is not always centre alligned it moves with autoalign. image But in web components it is always centre aligned , so even if it have space it render it in opposite direction. Here is the video for reference
> 
>  Screen.Recording.2025-06-30.at.5.12.33.PM.mov 
> while in react
> 
>  Screen.Recording.2025-06-30.at.5.14.27.PM.mov

@riddhybansal I've been digging into this, but it's been taking more time than expected. I realized that once the Popover parity work is complete, this should get resolved along the way. I’ve opened an issue for the Popover parity and left a comment noting that we should address it there.

packages/react/src/components/Toggletip/index.tsx
```diff
@@ -232,7 +232,7 @@ Toggletip.propTypes = {
   ]),
 
   /**
-   * Provide an offset value for alignment axis.
+   * Provide an offset value for alignment axis. Only takes effect when `autoalign` is enabled.
```
tay1orjones: This should've been marked experimental to match `autoAlign`

```suggestion
   * **Experimental:** Provide an offset value for alignment axis. Only takes effect when `autoalign` is enabled.
```

tay1orjones (approved):

riddhybansal:
> > So In react I can see the caret icon is not always centre alligned it moves with autoalign. image But in web components it is always centre aligned , so even if it have space it render it in opposite direction. Here is the video for reference
> > Screen.Recording.2025-06-30.at.5.12.33.PM.mov
> > while in react
> > Screen.Recording.2025-06-30.at.5.14.27.PM.mov
> 
> @riddhybansal I've been digging into this, but it's been taking more time than expected. I realized that once the Popover parity work is complete, this should get resolved along the way. I’ve opened an issue for the Popover parity and left a comment noting that we should address it there.

Sounds good to me !!

riddhybansal (approved):

packages/web-components/src/components/toggle-tip/toggletip.ts
```diff
@@ -98,9 +117,10 @@ class CDSToggletip extends HostListenerMixin(FocusMixin(LitElement)) {
     return html`
       <button
         aria-controls="${this.id}"
+        aria-label="Show information"
```
kennylam: Suggest creating a `label` property (or something similar), which defaults to "Show information" but can be modified by users for localization.

kennylam (approved):
This looks great! Just one small suggestion.

packages/web-components/src/components/toggle-tip/toggletip.ts
heloiselui: Great suggestion, thanks Kenny!

Kritvi-bhatia17 (commented):
Hey @heloiselui! This is looking great—thanks for the work on it! I just had a few quick questions:

1. I noticed the body text in the web component appears to be 16px, but based on Figma, the [documentation](https://carbondesignsystem.com/components/toggletip/style/#typography), and the React implementation, I believe it should be 14px (or 0.85rem). Could you double-check that?
   > image

2. I’m also not sure if the "Experimental auto align" feature in the web component is behaving the same way as it does in React. I’ve attached a video for comparison.
   > https://github.com/user-attachments/assets/ffaab866-3688-4764-a052-134b08126f65

3. Lastly, is there a functional difference between the defaultOpen and open props? 
   > image

heloiselui:
> Hey @heloiselui! This is looking great—thanks for the work on it! I just had a few quick questions:

@Kritvi-bhatia17 as we discussed, here are my comments, thanks again for reviewing this PR!

1. I found the same issue in other components. I came up with a temporary fix for the Accordion, but I also opened an issue to track this problem: https://github.com/carbon-design-system/carbon/issues/19847

2. Once we do the parity work for the Popover, this will definitely be fixed for the Toggletip too. Here's the issue for the Popover parity: https://github.com/carbon-design-system/carbon/issues/19837

3. I talked with Riddhi about it and removed the open from Storybook.

Kritvi-bhatia17:
> > Hey @heloiselui! This is looking great—thanks for the work on it! I just had a few quick questions:
> 
> @Kritvi-bhatia17 as we discussed, here are my comments, thanks again for reviewing this PR!
> 
> 1. I found the same issue in other components. I came up with a temporary fix for the Accordion, but I also opened an issue to track this problem: [[Bug]: <p> elements are not using $body-01 type token #19847](https://github.com/carbon-design-system/carbon/issues/19847)
> 2. Once we do the parity work for the Popover, this will definitely be fixed for the Toggletip too. Here's the issue for the Popover parity: [React/WC Parity: Popover #19837](https://github.com/carbon-design-system/carbon/issues/19837)
> 3. I talked with Riddhi about it and removed the open from Storybook.

Great @heloiselui! LGTM 🚀

Kritvi-bhatia17 (approved):

PR #19813, opened 2025-06-30
fix(contained-list): optional label

ariellalgilmore:
The contained List currently requires a Label. I don't believe the Label should be required and accessibility wise everything still passes without it. In Carbon Labs we want to use the ContainedList without the Label: https://github.com/orgs/carbon-design-system/projects/90/views/6?pane=issue&itemId=3157583280&issue=carbon-design-system%7Ccarbon-labs%7C668

### Changelog

**Changed**

- removed the required on the `label` and add a check if the `label` isn't given then don't render the heading

#### Testing / Reviewing

check the test story

can also make a real story but didn't think it was necessary. 

- [x] remove test story before merging!

## PR Checklist


kennylam (approved):
LGTM!

heloiselui (approved):
LGTM!

Issue #6251, opened 2020-06-10
Link should not prevent click events when no HREF


dabrad26:
When the <Link> component has no href a CSS method of disabling click is used.  This should probably be done via the JS if anything and not done via CSS.  It seems if a <Link has href respect it if not and has onClick use it.  And finally disable it if nothing (the click event shouldn't even be needed since the default style anchor with no href should be used.

```
.bx--link:not([href]):not(button) {
color: #c6c6c6;
cursor: not-allowed;
pointer-events: none;
-ms-touch-action: none;
touch-action: none;
}
```

Expected Behavior:

Remove the pointer-events and touch-action and if onClick is passed in set cursor:pointer on the Link.

There might be some back story to this I don't know.  So would like to find out why this is done.  There are several design requests to use link style for internal navigation actions. And also for those using Segment or tracking on click events or something like HTML5 history in react we want to call that verse calling an actual browser href.

Team: IBM Aspera

dabrad26:
@superkhau

superkhau:
+1, we ran into this issue on API Connect -- Had to work around it with a suggestion from @dabrad26 by adding an `href` prop (even though we're not linking anywhere, just triggering an `onClick` event).

To be more specific, if you don't add `href`, you get an unclickable link that looks like:

![image](https://user-images.githubusercontent.com/1617364/84323425-a3e72780-ab2b-11ea-8bdb-c62f2647d6cf.png)

# TLDR
We should mark `href` as required to get `onClick` to work properly OR remove the requirement of having to define `href` at all. Also, we should at least document the reasoning behind the current behaviour at https://www.carbondesignsystem.com/components/link/usage for posterity.

PR #19831, opened 2025-07-02
fix(link): handling disabled behavior of LinkComponent

### LINKED ISSUE #19729: [Bug]: Link still enabled when we set disabled field to true
### Package

@carbon/react

### Browser

Firefox

### Package version

v1.81.0

### React version

v19.1.0

### Description

so when setting the "disabled" field to true for Link, the Link can still be interacted with. the cursor just changes from the finger to the circle with a slash across it. please see view for more details

https://github.com/user-attachments/assets/0c96f1c9-fb02-4f1d-8876-16f2065570ca

### Reproduction/example

N/A

### Steps to reproduce

see video

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

NAS-IBM:
Closes #19729

### Changelog

**New**

- added event.preventDefault and event.stopPropagation for bubbling of the event for disabled state

**Changed**

- modified css behavior for disabled state.

#### Testing / Reviewing

in the disabled state, even if the DOM element is populated with href, the link will not navigate based on the target.

## PR Checklist

### DIFF
--- a/packages/react/src/components/Link/Link-test.js
+++ b/packages/react/src/components/Link/Link-test.js
@@ -5,10 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
-import Link from '../Link';
+import { Link } from '../Link';
 
 const prefix = 'cds';
 
@@ -129,6 +129,31 @@ describe('Link', () => {
     expect(document.body).toHaveFocus();
     await userEvent.tab();
     expect(document.body).toHaveFocus();
+  });
+
+  // check for disabled onclick handler
+  it('should not call onClick when disabled', async () => {
+    const onClick = jest.fn();
+    render(
+      <Link href="/" disabled onClick={onClick} className="some-class">
+        A simple link
+      </Link>
+    );
+    const link = screen.getByText('A simple link');
+    await userEvent.click(link);
+    expect(onClick).not.toHaveBeenCalled();
+  });
+
+  it('should call onClick when not disabled', async () => {
+    const onClick = jest.fn();
+    render(
+      <Link href="/" onClick={onClick} className="some-class">
+        A simple link
+      </Link>
+    );
+    const link = screen.getByText('A simple link');
+    await userEvent.click(link);
+    expect(onClick).toHaveBeenCalled();
   });
 
   describe('automated verification testing', () => {
--- a/packages/react/src/components/Link/Link-test.js
+++ b/packages/react/src/components/Link/Link-test.js
@@ -5,10 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { render, screen, fireEvent } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
-import { Link } from '../Link';
+import Link from '../Link';
 
 const prefix = 'cds';
 
--- a/packages/react/src/components/Link/Link.tsx
+++ b/packages/react/src/components/Link/Link.tsx
@@ -122,8 +122,25 @@ const LinkBase = React.forwardRef<
 
     const BaseComponentAsAny = (BaseComponent ?? 'a') as any;
 
+    const handleOnClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
+      if (disabled) {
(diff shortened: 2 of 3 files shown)


tay1orjones (commented):
This approach has been considered before but was avoided https://github.com/carbon-design-system/carbon/issues/6251 

I think we might still run into cases where interaction tracking scripts won't be able to capture the click events as expected. What do you think?

Also do you think preventing bubbling is in line with the details on this approach outlined in the comment in the file? 

https://github.com/carbon-design-system/carbon/blob/429a79259d66248b49931af3c0ebc3092a09fa79/packages/react/src/components/Link/Link.tsx#L114-L115

NAS-IBM:
> This approach has been considered before but was avoided #6251
> 
> I think we might still run into cases where interaction tracking scripts won't be able to capture the click events as expected. What do you think?
> 
> Also do you think preventing bubbling is in line with the details on this approach outlined in the comment in the file?
> 
> https://github.com/carbon-design-system/carbon/blob/429a79259d66248b49931af3c0ebc3092a09fa79/packages/react/src/components/Link/Link.tsx#L114-L115

Hi @tay1orjones,
The intended behavior is,
- If the `disabled` property is `false`, with or without href, onclick event is respected and its handlers are executed (as expected in #6251)
- But if it is only disabled,  even href attribute is added from outside the component, the navigation is prevented. JavaScript hook is required because anchor element does not natively support disabled state.

From the approach mentioned in the commented line, it is not mentioned about handling of injecting href attribute dynamically by the application. They mentioned only about removing href element and adding role=link and aria-disabled=true.

tay1orjones:
@NAS-IBM My main concern is that changing how events bubble could be a breaking change for some. Also, like the author mentions, attaching Javascript click events to `<a>` elements without `href` attributes is a bad authoring practice. So rather than modifying the browser behavior and preventing bubbling, my suggestion would be to provide a warning in dev instead.

```diff
     // Reference for disabled links:
     // https://www.scottohara.me/blog/2021/05/28/disabled-links.html
     if (!disabled) {
       linkProps.href = href;
     } else {
       linkProps.role = 'link';
       linkProps['aria-disabled'] = true;

+      // Providing an onClick to disabled links (without an href attribute) is 
+      // a bad authoring practice.
+      warning(
+        !rest.onClick,
+        '`Link` with `disabled` should not be given `onClick`. They do not' +
+        'have an `href` and are rendered as placeholder links. See more:' +  
+        'https://www.scottohara.me/blog/2021/05/28/disabled-links.html and' + 
+        'https://www.scottohara.me/note/2019/07/17/placeholder-link.html'
+      );
     }
```

Can you think of any issues with this approach? If someone encountered this error they'd need to change their usage to something like

```diff
 <Link
   disabled={isDisabled}
-  onClick={myClickHandler}
+  onClick={isDisabled ? undefined : myClickHandler}
   /> 
```

NAS-IBM:
> My main concern is that changing how events bubble could be a breaking change for some.

Please provide an example for my understanding.

@tay1orjones My understanding is, not handling event bubble is the issue to be addressed.

> Can you think of any issues with this approach?
Yes

For the sample provided,

```html
 <Link
   disabled={isDisabled}
-  onClick={myClickHandler}
+  onClick={isDisabled ? undefined : myClickHandler}
   /> 
``` 
If the control is given to the developer, still the developer can choose not to check if the link is disabled but provide the clickHandler like below,

```html
 <Link
   disabled={isDisabled}
   onClick={myClickHandler}
   /> 
``` 
where, the click handler would get executed in the production which is not an expected behavior. Also, this would not resolve this issue.

Also, there could be cases, disabled flag can be dynamic based on the events or states in the other part of the UI. Developers can not be expected to check the condition for executing the click handler.

In the solution provided, instead of the giving the control to the developer, the check is done inside the component.
in this case,
`event.preventDefault()` is equivalent to `undefined` in your suggestion.
`event.stopPropagation()` ensures even if the parent element has onclick behavior, it is not triggered since the click is made on the disabled anchor element.

tay1orjones (approved):
> If the control is given to the developer, still the developer can choose not to check if the link is disabled but provide the clickHandler

I think developers should retain this control. onClick fires natively in the browser in this situation. Developers shouldn't place onClick on disabled links, but it's still possible to do so. There is a chance someone is relying on this (for an admittedly narrow use case) and this could break it.  It's why I suggest a warning instead of outright blocking this behavior.

I agree it's a footgun though, so I'll hold these reservations for now. If we receive feedback on this change we can revisit it. 

Thanks for hearing me out and sharing your perspective on this! LGTM

riddhybansal (approved):

PR #19859, opened 2025-07-08
test(storybook): add test-only stories and hide them from the sidebar

### LINKED ISSUE #19805: Chromatic exploration: Replace controls-based VRT tests with purpose-build stories
When using Chromatic for VRT, stories cannot be configured via controls before being snapshotted. Instead, we must have new stories for each component state we want to snapshot. These new stories will not be of value to a consumer looking at the storybook for documentation, so they should be hidden from the left sidebar/nav.

## Tasks

- [ ] Create a new story for each test that we're configuring controls (via `args`) before snapshotting.
  - Currently there are only [3 usages](https://github.com/search?q=repo%3Acarbon-design-system%2Fcarbon+%22args%3A%22+path%3A*-test.e2e.js&type=code)
- [ ] For each story, hide it from the left nav using the new `tags` feature in storybook.

## Details on the `tags` feature in storybook

Here's more info from Chromatic on the tags feature and how to use it to hide "test only" stories from the left nav:

> ### Test-only stories
> One of the things we discussed today was the preference to keep your Storybook docs and sidebar clean, while expanding your test coverage as you run your POV. The great news is…I had forgotten that Storybook added specific built-in tags that can help make this possible! 🎉
> 
> Storybook has a recipe for docs-only stories, but for your use case there are stories you may want to keep out of the sidebar and docs, while still retaining the ability to run regression testing. In these situations, you’ll want to remove the `dev` and `autodocs` tags. This ensures they won’t appear in your sidebar and won’t show up in your autodocs, but you can still test them using Chromatic (or test runner, etc.). I’ll include a quick example below!
> 
> ```js
> // Replace your-framework with the framework you are using, e.g. react-vite, nextjs, vue3-vite, etc.
> import type { Meta, StoryObj } from '@storybook/your-framework';
> 
> import { Button } from './Button';
> 
> const meta = {
>   component: Button,
>   /*
>    * All stories in this file will:
>    * - Be included in the docs page
>    */
>   tags: ['autodocs'],
> } satisfies Meta<typeof Button>;
> export default meta;
> 
> type Story = StoryObj<typeof meta>;
> 
> export const Disabled: Story = {
>   /*
>    * This story will:
>    * - Be excluded from the docs page
>    * - Removed from the sidebar navigation
>    * - Still be a tested variant
>    */
>   tags: ['!dev', '!autodocs']
> }
> ```

Also a heads up: adding this in our stories will look slightly different as we're not using the same authoring pattern shown here. 

---

> [!IMPORTANT]  
> Please add implementation notes and feedback here:
> - https://github.com/carbon-design-system/carbon/issues/19810

tay1orjones:
Closes #19805

This adds three new stories that configure components with specific args for testing. These stories will be visually tested with Chromatic, while being hidden from the navigational sidebar. 

> [!IMPORTANT]  
> These stories are still included in the build, meaning they can be accessed by URL even when hidden from the sidebar.
>
> Additionally we discovered that there is no way to conditionally hide stories based on environment (production vs dev, for instance), so to debug vrt failures of these stories we'll have to either
> 1. Hit them by url directly
> 2. Temporarily reconfigure the tags from `['!dev', '!autodocs']` to `['dev', 'autodocs']`, and then switch it back before merging

### Changelog

**Changed**

- add stories to existing story files for 3 components that use args for vrt

#### Testing / Reviewing

- Go to the deploy preview, validate that these stories are not shown in the sidebar.
- Go to each story directly by url on the deploy preview
  - {storybookDeployBaseURL}/?path=/story/components-breadcrumb--breadcrumb-with-overflow-menu-size-small
  - {storybookDeployBaseURL}/?path=/story/experimental-statusindicators-unstable-iconindicator--default-with-size-20
  - {storybookDeployBaseURL}/?path=/story/experimental-statusindicators-unstable-shapeindicator--default-with-text-size-14
  - Ensure each of these bear the visual change present in the args (size difference, text size difference)
- In the chromatic status/ci check, review to make sure these stories are included and snapshotted. If they're not listed we'll need to debug why.

## PR Checklist

### DIFF
--- a/packages/react/src/components/Breadcrumb/Breadcrumb.stories.js
+++ b/packages/react/src/components/Breadcrumb/Breadcrumb.stories.js
@@ -79,6 +79,43 @@ BreadcrumbWithOverflowMenu.argTypes = {
   ...sharedArgTypes,
 };
 
+export const BreadcrumbWithOverflowMenuSizeSmall = (args) => {
+  console.log(process.env.IS_PRODUCTION);
+
+  return (
+    <Breadcrumb noTrailingSlash {...args}>
+      <BreadcrumbItem>
+        <a href="/#">Breadcrumb 1</a>
+      </BreadcrumbItem>
+      <BreadcrumbItem href="#">Breadcrumb 2</BreadcrumbItem>
+      <BreadcrumbItem data-floating-menu-container>
+        <OverflowMenu align="bottom" aria-label="Overflow menu in a breadcrumb">
+          <OverflowMenuItem itemText="Breadcrumb 3" />
+          <OverflowMenuItem itemText="Breadcrumb 4" />
+        </OverflowMenu>
+      </BreadcrumbItem>
+      <BreadcrumbItem href="#">Breadcrumb 5</BreadcrumbItem>
+      <BreadcrumbItem isCurrentPage>Breadcrumb 6</BreadcrumbItem>
+    </Breadcrumb>
+  );
+};
+
+BreadcrumbWithOverflowMenuSizeSmall.argTypes = {
+  ...sharedArgTypes,
+};
+
+/*
+ * This story will:
+ * - Be excluded from the docs page
+ * - Removed from the sidebar navigation
+ * - Still be a tested variant
+ */
+BreadcrumbWithOverflowMenuSizeSmall.tags = ['!dev', '!autodocs'];
+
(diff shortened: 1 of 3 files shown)


maradwan26 (approved):
LGTM !

kennylam (approved):

PR #19884, opened 2025-07-10
feat(filterable-multiselect): New SelectAll Feature

### LINKED ISSUE #17950: [Feature Request]: Select all filtered options in FilterableMultiSelect
### The problem

We need the ability for users to filter the list of options and then select everything the filter applies to with a single click.

### The solution

Implement Select all to FIlterableMultiSelect, as implemented for MultiSelect (https://github.com/carbon-design-system/carbon/issues/16117).

### Code of Conduct

### LINKED ISSUE #18455: [Feature Request]: Extend the select all feature to filterable multiselect
We have successfully implemented select all in multiselect, and now we need to implement the same behavior in filterable multiselect. This functionality should work identically to the existing multiselect implementation, maintaining consistency across our components.

### LINKED ISSUE #19768: [Feature Request]: SelectAll in FilterableMultiSelect
### The problem

Select All option is not implemented for the FilterableMultiSelect component. Not a big issue, but I believe could be a good enhacement on the component, not sure if was considered at some point.

Right now if you pass a `SelectAll` option as described in documentation cannot be selected/unselected, it keeps without the check.

### The solution

SelectAll in FilterableMultiSelect functionality should be available and have the same behavior as the SelectAll in MultiSelect component.

### Examples

_No response_

### Package

@carbon/react

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

maradwan26:
Closes #19768 
Closes #18455 
Closes #17950 

Added a Select All feature to Filterable MultiSelect.

Behaviour summary:
- When the selectAll item is clicked, all _filtered_ items are selected, this means that when items = {one, two, three}. Filtering for "three" and clicking selectAll only selects three, since it is the only filtered item present. Deselections also behave in this manner with filters.

- The checked/indeterminate/empty state of the selectAll item changes dynamically based on the filter. If only half of the filtered items are selected it'll be indeterminate. But, if the text input changes to filter just one, already selected item, it will be checked.

- The selectAll item is only present if it is usable. Meaning, if there are no filtered items (i.e. the dropdown is empty), there will be no selectAll item, since it has nothing to select. The same behaviour applies for if the only selectable item(s) are disabled.

- The selectAll item is always at the top of the list for usability, no matter what the filter is.

- The selectAll item is not included in the count of selected items in the `tag` on the left-hand-side of the text-input.

### Changelog

**New**

- Added SelectAll feature
- Added a new internal function "`toggleAll`" to `Selection.tsx` to help with the implementation of `handleSelectAllClick()`
- Added a "Filterable with Select All" story to showcase this new feature
- Added test cases for this new feature

**Changed**

- `InputKeyDownEnter` and `ItemClick` switch cases call `handleSelectAllClick()` instead of `onItemChange()` when the item clicked is a selectAll item
- `ListBox.MenuItem` now uses `Checkbox` instead of a `span`

**Removed**

- Removed a test case that tests the `span` inside of `ListBox.MenuItem` since the `span` has been removed.

#### Testing / Reviewing

- Go to the new "Filterable with Select All" story in the deploy preview
- Behaviours listed above should work as expected
- CI should pass

## PR Checklist


packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -1014,23 +1109,28 @@ export const FilterableMultiSelect = forwardRef(function FilterableMultiSelect<
...
+                          ) : (
+                            itemText
+                          )
+                        }
+                        checked={isChecked}
                         title={useTitleInItem ? itemText : undefined}
-                        className={`${prefix}--checkbox-label`}
+                        indeterminate={isIndeterminate}
+                        disabled={disabled}
                         data-contained-checkbox-state={isChecked}
```
Gururajj77: `data-contained-checkbox-state` can be removed, as it's not a prop that is required by the checkbox.

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -1014,23 +1109,28 @@ export const FilterableMultiSelect = forwardRef(function FilterableMultiSelect<
...
-                        id={`${itemProps.id}-item`}>
-                        {ItemToElement ? (
-                          <ItemToElement key={itemProps.id} {...item} />
-                        ) : (
-                          itemText
-                        )}
-                      </span>
+                        tabIndex={-1}
+                      />
```
Gururajj77, line 1132: And why do we need `tabIndex={-1}` here??

laurenmrice:
@maradwan26 Is this PR also adding this to WC, or just React?

maradwan26:
@laurenmrice This PR is just for React. Once its merged and everything looks good I'll make a new issue for the WC version

laurenmrice (approved):
This looks good to me! ⭐️

ariellalgilmore (commented):
looks goood, but when clicking on the "All roles" for some reason the cursor loses the padding 
https://github.com/user-attachments/assets/c2c97f42-78a8-4723-8e5e-b63e6d454c1e

maradwan26:
@ariellalgilmore Fixed! Thanks for pointing that out

ariellalgilmore (approved):
LGTM!

Gururajj77 (changes requested):

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
maradwan26: removed ✅

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
maradwan26: The tabIndex was added to pass [a CI test case](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/MultiSelect/__tests__/FilterableMultiSelect-test.js#L1126C2-L1156C4)

Gururajj77 (approved):

PR #19918, opened 2025-07-15
fix: adds warning for inccorect API usage on datepickerInput

### LINKED ISSUE #19712: [Bug]: DatePickerInput fails to display value under DatePicker range in readOnly mode
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.83.0 and v1.84.0 

### React version

v18.3.1

### Description

With `DatePicker` component in `readOnly` mode and `datePickerType="range"`, the 2nd child `DatePickerInput` component fails to display the value. Instead, it displays the placeholder value.

This bug was first introduced in @carbon/react v1.83.0 and persists in 1.84.0. In older versions 1.82.0 and prior, the bug is not present and everything works correctly.

The 1st screenshot is `readOnly` mode enabled, note the 2nd date field not displaying value. This is the bug.
Image

The 2nd screenshot is with `readOnly` mode disabled, note the 2nd date field displaying value as it should.
Image

### Reproduction/example

n/a

### Steps to reproduce

`npm install @carbon/react v1.83.0` or `npm install @carbon/react v1.84.0`

```
<DatePicker
                      id="startend"
                      datePickerType="range"
                      readOnly={readOnly}
                      disabled={isSubmitted}
                      onChange={(e) => {
                        setStartDate(e[0] && e[0].toLocaleDateString());
                        setEndDate(e[1] && e[1].toLocaleDateString());
                        setStartDateInvalid(false);
                        setEndDateInvalid(false);
                      }}
                    >
                      <DatePickerInput
                        id="startDate"
                        placeholder="mm/dd/yyyy"
                        labelText="Start date"
                        size="md"
                        value={startDate}
                        invalid={startDateInvalid}
                        invalidText="Valid date required."
                      />
                      <DatePickerInput
                        id="endDate"
                        placeholder="mm/dd/yyyy"
                        labelText="End date"
                        size="md"
                        value={endDate}
                        invalid={endDateInvalid}
                        invalidText="Valid date required."
                      />
</DatePicker>
```

### Suggested Severity

### Application/PAL

IBM Redbooks

### Code of Conduct

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/19712

This PR introduces a one-time console `warning` when the value prop is passed directly to the DatePickerInput component.

Our recent TypeScript migration and stricter prop validation in the DatePicker component exposed an incorrect API usage pattern that was previously "accidentally functional."

`value` is a fundamental HTML input attribute, making it intuitive for developers to try to control DatePickerInput directly. This warning provides immediate, explicit feedback that this is incorrect usage, guiding them to the proper API.

### Changelog

**New**

Adds a `warning` in the console if value is passed to `DatePickerInput`

#### Testing / Reviewing

pull down PR locally (warning is only thrown in non production env )
open "datepicker--read-only-range-bug-reproduction"  story
verify warning is seen once 

- [ ] test story needs to be removed before merging 

## PR Checklist


Gururajj77 (approved):
LGTM🚀

packages/react/src/components/DatePickerInput/DatePickerInput.tsx
```diff
@@ -189,6 +191,18 @@ const DatePickerInput = React.forwardRef(function DatePickerInput(
...
+      warning(
+        false,
+        `The 'value' prop is not supported on the DatePickerInput component. ` +
+          `For DatePicker components with 'datePickerType="range"', please ` +
+          `pass the 'value' prop (as an array of dates) to the parent ` +
+          `DatePicker component instead.`
+      );
+      didWarnAboutDatePickerInputValue = true;
+    }
+  }
```
tay1orjones:
```suggestion
  if (process.env.NODE_ENV !== 'production' && !didWarnAboutDatePickerInputValue) {
    warning(
      !('value' in rest),
      `The 'value' prop is not supported on the DatePickerInput component. ` +
        `For DatePicker components with 'datePickerType="range"', please ` +
        `pass the 'value' prop (as an array of dates) to the parent ` +
        `DatePicker component instead.`
    );
    didWarnAboutDatePickerInputValue = true;
  }
```

tay1orjones (approved):
LGTM once the test story is removed

ariellalgilmore (approved):

packages/react/src/components/DatePickerInput/DatePickerInput.tsx
2nikhiltom: Hey - I suspect this would always set the `didWarnAboutDatePickerInputValue` flag irrespective of the warning being shown or not. I changed it to
```
if (process.env.NODE_ENV !== 'production' && 'value' in rest && !didWarnAboutDatePickerInputValue) {
  warning(
    false,
    `The 'value' prop is not supported on the DatePickerInput component. ` +
      `For DatePicker components with 'datePickerType="range"', please ` +
      `pass the 'value' prop (as an array of dates) to the parent ` +
      `DatePicker component instead.`
  );
  didWarnAboutDatePickerInputValue = true;
}
```

PR #19955, opened 2025-07-17
Pagination.page select label text

### LINKED ISSUE #18516: [Feature Request]: Allow for passing custom label for page select in Pagination
### The problem

The `Pagination` component doesn't allow for passing custom label to page selector. The value is hard-coded (https://github.com/carbon-design-system/carbon/blob/fa1abc1b47b4daa5c52c9e9d8b03fdd65780e0d8/packages/react/src/components/Pagination/Pagination.tsx#L391).

### The solution

A prop to `Pagination` component could be introduced that would allow for providing custom label. It could accept a function with `totalPages` as argument.

### Examples

_No response_

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

### Tasks

- [ ] React Implementation

kadiryazgan:
Closes #18516

Currently the label of the select in Pagination component is hardcoded, thus cannot be localized.
This PR adds `pageSelectLabelText` property to the component to pass a custom function for the label value. It defaults to the previous hard-coded value if no function is provided, so it does not effect existing setups.

### Changelog
**New**
Added pageSelectLabelText property to Pagination component.


Gururajj77:
Hey @kadiryazgan, Thanks for the contribution, could you follow the specified format and write about the changes you made and how it affects the component? It helps us understand the changes you made and why it is been made.

kadiryazgan:
@Gururajj77 done.

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -180,6 +185,7 @@ const Pagination = React.forwardRef(function Pagination(
     pageNumberText: _pageNumberText = 'Page Number',
     pageRangeText = (_current, total) =>
       `of ${total} ${total === 1 ? 'page' : 'pages'}`,
+    pageSelectLabelText = (_current, total) => `Page of ${total} pages`,
```
tay1orjones: This could result in `Page of 1 pages`, right? Could it handle plurality for page/pages like what's done in `pageRangeText`? Also why have `_current` as a parameter if it's not used?

tay1orjones (changes requested):

packages/react/src/components/Pagination/Pagination.tsx
kadiryazgan: I wanted to preserve the previous hardcoded text as the default value, that's why it reads like that. But I changed the default value so that it reads `Page _current_ of _total_ page` (or `pages` if plural)

tay1orjones:
Closing in favor of https://github.com/carbon-design-system/carbon/pull/21699 - I fixed the merge conflict but couldn't push here for some reason. I left your contribution status in tact. Thanks again!

PR #19963, opened 2025-07-21
feat(NumberInput): add support for defaultStartValue

### LINKED ISSUE #19962: [Feature Request]: [NumberInput] Support defaultStartValue prop to define stepper start when input is empty; default to 0 if min < 0 and max > 0
### The problem

Add support for a new defaultStartValue prop in NumberInput to define the initial value for stepper when the input is empty.
If defaultStartValue is not provided, use logical defaults based on min and max.
Specifically, use 0 when min < 0 and max > 0.

### The solution

Introduce a new prop defaultStartValue in the NumberInput component.
This value will be used as the starting point when incrementing or decrementing from an empty input.
If not provided, fallback logic based on min and max will apply.

### Examples

`<NumberInput
            type="text"
            label="test-label"
            id="test"
            step={2}
            defaultStartValue={10}
            translateWithId={translateWithId}
          />`

### Package

@carbon/react

### Application/PAL

PAL, Graphite

### Business priority

### Available extra resources

_No response_

### Code of Conduct

abpaul1993:
Closes https://github.com/carbon-design-system/carbon/issues/19962

Introduced a new prop `defaultStartValue` to customize the starting value for stepper actions when the input is empty.  
Falls back to 0 when min < 0 and max > 0, or applies min/max based defaults otherwise.

image

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

## PR Checklist


tay1orjones (commented):
Hey, thanks for putting up this PR! I'm on the fence about this because it can make the UX of this field inconsistent for the user. 

Three questions I have:

1. Could you give a real world example of when this prop would be used and why the min/max increment/decrement default doesn't work?
2. If you (the developer) know what `defaultStartValue` should be, why not just provide a `defaultValue`?
3. If there's strong reasoning against 1. the min/max increment/decrement default and 2. a `defaultValue`, wouldn't the most predictable experience for the user be to just always start at `0` rather than some "random" value selected by the developer?

If this PR is accepted these three pieces information will be important for documentation purposes on the website and storybook to know when to recommend to use `defaultStartValue`.

tay1orjones:
@thyhmdo I put the `visual review` label on here for more of a UX design review based on my earlier comment, not for an actual visual review.

abpaul1993:
Hi @tay1orjones, For a controlled NumberInput, can defaultValue and value be used together?

JordanWSmith15:
Hey @tay1orjones 
So we ran in to some issues with our previous decision. Please read here:
https://ibm-ai-apps.slack.com/archives/C2K6RFJ1G/p1752756697413319?thread_ts=1749648029.085619&cid=C2K6RFJ1G
The Min and Max schema in Maximo is very large for most of our inputs (billions negative and positive) so hitting the - and + was resulting in very large numbers and did not feel useful.  Our testing team and designers helped us to determine what to do.
We still want to avoid setting 0 when the max or min schema would produce an error with that as a default, so we decided the best way to go about this was to just create a property. 

To answer your three questions:

 Could you give a real world example of when this prop would be used and why the min/max increment/decrement default doesn't work?
    **• A basic one I can think of is setting a year.  Min is 1950 and max of 2100, but for the steps, it makes a lot of sense to set current year (2025) for either + or - starting points.** 
    If you (the developer) know what defaultStartValue should be, why not just provide a defaultValue?
    **• Default value is typically not allowed in Maximo forms, as we verify the response from the user is real and was not skipped over (think of entering a meter reading on an asset inspection).** 
    If there's strong reasoning against 1. the min/max increment/decrement default and 2. a defaultValue, wouldn't the most predictable experience for the user be to just always start at 0 rather than some "random" value selected by the developer?
    **• I do think we can start at zero by default when zero is a valid entry, but we should also provide a prop to optimize the experience (as with my Year example above)** 
    FYI @abpaul1993 @sls-ca

sls-ca:
@tay1orjones using min and max as default start values is not a great UX, since, those values are never/rarely displayed to the user and are generally only displayed during validation.  How many times have you see "Please enter a number from -1.7976931348623157e+308 to +1.7976931348623157e+308" as the label on the field or placeholder?

The differences between a defaultStartValue and defaultValue has to do with the field initialization.  If you set a defaultValue then the field becomes populated, the end user doesn't know if that value came from the database (ie, was always there), and as such, they might go, "yeah, this is value I want" and then do nothing, but, that value is never stored to the db.  ie, the UX of showing a user a default value in a form as a value vs a placeholder is a very bad UX.

A defaultStartValue is basically a means of saying that given my range of possible min/max values this is the value that is mostly likely going to be used when the field is empty and the user hits +/=.  @JordanWSmith15 use case of years, eg.

This prop was intentionally created as not to cause side effects with the existing behaviour, so I'm a little surprised to see this much push back :(

For any given number input that is initialized with "nothing" I would expect the following
if there is no defaultStartValue
* if 0 is within the range of min/max then +/- would result in 0 (no min/max assumes any value, so it would also be 0)
* if 0 is not in the range of min/max then +/- would result in min/max
I think that intuitively from a UX point of view "empty" is "nothing" which is 0, so hitting +/- should either result in 0 first and then increment/decrement or it should start -/- of 0.

if there is a defaultStartValue, then the application is trying to offer a better UX than 0 (or the possible min/max).

Given that the current behaviour of NumberInput is that it doesn't follow the rules that I listed, and it assumes absolute min/max for an empty value, defaultStartValue is required and likely would be set to 0 in 99% of the cases.  Again, from a empty field with nothing, I doubt anyone would ever expect absolute min/max to appear when pressing +/-.  As it stands, if this were merged, it offers potential for a better UX and offers no sidefeccts to users that are already consuming the NumberInput component.

tay1orjones:
@JordanWSmith15 that's great clarifying info, thanks! That will help provide guidance on when it should be used. @abpaul1993 maybe something like this in the `Numberinput.mdx`:

> A step start value can be configured to provide the user a more beneficial starting point when the input is empty and an increment or decrement button is clicked. This can be used in situations where zero is not within the min/max or there is a more sensible default starting value, such as the current year when asking the user for a year. Since this allows the input to still initially be empty, it ensures on submission of the form that the user has actually interacted with the input and inserted a value.

---

@sls-ca I found this behavior is actually documented in the HTML spec, which I think we could mostly follow here to mimic browser defaults ([link](https://html.spec.whatwg.org/multipage/input.html#concept-input-min-zero:~:text=The%20step%20base%20is,Return%20zero.)). 

> I think that intuitively from a UX point of view "empty" is "nothing" which is 0, so hitting +/- should either result in 0 first and then increment/decrement or it should start -/- of 0.

Browsers do the latter - starts at +/- of `0`, so `-1` or `1`. I made a stackblitz with some example cases to look at this: https://stackblitz.com/edit/github-uvjmrhmc?file=src%2FApp.jsx

I also think we could reduce confusion by naming the prop `stepStartValue` (and maybe give it a default of `0`?). Then in the stepper logic the min/-1/1 could be handled. 

So all together my feedback would be:

* Add guidance to the .mdx overview docs for NumberInput
* Rename prop to `stepStartValue`
* if 0 is within the range of min/max, or there is no min or max
  * `+` click results in 1 
  * `-` click results in -1
* if 0 is not in the range of min/max 
  * `+` click results in `min`
  * `-` click results in `min`
  
Let me know what you think

sls-ca:
@abpaul1993 Please make the changes as per @tay1orjones's comment.

abpaul1993:
Hi @sls-ca and @tay1orjones I have updated my code as per the feedback.

packages/react/src/components/NumberInput/__tests__/NumberInput-test.js
```diff
@@ -920,7 +920,7 @@ describe('NumberInput', () => {
         expect(screen.getByLabelText('test-label')).toHaveValue('0');
       });
 
-      it('should begin incrementing from min when input is empty', async () => {
+      it('should begin incrementing from 1 when input is empty and o is in betwenn of min and max', async () => {
```
tay1orjones:
```suggestion
      it('should begin incrementing from 1 when input is empty and 0 is in between of min and max', async () => {
```

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -109,6 +109,11 @@ export interface NumberInputProps
    */
   formatOptions?: NumberFormatOptions;
 
+  /**
+   * Provide stepper start value when input is empty
```
tay1orjones:
```suggestion
   * Provide the value stepping should begin at when the input is empty
```

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -909,6 +922,11 @@ NumberInput.propTypes = {
    */
   min: PropTypes.number,
 
+  /**
+   * Stepper Start Value When Input is Empty.
```
tay1orjones:
```suggestion
   * Provide the value stepping should begin at when the input is empty
```

tay1orjones (approved):
A couple minor suggestions, otherwise LGTM 👍 Thanks!

riddhybansal (approved):

thyhmdo (approved):
Discussed with @tay1orjones Looks good to me

tay1orjones:
@abpaul1993 Shouldn't this also work when input type is number? I tried on the Default story and it doesn't have an effect.

https://deploy-preview-19963--v11-carbon-react.netlify.app/?path=/story/components-numberinput--default&args=allowEmpty:!true;stepStartValue:10;type:number

allowEmpty: true, stepStartValue: 10, type: number

abpaul1993:
@tay1orjones , i have not changed any behaviour when input type is number? do we need the same behaviour for this as well?

tay1orjones:
Yeah, the step calculation logic in `handleStep` runs for both type=text and type=number. `currentValue` is determined differently based on the type, but the step calculation thereafter is the same for both, right? Ideally the prop wouldn't be limited to type=text.

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -530,7 +536,14 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
           // When the field is empty (NaN), incrementing begins at min,
           // decrementing begins at max.
           // When there's no min or max to use, it begins at 0.
```
tay1orjones: I think this comment needs updated since it's no longer correct. Also is the last condition needed anymore since we don't want it to ever start at `max`?

```
} else if (direction === `down` && max) {
            rawValue = max;
```

abpaul1993:
Hi @tay1orjones , can you please review one more time? This time for type number also this property and stepper button features should work.

tay1orjones (approved):

PR #19965, opened 2025-07-21
fix(radio-button): skip disabled items in group keyboard navigation

### LINKED ISSUE #19959: [Bug]: Radio button disabled gets selected using keyboard
### Package

@carbon/web-components

### Browser

Chrome

### Package version

latest

### React version

_No response_

### Description

I am using Carbon web component radio button group with some disabled items and I've noticed that using the keyboard arrows, it the next item is the disabled radio button, it gets selected even though it does not get focus. You can see the incorrect behavior in the gif attached taken from Carbon storybook

I would expect that the focus/selected would go directly to the next non disabled button as it is in react 

Image

### Reproduction/example

https://web-components.carbondesignsystem.com/?path=/story/components-radio-button--vertical

### Steps to reproduce

1. Create a radio button group with 3 items, where the second item is disabled
2. Tab into the component, it should focus and select the first radio button
3. Navigate using Arrow key down: disabled item is selected
 

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

jose-biescas:
Closes #19959 

Adds check to skip disabled items when navigating radio buttons

### Changelog

*New*

- Add check to skip disabled items
- Added calculation to find next available radio button

*Changed*

- none

*Removed*

- none

#### Testing / Reviewing

Storybook for Vertical shows navigation skipping the disabled item.
Added extra radio buttons to storybook with disabled states to test.
yarn build succeeded
yarn test passed

## PR Checklist


jose-biescas:

2nikhiltom:
hey @jose-biescas Thanks for putting up this PR. The build and ci checks are currently failing , could you take a look before we review this

jose-biescas:
Hi @2nikhiltom, upon reviewing, I found 2 things actually.

1. Reading the build logs. I don't think the error is entirely accurate. I think it is because I added the `disabled` property to the `ManagedRadioButtonDelegate`, so that the `navigate` function was aware of the disabled state of a radio button; hence, skipping it if needed when navigating with keyboard. But this then requires `StructuredListRowRadioButtonDelegate` (which implements `ManagedRadioButtonDelegate`) to also implement the `disabled` state, which upon looking, seems like isn't a property that Structured List has.
2. I just tested both carbon/react and carbon/web-component `Structured List`, in React, the `Selection` version is non-navigable via keyaboard, but the web-component version is.

Regarding the build fails and how to fix. Since `StructuredListRowRadioButtonDelegate` doesn't need the `disabled` state. We can make the `disabled` property optional (?). 

The other option is to revert the changes I made to `navigate` and `ManagedRadioButtonDelegate` and apply the logic directly to the `_handleKeydown` event function of `CDSRadioButton`

2nikhiltom:
@jose-biescas  I think making the disabled property optional (disabled?: boolean) in the ManagedRadioButtonDelegate interface is the correct fix for the build failures. That will resolve the issue with StructuredListRowRadioButtonDelegate not implementing the disabled property.

2nikhiltom:
@jose-biescas The chnages looks good to me , just a small conflict and needs attention

tay1orjones (approved):

2nikhiltom (approved):
Thanks for making the changes

PR #19971, opened 2025-07-21
fix(invalid-text): fixed global styling for invalid-text

### LINKED ISSUE #19960: [Bug]: TextInput error text overlapping components below
### Package

@carbon/react

### Browser

Chrome

### Package version

1.87.0

### React version

18.3.1

### Description

The error text that appears below an `TextInput` component doesn't dynamically resize the spacing around it.. as a result, it overlaps content/components below the TextInput component.

error text is dynamic, where it will appear if `invalid=true` and `invalidText="<msg>"` is set.. And these error msgs can be short or long (could be 1 line or multi-line) msgs.

The current problem is, regardless of the error msg's length (1-line or multi-line msg), this error text field does not dynamically expand or give space for the text.. Instead, it just overlaps the content below. 
Image

I believe this use to work properly by dynamically giving space to the error text when visible, and collapses the extra empty space when not visible.. But now, it doesn't seem to work. I've also tried changing the css for the error text element (className `cds--form-requirement` and has id of `<id>-error-msg`), such as setting `height`, `margin` or `padding`, etc.. but it seems like this error element is floating or something, and not actually included as part of the layout.. (which probably is why it's not being dynamic). 
Image

I was just testing to see if there was anything we can do manually, but this SHOULD be resolved from the carbon library package.. As we have many `TextInput` fields all over the place, this should work out of box for all instances of the component.

### Reproduction/example

https://stackblitz.com/edit/owdgdbvv?file=src%2FApp.jsx

### Steps to reproduce

Just create an TextInput field with `invalid`/`invalidText` configured. Add some content below it to see it overlap (when the error appears).

In the reproduction example, just enter `-` to trigger invalid text scenario in the first `TextInput` field. You will see it overlap the content below and does not dynamically allocate space to that error text element.. Some error msgs can be long and be multi-lined, so that would overlap even more or get hidden.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

### LINKED ISSUE #19967: [Bug]: Checkbox warning and error icons are misaligned in React
Icons should be in-line with waring/error text
https://react.carbondesignsystem.com/?path=/story/components-checkbox--single

Image

Correct visual style is visible in Web components storybook.  https://web-components.carbondesignsystem.com/?path=/story/components-checkbox--single

maradwan26:
Closes #19960 
Closes #19967

There was a styling bug where invalidText would have the incorrect behaviour. For `TextInput` it would overlap with components below it, for `Dropdown` it would not show up at all when there are components below it, for `Checkbox` there would be alignment issues, etc.

This PR fixes this bug and re-implements the invalid-text overflow functionality when `cols` is used for `TextArea` for both React and Web Components since the original implementation was problematic.

### Changelog

**New**

- Added hidden test-only story for Dropdown: `?path=/story/components-textinput--test-invalid-text-no-overlap`
- Added hidden test-only story for TextInput: `?path=/story/components-textinput--test-invalid-text-no-overlap`

- Added `warnId` to the warn text in `TextArea` in React since it was missing

**Changed**

- Changed the implementation of the helper/invalid/warn-text overflow when the `cols` attribute was used in `TextArea` for React and WC:
    - _React_: Added refs for helper, invalid and warn texts and the wrapper which is then used to update their max-width accordingly in `useIsomorphicEffect` when the `cols` prop changes.
    - _WC_: Added a private `_measureWrapper()` function which, when `cols` is changed, measures the width of the wrapper and applies the appropriate styles to the helper/invalid/warn text.

**Removed**

- Removed `position: absolute` and `position: relative` styling that was used to implement text-overflow from React and WC `TextArea` since it was problematic
- Removed `.#{$prefix}--text-area--invalid {padding-inline-end: $spacing-08;}` from `_text-area.scss` since this was causing `TextArea` to increase its width past the set `cols` width when `invalid='true'`

#### Testing / Reviewing

- Go to https://deploy-preview-19971--v11-carbon-react.netlify.app/?path=/story/components-dropdown--test-invalid-text-no-overlap
    - Invalid text should not overlap with component below and should be visible
- Go to https://deploy-preview-19971--v11-carbon-react.netlify.app/?path=/story/components-textinput--test-invalid-text-no-overlap
    - Invalid text should not overlap with component below and should be visible
- Go to `Checkbox` in React storybook
    - Go to Single story
    - Invalid text should be aligned
- Go to `TextArea` in React storybook
    - Set `cols=2` and `invalid='true'`
    - Invalid text should wrap
- Go to `TextArea` in Web Components storybook
    - Set `cols=2` and `invalid='true'`
    - Invalid text should wrap
- Can also test with helper text and warn text to make sure they work as well

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.stories.js
+++ b/packages/react/src/components/Dropdown/Dropdown.stories.js
@@ -206,6 +206,7 @@ export const Default = (args) => {
         itemToString={(item) => (item ? item.text : '')}
         {...args}
       />
+      <Dropdown />
     </div>
   );
 };
--- a/packages/react/src/components/Dropdown/Dropdown.stories.js
+++ b/packages/react/src/components/Dropdown/Dropdown.stories.js
@@ -258,7 +258,7 @@ export const Test19971 = (args) => {
         itemToString={(item) => (item ? item.text : '')}
         {...args}
       />
-      <Dropdown id="test-2" items={items} />
+      <Dropdown titleText="test title " id="test-2" items={items} />
     </div>
   );
 };
--- a/packages/react/src/components/Dropdown/Dropdown.stories.js
+++ b/packages/react/src/components/Dropdown/Dropdown.stories.js
@@ -206,7 +206,7 @@ export const Default = (args) => {
         itemToString={(item) => (item ? item.text : '')}
         {...args}
       />
-      <Dropdown />
+      <Dropdown id="default-2" />
     </div>
   );
 };
--- a/packages/react/src/components/Dropdown/Dropdown.stories.js
+++ b/packages/react/src/components/Dropdown/Dropdown.stories.js
@@ -206,12 +206,63 @@ export const Default = (args) => {
         itemToString={(item) => (item ? item.text : '')}
         {...args}
       />
-      <Dropdown id="default-2" />
     </div>
   );
 };
 
 Default.args = {
+  ...sharedArgs,
+};
+
+// Test story, remove before merging
+export const Test19971 = (args) => {
+  const items = [
+    {
+      text: 'Lorem, ipsum dolor sit amet consectetur adipisicing elit.',
+    },
+    {
+      text: 'Option 1',
+    },
+    {
+      text: 'Option 2',
+    },
+    {
+      text: 'Option 3 - a disabled item',
+      disabled: true,
+    },
+    {
+      text: 'Option 4',
+    },
+    {
+      text: 'Option 5',
+    },
+    {
+      text: 'Option 6',
+    },
+    {
+      text: 'Option 7',
+    },
+    {
+      text: 'Option 8',
+    },
+  ];
+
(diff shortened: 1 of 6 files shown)


tay1orjones (commented):
This looks great to me. Do you think there's a way to add some VRT for this with the existing stories? Just trying to think of some ways we can cover this in tests to ensure it doesn't happen again. With chromatic we can have test-only stories, we could pair up on that if you'd like!

maradwan26:
@tay1orjones Sure! That sounds good. I've messaged you on slack

riddhybansal:
Found the same issue for combobox as well !! I feel this PR fixes that also 
[Slack thread](https://ibm-studios.slack.com/archives/C2K6RFJ1G/p1753866676174569)

tay1orjones (approved):
LGTM!

riddhybansal (approved):
LGTM 💯

PR #19975, opened 2025-07-22
fix(AILabel): fix types

wkeese:
Refs #17232.

Fix some of the types for Toggletip and AILabel.

In particular, make Typescript know about the `autoAlignBoundary` parameter.

The basic idea is that Toggletip extends Popover, and AILabel extends Toggletip.   Therefore, instead of repeating the definitions for `align` etc., leverage Popover's Typescript types and PropTypes. 

Note though that AILabel's `revertActive` property is confusing, as we are either extending `IconButton` or `Toggletip`, and they are quite different.

Also, to do the extension, I had to fix the backward (circular) dependency from Popover --> Toggletip.

Finally, I had to jump through hoops to get the `PropTypes` working, in order to avoid Typescript errors.  Unfortunately, it seems like the *.d.ts files are generated from the `propTypes` declarations rather than the Typescript declarations like they should be.

### Changelog

**Changed**

- Type declarations for Toggletip and AILabel.

#### Testing / Reviewing

I tested the changes locally against my app, and they seem to work.

## PR Checklist


packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap
```diff
@@ -9568,40 +9654,76 @@ Map {
...
-            "right-bottom",
-            "right-top",
-            "top-start",
-            "top-end",
-            "bottom-start",
-            "bottom-end",
-            "left-end",
-            "left-start",
-            "right-end",
-            "right-start",
```
tay1orjones, line 9593: losing this level of granularity in the public api snapshot is a bummer but not a huge deal. We'll gain it back in the next major when we'll remove the deprecated ones and it'll go back to just an array.

tay1orjones (approved):
thank you!

heloiselui (approved):

PR #20014, opened 2025-07-24
feat(selectall): Update selectAll behaviours for filterable-multiselect

### LINKED ISSUE #19941: [Feature Request]: Extend SelectAll feature for FilterableMultiSelect to Web Components.
### The problem

This issue aims to extend the SelectAll feature request for FilterableMultiSelect that was implemented in React [in this PR](https://github.com/carbon-design-system/carbon/pull/19884) to Web Components.

Due to `FilterableMultiSelect` being implemented as an attribute (`filterable`) in WC, the SelectAll feature technically already exists for filterable since users can just use the `select-all` attribute along with `filterable`. However, it will not follow the expected behaviour it should since it is not explicitly implemented for filterable.

### The solution

The following behaviours need to be present in the WC version:

- When the selectAll item is clicked, all filtered items are selected, this means that when items = {one, two, three}. Filtering for "three" and clicking selectAll only selects three, since it is the only filtered item present. Deselections also behave in this manner with filters.

- The checked/indeterminate/empty state of the selectAll item changes dynamically based on the filter. If only half of the filtered items are selected it'll be indeterminate. But, if the text input changes to filter just one, already selected item, it will be checked.

- The selectAll item is only present if it is usable. Meaning, if there are no filtered items (i.e. the dropdown is empty), there will be no selectAll item, since it has nothing to select. The same behaviour applies for if the only selectable item(s) are disabled.

- The selectAll item is always at the top of the list for usability, no matter what the filter is.

- The selectAll item is not included in the count of selected items in the tag on the left-hand-side of the text-input.

### Examples

See https://github.com/carbon-design-system/carbon/pull/19884

### Package

@carbon/web-components

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

maradwan26:
Closes #19941

Following up on #19884, this PR updates the selectAll behaviour for the `multiselect[filterable]` component in Web Components to also have the following behaviours:

- When the selectAll item is clicked, all filtered items are selected, this means that when items = {one, two, three}. Filtering for "three" and clicking selectAll only selects three, since it is the only filtered item present. Deselections also behave in this manner with filters.

- The checked/indeterminate/empty state of the selectAll item changes dynamically based on the filter. If only half of the filtered items are selected it'll be indeterminate. But, if the text input changes to filter just one, already selected item, it will be checked.

- The selectAll item is only present if it is usable. Meaning, if there are no filtered items (i.e. the dropdown is empty), there will be no selectAll item, since it has nothing to select. The same behaviour applies for if the only selectable item(s) are disabled.

- The selectAll item is always at the top of the list for usability, no matter what the filter is.

- The selectAll item is not included in the count of selected items in the tag on the left-hand-side of the text-input.

 
### Changelog

**New**

- Added `FilterableWithSelectAll` story to showcase the new feature
- Fixed an issue where the link for `WithToggleTipLabel` wasn't showing the correct styles in storybook

**Changed**

- Modified `_selectDidChange()` to account for when `select-all` and `filterable` are both used so that only filtered items are selected/de-selected
- Modified `_handleInput()` to hide select-all item when there are no selectable items
- Modified `_computeSelectAllState()` so that the indeterminate/checked state is determined based on the filtered items

#### Testing / Reviewing

- Go to the new "Filterable with Select All" story in the deploy preview
- Behaviours listed above should work as expected
- CI should pass

## PR Checklist

### DIFF
--- a/packages/web-components/src/components/multi-select/__tests__/multi-select-test.js
+++ b/packages/web-components/src/components/multi-select/__tests__/multi-select-test.js
@@ -1669,6 +1669,10 @@ describe('cds-multi-select', function () {
         'cds-multi-select-item[is-select-all]'
       );
 
+      // no items are selected
+      expect(selectAllItem.hasAttribute('selected')).to.be.false;
+      expect(selectAllItem.hasAttribute('indeterminate')).to.be.false;
+
       // filter to “foo”
       input.value = 'foo';
       input.dispatchEvent(new Event('input', { bubbles: true }));
--- a/packages/web-components/src/components/multi-select/__tests__/multi-select-test.js
+++ b/packages/web-components/src/components/multi-select/__tests__/multi-select-test.js
@@ -1047,56 +1047,6 @@ describe('cds-multi-select', function () {
       expect(items[0].hasAttribute('is-select-all')).to.be.true;
     });
 
-    it('should handle select-all with filterable multiselect', async () => {
-      const el = await fixture(html`
-        <cds-multi-select label="test-label" select-all filterable>
-          <cds-multi-select-item is-select-all
-            >Select All</cds-multi-select-item
-          >
-          <cds-multi-select-item value="apple">Apple</cds-multi-select-item>
-          <cds-multi-select-item value="banana">Banana</cds-multi-select-item>
-          <cds-multi-select-item value="cherry">Cherry</cds-multi-select-item>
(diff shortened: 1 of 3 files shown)


ariellalgilmore (approved):
LGTM!

2nikhiltom (approved):

PR #20018, opened 2025-07-25
fix: introduce hidelabel property to table rows

### LINKED ISSUE #20017: [a11y]: The <label> element for checkbox throws error for table with checboxes
### Package

@carbon/web-components

### Browser

Chrome

### Operating System

_No response_

### Package version

2.27.2

### React version

_No response_

IBM Equal Access Accessibility Checker

### Assistive technology

_No response_

### Description

The error msg `The <label> element does not have descriptive text that identifies the expected input` is shown for the checkboxes of table header and rows 

### WCAG 2.1 Violation

_No response_

### Reproduction/example

Simply try running test on web components table with checkboxes

### Steps to reproduce

Just run the accessibility tool on the table with checkboxes. The issue will pop up

### Suggested Severity

None

### Code of Conduct

r1shabhsharma:
Closes https://github.com/carbon-design-system/carbon/issues/20017

For a table with checkboxes, the accessibility issue was being shown because we had an empty label field for all the checkboxes. The changes in PR are:
- **To have some value in the label for checkboxes.** This was done to follow a similar pattern from `carbon-components-angular` repo
- **To introduce a new label for showing/hiding a label.** As we have some value in label, we might not want to show it as a default behaviour. By default the label stays hidden.
- Updated the html configs for checkbox to show the label only when both `hideCheckbox` and `hideLabel` are set to false.

**Before**

Screenshot 2025-07-25 at 6 54 01 PM

**After**

Screenshot 2025-07-25 at 6 52 44 PM

## PR Checklist


packages/web-components/src/components/data-table/table-row.ts
```diff
@@ -350,7 +357,7 @@ class CDSTableRow extends HostListenerMixin(FocusMixin(LitElement)) {
    * The `aria-label` attribute for the `<label>` for selection.
    */
   @property({ attribute: 'selection-label' })
-  selectionLabel = '';
+  selectionLabel = 'Select row';
```
tay1orjones, line 363: I think only this is needed to fix the issue.

tay1orjones (changes requested):
The label should always be hidden though, right? That's what we have for react.

https://github.com/carbon-design-system/carbon/blob/eeda36c23468a5022f0ab63d77ea12e46ee8e6b8/packages/react/src/components/DataTable/TableSelectRow.tsx#L126-L141

https://github.com/carbon-design-system/carbon/blob/eeda36c23468a5022f0ab63d77ea12e46ee8e6b8/packages/react/src/components/InlineCheckbox/InlineCheckbox.tsx#L136-L146

packages/web-components/src/components/data-table/table-row.ts
r1shabhsharma: Addressed
I got confused as we we were using `hideLabel` but at the same time we had a field to define the label. Thought we would like to keep this field optionally visible. But now clear

tay1orjones (approved):

r1shabhsharma:
Can we merge this PR? Or get some reviews please?
It has been open for a while now @Gururajj77 @emyarod

emyarod (commented):
looks good to me! just need to resolve some merge conflicts in the README and `.all-contributorsrc`

kennylam (approved):

PR #20059, opened 2025-07-30
Update FilterableMultiSelect.tsx to fix ShadowDOM

wfdcfs:
Closes #

{{short description}}

### Changelog

**New**

- Nothing new. This component previously worked in the shadowDOM. This PR is looking to reinstate that situation.

**Changed**

- FilterableMultiSelect component stopped working in ShadowDOM after v1.68.0. Since then I've been using 'patch-package' to manually update this file post installing of @carbon/react to add support for downshift props.

**Removed**

- Nothing removed. Support for downshift props re-instated

#### Testing / Reviewing

## PR Checklist

### DIFF
--- a/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
+++ b/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
@@ -677,6 +677,7 @@ export const FilterableMultiSelect = forwardRef(function FilterableMultiSelect<
     isItemDisabled(item, _index) {
       return (item as any)?.disabled;
     },
+    ...downshiftProps,
   });
   function stateReducer(state, actionAndChanges) {
     const { type, props, changes } = actionAndChanges;


wfdcfs:
I accept Developer Certificate of Origin

packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx
```diff
@@ -677,6 +677,7 @@ export const FilterableMultiSelect = forwardRef(function FilterableMultiSelect<
     isItemDisabled(item, _index) {
       return (item as any)?.disabled;
     },
+    ...downshiftProps,
```
tay1orjones, line 680: What mod did you make with patch-package? Just `...downshiftProps` onto `useCombobox` like this? I ask because to my knowledge this component has never done that. It's [not there in 1.67.0](https://github.com/carbon-design-system/carbon/blob/v11.67.0/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx). Is there something else in 1.68.0 that caused the break that you can find? Between those two versions there was only a single line changed in FilterableMultiSelect from [this PR](https://github.com/carbon-design-system/carbon/commit/9d5eea1e7645d5fbda220c3c55542d6115f16dde#diff-a1249d25d8548f5c2c1a100e428df6fb0f881e7f213b96994153fc4aede57f33R105)


This component is peculiar because two hooks from downshift are used, `useCombobox` and `useMultipleSelection`. `downshiftProps` is only passed to useMultipleSelection and is typed that way:

https://github.com/carbon-design-system/carbon/blob/715a06c270503f0f8ac00cafe1b011f7bf67b79f/packages/react/src/components/MultiSelect/FilterableMultiSelect.tsx#L166

If you need the ability to pass props to useCombobox, I'd suggest a slightly altered approach to avoid prop/type collision:

* Add `downshiftUseComboboxProps?: UseComboboxProps<ItemType>;`
* Add `downshiftUseMultipleSelectionProps?: UseMultipleSelectionProps<ItemType>;`
* Deprecate `downshiftProps` with `@deprecate` in the interface and `deprecate()` in the proptypes. This message should probably point folks to `downshiftUseMultipleSelectionProps` as that's what it was before.
* Update `MultiSelect.mdx` "downshiftProps" section to "Downshift props", alter to mention these new props and note that "the previous `downshiftProps` prop has been deprecated in favor of these"

PR #19636, opened 2025-06-12
fix(Menu): improve focus behavior for accessibility

heloiselui:
Closes #19551
Closes #16388

Fixes an accessibility issue where the `Menu` component in Carbon React did not correctly restore focus to the first focusable item if the user clicked outside the menu, for example. This broke keyboard navigation when re-entering the menu and did not align with expected ARIA behavior. 

Also fixes submenu focus behavior when navigating with `ArrowRight`, ensuring focus is correctly transferred to the submenu.

### Changelog

**New**

- Accessibility test to ensure the first focusable `MenuItem` receives focus on menu open
- Test to verify disabled items are skipped when determining initial focus

**Changed**

- `MenuItem` now checks if it’s the first focusable item and sets `tabIndex={0}` when needed
- Improved focus logic in `Menu` and `MenuItem` to better follow ARIA `Menu` guidelines
- Submenu now receives focus when opened via `ArrowRight` key, enabling correct `Esc`/`ArrowLeft` handling

**Removed**

- ~N/A~

#### Testing / Reviewing

- Go to the `React Deploy Preview` > `Menu` > `Default`
- Verify that focus automatically lands on the first enabled item in the main menu
- Click outside the menu, then press `Tab` and confirm that focus can return to the menu
- While you are in the main menu:
  - Navigate using `ArrowDown` and `ArrowUp`
  - Press `ArrowRight` on an item with a submenu (e.g., “Share with”) to open it
    -  Navigate using `ArrowRight` again, `ArrowDown` and `ArrowUp`
    - Press `Esc` to close the menu and ensure focus returns to `Share with`, or press `ArrowLeft` from inside the submenu to close it and return focus to the main menu.

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [X] Reviewed every line of the diff
- ~[ ] Updated documentation and storybook examples~
- [X] Wrote passing tests that cover this change
- [X] Addressed any impact on accessibility (a11y)
- [X] Tested for cross-browser consistency
- [X] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #20076, opened 2025-08-01
fix(menu): out-of-order keyboard nav when MenuItems added dynamically

### LINKED ISSUE #20004: [a11y]: Out-of-order keyboard navigation when MenuItems added dynamically
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.83

### React version

18

manual testing

### Assistive technology

none

### Description

# Background

React lets the application dynamically add/remove components.  So with code like this:

```jsx
<Menu>
        {items.map((item) => (
          <MenuItem key={item} label={item} />
        ))}
</Menu>
```

If you add a new item somewhere in the `items[]` array, React will create a new `MenuItem` component (i.e. a new `<li>` node) without disturbing the pre-existing ones.

# Problem

However, that dynamically added `MenuItem` breaks keyboard navigation.  Specifically, even if the dynamically added `MenuItem` is the first item in the `Menu`, it becomes the last navigable item with the down arrow key.

# Cause

In Menu.tsx, keyboard navigation is based on the order of the items in the `focusableItems[]` array, and `focusableItems[]` is based on `childContext.state.items[]`.

However, the order of the items in `childContext.state.items` is based on the order the `MenuItems` were *created*, not their logical order in the DOM.   Because each MenuItem registers itself when it's created, adding itself to the end of the `childContext.state.items[]` array.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://stackblitz.com/edit/github-ykbrpwcg-8pvggqsy?file=src%2FExample%2FExample.jsx

### Steps to reproduce

Just try pressing the "add menuitem" button and then navigation through the menu using up/down arrows.

Actually my test case works a little differently from my actual app.   In my actual app, the newly added MenuItem is last in the up/down arrow navigation, even though it's the second MenuItem in the list.  Not sure why the discrepancy.

### Suggested Severity

None

### Code of Conduct

maradwan26:
Closes #20004

There was an issue where adding `MenuItems` dynamically caused the keyboard navigation to iterate through `MenuItems` in the incorrect order. This PR fixes this bug for both React and Web Components. Some tab-indexing and focus issues have also been fixed for Web Components.

### React

**New**

- Added a test case to cover this issue

**Changed**

- Changed `registerItem` case in `MenuContext` to splice the list of items and insert the dynamically added item(s) in the correct index

**Removed**

- Removed `[isFocusable, setIsFocusable]` useState in `MenuItem` that was used to handle tab-indexing for MenuItem and just conditionally checked for disabled instead. Functionality should be the same.

### Web Components

**New**

- Allowed for menu-items to be tabbable, similarly to [#19636 ](https://github.com/carbon-design-system/carbon/pull/19636)for React
- Added `slotchange` event listener to call `_setActiveItems` and `_registerMenuItems` when new menu items are added
- Made `_setActiveItems` properly set menu items in the correct index

**Changed**

- Changed snapshot for menu-button to account for tab-index change

**Removed**

- Removed `tabindex = -1` being applied in multiple areas where it shouldn't

### Testing / Reviewing

- Go to `Bug Fix` and `Bug Fix Disabled` for both React and Web Components
- Add more menu items using the button
- Keyboard navigation should be in the correct order
- For Web Components, you should be able to return focus to the menu after clicking outside
- There should be no regression.

> [!NOTE]
> The two test stories `Bug Fix` and `Bug Fix Disabled` should be removed before merging

## PR Checklist


tay1orjones (approved):
LGTM!

kennylam (approved):

kennylam:
@maradwan26 LGTM, just need to remove test stories before it merges.

PR #20086, opened 2025-08-04
fix: add .js to datePicker rangePlugin import for ESM resolution

devadula-nandan:
Closes #

Fixes an import resolution issue caused by missing .js extension in the flatpickr rangePlugin import, which breaks under strict ESM environments like Webpack 5+.

Screenshot 2025-08-04 at 6 01 51 PM

### Changelog

**Changed**

- Added .js extension to flatpickr/dist/plugins/rangePlugin import in @carbon/web-components to fix ESM resolution in Webpack

#### Testing / Reviewing

Import the date-picker component and confirm it renders without build issues in a Webpack environment.

## PR Checklist


packages/web-components/src/components/date-picker/range-plugin.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 // @ts-nocheck
-import rangePlugin, { Config } from 'flatpickr/dist/plugins/rangePlugin';
+import rangePlugin, { Config } from 'flatpickr/dist/plugins/rangePlugin.js';
 import { Instance as FlatpickrInstance } from 'flatpickr/dist/types/instance';
 import { Plugin } from 'flatpickr/dist/types/options';
```
annawen1, line 11: should we add the `.js` extensions here too?

devadula-nandan: I don't think we need it for types, since it's not a runtime import. 🤔

tay1orjones (approved):
I guess the reason for this is the [`fullySpecified`](https://webpack.js.org/configuration/module/#resolvefullyspecified) config option (on by default)

flatpickr doesn't have `type: module` but `@carbon/web-components` does.

annawen1 (approved):

PR #20122, opened 2025-08-07
fix: fixed the break on datatable sortby

### LINKED ISSUE #20027: [Bug]: DataTable crashes on state.cellsById when calling renderProps sortBy
### Package

@carbon/react

### Browser

Chrome

### Package version

1.86.0

### React version

18.3.0

### Description

Upgrading to `carbon/react 1.86.0` or higher causes our UI to crash on the following line from the DataTable component on `state.cellsById`

```javascript
    const currentHeaders = Object.keys(state.cellsById).reduce((acc, cellId) => { 
```
Here is the crash stack
```
Uncaught TypeError: Cannot convert undefined or null to object
    at Object.keys (<anonymous>)
    at Space.js:24678:35
    at commitHookEffectListMount (react-dom.development.js:23184:26)
    at commitPassiveMountOnFiber (react-dom.development.js:24960:13)
    at commitPassiveMountEffects_complete (react-dom.development.js:24925:9)
    at commitPassiveMountEffects_begin (react-dom.development.js:24912:7)
    at commitPassiveMountEffects (react-dom.development.js:24900:3)
    at flushPassiveEffectsImpl (react-dom.development.js:27073:3)
    at flushPassiveEffects (react-dom.development.js:27018:14)
    at performSyncWorkOnRoot (react-dom.development.js:26110:3)
```

I isolated it to us calling sortBy from the renderProps. IE we are doing something like this:
```
<DataTable render={( { sortBy })=> {
   sortBy(someHeaderKey)

   return (<Table> .... </Table>)
}
```

We have been using this code for 3 years and it works on `1.85.0` so it is something specific to the update in `1.86.0` that has broken us. 

A colleague of mine has replicated a similar crash in [stackblitz on 1.86.0](https://stackblitz.com/edit/github-qikjysvv-wkfblepx?file=src%2FApp.tsx). You can see [this example on 1.85.0](https://stackblitz.com/edit/github-qikjysvv?file=src%2FApp.tsx) which is working with the same code. To trigger the crash, click the `Sort by xxx` buttons. From my colleague:

> The error is not an exact match, but i think the core issue is the same with the sortBy function removing certain properties from the table state causing them to be undefined

### Reproduction/example

https://stackblitz.com/edit/github-qikjysvv-wkfblepx?file=src%2FApp.tsx

### Steps to reproduce

Call `sortBy` from the DataTable renderProps

### Suggested Severity

### Application/PAL

IBM watsonx

### Code of Conduct

Gururajj77:
Closes #20027 

Fixes the custom `sortBy` render function that would throw undefined.

The problem:
- the previous state was not being preserved by the component's render function

```
 setState((prev) => getNextSortState(props, prev, { key: headerKey }));
 ```
- the previous sort state would be lost after a render because the render function doesn't have the access to return the new state by comparing it with the previous state.

The PR has the solution that preserves the state by spreading the previous state, applying the new sort state and returning it in the `handleSortBy` function.

### Changelog

**New**

- Tests to make sure this functionality doesn't break.

**Changed**

- The `handleSortBy` function now preserves the state by spreading the previous state, applying the new sort state and returns it.

#### Testing / Reviewing

- go to the test story added 
- click on the two buttons that can sort the rows as expected
- make sure it doesn't break the story like how the bug's stackblitz does.

### Note
- Remove the test story after two successful reviews.

## PR Checklist


heloiselui (approved):
LGTM!

tay1orjones (approved):
LGTM! @Gururajj77 just need to remove the test story

Issue #9208, opened 2021-07-15
Use Carbon tooltip instead of browser generated `title` tooltip for all components


griffindvs:
## Environment

> Operating system: MacOS Big Sur (11.4)

> Browser: Chrome 91

> Testing Tool: Manual

### What version of the Carbon Design System are you using?
`carbon-components: 10.38.0`
`carbon-components-react: 7.38.0`

### What did you expect to happen?
The `MultiSelect.Filterable`, `ToastNotification`, `InlineNotification`, and `NumberInput` Carbon React components each take prop inputs for text to be displayed in the default browser tooltip (internally using the `title` attribute). Some components just take a string while others take a function that returns a string (to allow for translation messages for singular/plural). These tooltips contain information that is not displayed elsewhere, so they should be resized when the browser is zoomed-in to allow those with varying levels of vision to read the text.

### What happened instead?
The default browser tooltips do not resize as the browser is zoomed-in. This means that when you go from 100% to 200% zoom on a webpage, the browser tooltip remains the same size. Because these tooltips are generated by the browser and merely generated from a `title` attribute, there is no workaround to force these tooltips to resize. This problem is noted in the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns) for the title attribute which notes that "if a tooltip effect is desired, it is better to use a more accessible technique" instead of this built-in tooltip. For Carbon, a more accessible technique could come in the form of integrating the custom Carbon tooltip into these components.

### What WCAG 2.1 checkpoint does the issue violate?
[Success Criterion 1.4.4 Resize Text](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html)

## Steps to reproduce the issue

1. Use the `InlineNotification` component as an example (though there are others).
2. Input a string for the `iconDescription` prop.
3. Build and view the webpage.
4. Zoom the browser to 200%.
5. The tooltip on the InlineNotification close button does not resize.

CodeSandbox:
https://codesandbox.io/s/carbon-tooltip-issue-n91ng?file=/src/index.js

## Additional information
Screenshot at 100% Zoom:
<img width="877" alt="Screen Shot 2021-07-15 at 2 07 30 PM" src="https://user-images.githubusercontent.com/18495025/125843738-89980e9c-7539-4597-bcc4-31ec244aabe1.png">

Screenshot at 200% Zoom:
[...]

PR #20154, opened 2025-08-11
feat: use title from rest prop on MenuItem 

### LINKED ISSUE #20090: [Feature Request]: Enable `MenuItem` to have a configurable `Tooltip`
### The problem

When you have a MenuItem component, I would like the ability to modify the tooltip shown when we hover above it. Right now, the title attribute of the inner div is being set to the label, but it would be good for that to be the default but if we pass on a parameter called title or tooltip or something then it would be updated with that. ~Right now if we try to set a title to the MenuItem component the inner div's title overrides whatever we set.~ _EDIT: This has been corrected in https://github.com/carbon-design-system/carbon/pull/20378_

### The solution

<MenuItem title="myToolTip" />

### Examples

_No response_

### Package

@carbon/react

### Application/PAL

_No response_

### Business priority

### Available extra resources

_No response_

### Code of Conduct

koyita:
Closes #20090 

MenuItem now shows title prop passed down when on hover, instead of the label prop.

### Changelog

**New**

-  none

**Changed**

- title={label}>. -> title={rest?.title || label}>

**Removed**

- none

#### Testing / Reviewing

- Create a new Menu with a MenuItem, 
- pass down a title prop which is different to the label prop, 
- verify by hovering mouse over MenuItem, tooltip should be title prop, not label.

## PR Checklist


koyita:

tay1orjones (commented):
What is the use case for the title to be different than the label?

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -272,7 +272,7 @@ export const MenuItem = forwardRef<HTMLLIElement, MenuItemProps>(
           <Text
             as="div"
             className={`${prefix}--menu-item__label`}
-            title={label}>
+            title={rest?.title || label}>
```
tay1orjones, line 275: I don't think it's a good idea to encourage the `title` be something different than the label. How would mobile and tablet users, or a keyboard only user access the text?

* https://www.tpgi.com/using-the-html-title-attribute-updated/
* https://inclusive-components.design/tooltips-toggletips/

We're hoping to eventually move away from `title` altogether, https://github.com/carbon-design-system/carbon/issues/9208

koyita: Hi, @tay1orjones  I've read the suggested reading and here is my take:

For the needs of my application in specific we wanted to add a tooltip to a MenuItem to give the user a slightly longer description of what we intend that button to do. In our case it opens a report in another page, giving the user slightly more context as to what type of report this is can avoid extra clicks for them.

 The reason for this is issue is: when I tried to just add the title argument and see if it would take I noticed that it did part way. In the way we are using MenuItem we are also including an icon using the renderIcon prop. When I hover the mouse over the icon I get the text I set as a title, but when I hover over the label text I just get the label text repeated.


```
<MenuItem
            label="Label"
            title="Longer Label"
            renderIcon={Add}
          ></MenuItem>
```

**Depending on where we hover the mouse this is what the tooltip shows:**

**On the icon:**

image

**on the label text:**

image

**I would like to be able to define in the title prop what to show on all of the component.**

 I find that having conflicting tooltips show up for the same component is confusing, that is reason I opened this issue, I would like to see the same tooltip all over the component instead of mismatching ones. If the dev doesn't pass through any titles, then its alright to have the label as the tooltip.

I understand the the title attribute in itself has evolved into only being useful for more traditional desktop or keyboard and mouse web applications. Which isn't quite compatible with mobile and tablet users, but understanding this, there are still many applications (like mine) which are intended to only be used in this way. Regarding the accessibility issues it may present maybe a good compromise would be that if a dev doesn't pass down an aria label but they do pass down a title, then we could set it as both?

Finally, the proposal you mention for moving away from title sounds like a good alternative but is it anywhere close to done? If not I think that still having access to the expected html attributes is important, at least meanwhile.

tay1orjones: Thanks for the additional info! I believe `Longer label` is shown because the icon is inheriting that `title` from the parent `li` (because `...rest` is on the `li`), whereas the `Text` has the `title={label}` so doesn't inherit it.

This can be fixed by moving `title={label}` from `Text` to the `li`. With that change every child will inherit the same value and it'll be consistent.

I found the `title` [was added](https://github.com/carbon-design-system/carbon/pull/15236) for really only one purpose - surfacing truncated `label`s. When the text truncates, the `...` is a visual indicator that there is more content. In your scenario there is no visual indication, but there could be. Why not add that additional helpful context to the label and let it truncate?

Using it for truncation is still subject to all the a11y/usability issues I linked earlier, but it's the best stopgap measure until a "real" tooltip or toggletip is implemented.

rcwakelin:
@tay1orjones 

**Background:**  In the past each engineer "type" would have their own softeard

We are creating new software (not web marketing pages) for researching and manufacturing high end semiconductor chips. Existing applications have been highly customized based on the needs of the individual engineers and semiconductor fab. (In many cases engineers write their own code)

There are two monitor sizes that the software is being designed for. It is not intended (or possible to be) for mobile or tablet use. 

**Audience:** Our audience is primarily male and older. These engineers need the visual acuity to review large numbers of photographs for "errors". We do not have visuals "issues" that are normally applied to web page (and international accessibility guidelines) other than color blindness. The rate seems high even for our audience. 

We have several "language/naming issues" that need to be overcome. First are engineers who don't have English as a their first language. They need more words that what appear menus, column heads, or names for x/y axis in charts. 

The second issue are the different "types" of engineers. A "PLY" Engineer and look at a value from a data base and use a "nickname" that their team applied to the data. A "WFT" Engineer's team look at the same value and use a different "nickname" they came up for the data. 

The third issue is the "work around" for the naming conventions. Each data base was developed on it own and no two databases use the same naming convention. To keep everything fair to different types engineers using the data, it was decided that the naming conventions would come from the database holding the information, even if it looks wrong. (its not feasible to restructure existing databases)

IBM pioneered the semiconductor and the industry has many IBM terms like "PLY" which only makes sense if your familiar with IBM semiconductor history. 

**Issue:** We rely heavily on tools tips and "?" (linking users into documentation) for more information. The current implementation of menu tool tips and looks broken.  We need a more consistent expression of tools from menus, to table heads to charting. This also meets our client contractual obligations. 

Attaching list of tools tips for the menus
Screenshot 2025-08-14 at 3 11 40 PM

Attaching screen shot of current implementation 
Screenshot 2025-06-30 at 2 13 54 PM

Accessible tooltips must be perceivable, operable, understandable, and robust. As you can see... current implementation is not enough. 

https://accessiblyapp.com/blog/tooltip-accessibility/

Happy to discuss at a Carbon call, let me know and I'll add to the sign up list.

tay1orjones:
> Happy to discuss at a Carbon call, let me know and I'll add to the sign up list.

@rcwakelin Yeah absolutely! Design office hours is tomorrow https://ec.yourlearning.ibm.com/w3/event/10522523

tay1orjones:
Hey thanks again for bringing this up! After discussing with the IBM Accessibility team, using `title` as a makeshift tooltip for anything other than the (truncated) label text is problematic and not something to be encouraged. The more correct solution would appear to be enabling MenuItem to use or be wrapped with the actual Carbon Tooltip. I've updated the original issue https://github.com/carbon-design-system/carbon/issues/20090 to reflect this new goal. 

I did put up a PR though to fix the inheritance issue that surfaced higher up in this thread, #20378. This should avoid mismatched titles going forward to always use the label text.

Also thanks for including such detailed use case information! Addressing this design gap without a tooltip will no doubt need some alternative design solutions. We're here to help - if you'd like to discuss and review ideas, attending design office hours is a great place to start.

With this I'm going to close this out but if you have any questions or concerns please let us know.

PR #20155, opened 2025-08-11
feat(dismissible-tag): make tooltip position configurable

### LINKED ISSUE #19891: [a11y]: Make the tooltip position configurable in DismissibleTag
### Package

@carbon/react

### Browser

_No response_

### Operating System

_No response_

### Package version

1.86.0

### React version

19.1.0

None, bumped into it during development

### Assistive technology

no

### Description

You are forcing the tooltip to be positioned at the bottom in [`DismissibleTag.tsx:190`](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tag/DismissibleTag.tsx#L190) and it makes for a bad time when overflow comes into play.

Image

### WCAG 2.1 Violation

Dunno, but the tooltip text is not visible fully

### Reproduction/example

Not much to do here, the description says it all.

### Steps to reproduce

Not much to do here, the description says it all.

### Suggested Severity

### Code of Conduct

maradwan26:
Closes #19891

The tooltip position for `DismissibleTag` was hardcoded to 'bottom', causing overflowing issues. This PR adds a requested prop/attribute that allows the user to configure the tooltip position for both React and Web Components.

### React

**New**

- Added `dismissTooltipAlign` prop that controls the dismiss tooltip alignment (defaults to `'bottom'`)
- Added controls for `dismissTooltipAlign` to React storybook

### Web Components

**New**

- Added `dismiss-tooltip-align` attribute that controls the dismiss tooltip alignment (defaults to `'bottom'`)
- Added controls for `dismissTooltipAlign` to WC storybook

#### Testing / Reviewing

- Go to `Tag/Dismissible` in both React and Web Components deploy preview
- `dismissTooltipAlign` control should work as expected

## PR Checklist


packages/react/src/components/Tag/DismissibleTag.tsx
```diff
@@ -45,6 +46,11 @@ export interface DismissibleTagBaseProps {
    */
   disabled?: boolean;
 
+  /**
+   * Specify the tooltip alignment for the dismiss button
+   */
+  dismissTooltipAlign?: PopoverAlignment;
```
tay1orjones: Convention across the codebase looks more centered on the name `tooltipAlignment`. With the `dismissTooltipLabel` prop already existing though it sort-of creates a conflicting convention.

I wish both these were `tooltipLabel` and `tooltipAlignment` but to avoid churn on the existing prop we could instead just split it down the middle and do `dismissTooltipAlignment`. What do you think?

```suggestion
  dismissTooltipAlignment?: PopoverAlignment;
```

maradwan26: Yeah that makes sense, I've renamed `dismissTooltipAlign` to `dismissTooltipAlignment` and `dismiss-tooltip-align` to `dismiss-tooltip-alignment`

tay1orjones (approved):
LGTM!

riddhybansal (approved):

PR #20239, opened 2025-08-14
fix(pagination): fix crash when using indefinite amount of pages

### LINKED ISSUE #19799: [Bug]: Pagination component crashes when trying to use indefinite pagination
### Package

@carbon/web-components

### Browser

Firefox

### Package version

latest

### React version

n/a

### Description

When trying to use the Pagination component without specifying the number of items nor pages, runtime errors occur:

`RangeError: invalid array length`

Attempting to set no. pages to the current page does not work either in our application. Setting the no. pages to the current page + 1 (so the dropdown has the next value ready) works but requires setting the dropdown to disabled, which is less than ideal.

### Reproduction/example

https://stackblitz.com/edit/github-kckev3uz

### Steps to reproduce

See the reproduction example. Click to the next page, observe runtime errors in console.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

maradwan26:
Closes #19799

This PR fixes a bug where the pagination component would crash when using an indefinite amount of pages, throwing a runtime error: `RangeError: invalid array length`.

### Changelog

**New**

- Added `const totalPagesSafe` variable, which should not be undefined, to be used instead of `totalPages` to create the array since `totalPages` had cases where it would be undefined

**Changed**

- Changed `Array.from(new Array(totalPages)..` to `Array.from(new Array(totalPagesSafe)...` in the `render()`

#### Testing / Reviewing

- Go to `Unknown pages and items` in deploy preview
- Click the right arrow to traverse through pages
- The component should not crash

## PR Checklist


ariellalgilmore (commented):
Looks good! Maybe we can just change the story name to match react though: https://react.carbondesignsystem.com/?path=/story/components-pagination--pagination-unknown-pages

maradwan26:
@ariellalgilmore 
Sounds good! I've pushed the changed story name

ariellalgilmore (approved):
awesome! thanks!

heloiselui (approved):
LGTM!

PR #20264, opened 2025-08-19
feat(toggle): update WC toggle functionality to match React

### LINKED ISSUE #20106: [a11y]: Accessibility violation due to typo in Toggle button web component
### Package

@carbon/web-components

### Browser

_No response_

### Operating System

_No response_

### Package version

2.33.0

### React version

_No response_

IBM Equal Access Accessibility Checker latest deployment

### Assistive technology

_No response_

### Description

A [typo](https://github.com/carbon-design-system/carbon/blob/a10d9b3823cf01009b37966405a5eb0e1bb7caee/packages/web-components/src/components/toggle/toggle.ts#L149) in the `cds-toggle` component is triggering the below accessibility violation. 

Image

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://web-components.carbondesignsystem.com/?path=/story/components-toggle--default

### Steps to reproduce

1. Open the Carbon Web Components Storybook.
2. Navigate to the default story of the Toggle button component.
3. Run an accessibility scan using the IBM Accessibility Checker browser extension.

### Suggested Severity

None

### Code of Conduct

### LINKED ISSUE #20107: React|WC Parity: Toggle
See parent issue https://github.com/carbon-design-system/carbon/issues/17992 for acceptance criteria.

Every PR needs a ux and visual review from the design team to ensure consistency. Please be sure PRs have [status: visual review 🔍](https://github.com/carbon-design-system/carbon/labels/status%3A%20visual%20review%20%F0%9F%94%8D) applied and request a review from @carbon-design-system/design

maradwan26:
Closes #20107 
Closes #20106 

This PR updates the `cds-toggle` component's functionality and features to match React and fixes some styling bugs.

### Changelog

**New**

- Added `cds-toggle-skeleton` component and its respective story `Toggle/Skeleton`
- Fixed an alignment styling bug when using `read-only` for both regular and small sizes
- Added `Toggle/With Accessible Labels` story
- Added `aria-labelledby` attribute and its respective behaviour and logic

**Changed**

- Renamed `readonly` storybook control to `readOnly` to abide by consistent camelCase usage
- Changed the functionality and logic of `hideLabel` to match React, it should:
    - when `label-a` and `label-b` are provided: _hide the label text and move it to the toggletext_
    - when there is no label text: _hide the toggle text_
- Changed `label-a` and `label-b` default values to `On` and `Off`, respectively

**Removed**

- Removed `aria-label` attribute in favour of `aria-labelledby`

#### Testing / Reviewing
- Go to `Toggle` in WC Deploy Preview
- `hideLabel` behaviour should match React
- `readOnly` styling should be fixed
    - _small_: checkmark should be properly aligned
    - _regular_: toggle circle should be properly aligned
- `Skeleton` and `With Accessible Labels` stories should work and match React
-  There should be no regression 

## PR Checklist

- ~~[] Wrote passing tests that cover this change~~


packages/web-components/src/components/toggle/toggle.ts
```diff
@@ -83,11 +83,17 @@ class CDSToggle extends HostListenerMixin(CDSCheckbox) {
...
+   */
+  @property({ type: String, attribute: 'aria-labelledby' })
+  ariaLabelledby?: string;
+
   /**
    * The text for the checked state.
    */
   @property({ attribute: 'label-a' })
-  labelA = '';
+  labelA = 'On';
```
heloiselui, line 96:
```suggestion
  labelA = 'Off';
```

packages/web-components/src/components/toggle/toggle.ts
```diff
@@ -111,7 +117,7 @@ class CDSToggle extends HostListenerMixin(CDSCheckbox) {
    * The text for the unchecked state.
    */
   @property({ attribute: 'label-b' })
-  labelB = '';
+  labelB = 'Off';
```
heloiselui, line 120:
```suggestion
  labelB = 'On';
```

heloiselui (commented):
Looking very good, just a small detail: 

- Pressing Enter does not toggle on/off like in React. The click handler should be on the button.

- Should we also update the Overview page to look similar to React, or are we only aligning functionality? I know we're discussing creating a template, but in the meantime, should we adjust things to make it look as close as possible?

- With Accessible Labels: In React, when I click on the “On/Off” text or the "Label", the toggle opens/closes. In Web Components it doesn’t.

maradwan26:
@heloiselui 

> Pressing Enter does not toggle on/off like in React. The click handler should be on the button.

Added ✅ 

>Should we also update the Overview page to look similar to React, or are we only aligning functionality? I know we're discussing creating a template, but in the meantime, should we adjust things to make it look as close as possible?

I think we can adjust things in the meantime, I've matched the `Overview` for the Toggle ✅ 

> With Accessible Labels: In React, when I click on the “On/Off” text or the "Label", the toggle opens/closes. In Web Components it doesn’t.

Fixed, this should also work now for External toggle label in the deploy preview ✅ 


> labelA = 'On'; old
  labelA = 'Off'; suggested change

I'm not sure if this should be done since it may be a breaking change to users. The React and WC versions of this component have the usage of labelA and labelB flipped. React has[ labelA for unchecked/off](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Toggle/Toggle.tsx#L45-L48) and WC [has labelA for checked/on](https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/toggle/toggle.ts#L86-L90). Reversing it now could be breaking since users will have to flip the labels back.

heloiselui (approved):

packages/web-components/src/components/toggle/toggle.ts
```diff
@@ -111,7 +117,47 @@ class CDSToggle extends HostListenerMixin(CDSCheckbox) {
...
+    const doc = this.ownerDocument || document;
+    const found: HTMLLabelElement[] = [];
+
+    if (this.id) {
+      found.push(
+        ...(Array.from(
+          doc.querySelectorAll(`label[for="${this.id}"]`)
+        ) as HTMLLabelElement[])
+      );
+    }
```
kennylam:
```suggestion
    const found = this.id
      ? [...doc.querySelectorAll<HTMLLabelElement>(`label[for="${this.id}"]`)]
      : [];
    }
```

`querySelectorAll` already returns a list, so you probably don't need the `Array.from`. And using `<HTMLLabelElement>` lets you avoid the manual casting.

packages/web-components/src/components/toggle/toggle.ts
```diff
@@ -111,7 +117,47 @@ class CDSToggle extends HostListenerMixin(CDSCheckbox) {
...
+    this._externalLabels = Array.from(new Set(found));
+    this._externalLabels.forEach((lbl) => {
+      lbl.addEventListener('click', this._onExternalLabelClick);
+    });
+  }
+
+  connectedCallback() {
+    super.connectedCallback?.();
+    this._attachExternalLabels();
+  }
```
kennylam, line 155:
```suggestion
  connectedCallback() {
    super.connectedCallback?.();
    this._attachExternalLabels();
  }
  
  disconnectedCallback() {
    super.disconnectedCallback?.();
    this._externalLabels.forEach(lbl =>
      lbl.removeEventListener('click', this._onExternalLabelClick)
    );
  }
```

We could add some cleanup to this too.

packages/web-components/src/components/toggle/toggle.ts
```diff
@@ -138,32 +184,46 @@ class CDSToggle extends HostListenerMixin(CDSCheckbox) {
...
+    } else {
+      stateText = checked ? labelA : labelB;
+    }
+
+    const labelId = id ? `${id}_label` : undefined;
+
+    const hasLabelText = (this.labelText ?? '') !== '';
+
+    const ariaLabelledby =
+      this.ariaLabelledby ?? (hasLabelText && labelId ? labelId : undefined);
```
kennylam:
```suggestion
      this.ariaLabelledby ?? (hasLabelText && labelId);
```

I think this is the does the same?

kennylam (commented):
@maradwan26 Nice! Just a few suggestions.

packages/web-components/src/components/toggle/toggle.ts
maradwan26: Yes, it'll have the same function thanks for pointing that out

maradwan26:
@kennylam 

Thanks for pointing these out! I've pushed all requested cleanup changes

laurenmrice (changes requested):
Hey @maradwan26 this is looking good! Just had a couple comments below:

### WC

I found some bugs in the Default story, which I am not sure is part of this PR but I am going to list it here and we can move it to another issue if needed.

**Default story**
- The “checked” prop in WC is referred to as the “toggled” prop in React. I think we should align that prop to be the same name for both WC and React and call it “toggled” unless there was a reason for the prop name difference.
- I would follow the same size prop names that are in React. Regular size should be Medium size.
- Read-only prop (small size): The checkmark should not be visible while toggled on, the circle should just remain the `$icon-primary` color.

**Skeleton story and With Accessible Labels story**
- There are no controls present in the Controls tab of this story like there are in React. Just wanted to make sure that was intentional.

**Small toggle story**
- The label for the toggle should just say “Label”.
- There are no controls present in the Controls tab of this story like there are in React. Just wanted to make sure that was intentional.

kennylam (approved):
Code LGTM!

maradwan26:
@laurenmrice 
> The “checked” prop in WC is referred to as the “toggled” prop in React...

So this is because toggle in WC inherits from checkbox so it also inherits its "checked" prop.[ I've opened an issue](https://github.com/carbon-design-system/carbon/issues/20353) to re-factor the toggle component so that it is no longer a child of checkbox, as we don't feel that it is necessary for toggle to inherit from checkbox.

> I would follow the same size prop names that are in React. Regular size should be Medium size.

Done ✅ 

> Read-only prop (small size): The checkmark should not be visible while toggled on, the circle should just remain the $icon-primary color.

Can you open an issue for this with a design spec? Because its the same for React as well

> Skeleton story and With Accessible Labels story, There are no controls ...

This is intentional for both React and WC to not have controls for `Skeleton` and `With Accessible Labels`. I've also emptied out the argTypes from the table in `Skeleton` with the exception of className.

> Small toggle story
> The label for the toggle should just say “Label”.

Done ✅ 

>There are no controls present in the Controls tab of ...

Done ✅ There were no controls for React either for this story so I've added controls there as well.

laurenmrice (approved):

PR #20266, opened 2025-08-20
fix(modal): update `preventCloseOnClickOutside` behavior to match the design spec

### LINKED ISSUE #20274: [Bug]: preventCloseOnClickOutside prop in ComposedModal does the opposite of what it logically should
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.89.0

### React version

v18.2.0

### Description

in ComposedModal https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview, when I place the prop preventCloseOnClickOutside, it makes close on clicking outside possible, and setting preventCloseOnClickOutside={false}, prevents close on click outside

### Reproduction/example

https://stackblitz.com/edit/dxmbykdv-1e5f19fj?file=src%2FApp.jsx

### Steps to reproduce

issue with truthy:
1) open modal
2) click outside of modal
3) close on click outside is not prevented, despite preventCloseOnClickOutside inside ComposedModal

issue with {false}
1) add ={false} to have preventCloseOnClickOutside={false}
2) open modal
3) click outside of modal
4) close is prevented, despite preventCloseOnClickOutside={false}

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

tay1orjones:
Follow-up to https://github.com/carbon-design-system/carbon/pull/19590
Closes #20274

I don't believe there is an issue created for this yet, but it was [surfaced in slack](https://ibm-studios.slack.com/archives/GD8JG6JR3/p1753374084075449) that with this code
```js
<ComposedModal open preventCloseOnClickOutside={true}>
  <ModalHeader>
    <p>header</p>
  </ModalHeader>
  <ModalFooter>
    <p>footer</p>
  </ModalFooter>
</ComposedModal>
```

The modal would still close when a click happens outside the modal, even though `preventCloseOnClickOutside` was passed. 

## Context

The changes in #19590 were guided by the root issue #13712. In the time since that was open in ~2023 there's been a shift in the html spec regarding this "outside-click-to-close" functionality. It's called "light dismiss" in the spec. `<dialog>`s that are non-modal (passive) have light dismiss by default. `<dialog>`s that are modal can also now have light dismiss, but it's opt in https://github.com/whatwg/html/issues/9373 

In our components, `preventCloseOnClickOutside` is the mechanism to control this "light dismiss". Conversely, Carbon design specs outline that non-passive modals should not be dismissible by outside clicks. 

Considering the movement in the spec, how it relates to our implementation, and discussing with other devs, we landed on the following for each type of modal:

* passive - close on outside by default
* passive w/ `preventCloseOnClickOutside` - do not close outside
* passive w/ `preventCloseOnClickOutside={false}` - close on outside
* non-passive - do not close on outside by default
* non-passive w/ `preventCloseOnClickOutside` - do not close on outside
* non-passive w/ `preventCloseOnClickOutside={false}` - close on outside with warning

> [!IMPORTANT]
> The main change here in this PR is that consumers can opt to break from the spec and still close on outside clicks with `preventCloseOnClickOutside={false}` in both `ComposedModal` and `Modal`

### Changelog

**Changed**

- Revised `preventCloseOnClickOutside` implementation in Modal and ComposedModal
- Updated warnings for when `preventCloseOnClickOutside={false}`
- Refactored tests for both
- Updated documentation for both to clarify this situation

#### Testing / Reviewing

- In the react storybook under ComposedModal there are 6 test stories that match the expected behavior listed above. 
  - With these you can test the original bug is fixed as well as the new updates for the opt-in w/ `preventCloseOnClickOutside={false}`
  - I don't think the deploy previews will show the warnings since it's technically not a dev environment. They'll pop locally though.

## PR Checklist

> [!CAUTION]
> Test stories need removed before merging


maradwan26 (approved):
LGTM

emyarod (approved):
looks good to me!

PR #20267, opened 2025-08-20
fix(CheckboxGroup): propagate props to Checkbox children

### LINKED ISSUE #20260: [Bug]: A `readOnly` `Checkbox` can be checked and unchecked
### Package

@carbon/react

### Browser

_No response_

### Package version

v1.84.0

### React version

_No response_

### Description

Users can check and uncheck a `readOnly` `Checkbox`.

https://github.com/user-attachments/assets/21658f51-2649-4228-9d1b-3e2c0290699d

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-checkbox--default&args=readOnly:!true

### Steps to reproduce

1. Go to the reproduction.
2. Select a `Checkbox` and observe the visual change indicating that it has been selected.

### Suggested Severity

None

### Application/PAL

DataPower

### Code of Conduct

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/20260

Propagated `CheckboxGroup` props to `Checkbox` children.

### Changelog

**New**

- Propagated `CheckboxGroup` props to `Checkbox` children.

#### Testing / Reviewing

It crossed my mind that I shouldn't have to hardcode the three properties that need to be propagated, since those properties may change in the future. However, I'm not sure there is a dynamic way to gather the properties that do and don't need to be propagated.

```sh
yarn test packages/react
```

## PR Checklist


packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -89,6 +96,44 @@ const CheckboxGroup: React.FC<CheckboxGroupProps> = ({
...
+      ) {
+        childProps.readOnly = readOnly;
+      }
+
+      if (
+        typeof warn !== 'undefined' &&
+        typeof child.props.warn === 'undefined'
+      ) {
+        childProps.warn = warn;
+      }
```
Gururajj77: Everything looks good, just a small improvement here




```suggestion
    const childProps = Object.assign(
  {},
  typeof invalid !== 'undefined' && typeof child.props.invalid === 'undefined' && { invalid },
  typeof readOnly !== 'undefined' && typeof child.props.readOnly === 'undefined' && { readOnly },
  typeof warn !== 'undefined' && typeof child.props.warn === 'undefined' && { warn }
);
```

Gururajj77 (changes requested):

packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
adamalston: > just a small improvement here

I don't see how that suggestion is an improvement. That code would result in a type error and has worse performance characteristics. I'm willing to update the implementation to be more concise, but I'd prefer a different approach than the one suggested.

Gururajj77 (approved):
Looks good now, thanks.

tay1orjones (approved):
Hardcoded's fine imo 👍 LGTM

PR #20304, opened 2025-08-21
fix: add linting comments

### LINKED ISSUE #20071: Turn on `.ts`/`.tsx` processing in eslint config
Right now eslint doesn't process typescript files. The reason for this from #19658: 

> Locally, doing so results in [#18991 (comment)](https://github.com/carbon-design-system/carbon/issues/18991#issuecomment-2770412818) and a lot of other problems getting caught which is what we want.

This was brought up again in https://github.com/carbon-design-system/carbon/pull/20060 and I suggested the following as one idea for tackling this:

1. Enable typescript file processing
1. Put ignore comments on each failure and include a comment with this issue url. As granular as possible: 
   * Rule-specific ignores instead of global/wildcard ignores
   * Prefer single-line ignores over multi-line/range ignores
   * Entire file ignores (still for individual rules only) as an absolute last resort
1. Submit this in a PR
1. Open a new issue (sub issue of this one) for removing all the ignores and chew away at them over time by searching for this original/parent issue url.
   * As we chew away at them, we remove the issue url and so treat this as a sort of checklist

This aims for benefits of:
* Gaining immediate lint coverage for all the .ts/.tsx code that doesn't get ignores applied
* Iterative, non-blocking work that can be put into small, more reviewable PRs instead of a long running branch with huge PR(s)
* Not having to manually sync an issue checklist with each ignore item that remains. We just issue PR after PR that can branch/stack as needed.

riddhybansal:
Closes #20071 

This PR enables ESLint processing for all TypeScript files (.ts and .tsx) across the codebase. Previously, these files were excluded from the linting process, which meant that many code quality and style issues were not being caught.

To avoid blocking development with a large number of linting errors, this PR takes a strategic approach. It automatically adds granular eslint-disable comments for all existing failures. Each comment includes a reference to this issue, creating a clear checklist for future, iterative cleanup work.

### Changelog

**New**

- Updated ESLint configuration to include TypeScript support and new rules. All existing linting errors in .ts and .tsx files are temporarily suppressed with inline comments.
- Some eslint issues are already fixed in the same PR as  Prettier's auto-formatting would have otherwise inserted a new line, making the comment ineffective.

#### Testing / Reviewing

There should not be any lint issues. Linter should not fail.

## PR Checklist


heloiselui (approved):
I ran it locally and everything looks good! Thanks for doing this, @riddhybansal!

tay1orjones (approved):

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -106,346 +106,355 @@ export interface MenuProps extends React.HTMLAttributes<HTMLUListElement> {
...
-  );
-
-  // Getting the width from the parent container element - controlled
-  let actionButtonWidth: number;
-  if (containerRef?.current) {
-    const { width: w } = containerRef.current.getBoundingClientRect();
-    actionButtonWidth = w;
-  }
+// eslint-disable-next-line react/display-name -- https://github.com/carbon-design-system/carbon/issues/20071
+const Menu = forwardRef<HTMLUListElement, MenuProps>(
```
tay1orjones: Maybe look at this one - I think it's impacting coverage for some reason

tay1orjones: Maybe this should be set to true https://eslint.org/docs/latest/rules/prefer-arrow-callback#allownamedfunctions

PR #20306, opened 2025-08-21
fix(IconTab): added disabled prop to IconTabProps

### LINKED ISSUE #20261: IconTabProps is missing the 'disabled' property
### Package

@carbon/react

### Browser

Chrome

### Package version

1.85.0

### React version

19.0.0

### Description

The documentation for Tabs → [Icon Tabs Only](https://react.carbondesignsystem.com/?path=/docs/components-tabs--overview#icon-tabs-only) shows usage of the disabled prop on <IconTab>.

Example from Docs:

`  <IconTab
      disabled
      label="Analyze"
    >`

However, when using this in my Next.js project, the compiler reports:
`Property 'disabled' does not exist on type 'IntrinsicAttributes & IconTabProps & RefAttributes<HTMLDivElement>'`

This means the disabled prop is not currently typed in IconTabProps, even though the docs (and runtime behavior) suggest it is supported.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/docs/components-tabs--overview#icon-tabs-only

### Steps to reproduce

- Install @carbon/react@1.85.0

- Import IconTab from @carbon/react

- Try to use <IconTab disabled label="...">...</IconTab>
 
- Observe TypeScript error.

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Vignesh-Loganathan-IBM-1:
Closes #20261

Added the disabled prop to the IconTabProps

### Changelog

**New**

- Added disabled prop to the IconTabProps

#### Testing / Reviewing

- [ ] In a typescript component use disabled in IconTab Component.

## PR Checklist


Vignesh-Loganathan-IBM-1:

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -1574,6 +1579,7 @@ const IconTab = React.forwardRef<HTMLDivElement, IconTabProps>(function IconTab(
     enterDelayMs,
     leaveDelayMs,
     label,
+    disabled = false,
```
adamalston: I have two questions:

1. Why does the prop need a default value?
2. Why does the prop need to be destructured here instead of being passed to `Tab` via the `rest` parameter?

The issue is a type issue. I don't think it requires a code change within the component.

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -1545,6 +1545,11 @@ export interface IconTabProps extends DivAttributes {
    */
   defaultOpen?: boolean;
 
+  /**
+   *  Specify whether your Tab is disabled.
```
adamalston: Should this comment be updated to refer to `IconTab` instead of `Tab`?

adamalston (commented):
Is there an existing test that has a `disabled` `IconTab`?

packages/react/src/components/Tabs/Tabs.tsx
Vignesh-Loganathan-IBM-1: updated

packages/react/src/components/Tabs/Tabs.tsx
Vignesh-Loganathan-IBM-1: @adamalston  got it . Reverted the changes and updated the PR

heloiselui (approved):
LGTM, thanks for your help!

packages/react/src/components/Tabs/Tabs.tsx
adamalston: Was the correct comment updated in https://github.com/carbon-design-system/carbon/commit/33ef48f2caff09c8c11dd8e993e5dfa23bed81c1? I'm looking at https://github.com/carbon-design-system/carbon/pull/20306/files and it doesn't appear so.

Vignesh-Loganathan-IBM-1: sorry @adamalston . previously updated the other comment . now its updated correctly.

Gururajj77 (approved):

riddhybansal (approved):

PR #20325, opened 2025-08-23
refactor(TextInput): port utils to typescript

adamalston:
No issue.

Ported `TextInput` utils to TypeScript.

### Changelog

**Changed**

- Ported `TextInput` utils to TypeScript.

#### Testing / Reviewing

```sh
yarn test packages/react
```

## PR Checklist


packages/react/src/components/TextInput/util.ts
```diff
@@ -5,33 +5,39 @@
...
+  sharedTextInputProps: Record<string, unknown>;
+  invalid: boolean;
+  invalidId: string;
+  warn?: boolean;
+  warnId?: string;
+  hasHelper?: boolean;
+  helperId?: string;
+}
+
+export const getTextInputProps = ({
```
maradwan26, line 32: Is there a reason for the name change from `textInputProps` to `getTextInputProps`? Tested locally and both work

maradwan26 (commented):
LGTM, just a small comment

packages/react/src/components/TextInput/util.ts
adamalston: The function is a getter. In my opinion, `textInputProps` doesn’t convey the function's purpose, but `getTextInputProps` does. I can change it back if that’s the preferred approach.

maradwan26 (approved):

packages/react/src/components/TextInput/util.ts
tay1orjones: I don't think it really matters here but _technically_ this returns a new reference every time, right? A derived output from inputs, kind of like a factory or assember or something. Whereas getter's usually imply a static and fully stable reference?

Again, I'm not arguing semantics, this is fine imo. Just some random thoughts 🙃 

re preferred approach - maybe it's worth exploring some guidelines in a pr to be documented in the dev docs

tay1orjones: The most uniform example of this naming convention is probably DataTable's public api w/ all the get...Props functions.

tay1orjones (approved):
LGTM!

PR #20378, opened 2025-08-28
fix(menuitem): fix `title` inheritance, move to `li`

tay1orjones:
It was identified in #20154 that if you placed a `title` on `MenuItem`, it would show when hovering the icon. When hovering the text, the label text would show (as intended). This fixes this mismatch so that hovering the MenuItem's icon and text always show the same title, the label.

> [!NOTE]
> `title` is not a prop on MenuItem and is not meant to be configurable, this is why it is ignored. For more context on why, see #20154.

### Changelog

**Changed**

- moved the `title` in MenuItem from `Text` to the parent `li`

#### Testing / Reviewing

- Review the Menu's `For Testing Title Inheritance` story
 - Hover over icon and text, the same title should be visible
 - Note that in the story I've placed a `title` that shouldn't be rendered

## PR Checklist

> [!CAUTION]
> ~Test story needs to be removed before merging~


ariellalgilmore (approved):
LGTM!

maradwan26 (approved):
LGTM, just the test story needs to be removed

PR #20390, opened 2025-08-29
test(toggle): add tests for toggle WC component

### LINKED ISSUE #20108: [toggle]: add unit tests
Add unit test cases for `toggle` component to match React.

maradwan26:
Closes #20108

Added test cases for toggle WC component.

### Changelog

**New**

- Added test cases for `cds-toggle`
- Added test cases for `cds-toggle-skeleton`

**Changed**
- Fixed `_handleChange()` in `cds-toggle` so that it dispatches events correctly
    - removed `indeterminate` from the event
    - changed eventChange to be `typeof CDSToggle`
- Moved `@click` listener from `<label>` to `<button>`
- Fixed `.checked="${checked}"` to `.checked=${checked}` in `render()`

**Changed**
- Removed `_handleKeyDown`, the toggle component should still work as expected by relying on the native button behaviours

#### Testing / Reviewing

- CI should pass
- There should be no regression in `cds-toggle`


heloiselui (commented):
Looks good, just a suggestion:

- Do we also need to create a test for the skeleton like in React?
- And for event tests for `cds-toggle-changed` (to check it emits on click/keyboard and does not emit when `disabled` or `readOnly`)?

maradwan26:
@heloiselui 

I've added new tests, and fixed up some of the older ones 🚀 . I had to fix the handleChange and rendering of the `cds-toggle` component for some tests to pass.

maradwan26:
@heloiselui

I've added new tests, and fixed up some of the older ones 🚀 .

I had to fix the following for cds-toggle for some tests to pass:
- `_handleChange()` now dispatches events correctly, I also removed the unnecessary indeterminate state
- Fixed `checked` attribute in the render
- Moved `@click` listener from the label to the button in the render
- Removed `_handleKeyDown` listener since its now dispatching too many events when toggling with Enter/Space resulting in the toggle going back and forth on a single click. We should be able to rely on the native button behaviour and have the same function.

There shouldn't be any regression for how `cds-toggle` functions from the [parity PR](https://github.com/carbon-design-system/carbon/pull/20264)

heloiselui (approved):
LGTM!

Gururajj77 (approved):

PR #20394, opened 2025-08-30
refactor: port `deprecateValuesWithin` to typescript

adamalston:
Closes N/A

Ported `deprecateValuesWithin` to TypeScript.

### Changelog

**Changed**

- Ported `deprecateValuesWithin` to TypeScript.

#### Testing / Reviewing

Check for type errors.

## PR Checklist

### DIFF
--- a/packages/react/src/prop-types/deprecateValuesWithin.ts
+++ b/packages/react/src/prop-types/deprecateValuesWithin.ts
@@ -1,22 +1,25 @@
 /**
- * Copyright IBM Corp. 2016, 2023
+ * Copyright IBM Corp. 2016, 2025
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
 
+import type { Requireable, Validator } from 'prop-types';
 import { warning } from '../internal/warning';
 
-const didWarnAboutDeprecation = {};
+type DeprecationTracker = Record<string, Record<string, boolean>>;
 
-export default function deprecateValuesWithin(
-  propType,
-  allowedValues,
-  propMappingFunction
-) {
-  return function checker(props, propName, componentName, ...rest) {
-    if (props[propName] === undefined) {
-      return;
+const didWarnAboutDeprecation: DeprecationTracker = {};
+
+export const deprecateValuesWithin = <T, U = unknown>(
+  propType: Validator<T> | Requireable<T>,
+  allowedValues?: readonly unknown[],
+  propMappingFunction?: (deprecatedValue: T) => U
+): Validator<T> | Requireable<T> => {
+  return (props, propName, componentName, ...rest) => {
+    if (typeof props[propName] === 'undefined') {
+      return null;
     }
 
     if (
@@ -45,4 +48,4 @@ export default function deprecateValuesWithin(
     }
     return propType(props, propName, componentName, ...rest);
   };
-}
+};
--- a/packages/react/src/prop-types/deprecateValuesWithin.ts
+++ b/packages/react/src/prop-types/deprecateValuesWithin.ts
@@ -12,10 +12,10 @@ type DeprecationTracker = Record<string, Record<string, boolean>>;
 
 const didWarnAboutDeprecation: DeprecationTracker = {};
 
-export const deprecateValuesWithin = <T, U = unknown>(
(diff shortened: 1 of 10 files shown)


packages/react/src/prop-types/deprecateValuesWithin.ts
```diff
@@ -1,22 +1,25 @@
...
-  allowedValues,
-  propMappingFunction
-) {
-  return function checker(props, propName, componentName, ...rest) {
-    if (props[propName] === undefined) {
-      return;
+const didWarnAboutDeprecation: DeprecationTracker = {};
+
+export const deprecateValuesWithin = <T, U = unknown>(
+  propType: Validator<T> | Requireable<T>,
```
heloiselui: Do we really need the union `Validator<T> | Requireable<T>` here?  
Since `Requireable<T>` already extends `Validator<T>`, using only `Validator<T>` might be enough. The union feels redundant, unless we want to keep it for semantic clarity.

packages/react/src/prop-types/deprecateValuesWithin.ts
```diff
@@ -1,22 +1,25 @@
...
-  propType,
-  allowedValues,
-  propMappingFunction
-) {
-  return function checker(props, propName, componentName, ...rest) {
-    if (props[propName] === undefined) {
-      return;
+const didWarnAboutDeprecation: DeprecationTracker = {};
+
+export const deprecateValuesWithin = <T, U = unknown>(
```
heloiselui: Do we actually need the generic `U` here? It types the `propMappingFunction`, but since the return type doesn’t expose `U`, we could simplify to` <T>`.

packages/react/src/prop-types/deprecateValuesWithin.ts
adamalston: That change results in type errors in all nine files where `deprecateValuesWithin` is used:

```diff
diff --git a/packages/react/src/prop-types/deprecateValuesWithin.ts b/packages/react/src/prop-types/deprecateValuesWithin.ts
index 76bb4ad8e..4fe73e70a 100644
--- a/packages/react/src/prop-types/deprecateValuesWithin.ts
+++ b/packages/react/src/prop-types/deprecateValuesWithin.ts
@@ -13,7 +13,7 @@ type DeprecationTracker = Record<string, Record<string, boolean>>;
 const didWarnAboutDeprecation: DeprecationTracker = {};
 
 export const deprecateValuesWithin = <T, U = unknown>(
-  propType: Validator<T> | Requireable<T>,
+  propType: Validator<T>,
   allowedValues?: readonly unknown[],
   propMappingFunction?: (deprecatedValue: T) => U
 ): Validator<T> | Requireable<T> => {
```

An alternative change might work instead:

```diff
diff --git a/packages/react/src/prop-types/deprecateValuesWithin.ts b/packages/react/src/prop-types/deprecateValuesWithin.ts
index 76bb4ad8e..f84444806 100644
--- a/packages/react/src/prop-types/deprecateValuesWithin.ts
+++ b/packages/react/src/prop-types/deprecateValuesWithin.ts
@@ -13,7 +13,7 @@ type DeprecationTracker = Record<string, Record<string, boolean>>;
 const didWarnAboutDeprecation: DeprecationTracker = {};
 
 export const deprecateValuesWithin = <T, U = unknown>(
-  propType: Validator<T> | Requireable<T>,
+  propType: Requireable<T>,
   allowedValues?: readonly unknown[],
   propMappingFunction?: (deprecatedValue: T) => U
 ): Validator<T> | Requireable<T> => {
```

packages/react/src/prop-types/deprecateValuesWithin.ts
adamalston: It doesn’t appear like we need it. I originally added it because there was an untyped variable in the util, but I don’t see that issue locally anymore. It should be safe to remove the parameter.

packages/react/src/prop-types/deprecateValuesWithin.ts
heloiselui: Yes, I think we can safely go with the second change.

heloiselui (approved):
Looks good, thanks Adam!

kennylam (approved):

PR #19666, opened 2025-06-16
refactor: standardize translation functionality across components

adamalston:
No issue.

Standardized translation functionality across components.

### Changelog

**New**

- Added types for all translation functions.

**Changed**

- Updated `style.md` with TypeScript examples to reflect the standardized pattern.
- Replaced non-standard translation key handling with consistent `translationIds`, `TranslationKey` types, and `defaultTranslateWithId` functions.
- Replaced `translateWithId` props with `TranslateWithId` interface extensions.
- Standardized all JSDoc comments for translation props since they're all referring to the same functionality.

**Removed**

- Removed `translationKeys` PropTypes because they were inconsistently applied across components. Either all components should have them or none should.

#### Testing / Reviewing

Reference commits: https://github.com/carbon-design-system/carbon/compare/6d6150ce59fbbf0bb1e5ccb7751a84e58643ab89...3ab6359a7162986ee0aa64604016aa02c35ad866

The goal of these changes is to introduce consistency into the translation functionality. I began looking into these changes when I noticed many implementations weren't following what was listed in `style.md`.

```sh
yarn test packages/react
```

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
- [x] Updated documentation and storybook examples
- [ ] ~Wrote passing tests that cover this change~
- [x] Addressed any impact on accessibility (a11y)
- [x] Tested for cross-browser consistency
- [x] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #20395, opened 2025-08-30
ci: address circular dependency warnings

adamalston:
Closes N/A

Refactored imports and export to prevent circular dependency warnings in CI.

### Changelog

**Changed**

- Refactored imports and export to prevent circular dependency warnings in CI.
- Updated the `Label` interface in `NumberInput.tsx` so that it no longer matches the component name.
- Moved the `Legend` component to `packages/react/src/components/Text/createTextComponent.tsx` to prevent a circular dependency

**Removed**

- Removed an unused `Label` text component.

#### Testing / Reviewing

https://github.com/carbon-design-system/carbon/pull/19666#issuecomment-3218212128

I’m not sure why the `LabelProps` (previously `Label`) interface is exported, as it isn’t used outside that file.

## PR Checklist

### DIFF
--- a/packages/react/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react/src/components/NumberInput/NumberInput.tsx
@@ -1015,7 +1015,7 @@ NumberInput.propTypes = {
   warnText: PropTypes.node,
 };
 
-export interface LabelProps {
+interface LabelProps {
   disabled?: boolean;
   hideLabel?: boolean;
   id?: string;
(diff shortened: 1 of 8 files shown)


packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -1040,6 +1040,8 @@ const Label: FC<Label> = ({ disabled, id, hideLabel, label }) => {
   return null;
 };
 
+// TODO: `Label` isn't exported. Why does it have `propTypes`? I have the same
+// question for the `HelperText` component too.
```
tay1orjones: Best practice for runtime checks before we had ts? idk honestly, probably to fix an old react or eslint warning or something about missing proptypes. You can remove if you'd like or leave in.

tay1orjones (approved):

kennylam (approved):

PR #20414, opened 2025-09-01
chore: `allowCustomValue` control updated in combobox's typeahead story 

### LINKED ISSUE #20409: [Bug]: In ComboBox Component not clearing the text, if the entries is not present when user click outside
### Package

@carbon/react

### Browser

Chrome

### Package version

1.90

### Description

 In the ComboBox component, when searching for an entry in the dropdown, if the entry is not found and the user clicks outside, the text should be cleared. However, it is currently not clearing. 
I’ve attached a screenshot for reference.

### Screenshots

Image

### Steps to reproduce

The issue is also reproducible at the docs.
Try to type any random string in the search bar, then click outside 
It should clear the text area, but it is not clearing

### Code of Conduct

Gururajj77:
Closes #20409 

the typeahead story now doesn't have `allowCustomValue` prop applied by default.
the control is updated to work properly with default value as false

#### Testing / Reviewing

- go to the typeahead story
- type anything that doesn't exist in the list
- click outside and see that it get's cleared.
- do the same after enabling the `allowCustomValue` control
- see that your typed characters are not cleared

## PR Checklist


riddhybansal (approved):

Kritvi-bhatia17 (changes requested):
Hi @Gururajj77, It seems to be working correctly when the Autocomplete prop is turned off. However, when it's enabled, the value gets cleared upon clicking outside the combobox — ideally, that shouldn't happen. The typed value should persist even clicking outside it.

https://github.com/user-attachments/assets/c0aae57d-654a-4323-891e-b5e4cdda2848

Kritvi-bhatia17 (approved):
LGTM 🚀

2nikhiltom (approved):
LGTM

PR #20415, opened 2025-09-01
fix(Search): use native input type="search"

### LINKED ISSUE #18031: [a11y]: TableToolbarSearch: use <input type=search>
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.70

### React version

18

none

### Assistive technology

_No response_

### Description

The [first rule of ARIA use](https://www.w3.org/TR/aria-in-html/#rule1) is:

> If you _can_ use a native HTML element [[HTML51](https://www.w3.org/TR/aria-in-html/#bib-HTML51)] or attribute with the semantics and behaviour you require **already built in**, instead of re-purposing an element and adding an ARIA role, state or property to make it accessible, **then do so**.

TableToolbarSearch currently renders:

```html
<input ... class="cds--search-input" role="searchbox" ... type="text" tabindex="0" value="">
```

As written in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/searchbox_role, it should be:

```html
<input type="search" />
```

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datatable-toolbar--default

### Steps to reproduce

N/A

### Suggested Severity

None

### Code of Conduct

heloiselui:
Closes #18031

Update `Search` and `TableToolbarSearch` to use `<input type="search">` instead of `<input type="text" role="searchbox">` to follow accessibility best practices.

This pattern may have been implemented originally to avoid inconsistent browser styles, ensure reliable support in older screen readers, and give Carbon full control over the input’s visual behavior. Since modern browsers and screen readers now provide consistent support for `<input type="search">`, keeping both is no longer necessary, and simplifying to `<input type="search">` provides a cleaner and more accessible solution.

### Changelog

**New**

- Added CSS resets for native `<input type="search">` browser decorations.

**Changed**

- Updated `Search` component: default `type` is now `"search"`.
- Deprecated the `role` prop. It will be removed in the next major release.
- Updated Storybook stories to use `type="search"`.
- Updated snapshots for `DataTable` and `TableToolbarSearch` to reflect `type="search"`.

**Removed**

- Default `role="searchbox"` from `Search` since `<input type="search">` already provides correct semantics.

#### Testing / Reviewing

- Go to `React Deploy Preview` > `Search` > `Default`.
  - Inspect the input in DevTools > it should render as `<input type="search">` without `role="searchbox"`.

image

- Verify `TableToolbarSearch` also renders `<input type="search">`:
  - Go to `React Deploy Preview` > `DataTable` > `Default`.
  - Inspect the `toolbar search` input in DevTools.
  - Confirm the input renders with `type="search"` and no `role="searchbox"`.

image

- Check in Chrome, Safari, and Firefox to confirm there are no visual regressions. (I've already reviewed it, but feel free to double-check.)

## PR Checklist


packages/react/src/components/Search/Search.tsx
```diff
@@ -112,7 +114,7 @@ export interface SearchProps extends InputPropsBase {
   size?: 'sm' | 'md' | 'lg';
 
   /**
-   * Optional prop to specify the type of the `<input>`
+   * Specify the type of the `<input>`, defaults to "search"
```
adamalston: Is it redundant to mention a prop's default value in its description when the Component API table in the Storybook already includes a dedicated column for default values?

packages/react/src/components/Search/Search.tsx
```diff
@@ -383,7 +385,9 @@ Search.propTypes = {
   renderIcon: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
 
   /**
-   * Specify the role for the underlying `<input>`, defaults to `searchbox`
+   * @deprecated Specify the role for the underlying `<input>`.
+   * No longer needed since `<input type="search">` already provides the correct semantics.
+   * This prop will be removed in the next major release of Carbon.
    */
   role: PropTypes.string,
```
adamalston: Should the `deprecate` util be used here?

packages/react/src/components/Search/Search.tsx
heloiselui: I totally get your point. I only included the default value in the description to reinforce it, but we can definitely remove it if it feels redundant.

packages/react/src/components/Search/Search.tsx
heloiselui: Thanks for catching this, I'll fix it.

emyarod (approved):
looks good to me!

riddhybansal (approved):
LGTM 💯

PR #20418, opened 2025-09-02
refactor: radio button parity

### LINKED ISSUE #19862: React/WC Parity: Radio-Buttons
### Acceptance Criteria

- [ ] See parent issue #17992 for baseline criteria
- [ ] Implement `defaultSelected`, `valueSelected`, and `required` attributes and add their respective test cases

> [!IMPORTANT]  

Gururajj77:
Closes #19862

Brings WC `<cds-radio-button>` and `<cds-radio-button-group>` functionality closer to their React counterparts, aligning event payloads, required/defaultChecked handling and readOnly behavior.

### Changelog

**New**

- Added `required` property on `<cds-radio-button>` mapped to the native `<input>`
- Added `defaultChecked` property on `<cds-radio-button>`, applied on first render if `checked` is not set
- Added `required` property on `<cds-radio-button-group>`, propagating it to child `<cds-radio-button>` elements

**Changed**

- Updated `cds-radio-button-changed` event detail to include `{ value, name, event, checked }` for parity with React `onChange`
- Updated `cds-radio-button-group-changed` event detail to include `{ value, name, event }` for parity with React `onChange`
- Enforced `readOnly` on `<cds-radio-button-group>` so that value does not update and change events are not fired when set
- Added `data-invalid` and `aria-invalid` attributes on `<cds-radio-button-group>` `<fieldset>` when `invalid` is true

**Removed**

- None

#### Testing / Reviewing

play around the default story and try to compare the components to make sure no missing features are present.

## PR Checklist


tay1orjones (approved):
Nice 👍

heloiselui (commented):
It looks good overall, just a few things I noticed:

- I think we need add the `AI label` to the table content in WC.

- In WC > Radio Button > Default, the second radio button label isn’t selected by default as it is in React.

- I think we should open a bug to fix the `disabled` behavior in both WC and React. When selected, the radio button label text can still be highlighted.

- `hideLabel` in React doesn’t behave the same way as in WC, I think this might be a bug.

- In WC, if I set both invalid: true and readOnly: true, the component still appears red. Same happens with disabled — it stays red, which is different from React.

packages/web-components/src/components/radio-button/radio-button.ts
```diff
@@ -397,6 +397,12 @@ class CDSRadioButton extends HostListenerMixin(FocusMixin(LitElement)) {
     hasAILabel
       ? this.setAttribute('ai-label', '')
       : this.removeAttribute('ai-label');
+
+    if (changedProperties.has('invalid') || changedProperties.has('readOnly')) {
+      if (this.readOnly) {
+        this.invalid = false;
+      }
+    }
```
maradwan26: I think instead of hard-setting the invalid attribute to false when readOnly is true we can just remove the red outline by adding a conditional `:not(:read-only))` to the [style](https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/radio-button/radio-button.scss#L71)

maradwan26 (commented):
LGTM, just a few comments

The label-text for the radio buttons with an AI-Label is not showing
Screenshot 2025-09-08 at 12 24 59 PM

This is probably because its picking up on spaces in the story's `render` for some reason, so removing the spaces could fix this

```    
<cds-radio-button label-text="Radio button label" value="radio-4">
    <cds-ai-label alignment="bottom-left">
        ${content}${actions}
    </cds-ai-label>
</cds-radio-button>
```

Gururajj77:
> LGTM, just a few comments
> 
> The label-text for the radio buttons with an AI-Label is not showing Screenshot 2025-09-08 at 12 24 59 PM
> 
> This is probably because its picking up on spaces in the story's `render` for some reason, so removing the spaces could fix this
> 
> ```
> <cds-radio-button label-text="Radio button label" value="radio-4">
>     <cds-ai-label alignment="bottom-left">
>         ${content}${actions}
>     </cds-ai-label>
> </cds-radio-button>
> ```

This is fixed in the latest commit

maradwan26 (approved):
LGTM

Gururajj77:
> It looks good overall, just a few things I noticed:
> 
> * I think we need add the `AI label` to the table content in WC.
- done, added the ai label hyperlink
> * In WC > Radio Button > Default, the second radio button label isn’t selected by default as it is in React.
- the 2nd radio button is default now
> * I think we should open a bug to fix the `disabled` behavior in both WC and React. When selected, the radio button label text can still be highlighted.
-  will raise a issue for this
> * `hideLabel` in React doesn’t behave the same way as in WC, I think this might be a bug.
- the `hideLabel` prop wasn't properly propagated in the react story, have modified to make it work like WC
> * In WC, if I set both invalid: true and readOnly: true, the component still appears red. Same happens with disabled — it stays red, which is different from React.
- this is fixed now.

riddhybansal (approved):

heloiselui (approved):
Thanks @Gururajj77 🚀

thyhmdo (approved):
Looks good

PR #20419, opened 2025-09-02
feat(datatable): select all button in wc batch actions

### LINKED ISSUE #19543: React|WC Parity: Data-table, select all
The web component batch action toolbar should support a way to render a "Select all" action, similar to what is supported in `@carbon/react`.

See [previous issue](https://github.com/carbon-design-system/carbon/issues/12595) for initial contribution to `TableBatchActions`.

Image

szinta:
Closes #19543 

Adds Select all button on Datatable batch actions toolbar, in web components

#### Testing / Reviewing

1. Open WC Storybook 
2. Navigate to  components ->Data Table -> Batch Actions -> Default story
3. Select any one row to see the Select all button in Batch action toolbar

## PR Checklist


packages/web-components/src/components/data-table/table-batch-actions.ts
```diff
@@ -130,6 +153,18 @@ class CDSTableBatchActions extends LitElement {
...
           ${formatSelectedItemsCount({ count: selectedRowsCount })}
         </p>
+        ${totalRowsCount > 0
+          ? html`
+              <span class="${prefix}--batch-summary__divider">|</span>
+              <button
+                class="${prefix}--btn ${prefix}--btn--primary ${prefix}--batch-summary__select-all"
+                @click=${handleSelectAll}>
+                <slot name="select-all-button-content"
+                  >Select all (${totalRowsCount})</slot
```
ariellalgilmore, line 163: can we make the "select all" text a prop or something? in react we have a tool to translate, but we don't have this in web components. but we need to make sure non english users won't have issues

ariellalgilmore (commented):
looks good just left one small comment. Also, is this button supposed to do something? I tried running it in react too, but it looks the same. Looking back at the issues it seems like the functionality will be handled separately just wanted to confirm.

packages/web-components/src/components/data-table/table-batch-actions.ts
szinta: Thanks for the suggestion! The "Select all" text is currently wrapped in a slot, which allows the end users to override the default "Select all" text with their own localized version. The same approach is applied for the "Cancel" button within the component. Though we don't have built-in translation tooling for web components like we do in React, I believe, this slot-based structure ensures flexibility for non English users.

szinta:
> looks good just left one small comment. Also, is this button supposed to do something? I tried running it in react too, but it looks the same. Looking back at the issues it seems like the functionality will be handled separately just wanted to confirm.

That's right. `Select All` button does not include any built in functionality

ariellalgilmore (approved):

riddhybansal (approved):
This LGTM 💯

alina-jacob (approved):
Looks good to me! 🚀 

I don't see the same feature in React though, am i missing something. I know @mattgalo contributed it [here](https://github.com/carbon-design-system/carbon/pull/14476), but i don't see it in our current library!

PR #20429, opened 2025-09-02
fix(textarea): fix helper-text "max-width: 0px" on hidden render

### LINKED ISSUE #20417: [Bug]: Helper text in TextArea has 0px width if not visible during first render
### Package

@carbon/react

### Browser

Safari

### Package version

1.90.0

### React version

19

### Description

When a TextArea is present but not visible during its first render (for example in a TabPanel or a CreateTearsheetStep form C4IBMP), the container width for its helper text is determined as `0px` making it break after each letter.

Image

This might be related to #19971

### Reproduction/example

https://stackblitz.com/edit/github-jfti5w9t?file=src%2FApp.jsx

### Steps to reproduce

- Implement a TextArea in a way that it is present in the DOM but not visible during it's first render
- Pass a helper text
- Make it visible

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

maradwan26:
Closes #20417

This PR fixes a bug where TextArea would have a 0px width if not visible during first render using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) for both React and Web Components.

### Changelog

**New**

- Added test stories `TextArea/Test With Tab` and `TextArea/Text Area Visibility Test` to both React and WC
    - _**should be removed before merging**_

**Changed**

- _React:_ Refactored `useIsomorphicEffect()` in `TextArea.tsx` to use a `ResizeObserver` that observes the `wrapperRef` and applies the calculated width when needed
- _Web Components:_ Refactored `updated()` in `textarea.ts` to use a `ResizeObserver` that observes `text-area__wrapper` and applies the calculated width when needed
    - This replaces the previous usage of the private `_prevCols` variable that was used to determine when the width should be measured
    - A `disconnectedCallback()` function was also added to cleanup and disconnect the observer

**Removed**

- Removed private `prevCols` from `cds-textarea`

#### Testing / Reviewing

- Go to `TextArea/Test With Tab` story in both React and WC
- Helper-text should not have a max width of 0 and should render as expected
- Go to `TextArea/Text Area Visibility Test` story in both React and WC
- Toggling visibility and/or toggling cols should render the helper-text as expected

## PR Checklist

> [!CAUTION]
> Test stories need to be removed before merging


packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -226,19 +226,29 @@ const TextArea = frFn((props, forwardRef) => {
...
+        }
+      });
+    };
+
+    const resizeObserver = new ResizeObserver(([entry]) => {
+      applyWidth(entry.contentRect.width);
     });
+    resizeObserver.observe(wrapperRef.current);
+
+    return () => resizeObserver.disconnect();
```
tay1orjones: Does this need to be guarded in case it's not defined from the early return up on line 236?


```suggestion
    return () => resizeObserver && resizeObserver.disconnect();
```

maradwan26: I don't think so because the observer is defined after the early return on line 236

tay1orjones: The cleanup function will always run on unmount

maradwan26: Oh I see ok thanks, good catch!

tay1orjones (approved):

heloiselui (approved):
LGTM, just need to remove the test story!

PR #20467, opened 2025-09-08
feat(DataTable-Dynamic): Add controls to Dynamic story

maradwan26:
No issue

There have been a couple times were users have mentioned the `DataTable/Dynamic` stories should also show how it would look using single-selection instead of multi-select. There are no controls in the `DataTable/Dynamic/Default` story in the React storybook, unlike the WC storybook, so there is currently no way for users to check this out for themselves in React.

This PR adds controls to the `DataTable/Dynamic/Default` story for React to allow users to customize the Dynamic DataTable like how they could in Web Components.

### Changelog

**New**

- Added controls to `DataTable/Dynamic/Default` story in React storybook

**Removed**

- Removed `table: { disable: true }` from `Default.argTypes` since its unnecessary

#### Testing / Reviewing

- Go to `DataTable/Dynamic/Default` in React storybook
- Controls should work

## PR Checklist


Gururajj77 (approved):

heloiselui (approved):
LGTM!

PR #20477, opened 2025-09-09
New Fluid text input web component

### LINKED ISSUE #20339: React|WC Parity: Create FluidTextInput
See parent issue #17992 for acceptance criteria.

> [!IMPORTANT]  

### LINKED ISSUE #20422: [fluid-text-input]: add unit test cases
Add unit test cases for fluid text inputs.

riddhybansal:
Closes #20422
Closes #20339  

New fluid textinput web component
Added test cases for the same.

### Changelog

**New**

- New fluid textinput web component
- Added test. cases for the same

#### Testing / Reviewing

Fluid text input should work as intended(React implementation)
Test should pass.

## PR Checklist


Gururajj77 (approved):
Looks good 🚀

riddhybansal:
There is a known [issue](https://github.com/carbon-design-system/carbon/pull/20477) with the toggletip

heloiselui (approved):
Looks good! 🚀

laurenmrice (changes requested):
@riddhybansal 

**Default story**

- This is happening in React as well, but when the component is disabled, you should not be able to interact with the Label, it should still have the disabled cursor. I can make a separate issue for this if needed, just let me know. 
- The `readOnly` prop has a capital O in React and its lowercase o in WC. I feel like this should be aligned with React unless this is how WC is doing it now across the board.

----

 **Default With Toggletip story**

- React needs to change its “Tooltip” terminology to “Toggletip”. WC is correct, it should be saying Toggletip. I can make a separate issue to update that instance for all fluid inputs in React.
- When trying to focus on the toggletip icon, the focus is obscured and cut off. I also can not close the toggletip by clicking anywhere outside of it but I should be able to, this seems like a problem stemming from the actual toggletip component in WC. Looks like you have an issue for that [here](https://github.com/carbon-design-system/carbon/issues/20490).
* The visual style of the open toggletip should be as follows in the spec and should match React:
Frame 1

----

**Skeleton story**

- The container is 65px, it should be 64px.

riddhybansal:
> * here

@laurenmrice 

- For the readonly state all the props comes from the non fluid component here its TextInput, but what I can see in most components in web components its `readonly` but some components like Select Multiselect its `readOnly`. Definitely an in consistent behaviour but changing anyone of them would be a breaking change. So ofcourse we need to create a seperate issue for that , and need to figure out how it needs to be done and need to fix multiple components. So we might need a generic issue for that as I can see this in NumberInput , TextArea, TextInput etc... Needs to figure out.

- For the disabled state the issue occurs accross all components , labels does not have a cursor so totally a style enhancement we need to do for all components, then it will reflect both in react and webcomponents. So we need to create an issue for that as well.

- For the renaming of tooltip to toggletip , its done for this component for both react and webcomponents. And I can see , you have already created issue for the same.

- All the toggletip visual issues are the issue related to toggletip. So I have updated the issue with all the information you provided [here](https://github.com/carbon-design-system/carbon/issues/20490).

- Skeleton state is fixed 💯

laurenmrice (approved):

PR #20501, opened 2025-09-10
fix(telemetry): patch package to v1.10.2

IgnacioBecerra:
Closes N/A

For a bit of context, I'll give a brief rundown of how `@ibm/telemetry-js` works. 

When any repo using carbon packages is doing its `npm install`, all carbon packages are being installed at once. To avoid them running our `postinstall` collection script at the same time and affecting CI performance, we make it so the first package that runs the script will become an IPC server. All others will become IPC clients, and send over their "work to do" over to the server. The server then processes all the package's sequentially, in order to avoid using up too many resources. 

However, if the package that becomes a server is using any version of telemetry that is not `v1.10.0` or higher, then it won't have the code required to run the Web Components scope for data collection. In other words, unless our Carbon Web Components package becomes the server, the chances of collecting WC data would be very rare, especially because we've found that CWC is one of the last packages to install, probably due to being one of the last carbon packages in a `package.json` if ordered alphabetically.

To circumvent this, this new `1.10.2` version will always have Web Components data collection work to be sent to its own WC IPC server, instead of the default server. This ensures that, if any repo is using any WC package, we will always be collecting WC data no matter what the default IPC server's version of telemetry is.

TL;DR: we are spinning up a WC-only IPC server to ensure WC data collection

### Changelog

**Changed**

- update `@ibm/telemetry-js` package to `v1.10.2`
- web components data will be collected under a different IPC server

#### Testing / Reviewing
N/A

## PR Checklist


emyarod (approved):
looks good to me!

heloiselui (approved):
LGTM!

PR #20505, opened 2025-09-11
fix(toggle): align readonly small toggle to design spec

### LINKED ISSUE #20367: [Bug] toggle read-only prop
#### Small size toggle
- Read-only prop set to true: The checkmark in the thumb/handle should not be visible while toggled on, the circle should just remain the `$icon-primary color.
`
---

#### Design spec

Image

maradwan26:
Closes #20367

This PR aligns the readonly small toggle to the design spec by removing the checkmark when it is in a `toggled` state.

### Changelog

**New**

- Added conditions to conditionally render the checkmark only if the toggle is not small and readonly

**Changed**

- Changed the toggle switch colour token from `$text-primary` to `$icon-primary`
    - This should not result in any visual changes since both tokens carry the same colour value. This is purely just for nomenclature consistency with the design spec

#### Testing / Reviewing

- Go to `Toggle/Default` in deploy preview for both React and Web Components
- In the controls set `size = small` and `readOnly=true`
    - [React link](https://deploy-preview-20505--v11-carbon-react.netlify.app/?path=/story/components-toggle--default&args=size:sm;readOnly:!true)
    - [WC link](https://deploy-preview-20505--v11-carbon-web-components.netlify.app/?path=/story/components-toggle--default&args=readOnly:!true;size:sm)
- There should be no checkmark visible inside of the toggle switch
- There should be no regression on the colour of the toggle switch, the token change should result in no visual changes

## PR Checklist

### DIFF
--- a/packages/react/src/components/Toggle/Toggle.tsx
+++ b/packages/react/src/components/Toggle/Toggle.tsx
@@ -203,7 +203,7 @@ export function Toggle({
         {labelText && <Text className={labelTextClasses}>{labelText}</Text>}
         <div className={appearanceClasses}>
           <div className={switchClasses}>
-            {isSm && (
+            {isSm && !readOnly && (
               <svg
                 aria-hidden="true"
                 focusable="false"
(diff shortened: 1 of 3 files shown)


emyarod (approved):
looks good to me!

sunny-babbar (approved):

PR #20516, opened 2025-09-12
fix(cds-table): resolve cds-table with mixed expand/non-expand rows

### LINKED ISSUE #17612: [Bug]: data table `expandable` prop does not work to dynamically make rows expandable
### Package

@carbon/web-components

### Browser

Chrome

### Package version

2.11.1

### React version

18.2.0

### Description

In our use case we would like only some rows to be expandable. The [storybook docs](https://web-components.carbondesignsystem.com/?path=/docs/components-datatable-basic--playground#expansion) mention
> In case you only want to set specific rows to be expandable, be sure to set the expandable prop only on that specific <cds-table-row> component, and have its adjacent sibling <cds-table-expanded-row> directly following it.

I've tried this a few different times with and without sorting / other props we've added and I can't get it to work. Setting expandable for the entire data-table works, but then there are rows that won't have content within the expandable section which isn't what we want.

It was also pointed out by @ariellalgilmore
> This looks to be a bug unfortunately.  I would also expect the expandable buttons shouldn't be visible if there are no  <cds-table-expanded-row>  even with the expandable prop

### Reproduction/example

https://stackblitz.com/edit/github-kbd9xw-xb63ek?file=index.html

### Steps to reproduce

1) Open the link
2) View the code in index.html that is authored using the format storybook docs suggest (setting expandable on a specific row with an adjacent sibling expanded row following it)
3) View the output that does not have one expandable row and one "normal" row.

### Suggested Severity

### Application/PAL

watsonx Assistant

### Code of Conduct

jeffchew:
Closes #17612

Fixes UI issues with expansion controls in tables with mixed expandable/non-expandable rows:
   - Only sets rows as expandable if they have expanded content
   - Maintains consistent table structure with proper styling for empty expansion cells

### Changelog

**Changed**

- Updated `_renderExpandButton` method to only render expand button if they actually have expanded content.
- Update to expandable row detection logic
- Added styles for empty expansion cells to fix highlighting and bottom border issues.

#### Testing / Reviewing

Stories for data table expansion have been modified to include a mix of expandable and non-expandable rows.

## PR Checklist


maradwan26 (approved):
LGTM !

tay1orjones (approved):

PR #20520, opened 2025-09-15
feat(NumberInput): add support locale format validation

abpaul1993:
Closes #

When entering meter readings with a wrong decimal separator there is no message indicating that the decimal separator is wrong and the meter reading is saved by ignoring the decimal separator.

RESULT:
A wrong meter reading is saved, e.g., 19 instead of 1.9.
EXPECTED RESULT:
The user is notified that the value entered is not correct.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

## PR Checklist


JordanWSmith15:
LGTM. Requirement was to show inline error rather than auto-correct formatting. So:
1,00.50 = Failure
1,0.50 = Failure
1.0,5 = Failure
1,000.50 = PASS

tay1orjones (commented):
Nice! It would be awesome if this could be a configurable option that is extensible to allow other consumer use cases. I think one way to do this could be:

* Add a new `validate` function prop that is called with a `value` and `locale`.
  * If it `return false` the input fails validation, `return true` passes validation, `undefined` would also pass validation for the case if they don't pass a function or it doesn't actually return anything.
* Pass `validate` as a new param to `getInputValidity()`
* In `getInputValidity`, call `validate` first, before the existing validation checks. This gives consumers the utmost control.
* Replace existing `isValidNumberInput` calls to call `getInputValidity` instead
* Use `normalizedProps.invalid` instead of `validFormat`
* Remove `setValidFormat` state and calls - these are right before/after setNumberValue, which is going to trigger a re-render anyway where then `getInputValidity` will be called again before `normalize`

The final piece would be to export the validation function. Maybe rename it, too - `validateNumberLocaleSeparators`, or `validateNumberSeparators`?
```diff
- const isValidNumberInput =
+ export const validateNumberSeparators = 
```

Usage in your app would look like
```js
import { NumberInput, validateNumberSeparators } from `@carbon/react`;

// ...

<NumberInput
  validate={validateNumberSeparators}
  // ...
/>
```

The benefits of this would be:
* No change in logic for existing users
* Opt-in is really simple
* Future validation functions we want to standardize can live and be maintained in `@carbon/react`
* `validate` accommodates any other use cases for additional custom validation

```js
<NumberInput
  validate={(value, locale) => {
    return localeSeparatorValidation(value, locale) && yourCustomValidation(value)
  }}
  ...
/>
```

Curious what you think about this, let me know!

tay1orjones:
@laurenmrice if the input is cleared on blur should it retain the invalid style/message or no?

https://github.com/user-attachments/assets/d32a402b-f19a-46dd-9684-90c10911f471

JordanWSmith15:
> @laurenmrice if the input is cleared on blur should it retain the invalid style/message or no?
> 2025-09-17.at.13.49.38-Components.NumberInput.-.With.Type.Of.Text.Storybook-Google.Chrome.mp4

agree this is probably not what we want and not what I see in production

laurenmrice:
@tay1orjones No, I don't think it should retain the invalid error message if the input is cleared on blur.

JordanWSmith15:
@tay1orjones  @abpaul1993  If you have "4,5" in the input, and something is implicitly coded by the application developer to clear the value, there is no longer an error, so we should not show the message anymore. 
However, I don't think we should be clearing things onBlur by default. We should never remove data onBlur from an input, whether it's valid or not. It's to the user to repair the error.

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -1073,9 +1169,26 @@ const HelperText = ({ disabled, description, id }: HelperTextProps) => {
...
+  invalid,
+  value,
+  max,
+  min,
+  validate,
+  locale,
+}) {
+  if (typeof validate === 'function') {
+    const result = validate(value, locale);
+    if (result !== undefined) {
```
sls-ca: I would say that if we are invalid we can return immediately.  But if we are "valid" here, we could still become invalid later.  ie, min/max, etc.  I see the validator does specifiy that there are 3 returns true/false/undefined but the number validation functions return 'true' for things like 'empty number', so what if things are required?

To keep things simple, I would focus on a false return as being invalid and return immediately.  Any `true` should just allow for further validation.

Unless the intention really is that the validation function should supercede any validation such as min/max, require, etc.  And if that is the case then we should update the locale parsing validation functions to return undefined where they currently return true.

sls-ca (changes requested):

JordanWSmith15:
LG2M

tay1orjones (approved):
LGTM 👍 Thank you!

sls-ca (approved):

sls-ca (approved):

ariellalgilmore (approved):
LGTM!

PR #20525, opened 2025-09-15
fix: let decorator accept any component instead of ailabel

### LINKED ISSUE #20494: [Bug]: decorator prop currently only accepts and renders AILabel component
### Package

@carbon/react

### Browser

_No response_

### Package version

1.90.0

### React version

_No response_

### Description

The `decorator` prop in most of the components is not rendering `custom` React elements, such as `Toggletip`
this behavior is a regression caused by a recent change that explicitly checks for and only renders the AILabel component in this [PR](https://github.com/carbon-design-system/carbon/pull/18971/files)

### Reproduction/example

https://stackblitz.com/edit/qaw8crdy?file=src%2FApp.jsx

### Steps to reproduce

pass a custom element to `decorator` prop and confirm it is not rendered 

```
<RadioButtonGroup
      legendText="Radio Button group"
      name="radio-button-default-group"
      decorator={
        <Toggletip autoAlign>
          <ToggletipButton label={'test'}>
            <span>i</span>
          </ToggletipButton>
          <ToggletipContent>{'test'}</ToggletipContent>
        </Toggletip>
      }
    >
```
workaround [example](https://stackblitz.com/edit/qaw8crdy-xwyjru8s?file=src%2FApp.jsx,package.json)  

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #20494

decorator prop accepts any component now, instead of only `AILabel`

> [!NOTE]  
> remove the test story before merging
### Changelog

**Changed**

- modified the normaliser to accept the passed component as it is, if it's not an `AILabel`
```
const normalizedDecorator = candidateIsAILabel
    ? cloneElement(candidate, { size: 'mini', kind: 'default' })
    : candidate;
```

#### Testing / Reviewing

- the test story in the RadioButton component, shows how the `ToggleTip` shows up after this change.

## PR Checklist


packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -611,14 +613,18 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
     // AILabel always size `mini`
     const candidate = slug ?? decorator;
     const candidateIsAILabel = isComponentElement(candidate, AILabel);
-    const normalizedDecorator = candidateIsAILabel
+    const normalizedDecorator: ReactNode = candidateIsAILabel
```
adamalston: This type is not required elsewhere. Why is it needed here?

```suggestion
    const normalizedDecorator = candidateIsAILabel
```

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -611,14 +613,18 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
...
     let isRevertActive: AILabelProps['revertActive'];
-    if (normalizedDecorator?.type === AILabel) {
-      isRevertActive = normalizedDecorator.props.revertActive;
+    if (
+      isValidElement(normalizedDecorator) &&
+      normalizedDecorator.type === AILabel
+    ) {
+      const asAILabel = normalizedDecorator as ReactElement<AILabelProps>;
+      isRevertActive = asAILabel.props.revertActive;
     }
```
adamalston: All of these types can be inferred by using `isComponentElement` or another type-safe implementation.

```tsx
const isRevertActive = isComponentElement(normalizedDecorator, AILabel)
  ? normalizedDecorator.props.revertActive
  : undefined;
```

However, since we already determined earlier in the file whether `candidate` is an `AILabel`, why do we need to perform this check again?

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -10,6 +10,8 @@ import cx from 'classnames';
 import PropTypes from 'prop-types';
 import React, {
   cloneElement,
+  isValidElement,
+  ReactElement,
```
adamalston:
```suggestion
  type ReactElement,
```

packages/react/src/components/NumberInput/NumberInput.tsx
Gururajj77: have made the changes as needed

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -611,15 +611,14 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
...
       ? cloneElement(candidate, { size: 'mini' })
-      : null;
+      : candidate;
 
     // Need to update the internal value when the revert button is clicked
-    let isRevertActive: AILabelProps['revertActive'];
-    if (normalizedDecorator?.type === AILabel) {
-      isRevertActive = normalizedDecorator.props.revertActive;
-    }
+    const isRevertActive = isComponentElement(normalizedDecorator, AILabel)
```
adamalston, line 711: https://github.com/carbon-design-system/carbon/pull/20525#discussion_r2352576057:

> [S]ince we already determined earlier in the file whether `candidate` is an `AILabel`, why do we need to perform this check again?

packages/react/src/components/NumberInput/NumberInput.tsx
adamalston: I don't see an answer to the question. Is the type required, or can it be removed?

packages/react/src/components/NumberInput/NumberInput.tsx
Gururajj77: This can be removed.

maradwan26 (approved):
LGTM

tay1orjones (approved):

maradwan26:
@Gururajj77 
Test story just needs to be removed to merge

packages/react/src/components/NumberInput/NumberInput.tsx
adamalston: I don't see a removal. Was these supposed to be a change here?

PR #20544, opened 2025-09-17
feat(storybook): globally filter controls and remove usages of table: { disable: true }

### LINKED ISSUE #20354: [Enhancement]: Use control filtering in Storybook to filter out controls we want to ignore
There are many instances within the React storybook where we are manually hiding controls by making a new `argTypes` for a story, and then going through each and every control and hiding it from the table. This can be tedious and makes the code-base overly verbose, especially when a component has many controls ([Example](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tag/Tag.stories.js#L124-L186)). This has been a reason to why many stories have controls such as 'children' that don't serve much purpose.

Storybook has a control filtering feature, where controls can simply be excluded from the table if they aren't wanted. See docs: https://storybook.js.org/docs/essentials/controls#filtering-controls

This has been used [in this commit](https://github.com/carbon-design-system/carbon/pull/20264/commits/fc3e656273180d50281e58f42375844162cb72c7) to hide every single control from the Skeleton story with the exception of `className` by simply passing in `(args)` to the story and using `include`. 
```
export const Skeleton = (args) => {
  return (
    <div>
      <ToggleSkeleton {...args} />
    </div>
  );
};

Skeleton.parameters = {
  controls: { include: ['className'] },
};
```

### Task:

The Storybook control filtering feature should be used _globally_ to hide un-wanted controls from _all stories_.

maradwan26:
Closes #20354

This enhancement globally filters controls in the React storybook and removes previous usages of `table: { disable: true },` in favour of [Storybook's include/exclude control filtering](https://storybook.js.org/docs/essentials/controls#filtering-controls), while also maintaining all argTypes for documentation purposes in the Overview pages.

The following controls are globally filtered:
- className
- children
- as
- decorator
- slug
- ref
- ariaLabel
    - This is because ariaLabel has been deprecated in favour of aria-label
- onChange, onClick, on....

### Changelog

**New**

- Added an [argTypesEnhancers](https://www.npmjs.com/package/@storybook/store/v/7.0.0-alpha.19#argtypes-enhancement) to `preview.js` that takes in an array of strings/regexes "`GLOBAL_EXCLUDED_CONTROLS`" of controls we want to globally filter out by applying `table: { disable: true },` to them

    - This method was used [due to a limitation with Storybook's parameters](https://storybook.js.org/docs/writing-stories/parameters#rules-of-parameter-inheritance). Globally filtering controls via `parameters = { controls: { exclude: .... },` would block the ability to filter controls specific to a single component's stories, the global filter would get overwritten. This then leads to us having to either reapply the global filter on top of a specific filter for each story, or use `table: { disable: true },`, both of which are methods that weaken this enhancement and its purpose. `argTypesEnhancers` allows for developers to use story specific control filtering with no issue.

    - The filter also ensures that controls/args aren't filtered out of the `Overview` pages, as even though they may not be used in the stories, they are useful as documentation for users.

**Changed**

- Changed usages of `table: { disable: true },` to `parameters = { controls: { exclude/include: .... },` throughout the codebase 

**Removed**

- Removed usages of `table: { disable: true },` to disable controls throughout codebase
- Removed usages of `table: { disable: true },` that hid argType from the `Overview` page throughout the codebase
    - There were _few_ exceptions to this. Like args that cannot be used with the current component (e.g. FluidTextArea and helperText, hideLabel, and light) either due to the args being inherited from a parent component, or not being supported (light).

#### Testing / Reviewing

- Stories should have the globally filtered controls not shown in their controls table
- ArgTypes should still be visible in Overview Page
- There shouldn't be unnecessary usages of `table: { disable: true },` in the storybook files in this branch
- Mention if there should be more/less controls added to the list of globally filtered controls
- CI should pass

## PR Checklist


tay1orjones (approved):
This cleans things up so much 🙌 LGTM

packages/react/.storybook/preview.js
```diff
@@ -368,4 +368,57 @@ const preview = {
...
+  'ariaLabel',
+  /^(?:on[A-Z]\w*)$/,
+];
+
+/**
+ * ArgTypes enhancer that enforces a global control filter across all stories.
+ * Args matching elements in the GLOBAL_EXCLUDED_CONTROLS array are not shown in the
+ * `controls` table
+ */
+export const argTypesEnhancers = [
```
tay1orjones, line 388: Worth noting this appears to be considered an experimental API https://github.com/storybookjs/storybook/issues/30749

maradwan26: Yeah, should I just add a comment, then?

Gururajj77 (approved):

PR #20094, opened 2025-08-05
fix(19982): fix toggletip-clickhandler

2nikhiltom:
Closes https://github.com/carbon-design-system/carbon/issues/19982

This PR fix ai-label tooltip closes unexpectedly when clicking slotted interactive content due to this.contains() not properly detecting slotted elements across shadow DOM boundaries

### Changelog

**New**

Added Test story to demo and test slotted content interactions ~(need to remove before merging - Done ✅)~
Added a new test case for this fix

**Changed**

Updated toggletip's  `_handleFocusOut` method to use` composedPath()` for proper slotted content detection

#### Testing / Reviewing

Open deploy preview - Navigate to AI Label > Test story 
click slotted button/input/link - tooltip should stay open
Click outside tooltip or press Escape - should close normally

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
~- [ ] Updated documentation and storybook examples~
~- [ ] Wrote passing tests that cover this change~
- [x] Addressed any impact on accessibility (a11y)
- [x] Tested for cross-browser consistency
- [x] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #20547, opened 2025-09-17
Revert "fix(19982): fix toggletip-clickhandler"

tay1orjones:
Reverts carbon-design-system/carbon#20094

@devadula-nandan can you provide the issue or context you mentioned why this needs to be reverted?

If reverted we'll need to issue a patch that includes this and https://github.com/carbon-design-system/carbon/pull/20548


tay1orjones:
Closing in favor of a fix in https://github.com/carbon-design-system/carbon/pull/20552

PR #20554, opened 2025-09-18
fix: move rest prop from button to th

### LINKED ISSUE #20460: [Bug]: props are passed into button element instead of th element
### Package

@carbon/react

### Browser

_No response_

### Package version

1.89.0

### React version

19.1.0

### Description

Conversation from Slack:
Person1: I was recently working on nested tables using Carbon Table and encountered an issue with setting the header width. When a column is not sortable, the implementation passes ...rest (all unused props) directly to the `<th>` element. In this case, the header width is applied correctly.
However, when a column is sortable, the implementation passes ...rest to the `<button>` inside the `<th>`. As a result, styles like width affect the button rather than the `<th>` itself, which causes the header width to be applied incorrectly.
From what I see, ...rest should probably always be passed directly to `<th>` to ensure consistent behavior.

Person2: I reviewed the code for the DataTable TableHeader component. It looks like the rest props are applied to the `<th>` element, but in the case of a sortable DataTable, it's applied to the `<button>` instead. As a result, I don't think you can directly pass styles here. The best way to apply styles would be to add a className to the parent element and then target the `<th>` inside it.
However, this is a legitimate bug that affects developer experience. I'd recommend filing an [enhancement request](https://github.com/carbon-design-system/carbon/issues/new/choose) with us to ensure ...rest props are consistently applied to the <th> element for both sortable and non-sortable columns.

### Reproduction/example

123

### Steps to reproduce

Create sortable table header with width property, width is being passed to button element instead of th element.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

Gururajj77:
Closes #20460

### Changelog

**Changed**

- Move {..rest} spread from button to <th>

#### Testing / Reviewing

pass some attributes like `width` to the `<TableHeader>`, and inspect that it get's applied to `<th>` 

## PR Checklist

### DIFF
--- a/packages/react/src/components/DataTable/TableHeader.tsx
+++ b/packages/react/src/components/DataTable/TableHeader.tsx
@@ -234,6 +234,7 @@ const TableHeader = React.forwardRef(function TableHeader(
 
   return (
     <th
+      {...rest}
       id={id}
       aria-sort={ariaSort}
       className={headerClasses}
@@ -247,8 +248,7 @@ const TableHeader = React.forwardRef(function TableHeader(
         type="button"
         aria-describedby={uniqueId}
         className={className}
-        onClick={handleClick}
-        {...rest}>
+        onClick={handleClick}>
         <span className={`${prefix}--table-sort__flex`}>
           <div className={`${prefix}--table-header-label`}>{children}</div>
           <Arrow size={20} className={`${prefix}--table-sort__icon`} />
--- a/packages/react/src/components/DataTable/TableHeader.tsx
+++ b/packages/react/src/components/DataTable/TableHeader.tsx
@@ -248,7 +248,8 @@ const TableHeader = React.forwardRef(function TableHeader(
         type="button"
         aria-describedby={uniqueId}
         className={className}
-        onClick={handleClick}>
+        onClick={handleClick}
+        {...rest}>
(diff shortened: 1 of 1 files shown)


riddhybansal (commented):
LGTM . But I just want @tay1orjones to chime up here if giving props access to both the elements are right or not. Or do we need to approach it in any other way.

tay1orjones:
The error present in CI is an example of why we probably shouldn't spread onto two different elements like this

```
TestingLibraryElementError: Found multiple elements by: [data-testid="test-id"]
```

This would apply to any additional props passed into the component. Attributes would be duplicated unless they're overridden by one specified after the `...rest`.

In the git history is there an indication as to why this was placed on the `button` instead of the `th`? My guess is there's a specific use case here that it supports and we'd need to not break that.

tay1orjones:
Also in looking at the original issue, I think there's better ways for the consumer to work around not being able to set `<TableHeader style={{'width': '##px'}} />`. Applying a `className` and using it to style the `th` is CSP-compliant and more secure than inline styles.

Gururajj77:
closing this PR as implementing this with fixing the tests would not make sense. Will look at this issue after understanding why `{...rest}` is applied to the `<button>` instead of `<th>`

PR #20561, opened 2025-09-18
refactor(toggle): deprecate `checked` in favour of `toggled`

### LINKED ISSUE #20353: Refactor `cds-toggle` to have new `toggled` attribute aliasing `checked`
<details><summary>Original issue contents</summary>
<p>
`cds-toggle` in web components, is a child component of `cds-checkbox`. This seems unnecessary since `cds-checkbox` is not in the render, and `checked` (which is from `cds-checkbox`) should not be used as an attribute for toggle, it should be `toggled`, like how it is in [React](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Toggle/Toggle.tsx#L96-L99).

The [_handleChange](https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/toggle/toggle.ts#L36-L56) function also has some elements coming `cds-checkbox` that don't seem necessary, like using the `eventChange` from `cds-checkbox`, instead of [its own](https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/src/components/toggle/toggle.ts#L175-L180), and an `indeterminate` state that isn't very useful, since `cds-toggle` doesn't have an `indeterminate` state.

```
protected _handleChange() {
    const { checked, indeterminate } = this._checkboxNode;
    if (this.disabled || this.readOnly) {
      return;
    }
    this.checked = !checked;
    this.indeterminate = indeterminate;
    const { eventChange } = this.constructor as typeof CDSCheckbox;
    this.dispatchEvent(
      new CustomEvent(eventChange, {
        bubbles: true,
        composed: true,
        detail: {
          indeterminate,
        },
      })
    );
  }
```
</p>
</details> 

After researching more into this, https://github.com/carbon-design-system/carbon/pull/20390/files#diff-a19415c58bc4b342888c2ba21944f95640bd604fcb0b2d3dbda69eeb2dcdf701, the work to do here is simply add `toggled` as a prop/attribute and have it be an alias for the existing `checked`. Mark `checked` as deprecated, pointing users to `toggled` instead. `checked` can be removed as part of the next major.

maradwan26:
Closes #20353

This PR deprecates the `checked` attribute in `cds-toggle` in favour of `toggled`. `checked` should be removed in the next major version. Until then, it still has the same functionality to prevent breaking changes to adopters using this attribute.

### Changelog

**New**

- Added new `toggled` attribute
- Added a test only story `Toggle/With Deprecated Checked` for reviewers to test that the `checked` attribute still retains its functionality
- Added `TODO` comments throughout the component's files to remove `checked` in the next major version and to also fix a naming inconsistency with `labelA` and `labelB`

**Changed**

- Updated `cds-toggle` to use `toggled`
- Deprecated `checked` attribute
- Refactored test cases to check for `toggled` attribute instead of `checked`
- Used aliasing to prevent breaking changes to users still using `checked`
    - Right now, `checked` and `toggled` should function the same with no functional/visual differences
    - I have also made sure that both attributes always show up in the DOM to continue supporting attribute-based query selection for `checked`.
 - Updated storybook controls to use `toggled`

#### Testing / Reviewing

- Go to `Toggle` in deploy preview
- All stories and controls should render and work as expected
- `Toggle/With Deprecated Checked` should render and work as expected
- CI should pass
- There should be no regression, there should be no visual/functional changes

## PR Checklist

> [!NOTE]
> The test story `Toggle/With Deprecated Checked` should be removed before merging


ariellalgilmore (approved):
LGTM! thanks for showing the deprecated story too!

tay1orjones (approved):

packages/web-components/src/components/toggle/toggle.stories.ts
```diff
@@ -89,14 +89,47 @@ export const Default = {
...
+      label-a="${labelA}"
+      ?disabled="${disabled}"
+      ?hideLabel="${hideLabel}"
+      label-text="${labelText}"
+      size="${size}"
+      ?read-only=${readOnly}
+      ?checked="${checked}"></cds-toggle>
+  `,
+};
+
```
tay1orjones:
```suggestion
```

PR #19908, opened 2025-07-14
fix: Composed modal and modal close on esc

riddhybansal:
Closes #19850 

Fix ESC key not working in ComposedModal  and modal when no element has focus

### Changelog

**New**

- Added document-level ESC key listener to ComposedModal and Modalfor consistent escape key behavior
- Added comprehensive test coverage for ESC key functionality regardless of focus state

#### Testing / Reviewing

- [ ] Open a ComposedModal or Modal with form inputs
- [ ] Click outside the modal to remove focus from any input field
- [ ] Press ESC key - modal should close (this would fail before the fix)
- [ ] Verify ESC key still works when modal elements have focus

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
- [ ] ~Updated documentation and storybook examples~
- [x] Wrote passing tests that cover this change
- [x] Addressed any impact on accessibility (a11y)
- [x] Tested for cross-browser consistency
- [x] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request guide](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/reviewing-pull-requests.md)

PR #20586, opened 2025-09-24
fix: closing of all modals when esc is clicked

### LINKED ISSUE #20426: [Bug]: Pressing `esc` should only close a single/focused modal
### Package

@carbon/react

### Browser

Chrome

### Package version

1.90

### React version

19

### Description

While there is not formal support for nested modals, our partner libraries make use of them. A recent change, probably https://github.com/carbon-design-system/carbon/pull/19908, made it so that a press of <kbd>esc</kbd> closes all open modals instead of just the topmost/focused one.

* https://github.com/carbon-design-system/ibm-products/issues/8174

IIRC the spec for `<dialog>` actually supports this idea now too. This should be fixed and a test written to cover it. It might be best for that test to have a comment or be wrapped in a describe block or something saying "nesting of modals is not formally supported but tested here to ensure downstream compatibility with partner libraries"

### Reproduction/example

https://deploy-preview-8171--carbon-for-ibm-products.netlify.app/?path=/story/components-tearsheet--stacked

### Steps to reproduce

open multiples and press esc, all are closed at once

### Suggested Severity

### Application/PAL

_No response_

### Code of Conduct

riddhybansal:
Closes #20426

Pressing esc should only close single top modal/composed modal.

### Changelog

**New**

- Added logic to check top modal first and then close it.
- Added test cases for the same.

#### Testing / Reviewing

I have added the test stories for the nested behaviour. Only top modal should close when esc is pressed in both modal and composed modal

## PR Checklist


Gururajj77 (approved):
LGTM

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -529,20 +533,33 @@ const ComposedModalDialog = React.forwardRef<
...
 
-    const handleEscapeKey = (event) => {
+    const handleEscapeKey = (event: globalThis.KeyboardEvent) => {
       if (match(event, keys.Escape)) {
-        event.preventDefault();
-        event.stopPropagation();
-        closeModal(event);
+        const allModalContainers = document.querySelectorAll(
+          '[role="dialog"][aria-modal="true"]'
+        );
```
tay1orjones: For the original fix in https://github.com/carbon-design-system/carbon/pull/19908 did you look into the previous fix from https://github.com/carbon-design-system/carbon/pull/11609?

If we could rely on the `onKeyDown` on the `<Layer>` to fire appropriately, it would be scoped to only the current modal and so I think it would close only the topmost modal, one at a time. No need for the document event handler.

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -529,20 +533,33 @@ const ComposedModalDialog = React.forwardRef<
...
+
+        // Proper visibility check that handles visibility:hidden
+        const visibleModals = Array.from(allModalContainers).filter((modal) => {
+          const styles = window.getComputedStyle(modal);
+          return styles.display !== 'none' && styles.visibility === 'visible';
+        });
```
tay1orjones: Could `enable-presence` help here? Or if we can remove the document event handler, I think this could be deleted?

tay1orjones (changes requested):
Can we do this without using an event handler on the document?

maradwan26 (commented):
This may be out of scope for this but I thought it'd be worth mentioning:

- The second nested modal can't be accessed via keyboard navigation

https://github.com/user-attachments/assets/8af10af3-c95d-4361-b2d4-283204fd1c5d


The esc closing fix works as expected though👍

riddhybansal:
Hi @mbgower I read in the [documentation](https://carbondesignsystem.com/components/modal/usage/) that Focus should  remain trapped in the dialog until it is closed. When a user clicks outside a modal that's set to remain open (not close on outside click), I've chosen to move focus to the primary action button. Does this implementation correctly maintain the focus trap requirement specified in Carbon's modal documentation?

mbgower:
> Hi @mbgower I read in the [documentation](https://carbondesignsystem.com/components/modal/usage/) that Focus should remain trapped in the dialog until it is closed. When a user clicks outside a modal that's set to remain open (not close on outside click), I've chosen to move focus to the primary action button. Does this implementation correctly maintain the focus trap requirement specified in Carbon's modal documentation?

I would call this question not directly related to this discussion. The issue here is to ensure that keyboard-based dismissals are granular, and applied in an appropriate hierarchy. What you are asking about concerns pointer actions.

Further, I do not see the guidance you are referring to specifically. Under passive modal I see:

> Dismissing a passive modal
> Passive modals are persistent until dismissed in one of the following ways.
> 
> - x: Clicking the close x icon in the upper right will close the modal without submitting any data and return the user to its previous context.
> - Click elsewhere: Clicking outside the passive modal area will automatically close the modal.
> - Esc: Pressing ESC on the keyboard

I _do_ notice that some of the other modals do not include the 'click elsewhere' guidance for dismissal.

In any event, I would definitely not advocate that clicking outside the modal should reposition the focus _in_ the modal. If clicking outside the modal does not dismiss the modal, then I conclude that the viewable screen outside the modal is truly inert, and clicking there will have no effect on the modal.

PR #20600, opened 2025-09-25
fix(modal): fix scroll on focus change

### LINKED ISSUE #20591: [Bug]: Modal body scrolls when focus changes
### Package

@carbon/react

### Browser

Chrome

### Package version

v1.91.0

### React version

v18.2.0

### Description

Swapping focuses on a scrollable modal body scrolls the modal to the bottom. This also prevents certain functions, like button onClicks from triggering. This seems to be for any focus swaps on the modal.

After the scroll, once an item is focused, click on it / using it will work.

I would expect it would not automatically scroll down to the bottom if focus changes, and I would expect button onClicks to work.

I know sev 1 may seem over severe for an issue like this that technically has a workaround, since once an item in the modal body has focus you can use it, but honestly this makes scrollable modals so frustrating to use that they may as well be unusable.

https://github.com/user-attachments/assets/64f5974d-1b27-4e82-a7b7-122415e17af8

### Reproduction/example

https://stackblitz.com/edit/eumm8zsk-xov5xz7n?file=src%2FApp.jsx

### Steps to reproduce

- Focus the modal body
- Click the edit button next to the `Edit start time` field
- The modal will scroll down instead of triggering the button's function. Furthermore, in this case you can't even see that you were unable to enter edit mode since it scrolled it off screen

### Suggested Severity

### Application/PAL

Financial Transaction Manager 

### Code of Conduct

maradwan26:
Closes #20591

Fixes a bug where the composed modal body would scroll to the bottom when focus changes

### Changelog

**New**

- Added a test stories for reviewers to test this PR, should be removed before merging.
    - `Modal/Composed Modal Scroll Bug Fix`
    - `Modal/Modal Scroll Bug Fix`

**Changed**

- Removed the for loop that was used to compute spacing in favour of `currentActiveNode.scrollIntoView({ block: 'center' });`

#### Testing / Reviewing

- Go to `Modal/Composed Modal Scroll Bug Fix` and `Modal/Modal Scroll Bug Fix`
- Click on the blue button
- The modal body should not scroll to the bottom

## PR Checklist


tay1orjones (approved):

ariellalgilmore (approved):
LGTM!

PR #20322, opened 2025-08-23
fix(header): adjust edge-aligned popover positioning

adamalston:
Closes https://github.com/carbon-design-system/carbon/issues/19240

Adjusted edge-aligned `Popover` positioning in the `Header`.

### Changelog

**Changed**

- Adjusted edge-aligned `Popover` positioning in the `Header`.

**Removed**

- Deleted a `Test` story that was seemingly overlooked during a previous pull request merge.

#### Testing / Reviewing

How to test: Check out my branch and set up a test story using the code from the issue.

I'm not sure how to add testing for this fix. Any help with that would be appreciated since I'm fixing a regression.

start

undefined

end

These changes also addressed the issue but they broke styling for other components.

```diff
diff --git a/packages/styles/scss/components/popover/_popover.scss b/packages/styles/scss/components/popover/_popover.scss
index ce5b361f1..477023493 100644
--- a/packages/styles/scss/components/popover/_popover.scss
+++ b/packages/styles/scss/components/popover/_popover.scss
@@ -235,7 +235,7 @@ $popover-caret-height: custom-property.get-var(
     > .#{$prefix}--popover
     > .#{$prefix}--popover-content {
     inset-block-end: 0;
-    inset-inline-start: 0;
+    inset-inline-start: 1rem;
     transform: translate(
       calc(-1 * $popover-offset),
       calc(100% + $popover-offset)
@@ -261,7 +261,7 @@ $popover-caret-height: custom-property.get-var(
     > .#{$prefix}--popover
     > .#{$prefix}--popover-content {
     inset-block-end: 0;
-    inset-inline-end: 0;
+    inset-inline-end: 1rem;
     transform: translate($popover-offset, calc(100% + $popover-offset));
   }

@@ -272,7 +272,7 @@ $popover-caret-height: custom-property.get-var(
     > .#{$prefix}--popover
     > .#{$prefix}--popover-caret {
     inset-block-end: 0;
-    inset-inline-end: 0;
+    inset-inline-end: 1rem;
     inset-inline-start: auto;
   }

@@ -281,7 +281,7 @@ $popover-caret-height: custom-property.get-var(
     > .#{$prefix}--popover-caret {
     inset-block-end: 0;
     inset-inline-end: auto;
-    inset-inline-start: 0;
+    inset-inline-start: 1rem;
     transform: translate(50%, calc($popover-offset));
   }
```

## PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
[...]

Issue #19837, opened 2025-07-03
React/WC Parity: Popover


heloiselui:
See parent issue https://github.com/carbon-design-system/carbon/issues/17992 for acceptance criteria.

-> include arguments inside the flip which allow for the auto adjusting in the correct place: 
https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Popover/index.tsx#L260-L298
https://floating-ui.com/docs/flip#flipalignment

this should fix the toggletip alignment #19790

> [!IMPORTANT]  
> Every PR needs a ux and visual review from the design team to ensure consistency. Please be sure PRs have [status: visual review 🔍](https://github.com/carbon-design-system/carbon/labels/status%3A%20visual%20review%20%F0%9F%94%8D) applied and request a review from @carbon-design-system/design

ariellalgilmore:
Make sure this gets included <img width="859" height="818" alt="Image" src="https://github.com/user-attachments/assets/46635544-68e3-47f3-b32f-ee398103bc59" />

PR #20607, opened 2025-09-26
fix(popover): alignment

ariellalgilmore:
Closes #N/A

in Carbon labs we noticed from #20322 was causing the popover  header to misaligned
image

Then, after talking to @laurenmrice we were that none of the carets seemed to be centered, which partially came from here https://github.com/carbon-design-system/carbon/pull/17964

Now the popovers should be correctly aligned while still maintaining the above fixes, but not causing the misalignment

### Changelog

**New**

- {{new thing}}

**Changed**

- adjusted popover styles
- added back long text definitionTooltip temporarily to check styles
- added alignment in UIShell's HeaderGlobalActions temporarly to check styles

**Removed**

- header specific styles the popover should be able to handle the styles

#### Testing / Reviewing

- Check all Popover stories (ltr and rtl)
- Check DefinitionTooltip stories (ltr and rtl)
- check tooltip stories

## PR Checklist

### DIFF
--- a/packages/react/src/components/Popover/index.tsx
+++ b/packages/react/src/components/Popover/index.tsx
@@ -327,6 +327,7 @@ export const Popover: PopoverComponent & {
               }),
             arrow({
               element: caretRef,
+              padding: 16,
             }),
             autoAlign && hide(),
           ],
--- a/packages/react/src/components/Tooltip/DefinitionTooltip.stories.js
+++ b/packages/react/src/components/Tooltip/DefinitionTooltip.stories.js
@@ -93,3 +93,72 @@ Default.argTypes = {
     },
   },
 };
+
+export const WithLargeText = (args) => {
+  const definition = 'Example definition';
+  return (
+    <p>
+      Custom domains direct requests for your apps in this Cloud Foundry
+      organization to a{' '}
+      <DefinitionTooltip openOnHover definition={definition} {...args}>
+        URL that you own. A custom domain can be a shared domain,
+      </DefinitionTooltip>{' '}
+      a shared subdomain, or a shared domain and host.
+    </p>
+  );
+};
+
+WithLargeText.args = {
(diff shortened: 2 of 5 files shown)


maradwan26 (changes requested):
The popover is misaligned for Tab Tip RTL:

Screenshot 2025-09-29 at 2 18 06 PM

LTR works fine, though.

The rest LGTM!

ariellalgilmore:
thanks @maradwan26 updated!

maradwan26 (approved):
LGTM !

emyarod (approved):
looks good to me!

laurenmrice (changes requested):
This is looking really good @ariellalgilmore. Just found a couple small bugs:

## React

### Popover
#### Default story
The dropshadow under the caret of the popover doesn’t look correct, and looks very faint. It should look like what is happening in the Experimental Auto Align story for example.

#### Feature flag > Floating styles story
The caret of the popover is not positioned 16px away from the edges of the popover when in a start or end position. It currently looks like it’s 8px. It should be 16px.

Frame 5


### Tooltip
#### Default story and Feature flag > Floating styles story
The following alignments are not correct:
- top-left
- top-right
- bottom-left
- bottom-right
- left-top
- left-bottom
- right-bottom
- right-top

-----

## WC

### Popover

#### Tab Tip story
When the popover is open, there is a top shadow that is created between the popover and the icon button. The shadow should not be visible there.

Screenshot 2025-10-02 at 10 45 59 AM

ariellalgilmore:
Thanks @laurenmrice!! The react bugs should be updated. Looks like web components hasn't had the parity work yet, so included your comment there to make sure it gets fixed with the parity: https://github.com/carbon-design-system/carbon/issues/19837#issuecomment-3366718476

laurenmrice (approved):
Looks great !

PR #20608, opened 2025-09-28
fix(select): use title prop for tooltip instead of first SelectItem

### LINKED ISSUE #20582: [Bug]: <Select> uses first <SelectItem> value as tooltip instead of respecting title prop Description
### Package

@carbon/react

### Browser

Chrome

### Package version

last

### React version

_No response_

### Description

When rendering a Carbon `<Select>`, the browser tooltip on the native `<select>` element shows the value of the first `<SelectItem>` instead of the selected or even the title prop passed to `<Select>`.

**Expected behavior**

If a initial value is set, it should be displaying the value of the selected item as tooltip.
If a title prop is provided, it should be forwarded to the native `<Select>` element and used as the tooltip.

**Actual behavior**

The tooltip shows the initial option’s value (first `<SelectItem>`) rather than the intended title or initial value.

### Reproduction/example

https://stackblitz.com/edit/github-vfphhms5?file=src%2FApp.tsx

### Steps to reproduce

Render the component the link above.

Hover over the component.

Tooltip shows "Monday" (the first option) instead of "Tooltip test".
Tooltip shows "Monday" (the first option) instead of "Tuesday" (selected value - supposed value when removing the title)

Setting value 2 as initial value -> initial tooltip shows "Monday" instead of "Tuesday"

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

Vignesh-Loganathan-IBM-1:
Closes #20582

Fixed Select component to properly handle value prop for tooltips, ensuring correct display of selected option text when using controlled components and  use title prop for tooltip instead of first SelectItem

### Changelog

**Changed**

Tooltip now correctly uses the `title` prop when provided, or falls back to the selected item's label. Previously, the tooltip incorrectly defaulted to the first `SelectItem` value regardless of selection or props. This fix improves accessibility and user experience when using the component in controlled mode.

#### Testing / Reviewing

- [x] Rendered the Select component with value and title props; verified tooltip shows the title or selected value correctly and no longer shows the first SelectItem value.
 
## PR Checklist

### DIFF
--- a/packages/react/src/components/Select/Select.tsx
+++ b/packages/react/src/components/Select/Select.tsx
@@ -185,14 +185,19 @@ const Select = React.forwardRef(
         React.isValidElement<SelectItemProps>(child)
     );
 
-    // Find the default option based on the specified defaultValue
-    const defaultOption = validChildren.find(
-      (child) => child.props?.value === other?.defaultValue
+    // Find the default option based on the specified defaultValue or value
+    const selectedValue = other?.value || other?.defaultValue;
+    const selectedOption = validChildren.find(
+      (child) => child.props?.value === selectedValue
     );
 
-    // Use the default option's text if available; otherwise, fallback to the first option's text
+    // Use the provided title prop, or the selected option's text if available;
+    // otherwise, fallback to the first option's text
     const initialTitle =
-      defaultOption?.props?.text || validChildren[0]?.props?.text || '';
+      other?.title ||
+      selectedOption?.props?.text ||
+      validChildren[0]?.props?.text ||
+      '';
 
     const [title, setTitle] = useState(initialTitle);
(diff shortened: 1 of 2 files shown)


Vignesh-Loganathan-IBM-1:

packages/react/src/components/Select/Select.tsx
```diff
@@ -471,3 +476,5 @@ Select.propTypes = {
 };
 
 export default Select;
+
+// Made with Bob
```
tay1orjones:
```suggestion
```

tay1orjones (changes requested):

tay1orjones (approved):
LGTM

maradwan26 (approved):
LGTM

PR #20637, opened 2025-10-02
fix: add role="group" to TableToolbar for proper accessibility

### LINKED ISSUE #20623: [a11y]: TableToolbar: inappropriate region role
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.89.0

### React version

18

n/a

### Assistive technology

no

### Description

Carbon currently renders the `TableToolbar` as a `<section>` with an `aria-label`, either a user-defined value or the default value of "data table toolbar".

This means that the `TableToolbar` is rendered with an implicit `region` role.  I think it's inappropriate for such a small thing to be rendered as a region:

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/region_role):

> The region role is an [ARIA landmark](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles#3._landmark_roles) role. The region role should be reserved for sections of content sufficiently important that users will likely want to navigate to the section easily and to have it listed in a summary of the page.

And similarly, [W3](https://www.w3.org/TR/WCAG20-TECHS/ARIA20.html) says that:

> The role of region should be used prudently, because if overused they can make the page overly verbose for screen reader users.

It would be nice to use role="toolbar" but the searchbox in the middle of the toolbar would make left-right arrow key navigation difficult.

I think any of these renderings would be more appropriate:

```html
<div role="group" aria-label="data table toolbar">
<section role="group" aria-label="data table toolbar">
<fieldset> <legend class="cds--visually-hidden">data table toolbar</legend> … </fieldset>
```

### WCAG 2.1 Violation

https://www.w3.org/TR/WCAG21/#name-role-value

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-datatable-toolbar--default

### Steps to reproduce

Just look at the generated HTML.

### Suggested Severity

None

### Code of Conduct

nehaaprasad:
**Closes #20623**

**Short Description:** Fixed TableToolbar accessibility issue by adding explicit `role="group"` to prevent inappropriate region landmark behavior.

### Changelog

**Changed**
- Added `role="group"` attribute to TableToolbar `<section>` element to override implicit region role
- Updated test snapshot to reflect new HTML structure

**Accessibility**
- Resolved a11y violation where TableToolbar created inappropriate landmark region for screen readers


maradwan26 (approved):
LGTM !

kennylam (approved):

PR #20640, opened 2025-10-02
fix(modal): remove gradient when body container height is short

### LINKED ISSUE #20259: [a11y]: MODAL - Loss of content is observed when page is magnified to 400%
### Package

@carbon/react

### Browser

Firefox

### Operating System

MacOS

### Package version

1.81.0

### React version

17.0.2

AVT2 manual testing by AVT team

### Assistive technology

_No response_

### Description

Issue description: Loss of content is observed when page is magnified to 400%

using storybook as an example : https://react.carbondesignsystem.com/iframe.html?globals=&args=&id=components-modal--danger-modal&viewMode=story

zoom in to 400%

Image

### WCAG 2.1 Violation

WCAG Mapping: 1.4.10 Reflow

### Reproduction/example

 https://react.carbondesignsystem.com/iframe.html?globals=&args=&id=components-modal--danger-modal&viewMode=story

### Steps to reproduce

https://react.carbondesignsystem.com/iframe.html?globals=&args=&id=components-modal--danger-modal&viewMode=story
zoom browser to 400%

### Suggested Severity

None

### Code of Conduct

maradwan26:
Closes #20259

Fixes an accessibility issue where the modal body content was difficult to interact with when its height was small due to being obscured by the gradient styling.

### Changelog

**New**

- Added a `useResizeObserver` to `Modal` and `ComposedModal` that adds the `cds--modal-scroll-content--no-fade` class (which removes the gradient styling) when the height of the modal body container is <= 300px (let me know if this value should be changed).

**Removed**
- Removed the `useIsomorphicEffect`, `useState`, and `debounce` that were used to compute the isScrollable boolean in favour of simply comparing `scrollHeight` to `clientHeight` since `useResizeObserver` allows `clientHeight` to update when it re-renders

#### Testing / Reviewing

- Gradient should be removed when the modal container body is short, allowing the modal content to be visible
- Check the test story `Composed Modal/With Scrolling Content Max Header` for an edge case where the header is extremely large, the content should still be accessible

## PR Checklist

> [!NOTE]
> The test story `Composed Modal/With Scrolling Content Max Header` should be removed before merging


packages/react/src/components/Modal/Modal.tsx
```diff
@@ -621,6 +621,23 @@ const ModalDialog = React.forwardRef(function ModalDialog(
...
+        el.style.setProperty('mask-image', 'none');
+      } else {
+        el.style.removeProperty('mask-image');
+      }
+    });
+    resizeObserver.observe(contentRef.current);
+
+    return () => resizeObserver.disconnect();
+  }, [contentRef]);
+
```
tay1orjones: A couple suggestions:

* `useEffect` isn't needed because changes to a ref [don't trigger a re-render](https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array).
* Using `useResizeObserver` instead will cut a ton of the boilerplate
* Applying a class feels like a more open-ended and reusable way to implement. Other framework authors, including our own wc, could use resizeobserver to apply the class just the same. App devs can reuse the class in other situations for their own use as well, or overwrite the styles on that class altogether.

```js
// put this up above the contentClasses definition on line ~490
const { height } = useResizeObserver(contentRef);

// ... then set the new class based on the height
const contentClasses = classNames(`${prefix}--modal-content`, {
    [`${prefix}--modal-scroll-content`]: hasScrollingContent || isScrollable,
    [`${prefix}--modal-scroll-content--no-fade`]: height >= 300,
  });

```

  
One bonus opportunity with this is we could loop in the isScrollable variable to this. Would make it simpler, get rid of the useIsomporphicEffect it's in, and remove a onresize callback from the window, too. 

```diff
// put this up above the contentClasses definition on line ~490
const { height } = useResizeObserver(contentRef);

- const [isScrollable, setIsScrollable] = useState(false);
+ const isScrollable = contentRef?.current?.scrollHeight > height

// ... then set the new class based on the height
const contentClasses = classNames(`${prefix}--modal-content`, {
    [`${prefix}--modal-scroll-content`]: hasScrollingContent || isScrollable,
    [`${prefix}--modal-scroll-content--no-fade`]: height >= 300,
  });

// ...delete everything down lower
-useIsomorphicEffect(() => {
-    if (contentRef.current) {
-      setIsScrollable(
-        contentRef?.current.scrollHeight > contentRef.current.clientHeight
-      );
-    }
-
-    function handler() {
-      if (contentRef.current) {
-        setIsScrollable(
-          contentRef.current.scrollHeight > contentRef.current.clientHeight
-        );
-      }
-    }

-    const debouncedHandler = debounce(handler, 200);
-    window.addEventListener('resize', debouncedHandler);
-    return () => {
-      debouncedHandler.cancel();
-      window.removeEventListener('resize', debouncedHandler);
-    };
-  }, []);
```

Similar/same approach should work for ComposedModal.

tay1orjones (commented):
Let me know what you think about this 👀

packages/react/src/components/Modal/Modal.tsx
maradwan26: @tay1orjones 

Yeah that's a really good idea the class can be useful for other use cases and this is gonna clean up the code a bunch

packages/react/src/components/ComposedModal/ComposedModal.tsx
```diff
@@ -83,40 +81,24 @@ export const ModalBody = React.forwardRef<HTMLDivElement, ModalBodyProps>(
   ) {
     const prefix = usePrefix();
     const contentRef = useRef<HTMLDivElement>(null);
-    const [isScrollable, setIsScrollable] = useState(false);
+
+    const { height } = useResizeObserver({ ref: contentRef });
+
+    const isScrollable =
+      !!contentRef.current &&
+      contentRef?.current?.scrollHeight > contentRef?.current?.clientHeight;
```
tay1orjones, line 93: * `isScrollable` implicitly relies on `height`. We could make that dependency more obvious for the future with a useMemo.
* With the double negate in place we should be able to remove the optional chaining.

```suggestion
    const { height } = useResizeObserver({ ref: contentRef });

    const isScrollable = React.useMemo(() => {
      const node = contentRef.current;
      return !!node && node.scrollHeight - node.clientHeight;
    }, [height, children]); // Would including children here make sense?
```

packages/styles/scss/components/modal/_modal.scss
```diff
@@ -292,6 +292,10 @@
       linear-gradient(to top, $layer 0, 2px, transparent 2px);
   }
 
+  .#{$prefix}--modal-scroll-content--no-fade {
+    mask-image: none !important; /* stylelint-disable-line declaration-no-important */
+  }
+
```
tay1orjones, line 303: Will this add enough specificity to avoid the `!important`?

```suggestion
  .#{$prefix}--modal-scroll-content.#{$prefix}--modal-scroll-content--no-fade {
    mask-image: none;
  }

```

tay1orjones (approved):
Two more suggestions but overall LGTM, could merge as is imo

packages/react/src/components/ComposedModal/ComposedModal.tsx
maradwan26: @tay1orjones 
I tested this locally and `eslint` throws an error, it considers the dependencies unnecessary

`493:6  warning  React Hook useMemo has unnecessary dependencies: 'children' and 'height'. Either exclude them or remove the dependency array  react-hooks/exhaustive-deps`

Should I just add a comment block explaining the dependency instead?

packages/styles/scss/components/modal/_modal.scss
maradwan26: Yes thanks! I just had to add some more selectors to target for when it has the decorator/slug classes too for the AI-Label variant

riddhybansal (approved):

heloiselui (approved):
LGTM, just need to remove the test

PR #20648, opened 2025-10-03
fix(NumberInput): allow deleting 0 before typing another digit

### LINKED ISSUE #20588: [Bug]: Number Input user cannot delete single digit value
### Package

@carbon/react

### Browser

_No response_

### Package version

1.76.0

### React version

_No response_

### Description

The user wants to replace the 0 with another digit, but is user cannot to delete the 0. The user has to type the 5 first resulting on 05, or 50 before removing the extra 0. The user should be able to delete the 0 before typing the 5

### Reproduction/example

Number Input in a Form activity cannot delete single digit value

### Steps to reproduce

Add a Number Input, and set initial value to 0 (zero)

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

Vignesh-Loganathan-IBM-1:
Closes #20588

this PR updates the handleOnChange logic in the NumberInput component to handle a specific edge case where users want to delete a leading 0 before entering another digit.

### Changelog

**Changed**

- Fixed issue where users couldn't delete a leading `0` before typing a new digit. Input handling logic (`handleOnChange`) was updated to support smoother editing of numeric values.

#### Testing / Reviewing

Rendered NumberInput with min={0} and verified that deleting a leading 0 works as expected.

## PR Checklist


Vignesh-Loganathan-IBM-1:

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -568,6 +568,20 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
       }
 
       if (type === 'number') {
+        // Special handling for empty input to allow deleting 0 before typing another digit
+        if (event.target.value === '') {
+          const state = {
+            value: allowEmpty ? '' : 0,
+            direction: 'down',
+          };
+          setValue('');
```
heloiselui:
```suggestion
          setValue(state.value);
```

heloiselui (commented):
Hey! Just a small thing, when the input is empty, you're sending `state.value` as 0 (if allowEmpty is false), but still calling `setValue('')`. Maybe it makes more sense to call `setValue(state.value)`.

heloiselui:
Hey, the CI test is failing. I think you need to check the tests, you can also before merging, you can run `yarn test` in the `packages/react` directory to verify locally.

packages/react/src/components/NumberInput/NumberInput.tsx
```diff
@@ -568,6 +568,20 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
...
+            direction: 'down',
+          };
+          setValue(state.value);
+
+          if (onChange) {
+            onChange(event, state);
+          }
+          return;
+        }
+
```
tay1orjones, line 584: Could you build this into the existing logic in the lines below instead of repeating it all up here and doing an early return?

tay1orjones:
I'm actually not sure we need this, I think it's working as expected. See https://github.com/carbon-design-system/carbon/issues/20588#issuecomment-3378461287

heloiselui:
Thank you very much for your contribution @Vignesh-Loganathan-IBM-1, we really appreciate the time you took to help with this issue. However, I believe we need more information from the author to better understand what needs to be improved. Because we're actually not sure we need this, as it seems to be working as expected (see #20588 comment).

I’ll be closing this PR for now. Once we have more details and the intention to move forward, we can reopen it and continue from where we left off.

PR #20653, opened 2025-10-06
Fix checked attribute not updating on TileGroup selection

### LINKED ISSUE #19738: [Bug]: TileGroup not setting `checked` attribute correctly on input element
### Package

@carbon/react

### Browser

Chrome

### Package version

1.85.0

### React version

17.0.2

### Description

Selecting different tiles within the TileGroup does not move the `checked` attribute to the correct input element. It stays on the initially selected tile and never changes. Functionality itself does not appear to be broken but this will cause problems for some assistive technologies.

### Reproduction/example

https://react.carbondesignsystem.com/?path=/story/components-tile--radio

### Steps to reproduce

This is seen in the storybook example.
https://react.carbondesignsystem.com/?path=/story/components-tile--radio

### Suggested Severity

### Application/PAL

IBM Cloud

### Code of Conduct

Vignesh-Loganathan-IBM-1:
Closes #19738

This PR resolves an accessibility issue where the checked attribute on input elements within the TileGroup component did not update when selecting a different tile. The attribute remained on the initially selected tile, potentially causing issues for assistive technologies that rely on accurate state representation.

**Changed**

- the RadioTile component correctly manages the `checked` attribute on the DOM element for better accessibility support. The implementation explicitly sets or removes the attribute using `setAttribute` and `removeAttribute`, which is important for assistive technologies to properly recognize the checked state of the radio button.

#### Testing / Reviewing

The tests verify that:

When checked={true}, the DOM element has the checked attribute set
When checked={false}, the DOM element does not have the checked attribute
The checked attribute updates correctly when the prop changes

## PR Checklist

### DIFF
--- a/packages/react/src/components/RadioTile/RadioTile.tsx
+++ b/packages/react/src/components/RadioTile/RadioTile.tsx
@@ -193,7 +193,27 @@ const RadioTile = React.forwardRef(
           tabIndex={!disabled ? tabIndex : undefined}
           type="radio"
           value={value}
-          ref={ref}
+          ref={(el) => {
+            // Set the checked attribute directly on the DOM element for assistive technologies
+            if (el) {
+              if (checked) {
+                el.setAttribute('checked', '');
+              } else {
+                el.removeAttribute('checked');
+              }
+            }
+
+            // Forward the ref
+            if (ref) {
+              if (typeof ref === 'function') {
+                ref(el);
+              } else {
+                (
+                  ref as React.MutableRefObject<HTMLInputElement | null>
+                ).current = el;
+              }
+            }
+          }}
           required={required}
         />
         <label {...rest} htmlFor={inputId} className={className}>
--- a/packages/react/src/components/RadioTile/RadioTile-test.js
+++ b/packages/react/src/components/RadioTile/RadioTile-test.js
@@ -126,6 +126,7 @@ describe('RadioTile', () => {
       expect(ref.current.type).toEqual('radio');
       expect(ref.current.value).toEqual('some test value');
     });
+
     it('should pass "required" prop to the input element', () => {
       render(
         <RadioTile required value="some test value">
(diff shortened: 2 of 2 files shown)


Vignesh-Loganathan-IBM-1:

heloiselui (approved):
Thanks for doing that @Vignesh-Loganathan-IBM-1, LGTM!

heloiselui:
@Vignesh-Loganathan-IBM-1 
Btw, don't forget to update the checklist in the template like you did in the other PRs. Thanks!
image

packages/react/src/components/RadioTile/RadioTile.tsx
```diff
@@ -193,7 +193,27 @@ const RadioTile = React.forwardRef(
...
+            // Forward the ref
+            if (ref) {
+              if (typeof ref === 'function') {
+                ref(el);
+              } else {
+                (
+                  ref as React.MutableRefObject<HTMLInputElement | null>
+                ).current = el;
+              }
+            }
```
kennylam, line 215:
```suggestion
            if (el) el.toggleAttribute('checked', !!checked);

            // Forward the ref
            if (typeof ref === 'function') {
              ref(el);
            } else if (ref) {
              (ref as React.MutableRefObject<HTMLInputElement | null>).current = el;
            }
```

kennylam (commented):
@Vignesh-Loganathan-IBM-1 thanks for doing this, this is great. Just a suggestion to make the conditions a little more concise.

tay1orjones:
Should `checked` be set like this? From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/checked#:~:text=The%20presence%20of,the%20defaultChecked%20property.):

> The presence of the HTML [checked](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#checked) attribute indicates the checkbox is checked by default. It does not indicate whether this checkbox is currently checked: if the checkbox's state is changed, this content attribute does not reflect the change; only the HTMLInputElement's checked IDL property is updated. The checked attribute is reflected by the [defaultChecked](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultChecked) property.

`RadioButton` doesn't imperatively update the attribute or anything, it just passes `checked` along to the underlying `input`.

PR #20658, opened 2025-10-06
chore(gh-workflows): update CI workflow logic

kennylam:
This fix addresses something I've been curious about for a while, where the `merge-playwright-reports` job always fails on draft PRs, like [this one](https://github.com/carbon-design-system/carbon/actions/runs/17974870004/job/51158011521?pr=20587#step:7:1). The issue is that the VRT jobs do not create any artifacts for dry runs (draft PRs). Since there is no draft PR condition in `merge-playwright-reports`, it tries to run it no matter what, then fails when it doesn't see any reports. All I've done is add a draft condition to the merge step so it matches the run conditions of the VRT job.

One other small change is in the VRT dry run step itself.

```diff
- github.event.pull_request.draft != false && github.base_ref != 'main'
+ github.event.pull_request.draft == true && github.base_ref != 'main'
```

The original condition `github.event.pull_request.draft != false` could incorrectly evaluate to `true` on `push` events, e.g. `push: branches: -main`. Since [push](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push) events don't include a `pull_request` object, `github.event.pull_request.draft` is `null`, and `null != false` is `true`. This caused the step to run on pushes even though it's meant for draft PRs.

This hasn't caused any problems so far, but I've updated the condition to `==true` to ensure it only runs in draft PRs to match the job's intention.

### Changelog

**Changed**

- update `merge-playwright-reports` to not run on draft PRs
- clean up VRT dry run logic

#### Testing / Reviewing

CI workflow should run and everything should pass. In future draft PRs, the `merge-playwright-reports` job should no longer run (and fail).

## PR Checklist


heloiselui (approved):
LGTM 🚀

tay1orjones (approved):

PR #20659, opened 2025-10-06
fix(select): use aria-describedby and visually hide label

### LINKED ISSUE #20649: [a11y]: (WebComponents) CDS-Select - There is no way to visually hide the label while maintaining accesiblity
### Package

@carbon/web-components

### Browser

Chrome

### Operating System

Windows

### Package version

2.27.1

### React version

N/A

Chrome Lighthouse

### Assistive technology

_No response_

### Description

When `label-text` is provided, and `hide-label` is true. The following Lighthouse error is reported:

 `Select elements do not have associated label elements.`

There doesn't seem to be any workaround to apply an aria-label to the html select component, which forces the user to render a label or have this error.

Possible solution: When hide-label is true, use label-text as aria-label (if provided).

### WCAG 2.1 Violation

WCAG 2.1 (A), WCAG 2.0 (A), WCAG 2.2 (A), Section 508, Trusted Tester, EN 301 549

### Reproduction/example

https://stackblitz.com/edit/github-xfkunee4-qfyqlqf8?file=index.html

### Steps to reproduce

Set `label-text` to 'Any value'
Set `hide-label` tor true

Scan component

### Suggested Severity

None

### Code of Conduct

olihf-dematic:
Closes #20649

Fixes the HTML select element in CDS-Select never having a defined `aria-label` even when `label-text` is provided

### Changelog

**New**

- `select`: Apply `aria-label` when `hide-label` is true, and `label-text` is defined

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

- On the storybook playground for CDS-Select (:
 - Ensure `label-text` is defined
 - Set `hide-label` to true
 - Verify an aria-label, with content `label-text` is applied to the html select element

## PR Checklist

### DIFF
--- a/packages/web-components/src/components/select/select.ts
+++ b/packages/web-components/src/components/select/select.ts
@@ -448,6 +448,7 @@ class CDSSelect extends FormMixin(LitElement) {
         id="input"
         class="${inputClasses}"
         ?disabled="${disabled}"
+        aria-label="${ifDefined(hideLabel ? labelText : undefined)}"
         aria-readonly="${String(Boolean(readonly))}"
         aria-invalid="${String(Boolean(invalid))}"
         aria-describedby="${ifDefined(!invalid ? undefined : 'invalid-text')}"
--- a/packages/web-components/src/components/select/select.ts
+++ b/packages/web-components/src/components/select/select.ts
@@ -421,6 +421,7 @@ class CDSSelect extends FormMixin(LitElement) {
     const labelClasses = classMap({
       [`${prefix}--label`]: true,
       [`${prefix}--label--disabled`]: disabled,
+      [`${prefix}--visually-hidden`]: hideLabel,
     });
 
     const helperTextClasses = classMap({
@@ -448,7 +449,6 @@ class CDSSelect extends FormMixin(LitElement) {
         id="input"
         class="${inputClasses}"
         ?disabled="${disabled}"
-        aria-label="${ifDefined(hideLabel ? labelText : undefined)}"
         aria-readonly="${String(Boolean(readonly))}"
         aria-invalid="${String(Boolean(invalid))}"
         aria-describedby="${ifDefined(!invalid ? undefined : 'invalid-text')}"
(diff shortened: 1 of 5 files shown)


olihf-dematic:

packages/web-components/src/components/select/select.ts
```diff
@@ -448,6 +448,7 @@ class CDSSelect extends FormMixin(LitElement) {
         id="input"
         class="${inputClasses}"
         ?disabled="${disabled}"
+        aria-label="${ifDefined(hideLabel ? labelText : undefined)}"
```
ariellalgilmore: Thanks for the submission! Ideally we would try to make sure the markup is the same as the [react select](https://react.carbondesignsystem.com/?path=/story/components-select--default&args=hideLabel:!true). Looks like instead of removing the `label` we should instead add a `cds--visually-hidden` class: https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Select/Select.tsx#L217. and then looks like the `aria-describedby` should be pointing to the helper-text. With these updates then I don't think we need to add the `aria-label`

ariellalgilmore (changes requested):

packages/web-components/src/components/select/select.ts
olihf-dematic: Hi, this makes sense. I've updated the PR to use the `cds--visually-hidden` class, and removed the `aria-label`

For the `aria-describedby` I've updated this to match the react version pointing to `helper-text`, but giving priority to `error-text`. I also noted that the `aria-describedby` was not actually linked to anything, as the `error-text` id wasn't applied to anything. So I've added these ids to the helper and error-text. These don't exactly match the react version, due to shadow dom encapsulation meaning they only need to be unique to that shadow dom.

ariellalgilmore (approved):
Thank you for the updates! this looks great!

Gururajj77 (approved):
LGTM

PR #20669, opened 2025-10-07
feat(popover): add new border and backgroundToken props

### LINKED ISSUE #20592: [Feature Request]: Popover: add border and background props
Design issue https://github.com/carbon-design-system/carbon-labs/issues/695
Related issue https://github.com/carbon-design-system/carbon/issues/20593

When Popover is used inside a Header, the default background (layer-01) can blend into content that uses a different theme. To address this, we need to add two additional props:

- [ ] New boolean `border` prop, should use `$border-subtle`
- [ ] New `background` prop, with the default being `layer` and an option for `background`. 

### Questions
* `background="background"` feels odd, but thats the token names so not sure of a less confusing option?
  * Maybe `backgroundToken="background"` ?
* ~Should we add a third option of `background-inverse` for consistency? Or is that confusing to have two ways to turn on highContrast~
  * This shouldn't interact with the existing `highContrast` prop because those tokens are already different values from the `$background`
    * Adding layers on the background might be an issue but not something to solve for right now
* Should this be baked into the uishell styles? `.cds--ui-shell .cds--popover {background: $background} `
  * leaning towards no, it should be configurable via a prop
* Consider adding these props to VRT

alisonjoseph:
Closes #20592

This PR adds the `border` and `backgroundToken` props to `Popover` for both React and Web Components.

### Changelog

**New**

- `border` prop in Popover for React and Web Components
- `backgroundToken` prop in Popover for React and Web Components

**Changed**

- Overhaul with how popover carets are styled, previously styles were directly on the "caret" div, they are now using the `::after` pseudoselctor for the background and `::before` for the border. Essentially it is two triangles stacked on top of each other to create the border effect. 
- WC: Tooltip now passes highcontrast attribute to tooltip-content for updated caret styling to be properly applied. 

#### Testing / Reviewing

Please test `Popover` and `Tooltip` in React and Web Components, be sure to test **all** combinations of Popover props: `align`, `autoAlign`, `dropShadow`, `caret`, `border`, `backgroundToken`, `highContrast`), and also check RTL.

## PR Checklist


ariellalgilmore:
Noticing in React the caret for the stories Tooltip-> Default and Popover -> feature flag -> Floating styles when positioning the caret at (top/bottom)-(start/end) the caret is not aligned
Deploy preview: 
Screenshot 2025-10-13 at 11 30 33 AM
Current: (the caret should be 1rem from the side, so the text and the caret align
Screenshot 2025-10-13 at 11 30 39 AM

ariellalgilmore:
Web components: 

I'm not seeing any dropshadow
Screenshot 2025-10-13 at 11 37 15 AM

alisonjoseph:
@ariellalgilmore I'm not seeing the `dropShadow` in the live web components storybook either, ~I'll take a look and see if I can easily fix here.~ **FIXED** 

And good catch on the tooltip, will take a look.

alisonjoseph:
@ariellalgilmore ready for a re-review.

heloiselui:
This looks great, I just noticed small things: 

**React:**

- `Popover React:`  `Tab Tip` > controls: `highContrast` > `true` > click in `isTabTip` "Set boolean > true/false" the icon shifts slightly closer to the top edge.

Pasted Graphic 1

**React & WC**
- `Popover Overview` > [Appearance](https://react.carbondesignsystem.com/?path=/docs/components-popover--overview#appearance), should we include `backgroundToken/border`?

thyhmdo (approved):
LGTM!

alisonjoseph:
@heloiselui I updated the docs, but I'm not seeing a difference in the icon in my preview vs. current storybook. 👀

ariellalgilmore (approved):
LGTM!

heloiselui:
> @heloiselui I updated the docs, but I'm not seeing a difference in the icon in my preview vs. current storybook. 👀

I think I was looking for all bugs possible (which there weren’t any, because you did an excellent work) that I ended up finding one that’s not related to your changes.😂

Thank you for updating the docs!

heloiselui (approved):

PR #20673, opened 2025-10-08
fix(slider): correct keyboard arrow navigation for decimal steps

### LINKED ISSUE #20036: [Bug]: Slider arrow key navigation buggy when using decimal step
### Package

@carbon/react

### Browser

Chrome

### Package version

1.84.0

### React version

18.3.1

### Description

The slider component doesn't work well when the step is a decimal number and the user tries to edit the number using the arrow keys. The arrow keys should work properly even when the step is a not a whole integer.

i believe the issue is because of the `Math.floor` [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Slider/Slider.tsx#L973). The slider works fine for negative numbers because the Math.floor is only being calculated on positive numbers.
```
  calcValueForDelta = (currentValue: number, delta: number, step = 1) => {
    const base =
      delta > 0 ? Math.floor(currentValue / step) * step : currentValue;
    const newValue = base + delta;
    const decimals = (step.toString().split('.')[1] || '').length;

    return Number(newValue.toFixed(decimals));
  };
```

https://github.com/user-attachments/assets/2b7e30f5-1b83-4e59-be62-46c70b14b639

### Reproduction/example

https://stackblitz.com/edit/github-hjyfzayi?file=src%2FApp.jsx

### Steps to reproduce

1. tab over until the slider dot is highlighted
2. press and/or hold left and right arrow keys to move the slider
3. notes:
    a. the slider will usually move in at least one direction for a few steps, but then gets stuck
    b. holding shift to activate the stepMultiplier can get it somewhat "unstuck", but after letting go of shift, it will quick get stuck again and no fine grain control is permitted.
    c. it works perfectly fine for negative numbers

### Suggested Severity

None

### Application/PAL

IBM Cloud - InstructLab

### Code of Conduct

sunny-kukkar-ibm:
Closes https://github.com/carbon-design-system/carbon/issues/20036

Fixed carbon slider issue for decimal steps for keyboard navigation.
root cause:  This is due to how Math.floor works with floating-point division. In calcValueForDelta function, for 0.3/0.1, JavaScript produce 2.9999999999999996, and Math.floor rounds it down to 2 resulting in 0.2 after multiplying by 0.1.
### Changelog

**Changed**

replaced Math.floor with Math.round in calcValueForDelta function. 
Updated the existing test case to validate this scenario.

#### Testing / Reviewing

CI should pass.

## PR Checklist

### DIFF
--- a/packages/react/src/components/Slider/Slider.tsx
+++ b/packages/react/src/components/Slider/Slider.tsx
@@ -1077,7 +1077,7 @@ export const Slider = (props: SliderProps) => {
    */
   const calcValueForDelta = (currentValue: number, delta: number, step = 1) => {
     const base =
-      delta > 0 ? Math.floor(currentValue / step) * step : currentValue;
+      delta > 0 ? Math.round(currentValue / step) * step : currentValue;
     const newValue = base + delta;
     // TODO: Why is the logical OR needed here?
     const decimals = (step.toString().split('.')[1] || '').length;
--- a/packages/react/src/components/Slider/__test__/Slider-test.js
+++ b/packages/react/src/components/Slider/__test__/Slider-test.js
@@ -1240,13 +1240,14 @@ describe('Slider', () => {
         await keyboard('{ArrowRight}');
         await keyboard('{ArrowRight}');
         await keyboard('{ArrowRight}');
+        await keyboard('{ArrowRight}');
 
         // Retrieve the last call to `onChange`.
         const lastCall = onChange.mock.calls[onChange.mock.calls.length - 1][0];
 
         // Assert that the value was updated correctly (i.e., not
         // 0.30000000000000004).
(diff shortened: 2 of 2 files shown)


sunny-kukkar-ibm:

emyarod (approved):
looks good to me!

tay1orjones (approved):
Looks great!

Convention for commits and PR titles is `type(scope): description` where the scope can usually be the impacted component. The remaining portion of the title can consist of a plain description of the work done. No need to include the issue number or duplicate "fix" description. 

Based on this I updated the title of this PR from 

`fix: 20036-carbon-slider decimal steps fix`

to

`fix(slider): correct keyboard arrow navigation for decimal steps`

PR #20690, opened 2025-10-09
fix(pagination): use container query for responsiveness

### LINKED ISSUE #17564: [Bug]: Allow smaller pagination component outside of just narrow screens
### The problem

I'm working on embedding the data-table and pagination component into our product watsonx Assistant chat. The pagination component has a smaller form factor, but it only appears when the screen size is close to mobile size.
Screenshot 2024-09-23 at 3 49 13 PM

For our use case our chat width is often quite narrow so it would be nice to use the smaller form factor of the component, however there is no good way to switch to the smaller form factor when it's only enabled by using a small screen.
Screenshot 2024-09-24 at 4 11 42 PM

(I'm marking this as high priority because we do have an upcoming release - however I can get around this by using my own css to hide parts of the components just like you guys do at this more narrow breakpoint, I just don't want to rely on that solution long term since it's brittle.)

### The solution

Add a prop to the pagination component that can be used to dynamically switch to the smaller form factor. The component could still operate as it does today when this prop is not set in order to not break people who are relying on the screen size breakpoint behavior. With this prop I could make the chat look like this, or if the chat window is larger I could use the original pagination component view:
Screenshot 2024-09-23 at 2 57 58 PM

### Examples

I don't have examples but I'm surprised Carbon is taking this approach here. For most components it seems functionality is exposed through props or breakpoints are made based off the parent components width. Setting break points off the screensize may be unexpected by consumers of Carbon if it's not a common pattern.

### Application/PAL

watsonx Assistant

### Business priority

### Available extra resources

_No response_

### Code of Conduct

annawen1:
Closes #17564

This PR solves an issue with pagination responsiveness where the mobile styles appear based on the viewport size rather than the container the pagination is in. This leads to problems shown below where the container element is smaller than the viewport size and should be displaying the mobile styles but doesn't because the viewport size is not at the mobile breakpoint.

Screenshot 2025-10-09 at 4 23 55 PM

To solve this, `@container` queries are used instead of `@media` queries. The container context is applied to the `cds-pagination` element as that takes the width of whatever parent/container it is placed in.

### Changelog

**Changed**

- use `@container` query instead of `@media` query

#### Testing / Reviewing

- Go to current pagination storybook: https://react.carbondesignsystem.com/?path=/story/components-pagination--default
- go into the inspector and change the width of the pagination container to `500px`
Screenshot 2025-10-10 at 12 19 50 PM
- Go to deploy preview and do the same: https://deploy-preview-20690--v11-carbon-react.netlify.app/?path=/story/components-pagination--default
- It should display the mobile styles 
- do the same for Web components storybook deploy preview to check

## PR Checklist

### DIFF
--- a/packages/styles/scss/components/pagination/_pagination.scss
+++ b/packages/styles/scss/components/pagination/_pagination.scss
@@ -42,6 +42,7 @@
     min-block-size: convert.to-rem(40px);
     overflow-x: auto;
 
+    // 42rem = 'md' breakpoint
     @container pagination (min-width: 42rem) {
       overflow: initial;
 
@@ -186,6 +187,7 @@
   .#{$prefix}--pagination__left {
     padding: 0 $spacing-05 0 0;
 
+    // 42rem = 'md' breakpoint width
     @container pagination (min-width: 42rem) {
       padding: 0 $spacing-05;
     }
--- a/packages/styles/scss/components/pagination/_pagination.scss
+++ b/packages/styles/scss/components/pagination/_pagination.scss
@@ -36,11 +36,13 @@
     justify-content: space-between;
     background-color: $layer;
     border-block-start: 1px solid $border-subtle;
+    container-name: pagination;
+    container-type: inline-size;
     inline-size: calc(100% - 1px);
     min-block-size: convert.to-rem(40px);
     overflow-x: auto;
 
-    @include breakpoint('md') {
+    @container pagination (min-width: 42rem) {
       overflow: initial;
 
       .#{$prefix}--pagination__control-buttons {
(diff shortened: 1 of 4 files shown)


maradwan26 (approved):
LGTM!

packages/styles/scss/components/pagination/_pagination.scss
```diff
@@ -36,11 +36,13 @@
...
     background-color: $layer;
     border-block-start: 1px solid $border-subtle;
+    container-name: pagination;
+    container-type: inline-size;
     inline-size: calc(100% - 1px);
     min-block-size: convert.to-rem(40px);
     overflow-x: auto;
 
-    @include breakpoint('md') {
+    @container pagination (min-width: 42rem) {
```
kennylam, line 46: I wonder if it's worth adding a comment at the first instance of these values to describe what 42rem is.

annawen1: done, I added the comment here and there

kennylam (approved):

PR #20707, opened 2025-10-13
fix: scope ai-label to the slot

devadula-nandan:
Closes #

this pr scopes ai-label to the assigned slot
it was observed that the previous logic and css effects every `cds-ai-label` placed inside the tile and we lose the flexibility to detach the slot.

### Changelog

**Changed**

- update scopes of the logics and css

#### Testing / Reviewing

try to remove slot="ai-label" and notice you have freedom to use ai-label wherever in the tile. just like a regular element

and this is what this pr does

https://github.com/user-attachments/assets/9638f8aa-cd5f-4d3d-9661-4e49112efdb9

## PR Checklist

### DIFF
--- a/packages/web-components/src/components/tile/tile.ts
+++ b/packages/web-components/src/components/tile/tile.ts
@@ -30,22 +30,19 @@ class CDSTile extends LitElement {
    * Handles `slotchange` event.
    */
   protected _handleSlotChange({ target }: Event) {
-    const hasContent = (target as HTMLSlotElement)
+    const hasSlottedAiLabel = (target as HTMLSlotElement)
       .assignedNodes()
-      .filter((elem) =>
-        (elem as HTMLElement).matches !== undefined
-          ? (elem as HTMLElement).matches(
-              (this.constructor as typeof CDSTile).aiLabelItem
-            ) ||
-            // remove reference of slug in v12
-            (elem as HTMLElement).matches(
-              (this.constructor as typeof CDSTile).slugItem
-            )
-          : false
-      );
-    if (hasContent.length > 0) {
-      this._hasAILabel = Boolean(hasContent);
-      (hasContent[0] as HTMLElement).setAttribute('size', 'xs');
+      .filter((elem) => {
+        if (!(elem instanceof HTMLElement)) {
+          return false;
+        }
+        const slot = elem.getAttribute('slot');
+        return slot === 'ai-label' || slot === 'slug';
+      });
+
+    if (hasSlottedAiLabel.length > 0) {
+      this._hasAILabel = Boolean(hasSlottedAiLabel);
+      (hasSlottedAiLabel[0] as HTMLElement).setAttribute('size', 'xs');
     }
     this.requestUpdate();
   }
--- a/packages/web-components/src/components/tile/tile.ts
+++ b/packages/web-components/src/components/tile/tile.ts
@@ -30,19 +30,22 @@ class CDSTile extends LitElement {
    * Handles `slotchange` event.
    */
   protected _handleSlotChange({ target }: Event) {
-    const hasSlottedAiLabel = (target as HTMLSlotElement)
+    const hasContent = (target as HTMLSlotElement)
       .assignedNodes()
-      .filter((elem) => {
-        if (!(elem instanceof HTMLElement)) {
-          return false;
-        }
-        const slot = elem.getAttribute('slot');
-        return slot === 'ai-label' || slot === 'slug';
-      });
-
-    if (hasSlottedAiLabel.length > 0) {
-      this._hasAILabel = Boolean(hasSlottedAiLabel);
-      (hasSlottedAiLabel[0] as HTMLElement).setAttribute('size', 'xs');
+      .filter((elem) =>
+        (elem as HTMLElement).matches !== undefined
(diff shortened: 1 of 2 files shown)


devadula-nandan:
we need the the ai-label to be mini and and xs, need it at a different place in the header etc.
and i was expecting if we don't pass slot="decorator" it becomes a regular component which we can render anywhere, but looks like this was a mistake in the styles and the logic. so here's the fix.

ariellalgilmore:
> we need the the ai-label to be mini and and xs, need it at a different place in the header etc. and i was expecting if we don't pass slot="decorator" it becomes a regular component which we can render anywhere, but looks like this was a mistake in the styles and the logic. so here's the fix.

I believe we can just do this when applying the `cds-ai-label` inside `cds-tile`:

`<cds-ai-label alignment="bottom-left" slot="">`. 

This way it doesn't get added into the `ai-label`/`slug`/ `decorator` slots

devadula-nandan:
> > we need the the ai-label to be mini and and xs, need it at a different place in the header etc. and i was expecting if we don't pass slot="decorator" it becomes a regular component which we can render anywhere, but looks like this was a mistake in the styles and the logic. so here's the fix.
> 
> I believe we can just do this when applying the `cds-ai-label` inside `cds-tile`:
> 
> `<cds-ai-label alignment="bottom-left" slot="">`.
> 
> This way it doesn't get added into the `ai-label`/`slug`/ `decorator` slots

thats the expectation which was not met

https://github.com/user-attachments/assets/540a78c0-0192-4140-bca3-f0760fc6c4cf

devadula-nandan:
https://github.com/user-attachments/assets/89d51ac3-8f97-46ff-9c38-cadae5925546

@ariellalgilmore if u look at the constructor it queries ai label but not the slotted one, which is the reason why it fails to detach from position and size

packages/web-components/src/components/tile/tile.ts
```diff
@@ -30,22 +30,19 @@ class CDSTile extends LitElement {
...
+        if (!(elem instanceof HTMLElement)) {
+          return false;
+        }
+        const slot = elem.getAttribute('slot');
+        return slot === 'ai-label' || slot === 'slug';
+      });
+
+    if (hasSlottedAiLabel.length > 0) {
+      this._hasAILabel = Boolean(hasSlottedAiLabel);
+      (hasSlottedAiLabel[0] as HTMLElement).setAttribute('size', 'xs');
```
ariellalgilmore: I think if we add the `slot=""` and just keep the styles you changed without changing the `ts` file it would fix the solution. The `handleSlotChange` function should only be updating elements within the `slug`/`ai-label`/`decorator` slot. When we add `slot=""` then the `handleSlotChange` would not be called on that `cds-ai-label`. The `cds-ai-label` default is `xs` so you won't see a size change unless you add a size.

```
<cds-ai-label alignment="bottom-left" slot="" size="sm">
            ${content}${actions}</cds-ai-label
          >
          

```

Screenshot 2025-10-13 at 8 06 43 AM

devadula-nandan: https://github.com/carbon-design-system/carbon/pull/20707#issuecomment-3397898263

devadula-nandan: the styles are for position
the ts updates are for size

ariellalgilmore: yes, but I am able to do the above size updates without any changes to the `tile.ts`. Can you try removing and updating the `cds-ai-label` to be like this:

```
<cds-ai-label alignment="bottom-left" slot="" size="sm">
            ${content}${actions}</cds-ai-label
          >

devadula-nandan: ill get back with a stackblitz. let me check it in an example.

devadula-nandan: https://stackblitz.com/edit/vitejs-vite-gz8pjxwd?file=package.json,index.html,src%2Fexample.ts%3AL35

devadula-nandan: i see if we add slot="" then it gets detached. which still looks wrong to me.

devadula-nandan: do u want me to go ahead and revert the ts changes?

ariellalgilmore: yes please revert the changes. thank you!

annawen1: Let's go ahead and revert the ts changes. Using the `slot=""` to solve this edge case issue is fine with me.

devadula-nandan: I've reverted the TS changes as per the suggestions.

Here’s a quick note on why I think passing `slot=""` to detach might be a bit confusing.
An average consumer may not realize this behavior unless it’s explicitly documented, and could wonder why the size doesn’t update after removing the slot attribute. assuming it was detached.

I’m curious if this is a convention we specifically follow in Carbon and I’m just missing it?
Eager to know the intention behind it.

Anyway, that’s my take on it!

ariellalgilmore (approved):
LGTM!

maradwan26 (approved):
LGTM !

PR #20712, opened 2025-10-13
MenuButton dropdown remains fixed when parent container scrolls, causing usability issues

### LINKED ISSUE #20693: [Bug]: MenuButton dropdown remains fixed when parent container scrolls, causing usability issues
### Package

@carbon/react

### Browser

Chrome

### Package version

1.92.1

### React version

18.3.1

### Description

When a MenuButton component is placed inside a scrollable container, the dropdown menu doesn't move with its parent container during scrolling. Instead, it remains fixed in its original screen position, causing a disconnect between the button and its dropdown menu. This creates a poor user experience as users scroll and find the menu is no longer visually connected to its trigger button.

**Video**

https://github.com/user-attachments/assets/2d770df3-19a4-44b8-8859-79e50914875e

**Potential solution**
We probably could expose the boundary option for floating-ui like we have in Popover. With that you'd be able to pass in a ref of the container. It would use that container instead of the viewport to calculate when to reflow and change alignment to "up" when you scroll.

### Reproduction/example

https://stackblitz.com/edit/f4oagpin?file=src%2FApp.jsx%3AL2,src%2Findex.scss

### Steps to reproduce

1. Place a MenuButton component inside a scrollable container and sufficient height to enable scrolling
2. Open the MenuButton dropdown
3. Scroll the container

**Actual Behaviour**
The dropdown menu remains fixed in its original screen position while the button moves with the scroll, causing them to become disconnected.

**Expected Behaviour**
The dropdown menu should maintain its position relative to the button as the container scrolls, staying visually connected to its trigger button.

### Suggested Severity

None

### Application/PAL

_No response_

### Code of Conduct

mariyageorge01:
Closes #20693 

When a MenuButton is rendered inside a scrollable container, the menu items doesn't maintain proper positioning relative to the button during scrolling. This occurs because the menu uses fixed positioning which disconnects it from its parent's scroll context, causing the menu to remain in its original position while the button moves with the scroll.

Solution:
The solution makes use of floating-ui's flip and hide middlewares to handle positioning and visibility. When scrolling occurs, the menu now stays properly aligned with the button, flips to maintain visibility when approaching container edges, hides when it would be clipped by the container and updates CSS classes to reflect the current placement.

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

## PR Checklist

### DIFF
--- a/packages/react/src/components/MenuButton/index.tsx
+++ b/packages/react/src/components/MenuButton/index.tsx
@@ -27,7 +27,6 @@ import {
   flip,
   size as floatingSize,
   autoUpdate,
-  type Boundary,
   type Placement,
   hide,
 } from '@floating-ui/react';
@@ -89,12 +88,6 @@ export interface MenuButtonProps extends ComponentProps<'div'> {
    * Specify a DOM node where the Menu should be rendered in. Defaults to document.body.
    */
   menuTarget?: Element;
-
-  /**
-   * Specify a bounding element to be used for calculating when to reflow and change alignment.
-   * The viewport is used by default. This allows the menu to stay within a scrollable container.
-   */
-  boundary?: Boundary;
 }
 
 // eslint-disable-next-line react/display-name -- https://github.com/carbon-design-system/carbon/issues/20452
@@ -110,7 +103,6 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
       menuAlignment = 'bottom',
       tabIndex = 0,
       menuTarget,
-      boundary,
       ...rest
     },
     forwardRef
@@ -145,12 +137,12 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
       fallbackPlacements: getFallbackPlacements(),
       fallbackStrategy: 'initialPlacement' as const,
       fallbackAxisSideDirection: 'start' as const,
-      boundary,
+      boundary: menuTarget,
       crossAxis,
     });
 
     if (!enableOnlyFloatingStyles) {
-      if (boundary) {
+      if (menuTarget) {
         middlewares.push(flip(getEnhancedFlipConfig(false)));
       } else {
         middlewares.push(
@@ -161,12 +153,12 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
       }
     }
 
-    if (boundary && enableOnlyFloatingStyles) {
+    if (menuTarget && enableOnlyFloatingStyles) {
       middlewares.push(flip(getEnhancedFlipConfig()));
     }
 
-    if (boundary) {
-      middlewares.push(hide({ boundary }));
+    if (menuTarget) {
+      middlewares.push(hide({ boundary: menuTarget }));
     }
 
     if (menuAlignment === 'bottom' || menuAlignment === 'top') {
@@ -381,22 +373,6 @@ MenuButton.propTypes = {
   menuTarget: PropTypes.instanceOf(
     typeof Element !== 'undefined' ? Element : Object
   ) as PropTypes.Validator<Element | null | undefined>,
-
-  /**
-   * Specify a bounding element to be used for calculating when to reflow and change alignment.
-   * The viewport is used by default. This allows the menu to stay within a scrollable container.
-   */
-  boundary: PropTypes.oneOfType([
-    PropTypes.oneOf(['clippingAncestors']),
-    PropTypes.elementType,
-    PropTypes.arrayOf(PropTypes.elementType),
-    PropTypes.exact({
-      x: PropTypes.number.isRequired,
-      y: PropTypes.number.isRequired,
-      width: PropTypes.number.isRequired,
-      height: PropTypes.number.isRequired,
-    }),
-  ]) as PropTypes.Validator<Boundary | null | undefined>,
 };
 
 export { MenuButton };
--- a/packages/react/src/components/MenuButton/index.tsx
+++ b/packages/react/src/components/MenuButton/index.tsx
@@ -27,6 +27,9 @@ import {
   flip,
   size as floatingSize,
   autoUpdate,
+  type Boundary,
+  type Placement,
+  hide,
 } from '@floating-ui/react';
 import { useFeatureFlag } from '../FeatureFlags';
 import { mergeRefs } from '../../tools/mergeRefs';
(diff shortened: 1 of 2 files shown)


packages/react/src/components/MenuButton/index.tsx
```diff
@@ -115,10 +117,48 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
     const prefix = usePrefix();
     const triggerRef = useRef<HTMLDivElement>(null);
     // eslint-disable-next-line @typescript-eslint/no-explicit-any -- https://github.com/carbon-design-system/carbon/issues/20452
-    let middlewares: any[] = [];
+    const middlewares: any[] = [];
```
tay1orjones, line 120: This entire middleware setup needs to be refactored. Instead of an empty array and imperatively `push()`ing options, it should be more declarative like all the other usages of floating-ui.

```diff
- middlewares.push(middleware1());
- if (aCondition) {
- 
-  // what if this actually needs to come before middleware1, for instance?
-  middlewares.push(middleware2()); 
- 
- }

+ middlewares: [
+  middleware1(),
+  aCondition && middleware2(),
+ ]
```

This way we're not running the risk of something being `push` ed too late impacting the order of the middleware's, which does matter.

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -115,10 +117,48 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
...
+      if (menuAlignment.includes('bottom')) {
+        return ['top', 'top-start', 'top-end'] as Placement[];
+      } else if (menuAlignment.includes('top')) {
+        return ['bottom', 'bottom-start', 'bottom-end'] as Placement[];
+      } else if (menuAlignment.includes('left')) {
+        return ['right', 'right-start', 'right-end'] as Placement[];
+      } else {
+        return ['left', 'left-start', 'left-end'] as Placement[];
+      }
+    };
```
tay1orjones, line 133: This can be much simpler, IIRC it's the same logic as OverflowMenu. Also `left` isn't an allowed option for menuAlignment.

https://github.com/carbon-design-system/carbon/blob/7a6088cde23048f072865639fd8b8f7cd8b11056/packages/react/src/components/OverflowMenu/next/index.tsx#L116-L123

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -115,10 +117,48 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
...
+
+    // Helper function to create enhanced flip middleware config
+    const getEnhancedFlipConfig = (crossAxis = true) => ({
+      fallbackPlacements: getFallbackPlacements(),
+      fallbackStrategy: 'initialPlacement' as const,
+      fallbackAxisSideDirection: 'start' as const,
+      boundary: menuTarget,
+      crossAxis,
+    });
```
tay1orjones, line 142: Same thing here, this should go directly into `middleware`. It should eliminate most/all of the typescript `as const` type stuff.

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -115,10 +117,48 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
...
+        );
+      }
+    }
+
+    if (menuTarget && enableOnlyFloatingStyles) {
+      middlewares.push(flip(getEnhancedFlipConfig()));
+    }
+
+    if (menuTarget) {
+      middlewares.push(hide({ boundary: menuTarget }));
```
tay1orjones, line 161: Also collapse into `middleware`

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -178,7 +218,42 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
...
+        refs.floating.current.classList.add(`${prefix}--menu--${placement}`);
+      }
+    }, [
+      floatingStyles,
+      refs.floating,
+      middlewareData,
+      placement,
+      open,
+      prefix,
+    ]);
```
tay1orjones, line 256: This isn't needed, all styles are applied via CSSOM instead of by classnames.

https://github.com/carbon-design-system/carbon/blob/20b4691222ebde34ae61de44cd3e029e4a4dc42d/packages/react/src/components/MenuButton/index.tsx#L167-L180

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -178,7 +218,42 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
...
         }
       });
-    }, [floatingStyles, refs.floating, middlewareData, placement, open]);
+
+      if (refs.floating.current && middlewareData.hide) {
+        refs.floating.current.style.visibility = middlewareData.hide
+          .referenceHidden
+          ? 'hidden'
+          : 'visible';
+      }
```
tay1orjones, line 227: This should be moved to live alongside where the styles are applied via CSSOM. This is how it's done in popover and most/all other components that utilize the `hide` middleware data provider

https://github.com/carbon-design-system/carbon/blob/053d1000b9ef2720f233aa0d8e82ce9ef792433a/packages/react/src/components/Popover/index.tsx#L360-L368

packages/react/src/components/MenuButton/index.tsx
```diff
@@ -115,10 +117,48 @@ const MenuButton = forwardRef<HTMLDivElement, MenuButtonProps>(
...
+        return ['left', 'left-start', 'left-end'] as Placement[];
+      }
+    };
+
+    // Helper function to create enhanced flip middleware config
+    const getEnhancedFlipConfig = (crossAxis = true) => ({
+      fallbackPlacements: getFallbackPlacements(),
+      fallbackStrategy: 'initialPlacement' as const,
+      fallbackAxisSideDirection: 'start' as const,
+      boundary: menuTarget,
```
tay1orjones, line 140: I don't think we should reuse `menuTarget` for this. It's using a portal under the hood in `Menu` which eventually is going to be removed. #16883 

It could also be useful for the boundary to be able to be something other than the target the menu is rendered into. I'm thinking like a scrollable region that wraps a content div where the menu is rendered or something. 

So it should be a separate prop, like in Popover. The name `autoAlignBoundary` still feels applicable for the prop, even though this component doesn't have an `autoAlign` prop.

tay1orjones (changes requested):
Thanks for diving into this! Functionally it looks okay I think. My comments below are mostly organizational fixes to align things with how the [other components that use `hide`](https://github.com/search?q=repo%3Acarbon-design-system%2Fcarbon%20hide%28&type=code) are doing things.

tay1orjones:
Hey @mariyageorge01, just checking in. Is this something you want to keep working on?

mariyageorge01:
I’d really love to contribute to this, but my current bandwidth is limited. I’m not sure when I’ll be able to resume work on it 🥺. 

I’m absolutely fine with someone else taking this up. Otherwise, I’ll resume work on it when my bandwidth allows 🙂.

heloiselui:
Hi, thanks for the effort on this so far! Since it’s been inactive for a bit, I’m going to close this for now. If you'd like to pick it back up, feel free to reopen. Thanks!

PR #20738, opened 2025-10-15
fix(FilterableMultiSelect): no aria-describedby helperText when invalid

### LINKED ISSUE #19644: [a11y]: 4.1.2 Name, Role, Value Violation from <FilterableMultiSelect /> aria-describedby label
### Package

@carbon/react

### Browser

Chrome

### Operating System

MacOS

### Package version

1.78.0

### React version

18.3.1

IBM Equal Access Accessibility Checker

### Assistive technology

_No response_

### Description

The issue we are experiencing is a 4.1.2 Name, Role, Value violation on our input element. This issue is happening because we are using the `helperText` prop on the `<FilterableMultiSelect />` component and that text when it is not present on the screen (it is not present for us when our input is invalid) hits the `helperId` and that will set the `helperId` to `filterablemultiselect-helper-text-${filterableMultiSelectInstanceId}`. This then causes issues because the `aria-describedby` is set to the `helperId` when `helperText` is `true`. Then when the `helperText` `<div>` disappears on our screen, the error message appears but the `filterablemultiselect-helper-text-${filterableMultiSelectInstanceId}` is not equal to the `id` of the `helperText` `<div>`

Image

Image

Image
Image
Image
Image

### WCAG 2.1 Violation

https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html

### Reproduction/example

https://ecasr6c4--run.stackblitz.io/

### Steps to reproduce

1.) Go to the link above: https://stackblitz.com/run?file=src%2FApp.jsx

2.) Open up the IBM Equal Access Accessibility Checker chrome tool

3.) Make sure the input is invalid and look for the 4.1.2 accessibility violation.

Image

### Suggested Severity

### Code of Conduct

tay1orjones:
Closes #19644

Small update to not place `aria-describedby` w/ the helperText id when the input is `invalid` and helperText is not shown/rendered.

### Changelog

**Changed**

- update filterablemultiselect aria-describedby logic w/ a new test

#### Testing / Reviewing

- Go to the ?path=/story/components-multiselect--filterable-with-select-all&args=helperText:something;invalid:!true story
- Make sure that the helperText prop is set to something and `invalid` is `true`
- Open the story in the popout iframe and run ac checker
- You should see no violations

## PR Checklist


kennylam (approved):

heloiselui (approved):
Tested and confirmed, the PR resolves this issue. 

I noticed we have the same problem when `warn` is true: `"The 'id' 'filterablemultiselect-helper-text-id-«r4»' specified for the ARIA property 'aria-describedby' value is not valid."` - not sure if you want open a new issue for that.

I also noticed that in the `Default` example, in controls, both `invalidText` and `warnText` use `"Whoopsie!"` as sample content. I think it would be nice to include that in all the other story examples as well, and add something similar for `helperText`, but just thinking out loud.

tay1orjones:
@heloiselui great catch, thanks! I updated to fix the `warn` bit

PR #20767, opened 2025-10-20
chore(deps): update lerna to v9

tay1orjones:
Replaces https://github.com/carbon-design-system/carbon/pull/20740 that was opened by renovate

This updates the project to lerna v9.

<details><summary>The info within this expansion is outdated, there is an upstream bug in nx blocking this PR, see my comment below</summary>
<p>

I originally had issues getting this to work with `typescript-config-carbon`. Lerna delegates `lerna run` orchestration to Nx. Nx seems to only creates task edges for workspace dependencies that it can prove live in the same repo.

`typescript-config-carbon`'s first build step is to delete the previous build artifacts. Packages depending on `typescript-config-carbon` were being built immediately due to lerna not knowing this is a workspace dependency. They would error because they expected the built files in `typescript-config-carbon` but they weren't there yet, `typescript-config-carbon` hadn't finished building.

To fix, I updated the versions to use the workspace protocol with `workspace:^0.8.0`. I ran `yarn lerna version minor --no-git-tag-version --no-push --yes` and verified that when we version the packages it will update these to `workspace:^0.8.0`. When we publish, the `workspace` protocol identifier should be removed by `lerna publish`. Final packages should have a normal `"typescript-config-carbon": "^0.8.0"` entry in their package.json.

</p>
</details> 

### Changelog

**Changed**

- updated lerna dep to v9
- updated `typescript-config-carbon` versions to use the workspace protocol

#### Testing / Reviewing

- You can test by pulling this down and run `yarn lerna version minor --no-git-tag-version --no-push --yes`
- Look at package.json files - the `workspace:` protocol entries should be increased by one minor version.

## PR Checklist

~- [ ] Tested for cross-browser consistency~


tay1orjones:
CI is failing due to an upstream bug in nx that is fixed but lerna hasn't been updated to use newer versions of nx in their allowed semver ranges. So this is blocked by https://github.com/lerna/lerna/pull/4242

kennylam:
@tay1orjones Bumping to v9.0.1 didn't resolve the build/ci issues. However, adding a `nx.json` file with a `dependsOn: build` seems to have fixed it.

nx.json
tay1orjones, line 1: @kennylam this LGTM 👍

tay1orjones (commented):
I can't approve my own PR but this looks good to me @kennylam @alisonjoseph @sunny-kukkar-ibm

ariellalgilmore (approved):
LGTM!

kennylam (approved):
LGTM!

PR #20780, opened 2025-10-22
feat(data-table): parity - make data-table header sticky

### LINKED ISSUE #17735: [Feature Request]: Allow table header to be sticky when table cells overflow
### The problem

On small form factors data tables sometimes need a horizontal scroll. When you have a table with buttons on the right side of the header these get lost in the overflow.

### The solution

A css solution is to use position: sticky with a width equal to the width of the parent container.

```
cds-table-header-title,
cds-table-header-description,
cds-table-toolbar {
  position: sticky;
  width: var(--cds-chat-table-width, auto);
  left: 0;
}
```

https://github.com/user-attachments/assets/5c4afc9b-798a-4767-b8a8-a16d6df55af8

This same solution could be implemented in carbon as a "stickyHeaderWidth" prop where the width can be passed to the web component.

I've also needed to implement styling to remove the padding from the `cds--data-table-header` class which is within a shadow root and can't be accessed to style. The workaround I came up with is setting negative margins on the header-title and header-description components to offset that padding, however in an ideal world this would not be necessary in this stickyHeader scenario.

```
// The title and description have a parent div around them with padding (cds--data-table-header) that causes the title
// and description to scroll horizontally even when those items have a fixed width (equal to their container size). If
// this was a React component we could just override the css, however because the container div that is setting this
// padding is in a shadow root, we can't select the div to override the styles. In order to remove this scroll the
// padding needs to be offset from the parent div using this negative margin "hack". Then the padding can be reapplied
// to the component we control so the title is properly indented but doesn't scroll.
cds-table-header-title,
cds-table-header-description {
  width: calc(var(--cds-chat-table-width, auto) - layout.$spacing-05 - layout.$spacing-05);
  margin-left: -#{layout.$spacing-05};
  margin-right: -#{layout.$spacing-05};
  padding: 0 layout.$spacing-05;
}
```

Note the calc on the width is to account for the new padding that is being added to these components.

### Examples

Sticky headers are a common pattern - the data table component just expects a much larger form factor at this time.

### Application/PAL

watsonx Assistant

### Business priority

### Available extra resources

_No response_

### Code of Conduct

annawen1:
Closes #17735 

This PR aligns the Web Component `cds-data-table` with the React version default behavior where the table header respects the parent width and remains sticky. The current implementation overflows out of the parent width and requires the table header to be scrolled to view the batch actions when in smaller viewport.

https://github.com/user-attachments/assets/2bf465b9-0daf-4d35-82fc-a5b1c7011d65

### Changelog

**Changed**

- switch the `display: table` attribute to the `data-table-content` element instead this is in-line with what is done in React
- ensure that the slotted `cds-table-head` and `cds-table-body` components are wrapped in parent divs to allow for the styling needed

#### Testing / Reviewing

Compare the current WC vs. React data table in small view port, then check the deploy preview - the header behavior should align 

## PR Checklist

### DIFF
--- a/packages/web-components/src/components/data-table/_table-action.scss
+++ b/packages/web-components/src/components/data-table/_table-action.scss
@@ -149,6 +149,7 @@
   @extend .#{$prefix}--batch-actions;
 
   box-sizing: border-box;
+  overflow-x: auto;
 }
 
 :host(#{$prefix}-table-batch-actions[active]) {
--- a/packages/web-components/src/components/data-table/_table-core.scss
+++ b/packages/web-components/src/components/data-table/_table-core.scss
@@ -12,8 +12,6 @@
 :host(#{$prefix}-table) {
   @extend .#{$prefix}--data-table;
 
-  display: table;
-
   ::slotted(#{$prefix}-table-head) {
     @include type-style('heading-01');
 
@@ -36,6 +34,11 @@
   inline-size: auto;
 }
 
+.#{$prefix}--data-table-content {
+  display: table;
+  inline-size: 100%;
+}
+
 :host(#{$prefix}-table[sticky-header]) {
   ::slotted(#{$prefix}-table-head),
   ::slotted(#{$prefix}-table-body) {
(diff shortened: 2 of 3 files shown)


Gururajj77 (approved):
Looks good 🚀

heloiselui (approved):
LGTM

riddhybansal (approved):

PR #20781, opened 2025-10-22
fix(slider): Disallow invalid & warn when readonly or disabled

### LINKED ISSUE #20731: [Slider]: Disallow invalid & warn when readonly or disabled

BiaBiruka:
Closes #20731

Fix the behavior for Slider and Two Handle Slider. When component is disabled or readOnly, it should not show waning nor invalid message.

### Changelog

**Changed**

- Instead of using readOnly state to check if should display warnings or not, Slider now also checks if component is disabled. This way, warnings will not be shown if the component is disabled/readOnly.

#### Testing / Reviewing

- Access the Slider playground on storybook (/?path=/story/components-slider--default) and change the state to "Invalid = true" and "disabled = true". The invalid message should not be shown.
- Repeat above test with "readOnly = true" instead of disabled. The invalid message should not be shown.
- Repeat the two tests above using "warn = true" instead of "invalid = true". The warning message on both scenarios should not be shown.

## PR Checklist

~Updated documentation and storybook examples~

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1546,6 +1546,13 @@
      "contributions": [
        "code"
      ]
+    },
+    {
+      "login": "BiaBiruka",
+      "name": "Beatriz Parolin",
+      "avatar_url": "https://avatars.githubusercontent.com/u/128196689?v=4",
+      "profile": "https://github.com/BiaBiruka",
+      "contributions": ["code"]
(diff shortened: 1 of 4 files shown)


BiaBiruka:

BiaBiruka:
recheck

PR #20808, opened 2025-10-26
fix(select): disallow invalid and warn when readonly or disabled

### LINKED ISSUE #20725: [Select]: Disallow invalid & warn when readonly or disabled

Mariam-Saeed:
Closes #20725 

Disallow invalid and warn when readonly or disabled in Select

### Changelog

**New**

- Added unit tests to confirm that the component correctly disallows both warnings and invalid states from showing up when `disabled` or `readOnly` are `true`.

**Changed**

- The React `Select` components now use the `useNormalizedProps` hook. This ensures all components apply a consistent logic to disallow `invalid` and `warn` states from appearing when `disabled` or `readonly` are `true`.
- The Web Component (WC) implementation was updated to include `normalizedProps`. This aligns the WC with the React library's prop normalization logic, specifically to disallow `invalid` and `warn` states when `disabled` or `readOnly` are `true`.

**Removed**

- {{removed thing}}

##### React Library

- Open the React Library Storybook and navigate to the `Select` component stories.
- **Test Disabled State:** Set the `disabled` prop to `true`. While `disabled` is true, set the `invalid` prop to `true` and confirm that the invalid styles/messages do not show up. Apply the same thing with the `warn` prop. 
- **Test ReadOnly State**: Set the `readOnly` prop to `true`. While `readOnly` is true, set the `invalid` prop to `true` and confirm that the invalid styles/messages do not show up. Apply the same thing with the `warn` prop. 

##### Web Component (WC)

- Open the WC Storybook and navigate to the `Select` component stories.
- **Test Disabled State:** Set the `disabled` attribute to `true`. While `disabled` is true, set the `invalid` attribute to `true` and confirm that the invalid styles/messages do not show up. Apply the same thing with the `warn` attribute. 
- **Test ReadOnly State**: Set the `readOnly` attribute to `true`. While `readOnly` is true, set the `invalid` attribute to `true` and confirm that the invalid styles/messages do not show up. Apply the same thing with the `warn` attribute. 

**Verify that the unit tests are passing, which confirms the correct behavior for `disabled` and `readOnly`.**

## PR Checklist


Mariam-Saeed:
**recheck**

packages/react/src/components/Select/Select.tsx
```diff
@@ -200,60 +199,56 @@ const Select = React.forwardRef(
...
       [`${prefix}--select--inline`]: inline,
       [`${prefix}--select--light`]: light,
-      [`${prefix}--select--invalid`]: invalid,
-      [`${prefix}--select--disabled`]: disabled,
+      [`${prefix}--select--invalid`]: normalizedProps.invalid,
+      [`${prefix}--select--disabled`]: normalizedProps.disabled,
       [`${prefix}--select--readonly`]: readOnly,
-      [`${prefix}--select--warning`]: warn,
+      [`${prefix}--select--warning`]: normalizedProps.warn,
       [`${prefix}--select--fluid--invalid`]: isFluid && invalid,
```
riddhybansal:
```suggestion
       [`${prefix}--select--fluid--invalid`]: isFluid && normalizedProps.invalid,
```

riddhybansal (changes requested):

riddhybansal:
Thanks for contributing just the small update above!!

.github/workflows/ci.yml
```diff
@@ -371,7 +371,7 @@ jobs:
           retention-days: 14
 
       - name: Download avt blob reports from GitHub Actions Artifacts
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
```
tay1orjones, line 374: ~I think this change should be made in a separate PR. It should use a SHA instead of the version tag. Also a comment on the end noting the version the SHA corresponds to. Dependabot/renovate will keep both up to date over time. There's examples of this elsewhere in this file.~

tay1orjones: Actually it looks like this was already done and this PR just needs updated with the latest from `main`

https://github.com/carbon-design-system/carbon/pull/20812

riddhybansal (approved):
LGTM, THANKS

Gururajj77 (approved):

PR #20818, opened 2025-10-27
Fix(time-picker): prevent invalid/warn states when disabled/readonly

### LINKED ISSUE #20727: [TimePicker]: Disallow invalid & warn when readonly or disabled

anuanto966:
Fixed Timepicker components in both React and Web Components to not show invalid/warn states when in disabled or readonly state, as users cannot interact with these components in those states.

Closes #20727 

### Changelog

**Changed**

- Instead of using readOnly state to check if should display warnings or not, Time Picker now also checks if component is disabled. This way, warnings will not be shown if the component is disabled/readOnly.

#### Testing / Reviewing
- Access the Slider playground on storybook (/?path=/story/components-time-picker--default) and change the state to "Invalid = true" and "disabled = true". The invalid message should not be shown.
- Repeat above test with "readOnly = true" instead of disabled. The invalid message should not be shown.
- Repeat the two tests above using "warn = true" instead of "invalid = true". The warning message on both scenarios should not be shown.

## PR Checklist


riddhybansal (changes requested):
Consider using useNormalizedInputProps in all components rather than creating separate validation logic:

normalizedProps = useNormalizedInputProps({
id, disabled, readOnly, invalid, invalidText, warn, warnText,
});

Then reference normalizedProps.invalid and normalizedProps.warn for validation states
Our input controls already use this pattern, which makes future fixes simpler (just update the hook once).

anuanto966:
> Consider using useNormalizedInputProps in all components rather than creating separate validation logic:
@riddhybansal This change is addressed now.

riddhybansal (approved):

riddhybansal (approved):
Thanks for doing this !! This looks perfect now.

alisonjoseph (approved):

PR #20823, opened 2025-10-27
fix(popover): update specificity for border and background-token

alisonjoseph:
Over in Carbon Labs we have Tooltips inside of Popovers and I noticed the new border/backgroundToken prop styles were bleeding through to the nested tooltip styles. This PR updates the styles to specifically only target the child so nested Tooltips keep their correct styling based on their own props. 

Example of what is happening over in Labs
Screenshot 2025-10-27 at 1 35 54 PM

> [!IMPORTANT]
> Test story will need to be deleted before merge 

### Changelog

**Changed**

- update border and backgroundToken styles for Popover to be more specific

## PR Checklist

### DIFF
--- a/packages/react/src/components/Popover/Popover.stories.js
+++ b/packages/react/src/components/Popover/Popover.stories.js
@@ -17,7 +17,6 @@ import { Settings } from '@carbon/icons-react';
 import { keys, match } from '../../internal/keyboard';
 import OverflowMenu from '../OverflowMenu/OverflowMenu';
 import OverflowMenuItem from '../OverflowMenuItem';
-import { Tooltip } from '../Tooltip';
 
 const prefix = 'cds';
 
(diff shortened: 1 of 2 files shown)


riddhybansal (approved):
LGTM 💯

packages/react/src/components/Popover/Popover.stories.js
```diff
@@ -17,6 +17,7 @@ import { Settings } from '@carbon/icons-react';
 import { keys, match } from '../../internal/keyboard';
 import OverflowMenu from '../OverflowMenu/OverflowMenu';
 import OverflowMenuItem from '../OverflowMenuItem';
+import { Tooltip } from '../Tooltip';
```
tay1orjones:
```suggestion
```

packages/react/src/components/Popover/Popover.stories.js
```diff
@@ -397,3 +398,22 @@ export const TabTipExperimentalAutoAlign = () => {
...
+          </p>
+
+          <Tooltip label="Close" open>
+            <button type="button">Tooltip test</button>
+          </Tooltip>
+        </PopoverContent>
+      </Popover>
+    </div>
+  );
+};
```
tay1orjones:
```suggestion
```

tay1orjones (approved):

